PR #190, opened 2014-02-23
Columninfo methods for schema fetching

johanneslumpe:
I rewrote the methods so they can be used with the query chain object, like this:

``` javascript
knex('tableName').columnInfo().then(function (columns) {
...
});
```

Right now the `columnInfo` method automatically terminates the query and calls `then` on the objects, forcing it to issue the query, because the response needs to be parsed first before being returned to the user. I'm sure you will want to change some things around, but the methods are working.

Now that I think of it, it would be nice to have a way, to format the response before it's being passed back to the user without force-calling `then`, since that actually breaks the ability to just call `toSql` on the object... Anyway, will push this now and then we can talk about it.

### DIFF
--- a/lib/clients/sqlite3/query/compiler.js
+++ b/lib/clients/sqlite3/query/compiler.js
@@ -1,6 +1,7 @@
 // Extend the base compiler with the necessary grammar
 module.exports = function(client) {
 
+  var _ = require('lodash');
   var QueryCompiler = require('../../../query/compiler')(client);
 
   return QueryCompiler.extend({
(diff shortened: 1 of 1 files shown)


souravsaraf:
Hi, 

knex('tableName').columnInfo()
does this one has a asCallback() ?

kibertoad:
@souravsaraf `asCallback` is a bluebird feature, and we are moving away from it, so I wouldn't recommend you to rely on it.
Why could you possibly want callback instead of async/await invocation, though?

PR #253, opened 2014-04-23
andWhereRaw method for clarity

AntJanus:
I noticed that there is a `orWhereRaw` method as well as an `andWhere` method (which has identical usage as a simple `where`) so I thought that for clarity's sake, it would make sense to add an `andWhereRaw` method.

Otherwise, it seems the naming conventions seem a little inconsistent.


willfarrell:
For other searching for this:

whereRaw, orWhereRaw, havingRaw, orHavingRaw are now just .where(knex.raw(rawQuery, bindings)), etc.

PR #316, opened 2014-06-13
Postgres-specific functions - schema management

dweinand:
_re-opening #286 (with some additional work) against master_

This code introduces a way for each dialect to have it's own query, schema, and migrate methods by implementing a *Interface on the dialect's Client_* object. For the postgres schema functions I followed the convention of having a methods.js file, but this approach doesn't require that.

Beyond that, it implements some postgresql-specific functionality:

* set/show search_path
* create/drop schemas
* hasSchema to check if a schema exists
* hasTable now will only look for tables in the current schema


marshals:
+1 this would be nice to see.  Need to be able to set search path easily.

tgriesser:
Hey @alliejanoch & @dweinand sorry for taking forever to say anything on this and other related tickets. 

I'm currently overhauling how a ton of stuff is implemented in the various dialects in the "next" branch to make extending with a lot of new features sane - currently any new feature ends up being a mess of string concatenation and extending or adding random methods throughout the library (which is my fault for how it's setup)

Once I've got things ready with that I'll make sure this ticket and any other PG specific features get a lot of attention - really would love to get more features in for that.

elhigu:
Hi, I started working on knex while ago and thought that I should go through old pull requests and get them in if possible or close otherwise if they are not going to be finished.

@dweinand do you think this is still relevant? If so, conflicts needs to be fixed and documentation of the new feature added to index.html.

dweinand:
I honestly haven't used knex in over a year and a half. But it looks like this is no longer valid due to #518

Issue #368, opened 2014-07-12
Mariasql fails with 'fields is not defined' then 'no database selected' when used with bookshelf


elliotf:
See tgriesser/bookshelf#414 and tgriesser/bookshelf#415

elhigu:
```
Fixed by setting db instead of database:

config = {
  client: 'mariasql',
  connection: {
    db: 'yourdb',
    user: 'your_user',
    host: 'your_host',
    password: 'your_pass'
  }
};
```

Should be fixed in mariasql driver by converting database attribute to db.

wubzz:
Closing due to #2480

PR #369, opened 2014-07-12
Add mariasql client back to integration tests

elliotf:
See #368


landongn:
This doesn't seem to resolve the problem at all; It seems like the flag, regardless of where it's set, is ignored elsewhere.  I've spent today tracking it down with little luck.

tgriesser:
Maria is part of the tests in 0.8

PR #471, opened 2014-09-10
Hooks

myndzi:
I don't know what's up with the end of file newline, all my editors seem to add it, but knex seems the outlier on this count as far as what's usual. I only added the 'after' hook to the query, but as you can see hooks should be easy to add wherever you want them. I was going to have a 'before' hook too but it really didn't make any sense.

### DIFF
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -1,8 +1,41 @@
 'use strict';
 
+var Promise      = require('bluebird');
+
 module.exports = function(Target) {
 var _ = require('lodash');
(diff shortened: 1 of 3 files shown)


diversario:
Could you also add a hook that runs before query is executed so it'd be possible to call, say, `.options()` on the query.

elhigu:

@myndzi do you think this is still relevant? If so, conflicts needs to be fixed and documentation of the new feature added to index.html.

elhigu:
Closing this since there hasn't been activity for a while...

PR #473, opened 2014-09-10
Added option to enable or disable quoting (quote)

vschoettke:
Added an option per client and per query to disable/enable quoting (`.quote(false/true)`) of non-reserved lowercase names (only a-z 0-9 and underscore) for table names, aliases and column names. Names may not start with a number. If quote(false) and the name does not match the criteria then it's automatically quoted. 

In case of the oracle dialect this also adds automatic conversion of all uppercase result-column-names to lowercase because oracle converts case insensitive names to uppercase. 

To allow the possibility to turn quotes on/off per query I had to remove the memoization in the Formatter. And pass the *Builder to the Formatter. Currently the formatter is reused for subqueries (because of the bindings) therefore changing the quote option in a subquery does not work correctly.

@tgriesser This is only beneficial if someone always wants to use lowercase for names and wants oracle compatibility. However there's really no need for this if all uppercase names are used instead.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -12,6 +12,9 @@ var EventEmitter = require('events').EventEmitter;
 // object attaches fresh constructors for each component
 // of the library.
 function Client(config) {
+  if (config && config.quote !== false) {
+    config.quote = true;
+  }
   this.initFormatter();
   this.initRaw();
   this.initTransaction();
--- a/lib/dialects/mysql/formatter.js
+++ b/lib/dialects/mysql/formatter.js
@@ -22,23 +22,49 @@ Formatter_MySQL.prototype.operators = [
   'rlike', 'regexp', 'not regexp'
 ];
 
+// based on http://docs.oracle.com/cd/B10501_01/win.920/a97249/ch3.htm
+Formatter_MySQL.prototype.reservedWords = [
+  'abort', 'accept', 'access', 'add', 'all', 'alter', 'and', 'any', 'array',
+  'arraylen', 'as', 'asc', 'assert', 'assign', 'at', 'audit', 'authorization',
+  'avg', 'base_table', 'begin', 'between', 'binary_integer', 'body', 'boolean',
+  'by', 'case', 'char', 'char_base', 'check', 'close', 'cluster', 'clusters',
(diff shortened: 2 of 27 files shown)


whitelynx:
@vschoettke Could this get rebased so it'll merge cleanly?

rhys-vdw:
@whitelynx Happy to merge if you're willing to rebase and create a new PR.

elhigu:
@vschoettke @whitelynx I didn't understand from the description, what is the motivation / real world use case for this feature? 

Why would you like to turn quoting off?

I'm not sure if this feature is worth of added complexity, either way this still needs documentation update to index.html and rebase.

Darkle:
I could really use this as I'm running into a bit of an issue with knex surrounding bound values with single quotes.

I'm using the full text search module of SQLite and am using the "NEAR group" method whose syntax is like so: 
```sql
MATCH 'NEAR(one two) three'
```
https://sqlite.org/fts5.html#section_3

Running the following knex raw method:
```javascript
knex.raw("select * from pages where rowid in ("+
            "select rowid from fts where fts match '? or NEAR(?)' order by rank"+
            ")", 
            ['foo bar', 'foo bar']
)
```
results in:
```sql
select * from pages where rowid in (select rowid from fts where fts match ''foo bar' or NEAR('foo bar')' order by rank)
```
which is a syntax error. 
Changing it to double quotes like this:
```sql
select * from pages where rowid in (select rowid from fts where fts match "'foo bar' or NEAR('foo bar')" order by rank") 
```
also generates an error.

rhys-vdw:
@Darkle This is the same as trying to interpolate `%?%` for a `LIKE` statement, right?. I think you'd be better off writing a helper function that builds these `NEAR` statements.

```js
function near(...strings, distance) {
  return `NEAR(${ strings.map(str => '"str"') }, ${ distance })`
}

knex('table').whereRaw('some_column MATCHES ?', near('foo', 'bar'))
```

I don't really underwhat what NEAR is for, and your foo bar example isn't helping. If you want further help on NEAR please open a new issue.

Darkle:
@rhys-vdw yes that does work as you suggested. Thanks for the info. 🙂

elhigu:
Moved this to issues since there hasn't been activity for a while.

PR #474, opened 2014-09-10
Fix sqlite3 rename column

myndzi:
This turned more into a general refactor of the column name replacement code. Besides the problems I've mentioned previously, I realized/encountered a couple more to do with self-referencing foreign keys. A general search and replace on the strings is undesirable unless we can guarantee that nothing bad will happen, which is also I assume why the replacement strings were qualified with commas and stuff. The new method breaks the query into a few pieces and performs replacements only in appropriate places. It does NOT handle expressions in the query flow, since they can basically be just any SQL, but this shouldn't matter for where/how it's being used?

Cases tested: Do nothing if no match, replace column name in string only in the column definition, and not matching string in a foreign key reference to a different table, replace column name in string and self-referencing foreign key

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -9,5 +9,5 @@
   "undef": true,
   "trailing": true,
   "unused": true,
-  "predef": [ "-Promise" ]
+  "predef": [ "-Promise", "before", "after" ]
 }
--- a/lib/dialects/mysql/schema/table.js
+++ b/lib/dialects/mysql/schema/table.js
@@ -8,6 +8,8 @@ var inherits = require('inherits');
 var Schema   = require('../../../schema');
 var helpers  = require('../../../helpers');
 
+var Promise  = require('../../../promise');
+    
 // Table Builder
 // ------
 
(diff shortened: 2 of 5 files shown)


myndzi:
This should be mergeable now...?

bendrucker:
Can you explain the need for the additions with mysql?

myndzi:
I wrote the sqlite3 code to support renaming foreign key columns. Then I wrote tests to verify that it worked. Then mysql broke under those same tests. The code now allows renaming of foreign key columns under mysql.

myndzi:
Ben: did you need more info or is this one good to go?

bendrucker:
I'm thrilled about this being untested but that's a symptom of things being too entangled and tests littered all over.

myndzi:
I assume you mean 'not thrilled'? What is the untested part you are concerned about? I added tests..?

bendrucker:
Too much code today...

You're 100% right on both counts. There could be some stricter testing on the recreation of the FKs but this is fine for now. Crazy complex workarounds are one of the reasons I'm so excited to split these things out into packages. We want to get things right for each database and situations like this is where that guarantee gets especially hairy.

tgriesser:
Just FYI, the "renames the column" test is broken on maria...

```js
KNEX_TEST_INTEGRATION_DIALECTS=maria npm run plaintest
```

```js
Error: ALGORITHM=COPY is not supported. Reason: Columns participating in a foreign key are renamed. Try ALGORITHM=INPLACE
```

It looks like the `getFKRefs` query isn't working for maria

myndzi:
Well, this was an sqlite patch; if the test breaks, it's because tbl.renameColumn breaks? I'm not familiar with any of the mysql/maria code yet. Looks like it's saying you can use an option (how does it get set?) to rename the columns in place, which is preferable anyway

myndzi:
So basically this functionality was simply broken and not tested before. When I added the tests for sqlite, mysql started bombing them since it didn't have a rename-column-with-foreign-key test before. I'm not sure why this is failing on maria but not mysql; if information_schema diverged between the two what do we want to do about it? Do we really want to have to detect whether it's mariadb or mysql and perform code differently? Any thoughts?

PR #511, opened 2014-09-30
implement support for postgres create/drop schema

tkellen:


hardchor:
Only found out about this on here. How come this isn't documented under http://knexjs.org/#Schema?

Issue #519, opened 2014-10-03
Conflict with Postgresql jsonb operators


metainfa:
Hi!

I have sql:

```sql
SELECT id, name, price, files->'photos'->0->>'name' as photo, annotation
FROM product
WHERE tags ?& array['17']
ORDER BY (tags->>'17')::int
LIMIT 10

```

## If i run this sql through node-postgres it work fine again:

```javascript
var pg = require('pg');

var connPg = "postgres://user:password@localhost/mydbname";
var clientPg = new pg.Client(connPg);
var sqlProductsByTags = [
  "SELECT id, name, price, files->'photos'->0->>'name' as photo, annotation ",
  "FROM product ",
  "WHERE tags ?& array['17'] ",
  "ORDER BY (tags->>'17')::int ",
  "LIMIT 10 "
];

clientPg.connect(function(err) {
  if(err) {
    return console.error('could not connect to postgres', err);
  }
  //console.log('SQL = ' + sqlProductsByTags.join(''));

  clientPg.query(sqlProductsByTags.join(''), function(err, result) {
    if(err) {
      return console.error('error select products by tags', err);
    }

    for(var i = 0, max = result.rows.length; i < max; i++) {
      console.log('name: ' +  result.rows[i].name);
    }

    clientPg.end();
  });
});
```
## If i run raw sql by knex, it fail with error (below listed): 

```javascript
var connPg = "postgres://user:password@localhost/mydbname"; 
var pg = require('knex')({
  client: 'pg',
  connection: connPg
});
var sqlProductsByTags = [
  "SELECT id, name, price, files->'photos'->0->>'name' as photo, annotation ",
  "FROM product ",
  "WHERE tags ?& array['17'] ",
  "ORDER BY (tags->>'17')::int ",
  "LIMIT 10 "
];

pg.raw(sqlProductsByTags.join(''))
```

This Error
```sh
{ [error: syntax error (at or near "$1")]
  name: 'error',
  length: 142,
  severity: 'ERROR',
  code: '42601',
  detail: undefined,
  hint: undefined,
  position: '98',
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  file: 'scan.l',
  line: '1053',
  routine: 'scanner_yyerror' }
```

I think that it fail because of i used jsonb operator '?&', precise only one symbol '?'
[Reference to PostgreSQL JSONB operators](http://www.postgresql.org/docs/9.4/static/functions-json.html)

[...]

PR #520, opened 2014-10-03
Fix for #519

tgriesser:


whitelynx:
@tgriesser Does this PR still make sense against master? If so, should it get merged?

PR #531, opened 2014-10-15
Do not lose domain during query runtime

salper:
It is not really a pull-request, more or less a question or a way to use domains on a Knex/Bookshelf based application.

The main problem resides in the way emitters and domains work together, and the fact that the pool is a collection of emitter singletons (created before or after inclusion in a domain).

I could use the same pattern which leads to unstack/restack domains before entering and after leaving a Knex/Bookshelf component. I'm just curious about your point of view, or the way used by people to bypass this domain caveat.


tgriesser:
Since domains are now deprecated, this isn't something we're going to address in knex.

PR #551, opened 2014-11-09
Fixes #550: MySQL dialect doesn't work when connection is specified by using the QueryBuilder

### LINKED ISSUE #550: MySQL dialect doesn't work when connection is specified by using the QueryBuilder.
```
var knex = require("knex")({ dialect: "mysql" });
knew.select().from("users").connection(myConnection).then(function (results) {
  // Do something.
});
```

**Observed result:**

There's a NPE in [interface.js#then](https://github.com/tgriesser/knex/blob/master/lib/interface.js#L27) because the runner is not loaded.

The [mysql dialect](https://github.com/tgriesser/knex/blob/master/lib/dialects/mysql/index.js#L21) initializes the Runner only if there exist a connection when it is created.

**Expected result:**

The mysql dialect must always initialize the runner (and the driver and the pool as well) in order to take into account the connection specified by `QueryBuilder#connection`.

seykron:


elhigu:
If still needed send a new pull request.

PR #577, opened 2014-11-24
Use pg-native for postgres dialect if available

### LINKED ISSUE #575: Ability to use pg-native
Hi.
Sorry if it was already answered.

Is there any way to use [pg-native](https://github.com/brianc/node-pg-native) PostgreSQL bindings instead of pure JS pg one? node-postgres documentation says that `require('pg').native` should be used instead of `require('pg')` (as in [Knex's PostgreSQL dialect](https://github.com/tgriesser/knex/blob/master/lib/dialects/postgres/index.js#L37)) in order to get access to the native binding.

Is there any common way to extend dialect or it need to be monkey-patched?

Thanks.

Kagami:
Fixes #575

### DIFF
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -34,7 +34,12 @@ Client_PG.prototype.dialect = 'postgresql';
 Client_PG.prototype.initDriver = function() {
   pg = pg || (function() {
     try {
-      return require('pg');
+      var pgn = require('pg');
+      try {
+        return pgn.native || pgn;
+      } catch (e) {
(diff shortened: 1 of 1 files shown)


Kagami:
Ok, I get the answer. Unfortunately `node-pg-query-stream` is not going to work with native bindings so closing this.

sheerun:
Can we support native bindings without streaming?

bendrucker:
We'll take a PR if someone can find a clean way to detect native bindings and throw with a clear error message when `stream` is called while they're used. Otherwise, no.

Issue #452, opened 2014-09-01
pool: connection ending in fatal error are not removed from the generic pool


syndr0m:
when having a fatal error on a mysql connection using knex 0.6.22 & pool & mysql driver,  the connection is not removed from the pool.

Further requests using this connection end in mysql driver error "cannot enqueue ... after fatal error" (PROTOCOL_ENQUEUE_AFTER_FATAL_ERROR)

a workaround :
```
var knex;

var connectionErrorHandler = function (connection, err) {
  if (connection && err && err.fatal) {
    if (connection.removedFromThePool)
      return;
    connection.removedFromThePool = true;
    knex.client.pool.genericPool.destroy(connection);
  }
};

var config =  {
  client: 'mysql',
  (...)
  pool: {
    (...)
    afterCreate: function (connection, callback) {
      connection.on('error', connectionErrorHandler.bind(null, connection));
      connection.on('end', connectionErrorHandler.bind(null, connection));
      callback();
    }
  }
};

knex = knexInit(config);
```

it might be linked to #206

wubzz:
The recent changes to pool handling in knex from 0.9 and onwards seems to have addressed most issues. We also discovered this kind of behaviour specifically for mysql2.

I'm going to close this and consider it resolved, but feel free to add a new issue if the error persists.

NemoAlex:
I just experienced this error on 0.14.0.

elhigu:
@NemoAlex do you have some test code to reproduce that?

shwuhk:
I have encountered into this issue too.
We use mysql with pool. This error happen when the connection is idle for a long time(like more than ten minutes) and then we do some query.
I suspect that this was happening because knex is trying to use a connection which is already timeout.
Any ideas?

elhigu:
@shwuhk I would recommend to not use 0.14 until 0.14.1 is released. There are couple of nasty issues which were missed when generic-pool was updated. I'm working on those and hope to get 0.14.1 out this week.

shwuhk:
I would try it again then when 0.14.1 is out. Thanks.

elhigu:
@shwuhk 0.14.1 it is out

PR #578, opened 2014-11-24
Do not add bad connections back to pool (fix for #452)

elliotf:
See #452.

This is a hacky workaround proof-of-concept conversation starter that
fixes the issue, but should be done in a cross-dialect way, and it
should also include tests.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -23,6 +23,14 @@ Pool.prototype.initialize = function() {
   return new GenericPool(_.defaults(this.config, _.result(this, 'defaults')));
 };
 
+function connectionErrorHandler(pool, connection, err) {
+  if (connection && err && err.fatal) {
+    if (connection.__knex__disposed) return;
+    connection.__knex__disposed = true;
+    pool.genericPool.destroy(connection);
+  }
+}
+
 // Some basic defaults for the pool...
 Pool.prototype.defaults = function() {
   var pool = this;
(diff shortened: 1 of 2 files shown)


tgriesser:
@elliotf is this one still good to be merged?

lib/pool.js
```diff
@@ -32,6 +40,11 @@ Pool.prototype.defaults = function() {
     create: function(callback) {
       pool.client.acquireRawConnection()
         .tap(function(connection) {
+          if ('function' === typeof connection.on) {
+            var errorHandler = connectionErrorHandler.bind(null, pool, connection);
+            connection.on('error', errorHandler);
+            connection.on('end',   errorHandler);
```
atiertant: 'end' event could be emit without any error

lib/pool.js
```diff
@@ -23,6 +23,14 @@ Pool.prototype.initialize = function() {
   return new GenericPool(_.defaults(this.config, _.result(this, 'defaults')));
 };
 
+function connectionErrorHandler(pool, connection, err) {
+  if (connection && err && err.fatal) {
```
atiertant: does not handle 'end' event if there is no err

atiertant:
an other problem is "what's happening to query using this bad connection ?" : Promise is never resolved or rejected ...

elhigu:

@elliotf do you think this is still relevant? If so, conflicts needs to be fixed also there was already couple of code comments. If PR is fixed I'll give it a review too.

elhigu:
Closing this for now since this hasn't been active, lets open this again if work continues... @wubzz does your changes to pooling do anything related to this issue?

Issue #593, opened 2014-12-13
Events


wubzz:
This isn't as much an issue as it is a request, but seeing as knex is already utilizing an EventEmitter, why not utilize this for other types of events alongside the 'query' event?

Don't get me wrong, the query event is helpful, but for instance an 'error' event being triggered upon reaching .catch() or callback({...}) would be great for logging purposes, in comparison to calling your Log functions anywhere a knex instance is being used (Say 3-5 places in a semi-big application).

Cheers.

PR #594, opened 2014-12-13
Add 'Error' event should a query fail.

wubzz:
In reference to Issue #593 , rather than requesting, figured I'd do pull request..

Tested on Postgresql, should work on rest. Purpose is to be able to listen on query errors should they occur, so that it can be used for logging purposes rather than placing your log-logic in the promise-reject/callback() functions of each Runner.query call. Convenient, and less code.

IE:
knex.on('error', myQueryErrorHandler);

### DIFF
--- a/knex.js
+++ b/knex.js
@@ -108,12 +108,15 @@ Knex.initialize = function(config) {
     Dialect = Clients[clientName]();
     client  = new Dialect(config);
 
-    // Passthrough all "start" and "query" events to the knex object.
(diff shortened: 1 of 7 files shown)


wubzz:
Any update or thoughts on this one?

Also may be affected by #855 .

wubzz:
I might redo this sometime this week, I still feel its a good event to have. For now, I'm closing this as the code changes are over 1 year old and will be easier to redo the changes than attempting to solve conflicts.

Issue #567, opened 2014-11-18
Migrations should be wrapped in a transaction automatically


ricardograca:
Whenever a migration fails it may leave the database in an inconsistent state, and a rollback won't be possible since the migration wasn't finished. I've looked through some of the closed issues and it seems some people already had problems due to this behavior. At least #312 and #148, but I didn't look very far back.

From my point of view a migration should always be wrapped inside a transaction that rolls back any changes in case of failure, if possible. It shouldn't be up to the user to do this manually, or is there a reason this can't be done automatically?

tgriesser:
This was handled by #622 and will be released soon.

PR #622, opened 2015-01-08
Wraps migrations in transations

zcei:
Hey,

I thought solving #567 would be a great start for contributing to `knex`, as I found this lovely project looking for a migration tool.

Tests are still running for me, but does anyone has a good idea on how to simulate a broken migration?
Thought about changing a column type, e.g. `enum` to `timestamp`, shouldn't have an automated conversion.
Seems not clean, though.

Hope to provide more contribution soon!

### DIFF
--- a/lib/migrate/index.js
+++ b/lib/migrate/index.js
@@ -155,6 +155,8 @@ Migrator.prototype._runBatch = function(migrations, direction) {
         })
         .then(function(batchNo) {
           return this._waterfallBatch(batchNo, migrations, direction);
+        }).catch(function(error) {
+          helpers.warn('migrations failed with error: ' + error.message);
         });
     });
 };
@@ -248,10 +250,9 @@ Migrator.prototype._waterfallBatch = function(batchNo, migrations, direction) {
 
     // We're going to run each of the migrations in the current "up"
     current = current.then(function() {
-      var transaction = knex.transaction(function(trx) {
-        return migration[direction](trx, Promise);
+      return knex.transaction(function(trx) {
+        return warnPromise(migration[direction](trx, Promise), 'migration ' + name + ' did not return a promise');
       });
-      return warnPromise(transaction, 'migration ' + name + ' did not return a promise');
     }).then(function() {
       log.push(path.join(directory, name));
       if (direction === 'up') {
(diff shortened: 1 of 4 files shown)


bendrucker:
> how to simulate a broken migration?

Add a valid column. Then run an invalid raw query.

zcei:
I'm not that familiar with MySQL:
Though I'm using a transaction, it creates the column from the invalid migration - and doesn't roll back.

However, for `postgres` it's working..

bendrucker:
[Yet another reason to not use MySQL](http://dev.mysql.com/doc/refman/5.1/en/implicit-commit.html). Forgot that the DDL implicitly commits changes. MySQL is the only DB without transaction support for schema building. It'll need to be documented but I see no problem with trying to use a transaction and then just letting it be implicitly committed.

zcei:
RLY?! wtf..
The statements mentioned there are IMO the most basic ones.
`CREATE TABLE, DROP TABLE, RENAME TABLE, ALTER TABLE`

Feeling confirmed in my love to `pg` :dancers: 

You think it's appropriate to check whether the used dialect is `mysql` & change assertion based on it? :construction:

tkellen:
<3 @zcei I have been meaning to do this for ages.

bendrucker:
The test needs special handling for mysql. But the code can just assume the transaction will work and let it get immediately committed in MySQL.

bendrucker:
@tgriesser I'll clean this up for a merge if you're cool with it. Shouldn't be a breaking change in any situation.

tgriesser:
What about situations where you already have your migrations wrapped in transactions? That was my one hesitance - I hadn't looked into what becomes of nested transactions, whether they're no-ops or not.

tgriesser:
Basically:

```js
knex.transaction = function(container) {
  return new client.Transaction(container);
};
```

should likely become:

```js
knex.transaction = function(container) {
  return container(client);
};
```

when the client is [created from a transaction](https://github.com/tgriesser/knex/blob/8c7af464395965f5273a9781360da088acb72501/lib/transaction.js#L25-L27). Eventually nested transactions will become `SAVEPOINT`s but for now this will make this merge possible without folks needing to change any existing migrations.

bendrucker:
Ah, correct. @zcei if you can figure this out and get a test against nested transactions I'll get this merged. Otherwise it'll be a little longer before I can review.

tgriesser:
@zcei can you change this PR to target the `pool` branch instead? I've got savepoints implemented there (see #786) so this should be possible

tgriesser:
Merged.

kimmobrunfeldt:
Is this yet released?

zcei:
`0.8.6`s only changes to current master are documentation fixes, so it's in a released version, yes.

PR #628, opened 2015-01-15
Fix bare subqueries in insert statements.

### LINKED ISSUE #627: Subqueries in insert statements
Using a subquery inside `.insert()` without wrapping it in a function yields an incorrect query.

Is there a reason for the requirement of wrapping the subquery in a function? I can't think of any use case where a user passing a subquery `QueryBuilder` instance to one of the functions would _not_ want it wrapped with parenthesis.

As far as implementation, we could change [Formatter#parameter()](https://github.com/tgriesser/knex/blob/master/lib/formatter.js#L33) to check for subqueries by checking if `value instanceof QueryBuilder` like [Formatter#checkRaw()](https://github.com/tgriesser/knex/blob/master/lib/formatter.js#L40) does.

----

Test cases
----------

### Case 1: A literal value

This query works as expected:
```javascript
db('geom_text').insert({description: 'a point!', geom: 'POINT(-71.064544 42.28787)'}).toString();
```
...this results in:
```
insert into "geom_text" ("description", "geom") values ('a point!', 'POINT(-71.064544 42.28787)')
```

### Case 2: A direct subquery

This one, however, fails:
```javascript
var subquery = db().select(db.raw('?', ['POINT(-71.064544 42.28787)']));
db('geom_text').insert({description: 'a point!', geom: subquery}).toString();
```
...resulting in:
```
insert into "geom_text" ("description", "geom") values ('a point!', select 'POINT(-71.064544 42.28787)')
```

### Case 3: A subquery returned by a function

This query, works as expected again:
```javascript
db('geom_text').insert({description: 'a point!', geom: function()
{
    return this.select(db.raw('?', ['POINT(-71.064544 42.28787)']));
}}).toString();
```
...resulting in:
```
insert into "geom_text" ("description", "geom") values ('a point!', (select 'POINT(-71.064544 42.28787)'))
```

whitelynx:
This updates `Formatter#checkRaw()` to pass the `parameter` flag to `outputQuery()` as `alwaysWrapped`.

Fixes #627.

### DIFF
--- a/lib/formatter.js
+++ b/lib/formatter.js
@@ -40,7 +40,7 @@ Formatter.prototype.checkRaw = function(value, parameter) {
   if (value instanceof QueryBuilder) {
     var query = value.toSQL();
     if (query.bindings) push.apply(this.bindings, query.bindings);
-    return this.outputQuery(query);
+    return this.outputQuery(query, parameter);
   }
   if (value instanceof Raw) {
     if (value.bindings) push.apply(this.bindings, value.bindings);
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -731,7 +731,7 @@ module.exports = function(qb, clientName, aliasName) {
         default: 'select * from "users" having "email" > ?'
       });
     });
-    
+
     it("nested having", function() {
       testsql(qb().select('*').from('users').having(function(){
         this.where('email', '>', 1);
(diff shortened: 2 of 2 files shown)


whitelynx:
Sorry for the whitespace changes... I have my editor set up to strip trailing whitespace in most files, and I forgot to check for that before making the pull request.

PR #641, opened 2015-01-22
FoundationDB SQL Layer Adapter

gjvargas:
I've added support for a FoundationDB SQL Layer knex adapter. I tried my best to follow the style of code throughout the project. The FDB SQL layer code was adapted from the PostgreSQL code. I also updated the test directory to include both unit tests and integration tests for the fdbsql adapter.

For more information about the newly supported database, see here: https://foundationdb.com/

Pleas don't hesitate to contact me for questions and comments.

Thanks!

### DIFF
--- a/lib/dialects/fdbsql/index.js
+++ b/lib/dialects/fdbsql/index.js
@@ -1,0 +1,138 @@
+'use strict';
+
+// FDB SQL Layer index
+// This file was adapted from the PostgreSQL index
+
+var _        = require('lodash');
+var inherits = require('inherits');
+
+var Client  = require('../../client');
+var Promise = require('../../promise');
+
+var pg;
+
+// Always initialize with the "QueryBuilder" and "QueryCompiler"
+// objects, which extend the base 'lib/query/builder' and
+// 'lib/query/compiler', respectively.
+function Client_FDB(config) {
+  Client.apply(this, arguments);
+  if (config.returning) this.defaultReturning = config.returning;
+  if (config.debug) this.isDebugging = true;
+  if (config.connection) {
+    // If no port has been defined,
+    if (!config.connection.port)
+      // set the port to the SQL Layer's default port, 15432.
+      config.connection.port = 15432;
+    this.initDriver();
+    this.initRunner();
+    this.connectionSettings = config.connection;
+    this.initPool();
+    this.pool = new this.Pool(config.pool);
+  }
+}
+inherits(Client_FDB, Client);
+
+Client_FDB.prototype.dialect = 'fdbsql';
+
+// Lazy load the pg dependency, since we might just be using
+// the client to generate SQL strings.
+Client_FDB.prototype.initDriver = function() {
+  pg = pg || (function() {
+    try {
+      return require('pg');
+    } catch (e) {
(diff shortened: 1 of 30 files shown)


bendrucker:
Please rebase this into a single commit when you have a chance and I'll review it. Knex's internals are starting to become unmanageable so please do the best you can to maximize test coverage. That's the only way to prevent regressions when this stuff needs to be rewritten.

gjvargas:
I've rebased as requested. Nearly every test that previously existed for postgreSQL runs for the fdbsql adapter now as well. Note that the ones that aren't properly tested against the FoundationDB SQL Layer are testing unsupported operations for the database. A couple of examples include the tests using the JSON and enum types.

tgriesser:
What Ben failed to mention though, is awesome work on putting this together @gjvargas!

Knex's internals are definitely getting a little hairy but should be looking really nice in the coming weeks as I'm going to dive back into things with a fresh look on modularity and coming up with an approach to use transducers throughout the codebase for consistent parsing and query generation.

All that said, Ben and I will take a look through and ask any questions.

lib/dialects/fdbsql/index.js
```diff
@@ -0,0 +1,138 @@
...
+inherits(Client_FDB, Client);
+
+Client_FDB.prototype.dialect = 'fdbsql';
+
+// Lazy load the pg dependency, since we might just be using
+// the client to generate SQL strings.
+Client_FDB.prototype.initDriver = function() {
+  pg = pg || (function() {
+    try {
+      return require('pg');
```
tgriesser: Is this supposed to be pg? or fdb?

gjvargas: It's correct as is. FoundationDB actually uses the pg network protocol for convenience, so I was able to use the same dependency as the PostgreSQL adapter.

gjvargas:
Awesome. Thanks guys!

nathanlws:
Hi @tgriesser @bendrucker, I worked with @gjvargas on this and would be happy to make any needed changes. Anything I can do to help wrap this up?

We have ones ready to go for Bookshelf and Ghost once this is in :)

bendrucker:
I'm fine with this. I'm restarting the last build to see what's up on the tests. As demand for other DBs increases, I'd like to generate multiple coverage reports specific to each DB so we can get a better handle on both DB-related status and stability.

bendrucker:
Can you fix the tests?

nathanlws:
> Can you fix the tests?

Sure!

SQL Layer isn't yet available in Travis so the integration tests will need to be skipped for CI. Is that ok?

bendrucker:
Yeah, that goes along with the "experimental" comment.

nathanlws:
Removed from default integration list and rebased. And looks like Travis is happy.

bendrucker:
Great. On my list for the weekend.

tgriesser:
Welp. Guess we'll see if there's any future here - https://www.npmjs.com/package/fdb

bendrucker:
Wow — time to revert this then:

http://www.macrumors.com/2015/03/24/apple-acquires-foundationdb/

PR #642, opened 2015-01-23
Added schema dump/load/drop/create/reset

SamDuvall:
I created a few more command line commands that mirror Rails commands I frequently used.
* knex schema:dump = Dump the schema the DB to an .sql file
* knex schema:load = Load the schema in a .sql file into the DB
* knex schema:create = Create the schema in the DB
* knex schema:drop = Drop the schema from the DB
* knex schema:reset = Run drop, create and load in sequence

### DIFF
--- a/knex.js
+++ b/knex.js
@@ -149,6 +149,14 @@ Knex.initialize = function(config) {
       }
     },
 
+    // Lazy-load / return a `SchemaLoader` object.
+    schemaLoader: {
+      get: function() {
+        if (!client.SchemaLoader) client.initSchemaLoader();
+        return new client.SchemaLoader(knex);
+      }
+    },
+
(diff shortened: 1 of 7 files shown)


bendrucker:
Much appreciated! For us to merge this, it'll need to work in all supported databases, not just MySQL. This is a perfect candidate for Tim and my upcoming project of modularizing/packagizing Knex, especially with respect to things like the schema builder and the migrations. 

If you don't mind digging into the Rails source, I'd really like to look at how they handle this functions.

santiagovazquez:
This is what I need. Thanks for your work @SamDuvall !

elhigu:
This would be nice feature if implemented properly for other databases and documented :+1:  Needs new pull request though.

PR #659, opened 2015-02-03
Don't die when pg client emits error

myndzi:


myndzi:
These tests are super whack. "sqlite3 dropTable has a dropTableIfExists method" -- dialects/postgres/index ???

I don't even know how to approach this.

myndzi:
There. The 'end' event may not be necessary, but I don't know enough about the inner guts of knex to really track that down. Tests pass, and forcing an error event via pg_terminate_backend doesn't bother my connection test script one bit.

bendrucker:
Will do my best to make time for this sometime this week

myndzi:
Do not merge this just yet. There's an outstanding issue causing connection leaks(!)

The fix is either in generic-pool-redux (ensure count is adjusted only by successfully destroyed objects) or in this patch (validate/synchronize pool destroy requests to ensure invalid or already-removed things aren't called)

The former makes more sense to me, and I'm uncertain how to correctly perform the latter.

myndzi:
Incompatible with https://github.com/tgriesser/knex/pull/693

When the above is added, will re-submit this one.

bendrucker:
Is this good to merge now?

tgriesser:
Doesn't look like it, still referencing `genericPool`

myndzi:
It wasn't synching up with my origin branch, so the change didn't get pushed; I rebased it onto upstream/master and force-pushed.

bendrucker:
If you can do a second rebase and get these squashed into a single commit this seems good to merge

myndzi:
Okay.

bendrucker:
Done! Thanks for your help on this

PR #660, opened 2015-02-03
Respect union wrap parameter as last argument

werkt:
union's wrap parameter was being ignored and misinterpreted when
building statements.  Maintain the spirit of the original implementation
(variable number of callbacks) consistent with other instances of
variadic-like behavior (select/columns), including handling list of
callbacks in array, while still processing the wrap parameter.

Updated documentation which was missing a reference to wrap, and
provided a unit test for wrapped unions.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1014,9 +1014,10 @@ knex.select('*').from('users').limit(10).offset(30)
 </pre>
 
     <p id="Builder-union">
-      <b class="header">union</b><code>.union(query)</code>
+      <b class="header">union</b><code>.union([*queries], [wrap])</code>
       <br />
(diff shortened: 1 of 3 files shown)


bendrucker:
Can you tack on a test with multiple unions?

bendrucker:
I know that was present before but it would be great to get it in now.

werkt:
Tests added.

bendrucker:
:) that's why tests are great. will review this this weekend.

tgriesser:
Great, thanks @werkt!

khalilTN:
Hi, is this already available on npm ?

PR #663, opened 2015-02-04
Don't call .then on promises when not giving callbacks

bbh-node:
This gets rid of the warning spam when running tests due to the (ab?)use of promise.then() in the .then-wrapping code


myndzi:
Hrm. I didn't know you could submit a pull request as another account. This is me, anyway.

myndzi:
How do I force a rebuild? This error doesn't have anything to do with this patch as far as I can tell; I just encounter it occasionally when running tests. My best guess is something somewhere isn't returning its promise and therefore some code is running before previous asynchronous calls are completed.

PR #664, opened 2015-02-04
Don't call .then on promises when not giving callbacks

myndzi:
This gets rid of the warning spam when running tests due to the (ab?)use of promise.then() in the .then-wrapping code


bendrucker:
See https://github.com/petkaantonov/bluebird/issues/482

PR #672, opened 2015-02-07
Adds build for node 0.12

halfdan:
**Note:** Just realized that sqlite3 doesn't build properly on v0.12 yet - doesn't make much sense merging at this point.


bendrucker:
Yeah we'll do this when the deps are ready

PR #686, opened 2015-02-13
Use db driver pool

myndzi:


myndzi:
How is this even a thing... my tests don't hit this error

PR #687, opened 2015-02-13
Use PG native driver pool

myndzi:
I should have separated these to start, I guess. This one passes tests for me at least.

### DIFF
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -8,13 +8,12 @@ var inherits = require('inherits');
 var Client  = require('../../client');
 var Promise = require('../../promise');
 
-var pg;
-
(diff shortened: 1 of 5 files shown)


myndzi:
This kind of thing is getting really tiresome... I don't even know how to approach fixing this. The tests pass on my dev box, and the broken test doesn't appear to have anything directly to do with the change.

myndzi:
Okay. The acquireRawConnection method had a side effect of setting client.version when it wasn't present; this was for feature detection for things like json columns. Travis apparently isn't using postgres 9.2(?), so when the client.version value was null it showed up the error in the version check logic. Now the logic is fixed and the version number is back.

myndzi:
Superseded by https://github.com/tgriesser/knex/pull/693

PR #693, opened 2015-02-19
Use Pool2 instead of Generic-pool-redux

myndzi:
Passes sqlite and postgres integration tests, leaving it to Travis for the others

### DIFF
--- a/lib/dialects/mysql/index.js
+++ b/lib/dialects/mysql/index.js
@@ -91,7 +91,7 @@ function connectionErrorHandler(client, connection, err) {
   if (connection && err && err.fatal) {
     if (connection.__knex__disposed) return;
     connection.__knex__disposed = true;
-    client.pool.genericPool.destroy(connection);
+    client.pool.destroy(connection);
   }
 }
 
--- a/lib/dialects/oracle/pool.js
+++ b/lib/dialects/oracle/pool.js
@@ -16,7 +16,7 @@ inherits(Pool_Oracle, Pool);
 
 Pool_Oracle.prototype.defaults = function() {
   return _.extend(Pool.prototype.defaults.call(this), {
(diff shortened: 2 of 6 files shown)


myndzi:
This pull request is incompatible with the other one for binding error handlers for postgres:

https://github.com/tgriesser/knex/pull/659

Once this one is accepted, I will update that one.

tgriesser:
Really awesome @myndzi! Let's kick the tires on this a bit and see if it causes any issues

@elliotf any chance you want to check this one out on master when you get a chance and see if it helps with issues you've also brought up previously regarding pooling

myndzi:
He would also need https://github.com/tgriesser/knex/pull/659 just to be clear. Besides actively tracking all connections, pool2 provides fire-and-forget .remove and .destroy methods that work whether the resource is checked out or not. They still have to be called, though, which is what the other patch is for in the case of idle resources from pg.

lib/dialects/mysql/index.js
```diff
@@ -91,7 +91,7 @@ function connectionErrorHandler(client, connection, err) {
   if (connection && err && err.fatal) {
     if (connection.__knex__disposed) return;
     connection.__knex__disposed = true;
-    client.pool.genericPool.destroy(connection);
+    client.pool.destroy(connection);
```
elliotf: should this be `destroy` or `release` like the others?

elliotf: Ah, nevermind.  This happens on error; the others happen when they're done with the connection.

myndzi: .release releases the resource to the pool
.remove removes the resource from the pool nicely
.destroy removes the resource from the pool dirtily

Knex only passes a remove handler so far so you should always call .remove

PR #706, opened 2015-02-24
Fix for out-of-order parameters when using a subquery in 'from'

### LINKED ISSUE #704: Incorrect order of query parameters when using subqueries
When building a query using a subquery, and using parameters in both, the resulting order of the parameters is incorrect. This specifically seems to happen when the main query uses `db.raw()` in its select clause.

Given this JS:
```javascript
var subquery = db().select().from('tbl').where({baz: 'some string'}).as('g');
var query = db().select(db.raw('?', [3]), 'g.geom').from(subquery).where({'g.bzzz': false});
console.log(query.toSQL());
console.log(query.toString());
```

...we get this output:
```javascript
{ method: 'select',
  options: undefined,
  bindings: [ 'some string', 3, false ],
  sql: 'select ?, "g"."geom" from (select * from "tbl" where "baz" = ?) as "g" where "g"."bzzz" = ?' }
```
```sql
select 'some string', "g"."geom" from (select * from "tbl" where "baz" = '3') as "g" where "g"."bzzz" = 'false'
```

...whereas we would expect to get:
```javascript
{ method: 'select',
  options: undefined,
  bindings: [ 3, 'some string', false ],
  sql: 'select ?, "g"."geom" from (select * from "tbl" where "baz" = ?) as "g"' }
```
```sql
select '3', "g"."geom" from (select * from "tbl" where "baz" = 'some string') as "g" where "g"."bzzz" = 'false'
```

It seems as if the subquery handling code is just naively prepending the subquery's parameters to the outer query's parameters, instead of inserting them between parameters from the outer's `select` clause and the ones from the outer's `where`.

whitelynx:
Fixes #704.

### DIFF
--- a/lib/query/compiler.js
+++ b/lib/query/compiler.js
@@ -16,7 +16,6 @@ function QueryCompiler(queryBuilder) {
   this.single      = queryBuilder._single;
   this.transacting = queryBuilder._transacting;
   this.grouped     = _.groupBy(queryBuilder._statements, 'grouping');
-  this.tableName   = this.single.table ? this.formatter.wrap(this.single.table) : '';
 }
 
 // Collapse the builder into a single object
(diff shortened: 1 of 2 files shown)


tgriesser:
Thanks, and thanks for including a test!

PR #740, opened 2015-03-17
fix(docs): fixes broken transaction link in docs

mgartner:


mgartner:
@bendrucker not sure why you re-did my change under your own commit, but I think it will be broken. The ID for the Transactions section is `Transactions` not `Transaction`.

https://github.com/tgriesser/knex/blob/master/index.html#L1371

bendrucker:
Your commit had `##` and it was easier to do from scratch than amend. Fixed.

PR #752, opened 2015-03-26
Adds test for escaping MySQL column names containing backticks

cspotcode:
- MySQL column and table names can contain backticks.
- They are escaped with a double-backtick, just like how PGSQL escapes doublequotes
- I am not 100% sure that the `default` SQL in this test is correct.  Someone should double-check

### DIFF
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -2273,15 +2273,15 @@ module.exports = function(qb, clientName, aliasName) {
 
     it('escapes queries properly, #737', function() {
       testsql(qb()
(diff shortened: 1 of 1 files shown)


cspotcode:
Oops, looks like I'll also need to update another test, here: https://github.com/tgriesser/knex/blob/13995d6936208fe0a968b9ae0f46a2f19faacffc/test/integration/builder/selects.js#L726-L731

If you think this bug is worth fixing I can update this PR with the other test and a fix.

elhigu:
New pull request will be needed to merge this test in.

PR #759, opened 2015-04-01
Don't use camelCase in aliased column names

### LINKED ISSUE #758: Rollback does not work with Redshift
Rollback clears all migrations when used with Redshift.

cpapazian:
The query to get the highest batch aliases the field to a camel-case name (`select max(batch) as batchNo`). Since Redshift downcases table and column names, the result comes back as `batchno` instead of `batchNo`. Using a snake-cased name instead fixes this.

Closes #758


cpapazian:
this is a fix for https://github.com/tgriesser/knex/issues/758

tgriesser:
Cool, thanks

Issue #711, opened 2015-02-25
Using where with an array should use IN query


mgartner:
It would be nice if `where` automatically used an `IN` statement when a value is passed that is an array, so that you don't have to explicitly call `whereIn`.

```js
knex('users').where({
  first_name: 'Test',
  role: ['owner', 'admin', 'user']
}).select('id')

// SELECT id FROM users WHERE first_name = 'Test' AND role IN ('owner', 'admin', 'user');
```

tgriesser:
Might be sort of convenient but then you hit a bunch of weird edge cases and it's a slippery slope to becoming more of a weird query DSL rather than (as close as we can) 1-to-1 with SQL, plus with Postgres' arrays it would get really ambiguous.

I would like to get something in here though that'd allow something like:

```js
knex('users').fromJS({
  where: {first_name: 'Test'},
  whereIn: {role: ['owner', 'admin', 'user']},
  select: 'id'
})
```

or something along those lines

PR #761, opened 2015-04-02
allows array value given to .where to be treated as .whereIn by the queryBuilder

T1B0:
syntaxic sugar, allows for 
```
qb().select('*').from('product').where({size : 100, category : ['goodie', 'gift']})
=> to generate : 
SELECT * FROM `product` WHERE size = 100 AND category IN ('goodies', 'gift');
```

### DIFF
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -175,8 +175,8 @@ QueryBuilder.prototype.andWhere = function(column, operator, value) {
   // are explicitly two arguments passed, so it's not possible to
   // do where('key', '!=') and have that turn into where key != null
   if (arguments.length === 2) {
-    value    = operator;
-    operator = '=';
+    value = operator;
+    operator = Array.isArray(value) ? 'in' : '=';
 
     // If the value is null, and it's a two argument query,
     // we assume we're going for a `whereNull`.
@@ -189,14 +189,12 @@ QueryBuilder.prototype.andWhere = function(column, operator, value) {
   var checkOperator = ('' + operator).toLowerCase().trim();
 
   // If there are 3 arguments, check whether 'in' is one of them.
-  if (arguments.length === 3) {
     if (checkOperator === 'in' || checkOperator === 'not in') {
-      return this._not(checkOperator === 'not in').whereIn(arguments[0], arguments[2]);
+      return this._not(checkOperator === 'not in').whereIn(column, value);
     }
     if (checkOperator === 'between' || checkOperator === 'not between') {
-      return this._not(checkOperator === 'not between').whereBetween(arguments[0], arguments[2]);
+      return this._not(checkOperator === 'not between').whereBetween(column, value);
     }
-  }
 
   // If the value is still null, check whether they're meaning
   // where value is null
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -126,7 +126,6 @@ module.exports = function(qb, clientName, aliasName) {
       });
     });
 
-
     it("where not", function() {
       testsql(qb().select('*').from('users').whereNot('id', '=', 1), {
         mysql: {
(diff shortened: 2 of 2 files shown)


tgriesser:
Thanks for the PR - this has come up a few times but I think for now we're going to stick with needing to be explicit, as postgres has an "array" type and so this might get pretty ambiguous.

bendrucker:
Thanks for the contribution! We decided there were too many ambiguous edge cases to do this (see https://github.com/tgriesser/knex/issues/711#issuecomment-76123594). 

However, you could definitely do this transformation via a helper in your app if you really do find yourself needing it often.

tgriesser:
Actually I'm looking to keep this one open, on second thought it could prove as a good simplified case of how it'll be possible to hook into the query construction in the next version to extend with something like this.

T1B0:
I've totally missed the similar #711 PR. I was actually surprised to see it's not the default behavior of the where function. 
Another way to meet similar use-case would have been to accept a specifically formatted object as value in where clause : 
```
select('*').from('foo').where({
    bar : 12,
    code : {operator : 'in', value : ['US', 'UK', 'AU']}
});
``` 
this would even allow more complex where clause building, by accepting all existing operators.

bendrucker:
It's not the default behavior for two reasons:

1. It abstracts away the SQL to an extent that's inconsistent with Knex
2. It's ambiguous when you consider the case of Postgres array columns.

T1B0:
I totally agree with you on 1/, it feels weird like a mongodb query {color : {$in : ['blue', 'green']}}.
About 2/ i think users of pg::array type is more an edge-case and are far fewer than users who'd expect an 'IN' operator for a js array - but that certainly just my opinion.

Anyway thanks for doing and maintaining this lib, it's really usefull and i enjoy working with it  :+1:

tgriesser:
I could see potentially supporting: 

```js
select('*').from('foo').where([
    ['bar', 12],
    ['code' , 'in', ['US', 'UK', 'AU']]
]);

```

T1B0:
you could even accept a mix of object and arrays like :
```
select('*').from('foo').where([
    {bar : 12},
    ['code' , 'in', ['US', 'UK', 'AU']]
]);
```
the point of this feature-request was for me to avoid the need of chunking my base parameter object to then do multiple call to where / whereIn. so this syntax would still allow me to do my query with only 1 call to where. Plus an array make it really easy to build your where clause by pushing params in it.

roytz:
@tgriesser so it's final that there won't be support for stuff like `.where({ name: ['Mike', 'John'] })`?
I find it very useful.

roytz:
Care to explain @elhigu ?
This is a great package which many uses and in my opinion, we should try and expand its features

elhigu:
@roytz Just a personal opinion. I wouldn't like to add any new APIs alternatives for doing the same thing with slightly different syntax. IMO current API already has too many syntaxes, which makes knex less robust that I would like it to be. Correct place to have these extended APIs would be some higher level libraries utilizing knex under the hood. 

We should try to prevent expanding it to support same things all over again and expand knex to support new type of queries, which has been impossible to create with current knex versions. We should also try to reduce current API to minimum having small consistent set of features, which are well tested, consistent and robust.

I also agree previous discussion that syntax would be ambiguous and that is one more reason to not add this.

roytz:
@elhigu I totally agree with your opinion about the current API having too many syntaxes - exactly why I thought we could use the same syntax for this so similar functionality.
If this library weren't to support this functionality, that's a whole other case. Since it does, I'm just suggesting using the existing syntax ( @T1B0 PR does that).
About the syntax being ambiguous, it's so common to do so and support input of string but also an array. (For example, almost every lodash function supports that)

elhigu:
@roytz https://github.com/tgriesser/knex/pull/761#issuecomment-88952428
> It's ambiguous when you consider the case of Postgres array columns.

roytz:
You mean for the case when searching for a complete specific array? Probably almost never happen.. This is not how postgres array should be used.. If you look for the whole array it means you should've stored it differently. Most array queries are to search for values inside the array so it just makes sense to support the latest option. Also, jsFrom doesn't support multiple arrays search. Since it's an object, it can only have one key called `whereIn` (or `andWhereIn`).

PR #762, opened 2015-04-03
Where array argz

T1B0:
following up on PR #761 - makes .where accepts an array of arguments


tgriesser:
Thanks for the PR but I think for now I'm going to avoid overloading the where operator with yet another way of doing things... I have added a `fromJS` method which allows you to build a query using something like:

```js
knex.fromJS({
   select: '*',
   from: 'users',
   where: {foo: bar},
   whereIn: ['id', [1, 2, 3]]
})
```

Hopefully that can suffice for the time being until I have a chance to think about it a bit more

PR #779, opened 2015-04-17
Allow raw columns in where clauses

santialbo:
This change allows to pass column as raw instead of having to pass the whole condition. This should be completely back compatible and allows the following syntax which is already allowed on `.whereIn` and `.whereBetween`:

```js
qb.where(knex.raw('"metadata"->\'width\''), operator, value)
```
I know that you can still achieve this by passing everything on the raw query but it was surprising to me that you couldn't do it this way. Feel free to close this if you think it's completely unnecessary.

regards


tgriesser:
Thanks so much!

PR #788, opened 2015-04-22
Wraps migrations in transations

tgriesser:
Moved from #622, merging against the "pool" branch which now has savepoint/nested transactions.


zcei:
Hej,

happy to see this! 
I had some problems getting the nested transactions working and then lost the focus :/

Nice that this seems to be fixed now :+1:

tgriesser:
Yep, still polishing thing a bit in the branch... should be soon.

PR #795, opened 2015-04-24
fix to work in browserify

calvinmetcalf:
using webpack to create a static build doesn't actually work because webpack uses different methods to look up the modules or something I don't know it through an error when I tried to bundle it, this doesn't and we can add more files to explicitly ignore if it gets too big


calvinmetcalf:
and fixed to avoid including the files you don't want it to include

Issue #809, opened 2015-05-03
Cannot Browserify


nolanlawson:
Environment:

```
$ browserify --version
10.0.0
$ node -v
v0.10.36
$ npm -v
2.7.0
$ uname -a
Darwin cinnabar.local 14.0.0 Darwin Kernel Version 14.0.0: Fri Sep 19 00:26:44 PDT 2014; root:xnu-2782.1.97~2/RELEASE_X86_64 x86_64 i386 MacBookAir6,2 Darwin
```

Steps to repro:

```
hub clone tgriesser/knex
cd knex
npm install -g browserify
browserify .
```

You'll get an error:

```
Error: ENOENT, open 'lodash/utility/noop'
```

reggi:
I think I'm running into the same issue here trying to compile browserify for node

```
Error: Cannot find module 'sqlite3' from '/Users/thomasreggi/Desktop/monorepo-reggi/packages/services/node_modules/knex/lib/dialects/sqlite3'
    at /Users/thomasreggi/Desktop/monorepo-reggi/packages/api/node_modules/resolve/lib/async.js:46:17
    at process (/Users/thomasreggi/Desktop/monorepo-reggi/packages/api/node_modules/resolve/lib/async.js:173:43)
    at ondir (/Users/thomasreggi/Desktop/monorepo-reggi/packages/api/node_modules/resolve/lib/async.js:188:17)
    at load (/Users/thomasreggi/Desktop/monorepo-reggi/packages/api/node_modules/resolve/lib/async.js:69:43)
    at onex (/Users/thomasreggi/Desktop/monorepo-reggi/packages/api/node_modules/resolve/lib/async.js:92:31)
    at /Users/thomasreggi/Desktop/monorepo-reggi/packages/api/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:117:15)
```

Here's my browserify script.

```
../../node_modules/.bin/browserify --node -s default -g browserify-aws-sdk -t babelify -o main.js index.js
```

nolanlawson:
That's really weird. In principle that shouldn't happen because of this: https://github.com/tgriesser/knex/blob/e3ea261cd167277354737408fdd5426bbce65776/package.json#L98

reggi:
@nolanlawson I think this is an issue with lerna :/

https://github.com/lerna/lerna/issues/212

reggi:
@nolanlawson I had the `--node` flag set which ignores the `browser` prop.

PR #810, opened 2015-05-03
#809 - fix browserification

nolanlawson:
My apologies for not adding any tests, but
it seems like you haven't reached the point
where you're browserifying knex and testing
it against WebSQL. If you'd like, though, a
simple command-line test could be:

    browserify . > /dev/null

In any case, I can confirm that after this PR
I am able to browserify knex.


calvinmetcalf:
I'm pretty sure you could replace all of these we either `false` because they are never called, or it would probably be shorter to just right a fricking noop

lib/noop.js
```diff
@@ -0,0 +1 @@
+module.exports = require('lodash/utility/noop')
```
calvinmetcalf: `module.exports = function noop (){};` is like 1000% times saner to user right here

nolanlawson: But what if lodash has performance improvements?

calvinmetcalf: It does not I checked

nolanlawson: joking, man :)

calvinmetcalf: I'm not, I wouldn't put it past jdd to have some sort of crazy `new Function` based black magic there

nolanlawson: hahaha, good point

PR #820, opened 2015-05-12
Fix pg.js regression

knownasilya:
Used to be able to use `pg.js` from #200


tgriesser:
`pg.js` is actually [deprecated](https://github.com/brianc/node-postgres-pure#deprecated) since it's now the default for the `pg` module (native bindings are now optional compilation step) - so you can just use the latest `pg` and you should be good to go

knownasilya:
Nice!

PR #825, opened 2015-05-14
Add a queryEnd event that contains the response

pursual:
This allows for better logging and debugging

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -116,6 +116,7 @@ assign(Runner.prototype, {
     this.builder.emit('query', assign({ __knexUid: this.connection.__knexUid }, obj));
     var runner = this;
     return this.client.query(this.connection, obj).then(function (resp) {
+      runner.builder.emit('queryEnd', assign({ __knexUid: runner.connection.__knexUid }, obj));
       return runner.client.processResponse(resp, runner);
     });
   }),


pursual:
Anyone?

rhys-vdw:
Hi @pursual, shouldn't the event get the response object as an argument?

Also `"queryEnd"` feels awkward. How about `"response"`?

Also please add documentation and tests.

chrisbroome:
Should the `__knexUid` be the same in both `this.connection.__knexUid` and `runner.connection.__knexUid`? If so I'd assign it to a temporary variable so that it doesn't change out from under you. The way I see the events working is that you can correlate `query` with `queryEnd` using the `__knexUid`. If this isn't the case, ignore my comment.

pursual:
Perhaps it should be the response object, I haven't explored that. It does work for logging as is, but I am not familiar with the source enough to say if this is the ideal method. I have used this code personally for logging, but it is kind of annoying every time I start a new project to have to insert it into Knex.

rhys-vdw:
@pursual My point was that it should get all relevant information - query and response.

PR will not be merged until tests and documentation are supplied.

ngoctranfire:
When will this be merged in?

rhys-vdw:
@ngoctranfire 

> PR will not be merged until tests and documentation are supplied.

Please read thread before asking questions.

rhys-vdw:
Actually this branch has gone completely stale, it targets the `/lib` directory which no longer exists. If anybody wishes to see this feature implemented please refer to discussion on #1203.

PR #861, opened 2015-06-05
Treat undefined values as null

xaka:
It used to be working in previous versions of knex (0.5.x), but it's broken in 0.8.x. Let's bring back the backward compatibility! :)

### DIFF
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -172,7 +172,7 @@ assign(Builder.prototype, {
 
       // If the value is null, and it's a two argument query,
       // we assume we're going for a `whereNull`.
-      if (value === null) {
+      if (value == null) {
         return this.whereNull(column);
       }
     }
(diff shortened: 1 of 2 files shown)


wubzz:
Is this really how it was in 0.6/0.7? My understanding was that undefined was simply considered redundant, leaving you with greater opportunity for dynamic where-objects being shipped into the querybuilder, without the querybuilder checking undefined columns to be/not be 'null'. If I want it checking column X for null, I would specify column X as null, not as undefined.

My memory could be serving me wrong, tho.

Personally kind of dislike it. Undefined and Null are not the same, and should not be treated as such. My two cents..

xaka:
@rhys-vdw Would you mind to take a look at this one please? For example https://github.com/felixge/node-mysql and https://github.com/brianc/node-postgres do exactly the same. We're in JS world and happen to be in situation where `undefined` and `null` are equal when it comes to translating values from system A to system B.

rhys-vdw:
Hi @xaka, I'm not familiar with the history of this feature, but I'm going to have to agree with @wubzz here. `undefined` and `null` are different concepts. Both JSON and SQL use `null` explicitly for empty values. `undefined` is a JS concept that represents a variable not existing - or the absence of a property.

This would reduce API flexibility and explicitness. Also it is a potentially breaking change for existing code. So thanks for taking the time to provide a PR, but I don't think I can merge this one, sorry.

PR #874, opened 2015-06-23
Fix oracle stream - regression

vschoettke:
The new usage => for function introduced an error in streaming for oracle.


vschoettke:
@tgriesser This may also apply to other places where you changed an inner function to the fat arrow notation.

tgriesser:
Oops, thanks!

Issue #184, opened 2014-02-19
timestamps not storing timezone


cultofmetatron:
The current <code>table.timestamp(name)</code> sets the dateTime type to timestamp. We are on postgresql but would like to be able to use timestamptz instead.

E-Rockalanche:
it would also be great if timestamps() uses the tz type in postgres

PR #876, opened 2015-06-24
default to timestamp with time zone for oracle

vschoettke:
In relation to #184 I also changed the default for `datetime()`, `timestamp()`, `timestamps()`and `time()` to `"timestamp with time zone"` for the oracle dialect. 
Like the postgres implementation, if the `without` param for `datetime(name, [without])` or `timestamp(name, [without])` is set to a truthy value then only `"timestamp"` is generated.


vschoettke:
@tgriesser the code works fine but only updates the code in the src folder and the tests. I think the travis script should build the release (lib) files and then execute the tests. Currently making a pull request that breaks tests does not warn.

I would also favor getting rid of the of the lib folder/any files that are generated by the build script in the repository to avoid extra commits that just change the compiled/generated data.

tgriesser:
Thanks!

> I would also favor getting rid of the of the lib folder/any files that are generated by the build script in the repository to avoid extra commits that just change the compiled/generated data.

The reason I've kept them in is that it allows people to pick specific sha commit-ish as they have in the past in their package.json, rather than only relying on the npm releases.

vschoettke:
@tgriesser If you don't update the build files every time you commit/merge, people will only get the npm releases anyways. Someone may also think a bug is fixed because it says so in the commit but it's still not fixed in the build files. 

I also think the danger of merging something non-working and detecting it way later is too high.

Issue #883, opened 2015-07-01
.where('col', '@>', [val]) .....doesn't work


smithb1994:
The Postgres contains operator '@>' doesn't work because the where statement doesn't allow arrays to be passed. Similar issue reference #879 and #791

```javascript
//get viewed pictures
knex('pictureModels')
    .select('p_id')
    .where('viewed', '@>', ['user2']);
```
`select "p_id" from "pictureModels" where "viewed" @> '{\"user2\"}'`
Notice the escaping on the array, the \ are queried... It should be `"viewed" @> '{"user2"}'`
#869 points this out too.

smithb1994:
So I did a little digging, and it looks like the postgres dialect produces the correct result, but it isn't using it by default. 
```javascript
console.log( knex.client.prepBindings([['cat','dog']]) );
```
outputs:
`'{"cat","dog"}'`

So it looks like there's something modifying the prepBindings before showing it, because the result I get for any raw query or somthing like 

```javascript
knex.where('col', '@>', ['cat','dog'])
```
I get `select * where "col" @> '{\"cat\",\"dog\"}'`
It's escaping with `\"` which we don't want, because instead of `"cat"` being queried, we are getting `\"cat\"` being queried.

smithb1994:
Okay, nevermind... the escaping was because I was sending all my queries .toString() which caused the confusion. I feel dump. :(

Issue #791, opened 2015-04-23
Using prepared statement with `WHERE somecol IN (?)` and array argument?


olalonde:
Is it possible to use prepared statement to do something like:

``` javascript
knex.raw('SELECT id FROM users WHERE id IN (?)', [ [1, 2, 3] ])
```

tgriesser:
No, though I'm looking into expanding the capabilities of argument interpolation. I'll consider this case.

olalonde:
Ok thanks.

tgriesser:
just curious, why not just do `select('id').from('users').whereIn('id', val)`?

olalonde:
I think my statement was a bit more complex than that and didn't have time to figure out the right knex methods Thanks for pointing this out though.

tgriesser:
Actually - so prepared statements don't convert arrays as you'd want there. Instead what I'm going to look at is creating a standard convention of interpolating values (potentially anything with a toSQL fn), so you could have some helper function on knex wrap the array so it knows to expand into the `(?, ?, ?)` format.

smithb1994:
@tgrlesser Any progress on this? I've run into this problem too.
Is there a temporary work around in the mean time?

smithb1994:
@tgriesser 
This feature is much needed for postgres functions array_append(), array_replace(), ect.
Is there a work around for passing arrays into knex.raw() ?

smithb1994:
Nevermind, you can already do what I wanted... I was just stupid enough to be executing my queries by doing .toString first.... I'm an idiot.

rhyek:
Any news on this? Or is this already possible?

elhigu:
Ops, wrong button. Anyways this could be probably considered duplicate since there are other issues about how array values are interpreted.

edorivai:
Usecase (Postgres):

```js
const countries = ['NL', 'BE', 'DE'];
knex('destination').select(
    knex.raw(`(CASE
        WHEN "countryCode" IN :countries THEN 'EU'
        ELSE 'WORLD'
    END) AS "region"`, { countries })
).toString();
```

Result:
```sql
select (CASE
        WHEN "countryCode" IN '{"NL","BE","DE"}' THEN 'EU'
        ELSE 'WORLD'
END) AS "region" from "destination"
```

Expected:
```sql
select (CASE
        WHEN "countryCode" IN ('NL','BE','DE') THEN 'EU'
        ELSE 'WORLD'
END) AS "region" from "destination"
```

[...]

PR #884, opened 2015-07-02
This kind of test case needs to be added

smithb1994:
As of right now, array bindings aren't working correctly. this kind of test needs to be implemented.
 #791 #883


smithb1994:
I'll see if I can find where the bug is first

Issue #869, opened 2015-06-17
It seems StringEscape for postgres is incorrect.


jchbh-duplicate:
According to the document of [String constant in postgres](http://www.postgresql.org/docs/8.3/interactive/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS), 
Unlike Mysql, the "(double quote) does not need to be escape to \", which cause the compare return null if anything match this inputs.

smithb1994:
+1

smithb1994:
@jchbh-duplicate  @tgriesser 
this bug has been causing me so much hassle, I make a hacky work around but it always keeps coming back up in different parts of my code. 
simple statements like `.where('col', '@>', [val])` don't work, because its querying for the `\`

tgriesser:
Closing in favor of #1661

Issue #828, opened 2015-05-15
apostrophes not properly escaped with pg


bblack:
on 0.8.5. apostrophes are escaped with backslashes, rather than the proper double-apostrophe. probably related to https://github.com/tgriesser/knex/issues/658 .

```
knex = require('knex')({client: 'pg'})
q = knex('Beer').where('name', 'ilike', "Mama's Little Yella Pils")
q2 = knex('Beer').where('name', "Mama's Little Yella Pils")
console.log(q.toString())
console.log(q2.toString())
```

> ```
select * from "Beer" where "name" ilike 'Mama\'s Little Yella Pils'
select * from "Beer" where "name" = 'Mama\'s Little Yella Pils'
```

**Comment by kayoub5** [comment]
for anyone have same issue, a temporary solution would be to do this
```sql
ALTER SYSTEM SET standard_conforming_strings=false
```
this will force PG to accept escaping with backslashes.
details:
http://www.postgresql.org/docs/current/static/config-setting.html
http://www.postgresql.org/docs/current/static/sql-altersystem.html
http://www.postgresql.org/docs/current/static/runtime-config-compatible.html#GUC-STANDARD-CONFORMING-STRINGS

victorquinn:
Any status update here? We're having the same problem.

Seems like a rather big problem that a query in Postgres can't contain an apostrophe so I'm surprised there hasn't been much movement here!

bblack:
I've just rebased #886 on master. I'd love to get some comments on it so we can get this fixed.

victorquinn:
I have a fork but would love ring in from @tgriesser -- I wasn't sure on the ideal approach here as the \' is in the sanitizer which has no knowledge of the type of db (Postgres v. other) so I felt like this belonged as a find/replace in the dialect (which would find/replace \' with '' for Postgres queries). Any thoughts?

Cheers,
Victor

bblack:
The escaping takes place in a function exposed by the `src/query/string` module, which is referred to in code as `SqlString`. The approach I took in #886 was to give each dialect its own `SqlString` property, which points to either a dialect-specific module (which I did for postgres), or the existing generic one (all other dialects).

victorquinn:
Nice, that makes sense but I didn't want to make that call because it's not my library and that adds additional complexity.

parkan:
+1 for this, please. This is a very serious bug, especially in a managed environment where the user can't `ALTER SYSTEM`

tgriesser:
[...]

Issue #658, opened 2015-02-03
Wrong escaping of double quotes


sokoloveg:
Example:
```javascript
    knex = new require('knex')({client: 'pg'});
    sql = knex('myTable').insert({myField1: 'Text "qwe" text...', myField2: "Text 'qwe' text..."}).toString();
    console.log(sql);
```
Output:
```
insert into "myTable" ("myField1", "myField2") values ('Text \"qwe\" text...', 'Text \'qwe\' text...')
```

Correct sql should be:
```
insert into "myTable" ("myField1", "myField2") values ('Text "qwe" text...', 'Text \'qwe\' text...')
```

tgriesser:
Closing in favor of #1661

PR #886, opened 2015-07-02
Per-dialect escaping

bblack:
Introduces per-dialect escaping so that e.g. single quotes can be correctly escaped in postgres.

Is this too hamfisted? Would it be better to refactor `src/string.js` so that some of it can be reused, or is it better to expect that the entire thing will be implemented per-dialect?

Addresses #869 and #828, and is a major step toward #658 .


parkan:
:+1:

incorrect single quote escaping with pg is a showstopper for users of managed pgsql like heroku so it would be great to see this merged

yusefnapora:
:+1: yes please!

rhys-vdw:
Hey @bblack. I've just seen this PR now. I'll try to have a look at this in the coming days, but to be honest I probably know the code base less well than yourself, ha.

Out of interest, could I ask why I've never encountered any problems with postgres and knex before? (I've certainly stored many apostrophes in my time)

bblack:
@rhys-vdw I doubt I know the code base better than you--this pull request is the first time I've looked at it.

As for why you haven't seen any problems with postgres and knex before: my guess is you've either used an older version of postgres than 9.1, or you have `standard_conforming_strings` off. (see http://www.postgresql.org/docs/9.2/static/runtime-config-compatible.html#GUC-STANDARD-CONFORMING-STRINGS and http://www.postgresql.org/docs/9.2/static/sql-syntax-lexical.html)

As a separate but related issue - those docs suggest that the SQL standard way to escape a single quote is by preceding it with another single quote. If true, it seems like that would be a better default behavior in knex.

chrisbroome:
Yeah the SQL standard way is to use two consecutive single quotes to escape a single quote in a string. This *should* be supported by every dialect.

rhys-vdw:
@bblack Sorry I haven't reviewed this one yet. Does this help solve #1021 perchance?

bblack:
Based on my understanding, no, but I can take a closer look later this week.

bblack:
@rhys-vdw i've pushed a commit that i think will solve #1021.

rhys-vdw:
Alright. @bblack, I've just reviewed the PR. For anyone interested, I discovered you can filter out the whitespace changes with this GitHub query param [`?w=1`](https://github.com/tgriesser/knex/pull/886/files?w=1).

If we are to take this approach it will require some changes. I'm uncomfortable with the amount of duplication between `src/dialects/postgres/string.js` and `src/string.js` as it stands.

I've just spent a lot of time trying to get familiar with the code and the nature of the problem, and I must admit I haven't got my head around it entirely.

> As for why you haven't seen any problems with postgres and knex before: my guess is you've either used an older version of postgres than 9.1, or you have standard_conforming_strings off.

If I were inserting and retrieving PostgreSQL via knex, would I see the added `\` on the way out? Or does knex filter this out, hiding the problem? Because I was definitely using a more recent version of PostgreSQL with default configuration.

Either way, @chrisbroome suggests that `''` is the standard escape for single quotes and is supported by other dialects. If this is the case then perhaps it's worth trying this change universally in `QueryString` for starters? If that works then #828 is down at least

src/dialects/maria/index.js
```diff
@@ -21,6 +21,8 @@ assign(Client_MariaSQL.prototype, {
 
   driverName: 'mariasql',
 
+  SqlString: require('../../query/string'),
```
elhigu: This should not be needed since `src/client.js` base class already sets it to `src/query/string.js`

src/dialects/mysql/index.js
```diff
@@ -34,13 +34,15 @@ assign(Client_MySQL.prototype, {
...
   QueryCompiler: QueryCompiler,
-  
+
   SchemaCompiler: SchemaCompiler,
 
   TableCompiler: TableCompiler,
 
   ColumnCompiler: ColumnCompiler,
 
+  SqlString: require('../../query/string'),
```
elhigu: This should not be needed since `src/client.js` base class already sets it to `src/query/string.js`

src/dialects/mysql2/index.js
```diff
@@ -26,11 +26,13 @@ assign(Client_MySQL2.prototype, {
   // The "dialect", for reference elsewhere.
   driverName: 'mysql2',
 
+  SqlString: require('../../query/string'),
```
elhigu: This should not be needed since `src/client.js` base class already sets it to `src/query/string.js`

src/dialects/oracle/index.js
```diff
@@ -38,6 +38,8 @@ assign(Client_Oracle.prototype, {
     return require('oracle')
   },
 
+  SqlString: require('../../query/string'),
```
elhigu: This should not be needed since `src/client.js` base class already sets it to `src/query/string.js`

src/dialects/postgres/index.js
```diff
@@ -34,6 +34,8 @@ assign(Client_PG.prototype, {
 
   SchemaCompiler: SchemaCompiler,
 
+  SqlString: require('./string'),
```
elhigu: `string.js` should be moved to `./query/string.js`

src/dialects/postgres/string.js
```diff
@@ -0,0 +1,115 @@
+'use strict'
```
elhigu: This should be made to be inherited from `src/query/string.js`, and both, base and inherited instance should be refactored to only override small part of class in inherited implementation. (SqlString must be instantiated as well to make inheritance work nice)

src/dialects/sqlite3/index.js
```diff
@@ -39,6 +39,8 @@ assign(Client_SQLite3.prototype, {
 
   TableCompiler: TableCompiler,
 
+  SqlString: require('../../query/string'),
```
elhigu: Shouldn't be necessary to set here since client base implementation already does that.

src/dialects/websql/index.js
```diff
@@ -20,6 +20,8 @@ inherits(Client_WebSQL, Client_SQLite3);
 
 assign(Client_WebSQL.prototype, {
 
+  SqlString: require('../../query/string'),
```
elhigu: Shouldn't be necessary to set here since client base implementation already does that.

elhigu:
Hi @bblack thanks for taking time to implement this! The approach of having separate `SqlString` implementation for each client sounds good. 

However as @rhys-vdw said there should be more shared functionality between base implementation and client implementation. Usually `knex` has been using inheritance to implement it check e.g. `dialects/*/query/compiler.js` or `dialects/*/transaction.js` etc.

Also`SqlString.escape` should be refactored in a way that implementation specific parts are moved to separate function, so that it is enough to rewrite smaller part in client specific implementation.

About the code I have two notes here and rest are in comments in  `changed files` tab.

1. In base implementation `string.js` is located in `src/query/string.js` so dialect specific class should be in `dialects/*/query/string.js` instead of `dialects/*/string.js`.

2. In few places one should use clients implementation of SqlString instead the one coming from require:
https://github.com/tgriesser/knex/pull/886/files#diff-cf27c1d543e886c89cd9ac8b8aeaf05bR127
https://github.com/tgriesser/knex/pull/886/files#diff-c8ff07bea2af41f7c0247b860dbd9d39R67
https://github.com/tgriesser/knex/pull/886/files#diff-c25c8a30a271df8c36bc7249d7b2be87R81

zacronos:
FWIW, I've been using knex a lot with Postgres 9.4.4 with `standard_conforming_strings` = `on`, and I've certainly stored/retrieved single-quotes via knex without running into an error.  And in fact, much of my use is on heroku, so I'm not sure why this bug would be a showstopper as @parkan suggested.

With that said, I don't want to confuse things -- it sounds like this PR is moving toward something good, so the question of when the bug is actually a problem may be moot.  I just wanted to add another data point since there is some confusion over when the problem raises its ugly head.

parkan:
@zacronos: I just tried some simple inserts and queries again and they do seem to succeed; however, we were definitely seeing errors like below with default heroku configuration (also 9.4.4/standard conforming strings on)

```
"insert into "tags" (name, user_id) select 'Shaquille O\'Neal', '66844f26-743f-11e5-a03f-c39c8223bd8f' where not exists (select 1 from "tags" where "name" = 'Shaquille O\'Neal' limit '1') - syntax error at or near "', '""
```

I think this may have to do with using `raw` calls, will investigate.

In the meantime, we've been running the @bblack fork and have not run into any trouble.

studds:
Like @parkan, I saw this error with raw calls, but haven't seen it otherwise. I used the `ALTER DATABASE SET standard_conforming_strings=false` workaround.

bblack:
@rhys-vdw @elhigu do you guys prefer to squash the changes or simply append new commits? (i have to rebase anyway to resolve conflicts with master).

elhigu:
Just rebase + add new commits is fine for me. It might make rebasing easier if you have commits, which just updates generated code separated from real changes. I usually drop generated code commits when doing `rebase -i`.

bblack:
I've addressed your comments (@rhys-vdw and @elhigu) with additional commits, and rebased to resolve conflicts.

The first few commits include build objects, and the last commit doesn't--is this okay, or do I need to `rebase -i` and remove those changes?

FrankZZ:
+1

elhigu:
@rhys-vdw @bblack Looks good to me. I like the way that class is kept as a set of static functions and just partly overridden :+1:

Vanuan:
It is actually fixed? I see that test still uses backslash for escaping:

```
      postgres: 'select * from "users" where "last_name" = \'O\'\'Brien\'',
```

bblack:
that's javascript escaping, since the string literal is given in single quotes. the resulting query looks like
```
select * from "users" where "last_name" = 'O''Brien'
```

PR #890, opened 2015-07-07
Fixed migration issue renameColumn is failed when there is foreign key reference in another table

taurencoder:
link to issue tgriesser/knex/issues/889


rhys-vdw:
Thanks @taurencoder. Nice work.

PR #896, opened 2015-07-13
fixed connection error handling for oracle dialect

helios1138:
right now connection error on oracle in the terminal looks like this:
```
TypeError: the target of promisifyAll must be an object or a function
```

### DIFF
--- a/lib/dialects/oracle/index.js
+++ b/lib/dialects/oracle/index.js
@@ -74,8 +74,8 @@ assign(Client_Oracle.prototype, {
     var client = this;
     return new Promise(function (resolver, rejecter) {
       client.driver.connect(client.connectionSettings, function (err, connection) {
-        Promise.promisifyAll(connection);
         if (err) return rejecter(err);
+        Promise.promisifyAll(connection);
         if (client.connectionSettings.prefetchRowCount) {
           connection.setPrefetchRowCount(client.connectionSettings.prefetchRowCount);
         }


rhys-vdw:
Looks like a straight forward change. Thanks.

vschoettke:
@rhys-vdw @helios1138 changed files in lib/ will be overwritten when the next compilation of the src/ happens.

@tgriesser you should really think about getting rid of the lib/ folder in the repository.

rhys-vdw:
@vschoettke I should be more careful. Should I force rollback the merge from master?

@helios1138 Could you please reapply these changes to the `/src` folder. Sorry, I missed that.

rhys-vdw:
@vschoettke Isn't it necessary that `/lib` is included for use as a dependency on other projects?

vschoettke:
@rhys-vdw It's only necessary if somebody uses the repository url as version in the package.json. If someone wants the latest (non-published) version a fork and building the libs themselves is still an option. Otherwise you have to maintain `/src` and `/lib` all the time.

For versions available via npm this is no problem because the built version is published.

Maintaining `/src` and `/lib` is also extra work because you have to test if the compiled that somebody commited to `/src`code actually passes because it's not currently tested with the unit tests. And like in your case accidentally accepting a `/lib` only fix will be problematic.

Also see #876 for my comments about that.

rhys-vdw:
@vschoettke I agree that is bad to have `/lib` under version control. But npm does allow direct commit targeting, and it is a feature I've used with knex after bugfixes in the past.

I wonder if it's possible to trigger a build (ie. populate `/lib`) as part of `npm install`. That seems the most appropriate behaviour.

vschoettke:
@rhys-vdw Yes It's possible to use the postinstall script. This also requires install babel as normal dependency. But if there is one generated package.json for publish (without babel) and one for the repository I think this would be a viable solution.

rhys-vdw:
> But if there is one generated package.json for publish (without babel) and one for the repository I think this would be a viable solution.

For publishing on services other than npm?

vschoettke:
Not other services, what I mean is that the `package.json` in the repository here includes babel as normal dependency and has the build step, and that the npm published version does not. In the published version the /src directory can also be omitted.

rhys-vdw:
@helios1138, are you planning to fix this PR to target `src/`. I've got to clean master before merging more. I can rewrite the changes myself if you haven't the time.

helios1138:
@rhys-vdw you mean like this: https://github.com/tgriesser/knex/pull/905?

Issue #889, opened 2015-07-07
Migration: cannot rename master table's column name if there's foreign key reference - Mysql


taurencoder:
I have upgraded knex from `0.7.*` to `0.8.6` recently and found our existing migrations doesn't work any more. There's some error thrown when exec the renameColumnName migration:
```
migrations failed with error: alter table `activities` drop foreign key `activity_categories_activity_id_foreign` - ER_CANT_DROP_FIELD_OR_KEY: Can't DROP 'activity_categories_activity_id_foreign'; check that column/key exists
```
After some investigation I found that actually the foreign key is not exist on master table `activities` but on its slave table `activity_categories`

taurencoder:
It appears to me that the issue is located in `dialects/mysql/schema/tablecompiler.js`

1. getFKRefs() function: the RC table should limit its `CONSTRAINT_SCHEMA` to current connected database, otherwise there's duplicated refs records returned
2. dropFKRefs() function: the table where foreign key to be dropped from should be `ref.TABLE_NAME` instead of `this.tableName()`
3. createFKRefs(): same as dropFKRefs(), use `ref.TABLE_NAME` instead of `this.tableName()`; and the fkName should be `ref.CONSTRAINT_NAME` instead of `ref.COLUMN_NAME`

I'm trying to submit an pull request to fix this..

PR #900, opened 2015-07-16
re-create fk with ref.CONSTRAINT_NAME instead of ref.COLUMN_NAME

taurencoder:
It's my fault that I missed one change in tgriesser/knex#889.


rhys-vdw:
Cheers.

PR #901, opened 2015-07-16
Add builder as first parameter to `queryBuilder.modify`

jevakallio:
As discussed in https://github.com/tgriesser/knex/pull/881#issuecomment-121974392


rhys-vdw:
Thanks!

PR #903, opened 2015-07-20
add LiveScript support back in to knex migrations

beppu:
We upgraded the version of knex we were using, and noticed that the migration system received a big overhaul since the last time we looked.  Unfortunately for us, it lost LiveScript support which our migrations were using, so this pull request should add it back in.

### DIFF
--- a/lib/bin/cli.js
+++ b/lib/bin/cli.js
@@ -71,7 +71,7 @@ function initKnex(env) {
 function invoke(env) {
 
   var pending,
-      filetypes = ['js', 'coffee'];
+      filetypes = ['js', 'coffee', 'ls'];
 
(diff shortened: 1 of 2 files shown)


khoerling:
+1

rhys-vdw:
Nice work. Thanks.

MadcapJake:
Shouldn't this be added to `src` version too?

beppu:
@MadcapJake I'm not sure what you mean by `src` version, but I did just realize I didn't provide a stub for the `seed:make` subcommand.  If that's what you meant, then, yes I do need to add that.

If you meant something else, could you clarify, please.

MadcapJake:
Doesn't this project build from `src` into `lib`? Because if so, you modified the built version rather than the source version.

beppu:
@MadcapJake You're right.  I totally missed that.  I guess I didn't expect built files to be checked in.

Anyway, I will undo the previous commits and redo them the right way.  Thanks for noticing that.

MadcapJake:
No prob!  It looks like other commits may have done the same thing so take a look at a diff/status of `lib` and you might want to commit only the changes you made.

rhys-vdw:
@beppu Gosh. That's the second time I've merged something targeting `/lib`. I need to be more careful.

PR #908, opened 2015-07-23
Add select lock arguments

promag:
This feature allows to specify `OF table_name` and `NOWAIT`in a `SELECT`query with `FOR UPDATE` and `FOR SHARE` when the dialect is `postgres`.

Reference:
`[ FOR { UPDATE | SHARE } [ OF table_name [, ...] ] [ NOWAIT ] [...] ]`

### DIFF
--- a/index.html
+++ b/index.html
@@ -1312,7 +1312,7 @@ knex.transaction(function(trx) {
 </pre>
 
     <p id="Builder-forUpdate">
-      <b class="header">forUpdate</b><code>.transacting(t).forUpdate()</code>
+      <b class="header">forUpdate</b><code>.transacting(t).forUpdate([tables], [nowait])</code>
       <br />
       Dynamically added after a <a href="#Builder-transacting">transaction</a> is specified,
       the <tt>forUpdate</tt> adds a <tt>FOR UPDATE</tt> in PostgreSQL and MySQL during a select statement.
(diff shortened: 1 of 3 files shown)


promag:
@tgriesser WDYT?

promag:
@rhys-vdw WDYT?

elhigu:

@promag This looks like a good addition, do you think this is still relevant? If so, conflicts needs to be fixed.

promag:
@elhigu Sure! Not sure if I should include just the first commit.

elhigu:
@promag I reviewed the code, looks like the test case is missing. Also there was some strangeness when reading arguments and documentation. It did look like you are supporting actually following type of parameters:

```js
.forUpdate('table1')
.forUpdate('table1','table2') 
.forUpdate(true) 
.forUpdate('table1','table2',true)
``` 

I'll leave doc suggestion in line comments.

index.html
```diff
@@ -1312,7 +1312,7 @@ <h3 id="Builder-wheres">Where Clauses</h3>
 </pre>
 
     <p id="Builder-forUpdate">
-      <b class="header">forUpdate</b><code>.transacting(t).forUpdate()</code>
+      <b class="header">forUpdate</b><code>.transacting(t).forUpdate([tables], [nowait])</code>
```
elhigu: Maybe this is a bit more clear spec of arguments `.forUpdate('table1', 'table2',... , [nowait])` parameter reading did seem so creative, that it could be good to explain with examples all the variations.

elhigu:
@promag Any chances of getting those examples about possible parameters, since that `forUpdate([tables], [nowait])` is a bit ambiguous?

elhigu:
Should be implemented in docs repo. Adding issue there and link to this PR.

Issue #915, opened 2015-07-25
Feature request: `join table as alias`


olalonde:
Would be nice if the `.join` api exposed an `.as` method to alias the table being joined. Would be useful for queries like `knex('users').join('users', function () { this.as('referrer'); })`.

olalonde:
Sending a PR soon.

olalonde:
Ok so I implemented the feature but when executing the query, the rows don't parse correctly into an object.

olalonde:
Oh ok, the problem was that both table had same column names and were "shadowing" each others. Fixed by "manually" prefixing all joined columns with `.select()`. Ready to send PR.

elhigu:
@olalonde  I was just going through old PRs looks like the one you sent didn't work so I'm closing it for now. This would be nice addition to have. 

In addition to code and test PR should also have `index.html` update with documentation.

absolux:
Something like below does not work ?
```js
 knex('users').join('users as second', 'users.id', 'second.other_id') 
```

elhigu:
@absolux yeah it works, I suppose this can be closed...

PR #916, opened 2015-07-25
.as method on JoinClause objects. See issue #915

olalonde:
See issue https://github.com/tgriesser/knex/issues/915. Oops sorry, just noticed vim auto deleted white characters.


elhigu:
Added comment to #915 that more complete / working PR would be needed.

PR #917, opened 2015-07-27
PostgreSQL native support

nkt:


pierreinglebert:
Look at this PR why it's been refused previously : https://github.com/tgriesser/knex/issues/575

tkellen:
:+1:

nkt:
@rhys-vdw ping

rhys-vdw:
Hi @nkt. I've read through the related past issues. (Thanks for the link, @pierreinglebert.)

The rejection reason for the previous PR is [here](https://github.com/tgriesser/knex/pull/577#issuecomment-70412269). Tests were failing because streaming API is/was unsupported by pg-native.

I'll defer to @bendrucker's requirements: that the streaming error, if fired, be caught and and another error with a clear explanation be thrown. However, currently the streaming tests are passing in your branch. Does this mean that pg-native now supports streaming, or that tests are not actually running using pg-native? If the latter is the case, I think we'll need test coverage to be expanded to run both pg and pg-native before merging into master.

zacronos:
Clearly something has changed -- the substance of the 2 PRs is functionally the same, but the prior PR was causing tests to fail, where this PR is passing all tests.

I've tried to dig in and see what has changed, but I don't see anything on the surface.  The streaming test that failed before is unchanged.  The `pg-query-stream` library has had some updates, but there doesn't seem to be mention of supporting `pg.native`; in fact, according to the author, this would require major changes: https://github.com/brianc/node-pg-query-stream/issues/8.  Though of course it is possible that is no longer true, perhaps with an underlying change to one of its dependencies, such as `pg-cursor`.

So, it seems unlikely the tests are running with `pg.native`.  I don't understand why this didn't come up before, but I would guess that `pg-native` needs to be added to `devDependencies` in order for the tests to use `pg.native`.

zacronos:
Regarding the requirements from @rhys-vdw: I think that is good.  However, if it is still impossible to stream while using `pg.native` (which I am assuming is the case until it is proven otherwise), then no PR which simply uses `pg.native` by default and falls back to `pg` (as this PR does) will be acceptable, even with the added error handling -- existing apps that use streaming and happen to have `pg.native` installed would break.

An acceptable PR would also need to provide some option for controlling the use of `pg.native`.  The default should still be `pg`, for backward compatibility with apps that use streaming, and there should be an option available to explicitly tell knex to use `pg.native` (in which case I would imagine it is preferable to throw if `pg.native` isn't available, rather than silently fall back to `pg`).  Note that this approach should also make it more straightforward for test coverage to include both `pg` and `pg.native`.

Last note: With regard to running tests on both `pg` and `pg.native`, is there a clean way to allow the streaming calls to fail (with a specific, expected error) when run on `pg.native`, even though those same tests should pass for `pg` and other dialects?  That could be the toughest part of this.  If @nkt and others don't want to work on the more complex PR that's been outlined, I may put some time toward it -- but if so, I would appreciate some help figuring out how to approach this testing issue.

rhys-vdw:
> However, if it is still impossible to stream while using pg.native (which I am assuming is the case until it is proven otherwise), then no PR which simply uses pg.native by default and falls back to pg (as this PR does) will be acceptable, even with the added error handling

Agree 100%. One possible approach would perhaps be to add a different client name (eg. `pgNative`). Perhaps even an entirely new `Client` type that disables streaming. It could just inherit from [`Client_PG`](https://github.com/tgriesser/knex/blob/5e65678b32fa639843902d39603ab2050d84f10f/src/dialects/postgres/index.js) and override the [`_driver`](https://github.com/tgriesser/knex/blob/5e65678b32fa639843902d39603ab2050d84f10f/src/dialects/postgres/index.js#L39-L41) to return `pg.native`, and replace [`_stream`](https://github.com/tgriesser/knex/blob/5e65678b32fa639843902d39603ab2050d84f10f/src/dialects/postgres/index.js#L105-L113) with a function that throws a detailed error message.

I think the streaming tests are [here](https://github.com/tgriesser/knex/blob/5e65678b32fa639843902d39603ab2050d84f10f/test/integration/builder/selects.js#L107-L136) and also there seems to be a test [here](https://github.com/tgriesser/knex/blob/5491e7083cc8e7cdaf54d98a0efd121c310003fb/test/tape/stream.js). I know that Tim in general is preferring to move towards using Tape for tests, but the framework doesn't seem so well fleshed out yet - so do what's easier. I think maybe separating the streaming tests from `selects.js` and moving them into a new stream file to be included [here](https://github.com/tgriesser/knex/blob/5e65678b32fa639843902d39603ab2050d84f10f/test/integration/suite.js) would be appropriate.

Or alternatively you could do something like this:

```js
```

zacronos:
That sounds like a great plan.  I haven't spent a lot of time looking around the knex codebase, so I really appreciate the detailed suggestions (and links to the appropriate locations in the code).

rhys-vdw:
@zacronos Nor have I, so you may find a cleaner solution. Holler if you'd like to discuss anything else.

elhigu:
Has anyone actually benchmarked, what is performance difference between `pg` and `pg-native`? According to our profiling, overhead caused by `pg` has been really small compared to all other parts of execution...

PR #919, opened 2015-07-27
added LiveScript support for migrations correctly #903

beppu:
Previously, I had directly modified files in `lib/` because I didn't realize they were generated from files in `src/`.  I have since reverted those changes and changed the right files, and went through the build process to generate the files that needed to be generated.

Apologies to @rhys-vdw and thanks to @MadcapJake for noticing my mistake before it was too late.


rhys-vdw:
Thankyou. :-)

Issue #892, opened 2015-07-07
Timed-out connection causes crash


mfhholmes:
Hey,

I'm using express-sessions, connect-session-knex, and mysql2 modules. Mysql database running on localhost on Ubuntu VM. Left the VM running overnight, first test of the day crashed the server with:

```
Unhandled rejection Error: select `sess` from `sessions` where `sid` = '35983279-598a-4e88-8f1e-d9276e3859c1' and CAST('2015-07-07T13:24:05.376Z' as DATETIME) <= expired - This socket has been ended by the other party
    at Socket.writeAfterFIN [as write] (net.js:297:12)
    at Connection.write (/vagrant/national/node_modules/mysql2/lib/connection.js:136:15)
    at Connection.writePacket (/vagrant/national/node_modules/mysql2/lib/connection.js:150:10)
    at Query.start (/vagrant/national/node_modules/mysql2/lib/commands/query.js:39:14)
    at Query.Command.execute (/vagrant/national/node_modules/mysql2/lib/commands/command.js:34:20)
    at Connection.handlePacket (/vagrant/national/node_modules/mysql2/lib/connection.js:310:28)
    at Connection.addCommand (/vagrant/national/node_modules/mysql2/lib/connection.js:326:10)
    at Connection.query (/vagrant/national/node_modules/mysql2/lib/connection.js:397:15)
    at /vagrant/national/node_modules/knex/lib/dialects/mysql/index.js:92:18
    at tryCatcher (/vagrant/national/node_modules/knex/node_modules/bluebird/js/main/util.js:24:31)
    at Promise._resolveFromResolver (/vagrant/national/node_modules/knex/node_modules/bluebird/js/main/promise.js:462:31)
    at new Promise (/vagrant/national/node_modules/knex/node_modules/bluebird/js/main/promise.js:69:37)
    at Client._query (/vagrant/national/node_modules/knex/lib/dialects/mysql/index.js:88:12)
    at Client.query (/vagrant/national/node_modules/knex/lib/client.js:127:24)
    at Runner.<anonymous> (/vagrant/national/node_modules/knex/lib/runner.js:118:24)
    at Runner.tryCatcher (/vagrant/national/node_modules/knex/node_modules/bluebird/js/main/util.js:24:31)
```

Looks to me like the connection timed out, the websocket got closed by MySQL, and Knex didn't pick it up and reconnect. Hence this looks like a knex issue rather than an express-sessions issue - happy to post the issue there if I'm wrong ;)

[...]

PR #922, opened 2015-07-28
Provide Pool2 a ping method to test a connection

vellotis:
#892

This is how it should be done for Pool2. Every client dialect must implement its ping method for the used driver. At the moment only mysql/2, maria and pg native drivers are capable of doing this kind of check. 

If ping method fails then connection will be destroyed from the pool. Otherwise (like at the moment) Pool2 resource pool thinks, that the connection is still valid. So it tries to use it for the queued requests over and over again without releasing it and acquiring a new one.

Sadly I don't have time to set up all the databases to run the tests. I would be glad if somebody could run them and confirm the merge.


zachary822:
I would like to, as a user, add to this that it works nicely for mysql2.
I changed the database wait_timeout and interactive_timeout to 1 second.
The original knex would throw an error saying "This socket has been ended by the other party".
It does not happen with this commit.

indeyets:
@rhys-vdw can you check this one please? (sorry if I'm wrong to ping you, but it looks like you're doing most of merges these days)

rhys-vdw:
@indeyets No problem. I saw this email but haven't gotten around to reviewing yet. Just having a look now...

So this provides a generic way to ping a knex instance to check if it's up. I've got a few questions...

> If ping method fails then connection will be destroyed from the pool. 

Where is the code that handles this? I can't see it... :-S

@zachary822:
> The original knex would throw an error saying "This socket has been ended by the other party".
> It does not happen with this commit.

How could this be the case? Wouldn't that only happen if you luckily called `ping` after the db went down, but before your next query?

> Sadly I don't have time to set up all the databases to run the tests. I would be glad if somebody could run them and confirm the merge.

There are no new tests. Tests do appear to be failing in Travis, but I'm not sure if it's related.

[This concerns me](https://github.com/vellotis/knex/blob/ft-pool2-ping-connection/src/client.js#L192-L198). Is it okay to say the database is working when a client does not provide a `ping` feature? I'd probably prefer an NYI or even just an undefined exception here, unless there is some reasoning for this.

Steps to getting this merged (with my apologies to @vellotis for delayed review):

 - Answer above questions for me.
 - Add support for every dialect.
 - Add tests for `ping` in all use cases.
 - Add documentation explaining the feature to users.

zachary822:
I think it has something to do with [this](https://github.com/myndzi/pool2/blob/master/lib/pool.js#L86-L86) in Pool2. You can provide a ping function to pool2 for it to call when it needs to ping the database (haven't got around to seeing when it does ping the database). The default pool2 ping implementation looks like a noop.

vellotis:
As I have a need to deal with knex again I merged current tgriesser master into it.
As well I would write these missing tests. To follow some good practise, under where should I write these tests?

wubzz:
@vellotis @rhys-vdw From my understanding this will expose a ping method on the knex client to check if the connection is still valid? 

If so, then in regards to #1198 and its related issues, I would prefer and recommend that we use `Pool2` to handle this internally in knex instead of having the user manually run ping commands. 

Having default ping functions would be needed either way, so the functions created in the PR are still valid, might just need some tweaks. I'm just implying they would be better of used as defaults for `config.pool.ping` when it is undefined.

vellotis:
@wubzz, you are correct. #1198 is really closely related to current PR. I ran to the same problem long time ago. And I'm willing to contrubute to solve this.

I would just need some advice where should I write the tests. So everyone could be happy to finally 

@rhys-vdw Pinging methods:
  - maria :white_check_mark:
  - mssql :grey_question:
```javascript
    @driver.Connection::connect( config, cb )
```
  - mysql/mysql2 :white_check_mark:
  - postgres :white_check_mark:
  - oracle :grey_question: 

    https://github.com/oracle/node-oracledb/blob/02f74fabf09f039030c502cef9d74093b440a46f/doc/api.md#connectionhandling
```javascript
    oracledb.getConnection( settings, cb )
```
  - strong-oracle :grey_question:

    https://github.com/oracle/node-oracledb/blob/02f74fabf09f039030c502cef9d74093b440a46f/doc/api.md#connectionhandling
```javascript
    oracle.connect( settings, cb )
```
  - sqlite3 :question: 
  - websql :x:  (Depracted since November 18, 2010)

@rhys-vdw @ "This concerns me". I don't see the actual reference any more as the reference is against master branch not commit. But I suppose that the problem is resolved by https://github.com/myndzi/pool2/blob/2130ba5162de17b9311228ef97d071f7739ab1ce/lib/resource-request.js what is the outcome of my discussion @ https://github.com/myndzi/pool2/issues/15

wubzz:
@vellotis Talked this over in #1198 and currently the idea is:

- Add default `ping` method for *all* dialects
- Ping method can be overwritten by explicitly defining `config.pool.ping` as a function, otherwise always use the dialect default.

Have you tested the ping methods in your PR against cases such as killing the connection on the server side? My assumption was that running a `SELECT 1` query would be the most efficent ping test, but I could very well be wrong.

Example postgres with pgAdmin:

```javascript
knex.raw('SELECT version()')
.then(() => {
debugger; //pgAdmin -> server status -> kill all connections
return knex.raw('SELECT version()')
})
.then(() => {
})
.catch((err) => {
//Should not reach this point
});
```

PR #928, opened 2015-08-03
Fixes an issue where a transaction never rolls back as expected.

OronNadiv:
This PR fixes an issue where a connection may have a transaction that was never rolled back.

Steps to reproduce:
1. Acquire connection.
2. Create transaction (BEGIN).
3. Execute [`container(transactor)`](https://github.com/tgriesser/knex/blob/master/lib/transaction.js#L38).
4. Instead of returning a promise, [`container(transactor)`](https://github.com/tgriesser/knex/blob/master/lib/transaction.js#L38) throws an exception.

Result: Transaction will not be rolled back.  Next time someone uses the same connection, their changes will be done as part of the transaction.  This may result with the changes that are not saved to the DB.

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -34,19 +34,15 @@ function Transaction(client, container, config, outerTx) {
     init.then(function () {
       return makeTransactor(_this, connection, trxClient);
     }).then(function (transactor) {
-
-      var result = container(transactor);
-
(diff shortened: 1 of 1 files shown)


jurko-gospodnetic:
@wubzz - this one can be closed as the underlying problem has been resolved by pull request #1257.

PR #946, opened 2015-08-24
Postgres question mark escaping support

### LINKED ISSUE #519: Conflict with Postgresql jsonb operators

elhigu:
Fixes #519 and #888 by adding support for `\\?` escape sequence to postgresql raw queries, which does not have bindings given. 

Uses solution based on https://github.com/tgriesser/knex/compare/qs-escapement which converts escaped question marks to place holders before replacing `?` with `$1..$n` variables of postgresql prepared statement.

Plain ignoring position binding replacement if no bindings was given (suggested in #888) did not work well, because in the end where replacement is done, may have bindings from various where statements as shown in example below.

After final SQL is created, placeholders are converted to single question marks.

e.g.
```js
knex.from('table')
  .whereRaw("(json->'field')::jsonb \\?& array['keyOne', 'keyTwo']")
  .where('id', '>', 1)
```
Is processed in few steps:
```
1# select * from "table" where (json->'field')::jsonb \\?& array['keyOne', 'keyTwo'] and "id" > ?
2# select * from "table" where (json->'field')::jsonb $Q& array['keyOne', 'keyTwo'] and "id" > ?
3# select * from "table" where (json->'field')::jsonb $Q& array['keyOne', 'keyTwo'] and "id" > $1
4# select * from "test_table_two" where (json->'field')::jsonb ?& array['keyOne', 'keyTwo'] and "id" > $1
```


soldair:
this looks like it would be a good enough fix for #888 too.

rhys-vdw:
I hate to break it to you, @elhigu, but none of your commits target source files. They're all changes to the compiled output. If you were to compile now you'd lose everything. You'll need to somehow retarget all these changes on `/src`.

rhys-vdw:
@elhigu. Sorry, I see now that that's all the changes from previous commits. Looking now:

I wonder if we can't do something like this [here](https://github.com/elhigu/knex/blob/postgres-question-mark-escaping-support/src/dialects/postgres/index.js#L106-L113):
```js
  positionBindings: function(sql) {
    var questionCount = 0;
    return sql.replace(/(?!\\)(\?)/g, function() {
      questionCount++;
      return '$' + questionCount;
    });
```

That's two fewer calls to `replace`, and two fewer strings created as a temporary value.

elhigu:
@rhys-vdw I think the written solution is more clear. I don't believe that this regex is bottle neck in execution, so call count should not be relevant. Though of course I will do the change if necessary to get this merged :) Also estimating performance of regex is quite hard without testing.

elhigu:
@rhys-vdw I got interested to see actual difference and wrote fixed version of both expressions. Looks like actually making 3 simple calls instead of one more complex one is slightly faster.

1 replace version (I had to add some fixes there to make \? -> ? conversion happen and to fix questionCount incrementing to be ignored if escaped ? was found):

```js
Mikaels-MBP:~ mikaelle$ cat test1.js ; time node test1.js 
a = function(sql) {
  var questionCount = 0;
    if (escapes.length % 2) { 
      return '?'; 
    } else {
      questionCount++;
      return '$' + questionCount;
    }
  });
}

console.log(a("What we have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ?"));

for (var i=0; i < 1000000; i++) {
  a("What we have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ?");
}
What we have here ? $1 have here ? $2 have here ? $3 have here ? $4 have here ? $5 have here ? $6 have here ? $7 have here ? $8

real	0m6.511s
user	0m6.484s
sys	0m0.070s
````

3 replaces version:
```js
Mikaels-MBP:~ mikaelle$ cat test2.js ; time node test2.js 
a = function(sql) {
    var questionCount = 0;
    return sql.replace(/\\\?/g, function() {
      return '$Q';
    }).replace(/\?/g, function() {
      questionCount++;
      return '$' + questionCount;
    }).replace(/\$Q/g, function() {
      return '?';
    });
}

console.log(a("What we have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ?"));
for (var i=0; i < 1000000; i++) {
  a("What we have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ? have here \\? ?");
}
What we have here ? $1 have here ? $2 have here ? $3 have here ? $4 have here ? $5 have here ? $6 have here ? $7 have here ? $8

real	0m5.168s
user	0m5.142s
sys	0m0.058s
Mikaels-MBP:~ mikaelle$ 
```

elhigu:
@rhys-vdw I gave this one more thought... Actually one regex has an advantage over the three separate ones. Now if user has `$Q` string in his query, it will be replaced by `?` even that it was not originally `\?`, so the solution without temporary `$Q` is more compatible.

elhigu:
Updated pull request with more compatible substitution suggested by @rhys-vdw

rhys-vdw:
@elhigu. Thanks for profiling that. I had incorrectly assumed that building three whole query strings would be more expensive. Good example of why profiling is always necessary when discussing performance.

I hadn't considered the other advantage, but I'm glad I provided something useful. :wink:

This all looks great.

PR #952, opened 2015-08-27
Specify how migration cli chooses environment

tschuy:
I don't know why this wasn't immediately obvious to me, but it wasn't.

EDIT: actually, looks like there is a command line flag that isn't documented. I've added a short sentence on it.


rhys-vdw:
Nice. I too was unaware of this flag.

PR #956, opened 2015-08-30
Improve tests and docs for migrate API

harriha:
Summary of changes:

- ~~Fix some typos from migrate API docs~~ (someone else was faster)
- Add a mention about `disableTransactions` config option
- Add descriptions for each config option
- Remove mention of config option `database` (couldn't find out where that would be used, hope I got this right?)
- Add basic `tape` tests for the migrator

Proof-reading the config option descriptions is probably a good idea.

The changes to actual code are sort-of cosmetic, just thought to make it easier to find the config default values. Was thinking of asserting for valid values in `setConfig()`, but didn't want to start breaking apps which are potentially feeding in invalid/old config values.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2891,6 +2891,7 @@ $ knex seed:run
       <li><tt>directory</tt>: a relative path to the directory containing the migration files (default <tt>./migrations</tt>)</li>
       <li><tt>extension</tt>: the file extension used for the generated migration files (default <tt>js</tt>)</li>
       <li><tt>tableName</tt>: the table name used for storing the migration state (default <tt>knex_migrations</tt>)</li>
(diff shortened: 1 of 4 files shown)


elhigu:
Thanks for the PR and I'm really sorry that this hasn't got any attention before (I just stumbled here while cleaning up old stuff). 

Code / documentation did look fine to me, would you like to fix conflicts here?

harriha:
Hi @elhigu, 

Whoops, forgot myself that this wasn't merged yet, could've taken an action myself earlier, too. But anyway, I'll take a fresh look at this, I assume most of the changes are valid still (haven't been following changes lately too actively but this is now a good reason to catch up).

harriha:
@elhigu, PR fixed and good to go as far as I can see, thanks for taking this forward.

elhigu:
Thanks!

Issue #848, opened 2015-05-28
Migrations: Can't "CREATE INDEX CONCURRENTLY" anymore


torarvid:
Hi. Not sure if this is a regression, but with knex 0.7.x I could have a migration where I added a raw command to do "CREATE INDEX CONCURRENTLY".

Now I get `CREATE INDEX CONCURRENTLY cannot run inside a transaction block`. Is this not possible at all anymore, or is there a trick to make it work? :)

bendrucker:
For the moment you'll need to follow #834 and disable migration transactions entirely. At some point you'll be able to set this on a per-migration basis.

rhys-vdw:
This has been address by #957.

Issue #834, opened 2015-05-19
no way to disable transactions in knex 0.8.0+ migrations


jurko-gospodnetic:
`knex 0.8.0` started wrapping all `knex` migration steps in separate transactions, and there seems to be no way to turn this behaviour off.

Problem with that is that there are some databases and some commands that just can not be run under an active database transaction. For example, running a `ALTER TYPE ... ADD VALUE` command on Postgres will fail in such scenario (http://www.postgresql.org/docs/9.4/static/sql-altertype.html).

This is also related to a problem of knex reporting such failures by just logging an end-user warning message and not reporting it on as a regular promise rejection, but I'll log that as a separate issue.

I have not yet tested whether and how I can manually stop and restart `knex`'s database transaction around such problematic commands, but since the whole migration steps can not in theory be safely run under a transaction, there really should be a way to turn this behaviour off.

tgriesser:
Ah wow, that's interesting - didn't realize you couldn't do that. Did a little digging and this seems to be how rails fixed it - via disabling ddl transactions on a [per-migration basis](https://github.com/rails/rails/commit/b337390889cb4a9f80ed08daf072a043f0e7ddf3)

I'll look to get something in there to turn this off... would you prefer a flag to have them disabled for all migrations or one which disables on a per-file basis?

jurko-gospodnetic:
Well, for easy backward compatibility, it would definitely be good to be able to disable them all together.

Allowing disabling them for specific files seems a logical next step, but is not really I need at this moment so haven't given much thought to how the interface for that would function.

tgriesser:
Should now be able to add a `disableTransactions` option in the migration config and it will not try to wrap things in a transaction.

jurko-gospodnetic:
Thanks, will give this a try!

Commit's a b*tch to review though, thanks to overwhelming ES5 --> ES6 code changes. :-)

tgriesser:
Sorry about that, only real functional change was [here](https://github.com/tgriesser/knex/blob/01bae6a598bd11cf8fea9e47544ab5d2bcbb8ca2/lib/migrate/index.js#L224-L241)

PR #957, opened 2015-08-30
Allow disabling transaction per migration

harriha:
Hi,

As discussed in #848 and #834, there's cases where one needs to disable transactions on per-migration basis. Ran into a such case myself, and thus ended up taking my chances with this feature.

I couldn't find any hints whether this feature or migration-specific configs in general would be in the works, thus jumped right in to suggest something. If there's some activity around this already, happy to modify this PR accordingly.

The change in the code seemed straightforward, was wondering a bit how to set the actual configs in migration files. The end result below - the reason I didn't go with the current `config.disableTransactions` approach is that I wanted to get rid of that negation there. If you, however, feel that being consistent here would be better, I can change `config.transaction` -> `config.disableTransaction`, for example.

Tests became maybe a bit messy since had to create multiple sets of test migration files, but not much alternatives afaics.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2484,6 +2484,26 @@ $ knex seed:run
       <tt>directory</tt>, <tt>extension</tt>, and <tt>tableName</tt> for the migrations.
     </p>
 
+    <h4 id="Migrations-API-transactions">Transactions in migrations</h4>
+
+    <p>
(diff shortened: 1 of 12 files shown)


rhys-vdw:
Hi @harriha. Very thorough work. Thanks for your contributions, this is a well needed feature.

rhys-vdw:
Alright, I merged this one manually (there were merge conflicts in the compiled output after merging your other PR).

harriha:
Thanks, good to see this getting in :+1: . The issue #848 is related, could potentially be closed.

PR #958, opened 2015-08-30
Allow zero in join clause, fixes #953

### LINKED ISSUE #953: leftJoin can't give a value of 0
I have a statement with

```javascript
knex('xxx').leftJoin('abc as a', 'a.field', 0);
```

And it generate a `QUERY` of : `select * from xxx left join abc as a on a.field =` , the **0** is missing. Perhaps you write a code as followering:

```javascript
if (arguments[1]) {

}
```

which should be:

```javascript
if (typeof arguments[1] === 'undefined') {

}
```

:frowning:

harriha:
Hi,

Took the liberty to look into #953 and found the problem as well. Please let me know if there's any issues with this fix.

### DIFF
--- a/lib/connection/index.js
+++ b/lib/connection/index.js
@@ -1,12 +1,6 @@
 'use strict';
 
-Object.defineProperty(exports, '__esModule', {
-  value: true
-});
-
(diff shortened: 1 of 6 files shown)


rhys-vdw:
Thanks @harriha. Looks like a clean fix.

Issue #332, opened 2014-06-19
Inserting 'undefined' into a DEFAULT field violates NOT NULL constraint in SQLite


coolaj86:
```bash
insert into "schedules" ("created_at", "dtstart", "event", "next", "previous", "rrule", "until", "updated_at", "xattrs") values (?, ?, ?, ?, ?, ?, ?, ?, ?)
[ Thu Jun 19 2014 13:07:34 GMT-0600 (MDT),
  1403204974631,
  '{"type":"log"}',
  1403204974631,
  null,
  'UNTIL=20140619T191234Z;FREQ=DAILY;COUNT=1',
  1403205154631,
  Thu Jun 19 2014 13:07:34 GMT-0600 (MDT),
  undefined ]
```

I believe the `undefined` and corresponding key should be omitted, otherwise `undefined` -> `null` takes precedence over `DEFAULT` and causes an error (whereas in postgresql `DEFAULT TO` seems to be processed whenever a null value is passed).

Also, I don't think the javascript date object is being properly stringified for a SQLite's timestamp field, but I haven't confirmed that.

elhigu:
With sqlite3 undefined keys shouldn't be given if there are null constraint. I wouldnt like to make spec how insert work to behave different way when inserting single row or multiple rows. 

Completely ignoring undefined will work only when inserting single rows. When inserting multiple rows dropping undefined values is not even possible since one cannot give different set of columns for different rows...

Issue #662, opened 2015-02-04
Inserting multiple items ignores defaults for missing values, but sets NULL instead


elhigu:
Documentation says:
```
// Normalizes for empty keys on multi-row insert:
knex('coords').insert([{x: 20}, {y: 30},  {x: 10, y: 20}])
Outputs:
insert into `coords` (`x`, `y`) values (20, NULL), (NULL, 30), (10, 20)
```

Would it be possible to change this behavior to actually use DEFAULT instead of NULL?

Adding implicit NULL causes e.g. not null constraint failure in case where one would like to set default value.

bendrucker:
I think this is sensible. @tgriesser?

demisx:
+1 Just got bitten by it again. Once the seed file has some values for fields, their defaults never kick in.

elhigu:
Giving DEFAULT as special value seems to be supported by mariadb and postgresql... Is there any preference how this should be supported? 

If I understood correctly it did seem that default value was meant to be able to be set in database configuration (https://github.com/tgriesser/knex/blob/master/src/dialects/postgres/index.js#L20).

To me it sounds reasonable to have this as a configuration option, since this feature cannot be implemented e.g. for sqlite and I would like to keep default behaviour of knex consistent between dialect implementations (would be a bit awkward if defaults are set correctly for mysql and postgresql, but differently for sqlite).

I already hacked together simple implementation for this, if the configuration selectable value approach is fine I could start putting clean implementation together with tests and documentation.

elhigu:
Looks like https://github.com/tgriesser/knex/pull/962 has also some discussion about multi insert.

chrisbroome:
:+1: I was very surprised by this behavior also.

Unless there is a **very** good reason not to, my columns are always `NOT NULL` with `DEFAULT` values. The current behavior assumes that every column is nullable which is a really really bad practice for relational schema design.

MatthewPhinney:
As far as I can tell this was not completely fixed until #1269, included in version 11.

nik-lampe:
For me this still happens when I run the seeds programatically.

Running 
 await knex.seed.run({ specific: 'dev.js', env: 'test' }) 
sets `null` for the missing values
whereas running 
knex seed:run --specific dev.js --env test
sets `DEFAULT`

PR #962, opened 2015-09-02
Don't insert or update columns with undefined values

### LINKED ISSUE #961: undefined attributes are not handled properly
When saving to the database values that are undefined should not be saved currently they art set to NULL.  See tgriesser/bookshelf#891

b3njamin:
Fixes #961. This is unable to work on bulk inserts. 

Not set up to do integration tests, was able to run through unit tests fine.

### DIFF
--- a/lib/query/compiler.js
+++ b/lib/query/compiler.js
@@ -78,7 +78,6 @@ assign(QueryCompiler.prototype, {
     } else if (typeof insertValues === 'object' && _.isEmpty(insertValues)) {
       return sql + this._emptyInsertValue;
     }
-
     var insertData = this._prepInsert(insertValues);
     if (typeof insertData === 'string') {
       sql += insertData;
@@ -360,6 +359,7 @@ assign(QueryCompiler.prototype, {
     var i = -1;
     while (++i < data.length) {
       if (data[i] == null) break;
+      data[i] = _.omit(data[i], _.isUndefined);
       if (i === 0) columns = Object.keys(data[i]).sort();
       var row = new Array(columns.length);
       var keys = Object.keys(data[i]);
@@ -388,6 +388,7 @@ assign(QueryCompiler.prototype, {
 
   // "Preps" the update.
   _prepUpdate: function _prepUpdate(data) {
+    data = _.omit(data, _.isUndefined);
     var vals = [];
     var sorted = Object.keys(data).sort();
     var i = -1;
--- a/lib/query/compiler.js
+++ b/lib/query/compiler.js
@@ -359,7 +359,6 @@ assign(QueryCompiler.prototype, {
     var i = -1;
     while (++i < data.length) {
       if (data[i] == null) break;
-      data[i] = _.omit(data[i], _.isUndefined);
       if (i === 0) columns = Object.keys(data[i]).sort();
       var row = new Array(columns.length);
       var keys = Object.keys(data[i]);
--- a/src/query/compiler.js
+++ b/src/query/compiler.js
@@ -57,7 +57,7 @@ assign(QueryCompiler.prototype, {
     }
     return _.compact(statements).join(' ');
   },
-  
+
   pluck: function() {
     return {
       sql: this.select(),
@@ -78,13 +78,12 @@ assign(QueryCompiler.prototype, {
     } else if (typeof insertValues === 'object' && _.isEmpty(insertValues)) {
       return sql + this._emptyInsertValue
     }
-
     var insertData = this._prepInsert(insertValues);
     if (typeof insertData === 'string') {
       sql += insertData;
     } else  {
       if (insertData.columns.length) {
-        sql += '(' + this.formatter.columnize(insertData.columns) 
+        sql += '(' + this.formatter.columnize(insertData.columns)
         sql += ') values ('
         var i = -1
         while (++i < insertData.values.length) {
@@ -124,14 +123,14 @@ assign(QueryCompiler.prototype, {
         if (stmt.distinct) distinct = true
         if (stmt.type === 'aggregate') {
           sql.push(this.aggregate(stmt))
-        } 
+        }
         else if (stmt.value && stmt.value.length > 0) {
           sql.push(this.formatter.columnize(stmt.value))
         }
       }
     }
     if (sql.length === 0) sql = ['*'];
-    return 'select ' + (distinct ? 'distinct ' : '') + 
+    return 'select ' + (distinct ? 'distinct ' : '') +
       sql.join(', ') + (this.tableName ? ' from ' + this.tableName : '');
   },
 
@@ -357,7 +356,7 @@ assign(QueryCompiler.prototype, {
     if (statement.not) return 'not ' + str;
     return str;
   },
-  
+
   _prepInsert: function(data) {
     var isRaw = this.formatter.rawOrFn(data);
     if (isRaw) return isRaw;
@@ -367,6 +366,7 @@ assign(QueryCompiler.prototype, {
     var i = -1
     while (++i < data.length) {
       if (data[i] == null) break;
+      data[i] = _.omit(data[i], _.isUndefined)
       if (i === 0) columns = Object.keys(data[i]).sort()
       var row  = new Array(columns.length)
       var keys = Object.keys(data[i])
@@ -395,6 +395,7 @@ assign(QueryCompiler.prototype, {
 
   // "Preps" the update.
   _prepUpdate: function(data) {
+    data = _.omit(data, _.isUndefined)
     var vals   = []
     var sorted = Object.keys(data).sort()
     var i      = -1
--- a/src/query/compiler.js
+++ b/src/query/compiler.js
@@ -366,7 +366,6 @@ assign(QueryCompiler.prototype, {
     var i = -1
     while (++i < data.length) {
       if (data[i] == null) break;
-      data[i] = _.omit(data[i], _.isUndefined)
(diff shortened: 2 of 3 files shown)


vschoettke:
@rhys-vdw This is a breaking change and should cause a 0.9.0 version.

rhys-vdw:
@vschoettke :+1:

Do you agree with the concept?

rhys-vdw:
@b3njamin Could you explain the bulk insert limitation? I'd prefer consistency.

rhys-vdw:
Also adding a test case for #332 would be nice.

b3njamin:
@rhys-vdw From my understanding in both mysql and postgres bulk inserts require a consistent number of columns to values. Currently for updates & single inserts it removes the column from being saved. For multiple inserts it will set the column to NULL (which again is incorrect the only option i guess would be to set it to the default value)

Insert (single)
```js
knex('users').insert({name: 'Jane Doe', dob: undefined})
```
```sql
INSERT INTO users (`name`) values ('Jane Doe')
```
Update
```js
knex('users').where('id', '=',1)
  .update({
    name: 'Jannet Doe'
    dob: null
  })
```
```sql
UPDATE `users` SET `name` = 'Jannet Doe' WHERE `id` = 1
```

Insert (multiple) at present
```js
knex('users').insert([{name: 'John Doe', dob: '1991-10-07'}, {name: 'Jane Doe', dob: undefined}])
```
```sql
insert into `users` (`name`, `dob`) values ('John Doe', '1991-10-07'), (Jane Doe, NULL)
```

Insert (multiple) potentially
```js
knex('users').insert([{name: 'John Doe', dob: '1991-10-07'}, {name: 'Jane Doe', dob: undefined}])
```
```sql
insert into `users` (`name`, `dob`) values ('John Doe', '1991-10-07'), (Jane Doe, DEFAULT)
```

rhys-vdw:
@b3njamin - I think it would be wrong to merge this without addressing the bulk insert case.

I agree with using `DEFAULT` where `undefined` is specified. (Actually it would be fine to have `DEFAULT` for `undefined` keys in the individual insert case as well, if that makes it easier.)

vschoettke:
@rhys-vdw I try not to use the `null` type in JS as it's just additional overhead. So if someone wants to insert a parameter (existing key) with value `undefined` it should use that in the query.

b3njamin:
@rhys-vdw currently bulk inserts handle columns to value issue by setting the columns to `undefined` (or `NULL`), do we address this so inserts set `undefined` values to `DEFAULT`. But only on inserts, updates just remove undefined value.

rhys-vdw:
@vschoettke, `null` is an interesting/controversial concept. I'll provide some arguments for why I think this is the desired behaviour and hopefully we can come to an agreement.

**Null is explicit**

Generally I prefer to take the option that supports developer intention. Explicitly setting a value to `null` which is analogous with the SQL concept of `NULL` satisfies this. `undefined` is less clear, therefore less explicit.

**`undefined` means 'does not exist'**

A variable being not defined is distinct from it being `null`. JavaScript treats `undefined` values similarly to deleted values in many ways.

```js
var x = {a: undefined};
x.a === x.b // -> true
```

```js
var x = {a: undefined};
JSON.stringify(x) // -> '{}'
```

**Knex treats missing keys as default for single insert, but as NULL for bulk**

One aspect of the current behaviour is that it treats missing keys as undefined - therefore assigning them a `NULL` value.

```js
var a = {a: 'a'};
var b = {b: 'b'};

knex('table').insert(a).toString()
// ->  "insert into "table" ("a") values ('a')"

knex('table').insert([a, b]).toString(),
// ->  "insert into "table" ("a", "b") values ('a', NULL), (NULL, 'b')"
```

The absence of a field in an individual insert means that it will be set to `DEFAULT` value. But when we insert this same object in an array it will be `NULL`. This is inconsistency and could lead to hard to find errors.

To make this behaviour consistent we must treat both absent fields and `undefined` fields as if they were equivalent (as `JSON.stringify` does)

**There is no way to specify `DEFAULT` for one cell value within bulk insert statement**

An additional note on the above behaviour - there is no way to work around it. You can't do a bulk insert with `DEFAULT` values.

rhys-vdw:
> do we address this so inserts set undefined values to DEFAULT. But only on inserts, updates just remove undefined value.

@b3njamin, I believe setting missing columns and `undefined` columns to `DEFAULT` is the correct behaviour here (as outlined in previous post).

In the case of `.update()`, not providing a value (or providing `undefined`) makes sense to not alter the value in question, and should be ignored.

As @vschoettke pointed out, these are breaking changes so some consideration should be made.

@tgriesser, do you have an opinion on this?

b3njamin:
@rhys-vdw with you on this! I think its pretty clear to that undefined `!= null`, and as such should be handled differently. This also fix #332 in regards to the `NOT NULL` constraints.

b3njamin:
@rhys-vdw @tgriesser So I've updated the code so just affects updates, not 100% sure how to change the insert code to use default values, might need someone else to do this, or someone to point me in the right direction.

tkellen:
The keyword DEFAULT can be used in lieu of a real value.

b3njamin:
@tkellen Sure, but where is the best place to put this. `_prepInsert`?

tkellen:
Whoops, sorry for the drive-by (totally unhelpful, obvious) comment.

rhys-vdw:
@b3njamin, sorry for the delay getting back to you.

I spoke to @tgriesser last night and got the green light on this issue. I'd love to see this change merged into master.

Are you still interested in helping out with this one? I'm guessing you got stuck, so could you give an update on what's blocking? Hopefully I, or somebody else, will be able to provide some guidance.

chrisbroome:
@rhys-vdw @b3njamin I have to say I disagree with the update behavior. I think there's a big difference between the following:

`{name: 'joe', dob: undefined}`, should produce the following sql:

```SQL
UPDATE user SET name = 'joe', dob = DEFAULT;
````

`{name: 'joe'}`, should produce the following sql:

```SQL
UPDATE user SET name = 'joe';
```

Basically, if a key is present in the JavaScript object's *own enumerable properties*, then it should be sent to the database. Here is a table with what I think the behavior should be.

* If the key is present but `undefined`, it should be set to `DEFAULT`.
* If the key is present and `null`, it should be set to `NULL`.
* If the key is present and any other value, send that value to the database.
* If the key is not present, do not send anything for the value (obviously).

If only specific fields should be updated, I think the *application code* should whitelist the fields before the operation. I think that whitelisting is out of scope for *library code*. Whitelisting attributes for updates and inserts is a good practice anyway to avoid mass assignment attacks. Here's an example of what I'm talking about:

```JavaScript
var _ = require('lodash');

function updateUser(attributes) {
  var whitelistedFields = ['name'];
  var whitelistedAttribute = _.pick(attributes, whitelistedFields);
  return knex.from('user').update(attributes);
}

var sql = updateUser({name: 'joe', dob: undefined}).toString();
```

The `sql` var should contain the string from the second example above.

rhys-vdw:
@chrisbroome: This is not intended to make `undefined` a shorthand for `default`. In the case of an update the key should not be inserted - as it has no value. The choice of using `default` is because it's an alternative to `NULL`, which assumes that columns are nullable. When bulk inserting, it is possible for some row values to have missing keys. In this case we need a sensible default - currently the default is `NULL`.

In your example `default` should be specified explicitly:

```js
var sql = updateUser({name: 'joe', dob: knex.raw('default')}).toString();
```

elhigu:
I had hard time to understand exactly what this pull request was about from the discussion above.... so just to be clear this pull request is about supporting the following testcase (correct me if I'm wrong):

```js
  it("should not update columns undefined values", function() {
    testsql(qb().update({'email': 'foo', 'name': undefined}).table('users').where('id', '=', 1), {
      mysql: {
        sql: 'update `users` set `email` = ? where `id` = ?',
        bindings: ['foo', 1]
      },
      default: {
        sql: 'update "users" set "email" = ? where "id" = ?',
        bindings: ['foo', 1]
      }
    });
  });
```

Part discussed about inserting multiple values is already implemented.

b3njamin:
@elhigu Yup

elhigu:
@b3njamin I think implementation looks good / correct. There was many unnecessary whitespace changes making it a bit hard to read though. 

Could you fix the conflicts (I suppose they come from generated code) and squash the commits so that invalid / reverted code from the first commit won't get to the codebase?

elhigu:
@b3njamin do you have any plans of completing this one? Just checking, no pressure :)

b3njamin:
@elhigu I can merge, but the implementation isn't complete in regards to bulk inserts, as mentioned, quite far above.

elhigu:
@b3njamin I have implemented support for undefined keys in bulk inserts already couple months ago https://github.com/tgriesser/knex/issues/662 . No need to address that issue in this pull request.

b3njamin:
@elhigu cool - ill rebase master into it - any need to add any user facing documentation for this change?

elhigu:
@b3njamin yes, please I had forgot the documentation of the new feature completely. One line in insert / update docs in index.html should do. Maybe just adding e.g. `ignoredKey: undefined` to one of the examples would be easiest to understand?

elhigu:
@b3njamin we also removed all generated lib/ code from repo so resolving conflicts should be easier now :)

elhigu:
I fixed this without ws changes + documentation in #1174. Thanks for making this happen @b3njamin :+1:

PR #964, opened 2015-09-03
Clarify documentation for defaults for precision and scale

gangstead:
Had to go digging through the source code to find out how my numeric columns were getting their defaults.  I added a line to the float and decimal columns types' documentation to make it clear.  This is in line with other defaults that are specified in the docs.


rhys-vdw:
Thanks. :+1:

Issue #897, opened 2015-07-14
Allow bin/knex to use DB_URL to connect


b3njamin:

b3njamin:
#976

elhigu:
Could you just write reading value from DB_URL / DATABASE_URL to knexfile.js instead of adding hard coded environment variables to knex client? 

There is an example how to use environment variables from knex file in http://knexjs.org/#knexfile

```js
module.exports = {
  client: 'pg',
  connection: process.env.DATABASE_URL || {
    user: 'me',
    database: 'my_app'
  }
};
```

I'm against of having this included, because I don't see that this feature adds any value to the project. @rhys-vdw @tgriesser could you close this issue and related pull request if you agree?

b3njamin:
Id argue that it adds the ability to run knex executable without the need for a file. A lot of popular tools allow the overwriting of file based configuration by environment variables including rails. See here http://guides.rubyonrails.org/configuring.html#configuring-a-database

elhigu:
I understand very well that many tools offer setting some parameters through environment variables and when I'm deploying software in cloud servers etc. environment variables is my primary way to setup configuration for pretty much everything (especially secret stuff that cannot be put to code repository).

What I'm not getting is the use case why would you like to use `knex` client tool with only `DB_URL` set? I mean that tool is made for running seed files / migrations and you need to have `knexfile.js` to tell where to find your migrations / seed files that you are trying to run.

Using `knex` client just with `DB_URL` set without any additional configuration doesn't make sense to me. 

Also if you have seeds / migration files to run with `knex` in your repository, why not to include `knexfile.js` as well and map there reading values from all the environment variables you like? Having that extra file doesn't hurt anyone and it gives you a lot more powerful way to use environment variables to configure your software.

Only use case what I can think of for this feature is when there are some ready made migrations and `knexfile.js` which you don't like to modify and you would like to only override database connection part of the configuration to run migrations to other DB... Could you explain your use case why do you need this feature?

Other thing I left out was that if one is going to make using `DB_URL` env variable supported directly in client, we should consider well, which variable name to use and if that or value from `knex` configuration should be primary choice in case where both are found (probably knexfile.js would be the primary one).

rhys-vdw:
I agree with @elhigu. This use case is already supported by knex by referencing `process.env` in your knexfile.

b3njamin:
Supporting it in the client is definitely better option. Its just cleaner to foego the knexfile and it also follows conventions set by other libraries.

PR #976, opened 2015-09-15
Knex CLI can read the config from either a file or env vars

b3njamin:


elhigu:
Left comments to related issue #897

rhys-vdw:
Thanks for the PR @b3njamin, but we wont be merging this one (as discussed in #897).

PR #981, opened 2015-09-17
Fix 'migrate.latest()' when at base migration

techniq:
Getting `Knex:warning - migrations failed with error: Cannot read property '0' of undefined` when running `knex.migrate.latest(config)` when at the base migration (no records in migrations table)


techniq:
Disregard.  This was a residual issue from `mock-knex` where it was removing the mocked client.

Issue #855, opened 2015-06-01
Query Event 8.6 not triggering like before


wubzz:
Hey.

Updated my knex package from 7.6 -> 8.6 (big leap, I know). By doing so, it appears the Query Event is not being triggered as it used to. Is this something you're aware of?

Example code:
```javascript
knexDb.on('query', function() {
var debugVariable = ''; //Never get here, not triggering
});
```

Everything is run through transactions (if that matters?). We're very dependant on this event as we append SchemaName (Postgres) to the queries.

I couldn't find anything relating to this in the changelog, but please do correct me if I missed it.

Thanks in advance.

wubzz:
It seems to have broken between 0.7.6 -> 0.8.0

tgriesser:
Ah, yep transactions got a big internal overhaul and I think I forgot to proxy the events on the "transaction" client back through to the original client.

Also, are you saying that you're using the event to mutate the object passed to the event, modifying the sql before it goes into the query? If so I might need to change else to support this behavior.

wubzz:
@tgriesser Ah, I see, thanks for the quick reply. I'll stay 7.6 for now and wait for fix. :)

And yes, that's what's being done. The sql property of the data object gets manipulated by adding the SchemaName to all relevant places of the query through a regex. This is done because I saw no other (or better) way of doing it.

Example:
```sql
db('Year').select().toString()
select * from "Year"
```
<--- Event --->
```sql
select * from "MySchema"."Year"
```

tgriesser:
In that case you should be able to do `db('MySchema.Year').select().toString()`, no?

wubzz:
Surprisingly, that did work.. I recall having tried that before and received **"MySchema.Year"** instead of **"MySchema"."Year"** which is the original cause opting for the query event.

I'll keep that in mind, might make life easier.

PR #984, opened 2015-09-19
Query event in Transaction should be proxied to the main client

wubzz:
As discussed with @tgriesser in #855 the query event is no longer triggered onto the main client when in a transaction. It was simply a miss during the transaction overhaul, according to tgriessers's comments.

Example code:
```
dbClient.on('query', function(obj) {
	console.log('onQuery'); //Works in knex < 8.0.0, fails in >= 8.0.0
	return obj;
});

dbClient.transaction(function(tr) {
	return tr.raw('SELECT version()').then(tr.commit).catch(tr.rollback);
}).then(_.noop).catch(_.noop);
```

There may be better ways of fixing this, I don't know.

### DIFF
--- a/lib/dialects/websql/transaction.js
+++ b/lib/dialects/websql/transaction.js
@@ -24,6 +24,7 @@ function makeClient(trx, client) {
 
   trxClient.on('query', function (arg) {
     trx.emit('query', arg);
+    client.emit('query', arg);
   });
   trxClient.commit = function () {};
   trxClient.rollback = function () {};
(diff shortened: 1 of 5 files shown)


rhys-vdw:
Hi @wubzz. That looks reasonable to me, and looks like it closes #855.

I've made a rule not to merge anything without tests as I am also not that familiarized with the codebase. So please add a couple of tests for these events and I'll merge it in.

wubzz:
@rhys-vdw Truthfully having issues installing dependecies for knex on my windows platform, and can't get it to work. Seems to primarily be mariasql where I have issues..

I commited a test hoping that I added it in the right place (also lack experience with the project structure/codebase), and hoping CI will pull through.

Pretty straight-forward test, tho.

PR #986, opened 2015-09-21
Document connectionless use case

juanjoLenero:


rhys-vdw:
Thanks @juanjoLenero. :+1:

explicitcall:
any reason why this documentation was removed in the later releases?

wubzz:
Not sure why *everything* was removed. The only part that was actually wrong in this doc in newer knex was:
```
Just pass in an empty object when initializing the library. Specify a client if you are interested in a particular flavour of SQL.
```

Since now knex requires a client to be specified. But back when this PR was created, it was all valid.

explicitcall:
Hi @elhigu, is to worth resubmitting this PR? 

Converting knex queries to strings to later reuse with own client object (or even other libraries) works for me with the latest version, I don't see why adding information about this back to the docs page would hurt anything.

elhigu:
Yep, looks like there is no explicit mention about using knex without connection (even that there are some examples of it). There should be some section added back, which tells that knex can be used as a query builder if only `client` parameter is passed in configuration. And that section should mention that one can use toSQL().toNative() method to get dialect specific SQL and bindings out in as separate strings and to mention that toString() shouldn't be used for that conversion.

@explicitcall yes PR of some documentation about the use case would be welcome. Nowadays documentation is found from separate repository.

PR #987, opened 2015-09-21
Add "udt_name" attribute to the columnInfo() method

sanches89:
This is useful when the field is set to array.

### DIFF
--- a/lib/dialects/postgres/query/compiler.js
+++ b/lib/dialects/postgres/query/compiler.js
@@ -90,6 +90,7 @@ assign(QueryCompiler_PG.prototype, {
         var out = _.reduce(resp.rows, function (columns, val) {
           columns[val.column_name] = {
             type: val.data_type,
+            udtType: val.udt_name,
             maxLength: val.character_maximum_length,
             nullable: val.is_nullable === 'YES',
             defaultValue: val.column_default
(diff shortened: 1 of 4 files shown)


rhys-vdw:
Looks reasonable. Any idea why this (and [all the others](http://www.postgresql.org/docs/9.4/static/infoschema-columns.html)) were omitted?

sanchesic89:
Maybe for consistency between databases, but in my tests the other informations, apparently, is not so relevant. 
I don't know if is interesting add a parameter in the method to show all column informations.

rhys-vdw:
So long as a minimum set of values is shared between all implementations then consistency shouldn't be an issue. However, if my understanding of the docs is correct, might it be more appropriate to do this?

```js
columns[val.column_name] = {
  type: val.data_type === 'USER-DEFINED' ? val.udt_name : val.data_type;
  maxLength: val.character_maximum_length,
  nullable: (val.is_nullable === 'YES'),
  defaultValue: val.column_default
}
```

sanchesic89:
I agree with you and I forgot about [User Types](http://www.postgresql.org/docs/9.4/static/xtypes.html). However you suggestion won't solve the array problem.
I make same changes in my repository and  now I believe that is right.

elhigu:
Closing due to inactivity.

PR #994, opened 2015-09-23
Query event in Transaction should be proxied to the main client

wubzz:
As discussed with @tgriesser in #855 the query event is no longer triggered onto the main client when in a transaction. It was simply a miss during the transaction overhaul, according to tgriessers's comments.

Example code:
```
dbClient.on('query', function(obj) {
	console.log('onQuery'); //Works in knex < 8.0.0, fails in >= 8.0.0
	return obj;
});

dbClient.transaction(function(tr) {
	return tr.raw('SELECT version()').then(tr.commit).catch(tr.rollback);
}).then(_.noop).catch(_.noop);
```

There may be better ways of fixing this, I don't know.

### DIFF
--- a/lib/dialects/websql/transaction.js
+++ b/lib/dialects/websql/transaction.js
@@ -24,6 +24,7 @@ function makeClient(trx, client) {
 
   trxClient.on('query', function (arg) {
     trx.emit('query', arg);
+    client.emit('query', arg);
   });
   trxClient.commit = function () {};
   trxClient.rollback = function () {};
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -208,6 +208,7 @@ function makeTxClient(trx, client, connection) {
 
   trxClient.on('query', function (arg) {
     trx.emit('query', arg);
+    client.emit('query', arg);
   });
 
   var _query = trxClient.query;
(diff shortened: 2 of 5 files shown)


wubzz:
@rhys-vdw Had to open a new one, tests weren't being trigger oncommit on the other PR ( #984 ) ...

Seems to pull through.

wubzz:
@rhys-vdw Did I miss something in this PR before it can be merged? Please do let me know if that's the case. I was kinda expecting to see it in 0.9 couple of weeks back.

rhys-vdw:
Hey @wubzz. Sorry for the lack of correspondence. This looks good!

One question before I merge - why does websql transaction need its own copy of this code? Is it not possible to add this code in some shared code?

I'll do another release for you after this gets merged.

wubzz:
@rhys-vdw I can't give an honest to god answer to this, but based around the fact that the two classes of which these changes are contained (Transaction_WebSQL and no WebSQL Transaction) are very different, I decided to keep it seperate. They both contain their own `trxClient` event registrant before my changes, so I just added upon that.

rhys-vdw:
Yep, no problem @wubzz. Sounds good. Thanks for the PR.

PR #1009, opened 2015-10-05
Fix reference to undefined `client.initialize` method

### LINKED ISSUE #866: Error reinitializing connection pool
It was mentioned [here](https://github.com/tgriesser/knex/issues/724#issuecomment-97991976) that since `0.8.x` it should be possible to call `knex.initialize()` after `knex.destroy()` to re-initialize a connection pool.

However, I am getting the error below:

```
Object [object KnexClient] has no method 'initialize'
at Function.initialize (./node_modules/knex/lib/util/make-knex.js:44:21)
```

Using `knex@0.8.6` and `sqlite3@3.0.8`.

I have created a [Gist](https://gist.github.com/blah238/3292ef7881b3dd3c7269) demonstrating the issue.

It seems that [here](https://github.com/tgriesser/knex/blob/5e65678b32fa639843902d39603ab2050d84f10f/src/util/make-knex.js#L43) the `make-knex` utility attempts to call a non-existent `initialize` function on the client object. Is this functionality just not implemented yet, or was I misunderstanding the intention?

I could work around this by creating a new Knex instance (and remembering to replace every reference to the previous instance, which is not always straightforward given things like module caching and 3rd party code), but it would simplify things greatly by being able to reuse the same instance and just reinitializing its connection pool.

blah238:
Fixes #866

### DIFF
--- a/lib/util/make-knex.js
+++ b/lib/util/make-knex.js
@@ -41,7 +41,7 @@ module.exports = function makeKnex(client) {
 
     // Typically never needed, initializes the pool for a knex client.
     initialize: function initialize(config) {
-      return client.initialize(config);
+      return client.initializePool(config);
     },
 
     // Convenience method for tearing down the pool.
--- a/src/util/make-knex.js
+++ b/src/util/make-knex.js
@@ -40,7 +40,7 @@ module.exports = function makeKnex(client) {
 
     // Typically never needed, initializes the pool for a knex client.
     initialize: function(config) {
-      return client.initialize(config)
+      return client.initializePool(config)
     },
 
     // Convenience method for tearing down the pool.
--- a/test/integration/suite.js
+++ b/test/integration/suite.js
@@ -4,6 +4,7 @@
 
 module.exports = function(knex) {
   var sinon = require('sinon');
+  var Pool2 = require('pool2');
 
   describe(knex.client.dialect + ' | ' + knex.client.driverName, function() {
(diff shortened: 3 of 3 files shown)


rhys-vdw:
Hi @blah238. Great. Could you please provide a test for this to prevent regressions?

blah238:
@rhys-vdw added a test that passes for me under SQLite3. Not sure why it's failing for mariasql.

blah238:
@rhys-vdw does the Travis build not run the Babel build? The error would indicate that it's not rebuilding `/lib`, and I gather that PRs are not to change the contents of `/lib` either.

rhys-vdw:
I've never investigated the CI system. If that's the case we have a problem.

blah238:
Yeah sorry I deleted that comment, but it would appear that Travis does not have an `npm run dev` step, and you have to update `/lib` by running `npm run dev` locally and including those changes in your commit for it to be reflected in Travis.

Seems like a lot of PRs might have invalid tests because of this.

blah238:
This is noted in [CONTRIBUTING.md](https://github.com/tgriesser/knex/blob/4304e38115bc27a6732cbdefd9f92c5bcee2094f/CONTRIBUTING.md#how-to-contribute-to-knexjs) BTW, which is why I deleted my comment.

blah238:
Closing in favor of #1014.

PR #1011, opened 2015-10-06
Ensure /lib is up to date during CI tests

blah238:
As noted in this [comment](https://github.com/tgriesser/knex/pull/1009#issuecomment-146021024) it looks like Travis doesn't have an `npm run dev` step, so PRs which do not remember to update `/lib` will not be tested against the built source.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ node_js:
 before_script:
   - psql -c 'create database knex_test;' -U postgres
   - mysql -e 'create database knex_test;'
+  - babel -L -D src/ --out-dir lib/
 
 after_script:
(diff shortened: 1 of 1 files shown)


rhys-vdw:
I wonder if it might be more appropriate to add the build command into the test script in package.json. That way it'll work locally also. What do you think?

blah238:
@rhys-vdw sounds good. I've also been using an `npm run babel` command locally since I've found that `npm run dev` would often hang (unkillable by Ctrl-C'ing). I've also had trouble running `npm test` on Windows but I think I've fixed that now. Closing this in favor of PR #1019.

Issue #921, opened 2015-07-28
Out of order resource error handling


myndzi:
Split off from https://github.com/tgriesser/knex/issues/842

Thanks to a lot of back and forth with @zacronos, we identified a problem in the postgres dialect (and probably others) with error and resource handling.

Knex uses `Promise.using()` with a disposer to wrap requests:
https://github.com/tgriesser/knex/blob/0.8.6/lib/runner.js#L135-L142

...but the underlying resource allocation binds `end`/`error` events, with a function that destroys the resource from the pool:

https://github.com/tgriesser/knex/blob/0.8.6/lib/dialects/postgres/index.js#L67-L68

https://github.com/tgriesser/knex/blob/0.8.6/lib/dialects/postgres/index.js#L160-L166

The result is: if, during a query, the connection emits 'end' or 'error', the resource is destroyed from the pool; the query runner ends, and the `ensureConnection`'s disposer attempts to release the (now destroyed) resource back to the pool.

Additionally, if this was an error event, the error message might be getting swallowed?

This isn't a straightforward patch and may require some refactoring; I am not quite sure how to approach it. The destroy and release should be on the same level of abstraction, but they aren't.

myndzi:
Simplified test case:

    'use strict';

    var knex = require('knex')({
      client: 'pg',
      connection: {
        // connection info
      },
      pool: {
        min: 1,
        max: 1
      }
    });

    knex.transaction(function (trx) {
      return trx.select('id').from('sites').limit(1)
      .then(function (res) {
        var res = knex.client.pool.pool.keys()[0];
        res.emit('error', new Error('foo'));
      });
    });

I don't believe the transaction is necessary, but it was the simplest way I was able to ensure that I emitted the error while the resource was 'checked out' by knex.

devTristan:
This is relevant to me in that it is currently ruining my life.

devTristan:
Ran your script, got this, if it helps at all:

```
node horror.js 
Knex:Error Pool2 - Error: Pool.release(): Resource not member of pool
```

seamusabshere:
Could this also be a source of the dreaded `[Error: Connection terminated]` errors?

myndzi:
I didn't have time to dig into this when I submitted this issue, but there's a PR that I believe should help, now:

https://github.com/tgriesser/knex/pull/973

[...]

PR #1014, opened 2015-10-08
Change pool initialization to be asynchronous

blah238:
### Summary:

Pool initialization is no longer done implicitly when instantiating Knex or a Knex client. Pool initialization ~~must~~ may be performed explicitly, by calling either `knex#initialize` or `client#initializePool`. These are both now asynchronous methods that can either take a callback or return a promise. *Edit:* If the pool is not initialized explicitly, it will be initialized upon the first database connection being acquired.

### Rationale:

Pool initialization should always be performed asynchronously, because pool destruction (which may occur during pool initialization if a pool was previously created) occurs asynchronously. Additionally, constructors should not cause side effects and should always run synchronously.

**NOTE:** ~~This is a breaking API change in that users who were relying on implicit pool initialization will have to change their code to explicitly initialize the pool.~~ This should be backwards-compatible, as pool initialization will be done implicitly upon the first connection being acquired.

Supersedes PR #1009 (Fix reference to undefined `client.initialize` method)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -157,7 +157,6 @@ assign(Client.prototype, {
         }
       }).then(function () {
         client.pool = new client.Pool(assign(client.poolDefaults(config.pool || {}), config.pool));
-        client.pool.setMaxListeners(0); // Suppress warnings about too many event listeners
         client.pool.on('error', function (err) {
           helpers.error('Pool2 - ' + err);
         });
@@ -235,20 +234,10 @@ assign(Client.prototype, {
   // A callback may also be used instead of a promise.
   releaseConnection: function releaseConnection(connection, callback) {
     var pool = this.pool;
-    var promise = new Promise(function (resolver, rejecter) {
-      var errorHandler, successHandler;
+    var promise = new Promise(function (resolver) {
       debug('releasing connection to pool: %s', connection.__knexUid);
-      errorHandler = function (err) {
-        pool.removeListener('drain', successHandler);
-        rejecter(err);
-      };
-      successHandler = function () {
-        pool.removeListener('error', errorHandler);
-        resolver();
-      };
-      pool.once('error', errorHandler);
-      pool.once('drain', successHandler);
       pool.release(connection);
+      resolver();
     });
     // Allow either a callback or promise interface for releasing connections.
     if (typeof callback === 'function') {
--- a/lib/client.js
+++ b/lib/client.js
@@ -38,9 +38,6 @@ function Client() {
   this.connectionSettings = cloneDeep(config.connection || {});
   if (this.driverName && config.connection) {
     this.initializeDriver();
-    if (!config.pool || config.pool && config.pool.max !== 0) {
-      this.initializePool(config);
-    }
   }
 }
 inherits(Client, EventEmitter);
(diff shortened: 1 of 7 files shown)


blah238:
Not sure why the 0.10.x build is failing but 0.12.x is passing: https://travis-ci.org/tgriesser/knex/builds/84410455

rhys-vdw:
> NOTE: This is a breaking API change in that users who were relying on implicit pool initialization will have to change their code to explicitly initialize the pool.

I don't disagree with your rationale, but I'm not going to be able to merge a breaking change into master. This project is too popular to break something that fundamental. I'd suggest maybe offering a constructor option `initialize` which would be `true` by default (ie. has the current knex behaviour), but when activated requires explicit initialization. I'd want to run this by @tgriesser as well (including my suggestion).

```js
var knex = Knex({ client: {...}, initialize: false });

knex.initialize()
.then(startApp)
.catch(...);
```

One of the advantages to the current pattern is that it makes use code much simpler (as it doesn't require waiting on the async `initialize` before using `knex`). Async patterns can be a bit confusing for less experienced JS coders, so I think this does have its merits.

I'm more inclined to merge your fix #1009, which is pretty clear cut - are you still happy for that to be merged in the meantime?

blah238:
@rhys-vdw thanks for your feedback! As an alternative to adding an `initialize` option, I've just pushed a commit that adds back implicit pool initialization to `client#acquireConnection`, which should solve this in a backwards-compatible way.

There are a few other improvements as well. I'll squash it after you've had a chance to review it. Looks like all tests are passing.

rhys-vdw:
Thanks @blah238. This is sitting in my to do list.

I've noticed that you've done some good work on Knex recently. I was trying to contact you privately to see if you'd be interested in collaborator status, but I can't find your contact email. If you're interested in discussing that please email me: [rhys.vdw@gmail.com](mailto:rhys.vdw@gmail.com).

lib/client.js
```diff
@@ -151,15 +148,30 @@ assign(Client.prototype, {
...
+  initializePool: function initializePool(config, callback) {
+    var client = this;
+    var promise = Promise['try'](function () {
+      return Promise['try'](function () {
+        if (client.pool) {
+          return client.destroy();
+        }
+      }).then(function () {
+        client.pool = new client.Pool(assign(client.poolDefaults(config.pool || {}), config.pool));
+        client.pool.setMaxListeners(0); // Suppress warnings about too many event listeners
```
rhys-vdw: Why suppress warnings here?

src/client.js
```diff
@@ -148,15 +145,31 @@ assign(Client.prototype, {
...
+    var client = this
+    var promise = Promise.try(function() {
+      return Promise.try(function() {
+        if (client.pool) {
+          return client.destroy()
+        }
+      })
+      .then(function() {
+        client.pool = new client.Pool(assign(client.poolDefaults(config.pool || {}), config.pool))
+        client.pool.setMaxListeners(0) // Suppress warnings about too many event listeners
```
rhys-vdw: Why suppress warnings here?

blah238: @rhys-vdw because I added per-connection listeners in [`client#releaseConnection`](https://github.com/tgriesser/knex/pull/1014/files#diff-50cfa59973c04321b5da0c6da0fdf4feR249) and it would spam warnings about having more than 10 event listeners whenever you have several connections open. Maybe @myndzi (from [Pool2](https://github.com/myndzi/pool2)) has some input here on whether there is another way to wait for connections to be released.

rhys-vdw: Is it possible to set it to a high number that will still be triggered when something goes wrong, without it being triggered in the course of normal proceedings? I assume this event is only used internally, so we might be able to calculate the max number of listeners from max connections?

blah238: I'm not sure. I can try it -- however I am not that familiar with Pool2's inner workings. I just went into the source for [pool#release](https://github.com/myndzi/pool2/blob/02e1ecd496857ba5799daa0c5d84e1b3ce694418/lib/pool.js#L184) and saw what events it emitted and went from there.

One of the issues I was trying to fix was that previously, the promise would never reject, but maybe that's because of how Pool2 is designed.

However, seeing issues like https://github.com/tgriesser/knex/issues/921 now makes me hesitant to change too much at once, so maybe it would be more prudent to revert the changes to this function, perhaps just keeping the ability to use a callback.

Might be good to get some input from @myndzi and @tgriesser on the "release resource" error handling story here.

blah238:
@rhys-vdw I just went ahead and reverted the error handling changes to `client#releaseConnection`. Don't think I want to take that on just yet.

rhys-vdw:
> I just went ahead and reverted the error handling changes to client#releaseConnection.

Thanks. I just got to work, I'll review this evening.

> Don't think I want to take that on just yet.

No stress.

elhigu:
@blah238 I'm doing some cleaning of old pull requests, what do you think about this one? Is this pull request still relevant or should it be moved to issues for discussion maybe?

blah238:
Hi @elhigu, I think it's still needed since AFAIK the issue in the PR this PR superseded ([Fix reference to undefined `client.initialize` method](https://github.com/tgriesser/knex/pull/1009)) is still present, and the synchronous pool initialization API is still inconsistent with pool destruction.

Needs to be rebased of course, been meaning to do that...

blah238:
I'll wait until https://github.com/tgriesser/knex/pull/1279 has been merged since it would conflict with that (Bluebird API changes).

johanneslumpe:
@rhys-vdw @blah238 Since #1279 has been merged, is this going to be merged in as well? I'm wondering if this would help with issues like https://github.com/tgriesser/knex/issues/1481?

blah238:
@rhys-vdw @johanneslumpe I've been out of the loop on Knex development for a while and haven't even looked at what's new in Bluebird, so I think I will have to leave it to another contributor to pick up where this left off. PRs welcome as always of course.

tgriesser:
Closing since we've change pool implementations and so a lot of this code doesn't work anymore. Also I believe the new pool implementation should be a lot simpler than it was previously.

PR #1018, opened 2015-10-12
Add commit and rollback events.

olalonde:


olalonde:
Needed for https://github.com/tgriesser/bookshelf/issues/838

rhys-vdw:
Hey @olalonde. This will require tests and documentation before it can be merged. Can we simplify the event names to just `rollback` and `commit` as well?

Also, I'm still curious about the use case for these...

olalonde:
Thanks for the feedback. I took the names from Rails ActiveRecord: http://guides.rubyonrails.org/active_record_callbacks.html#available-callbacks though I'm not opposed to changing the event names to `rollback` and `commit`. The use case is actually for Bookshelf (tgriesser/bookshelf#838).

Here's a made up use case:

You want to send a welcome email to a user after his account is created. The `created` event kind of works but in more complex transactions the create event might fire before the transaction is committed and if the transaction is rolled back later on, the email will be sent although the user wasn't really created. You could work around this problem by not using events but bookshelf events usually make code easier to understand.

rhys-vdw:
@olalonde Sorry I didn't respond earlier. Yes, tests are still maintained and running on CI. There is a minimal setup guide [here](https://github.com/tgriesser/knex/blob/master/CONTRIBUTING.md#integration-tests). If you're stuck you can always just push the tests and watch them fail on Travis (click 'show all checks') in this thread.

> I took the names from Rails ActiveRecord

Yeah, underscores make sense for Ruby. Seems weird having them in a JS library.

> You could work around this problem by not using events but bookshelf events usually make code easier to understand.

You can attach your handlers to `then` or `catch` in your register user function. You might want to have a method for persisting a user model that doesn't require an email to be sent out - generally it's better to decouple emailing from models.

```js
bookshelf.transaction(transacting =>
  User.forge(data).save(null, {transacting})
    .then(doSomeOtherThing({transacting})
).then(sendWelcomeEmail);
```

olalonde:
> You can attach your handlers to then or catch in your register user function. You might want to have a method for persisting a user model that doesn't require an email to be sent out - generally it's better to decouple emailing from models.

Sure.. was just making this up. I need this for something else which was too long to describe here.

rhys-vdw:
@olalonde No probs, as I said, will accept with aforementioned changes.

olalonde:
Thanks. Changed the event names to 'commit' and 'rollback'.

rhys-vdw:
Sorry, I was referring to the tests and documentation as well. No point adding a feature if nobody knows about it.

jpstone:
I know about it! :) Has this PR been merged?

rhys-vdw:
No merge without tests and docs, sorry. @jpstone, if you want to checkout this branch and do the necessary things I'll do a merge and release.

olalonde:
Just wrote tests. Not sure where docs are supposed to be.

olalonde:
Oh, I suppose it's in `index.html`. Just added some doc in the transaction example.

rhys-vdw:
> PS: you guys should consider git ignoring ./lib and use a prepublish script instead.

Definitely. Will happen at some stage. So much noise in the diffs.

The tests are failing because `knexfile.js` has been deleted.

Sorry for the back and forth, but I was hoping you'd document the events in the [events section](http://knexjs.org/#Interfaces-Events) including the arguments passed etc.

olalonde:
Added events to events section as suggested, also re-added knexfile.js (oops).

rhys-vdw:
[Failed on mariasql](https://travis-ci.org/tgriesser/knex/jobs/91591608#L1342) apparently.

olalonde:
Strange. I don't have mariasql and don't actually need that feature anymore in my code anymore so it's hard justifying so much time on a ~4 liner :) I'd be happy if someone can take over this pull or otherwise feel free to close it.

rhys-vdw:
Okay @olalonde.

I'm closing this because I still don't have a use case - and your not needing it makes me even more skeptical.

@jpstone If you want to revive this please respond with a use case.

PR #1019, opened 2015-10-12
NPM test script improvements

blah238:
- Add `npm run babel` command and run it on `npm test` (supersedes PR #1011 - Ensure /lib is up to date during CI tests)
  - Fix `npm test` command not working on Windows due to Istanbul config issue (see https://github.com/gotwarlost/istanbul/issues/90)
  - Update `CONTRIBUTING.md`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,8 @@
 ## How to contribute to Knex.js
 
-* Make changes in the `/src` directory and run `npm run dev` to
-update the files in `/lib`
+* Make changes in the `/src` directory and run `npm run babel` (runs once and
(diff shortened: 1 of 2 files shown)


rhys-vdw:
Hey @blah238. Great work.

FYI, I found out the other day that you can do `$(npm bin)/_mocha` instead of `node_modules/mocha/bin/_mocha`.

blah238:
Thanks @rhys-vdw!

BTW I don't think `$(npm bin)/_mocha` would work in this case either (resolves to a Bash script, while Istanbul is looking for a .js file).

rhys-vdw:
@blah238 It gets expanded before being passed as an argument, I believe, and is fine. Makes no difference to this PR of course.

PR #1023, opened 2015-10-15
Array values supported on object passed to where functions

chuym:
This PR adds support for passing objects containing array values to the `where` function. This allows for shorter and more concise syntax when querying for anything else than `=`. Example:

Previously:

`Player.where('team', 'foo').andWhere('score_record', '>=', 500).fetchAll();`

Now with this change:

`Player.where({
  team: 'foo',
  score_record: ['>=', 500]
}).fetchAll();
`


chuym:
There are a few issues with my fork. Will reopen later

PR #1024, opened 2015-10-15
Adding support for array values in object passed to 'where' functions

chuym:
This PR adds support for passing objects containing array values to the `where` function. This allows for shorter and more concise syntax when querying for anything else than `=`. Example:

Previously:

`Player.where('team', 'foo').andWhere('score_record', '>=', 500).fetchAll();`

Now with this change:

`Player.where({
  team: 'foo',
  score_record: ['>=', 500]
}).fetchAll();
`

### DIFF
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -239,7 +239,7 @@ assign(Builder.prototype, {
     var boolVal = this._bool();
     var notVal = this._not() ? 'Not' : '';
     for (var key in obj) {
-      this[boolVal + 'Where' + notVal](key, obj[key]);
+      this[boolVal + 'Where' + notVal].apply(this, [key].concat(obj[key]));
     }
     return this;
   },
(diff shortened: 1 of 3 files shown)


chuym:
Looks like the failure was caused by a timeout running a test that doesn't go through the changes in this PR.

rhys-vdw:
@chuym Not sure what the problem is but I reran the test and it's worked this time. I'll have to come back to the review though.

elhigu:
Since this is a new feature it also needs documentation written in index.html

tgriesser:
I'm not sure about this one, I think it gets potentially ambiguous when dealing with array values in Postgres. I'd like to think about it a little more

elhigu:
Cleaning up old pull requests. Moved discussion of this to issue https://github.com/tgriesser/knex/issues/1281

PR #1029, opened 2015-10-21
[doc] Update Examples of defaultTo in Docs  [closes #699]

### LINKED ISSUE #699: Update default timestamp creation docs
Knex now has `knex.fn.now()` which should be documented [here](http://knexjs.org/#Schema-timestamp). This was addressed in #310.

Somehow it seems that the documented solution for SQLite there doesn't work anymore. Using knex.raw("date('now')") produces a syntax error with the latest version of the npm sqlite3 package, which ships libsqlite 3.8.x.

codydaig:
Updated the defaultTo examples in the documentation to reflect the new `knex.fn.now()` function

Refrence #699

### DIFF
--- a/index.html
+++ b/index.html
@@ -1813,8 +1813,8 @@ knex.schema.raw("SET sql_mode='TRADITIONAL'")
       is passed as the second argument.
       <br />
       Note that the method for defaulting to the current datetime varies from one database to another.
-      For example: PostreSQL requires <tt>.defaultTo(knex.raw('now()'))</tt>,
-      but SQLite3 requires <tt>.defaultTo(knex.raw("date('now')"))</tt>.
+      For example: PostreSQL requires <tt>.defaultTo(knex.fn.now())</tt>,
+      but SQLite3 requires <tt>.defaultTo(knex.fn.now())</tt>.
     </p>
 
     <p id="Schema-timestamps">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1813,8 +1813,8 @@ <h3 id="Schema-Building">Schema Building:</h3>
       is passed as the second argument.
       <br />
       Note that the method for defaulting to the current datetime varies from one database to another.
-      For example: PostreSQL requires <tt>.defaultTo(knex.raw('now()'))</tt>,
-      but SQLite3 requires <tt>.defaultTo(knex.raw("date('now')"))</tt>.
+      For example: PostreSQL requires <tt>.defaultTo(knex.fn.now())</tt>,
```
squarejaw: Might be a good idea to fix the typo in "PostgreSQL" that exists in the original line. :wink:

squarejaw:
The way this is worded is a bit confusing since the method doesn't vary any more. Both cases use `.defaultTo(knex.fn.now())`.

codydaig:
@squarejaw Thanks for the feedback, I have made the requested changes. How does that look?

squarejaw:
Looks like there's some extra whitespace to the left of the example code, but other than that it looks good.

codydaig:
@squarejaw Whoops...... Thanks! Fixed.

rhys-vdw:
Looks good. Thanks guys.

PR #1046, opened 2015-11-04
Distinct modifier support for aggregate functions

### LINKED ISSUE #1028: How to do `count(distinct)` ?
I have been unable to find a way with the knex API to perform a count distinct, E.g. `select count(distinct things.id) as aggregate from things...`

`select distinct count(things.id)` is very different to `select count(distinct things.id)`

`.distinct().count()` makes the former, and putting `.count('distinct things.id')` results in an incorrectly quoted query: `select count("distinct things"."id")`

I get the feeling I'm going to have to structure the query as `select count(*) 
from (select distinct things.id from things....`, I'm just hoping there is a way to do it without using subqueries.

Any ideas at all would be very greatly appreciated.

ErisDS:
This adds support for `count(distinct *)` as discussed in #1028.

I have also added support for `avg(distinct *)` and `sum(distinct *)` as both of these seemed valuable, I have not added distinct support for `min` and `max` as I don't think that adding distinct to those aggregates can impact the outcome. If it is felt that these should be added, I'm more than happy to update the PR.

There may be some confusion here between the standard distinct expression, and the distinct expression that goes inside of an aggregate function. I'm not sure whether to refer to it as 'inner distinct' or perhaps 'aggregate distinct'. When building the statement, I used `aggregateDistinct` to differ from the standard `distinct`.

I should probably also rename both `_distinct()` and `_distinctFlag` to also clarify the distinction between the distincts (see what I did there? 😛) but wanted to ask if there were thoughts on the name - naming things is always the hardest part.

I have tested this locally against sqlite3 only.

More than happy to provide more tests or other updates as necessary.

----

closes #1028 

- add support for count(distinct *), avg(distinct *) and sum(distinct *)
- min and max don't really make sense with distinct, so didn't add those


elhigu:
Looks perfect to me. Only thing I would have changed would have been to add changed code and autogenerated files in separate commits :) @rhys-vdw any comments to this one? It had good tests and implementation was pretty much the same with e.g. `not` modifier.

ErisDS:
I can always pull changes out into separate commits if that is desirable, however I think having separate commits for src and built code would be undesirable - it is not useful history & creates a point at which things will not work as you expect. 

I've tried to implement this in an idiomatic way so it fits the library - I felt the naming was definitely confusing so I've updated that. These names are long, but other than calling it `aggDistinct` I wasn't sure where to go with it :grin: happy to change it.

Also updated the docs, good shout @elhigu. Also happy to modify / improve them if anyone has a better idea.

lib/query/builder.js
```diff
@@ -26,6 +26,7 @@ function Builder(client) {
   this._joinFlag = 'inner';
   this._boolFlag = 'and';
   this._notFlag = false;
+  this._aggregateDistinctFlag = false;
```
rhys-vdw: No need to store this on `QueryBuilder` instance.

lib/query/builder.js
```diff
@@ -565,6 +566,21 @@ assign(Builder.prototype, {
     return this._aggregate('avg', column);
   },
 
+  // Retrieve the "count" of the distinct results of the query.
+  countDistinct: function countDistinct() {
+    return this._aggregateDistinct(true).count.apply(this, arguments);
```
rhys-vdw: Instead call `return this._aggregate('sum', column, true)`, and have `_aggregate` take a distinct flag.

lib/query/builder.js
```diff
@@ -742,6 +758,17 @@ assign(Builder.prototype, {
     return ret;
   },
 
+  // Helper to get or set the "distinctFlag" value.
+  _aggregateDistinct: function _aggregateDistinct(val) {
```
rhys-vdw: No longer necessary.

rhys-vdw:
@ErisDS @elhigu: Having generated code in the repository at all is undesirable. Don't worry about it. The solution is to add it to the `.gitignore`, but there will have to be some tweaks to the release script first.

Sorry for the wait, @ErisDS. I approve of the API change.

One thing to change in the implementation before I can merge:

Instead of adding an `_aggregateDistinct()`, just pass an extra argument directly to [`_aggregate`](https://github.com/ErisDS/knex/blob/issue-1028/src/query/builder.js#L782-L791).

```js
  // Retrieve the sum of the distinct values of a given column.
  sumDistinct: function(column) {
    return this._aggregate('sum', column, true);
  },
```

ErisDS:
I have updated this PR with the suggested changes. Reduced surface area :+1: 

I think whether or not generated code should be checked in depends a lot on the project and how it works - there are pros and cons from a git perspective. In this case, as the documentation in the repo depends on the generated code, I don't think it's so bad to have it checked in. I didn't realise how the docs worked at first - but actually, writing docs is a great extra check that your code does what you expect - in fact, the docs almost work like a REPL, very handy.

May I (in a separate PR) update the contribution guidelines with what I learned about the documentation, and that an update there is needed when making api changes?

rhys-vdw:
> I have updated this PR with the suggested changes.

Great, thanks for that.

> I think whether or not generated code should be checked in depends a lot on the project and how it works - there are pros and cons from a git perspective.

The included `knex.min.js` script matters in the `gh-pages` branch, so that can be worked around. Docs, as you are obviously aware, are not generated content. I think the docs would be the same other than missing the script outside of `gh-pages`.

> May I (in a separate PR) update the contribution guidelines with what I learned about the documentation, and that an update there is needed when making api changes?

Yes, please. :+1:

ErisDS:
Another reason why it may be desirable to keep the generated files in git - I just changed my package.json file to `"knex": "git://github.com/tgriesser/knex.git#bd97610"` to check if this fixes an issue I'm encountering in my project (it doesn't... that's another story). 

I know there are both pros and cons to keeping built files in the repo, just wanted to suggest that if it isn't causing major headaches (I don't know if it is or isn't ), changing it may result in some you aren't expecting :wink:

rhys-vdw:
@ErisDS That case can be addressed with an `install` script in `package.json`.

rhys-vdw:
I've done a bit of research on the matter recently as I just made this change to Bookshelf. I'm fairly convinced it's the right way. Note that the generated code is not necessarily correct at each commit - so keeping `lib/` under version control could be more problematic than you realize. Also nobody want to be solving merge conflicts in compiled output.

tgriesser:
I think that having `/src` with a `postinstall` script as react-router does it will be the way to go here, https://github.com/rackt/react-router/blob/620d2703b6314a03b8a54a1fccaac097d436567a/npm-scripts/postinstall.js that way if you have a copy off of github it should also work so long as you have babel installed.

rhys-vdw:
@tgriesser Problem here is that the global version of babel may be incorrect. IMO it's totally valid to call `npm install babel` in the postinstall script and use the `devDependencies` version of Babel. Babel has just undergone some massive changes so you can't rely on configuration being compatible across the board.

I've been advised that the `publish` script is technically the correct place for this. It is also run on `install`, but will run before a release.

rhys-vdw:
[Here](https://github.com/tgriesser/bookshelf/blob/a97fd456e1d0b36cf5990dfd2c4b1c0eaf2c981d/package.json#L14) is how I handled it in Bookshelf. Still requires some work.

ricardogama:
@tgriesser @rhys-vdw This feature is awesome, when can we expect a new release?

Issue #1003, opened 2015-09-29
knex.raw identifier binding ?? does not quote Table/Column name correctly


elhigu:
When I try to use knex.raw to convert case sensitive table / column names to quoted sql identifiers, it does not seem to work as expected:

```js
knex('myTable').where('MyTable.myColumn', '=', knex.raw('??', ['OtherTable.myColumn']) ).toString()
```
outputs:
```sql
select * from "myTable" where "MyTable"."myColumn" = "OtherTable.myColumn"
```

when last column reference should be quoted `"OtherTable"."myColumn"`. Should this work or am I just trying to do it wrong? 

ps. I know I can use knex client internal functions for quoting this, but that is not exposed official API...

rhys-vdw:
What is `??`? It doesn't seem to be documented.

chrisbroome:
@rhys-vdw It's for quoting things that should be identifiers like table names, column names, etc.

For what it's worth, I've also experienced this issue, and I've had to work around it by creating `knex.raw('??.??', [tableName, columnName])` objects and passing them around while `toString`ing in strategic places to make it work. It's definitely not the most ideal situation, but I haven't had to parameterize table/view/column names enough in daily usage for it to be a big deal.

rhys-vdw:
Thanks @chrisbroome. I see there is a test [here](https://github.com/tgriesser/knex/blob/775d6c2af7106b38d53083bd5185c189f2fcd01f/test/tape/raw.js#L12-L18).

We can't automatically split on the `.` because it's possible to have period in a table name, no?

To close this issue, the knew.raw documentation should be updated to include this operator. Probably should include an example of the `??.??` usage.

elhigu:
@chrisbroome nice work around, hadn't thought of that :)

@rhys-vdw I still think that this is a real bug instead of just missing documentation. Even the basic usage of `knex` does not work well if one has `.` in table names. If you have them you can't do any complex queries e.g. having joins and having to refer multiple tables in query.

For example this is how you do simple join:
```js
knex('MyTable').join('OtherTable', 'MyTable.otherId', '=', 'OtherTable.id').toString();
```
which renders to following SQL (dots separate table and column):
```sql
select * from "MyTable" inner join "OtherTable" on "MyTable"."otherId" = "OtherTable"."id"
```

With `Dotty.Table` query above outputs totally invalid output:
```js
[...]

PR #1050, opened 2015-11-06
Fix knex raw indentifier escaping and named parameter regexp and add documentation

elhigu:
Mainly fixed issue #1003 and added documentation. Also accidentally found bug from named parameter replacement regexp  and wrote tests + fixed that too.


chainlink:
+1

rhys-vdw:
Looks great, thanks @elhigu.

Issue #902, opened 2015-07-17
Using '.where' with 'knex.raw' statements and '.fetchOne' from bookshelf 0.8.1 causes an error.


wilsontgh:
```
{ [error: select "cm_vehicles".* from "cm_vehicles" where details ->> 'last_fine_check' < $1 or (details ->> 'last_fine_check') is null order by (details ->> 'last_fine_check') ASC limit $2 - bind message supplies 1 parameters, but prepared statement "" requires 2]
  name: 'error',
  severity: 'ERROR',
  code: '08P01',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  file: 'postgres.c',
  line: '1500',
  routine: 'exec_bind_message' }
```
Here are the statements:
```
qb.where(knex.raw("details ->> 'last_fine_check' < ?", oneDayAgo))
			.orWhere(knex.raw("(details ->> 'last_fine_check') is null"))
			.orderBy(knex.raw("(details ->> 'last_fine_check')"), "ASC");

Model.Vehicle.collection().query(query).fetchOne().then(...)
```
This happens if I use knex.js 0.8.6 but does not happen in knex.js 0.7.6.

rhys-vdw:
Hi @wilsontgh, sorry for the late reply.

The problem here is that `.raw` requires its bindings in an array.

This should work:

```js
qb.where(knex.raw("details ->> 'last_fine_check' < ?", [oneDayAgo]))
            .orWhere(knex.raw("(details ->> 'last_fine_check') is null"))
            .orderBy(knex.raw("(details ->> 'last_fine_check')"), "ASC");
```

An argument could certainly be made for raising an error if a non-array value is supplied as the "bindings" argument for `raw`. Either way the documentation should be made more clear.

By the way, you can simplify your code like so:

```js
qb.whereRaw("details ->> 'last_fine_check' < ?", [oneDayAgo])
  .orWhereRaw("(details ->> 'last_fine_check') is null")
  .orderByRaw("(details ->> 'last_fine_check') asc");
```

rhys-vdw:
Request PR for the following changes:

 - Update documentation to clarify that `knex.raw(expression, bindings)` accepts only an array as its `bindings` argument.
 - Raises an error if bindings is supplied other than `null`, `undefined` or an instance of `Array`.

[...]

PR #1051, opened 2015-11-06
Checking for plain object instead object on raw queries

chuym:
The query builder checks if the bindings are objects in the following form:

    if (this.bindings && typeof this.bindings === 'object') {

This would result in an unexpected behavior if passed an object other than a plain object. For example, if passing a `Date` object, the workaround would be to pass the `Date` object in an array, but that's not very intuitive, given that the `where` function you can pass the `Date` object directly.

This patch uses lodash's `isPlainObject` to do the above check, now we can pass a Date object directly on a raw clause in the same fashion as in the `where` method.

### DIFF
--- a/lib/raw.js
+++ b/lib/raw.js
@@ -7,6 +7,7 @@ var inherits = require('inherits');
 var EventEmitter = require('events').EventEmitter;
 var assign = require('lodash/object/assign');
 var reduce = require('lodash/collection/reduce');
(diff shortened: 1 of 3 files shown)


chuym:
This is a fix for #902

rhys-vdw:
Looks great, thanks. Sorry to not have responded to #902. Thanks for taking the initiative to find the problem and fix it.

PR #1066, opened 2015-11-20
Add migration status method

dustinmartin:
Returns 0 if all migrations are run and DB is up to date. Return negative number if DB is missing migrations. Return positive number if DB is ahead (rare case, e.g. if migrations were run then code is reverted).

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -10,5 +10,5 @@
   "undef": true,
   "trailing": true,
   "unused": true,
-  "predef": [ "-Promise", "before", "after" ]
+  "predef": [ "-Promise", "before", "after", "beforeEach", "afterEach" ]
 }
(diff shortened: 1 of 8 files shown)


dustinmartin:
I've been looking into the failure of this PR and #1067 with no luck. Locally I was running the tests against SQLite and have been attempting to run them against Postgres too (since that is where the failures are happening) but that has been harder than expected. 

I'm seeing the following when running against Postgres:
screen shot 2015-11-22 at 8 44 37 pm

That said, it seems odd that these failures are Postgres specific but I could be missing something. 

I'd be very grateful for some assistance.

dustinmartin:
Any feedback on this PR?

elhigu:
looks good to me.

PR #1067, opened 2015-11-20
Add migration locking

dustinmartin:
Add migration locking so multiple services cannot try to run migrations at same time. This added a new lock table. If migrations are locked and migrations are run by another service it results in an error.

### DIFF
--- a/src/migrate/index.js
+++ b/src/migrate/index.js
@@ -9,6 +9,13 @@ var mkdirp   = require('mkdirp');
 var Promise  = require('../promise');
 var helpers  = require('../helpers');
 var assign   = require('lodash/object/assign');
+var inherits = require('inherits');
+
+function LockError() {
+  this.name = 'MigrationLocked';
+  this.message = 'migrations failed: migration in progress';
+}
+inherits(LockError, Error);
 
 // The new migration we're performing, typically called from the `knex.migrate`
 // interface on the main `knex` object. Passes the `knex` instance performing
@@ -18,8 +25,6 @@ export default class Migrator {
   constructor(knex) {
     this.knex   = knex
     this.config = this.setConfig(knex.client.config.migrations);
-    this.LockError = function() {}
-    this.LockError.prototype = Object.create(Error.prototype);
   }
 
   // Migrators to the latest configuration.
@@ -109,24 +114,11 @@ export default class Migrator {
     var table = this.config.tableName;
     var lockTable = this._getLockTableName(table);
     return this.knex.schema.hasTable(table)
-      .then((exists) => {
-        if (!exists) return this._createMigrationTable(table);
-      })
-      .then(() => {
-        return this.knex.schema.hasTable(lockTable)
-          .then((exists) => {
-            if (!exists) return this._createMigrationLockTable(lockTable);
-          });
-      })
-      .then(() => {
-        return this.knex(lockTable)
-          .select('*')
-          .then((data) => { 
-            if (!data.length) {
-              return this.knex(lockTable).insert({ is_locked: false });
-            }
-          });
-      });
+      .then(exists => !exists && this._createMigrationTable(table))
+      .then(() => this.knex.schema.hasTable(lockTable))
+      .then(exists => !exists && this._createMigrationLockTable(lockTable))
+      .then(() => this.knex(lockTable).select('*'))
+      .then(data => !data.length && this.knex(lockTable).insert({ is_locked: false }));
   }
 
   // Create the migration table, if it doesn't already exist.
@@ -149,21 +141,19 @@ export default class Migrator {
     return tableName + '_lock';
   }
 
-  _isLocked() {
+  _isLocked(trx) {
     var tableName = this._getLockTableName(this.config.tableName);
     return this.knex(tableName)
+      .transacting(trx)
+      .forUpdate()
       .select('*')
-      .then(function(data) {
-        if (data[0].is_locked) {
-          return true;
-        }
-        return false;
-      });
+      .then(data => data[0].is_locked);
   }
 
-  _lockMigrations() {
+  _lockMigrations(trx) {
     var tableName = this._getLockTableName(this.config.tableName);
     return this.knex(tableName)
+      .transacting(trx)
       .update({ is_locked: true });
   }
 
@@ -175,37 +165,39 @@ export default class Migrator {
 
   // Run a batch of current migrations, in sequence.
   _runBatch(migrations, direction) {
-    return this._isLocked()
-      .then(isLocked => {
-        if (isLocked) {
-          throw new Error('migrations failed: migration in progress');
-        }
-        return Promise.all(_.map(migrations, this._validateMigrationStructure, this))
-      })
-      .then(() => this._lockMigrations())
-      .then(() => this._latestBatchNumber())
-      .then(batchNo => {
-        if (direction === 'up') batchNo++;
-        return batchNo;
-      })
-      .then(batchNo => {
-        return this._waterfallBatch(batchNo, migrations, direction)
-      })
-      .then(() => this._unlockMigrations())
-      .catch(this.LockError, error => { 
-        helpers.warn('migrations failed with error: ' + error.message);
+    return this.knex.transaction(trx => {
+      return this._isLocked(trx)
+        .then(isLocked => {
+          if (isLocked) {
+            throw new LockError();
+          }
+          return Promise.all(_.map(migrations, this._validateMigrationStructure, this))
+        })
+        .then(() => this._lockMigrations(trx));
+    })
+    .then(() => this._latestBatchNumber())
+    .then(batchNo => {
+      if (direction === 'up') batchNo++;
+      return batchNo;
+    })
+    .then(batchNo => {
+      return this._waterfallBatch(batchNo, migrations, direction)
+    })
+    .then(() => this._unlockMigrations())
+    .catch(LockError, error => { 
+      helpers.warn('migrations failed with error: ' + error.message);
+      throw error;
+    })
+    .catch(error => {
+      helpers.warn('migrations failed with error: ' + error.message)
+      // If the error was not due to a locking issue, then
+      // remove the lock. Otherwise, leave it in place to
+      // force manual intervention
+      return this._unlockMigrations().finally(function() {
         throw error;
-      })
-      .catch(error => {
-        helpers.warn('migrations failed with error: ' + error.message)
-        // If the error was not due to a locking issue, then
-        // remove the lock. Otherwise, leave it in place to
-        // force manual intervention
-        return this._unlockMigrations().then(function() {
-          throw error;
-        });
       });
-  }
+    });
+}
 
   // Validates some migrations by requiring and checking for an `up` and `down` function.
   _validateMigrationStructure(name) {
--- a/src/migrate/index.js
+++ b/src/migrate/index.js
@@ -186,7 +186,7 @@ export default class Migrator {
     .then(() => this._unlockMigrations())
     .catch(LockError, error => { 
       helpers.warn('migrations failed with error: ' + error.message);
-      throw error;
+      return error;
     })
     .catch(error => {
       helpers.warn('migrations failed with error: ' + error.message)
@@ -194,8 +194,18 @@ export default class Migrator {
       // remove the lock. Otherwise, leave it in place to
       // force manual intervention
       return this._unlockMigrations().finally(function() {
-        throw error;
+        return  error;
       });
+    })
+    .then(result => {
+      // Throw the error so calling code can respond
+      // to any problems if needed
+      if (result instanceof Error) {
+        throw result;
+      }
+      else {
+        return result;
+      }
     });
 }
 
--- a/src/migrate/index.js
+++ b/src/migrate/index.js
@@ -92,9 +92,16 @@ export default class Migrator {
   // dependent on the migration config settings.
   _ensureTable() {
     var table = this.config.tableName;
+    var lockTable = this._getLockTableName(table);
     return this.knex.schema.hasTable(table)
       .then((exists) => {
         if (!exists) return this._createMigrationTable(table);
+      })
+      .then(() => {
+        return this.knex.schema.hasTable(lockTable)
+          .then((exists) => {
+            if (!exists) return this._createMigrationLockTable(lockTable);
+          });
       });
   }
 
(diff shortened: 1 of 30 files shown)


elhigu:
Checkout `test/knexfile.js` you should create database called `knex_test` and user called `postgres` which has full access to that DB. There are different user names for different users. Let me know if you have still problems. 

Would be nice to update  `CONTRIBUTING.md` with examples how to add db / needed user for each database.

dustinmartin:
@elhigu Thanks! I had been trying to create a custom config but creating a default user has worked better. I'll get this PRs fixed now.

I'll also submit a change to CONTRIBUTING.md about creating a new postgres user.

dustinmartin:
Hey all, any feedback on this PR?

src/migrate/index.js
```diff
@@ -108,9 +115,49 @@ export default class Migrator {
...
+      .count('is_locked as count')
+      .then(function(data) {
+        if (data[0].count > 0) {
+          return true;
+        }
+        return false;
+      });
+  }
+
+  _lockMigrations() {
```
elhigu: This probably still allows multiple connections to take lock in the same time and it just creates multiple rows... 

You could initialize table with single row and use transaction to make sure you were the only one who read and set the value of that row e.g. with http://knexjs.org/#Builder-forUpdate

src/migrate/index.js
```diff
@@ -119,10 +166,22 @@ export default class Migrator {
       .then((batchNo) => {
         return this._waterfallBatch(batchNo, migrations, direction)
       })
+      .then(() => this._unlockMigrations())
```
elhigu: Maybe just `.finally(() => this._unlockMigrations())` after the `.catch(...)` part. Then you won't need separate unlock in `.catch`block either.

src/migrate/index.js
```diff
@@ -108,9 +115,49 @@ export default class Migrator {
...
+    return this.knex(tableName).del();
+  }
+
   // Run a batch of current migrations, in sequence.
   _runBatch(migrations, direction) {
-    return Promise.all(_.map(migrations, this._validateMigrationStructure, this))
+    return this._isLocked()
+      .then((isLocked) => {
+        if (isLocked) {
+          throw new Error('migrations failed: migration in progress');
```
elhigu: You can here create own Error e.g. LockingError() type for this, in that way you can have separate error handler in check examples from here: http://bluebirdjs.com/docs/api/catch.html.

src/migrate/index.js
elhigu: Add another error handler `.catch(LockingError, () => {})` which will do nothing in case if locking failed

src/migrate/index.js
```diff
@@ -119,10 +166,22 @@ export default class Migrator {
...
+      .then(() => this._unlockMigrations())
       .catch((error) => {
         helpers.warn('migrations failed with error: ' + error.message)
-        throw error
-      })
+
+        // If the error was not due to a locking issue, then
+        // remove the lock. Otherwise, leave it in place to
+        // force manual intervention
+        if (error.message !== 'migrations failed: migration in progress') {
```
elhigu: And here checking by error type instead of error message would be cleaner (changing message above would break this otherwise...).

test/integration/migrate/index.js
```diff
@@ -34,6 +44,28 @@ module.exports = function(knex) {
...
+                throw new Error('then should not execute');
+              });
+          })
+          .catch(function(error) {
+            expect(error).to.have.property('message', 'migrations failed: migration in progress');
+            // Clean up lock for other tests
+            return knex('knex_migrations_lock').del();
+          });
+      });
+
```
elhigu: Maybe one more test showing that if migration fails, but not because of locking it should still release the lock.

elhigu:
To me feature looks good, correct stuff in correct places, but I would like to see more robust way to do the actual locking.

src/migrate/index.js
dustinmartin: @elhigu I updated the method of locking to use a single record which gets updated but I’m unclear what you mean by using forUpdate in this circumstance. My understanding is that the row will already be locked while it get’s updated. But maybe I misunderstood something.

src/migrate/index.js
```diff
@@ -18,6 +18,8 @@ export default class Migrator {
   constructor(knex) {
     this.knex   = knex
     this.config = this.setConfig(knex.client.config.migrations);
+    this.LockError = function() {}
```
elhigu: `knex` includes `inherits` library and its been used in most of the places: 

e.g. 
```js
var inherits       = require('inherits')
var Client         = require('../../client')

function Client_PG(config) {
  Client.apply(this, arguments)
}
inherits(Client_PG, Client)
```

src/migrate/index.js
```diff
@@ -105,9 +107,25 @@ export default class Migrator {
...
   _ensureTable() {
     var table = this.config.tableName;
+    var lockTable = this._getLockTableName(table);
     return this.knex.schema.hasTable(table)
       .then((exists) => {
         if (!exists) return this._createMigrationTable(table);
+      })
+      .then(() => {
+        return this.knex.schema.hasTable(lockTable)
+          .then((exists) => {
```
elhigu: This `.then` could be in the same level with main promise flow... I don't see any reason here to use nested thens. So end result would be something like: 
```js
.then(() => this.knex.schema.hasTable(lockTable))
.then((exist) => !exist && this._createMigrationLockTable(lockTable))
.then(() => {
  return this.knex(lockTable)...
}
```

elhigu: Also... there might be multiple is_locked rows added if multiple processes are racing to run migrations for the first time, but it probably is not a problem, since only the first row is used :)

src/migrate/index.js
```diff
@@ -121,21 +139,72 @@ export default class Migrator {
...
+  _getLockTableName(tableName) {
+    return tableName + '_lock';
+  }
+
+  _isLocked() {
+    var tableName = this._getLockTableName(this.config.tableName);
+    return this.knex(tableName)
+      .select('*')
+      .then(function(data) {
+        if (data[0].is_locked) {
```
elhigu: `is_locked` is already boolean, could you just return `.then((data) => data[0].is_locked)`

src/migrate/index.js
elhigu: It is locked while it is updated, but reading has already been done before updating, which causes the problems. 
I mean this kind of scenario:
```
Process 1 read lock row: locked: false
Process 2 read lock row: locked: false
Process 1 write lock => locked : true
Process 2 write lock => locked : true
Process 1 start running migrations
Process 2 start running migrations
```
If you make transaction before reading the lock and say that when I select this row I'm going to update it later on in this same transaction, it applies read lock for that row and it is not released until transaction ends. Scenario above will change to:
```
Process 1 start transaction
Process 2 start transaction
Process 1 select read lock row for update: locked: false // select for update causes read lock for row
Process 2 select read lock row for update... // row has already been selected for update in other transaction... must wait
Process 1 write lock => locked : true
Process 1 end transaction
Process 2 read lock row: locked: true // row's read lock was freed... we can continue now
Process 1 start running migrations
Process 2 exit
Process 1 set lock = false
Process 1 exit
```

src/migrate/index.js
```diff
@@ -121,21 +139,72 @@ export default class Migrator {
...
+      .catch(this.LockError, error => { 
+        helpers.warn('migrations failed with error: ' + error.message);
+        throw error;
       })
+      .catch(error => {
+        helpers.warn('migrations failed with error: ' + error.message)
+        // If the error was not due to a locking issue, then
+        // remove the lock. Otherwise, leave it in place to
+        // force manual intervention
+        return this._unlockMigrations().then(function() {
```
elhigu: use `.finally` instead of then, to make sure that error is thrown also if unlock fails.

src/migrate/index.js
```diff
@@ -121,21 +139,72 @@ export default class Migrator {
...
+      .update({ is_locked: false });
+  }
+
   // Run a batch of current migrations, in sequence.
   _runBatch(migrations, direction) {
-    return Promise.all(_.map(migrations, this._validateMigrationStructure, this))
+    return this._isLocked()
+      .then(isLocked => {
+        if (isLocked) {
+          throw new Error('migrations failed: migration in progress');
```
elhigu: Shouldn't this be LockingError thrown to prevent releasing the lock? This case should be catched by the test cases.

elhigu:
@dustinmartin Thanks for the update, I would really like to get this feature to `knex`. There were still some problems and tests are breaking though. I left the comments in overview tab.

src/migrate/index.js
dustinmartin: Oops! I did make the change here but I must have accidentally undid it. Good catch.

dustinmartin:
@elhigu Thanks for all the help/feedback. I *think* I got everything fixed but take another look at let me know if it could be further refined.

src/migrate/index.js
```diff
@@ -121,22 +131,84 @@ export default class Migrator {
...
+    var tableName = this._getLockTableName(this.config.tableName);
+    return this.knex(tableName)
+      .transacting(trx)
+      .forUpdate()
+      .select('*')
+      .then(data => {
+        // This is to handle the different databases
+        // data[0].is_locked == false will also work but
+        // JSHint requires === which won't do triple equals
+        if (data[0].is_locked !== false &&
```
elhigu: This looks really error prone... would this be easier if `is_locked` column type would be e.g. integer or even string?

src/migrate/index.js
```diff
@@ -121,22 +131,84 @@ export default class Migrator {
...
-        throw error
-      })
-  }
+    return this.knex.transaction(trx => {
+      return this._isLocked(trx)
+        .then(isLocked => {
+          if (isLocked) {
+            throw new LockError();
+          }
+          return Promise.all(_.map(migrations, this._validateMigrationStructure, this))
```
elhigu: Verifying migration files could be after the locking transaction. It is a bit strange here in a middle of locking.

src/migrate/index.js
```diff
@@ -121,22 +131,84 @@ export default class Migrator {
...
+      if (direction === 'up') batchNo++;
+      return batchNo;
+    })
+    .then(batchNo => {
+      return this._waterfallBatch(batchNo, migrations, direction)
+    })
+    .then(() => this._unlockMigrations())
+    .catch(LockError, error => { 
+      helpers.warn('migrations failed with error: ' + error.message);
+      throw error;
```
elhigu: Probably this line should not pass error on, because next handler will then catch newly thrown error and unlock the migration table.

elhigu: This case could be tested by trying to start 3 migrations at the same time... 

first takes a lock and starts running, 
second one fails and exit and releases the lock, 
third one will take lock and start running migrations at the same time with the first one.

test/integration/migrate/index.js
```diff
@@ -118,6 +129,59 @@ module.exports = function(knex) {
...
+          .update({ is_locked: true })
+          .then(function() {
+            return knex.migrate.latest({directory: 'test/integration/migrate/test'})
+              .then(function() {
+                throw new Error('then should not execute');
+              });
+          })
+          .catch(function(error) {
+            expect(error).to.have.property('message', 'migrations failed: migration in progress');
+            // Clean up lock for other tests
```
elhigu: Check here that is_locked is still true.

test/integration/migrate/index.js
```diff
@@ -118,6 +129,59 @@ module.exports = function(knex) {
...
+            return knex('knex_migrations_lock').select('*')
+          })
+          .then(function(data) {
+            // This is to handle the different databases
+            // data[0].is_locked == false will also work but
+            // JSHint requires === which won't do triple equals
+            if (data[0].is_locked !== false &&
+                data[0].is_locked !== 0 &&
+                data[0].is_locked !== '0') {
+              throw new Error('is_locked should be false');
```
elhigu: Also in these is better to use `expect()` to tell about errors e.g. `expect(is_locked).to.equal(true)` that way it is easier to understand from code that this is assertion to check correct functionality and not just an internal error in tests.

elhigu:
@dustinmartin Code is starting to look good, but I got just few more remarks / potential bugs when I checked through it again.

src/migrate/index.js
dustinmartin: Maybe an int would be better. Each DB seems to handle booleans differently so that might makes things much simpler. I'll give it a shot.

dustinmartin:
@elhigu Good morning. More changes for you to review.

src/migrate/index.js
```diff
@@ -121,22 +131,83 @@ export default class Migrator {
...
-      })
-  }
+    return this.knex.transaction(trx => {
+      return this._isLocked(trx)
+        .then(isLocked => {
+          if (isLocked) {
+            throw new LockError();
+          }
+        })
+        .then(() => this._lockMigrations(trx));
```
elhigu: Lets assume that here _lockMigrations throw an error, it wont be LockError so it will be passed until the end to normal error handler. But before reaching error handler there is still a chance that some other process did took lock correctly and now lock will be accidentally freed again :)
Some temporary connection error might trigger this. 
Probably the most correct functionality is that only the very same process who acquired the lock must be the one who releases it. So we should catch any exception thrown from transaction and throw LockError instead.

src/migrate/index.js
```diff
@@ -121,22 +131,83 @@ export default class Migrator {
...
+    })
+    .then(() => Promise.all(_.map(migrations, this._validateMigrationStructure, this)))
```
elhigu: So one more catch is needed here to convert any Error thrown from transaction to be `LockError`.

src/migrate/index.js
```diff
@@ -121,22 +131,83 @@ export default class Migrator {
...
-  }
+    return this.knex.transaction(trx => {
+      return this._isLocked(trx)
+        .then(isLocked => {
+          if (isLocked) {
+            throw new LockError();
+          }
+        })
+        .then(() => this._lockMigrations(trx));
+    })
```
elhigu: So one more catch is needed here to convert any Error thrown from transaction to be `LockError`.

src/migrate/index.js
```diff
@@ -121,22 +131,83 @@ export default class Migrator {
...
+    .then(batchNo => {
+      if (direction === 'up') batchNo++;
+      return batchNo;
+    })
+    .then(batchNo => {
+      return this._waterfallBatch(batchNo, migrations, direction)
+    })
+    .then(() => this._unlockMigrations())
+    .catch(LockError, error => { 
+      helpers.warn('migrations failed with error: ' + error.message);
```
elhigu: Also after this change it is possible if migration process crashes the table remains locked. One could mention in error message also how to unlock migrations table manually.

src/migrate/index.js
```diff
@@ -121,22 +131,83 @@ export default class Migrator {
...
+      return error;
+    })
+    .catch(error => {
+      helpers.warn('migrations failed with error: ' + error.message)
+      // If the error was not due to a locking issue, then
+      // remove the lock. Otherwise, leave it in place to
+      // force manual intervention
+      return this._unlockMigrations().finally(function() {
+        return  error;
```
elhigu: if this is `throw error`then one wont need the last then part, which just throws returned error... also `return Promise.reject(error);` would work.

elhigu:
@dustinmartin thanks for the update. I still did find couple of issues, pretty minor ones, but still possible error cases causing invalid results.
If you are getting tired of these review bounces, just tell me, this is already like 98% ready, so I can finish this off myself if necessary :)

dustinmartin:
@elhigu Thanks for the feedback. If you could could finish up the last couple things that would be great. I've got a couple other higher priorities that came up so I need to take care of those. Thanks!

elhigu:
@dustinmartin sure, I'll finish this when I get spare moment

elhigu:
New pull request made on top of @dustinmartin's work found here https://github.com/tgriesser/knex/pull/1094

PR #1094, opened 2015-12-13
Prevent multiple connections from running migrations at the same time

elhigu:
* Adds locking when migrations are started
* forceFreeMigrationsLock function to allow removing the lock manually

### DIFF
--- a/src/migrate/index.js
+++ b/src/migrate/index.js
@@ -11,9 +11,9 @@ var helpers  = require('../helpers');
 var assign   = require('lodash/object/assign');
 var inherits = require('inherits');
 
-function LockError() {
+function LockError(msg) {
   this.name = 'MigrationLocked';
-  this.message = 'migrations failed: migration in progress';
+  this.message = msg;
 }
 inherits(LockError, Error);
 
@@ -112,7 +112,7 @@ export default class Migrator {
   // dependent on the migration config settings.
   _ensureTable() {
     var table = this.config.tableName;
-    var lockTable = this._getLockTableName(table);
+    var lockTable = this._getLockTableName();
     return this.knex.schema.hasTable(table)
       .then(exists => !exists && this._createMigrationTable(table))
       .then(() => this.knex.schema.hasTable(lockTable))
@@ -137,12 +137,12 @@ export default class Migrator {
     });
   }
 
-  _getLockTableName(tableName) {
-    return tableName + '_lock';
+  _getLockTableName() {
+    return this.config.tableName + '_lock';
   }
 
   _isLocked(trx) {
-    var tableName = this._getLockTableName(this.config.tableName);
+    var tableName = this._getLockTableName();
     return this.knex(tableName)
       .transacting(trx)
       .forUpdate()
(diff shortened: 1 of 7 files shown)


elhigu:
@rhys-vdw could you give this the final review. @dustinmartin made this almost ready, but didn't have time to give it a last touch in #1067.

rhys-vdw:
@elhigu Yep, will review this evening.

src/migrate/index.js
```diff
@@ -121,22 +138,89 @@ export default class Migrator {
...
+      .transacting(trx)
+      .update({ is_locked: 1 });
+  }
+
+  _getLock() {
+    return this.knex.transaction(trx => {
+      return this._isLocked(trx)
+        .then(isLocked => {
+          if (isLocked) {
+            throw new Error("Migration table is are already locked");
```
rhys-vdw: "is are" already locked.

src/migrate/index.js
```diff
@@ -121,22 +138,89 @@ export default class Migrator {
...
+  _lockMigrations(trx) {
+    var tableName = this._getLockTableName();
+    return this.knex(tableName)
+      .transacting(trx)
+      .update({ is_locked: 1 });
+  }
+
+  _getLock() {
+    return this.knex.transaction(trx => {
+      return this._isLocked(trx)
```
rhys-vdw: Am I correct in thinking you could still have a problem here unless specifying 'repeatable reads' as the isolation level?

(I may be way off here)

src/migrate/index.js
```diff
@@ -121,22 +138,89 @@ export default class Migrator {
...
-      .then((batchNo) => {
-        if (direction === 'up') batchNo++;
-        return batchNo;
-      })
-      .then((batchNo) => {
-        return this._waterfallBatch(batchNo, migrations, direction)
-      })
-      .catch((error) => {
+    return this._getLock()
+    .catch(err => {
```
rhys-vdw: This `catch` statement could be moved into `_getLock` (after transaction handler) - feels more sensible to fire the `LockError` from there. (or perhaps I'm just bike shedding)

rhys-vdw:
@elhigu Looking good.

src/migrate/index.js
elhigu: What kind of problem? When the transaction of the first connection is has ran `select ... for update;` and then the transaction of the second connection tries to run the same `select ... for update;` the second connection will wait until the first transaction ends before returning result for that query.

src/migrate/index.js
elhigu: Agreed :) I'll change this.

src/migrate/index.js
elhigu: will fix.

src/migrate/index.js
rhys-vdw: I believe that is only the case if the isolation level is set to 'repeatable reads' or higher.

[Wikipedia](https://en.wikipedia.org/wiki/Isolation_(database_systems)#Read_committed) on 'read committed' isolation level or lower:
> ...read committed is an isolation level that guarantees that any data read is committed at the moment it is read. It simply restricts the reader from seeing any intermediate, uncommitted, 'dirty' read. It makes no promise whatsoever that if the transaction re-issues the read, it will find the same data; data is free to change after it is read.

So, if I understand correctly, at this isolation it is possible to have two transaction that issue a `select` before either transaction `update`s the lock table. This means they would each redundantly `update` the table and try to execute the migrations. Of course the chance of this happening would be greatly reduced since they would need to line up perfectly.

I'm not sure exactly how this would be configured since Knex doesn't have an isolation level interface - so perhaps this is a moot point.

rhys-vdw: @elhigu Oops, missed `forUpdate`. Don't mind me.

dustinmartin:
@elhigu Thanks or finishing this up. Very much appreciated.

When do you think the next release of knex will be that will include these changes?

rhys-vdw:
@dustinmartin Should do one soon. @elhigu is there anything else you'd like to see in the next release?

elhigu:
@rhys-vdw there is still that one pull request about sqlite3 dialect escaping which I would like to get in. It fixes also some regressions from postgres string literal escaping from getting to release.

PR #1100, opened 2015-12-17
2616: removed double words

IsaacLessard:
deleted 'specifies' from ...object, which may specify specifies the... on line 2616


blah238:
Thanks @IsaacLessard!

Issue #823, opened 2015-05-13
There is no pool defined on the current client


cavinsmith:
Knex 0.8.4
pg 4.3.0
Postgres 9.4
Centos6 

FYI: Knex 7.6 @ same conditions works perfeclty.

blah238:
I think you meant Knex 0.8.4, but I'm also seeing this after updating from 0.8.3.

cavinsmith:
@blah238 yep, 0.8.4

kswope:
Same problem.  I'm a knex noob so I just stumbled into the solution in minutes.  Must not be a default pool setting in the connection string anymore. 

Here is the example from the docs
```
 var knex = require('knex')({
  client: 'mysql',
  connection: {
    host     : '127.0.0.1',
    user     : 'your_database_user',
    password : 'your_database_password',
    database : 'myapp_test'
  },
  pool: {
    min: 0,
    max: 7
  }
});
```

pkinney:
This is occurring if no pool configuration is defined when creating the client.

With the logic added in https://github.com/tgriesser/knex/commit/054be4606808e49577c931b78e20ee2928c87d89, if not pool object is specified in the config options, the pool is not initialized.

```javascript
if (config.pool && config.pool.max !== 0) {
  this.initializePool(config);
}
```

kswope:
Is this a production ready library?  I've never seen a lib completely remove a default setting suddenly.  Usually if such a show stopping change was to be made a deprecation warning would show up in the logs somewhere.  Now I'm afraid to move forward with this.

blah238:
@kswope I assume that since it wasn't documented and such deprecation notices have been made in the past that this is just a bug. I could be wrong, of course.

natorojr:
@kswope Knex is definitely a production ready library -- and an awesome one at that.

Unfortunately, seems like there was a slight oversight in commit https://github.com/tgriesser/knex/commit/054be4606808e49577c931b78e20ee2928c87d89

I'm sure Tim did not intend to break existing code. 

I think the fix is simply:

```javascript
if (!config.pool || (config.pool && config.pool.max !== 0)) {
  this.initializePool(config);
}
```
I'll submit a pull request in a few hours.

In the meantime, you'll need to explicitly add pool entry to your config.

tgriesser:
Wow. Can't believe I missed this, the tests weren't even doing what I'd meant them to either - sorry for the breakage here.

natorojr:
[...]

PR #1101, opened 2015-12-17
Fix error with CLI trying to select environment config for "basic configuration"

galenandrew:
## Issue

Per the docs for **CLI knexfile.js**, the following configuration should be supported:
> Basic configuration:
> 
```javascript
module.exports = {
  client: 'pg',
  connection: process.env.DATABASE_URL || {
    user: 'me',
    database: 'my_app'
  }
};
```

Per the following conditional block in [`/src/bin/cli.js#L56-L59`](https://github.com/tgriesser/knex/blob/master/src/bin/cli.js#L56-L59) (introduced in #527) there is no check to make sure `config[environment]` exists.
```javascript
  if (environment) {
    console.log('Using environment:', chalk.magenta(environment));
    config = config[environment];
  }
```
**This results in an undefined config when using the basic configuration**.

## Pull Request Details

This pull request includes:

- `config` assignment defaults to itself when `config[environment]` does not exist
- update to the docs indicating that `NODE_ENV` is also used to set the environment in Knex CLI<br>_(this was my issue originally…I did not know the CLI environment defaults to `NODE_ENV`)_

_NOTE: I branched off `master`, but after following the CONTRIBUTING guidelines other files in `lib` showed changes after building that my code would not have affected. My assumption is that the latest build in `master` was dirty so I have only included src changes in this PR._

### DIFF
--- a/index.html
+++ b/index.html
@@ -2519,12 +2519,15 @@ $ knex migrate:latest
 </pre>
 
     <p>
-      You can also pass the <tt>--env</tt> flag to select an alternative
-      environment:
+      You can also pass the <tt>--env</tt> flag or set <tt>NODE_ENV</tt> to select an alternative environment:
     </p>
 
 <pre>
 $ knex migrate:latest --env production
+
+# or
+
+$ NODE_ENV=production knex migrate:latest
 </pre>
 
     <h2 id="Seeds">Seed files</h2>
--- a/src/bin/cli.js
+++ b/src/bin/cli.js
@@ -53,9 +53,10 @@ function initKnex(env) {
   if (!environment && typeof config[defaultEnv] === 'object') {
     environment = defaultEnv;
   }
+
   if (environment) {
     console.log('Using environment:', chalk.magenta(environment));
-    config = config[environment];
+    config = config[environment] || config;
   }
 
   if (!config) {


rhys-vdw:
Fixes #1103.

rhys-vdw:
@galenandrew, thanks for the PR!

re #823:

> This of course made the migrations and CLI work properly (although does not follow the docs), but then introduced the pooling error because the config object being passed into require('knex')(config) included the nested environment configs.

I'm thinking that `Knex()` might also accept a "non-basic" config (ie. `{ development: .., production: ...}`) for complete parity between `knexfile.js` and `Knex()`. What do you think @galenandrew, @tgriesser?

galenandrew:
@rhys-vdw I love that idea (updating `Knex()` to accept a non-basic config). What are your thoughts on having it fallback to looking for a knexfile if no config is passed into the `Knex()` constructor?

rhys-vdw:
@galenandrew I considered that and think that it is probably not a good idea. I'd prefer for it to be supplied explicitly. IMO `Knex()` should create a connection-less `QueryBuilder` interface, rather than searching for a default connection silently. (Currently this is achieved by calling `Knex({})`.)

galenandrew:
Ah…you make a good point.

PR #1109, opened 2015-12-23
Add option to timestamps to be not null and default to the current timestamp

### LINKED ISSUE #547: should timestamps() set default values for updated_at and created_at?
I would expect this to be the case however it currently does not behave in this ~~manor~~ manner.

kmikael:
This closes #547. I've added a third parameter to `timestamps` so it will be backwards compatible.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1875,6 +1875,12 @@ knex.schema.raw("SET sql_mode='TRADITIONAL'")
       <br />
       Adds a <tt>created_at</tt> and <tt>updated_at</tt> column on the database,
       setting these each to <a href="#Schema-dateTime">dateTime</a> types.
+      <br />
+      When true is passed as the first argument a <tt>timestamp</tt> type is used.
+      <br />
+      Both colums default to being not null and the current timestamp when true
+      is passed as the second argument.
+      <br />
     </p>
 
     <p id="Schema-binary">
--- a/src/schema/tablebuilder.js
+++ b/src/schema/tablebuilder.js
@@ -144,12 +144,13 @@ _.each(columnTypes, function(type) {
     // The "timestamps" call is really a compound call to set the
     // `created_at` and `updated_at` columns.
     if (type === 'timestamps') {
-      if (args[0] === true) {
-        this.timestamp('created_at');
-        this.timestamp('updated_at');
-      } else {
-        this.datetime('created_at');
-        this.datetime('updated_at');
+      var col = (args[0] === true) ? 'timestamp' : 'datetime';
+      var createdAt = this[col]('created_at');
+      var updatedAt = this[col]('updated_at');
+      if (args[1] === true) {
+        var now = this.client.raw('CURRENT_TIMESTAMP');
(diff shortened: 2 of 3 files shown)


elhigu:
Needs still updates to documentation in `index.html`. Otherwise looks good to me.

rhys-vdw:
Great, this should ultimately be the default. `now()` is not supported by MS SQL, so really we need a portable function `now()` for each dialect to do this properly.

rhys-vdw:
Just to be clear: I think it will be okay to merge this as is for now. If you are interested in going the extra mile and adding in per-dialect support for `now()` I'd like to make this the default behaviour.

kmikael:
If I used `.raw('CURRENT_TIMESTAMP')` instead of `.raw('now()')`. I think that would work fir the majority of SQL databases?

kmikael:
So should I add documentation to `index.html` mentioning that if `true` is passed as the second argument that default values will be used?

Btw, I remember that the documentation used to mention something about the first parameter. i didn't see anything in the history. But the documentation for `timestamps` doesn't mention anything anymore…

kmikael:
I've used `CURRENT_TIMESTAMP` like I mentioned and I've updated the documentation for `timestamps`.

elhigu:
thanks kmikael :+1:

galenandrew:
As this changes has not been released/published yet, can we continue the discussion around the point @rhys-vdw makes in his [earlier comment](https://github.com/tgriesser/knex/pull/1109#issuecomment-167575960) about having the default behavior? Essentially making the inverse logic – if no arguments are passed, the columns default to timestamp/now and setting the second argument to `false` set's them to null.

Thoughts? Happy to submit a PR if this is the desired behavior.

kmikael:
@galenandrew In my opinion, setting the current timestamp should be the default behavior.

galenandrew:
After looking at #1209, I also think it's worth noting that the default behavior should include setting "on update current timestamp" for the `updated_at` field.

For instance…
```js
table.timestamps()
```
…would set the following
```sql
"created_at" datetime default CURRENT_TIMESTAMP
"updated_at" datetime default CURRENT_TIMESTAMP on update CURRENT_TIMESTAMP
```

I'll work on a PR soon, but have a few other things on my plate ATM. If anyone else wants to take a stab at a PR :+1:!

rhys-vdw:
> `"updated_at" datetime default CURRENT_TIMESTAMP on update CURRENT_TIMESTAMP`

@galenandrew This might not be desirable as a default.  Also is that this would make it difficult to change records without triggering the update. There are times when a DBA might want to modify records en masse (in a data migration, say) while preserving the existing `updated_at` value. Not sure if this is compelling.

Also note that it is [less trivial in dialects other than MySQL](http://pointbeing.net/weblog/2008/03/mysql-versus-postgresql-adding-a-last-modified-column-to-a-table.html).

If you'd like to continue this discussion please open a new issue as this one is closed.

Issue #1052, opened 2015-11-10
Transaction promise is mutating after being resolved (resolves to undefined after a while)


gajus:
Detail description https://github.com/petkaantonov/bluebird/issues/863

elhigu:
Verified bug, here is test case:

```js
test('#1052 - transaction promise mutating', function() {
  var transactionReturning = knex.transaction(function(trx) {
    return trx.insert({id: 1, name: 'A'}).into('test_table');
  });
  return Promise.all([transactionReturning, transactionReturning])
    .spread(function (ret1, ret2) {
      if (ret1 !== ret2) throw new Error("Promise was mutated when it was resolved twice.");
    });
});
```

chuym:
Bluebird checks if the passed objects are instances of the `Promise` class in the `all` method on [this line](https://github.com/petkaantonov/bluebird/blob/master/src/promise_array.js#L21), otherwise it checks if it's a "Promise-like" class, (which `Transaction` is) and then does some failover functionality I haven't delved into yet. 

A quick workaround is to pass the `Promise` object and that should work:
````
test('#1052 - transaction promise mutating', function() {
  var transactionReturning = knex.transaction(function(trx) {
    return trx.insert({id: 1, name: 'A'}).into('test_table');
  });
  return Promise.all([transactionReturning._promise, transactionReturning._promise])
    .spread(function (ret1, ret2) {
      if (ret1 !== ret2) throw new Error("Promise was mutated when it was resolved twice.");
    });
});
````
I sent a fix on #1114 which makes `Transaction` a "Bluebird-like" promise object

**Comment by rkaw92** [comment]
This is really bad. Here is a test case which depicts just how dangerous this is:
```js
'use strict';

const knex = require('knex');
const connection = knex({
	client: 'sqlite3',
	connection: ':memory:'
});

const initPromise = connection.transaction(function(transaction) {
	// The body of the transaction probably does not matter:
	return transaction.schema.createTable('users', function(table) {
		table.text('userID');
		table.text('password');
	});
});

initPromise.then(function() {
	console.log('First .then() fulfilled');
	// Introduce the mutation:
	initPromise.then(function() {
		throw new Error('Promise overwriter');
	});

	setTimeout(function() {
		initPromise.then(function() {
			console.log('Second .then() fulfilled');
		}, function(error) {
			console.log('Second .then() ERROR: %s', error);
		});
	}, 2000);
});
```

[...]

PR #1114, opened 2015-12-27
Issue #1052 patch

chuym:
This patch fixes issue #1052 

This patches makes the `Transaction` class compatible with Bluebird's inner functions, notably `Promise.all`, where an iterable of `Promise` objects is expected or an object that responds to the `Promise` interface (like `Transaction`). to do so, two changes were made:
1. The `Transaction._promise` property was renamed to `Transaction._promise0`, in order to comply with [Bluebird](https://github.com/petkaantonov/bluebird/blob/v2.9.24/src/thenables.js)
2. An new instance method `Transaction._then` was defined as well for the same reasons above but for [this line](https://github.com/petkaantonov/bluebird/blob/v2.9.24/src/thenables.js#L16)

An alternate solution is to make `Transaction` inherit from `Promise`, this way it will make it a Promise instance, but that may lead to unexpected behavior given how `Transaction` is proxying function calls to the inner promise object. Maybe a refactor is due to make `Transaction` a `Promise` first, and then provide an interface to make SQL calls on it?

### DIFF
--- a/lib/bin/cli.js
+++ b/lib/bin/cli.js
@@ -53,9 +53,10 @@ function initKnex(env) {
   if (!environment && typeof config[defaultEnv] === 'object') {
     environment = defaultEnv;
   }
+
   if (environment) {
     console.log('Using environment:', chalk.magenta(environment));
-    config = config[environment];
+    config = config[environment] || config;
   }
 
   if (!config) {
--- a/lib/connection/index.js
+++ b/lib/connection/index.js
@@ -1,12 +1,6 @@
 'use strict';
 
-Object.defineProperty(exports, '__esModule', {
-  value: true
-});
-
-var _createClass = (function () { function defineProperties(target, props) { for (var i = 0; i < props.length; i++) { var descriptor = props[i]; descriptor.enumerable = descriptor.enumerable || false; descriptor.configurable = true; if ('value' in descriptor) descriptor.writable = true; Object.defineProperty(target, descriptor.key, descriptor); } } return function (Constructor, protoProps, staticProps) { if (protoProps) defineProperties(Constructor.prototype, protoProps); if (staticProps) defineProperties(Constructor, staticProps); return Constructor; }; })();
-
-var _get = function get(_x, _x2, _x3) { var _again = true; _function: while (_again) { var object = _x, property = _x2, receiver = _x3; _again = false; if (object === null) object = Function.prototype; var desc = Object.getOwnPropertyDescriptor(object, property); if (desc === undefined) { var parent = Object.getPrototypeOf(object); if (parent === null) { return undefined; } else { _x = parent; _x2 = property; _x3 = receiver; _again = true; desc = parent = undefined; continue _function; } } else if ('value' in desc) { return desc.value; } else { var getter = desc.get; if (getter === undefined) { return undefined; } return getter.call(receiver); } } };
+exports.__esModule = true;
 
 function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError('Cannot call a class as a function'); } }
 
@@ -20,7 +14,7 @@ var Connection = (function (_EventEmitter) {
   function Connection(connection) {
     _classCallCheck(this, Connection);
 
-    _get(Object.getPrototypeOf(Connection.prototype), 'constructor', this).call(this);
+    _EventEmitter.call(this);
     this.connection = connection;
 
     // Flag indicating whether the connection is "managed",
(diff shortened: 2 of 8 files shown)


chuym:
Looks like the test timed out on its own?

rhys-vdw:
Hey @chuym,

This looks good. I reran that test and it's resolved nicely. Happy to merge this, but could you give me some context about what changed here?

I'm wondering how much work it would be to remove Bluebird as a dependency to Knex and require consumer code to initialize Knex by passing a preferred `Promise` constructor - eg. Bluebird's `Promise`, or the new standard ES Promise. Any thoughts on this?

chuym:
@rhys-vdw Sure thing, the changes are done on the `Transaction` class, right now the `Transaction` class acts as a Proxy for the inner bluebird's `Promise` object by [exposing a few methods](https://github.com/tgriesser/knex/blob/master/src/transaction.js#L261) and then calling them in the context of the `Promise` itself, this allows to work on `Transaction` objects as if they were `Promise` instances in principle.

The bug reported happens when using bluebird's `Promise.all` method and passing, `Transaction` instances to it, bluebird does a check if it is a "thenable" object (that is, it implements a sort of Promise interface), which `Transaction` does, when that happens it does an additional check to see if it is a bluebird's `Promise` interface, for that to happen the object must have a property named `_promise0` [see this line](https://github.com/petkaantonov/bluebird/blob/master/src/thenables.js#L50) if that's the case, then it uses the object's `then` method at all times. I believe this is what we want to do in knex with the `Transaction` object right now since we are exposing bluebird's Promise API.

The previous implementation does not passes the bluebird's check if it's a bluebird Promise, so it takes another path, basically bluebird creates a helper promise, uses its own resolver/reject functions but calls the passing object `then` function, [check this line](https://github.com/petkaantonov/bluebird/blob/master/src/thenables.js#L62)

If you check a bit below, you can see bluebird's [resolver functions](https://github.com/petkaantonov/bluebird/blob/master/src/thenables.js#L70) they don't return the value, so this means that if we call `then` again, the value is lost.

This is perhaps an upstream bug with bluebird's promise (I'll investigate further) but if the `Transaction` class complies with the `isAnyBluebirdPromise` check, then it will work as expected because it will assume it is a `Promise` instance. This is why I pointed out it may be better for `Transaction` to be a `Promise` first, then implement the query methods on top of it.

ES6 promises is minimal and doesn't provide some of the functions currently used on knex, like `spread`, `tap` among others, an alternative would be to define these functions as util functions in knex in case they are not defined, this may be a risky move given the high reliance on bluebird's promises right now, but could provide the additional benefit of having a clearer `Promise` interface

rhys-vdw:
Thanks for the detailed explanation, @chuym.

> If you check a bit below, you can see bluebird's resolver functions they don't return the value, so this means that if we call then again, the value is lost.

That's confusing. I'm going to look into this more later.

> when that happens it does an additional check to see if it is a bluebird's Promise interface, for that to happen the object must have a property named _promise0

Hm. Okay, that makes sense. I'm a little uncomfortable with depending on an implementation detail of Bluebird - since we'll be liable to run into similar problems again. Also we're tricking Bluebird into thinking that we're supplying a Bluebird promise, but obviously it is not! Inheriting from a promise seems wrong - but then again so does creating a "thenable" that is not a plain promise - and this is how all of Knex works.

This sounds like the correct approach for Knex for now.

> ES6 promises is minimal and doesn't provide some of the functions currently used on knex, like spread, tap among others

It's true - and most notably they do not support multiple arguments. I'm not sure if that's a feature used by Knex, but the others are simple to work around:

```js
Promise.all([ap, bp, cp]).spread((a, b, c) =>
// becomes
Promise.all([ap, bp, cp]).then(([a, b, c]) =>
```

> an alternative would be to define these functions as util functions in knex in case they are not defined

I'm thinking we can write simple helpers for `tap`, `props` etc. without actually mutating the `Promise` instance passed to the knex constructor:

```js
tap = (func) => (value) => { func(); return value; }

Promise.resolve('a')
  .then(tap(() => console.log('blah!'))
  .then(a => console.log('this is the first letter of the alphabet: ', a));
```
```js
props = (obj) => Promise.all(_.values(obj)).then(resolved =>
  _.zipObject(_.keys(obj), resolved)
);
```

chuym:
@rhys-vdw I've sent a PR for Bluebird's that should solve the problem encountered here, basically there is a type of thenable object that is not correctly handled by bluebird, that is, a thenable object that keeps its resolved value in an internal property here: https://github.com/petkaantonov/bluebird/pull/935

This could be considered as a temporary patch until the issue is fixed upstream or we can wait until the issue is fixed.

Another issue I see is that users may expect the Promises returned by various knex functions, and even bookshelf functions to a certian extent, to behave like bluebird, or at least to expose the same API as bluebird, that is, end users may be using methods like `spread` or `tap` from returned Promises object or expect methods like `all` to behave a certain way.

rhys-vdw:
@chuym I'm waiting to see how the Bluebird PR pans out... I noticed your fiddle actually doesn't log `undefined` where it's meant to be demonstrating failure.

elhigu:
This doesn't look like a correct fix (relays on bluebird implementation).

Issue #138, opened 2013-12-17
Replication - Master/Slave


whito:
Question or Feature Request: 

Is it currently possible to utilize replication (master/slave) with knex connections? Perhaps be able to pass an array of different connections during initialize and define their capabilities (read only, read write, write)

This way you can set up multiple databases on different boxes for scaling, redundancy, performance etc.

jonathandelgado:
@tgriesser Are you still interested in implementing this? I'm not familiar enough with Postgres, but I can throw in a PR for the MySQL poolCluster solution I mentioned earlier in the issue.

daviddutch:
@tgriesser any news on this? I would need this on Postgres.

rvlander:
+1

rvlander:
Any news on this? If nothing have been done I will have to do something myself. Did you have any plan, ideas, vision on how to implement it?

rhys-vdw:
@rvlander It doesn't sound like it. You're encouraged to try this as it seems like a popular feature. I'm not sure what Tim's status is at the moment but he's been busy. He's requested above that somebody implements it, so go for it. I can't provide much guidance but can merge when you're done.

jonathandelgado:
@rvlander If you need this now, just create a pool of knex instances and round-robin between them. That is what I am doing currently and it works great. You might need some kind of health monitoring depending on your application.

avimar:
I plan to have a mysql master-master, and for maintenance purposes would shut one down for 3-30 minutes. Is there a way to do that without downtime?

If a query times out to mysql, will it try the next connection?

Is there a way get mysql to stop responding to pings but finish the current query, so it times out from the pool? Would it keep trying and add itself back later?

codematix:
+10

Eli-Goldberg:
+10

ewdave:
Hi, any update on this feature on knex yet?

elhigu:
Moving discussion to https://github.com/tgriesser/knex/issues/2253 where has been collected links to many related issues.

PR #1129, opened 2016-01-08
WIP: adds suport for cluster of nodes.

rvlander:
This PR aimed at beeing able to specify a cluster of nodes to Knex (each node running a instance of the database). It is particularly useful if you have want to use a multi-master replication system #138. 

Currently only PostgreSQL is supported by this PR, but a little work could be done to make it available to all dialects.

Here is an example of config that needs to be passed to Knex:

``` javascript
var dbConfig = {
  client: 'postgresbdr',
  connections: [
    {
      host: '192.168.1.1',
      port: '5432',
      database: 'blog',
      user: 'user',
      password: 'pass'
    },
    {
      host: '192.168.1.2',
      port: '5432',
      database: 'blog',
      user: 'user',
      password: 'pass'
    }
  ]
};
```

With this config Knex tries to do its request on whatever node of the cluster.

Any comments and help are welcome!

### DIFF
--- a/.jshintignore
+++ b/.jshintignore
@@ -1,0 +1,2 @@
+test/coverage/.
+test/integration/migrate/migration/.
--- a/.jshintrc
+++ b/.jshintrc
@@ -10,5 +10,6 @@
   "undef": true,
   "trailing": true,
   "unused": true,
+  "esnext": true,
   "predef": [ "-Promise", "before", "after", "beforeEach", "afterEach" ]
 }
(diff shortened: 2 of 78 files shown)


tgriesser:
Hey @rvlander sorry it's taken me so long to reply to this.

For 1.0, I'm planning on making it so end-users can easily provide their own pool implementation, so folks can configure strategies for acquiring pooled connections. Will close this ticket once that happens

elhigu:
I don't think posgresql specific solution is way to go here.

PR #1142, opened 2016-01-18
Returning doc update

jurko-gospodnetic:
`returning()` docs did not explain the parameter structure passed to this API.

It accepts a single parameter which can be either a string or a list of strings and depending on which you send, the final query result is either a list of column values or a list of objects containing column values as properties.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1246,6 +1246,11 @@ knex('coords').insert([{x: 20}, {y: 30},  {x: 10, y: 20}])
       <br />
       Only utilized by PostgreSQL databases, the returning method specifies which column should be returned
       by the <a href="#Builder-insert">insert</a> and <a href="#Builder-update">update</a> methods.
+      <br />
+	  Passed <tt>column</tt> parameter may be a string or an array of strings. When passed in a string,
+	  makes the SQL result be reported as an array of values from the specified column. When passed in an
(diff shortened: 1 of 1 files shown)


jurko-gospodnetic:
Ping?

wubzz:
@rhys-vdw Are you ok with merging this? It checks out.

It would make more sense to have `.returning` always return Objects to have a consistent resultset, but that would be a **breaking** change. Is that something to consider?

jurko-gospodnetic:
> It would make more sense to have .returning always return Objects to have a consistent resultset, but that would be a breaking change. Is that something to consider?

Yeah, I think the interface would have been nicer if that were the case, but it's definitely something that should not be done in this pull request. :smile: And yup, that would likely break a LOT of code. But at least the fireworks would be nice... :laughing:

rhys-vdw:
> It would make more sense to have .returning always return Objects to have a consistent resultset

Agreed. Consistency is good. I guess that functionality might be supported by a function with a name like `pluck` or something? I'm not sure that this is really important. If either of you think it would be good to change, please open an issue.

> Are you ok with merging this?

Generally speaking any documentation updates that clarify how to use existing functionality is good. It will be simple to deprecate this later if necessary. Perhaps a consideration before v1.0.

PR #1143, opened 2016-01-18
Create If Not Exists

henrygiocondo:
The IF condition was wrote wrong, instead of 'create if not exists' it was 'create if exists'

### DIFF
--- a/lib/dialects/mssql/schema/tablecompiler.js
+++ b/lib/dialects/mssql/schema/tablecompiler.js
@@ -21,7 +21,7 @@ assign(TableCompiler_MSSQL.prototype, {
 
   createAlterTableMethods: ['foreign', 'primary', 'unique'],
   createQuery: function createQuery(columns, ifNot) {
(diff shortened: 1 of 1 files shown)


rhys-vdw:
Any chance we could get a test for this one?

henrygiocondo:
How do i? Im using my correction on MSSQL and its ok... its just a logic
case... object_id returns the id of the table, so, it was checkin if
object_id is not null to create, if it is not null then it already exists
on the database...

I just changed it to object_id is null (the object does not exists on the
db) so create.

If you need a test just tell me how to do.


2016-01-18 20:40 GMT-02:00 Rhys van der Waerden <notifications@github.com>:

rhys-vdw:
@henrygiocondo, I just noticed that you've modified the wrong file. You must edit the file under `/src`, not `/lib`. `/lib` is generated from `/src` by Babel.

Also, if you don't mind doing the test, just add this case to this file: https://github.com/tgriesser/knex/blob/master/test/unit/schema/mssql.js

Thanks. :+1:

henrygiocondo:
I changed the file on src and created the test case. Closing

Issue #599, opened 2014-12-16
Ability to define `collate` for a column in MySQL?


danielgindi:
Mainly for case-sensitive etc.

Any way for doing that today without jumping through [too many] hoops?

elhigu:
thanks @alexrsagen !

danielgindi:
Thanks @alexrsagen ! :-)

mercmobily:
This should have been done for all dialects. Oh well...

PR #1147, opened 2016-01-19
Add collate for columns (MySQL)

alexrsagen:
This had been requested in #599 a year ago, but a PR was never created.

I have included a test that passed when i ran it.


alexrsagen:
@rhys-vdw Please have a look at this.

meepdarknessmeep:
very useful please add

tjwebb:
@alexrsagen can you add docs for this feature? Otherwise, lgtm

alexrsagen:
@tjwebb Is this okay? 2fae475

alexrsagen:
I have no idea how that one check failed, but i'm sure i didn't touch that, and the only thing i just changed was index.html... It said it timed out, so must be slow vm or old Node.js version causing that.

elhigu:
I re-triggered the test :) it was just usual timeout when some test doesn't complete in 5000ms... Not sure if there is real problem in somewhere in our codebase or is it just random lagging of travis, which causes it. It happens maybe once in every 20 build or so.

alexrsagen:
@elhigu Great, thanks :)

I think the timeout issue is more than likely a problem with travis, as they create a very small VM or container for every test they run, so they can't use many resources.

Could you or another collaborator merge this please? Currently i'm using `table.string("email").comment("' collate 'utf8_unicode_ci")` to get past without the function i'm committing here :)

elhigu:
@alexrsagen code looks fine to me. We are trying to push out 0.10 so I would rather merge this after it is released... we hope to be able to make releases in faster cycle in future.

alexrsagen:
@elhigu Ah i see. Thanks for clarifying :)

PR #1149, opened 2016-01-21
Remove generated lib from repository and run build in prepublish

elhigu:
Generated code in lib directory has caused lots of problems, when trying to get multiple pull requests to be merged to repo at the same time. It also generated lots of noise in code reviewing.


rhys-vdw:
Awesome!

PR #1150, opened 2016-01-23
Updates version ref in npm install

TomNeyland:


tgriesser:
Thanks - I think we decided to just take out the version in the docs alltogether, in which case npm will install the latest.

Issue #483, opened 2014-09-12
Document index, constraint, timestamp col drop methods


danschumann:
I'm looking to put a remove index command on a `down` migration, but not seeing one in the docs, is there a removeIndex?

djmadcat:
+1

blah238:
FYI `dropIndex` is now documented: https://github.com/tgriesser/knex/blob/05ffb457a885c9a2d27a0684b57cbcf45cbb0456/index.html#L1908

techniq:
Also need to document `foreign()` to create foreign key references after a column already exists.  I just replaced some existing foreign key references so I could add cadcade deletions using:

```js
exports.up = function(knex, Promise) {
  return knex.schema
    .table('review', function (table) {
      table.dropForeign('event_id')
      table.foreign('event_id')
        .references('event.id')
        .onDelete('CASCADE')
        
      table.dropForeign('user_id')
      table.foreign('user_id')
        .references('user.id')
        .onDelete('CASCADE')
    })
})
```

rhys-vdw:
Thanks @techniq, updated Ben's post.

rhys-vdw:
Would you be interested in supplying a PR, @techniq?

techniq:
@rhys-vdw See PR #1156

rhys-vdw:
Thanks @techniq.

techniq:
Welcome @rhys-vdw. You can check those off of @bendrucker's list ;)

rhys-vdw:
Already done.

techniq:
Thanks.  My page needed a hard refresh to see.  Guess just comments are websocket streamed in.

rhys-vdw:
I figured something like that. I'll update `gh-pages` later when I'm back from work.

PR #1156, opened 2016-01-27
Add docs for table foreign key modification functions, foreign/dropForeign

techniq:
Issue #483


rhys-vdw:
Beautiful. Cheers @techniq

techniq:
Had an extra `>` and missing `)` after reviewing a little more closely.  See #1157

PR #1161, opened 2016-01-27
[Docs] Fixes #1148, closes errant <code> tag in docs

### LINKED ISSUE #1148: `references` function taking <table>.<column> syntax is undocumented
It's used in the [Bookshelf docs](http://bookshelfjs.org/#associations) but not shown in the Knex docs. It needs to be documented that this can be used instead of saying the slightly longer `references(<column>).inTable(<table>)`

lourd:
- There was a closing `</column>` instead of `</code>`


elhigu:
Looks good thanks!

PR #1163, opened 2016-01-28
Feature/add new event query error

wubzz:
I made these changes back in 2014 (like v0.7?) but it was never merged, unsure why. However, tg and ben both seemed ok with it at the time, so I'm giving it another go. References: #594 #593 

Use case example:
```javascript
knex.on('query-error', function(error, obj) {
appLogger.log(error);
});
```


elhigu:
Looks good, could you add also documentation to index.html for this feature?

wubzz:
@elhigu Certainly. Is that text sufficient enough? Kinda bad with docs..

Also, I changed the text for the "query" error from `Useful for logging all` -> `Useful for logging all queries throughout your application.` Noticed it spontaneously, and it seemed odd.

index.html
```diff
@@ -2428,10 +2429,33 @@ <h3 id="Interfaces-Events">Events</h3>
 <code class="js">knex.select('*')
   .from('users')
   .on('query', function(data) {
-    app.log(data);
+  app.log(data);
```
elhigu: Does example still look correct in browser? looks like indentation was removed here by an accident.

index.html
```diff
@@ -2428,10 +2429,33 @@ <h3 id="Interfaces-Events">Events</h3>
...
   })
   .then(function() {
-    // ...
+  // ...
```
elhigu: and here too

index.html
```diff
@@ -2428,10 +2429,33 @@ <h3 id="Interfaces-Events">Events</h3>
...
+
+
+<pre>
+<code class="js">knex.select(['NonExistentColumn'])
+  .from('users')
+  .on('query-error', function(error, obj) {
+    app.log(error);
+  })
+  .then(function() {
+  // ...
```
elhigu: indentation

index.html
```diff
@@ -2428,10 +2429,33 @@ <h3 id="Interfaces-Events">Events</h3>
...
+  })
+  .catch(function(error) {
+  // Same error object as the query-error event provides.
```
elhigu: indentation

src/client.js
```diff
@@ -130,8 +130,9 @@ assign(Client.prototype, {
     debugQuery(obj.sql)
     return this._query.call(this, connection, obj).catch(function(err) {
       err.message = SqlString.format(obj.sql, obj.bindings) + ' - ' + err.message
+      this.emit('query-error', err, obj)
       throw err
-    })
+    }.bind(this))
```
elhigu: with an arrow function bind shouldn't be needed so this would look slightly cleaner

elhigu:
Docs were nice! Looks like there was some unintentional edits too. Thanks a lot for this pq :+1:

wubzz:
Sorry about that, should be fixed now.

marcbachmann:
In the docs, that feature is only mentioned by one short sentence. It's also in the changelog, where I found it. Maybe we can add an example with the global event listener.

wubzz:
@marcbachmann Not sure what you mean by this. Each event has its own pretty extensive description in the docs, including code examples. [Reference](http://knexjs.org/#Events-query-error)

marcbachmann:
`knexInstance.on('query-error')` isn't documented.
The example only shows `queryInstance.on('query-error')`

elhigu:
@marcbachmann open pull request with documentation fixes if you feel like it 👍

Issue #1118, opened 2015-12-29
orWhere with object (2+ properties) as argument


titarenko:
Hi! First of all, thank you for a great library!

I'm wondering why
```js
.where({ field: 'name', value: 'Bob' })
.orWhere({ field: 'ip', value: '127.0.0.1' })
```
generates following SQL
```sql
where ("field" = $1 and "value" = $2 or "field" = $3 or "value" = $4)
```
I was expecting that result should be "and-or-**and**", but not "and-or-**or**".

I found #205, where you are advising to wrap "and" parts into "where(function)", but that construction is less readable (well, not a big issue, but worth to avoid for simplicity), that one might expect.

titarenko:
I can assume this was done for simpler code in cases of "where this or that or that", having ability to specify that in single builder call (passing an object).

But could you please mention that in docs, because currently I cannot find "orWhere" with object (with 2+ properties) as an argument on http://knexjs.org.

P. S. Again, thanks for such cool module!

rhys-vdw:
Right, so the expected output should be:

```js
where "field" = $1 and "value" = $2 or ("field" = $3 and "value" = $4)
```

Right?

titarenko:
Yes, at least that is what I was expecting intuitively (if I pass field-value pairs in a *single object*, I intuitively assume they represent *single entity* and *tied very closely together*, so I'm expecting "and" between them). 

But again, as I mentioned in the second message, after a bit deeper thinking I realised, that "or" in this case is also reasonable, since there should be builder method which allows you to write
`.orWhere({ a: 'b', c: 'd' })` instead of `orWhere('a', 'b').orWhere('b', 'c')` to achieve `"a" = $1 or "b" = $2`. 

So essentially, as for me, it's fine to leave implementation as it is, but I think there should be an example dedicated to this case on http://knexjs.org page.

wubzz:
+1, this issue got me good several times. Reading the code `.orWhere({...})` certainly makes you think it's an *and* when in reality it generates *or*.

rhys-vdw:
Yep. Not sure why I had this marked as a discussion. Definitely need a PR to fix this broken behaviour.

wubzz:
This change will generate the query @rhys-vdw wrote above:

```javascript
// Adds an `or where` clause to the query.
  orWhere: function orWhere() {
    this._bool('or');
    var obj = arguments[0];
    if(_.isObject(obj)) {
      return this.whereWrapped(function() {
[...]

PR #1164, opened 2016-01-28
Bugfix/or where with object should be and

wubzz:
This is a PR for #1118. There may be better ways of solving this..

Previously, `orWhere({...})` would generate *or-or-or*, it will now generate a wrapped *and-and-and* instead.


rhys-vdw:
@wubzz This looks good to me. I think it might deserve an example in the docs, if you don't mind.

wubzz:
@rhys-vdw I agree. Added example code with a description. Hope the description will suffice.

rhys-vdw:
Thanks dude. :+1:

PR #1168, opened 2016-01-29
Proxy query-error event in transaction

wubzz:
@elhigu In regards to #1163 that was merged yesterday, I forgot to take this into consideration. I fixed the same bug for the normal 'query' event back in Sept 2015, so that reminded me..


wubzz:
No idea what happened during those CI builds, but surely wasn't caused by this change.

PR #1169, opened 2016-01-29
Debug with raw shows no info

### LINKED ISSUE #1088: `knex.raw()` doesn't print queries when debugging
If I use a `knex.raw()` statement with a config like: `require(knex)({ client: 'pg', debug: true })`, then nothing shows up. But if I use the query builder, it shows up. Am I doing something wrong?

```js

var knex = require('knex')({ client: 'pg', debug: true });

return knex.raw('SELECT id FROM users');
```

Debugging doesn't print anything. However, if I use the builder..

```js

var knex = require('knex')({ client: 'pg', debug: true });

return knex('users').select('id');
```

I get the expected output in the console.

wubzz:
This fixes #1088 where `.raw` queries are not being logged when `debug` is set to true in the options/config.

I realize there's a deprecation TODO right above this change, but until said deprecation is in place, I reckon it should still be fixed.


blah238:
Thanks @wubzz!

Issue #1000, opened 2015-09-29
Add .filter() method for resultsets.


VynceMontgomery:
Just like .map and .reduce, please add a .filter that works just like -- because it is -- bluebird's .filter().

rhys-vdw:
Hi @VynceMontgomery. I'll accept a PR for this one.

PR #1170, opened 2016-01-30
Add .filter() to resultsets

wubzz:
This is for issue #1000 .

Example:

```javascript
knex('test')
.select()
.filter(function(item) {
	return item.test > 10;
})
.then(function(items) {
})
.catch(function(error) {
});
```

Not entirely sure on the use cases of this since there's `.where` and whatnot, but who am I to judge. :)

### DIFF
--- a/index.html
+++ b/index.html
@@ -224,6 +224,7 @@
       <li>&nbsp;&nbsp;– <a href="#Promises-reduce">reduce</a></li>
       <li>&nbsp;&nbsp;– <a href="#Promises-bind">bind</a></li>
       <li>&nbsp;&nbsp;– <a href="#Promises-return">return</a></li>
+      <li>&nbsp;&nbsp;– <a href="#Promises-filter">filter</a></li>
       <li><b><a href="#Interfaces-Callbacks">Callbacks</a></b></li>
       <li>&nbsp;&nbsp;– <a href="#Interfaces-asCallback">asCallback</a></li>
       <li><b><a href="#Interfaces-Streams">Streams</a></b></li>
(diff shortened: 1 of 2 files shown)


wubzz:
Would it also make sense to append `.first()`, `.last()` and `.chunk()` to the interface for convenience? Should be something simple as

```javascript
var Promise = require('bluebird');
 _.each(['first', 'last'], function(method) {
    Target.prototype[method] = function() {
      var then = this.then();

      return new Promise((resolver) => {
        then.then((items) => {
          resolver(_[method](items));
        });
      });
    };
  });
```

IE

```javascript
knex('test')
.select()
.filter(function(item) {
    return item.test > 10;
})
.first()
.then(function(item) {
})
.catch(function(error) {
});
```

rhys-vdw:
Apparently I had marked #1000 as "PR Please" but have since changed my mind on the matter. Sorry @wubzz!

I think maybe hold off on aliasing all the Bluebird shorthands. It might be preferable to completely rip out Bluebird and allow the client to pass in their own standard compliant `Promise` object. This way native promises could be supported, but Bluebird can be used if preferred.

I'd like to have a discussion around this before we add more promise related methods to the API.

More specifically I think that as a SQL library we ought to encourage correct usage and discourage "incorrect" usage. As you said, there is typically no real reason to use `.filter` when `.where` is available.

> Would it also make sense to append .first(), .last() and .chunk() to the interface for convenience?

@wubzz We already have [`first()`](http://knexjs.org/#Builder-first), but it does not return a Promise.

I'm a bit wary about methods that mean doing unnecessary processing (eg. fetching 1000 rows and discarding 999 of them).

wubzz:
@rhys-vdw Yeah, that makes sense to me. I did fail to see a use case for this as stated above. I'm gonna close this one.

Issue #965, opened 2015-09-04
knex.raw arguments in 0.8.6


wubzz:
I'm currently on hold with upgrading from 0.7.6 to 0.8.6 due to the knex events not being fixed in 0.8.6 yet, but I figured I could upgrade to test if theres anything else in my app that would need fixing.

Doing this, I found that the argument handling for `knex.raw` must have changed, and I'm just unsure whether this was intentional or not. (Nothing in changelog)

**0.8.6**
```javascript
knex.raw('SELECT schema_name FROM information_schema.schemata WHERE schema_name = ?', 'Test').toString();
//SELECT schema_name FROM information_schema.schemata WHERE schema_name = 'T'
//Also throws database error: 08P01 - bind message supplies 4 parameters, but prepared statement "" requires 1
```

**0.7.6**
```javascript
knex.raw('SELECT schema_name FROM information_schema.schemata WHERE schema_name = ?', 'Test').toString();
//SELECT schema_name FROM information_schema.schemata WHERE schema_name = 'Test'
```

Was this intentional?

wubzz:
I should add that sending `['Test']` in 0.8.6 does work, but not if you send the string alone. Instead it appears to run `split` on the string..

rhys-vdw:
Just a guess, I think it is possible to bind an array as a value. This
would cause ambiguity between applying an array as a binding and an array
of bindings.

If this is the case then it should error rather than splitting the string.

wubzz:
@rhys-vdw I agree it should throw an error to easily detect broken parts of the apps.

wubzz:
@rhys-vdw Would you agree that allowing either String or Array is the better option here? Much like nearly every other function in exposed in knex, such as the tablebuilder. (primary('id'), primary(['id'])

That way knex can be backwards-compatible in regards to this issue. It's a quick fix changing the argument to an Array if the input is a String. See: https://github.com/tgriesser/knex/blob/master/src/raw.js#L29

elhigu:
Fixed in #1172

PR #1172, opened 2016-01-30
Make raw bindings backwards compatible

wubzz:
I explained this in #965 where the handling of bindings for `.raw` was changed in v0.8+ disallowing the use of for example:

```javascript
knex.raw('username = ?', 'Test');
//generates username = 'T' with 4 bindings ['T', 'e', 's', 't']
```

So in the spirit of not breaking applications upon upgrading knex version, this is my suggested solution. It doesn't make any sense to enforce Array bindings while the rest of the knex interface is very flexible, and it also allows for easier-to-read code.

```javascript
knex.raw('username = ?', 'Test');
//now generates username = 'Test'
```

I added a test case so that this does not break again.

Edit: A good argument to why this should be merged, I am going to quote http://knexjs.org/
>**Upgrading 0.7 -> 0.8**
>
>There should be no major changes breaking the external API. If you encounter issues, please open a ticket.

Yet this broke in 0.8.

### DIFF
--- a/src/raw.js
+++ b/src/raw.js
@@ -27,7 +27,7 @@ assign(Raw.prototype, {
   set: function(sql, bindings) {    
     this._cached  = undefined
     this.sql      = sql
-    this.bindings = _.isObject(bindings) || _.isUndefined(bindings) ?  bindings : [bindings]
+    this.bindings = (_.isObject(bindings) || _.isUndefined(bindings)) ?  bindings : [bindings]
 
     return this
   },
--- a/src/raw.js
+++ b/src/raw.js
@@ -6,6 +6,7 @@ var EventEmitter  = require('events').EventEmitter
 var assign        = require('lodash/object/assign')
 var reduce        = require('lodash/collection/reduce')
 var isPlainObject = require('lodash/lang/isPlainObject')
+var _             = require('lodash');
 
 function Raw(client) {
   this.client   = client
@@ -26,7 +27,8 @@ assign(Raw.prototype, {
   set: function(sql, bindings) {    
     this._cached  = undefined
     this.sql      = sql
-    this.bindings = bindings
+    this.bindings = _.isObject(bindings) || _.isUndefined(bindings) ?  bindings : [bindings]
+
     return this
   },
 
--- a/index.html
+++ b/index.html
@@ -2134,6 +2134,9 @@ knex('users')
   .orWhere(
     knex.raw('accesslevel = ?', 1)
   )
+  .orWhere(
+    knex.raw('updtime = ?', new Date())
+  )
 </pre>
 
     <p>
--- a/index.html
+++ b/index.html
@@ -2124,6 +2124,18 @@ knex('users')
   }))
 </pre>
 
+    <p>For simpler queries where one only has a single binding, <tt>.raw</tt> can accept said binding as its second parameter.</p>
+
+<pre class="display">
+knex('users')
+  .where(
+    knex.raw('LOWER("login") = ?', 'knex')
+  )
+  .orWhere(
+    knex.raw('accesslevel = ?', 1)
+  )
+</pre>
+
     <p>
       To prevent replacement of <tt>?</tt> one can use escape sequence <tt>\\?</tt>.
     </p>
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -3109,4 +3109,33 @@ describe("QueryBuilder", function() {
       }
     });
   });
+
+  it('#965 - .raw accepts Array and Non-Array bindings', function() {
+    var expected = function(expectedBindings) {
+      return {
+        mysql:   {
+          sql:      'select * from `users` where username = ?',
+          bindings: expectedBindings
+        },
+        mssql:   {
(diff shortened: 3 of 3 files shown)


wubzz:
@elhigu Ima ping you in on this one as well if that's ok. In my opinion this is a crucial fix before next release.

src/raw.js
```diff
@@ -26,7 +27,8 @@ assign(Raw.prototype, {
   set: function(sql, bindings) {    
     this._cached  = undefined
     this.sql      = sql
-    this.bindings = bindings
+    this.bindings = _.isObject(bindings) || _.isUndefined(bindings) ?  bindings : [bindings]
```
elhigu: I would like to add parenthesis here for clarity... 

I was reading the line first the way that in case when _.isObject is true, true would be assigned to `this.bindings` since part after `||` won't need to be ran :) looks like the ternary operator makes the difference here...

elhigu:
@wubzz for the arguments why this should be merged... we really can break backwards compatibility on every 0.x change... and actually we are breaking it all the time. And the mention in changes between 0.7 -> 0.8 Tim was saying that most of the APIs should stay the same... this one might be intentional or not.

That being said I'm all in to support those shortcuts being able to pass non array / object value to raw to prevent boilerplate. 

However it needs to be added to documentation exactly what parameters `knex.raw` may take and how they are interpret. Specially the case how to give an array to be bound (weird [[]]) syntax or actually I would prefer named bindings for that `raw(":theArray", { theArray: [] })`.

@rhys-vdw If it is ok to you I would like to merge this one after some minor changes.

wubzz:
From my experience breaking backwards compatibility should only be done when there's a good reason to do it. Something like this definitely isn't a good enough cause, even less so since it wasn't in the changelog. I understand it may not have been intentional, and I'm not blaming anyone that it broke, my clarity is just that it gets fixed. With a test case in place, at least history won't repeat itself.

I'll try to update the documentation as best I can.

src/raw.js
wubzz: My bad, I do agree.

elhigu:
> From my experience breaking backwards compatibility should only be done when there's a good reason to do it.

Agree :+1:

elhigu:
Btw. code looks good to me, so if @rhys-vdw says its fine I'll merge :+1:

rhys-vdw:
Can we get a test case for this?

```js
knex.raw('created_at > ?', new Date());
```

I'm still a bit wary of this change in general. The `[[]]` thing is pretty dirty. There's still instruction for that in the docs. Also, if we're accepting a single argument, wouldn't it follow to support `raw(query, ...args)`?

elhigu:
@rhys-vdw the `[[]]` case is already there and it is pretty dirty, not sure what could be done about it except avoid doing it :) 

How system should behave in case of `knex.raw('created_at > ?', new Date());` should it fail and tell user to use `[new Date()]` or should it work and output 

```
> knex.raw('wat < ?', [new Date()]).toString()
'wat < \'2016-02-04T09:21:00.091+02:00\''
```

Or should we allow using it as an object as it currently works enabling this kind of insanity
```
a = new Date()
Thu Feb 04 2016 09:26:55 GMT+0200 (EET)
> a.foo = 'OMG'
'OMG'
> knex.raw('wat < :foo', a).toString()
'wat < \'OMG\''

```

I have nothing against of supporting also `raw(query, ...args)` syntax, but I don't see it as a must for this pull request.

In the end to me any of the choices how to handle special objects e.g. `Date` is fine. It all is extra functionality on top of knex 0.9... 

Knex already has code, that treats Date objects specially, so if it is documented that one can do `knex.raw('?', new Date())` without an array syntax it sounds good to me.

wubzz:
Currently using `new Date()` does not call .tostring() as far as I could tell. The generated string is built out of the Date object, see https://github.com/tgriesser/knex/blob/master/src/query/string.js#L20 for reference.

Example output
```javascript
... where date = '2016-02-04 17:03:01.599'
```

I added this to the test.

rhys-vdw:
> the [[]] case is already there and it is pretty dirty

Yeah, but before this PR it's consistent. Every value must be provided in an array. No surprises. I actually think the correct API here is to either support array only or varargs only. That way we eliminate the inconsistent case.

Since this is a PR to restore back compatability I'm happy to go with this for now. We still need an instruction in the docs to warn about passing an array before this can be merged.

rhys-vdw:
Merged. Cheers @wubzz.

wubzz:
@rhys-vdw Perhaps completely unrelated, but any thoughts on the arrival of 0.9.1 ? Nearly 150 commits since 0.9.0.

Currently we're still using 0.7.6 at work due to a few bugs that afaik are fixed in 0.9.1, primarily the event related bugs like proxying `query` from a transaction object onto the main client. (Used for logging and such)

rhys-vdw:
@wubzz Yes. I was going to do it earlier today but got stuck trying to fix up #1180. Hoping to have it done by Monday. I think it will be 0.10.0 though due to breaking changes.

wubzz:
@rhys-vdw Sounds good, cheers. :+1:

PR #1174, opened 2016-02-01
Removes object keys with value undefined from being added to DB in update.

elhigu:
Clean version of pq #962 with documentation. Fixes issue #961.


rhys-vdw:
Awesome. :+1:

PR #1175, opened 2016-02-01
Prevent attempt to create migrations table again if it already exists

ArsalanDotMe:
There is an issue where if you are creating migrations table inside a schema, knex tries to create the table every time even though the table is already created. The real issue is that knex migrations do not handle custom schemas but this is a quick fix to handle those until knex rolls out better support for migrations in multiple schemas.


elhigu:
Seems legit +1

PR #1176, opened 2016-02-01
Support connections to EnterpriseDB Postgres databases.

al3x:
[EnterpriseDB](http://www.enterprisedb.com/) is a distribution of Postgres. Knex should be compatible with EnterpriseDB as-is, but because it's looking for the string `Postgres` in the result of `SELECT version();`, connections to EnterpriseDB databases currently fail.

### DIFF
--- a/src/formatter.js
+++ b/src/formatter.js
@@ -86,6 +86,10 @@ assign(Formatter.prototype, {
     return this._wrapString(value + '');
   },
 
+  wrapAsIdentifier: function wrapAsIdentifier(value) {
+    return this.client.wrapIdentifier((value || '').trim());
+  },
+
   alias: function(first, second) {
     return first + ' as ' + second;
   },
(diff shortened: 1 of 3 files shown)


elhigu:
@al3x Looks like tests are failing any ideas why?

al3x:
@elhigu Not sure. Most Postgres tests are passing?

src/formatter.js
```diff
@@ -86,6 +86,10 @@ assign(Formatter.prototype, {
     return this._wrapString(value + '');
   },
 
+  wrapAsIdentifier: function wrapAsIdentifier(value) {
+    return this.client.wrapIdentifier((value || '').trim());
+  },
+
```
elhigu: Why these changes are here?

elhigu:
Implemented in #1993

Issue #1171, opened 2016-01-30
Knex object hangs program inside transactions


LyonesGamer:
In the case of a code sample like this:

    const knex = require('knex')({...});

    knex.transaction(function(trx) {
        knex('table).select('id').first().then(function(id) {
            trx.update({'foo': 4}).where({'id': id}).then(function() {
            }
        }
    }

The call to knex() causes the application to hang indefinitely. While I can understand wanting the user to use trx inside a transaction, it would be better for it to return an error than hang.

elhigu:
It shouldn't hang if you have DB connections available to do other request outside of the transaction inside a transaction... This might be related to this issue https://github.com/tgriesser/knex/issues/1040

elhigu:
How is your connection pool settings? Could you try to make pool bigger to test if it still hangs to verify if the error is because of not having enough available connections to complete transaction?

LyonesGamer:
I don't define connection pool settings, but that did encourage me to look at those settings, and noticed this:

"This connection pool has a default setting of a min: 2, max: 10 for the MySQL and PG libraries, and a single connection for sqlite3 (due to issues with utilizing multiple connections on a single file)."

I am using SQLite, so this would appear to be the issue.

elhigu:
@LyonesGamer next release should throw an error after certain timeout if there are no available connections

Issue #1040, opened 2015-10-30
More Transactions than max pool connections kills knex


ArdentZeal:
Hello,

when starting more transactions then there are connections allowed in the pool, knexjs hangs indefinitly and does not throw an error. I would expect knexjs to queue all transactions, then start one after another as soon as a connection frees up.

To get a better understanding:

* Are transactions always executed on 1 connection? 
* When I start a transaction and do some read calls without the transacting(trx) part, because I know the data was there before the transaction started anyway --> more than 1 connection is used? Or do these read calls also use the transaction connection?

Kind Regards,
ArdentZeal

vschoettke:
Every knex SQL command that uses `.transacting(trx)` or uses `trx` instead of `knex` (where trx is the callback parameter of `.transaction()`) is executed within the same connection.

If you don't specify/use the transaction (trx) object the command will run outside of the transaction.

ArdentZeal:
> If you don't specify/use the transaction (trx) object the command will run outside of the transaction.

I know the theory that far, I am interested specifically in the connection part of the question. So if a query runs outside the transaction, it **has to use** a different connection and can not be executed on the transaction connection?

### I am assuming following scenario might be happening

I have 5 transactions running simultaneously with a connection pool of 5 and there are some calls in my logic which do not use .transacting(trx). These calls would need at least one seperate connection to do their stuff, which is not possible because all of the available connections are blocked by the running transactions. ==> Deadlocked

is this possible to happen or can knex use the transaction connections to finish up the calls which are not .transacting(trx).

Kind Regards,
ArdentZeal

vschoettke:
Knex will not reuse connection with a transaction for other SQL commands outside of that transaction (and it should not because doing so may change the query result).

Regarding your example, the deadlock will be resolved by the `acquireTimeout` from the pool2 so the pending queries will fail after the specified timeout.

I highly recommend using the `trx` object as `knex` replacement since `.transacting(trx)` is easily missed. See http://knexjs.org/#Transactions for an example.

ArdentZeal:
> Regarding your example, the deadlock will be resolved by the acquireTimeout from the pool2 so the pending queries will fail after the specified timeout.

Thanks for all the help so far, but there is no timeout happening. Is the default timeout no timeout? 

Kind Regards,
ArdentZeal

elhigu:
I think what you are doing there is that you start 5 transactions:
```js
[...]

PR #1177, opened 2016-02-01
Bugfix/knex hangs when pool is filled with trs

wubzz:
@tgriesser @elhigu @rhys-vdw In regards to #1040 and #1171 , this change will throw a Timeout error including information about the query that was run, when acquiring a connection is not possible or simply takes too long.

I am in no way implying that this is the correct way of fixing this, but it does work. The issues mentioned it needs to be configurable, so a new option has been added to the knex options object. (not the 'Pool' object since technically it has no relevance there)

Not so sure the hacky stuff done in the test file is allowed, but couldn't figure out a better way of doing it.. :)

### DIFF
--- a/test/integration/builder/transaction.js
+++ b/test/integration/builder/transaction.js
@@ -294,7 +294,7 @@ module.exports = function(knex) {
         trx.raw('SELECT 1 = 1').then(function() {
 
           //No connection is available, so try issuing a query without transaction.
-          //Since there is no available connection, it should throw a timeout error based on knex pool config.
+          //Since there is no available connection, it should throw a timeout error based on `aquireConnectionTimeout` from the knex config.
           knexDb.raw('select * FROM accounts WHERE username = ?', ['Test'])
               .then(function(res) {
                 expect(res).to.not.exist();
--- a/test/integration/builder/transaction.js
+++ b/test/integration/builder/transaction.js
@@ -291,7 +291,7 @@ module.exports = function(knex) {
 
       //Create a transaction that will occupy the only available connection, and avoid trx.commit.
      knexDb.transaction(function(trx) {
-        trx.raw('SELECT version()').then(function() {
+        trx.raw('SELECT 1 = 1').then(function() {
 
           //No connection is available, so try issuing a query without transaction.
           //Since there is no available connection, it should throw a timeout error based on knex pool config.
--- a/test/integration/builder/transaction.js
+++ b/test/integration/builder/transaction.js
@@ -3,6 +3,8 @@
 'use strict';
 
 var Promise = testPromise;
+var Knex   = require('../../../knex');
+var _ = require('lodash');
 
 module.exports = function(knex) {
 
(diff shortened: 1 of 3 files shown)


elhigu:
@wubzz Awesome! I'll check this later on today.

test/integration/builder/transaction.js
```diff
@@ -277,6 +279,41 @@ module.exports = function(knex) {
...
+              .catch(function(error) {
+                expect(error.bindings).to.be.an('array');
+                expect(error.bindings[0]).to.equal('Test');
+                expect(error.sql).to.equal('select * FROM accounts WHERE username = ?');
+                expect(error.message).to.equal('Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?');
+                trx.commit();
+              });
+        }).catch(trx.rollback);
+      })
+      .then(function() {
```
elhigu: IIRC you could remove `done()` calls and just return a promise from test case to make code slightly more robust.

test/integration/builder/transaction.js
```diff
@@ -277,6 +279,41 @@ module.exports = function(knex) {
 
     });
 
+    it('#1040, #1171 - When pool is filled with transaction connections, Non-transaction queries should not hang the application, but instead throw a timeout error', function(done) {
+      this.timeout(15000);
+      //To make this test easier, I'm changing the pool settings to max 1.
+      var knexConfig = _.clone(knex.client.config);
+      knexConfig.pool.min = 0;
+      knexConfig.pool.max = 1;
+      knexConfig.acquireConnectionTimeout = 10000;
```
elhigu: Does this test take now 10 seconds to run? If so could we set timeout to < 1 sec to prevent test from taking so long?

test/integration/builder/transaction.js
```diff
@@ -277,6 +279,41 @@ module.exports = function(knex) {
...
+
+      //Create a transaction that will occupy the only available connection, and avoid trx.commit.
+     knexDb.transaction(function(trx) {
+        trx.raw('SELECT 1 = 1').then(function() {
+
+          //No connection is available, so try issuing a query without transaction.
+          //Since there is no available connection, it should throw a timeout error based on `aquireConnectionTimeout` from the knex config.
+          knexDb.raw('select * FROM accounts WHERE username = ?', ['Test'])
+              .then(function(res) {
+                expect(res).to.not.exist();
```
elhigu: Is intention of this to always fail if this then block is executed? In that case `expect(false).to.be.ok()` would be more readable to me, since it fail independently from `res`.

test/integration/builder/transaction.js
```diff
@@ -277,6 +279,41 @@ module.exports = function(knex) {
...
+              })
+              .catch(function(error) {
+                expect(error.bindings).to.be.an('array');
+                expect(error.bindings[0]).to.equal('Test');
+                expect(error.sql).to.equal('select * FROM accounts WHERE username = ?');
+                expect(error.message).to.equal('Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?');
+                trx.commit();
```
elhigu: If you return promise from transaction, calling trx.commit / trx.rollback wouldn't be necessary. Not sure why you commit transaction on error though :)

wubzz: I called commit to release the connection and enter the transaction's callback which in turn called done(), but as you mentioned simply returning knexDb.transaction is easier than using 'done'. Switching to that! :+1:

elhigu:
Generally code looks really good :) There were few things I didn't understand completely. Only serious issue of those was the 10 second timeout during testrun.

wubzz:
Yeah, not sure what I was thinking with the 10secs.. Changed it to 1sec. Simplified the test code as well.

elhigu:
@wubzz thanks a lot, this was really important feature :+1: most of the time people had no idea why their queries just hang.

PR #1181, opened 2016-02-02
fixes bug when using dot in as identifier

a0viedo:
I've encountered what I think is a bug on the formatting of SQL statements. Take the following example:
```js
var knex = require('knex')(config);
console.log(knex.select('id as foo.bar').from('users').toSQL());
```
This would print
```
{ method: 'select',
  options: {},
  bindings: [],
  sql: 'select "id" as "foo"."bar" from "users"' }
```
like trying to access a field in the alias but instead should be formatted as
```
{ method: 'select',
  options: {},
  bindings: [],
  sql: 'select "id" as "foo.bar" from "users"' }
```

### DIFF
--- a/src/formatter.js
+++ b/src/formatter.js
@@ -86,6 +86,10 @@ assign(Formatter.prototype, {
     return this._wrapString(value + '');
   },
 
+  wrapAsIdentifier: function wrapAsIdentifier(value) {
+    return this.client.wrapIdentifier((value || '').trim());
+  },
+
   alias: function(first, second) {
     return first + ' as ' + second;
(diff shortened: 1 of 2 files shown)


elhigu:
As far as I know, the correct format is `"foo"."bar"`. @a0viedo  why do you think it is wrong?

elhigu:
Looks like test error was just a timeout, I retriggered the build.

elhigu:
@a0viedo good catch and thanks for the fix!

a0viedo:
@elhigu thank you!

Issue #757, opened 2015-03-31
batch insert breaks after 5k lines with ca 65 columns


ArdentZeal:
Hello there,

I tried to write an Excel import for my application. After importing smaller excelfiles (1000ish lines), I tried a big one with ca 20000. The excel import stopped working.

I debugged it down to the parameterized queries, which seem to not be able to handle all the different parameters and is ULTRA slow at that point too.

To workaround that, I used the toString() method and applied the query then with knex.raw.
The problem with this approach is that the 5k line limit is still valid. Even the toString()-Method seems to be at some kind of end with too much data.

So my second workaroung is to bundle my data in smaller packets (1000) and to toString() that (to be fast again), which works but is really really ugly.

```
 var sqlquery = '';

                var batchSize = 1000;
                var newListEntriesLength = newListEntries.length;
                // build sql string in batchSize packs
                for(var i = 0; (i * batchSize) < newListEntriesLength; i++) {
                    var slicedNewEntries = newListEntries.splice(0, batchSize);

                    if(slicedNewEntries.length > 0) {
                        sqlquery += _strucd.knex(tableName).insert(slicedNewEntries).toString() + '; ';
                    } 
                }

                // Add leftovers
                if(newListEntries.length > 0) {
                    sqlquery += _strucd.knex(tableName).insert(newListEntries).toString() + ';';
                }
                _strucd.knex.raw(sqlquery)
                    .then(function(results) {
                        return callback(null, results);
                    }, function(err) {
                        return callback(err);
                    });
```

1. Am I missing something? 
2. Is it really that complicated to do fast batch insert with several 1000 lines?
3. Is there a better way to build an insert query, without doing everything it manually?

Kind Regards,
Matthias

tgriesser:
I'm going to dig into this a bit and see if I can improve things to make this possible

ArdentZeal:
Ah Im using Postgres underneath, if that matters at all.

tgriesser:
At the very least I'll make it so there's a (configurable) batch limit so the ugly code is pushed into the library and not on you to produce

catmsred:
Just checking in to see if there's been any work on this?  I'm also working with big imports (~16000 lines) on MariaDB, and I'm having similar problem.  If I let the query run long enough, then I eventually get the following error:

```
[project]/node_modules/mariasql/lib/Client.js:367
      this._client.query(this._curResults._query, this._curResults._useArray);
                   ^
Error: Not ready to query
    at Error (native)
[...]

PR #1182, opened 2016-02-02
knex.batchInsert

wubzz:
This was discussed in #757. Couldn't see Tim's code in that issue so not sure how he originally implemented this, but hopefully I'm not too far of.

Ideas for performance improvements is more than welcome. It still takes a bit to insert 5k+ rows.

In short it will create a transaction and insert all the rows within that transaction.

```javascript
knex.batchInsert(tableName, rows, chunkSize = 1)
```

### DIFF
--- a/src/util/make-knex.js
+++ b/src/util/make-knex.js
@@ -37,8 +37,7 @@ module.exports = function makeKnex(client) {
     batchInsert: function(table, batch, chunkSize) {
       chunkSize = _.isNumber(chunkSize) ? chunkSize : 1;
 
-      return new Promise((resolver, rejecter) => {
-        this.transaction((tr) => {
+      return this.transaction((tr) => {
 
           //Avoid unnecessary call
           if(chunkSize !== 1) {
@@ -50,11 +49,7 @@ module.exports = function makeKnex(client) {
           }))
           .then(tr.commit)
           .catch(tr.rollback)
-
         })
-        .then(resolver)
-        .catch(rejecter)
-      })
     },
 
     // Runs a new transaction, taking a container and returning a promise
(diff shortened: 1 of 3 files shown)


test/integration/builder/inserts.js
```diff
@@ -647,6 +647,37 @@ module.exports = function(knex) {
       });
     });
 
+
+    it('#757 - knex.batchInsert(tableName, bulk, chunkSize)', function (done) {
+      this.timeout(20000);
```
elhigu: Just to make sure that batch insert does what it is supposed to dataset could be e.g. 100 items, which are added in batch of 30 items or so? We need to avoid long running tests.

test/integration/builder/inserts.js
```diff
@@ -647,6 +647,37 @@ module.exports = function(knex) {
...
+              var item = {};
+              for(var x = 0; x < 30; x++) {
+                item['Col' + x] = fiftyLengthString;
+              }
+              items.push(item);
+            }
+
+            return knex.batchInsert('BatchInsert', items, 1);
+          })
+          .then(function() {
```
elhigu: returning promise is a bit nicer here

src/util/make-knex.js
```diff
@@ -32,6 +34,24 @@ module.exports = function makeKnex(client) {
...
+      chunkSize = _.isNumber(chunkSize) ? chunkSize : 1;
+
+      return this.transaction((tr) => {
+
+          //Avoid unnecessary call
+          if(chunkSize !== 1) {
+            batch = _.chunk(batch, chunkSize)
+          }
+
+          Promise.all(batch.map((items) => {
```
elhigu: returning promise from transaction should call trx.commit / trx.rollback automatically if I recall correctly

elhigu:
@wubzz thaks for PR, I left few comments on the code. Also documentation should be added for the new feature. The long running test was the only serious issue in the code. Other stuff are more just opinions :)

wubzz:
@elhigu Thanks for the feedback. I've updated the test and added info to the docs. There was no obvious section in the docs to place this in so I added a 'Utility' section. Let me know if I need to change it! :)

Issue #614, opened 2015-01-02
Where with subquery


ekulabuhov:
Hi!

In your examples you have:
`knex('accounts').where('id', 'in', subquery)`

If you replace 'in' with any other operator (i.e. >, <, =, etc.) the subquery won't be wrapped in parenthesis anymore. Testing this with PG produces invalid query.

Thanks!

sandro-csimas:
Here is how solve:
````
return UserEvent.collection().query(function(qb) {
  qb.where('circle_id', circle.id)
  qb.whereIn('user_id', entities.getIds(users))
  qb.whereIn('type', ['user_leaves', 'user_arrives'])
  qb.where('id', function() {
    this.select('last_user_event.id')
    .from('user_event as last_user_event')
    .whereRaw('last_user_event.user_id = user_event.user_id')
    .orderBy('last_user_event.occurrence_date', 'desc')
    .limit(1);
  });
})
````
Will be better if knex support subqueries as you said for the other operators.

elhigu:
This seems to work in 0.9

PR #1184, opened 2016-02-04
Allows builder instances to be used as scalar values

particlebanana:
I'm not sure if it was intentional or not to only allow subqueries using a builder instance in `whereIn` statements but this fix makes it so that they can be used elsewhere. It's quite handy to build instances programmatically rather than have to build up the function in the grouped chain.

I'm not super familiar with the codebase so if this breaks other stuff let me know. Should work similar to the way the `whereIn` compiler works though.

cc #614

### DIFF
--- a/src/formatter.js
+++ b/src/formatter.js
@@ -80,6 +80,9 @@ assign(Formatter.prototype, {
     if (typeof value === 'function') {
       return this.outputQuery(this.compileCallback(value), true);
     }
+    if (value instanceof QueryBuilder) {
+      return this.parameterize(value);
+    }
     raw = this.unwrapRaw(value);
     if (raw) return raw;
     if (typeof value === 'number') return value;
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -157,6 +157,15 @@ describe("QueryBuilder", function() {
     });
   });
 
+  it("allows for subqueries to be used inside select", function() {
+    var subquery = qb().select('total').from('pets').as('pet_total');
+    testsql(qb().select(['name', subquery]).from('public.users'), {
+      mysql: 'select `name`, (select `total` from `pets`) as `pet_total` from `public`.`users`',
+      mssql: 'select [name], (select [total] from [pets]) as [pet_total] from [public].[users]',
+      default: 'select "name", (select "total" from "pets") as "pet_total" from "public"."users"'
+    });
+  });
+
   it("basic table wrapping", function() {
     testsql(qb().select('*').from('public.users'), {
       mysql: 'select * from `public`.`users`',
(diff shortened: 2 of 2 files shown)


src/formatter.js
```diff
@@ -80,6 +80,9 @@ assign(Formatter.prototype, {
     if (typeof value === 'function') {
       return this.outputQuery(this.compileCallback(value), true);
     }
+    if (value instanceof QueryBuilder) {
+      return this.parameterize(value);
+    }
```
elhigu: `raw = this.unwrapRaw(value);` in the next line already runs converting QueryBuilder instance to query...

elhigu:
Hi...  @particlebanana What am I missing? Doesn't this already work exactly the way those test cases are testing? I mean I just tried this out on knex 0.9 without this with following

```js
k = require('knex')({client: 'pg'})
sq = k.select('col').from('table');
sqAs = k.select('col').from('table').as('asSomething');
k('table').select('someCol', sqAs).where('total', sq).toString()
// select "someCol", (select "col" from "table") as "asSomething" from "table" where "total" = (select "col" from "table")
```

particlebanana:
Yup @elhigu terribly sorry about that. Looks like I had tests failing for different reasons.

elhigu:
@particlebanana no worries :)

particlebanana:
Just for anyone coming here by search I tracked down the issue. If you don't provide an `as` to the query builder instance used in the subquery, the result will not be wrapped unless it's used as a scaler value inside the `where` clause. So the workaround is to always add `.as()`.

This mainly affects the `SELECT` and `FROM` clauses.

So in the example above by @elhigu the following produces an invalid query:

```javascript
var k = require('knex')({client: 'pg'})
var sq = k.select('col').from('table');
var sq2 = k.select('col').from('table');
k('table').select('someCol', sq2).where('total', sq).toString();
// select "someCol", select "col" from "table" from "table" where "total" = (select "col" from "table")
```

```javascript
var k = require('knex')({client: 'pg'})
var sq = k.select('col').from('table');
var sq2 = k.select('col').from('table');
k.from(sq).select('someCol', sq2).where('total', '>', sq).toString();
// select "someCol", select "col" from "table" from select "col" from "table" where "total" > (select "col" from "table")
```

Adding `.as()` fixes this:

```javascript
var k = require('knex')({client: 'pg'})
var sq = k.select('col').from('table').as('foo');
var sq2 = k.select('col').from('table').as('bar');
k.from(sq).select('someCol', sq2).where('total', '>', sq).toString();
// select "someCol", (select "col" from "table") as "bar" from (select "col" from "table") as "foo" where "total" > (select "col" from "table") as "foo"
```

PR #1194, opened 2016-02-10
Default batch size 1000 for batchInsert and validate batch size parameter to be > 0

### LINKED ISSUE #1113: Slow bulk inserts + possible beginnings of a solution
First off, thank you for this amazing library.  I went through about 3 different database libraries for Node and I'm totally sold on KnexJS!

I have toyed around with this for a small while because I was experiencing slow behavior trying to use knex.insert() with a large array in PostgreSQL.  For inserting about 40,000 rows, my program was not running to completion even after waiting for about 10 minutes.  However, I wrote a hacky version of insert that doesn't use PostgreSQL bind variables, and this seems to go a lot faster.  I haven't done a lot more testing or research, but wanted to bring this up in case anyone else had run into this/has more information.  

This is function I wrote that seems to do bulk inserts much faster.  Not suggesting we actually use this (lots of problems with it), but more as a reference: https://gist.github.com/saikat/a5424c23ba29e9e13589

As an anecdotal piece of information, Sequelize seems to be really fast on bulkCreate.  It could be useful to see what they are doing.

elhigu:
Kind of fixes #1113

### DIFF
--- a/src/util/make-knex.js
+++ b/src/util/make-knex.js
@@ -35,7 +35,7 @@ module.exports = function makeKnex(client) {
     },
 
     batchInsert: function(table, batch, chunkSize) {
-      chunkSize = _.isNumber(chunkSize) ? chunkSize : 1;
+      chunkSize = (_.isNumber(chunkSize) && chunkSize > 0) ? chunkSize : 1000;
 
       return this.transaction((tr) => {
 
@@ -44,11 +44,9 @@ module.exports = function makeKnex(client) {
             batch = _.chunk(batch, chunkSize)
           }
 
-          Promise.all(batch.map((items) => {
+          return Promise.all(batch.map((items) => {
             return tr(table).insert(items)
-          }))
-          .then(tr.commit)
-          .catch(tr.rollback)
+          }));
         })
     },
 
--- a/index.html
+++ b/index.html
@@ -2275,7 +2275,7 @@ knex.select('e.lastname', 'e.salary', subcolumn)
 
     <p>The <tt>batchInsert</tt> utility will insert a batch of rows wrapped inside a transaction (which is automatically created), at a given <tt>chunkSize</tt>.</p>
     <p>It's primarily designed to be used when you have thousands of rows to insert into a table.</p>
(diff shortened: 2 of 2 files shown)


wubzz:
The default is also listed in the index.html.

elhigu:
Good catch, updating...

src/util/make-knex.js
```diff
@@ -35,7 +35,7 @@ module.exports = function makeKnex(client) {
     },
 
     batchInsert: function(table, batch, chunkSize) {
-      chunkSize = _.isNumber(chunkSize) ? chunkSize : 1;
+      chunkSize = (_.isNumber(chunkSize) && chunkSize > 0) ? chunkSize : 1000;
```
rhys-vdw: I think this should really throw `TypeError` if a non-number is supplied.

rhys-vdw: Or negative.

elhigu: @rhys-vdw yep, I was thinking about it, but felt lazy to implement it in this pull request. I'll add another for that.

src/util/make-knex.js
```diff
@@ -44,11 +44,9 @@ module.exports = function makeKnex(client) {
             batch = _.chunk(batch, chunkSize)
           }
 
-          Promise.all(batch.map((items) => {
+          return Promise.all(batch.map((items) => {
             return tr(table).insert(items)
-          }))
-          .then(tr.commit)
-          .catch(tr.rollback)
+          }));
```
jshjohnson, line 49: Why was the commit/rollback lines removed here? Since upgrading the version `0.95.11`, all `batchInsert` calls are hanging and I think this may be why

kibertoad: Good catch, I'll take a look.

kibertoad: @jshjohnson Having checked the code and the tests, I believe that code behaves correctly. Autocommit/autorollback happens when no explicit transaction is passed, and when explicit transaction is passed, knex should not be committing automatically, because user might want to perform some additional actions within the same transaction. Hence commit or rollback need to be invoked explicitly on that transaction by the user.

PR #1199, opened 2016-02-13
Fixed regression running migrations through cli

### LINKED ISSUE #1197: migrate fails PostgreSQL 9.5
Ever since I upgraded to PostgreSQL 9.5, whenever I try to migrate, it fails with the following message (added the DEBUG info):

I'm using knex 0.10.0-rc1 CLI. I've tried the exact same migration code on an existing different project that works with a PostgreSQL 9.4 database and it succeeds. But just to check, this is the code used:

```
  return Promise.all([
  	knex.schema.createTable('artists', function (table) {
  		table.increments('ID');
  		table.string('name');
  	}),
  	knex.schema.createTable('albums', function (table) {
  		table.increments('ID');
  		table.string('title').notNullable();
  		table.integer('artist_ID').references('ID').inTable('artists').notNullable();
  		table.string('release_date').notNullable();
  		table.text('description').notNullable();
  		table.string('spotify_link');
  		table.string('playstore_link');
  		table.string('rym_link');
  		table.string('album_art').notNullable();
  	}),
  	knex.schema.createTable('lists', function (table) {
  		table.increments('ID');
  		table.string('title');
  		table.string('category');
  	}),
  	knex.schema.createTable('best_videos_songs', function (table) {
  		table.increments('ID');
  		table.integer('artist_ID').references('ID').inTable('artists').notNullable();
  		table.string('title').notNullable();
  		table.string('rank').notNullable(); // 0 = honorary
  		table.string('youtube_link').notNullable();
  		table.integer('type').notNullable(); // 0 = video, 1 = song
  		table.string('year').notNullable();
  	}),
  	knex.schema.createTable('best_albums', function (table) {
  		table.increments('ID');
  		table.integer('album_ID').references('ID').inTable('albums').notNullable();
  		table.string('rank').notNullable();
  		table.string('year').notNullable();
  	})
  ]).then(function () {
  		return knex.schema.createTable('lists_albums', function (table) {
  			table.integer('list_ID').references('ID').inTable('lists').notNullable();
  			table.integer('album_ID').references('ID').inTable('albums').notNullable();
  			table.primary(['list_ID', 'album_ID']);
  		});
  	});
};

  return knex.schema.dropTable('lists_albums')
  	.then(function () {
  		return Promise.all([
  			knex.schema.dropTable('lists'),
  			knex.schema.dropTable('best_videos_songs'),
  			knex.schema.dropTable('best_albums'),
  			knex.schema.dropTable('albums'),
  			knex.schema.dropTable('artists')
  		]);
  	});
};
```

Could this be an issue on my end, or is this because of PostgreSQL 9.5?

elhigu:
Fixes  #1197

### DIFF
--- a/src/migrate/index.js
+++ b/src/migrate/index.js
@@ -196,7 +196,7 @@ export default class Migrator {
     .then(batchNo => {
       return this._waterfallBatch(batchNo, migrations, direction)
     })
-    .then(() => this._freeLock())
+    .then(retVal => this._freeLock().then(() => retVal))
     .catch(error => {
       var cleanupReady = Promise.resolve();
 
--- a/test/integration/migrate/index.js
+++ b/test/integration/migrate/index.js
@@ -243,7 +243,10 @@ module.exports = function(knex) {
 
     describe('knex.migrate.rollback', function() {
       it('should delete the most recent batch from the migration log', function() {
-        return knex.migrate.rollback({directory: 'test/integration/migrate/test'}).then(function() {
+        return knex.migrate.rollback({directory: 'test/integration/migrate/test'}).spread(function(batchNo, log) {
+          expect(batchNo).to.equal(1);
+          expect(log).to.have.length(2);
+          expect(log[0]).to.contain('test/integration/migrate/test');
           return knex('knex_migrations').select('*').then(function(data) {
             expect(data.length).to.equal(0);
           });


src/migrate/index.js
```diff
@@ -196,7 +196,7 @@ export default class Migrator {
     .then(batchNo => {
       return this._waterfallBatch(batchNo, migrations, direction)
     })
-    .then(() => this._freeLock())
```
dannyfritz: If you're using Bluebird, a passthrough such as `.tap` might be cleaner. Same thing though!
http://bluebirdjs.com/docs/api/tap.html

src/migrate/index.js
```diff
@@ -196,7 +196,7 @@ export default class Migrator {
...
+    .then(retVal => this._freeLock().then(() => retVal))
```
dannyfritz: If you're using Bluebird, a passthrough such as .tap might be cleaner. Same thing though!
http://bluebirdjs.com/docs/api/tap.html

elhigu: I was thinking about it, but noticed that I cannot use `.tap` since if it throws an error control won't go to catch block in promise chain. 

Other reason why I don't like to use it is that I prefer nowadays to use node's native promise since es6 has it now without need to import bluebird.

rhys-vdw: > I was thinking about it, but noticed that I cannot use .tap since if it throws an error control won't go to catch block in promise chain (as far as I know).

FYI This is the not the case. `tap` would be fine.

Issue #1198, opened 2016-02-12
Not able to handle when server ends the connection


woopstar:
I'm seeing this problem because there is a connection timeout for 600 seconds from the MySQL server. Then a connection is closed from the server instead of the client.

When Knex then tried to execute a query, the following error comes:

```
Unhandled rejection Error: BEGIN; - This socket has been ended by the other party
    at Socket.writeAfterFIN [as write] (net.js:297:12)
    at Connection.write (/node_modules/mysql2/lib/connection.js:136:15)
    at Connection.writePacket (/node_modules/mysql2/lib/connection.js:150:10)
    at Query.start (/node_modules/mysql2/lib/commands/query.js:39:14)
    at Query.Command.execute (/node_modules/mysql2/lib/commands/command.js:34:20)
    at Connection.handlePacket (/node_modules/mysql2/lib/connection.js:310:28)
    at Connection.addCommand (/node_modules/mysql2/lib/connection.js:326:10)
    at Connection.query (/node_modules/mysql2/lib/connection.js:397:15)
    at /node_modules/knex/lib/dialects/mysql/index.js:92:18
    at tryCatcher (/node_modules/knex/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._resolveFromResolver (/node_modules/knex/node_modules/bluebird/js/main/promise.js:480:31)
    at new Promise (/node_modules/knex/node_modules/bluebird/js/main/promise.js:70:37)
    at Client._query (/node_modules/knex/lib/dialects/mysql/index.js:88:12)
    at Client.query (/node_modules/knex/lib/client.js:127:24)
    at /node_modules/knex/lib/transaction.js:219:21
    at tryCatcher (/node_modules/knex/node_modules/bluebird/js/main/util.js:26:23)
```

wubzz:
Vague interpretation of what you mean by "Not able to handle" I'm guessing you say that because the error was thrown rather than handled? (Missing `.catch` block?)

Or are you saying knex should attempt to open up a new connection when this occurs?

woopstar:
I believe Knex should attempt to open a connection when this occours. If a pool object exists, but the sql server has closed the connection and you're trying to send a query. In stead of failing, it should first attempt to reconnect and if not able to do so, then fail?

elhigu:
@woopstar I agree knex should be able to handle that case and just throw away the connection and try to get new one if DB side had closed it. Then if reconnection doesn't work pool can throw connection timeout exception...

[...]

PR #1200, opened 2016-02-13
Bugfix/acquire new connection when socket closed

wubzz:
@elhigu @woopstar In reference to #1198 this will "retry" a runner's job once it fails due to the socket being closed by the server. A few key notes here that had me going back and forth a bit..

1. The `Runner.run` function seems to be the best place to put this to avoid duplicating code between `Runner.query` and `Runner.queryArray` and to avoid having to move when/where events are being emitted.

2. The error thrown does not contain a code, only a message. In the listed issue, the message for mysql was `This socket has been ended by the other party` while in Postgres it's something like `Socket is closed` -- So I have no clue how to globally distinguish the specific error across all dialects.

3. I avoided recursion here since retrying more than once does not sound like a good idea.

4. I have no clue how to write a test for this since the connections need to be closed by the server. Any ideas?

Test code used

```javascript
knex.raw('SELECT version()')
.then(function() {
debugger; //Manually close connections on the server
return knex.raw('SELECT version()');
})
.then(function(result) {
//Result is present, query success.
}).catch(function(error) {
//Previously 'socket closed' error.
});
```

Code may need some cleanup, putting this up to see if I'm on the right track.

### DIFF
--- a/src/runner.js
+++ b/src/runner.js
@@ -40,7 +40,7 @@ assign(Runner.prototype, {
       .then(resolver)
       .catch((error) => {
           //TODO: -- Temporary check during test-phase. How to properly check this globally across all dialects?
-          if(error.message.indexOf('socket') !== -1) {
+          if(error.message.toLowerCase().indexOf('this socket is closed') !== -1) {
             //Socket closed, so kill connection and try one more time with a fresh connection.
             runner.connection.__knex__disposed = true;
             runner.client.pool.destroy(runner.connection);
--- a/src/runner.js
+++ b/src/runner.js
@@ -25,22 +25,32 @@ assign(Runner.prototype, {
   // a single connection.
   run: function() {
     var runner = this
-    return Promise.using(this.ensureConnection(), function(connection) {
-      runner.connection = connection;
 
+    return new Promise((resolver, rejecter) => {
       runner.client.emit('start', runner.builder)
       runner.builder.emit('start', runner.builder)
+
       var sql = runner.builder.toSQL();
 
       if (runner.builder._debug) {
         console.log(sql)
       }
 
-      if (_.isArray(sql)) {
-        return runner.queryArray(sql);
-      }
-      return runner.query(sql);
-
+      return runner._execSQL(sql)
+      .then(resolver)
+      .catch((error) => {
+          //TODO: -- Temporary check during test-phase. How to properly check this globally across all dialects?
+          if(error.message.indexOf('socket') !== -1) {
+            //Socket closed, so kill connection and try one more time with a fresh connection.
+            runner.connection.__knex__disposed = true;
+            runner.client.pool.destroy(runner.connection);
+            runner.connection = null; //ensureConneciton will create a new connection
+            return runner._execSQL(sql)
+              .then(resolver)
+              .catch(rejecter)
+          }
+          rejecter(error)
+      })
     })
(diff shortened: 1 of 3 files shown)


src/runner.js
```diff
@@ -61,6 +65,24 @@ assign(Runner.prototype, {
...
   },
 
+  _run: function() {
+    var runner = this
+    return Promise.using(this.ensureConnection(), function(connection) {
+      runner.connection = connection;
+
+      var sql = runner.builder.toSQL();
+
+      if (runner.builder._debug) {
```
wubzz: `sql` and if-statement should be moved to `run()` instead of `_run()` to avoid double `console.log` when a retry is called. And a better name for the function as well.

woopstar:
I was looking at https://github.com/felixge/node-mysql/issues/761, and saw the following line:

`if (err.code === 'PROTOCOL_CONNECTION_LOST') {`

Maybe it would help detecting when a socket is closed?

woopstar:
It also seems to be mentioned here https://github.com/sidorares/node-mysql2/issues/57

wubzz:
Thanks for that info @woopstar , means checking for mysql shouldn't be a problem.

Debugging the pg library it seems that it simply proxies the error to the `.query` callback function, and it only contains the message property, and no `code`.

Not sure how SQLlite or mssql handles this..

elhigu:
@wubzz I think that 1 retry try is good for initial implementation, I'll try to have time to check this out soon. Sorry for delay.

wubzz:
@elhigu Commits should be fixed now. Currently the big issue with this PR is distinguishing the Error uniquely across all dialects. MySQL has a code, Postgres has only a string, and not sure on the others.

elhigu:
@wubzz probably sqlite never does this, since it supports only one connection. No idea of oracle / mssql... we should do a bit of googling to checkout from every driver if they support some way to recognize lost connection. Is there any other nodejs ORM which does recognize this case, how they do it?

For testing at least with postgresql you can write a query which closes all active connections, but I'm not sure if that is same kind of case that naturally lost connection... I would like to find out to make this some robust and testable way.

wubzz:
@elhigu I tried writing a test using pg's `pg_terminate_backend` but it seems that does not throw the same kind of error.

I was however able to reproduce the identical error by nullifying the connection's socket handle.
(Runner)
```
this.connection.connection.stream._handle = null;
```

This may seem impractical, but works:

```javascript
it('When Socket is closed by server, acquire a new connection and retry running the query.', function() {
				return knex.raw('SELECT version()')
					.then(() => {
						var qBuilder = knex.raw('SELECT version()');
						
						var Runner = qBuilder.client.runner(qBuilder);
						Runner.run();

						qBuilder.once('query', function() {
							Runner.connection.connection.stream._handle = null;
						});

						return qBuilder;
					})
					.catch((error) => {
						expect(true).to.equal(false);
					})
			});
```

wubzz:
I have been testing back and forth with mysql but cannot reproduce the error by closing connections via administrative tools or by introducing a connection timeout as mentioned in the issue.

This leads me to question the postgres error I've been seeing as well, it may not be of the same type described in the issue.

elhigu:
@wubzz I left comment to #1198 I think that after all this approach is not viable. Sorry about my bad advices earlier :(

PR #1201, opened 2016-02-14
Add .timeout(ms) to QueryBuilder

wubzz:
A feature suggestion that adds timecontrol over queries to the querybuilder that is being run in an application, similarly to how you can now control the Timeout for acquiring a new conection in 0.10.

Probably not often that this would have a use case, but in complex queries where you have tons of joins and tons of data, this could prove useful. If not for anything else, it can be used for finding queries that are slowing down an app.

This is not a requested feature or anything that I am personally in need of, just something extra.

### DIFF
--- a/index.html
+++ b/index.html
@@ -67,6 +67,7 @@
     </a>
     <ul class="toc_section">
       <li>– <a href="#Builder-main"><b>constructor</b></a></li>
+      <li>– <a href="#Builder-timeout">timeout</a></li>
       <li>– <a href="#Builder-select">select</a></li>
       <li>– <a href="#Builder-as">as</a></li>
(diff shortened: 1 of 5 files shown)


elhigu:
@wubzz if anyhow possible could you rebase this on master and squash some of those back and forth commits? `git commit --amend` is good way to add fixes to last commit that you did if you find that there was a problem in the commit you just put in. Also `git rebase -i origin/master` gives you pretty nice control what is getting in and how.

wubzz:
@elhigu Been trying a Git client with no visual representation built it, so it's a bit odd. I'll try get it sorted. Worst case I'll cherry-pick on a new branch and open a new one.

PR #1205, opened 2016-02-15
Fix CLI failing with TypeError if knex installation missing

jharting:
When run without local knex installation CLI currently fails with a nice error message in some cases and with TypeError with other cases (migrate:make, seed:make). This changes aligns the behavior to always show the expected error message.


elhigu:
@jharting thanks!

PR #1209, opened 2016-02-16
More control over timestamp defaults and auto updating.

fiveladdercon:
This is yet another proposal/fix about dealing with timestamps and NOW (i.e. #1109 etc.).

Firstly, I propose that an empty or undefined .defaultTo() on a timestamp results in a NOW default across the dialects.  This abstracts NOW and eliminates the need for a special NOW symbol in user code:

```
.timestamp('modified_at').defaultTo()
 --> modified_at timestamp default CURRENT_TIMESTAMP

.timestamp('modified_at').defaultTo(undefined)
 --> modified_at timestamp default CURRENT_TIMESTAMP
```

This change is made in the ColumnCompiler baseclass and permits per-dialect overrides as required.  The MySQL dialect overrides the baseclass as a demonstration and means the baseclass behavior can be changed if there is a better option across the dialects.

Secondly, I've added an .updating() chainable:

```
.timestamp('modified_at').defaultTo().updating()
-->  modified_at timestamp default CURRENT_TIMESTAMP on update CURRENT_TIMESTAMP`
```

This splits off the on update from the default because both options are valid on their own (in different tables):

```
x.timestamp('a').defaultTo() --> a timestamp default CURRENT_TIMESTAMP
y.timestamp('b').updating() --> b timestamp on update CURRENT_TIMESTAMP

```

This is only implemented in the MySQL dialect because I'm not sure of it's support in other dialects.

The last change is a convenience for a default of '0000-00-00 00:00:00' in non TRADITIONAL SQL mode:

```
.timestamp('a').defaultTo(0)
--> a timestamp default 0
```

Again this change is only in the MySQL dialect.  The integration test temporarily disables TRADITIONAL mode.


elhigu:
Closing due to inactivity. Also I think this should be implemented for every dialect to be merged in.

PR #1213, opened 2016-02-19
Ensure each line of query is less than 2500 characters

AiNoKame:
Using strong-oracle and sqlplus, sqlplus has a command line limit of 2500 characters, and I have an eager loading select query with 6 joins, which surpasses the limit. I can perform the query from the sqlplus client by just reading from a file with the query and having a carriage return at at the ~2499 character mark (outside of an aliased string), but knex's sql query currently contains one string and not an option to read a long, split up query from a file. This PR just inserts `\n`(s) in the sql query string to allow long queries.

### DIFF
--- a/src/dialects/oracle/index.js
+++ b/src/dialects/oracle/index.js
@@ -119,6 +119,35 @@ assign(Client_Oracle.prototype, {
     });
   },
 
+  // Ensures query is less than 2500 characters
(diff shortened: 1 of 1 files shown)


AiNoKame:
I'd love to know if there's a better place to put this function :smiley:

PR #1226, opened 2016-02-25
Fixed nested transactions

jurko-gospodnetic:
The original code for handling nested transactions did not work correctly when you had sibling nested transactions, i.e. two or more transactions all living inside the same parent transaction and latter ones getting started after their predecessor sibling completes (rolls back or commits). In such cases, knex would hang on any DB operation attempted using a non-initial sibling transaction.

Basic cause was in knex _accidentally_ waiting for that same transaction to complete before allowing someone else to access the transaction's database connection. :smiling_imp:

Basic bug fixed in d0836b8f880373836b153e922f401496e2914466.

Detailed tests added for handling sibling tested transactions in 675c684d3bd6b55ed63e4a6d3035754d2964cf8d. You can run those tests to reproduce the hang if you run them without the fix applied.

The three later commits (e0e77dfaf8dd13719bbebe1279d4a6d601595be2, f38f901bd5745399e129635febf2db35bfcae2f3, 675c684d3bd6b55ed63e4a6d3035754d2964cf8d) then just clean up the related code. And in the end the code is simplified enough that the original fix becomes unnecessary to begin with. :smile:

Hope this helps.

Best regards,
  Jurko Gospodnetić

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -63,7 +63,7 @@ function Transaction(client, container, config, outerTx) {
   this._childQueue = []
 
   // The queue is a noop unless we have child promises.
-  this._queue = this._queue || Promise.resolve(true)
+  this._queue = Promise.resolve(true)
 
   // If there's a wrapping transaction, we need to see if there are 
   // any current children in the pending queue.
@@ -74,9 +74,7 @@ function Transaction(client, container, config, outerTx) {
     if (outerTx._childQueue.length > 0) {
 
       var previousSibling = outerTx._childQueue[outerTx._childQueue.length - 1]
-      this._queue = this._queue.then(function() {
-        return Promise.settle([previousSibling])
-      })
+      this._queue = Promise.settle([previousSibling])
 
     }
 
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -73,8 +73,9 @@ function Transaction(client, container, config, outerTx) {
     // settles (commit or rollback) and then we can continue.
     if (outerTx._childQueue.length > 0) {
 
+      var previousSibling = outerTx._childQueue[outerTx._childQueue.length - 1]
       this._queue = this._queue.then(function() {
-        return Promise.settle(outerTx._childQueue[outerTx._childQueue.length - 1])
+        return Promise.settle([previousSibling])
       })
 
     }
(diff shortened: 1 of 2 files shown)


jurko-gospodnetic:
Note that this pull request does not attack the issue of why we'd want to try to synchronize sibling nested transactions in the first place, but only makes sure that the logic already implemented no longer breaks.

The current code, due to the way transactions perform their `SAVEPOINT` commands directly at transaction startup, already depends on callers keeping such sibling nested transactions in sync by properly chaining their promises. Thus, it might be that this internal synchronization is either not needed, or needs to be fixed to work for cases when the caller does not chain those promises correctly. But I think that's a topic for a separate story. As there are no tests related to that, I was loath to do that cleanup here lest I confuse the original hang issue that needed to be fixed.

jurko-gospodnetic:
Ping?

@wubzz & @rhys-vdw - you guys appeared to take a look at my last pull request, not sure who's the correct person to ping about this one or who gets notifications about newly added pull requests here.

wubzz:
@jurko-gospodnetic I'm not too familiar with the code for transactions, but the changes here makes sense to me and it seems you thoroughly tested this with several new tests and also updated the inline documentation so I am perfectly fine with merging this.

Thanks!

src/transaction.js
```diff
@@ -238,7 +222,7 @@ function makeTxClient(trx, client, connection) {
     })
   }
   trxClient.acquireConnection = function() {
-    return trx._queue.then(function() {
+    return Promise.settle([trx._previousSibling]).then(function () {
```
elhigu: Is it ok here to continue if `_previousSibling` did fail?

jurko-gospodnetic: This behaviour is unchanged. And from how I read the original code, the idea is that if we have a set of transactions like this:

```
TP (parent)
    - TS1 (first sibling)
    - TS2 (second sibling)
```

TS2 should be allowed to proceed (i.e. acquire its connection so it can perform actual database operations) only after TS1 has completed its work. Whether TS1 got committed or rolled back.

If we want TS2 to be run only if TS1 gets committed, then we should created TS2 in a `.then()` block chained to the TS1 resolution promise, and not just create it as a completely unrelated sibling transaction.

PR #1242, opened 2016-03-02
Update lodash from 3.7.0 -> 4.6.0

### LINKED ISSUE #1214: Update Lodash dependency from v3 to v4
The lodash file structure has been changed making the following lines break with an error:
```MODULE_NOT_FOUND,Cannot find module 'lodash/object/assign'```
So basically the problem as I see it is deep require statements similar to:
```var assign = require('lodash/object/assign');```
Not sure why this these weren't like more like:
```var assign = require('lodash').assign;```
in the first place but here is an idea where to start.

wubzz:
Fixes #1214 . Updates lodash to v4, which affects more or less all project files. I would only include this PR for a big release, even if tests pass.


wubzz:
Sorry, it looks like my IDE went apeshit on the indentation.. I'll fix this.

blah238:
:+1: great work @wubzz! Since this is such an integral dependency I agree it would be wise to bump at least the minor version.

elhigu:
@wubzz looks good, but need to resolve conflict before being able to merge

wubzz:
@elhigu Yeah, no problem, I'll resolve those. Though, I'm not sure if we want to merge this already or not. I'm thinking 0.11 to be safe.

elhigu:
If we merge this now, next knex release will be 0.11 :) On the other hand if we haven't made any breaking changes since 0.10 we could roll out 0.10.1 already.

elhigu:
@wubzz next release will be 0.11 so no need to be on hold with this

wubzz:
@elhigu Sweet, I'll get this done sometime this week. Do you think Upgrading Bluebird would be appropiate at the same time? There's an issue for this too.

PR #1247, opened 2016-03-04
Update docs for returning to include multiple columns

caelon:
Added a use case and an example for .returning


wubzz:
@caelon Thanks!

PR #1250, opened 2016-03-07
Fixes inTable(undefined) silently fails bug from Issue #1220

### LINKED ISSUE #1220: inTable(undefined) fails silently
table.foreign('foreignId').references('id').inTable(undefined);

fails silently whereas

    table.foreign('foreignId').references('id').inTable('nonExistentTable');

throws an error (env: knex 0.10.0, pg: 4.4.6, postgresql: 9.5, node: 5.6.0). Is that expected behaviour or is it a bug? I had a hard time to figure out why my foreign key wasn't created...

ngoctranfire:
Fixes #1220, https://github.com/tgriesser/knex/issues/1220, inTable(undefined) fails silently error by making sure it throws an error if the user passes an undefined table name.

### DIFF
--- a/src/schema/tablebuilder.js
+++ b/src/schema/tablebuilder.js
@@ -190,8 +190,8 @@ TableBuilder.prototype.foreign = function(column) {
         foreignData.references = pieces ? pieces[0] : tableColumn;
         return {
           on: function(tableName) {
-            if (typeof tableName === 'undefined') {
-              throw new Error("table name is of type 'undefined'. Please pass in a valid table name");
+            if (typeof tableName !== 'string') {
+              throw new TypeError(`Expected tableName to be a string, got: ${typeof tableName}`);
             }
             foreignData.inTable = tableName;
             return returnObj;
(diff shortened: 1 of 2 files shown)


src/schema/tablebuilder.js
```diff
@@ -190,6 +190,9 @@ TableBuilder.prototype.foreign = function(column) {
         foreignData.references = pieces ? pieces[0] : tableColumn;
         return {
           on: function(tableName) {
+            if (typeof tableName === 'undefined') {
+              throw new Error("table name is of type 'undefined'. Please pass in a valid table name");
```
wubzz: @ngoctranfire Would you be ok with checking if `typeof tableName !== 'string'` and throwing a `TypeError` instead? It would make more sense considering supplying anything other than a string should be failing anyways.

Error message could be something like `Expected tableName to be a string, got: --typeof tableName--`

ngoctranfire: Yup. Going to fix it asap right now.

wubzz:
This looks good now. Thanks!

PR #1251, opened 2016-03-07
Update regex for named bindings

### LINKED ISSUE #1128: "Module not found" errors when using knex with webpack
I am trying to use knex in a project I'm working on. The problem I'm having is that when I try to run webpack it is tracing the requires and attempting to pull in drivers/dialects that I'm not using and don't have installed.

Any thoughts on how this can be solved?

`````` log
ERROR in ./~/mariasql/lib/Client.js
Module not found: Error: Cannot resolve 'file' or 'directory' ../build/Debug/sqlclient in ~/project/node_modules/mariasql/lib
 @ ./~/mariasql/lib/Client.js 17:10-45

ERROR in ./~/knex/lib/dialects/sqlite3/index.js
Module not found: Error: Cannot resolve module 'sqlite3' in ~/project/node_modules/knex/lib/dialects/sqlite3
 @ ./~/knex/lib/dialects/sqlite3/index.js 33:11-29

ERROR in ./~/knex/lib/dialects/mysql2/index.js
Module not found: Error: Cannot resolve module 'mysql2' in ~/project/node_modules/knex/lib/dialects/mysql2
 @ ./~/knex/lib/dialects/mysql2/index.js 33:11-28

ERROR in ./~/knex/lib/dialects/mysql/index.js
Module not found: Error: Cannot resolve module 'mysql' in ~/project/node_modules/knex/lib/dialects/mysql
 @ ./~/knex/lib/dialects/mysql/index.js 35:11-27

ERROR in ./~/knex/lib/dialects/oracle/index.js
Module not found: Error: Cannot resolve module 'oracle' in ~/project/node_modules/knex/lib/dialects/oracle
 @ ./~/knex/lib/dialects/oracle/index.js 40:11-28

ERROR in ./~/knex/lib/dialects/postgres/index.js
Module not found: Error: Cannot resolve module 'pg' in ~/project/node_modules/knex/lib/dialects/postgres
 @ ./~/knex/lib/dialects/postgres/index.js 46:11-24

ERROR in ./~/knex/lib/dialects/postgres/index.js
Module not found: Error: Cannot resolve module 'pg-query-stream' in ~/project/node_modules/knex/lib/dialects/postgres
 @ ./~/knex/lib/dialects/postgres/index.js 132:50-76

ERROR in ./~/knex/lib/dialects/strong-oracle/index.js
Module not found: Error: Cannot resolve module 'strong-oracle' in ~/project/node_modules/knex/lib/dialects/strong-oracle
 @ ./~/knex/lib/dialects/strong-oracle/index.js 15:9-33```
``````

### LINKED ISSUE #1228: Cannot seem to use named bindings for arrays
```
var knex = require('knex');
var postgresConn = 'postgresql://user:pw@host:port/db?tcpKeepAlive=true';

var redshift = knex({
    client: 'pg',
    connection: postgresConn
});


var sql = redshift.raw('select * from schema.table where col_1 in (:col_1_values) and col_2 = :col_2_value',
    {col_1_values: [[1, 2, 3]], col_2_value: 99}).toString();

console.log(sql);
```

Expected output:
select * from schema.table where col_1 in (1,2,3) and col_2 = '99'

Actual output:
select * from schema.table where col_1 in (:col_1_values) and col_2 = '99'

knex version: 0.10.0
pg version: 3.6.2

### PR DESCRIPTION
**Pull request by wubzz**
Fixes #1228  . I'm unsure why the first condition of the regular expression was there, as it shouldn't matter at all?

It should now properly match named bindings, and will not require a space before the binding *(see related issue)*.

I also added a test for this based on examples in documentation.

See http://regexr.com/3cv16

------------------------
**TODO**
- [x] Change regular expression so that spaces are not mandatory for named bindings
- [x] If a binding value is undefined, leave the binding in the sql untouched
- [x] Update docs explaining these changes in detail.

### DIFF
--- a/src/raw.js
+++ b/src/raw.js
@@ -127,7 +127,7 @@ function replaceKeyBindings(raw) {
   var client   = raw.client
   var sql      = raw.sql, bindings = []
 
-  var regex = new RegExp('(^|\\s)(\\:\\w+\\:?)', 'g')
+  var regex = new RegExp('(\\:\\w+\\:?)', 'g')
   sql = raw.sql.replace(regex, function(full) {
     var key = full.trim();
     var isIdentifier = key[key.length - 1] === ':'
--- a/src/raw.js
+++ b/src/raw.js
@@ -128,11 +128,15 @@ function replaceKeyBindings(raw) {
   var sql      = raw.sql, bindings = []
 
   var regex = new RegExp('(\\:\\w+\\:?)', 'g')
+  var emptyBindings = [];
   sql = raw.sql.replace(regex, function(full) {
     var key = full.trim();
     var isIdentifier = key[key.length - 1] === ':'
     var value = isIdentifier ? values[key.slice(1, -1)] : values[key.slice(1)]
-    if (value === undefined) return ''
+    if (value === undefined) {
+      emptyBindings.push(full);
+      return full;
+    }
     if (value && typeof value.toSQL === 'function') {
       var bindingSQL = value.toSQL()
       if (bindingSQL.bindings !== undefined) {
@@ -146,6 +150,10 @@ function replaceKeyBindings(raw) {
     bindings.push(value)
     return full.replace(key, '?')
   })
+
+  emptyBindings.map(function(binding) {
+    sql = sql.replace(new RegExp(` ${binding}|${binding} |${binding}`, 'g'), '');
+  });
 
   return {
     method: 'raw',
--- a/src/raw.js
+++ b/src/raw.js
@@ -128,13 +128,11 @@ function replaceKeyBindings(raw) {
   var sql      = raw.sql, bindings = []
 
   var regex = new RegExp('(\\:\\w+\\:?)', 'g')
-  var emptyBindings = [];
   sql = raw.sql.replace(regex, function(full) {
     var key = full.trim();
     var isIdentifier = key[key.length - 1] === ':'
     var value = isIdentifier ? values[key.slice(1, -1)] : values[key.slice(1)]
     if (value === undefined) {
-      emptyBindings.push(full);
       return full;
     }
     if (value && typeof value.toSQL === 'function') {
@@ -150,10 +148,6 @@ function replaceKeyBindings(raw) {
     bindings.push(value)
     return full.replace(key, '?')
   })
-
-  emptyBindings.map(function(binding) {
-    sql = sql.replace(new RegExp(` ${binding}|${binding} |${binding}`, 'g'), '');
-  });
 
   return {
     method: 'raw',
--- a/index.html
+++ b/index.html
@@ -2224,18 +2224,27 @@ knex('users')
   .orWhere(knex.raw('status <> ?', [1]))
   .groupBy('status')
 </pre>
+
+    <p>
+      Positional bindings <tt>?</tt> are interpreted as values and <tt>??</tt> are interpreted as identifiers.
+    </p>
+
 <pre class="display">
-// positional bindings ? is interpret as value and ?? as identifier
 knex('users').where(knex.raw('?? = ?', ['user.name', 1]))
 </pre>
 
+    <p>
+      Named bindings such as <tt>:name</tt> are interpreted as values and <tt>:name:</tt> interpreted as identifiers.
+      Named bindings are processed so long as the value is anything other than <tt>undefined</tt>.
+    </p>
+
 <pre class="display">
-// named bindings :name is interpret as value and :name: as identifier
 knex('users')
-  .where(knex.raw(':name: = :thisGuy or :name: = :otherGuy', {
+  .where(knex.raw(':name: = :thisGuy or :name: = :otherGuy or :name: = :undefinedBinding', {
     name: 'users.name',
     thisGuy: 'Bob',
-    otherGuy: 'Jay'
+    otherGuy: 'Jay',
+    undefinedBinding: undefined
   }))
 </pre>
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by wubzz** [comment]
From the failing tests I can assume the previous condition in the regexp was for preventing an extra space if a named binding was left undefined. I'll try find another solution to that issue.

**Comment by wubzz** [comment]
@elhigu Would you mind reviewing this? I think you're the right person as you worked with this code previously.

**Comment by elhigu** [comment]
@wubzz sorry, I hadn't went through github notifications last few days. I'll check this out tomorrow. btw. pinging me in knex gitter chat might be easier to get my attention time to time.

File: src/raw.js [hunk 1]
```diff
@@ -116,12 +116,16 @@ function replaceKeyBindings(raw) {
   var client   = raw.client
   var sql      = raw.sql, bindings = []
 
-  var regex = new RegExp('(^|\\s)(\\:\\w+\\:?)', 'g')
+  var regex = new RegExp('(\\:\\w+\\:?)', 'g')
```
elhigu: Earlier bindings has been working in a way that there had to be space before start of named binding. Old regexp was: `\\s(\\:\\w+\\:?)` so I had changed it to allow also be in start of line `(^|\\s)(\\:\\w+\\:?)`. 

I don't know about original motivation why named bindings were not allowed to be used without leading white space.

elhigu:
@wubuzz I believe that there shouldn't be any reason why named bindings would require leading whitespace. They are just converted to positional bindings anyways and as far as I know one can give them without leading spaces.

src/raw.js
```diff
@@ -136,6 +140,10 @@ function replaceKeyBindings(raw) {
     return full.replace(key, '?')
   })
 
+  emptyBindings.map(function(binding) {
+    sql = sql.replace(new RegExp(` ${binding}|${binding} |${binding}`, 'g'), '');
```
elhigu: Why trailing / leading space is removed when no bindings?

wubzz: The previous regular expression matched by forcing a trailing space, so it would convert the following:

`SELECT :col`
to
`SELECT`

Whereas removing the space requirement from the regular expression would instead convert

`SELECT :col`
to
`SELECT ` *(Space remains)*

Generally this shouldn't be an issue execution wise, but from the failing tests I figured why not keep it that way. Basically if a binding is supplied in the query like `:col`, but the binding value is undefined, the previous could would remove that binding from the query and its trailing space. That's what I'm doing in that code to supplement the removal of space from the regex.

wubzz: See [this test](https://github.com/tgriesser/knex/blob/master/test/tape/raw.js#L78) for reference.

elhigu: @wubuzz probably the best choice here would be just to replace named binding with ? and if value is not found from bindings, then it would do the same that giving undefined to positional binding:

```js
require('knex')({client: 'pg'}).raw('?', [undefined]).toString()
// 'DEFAULT'
require('knex')({client: 'pg'}).raw('?').toString()
// '?'
```

So e.g.

```js
require('knex')({client: 'pg'}).raw(':binding').toString()
// ':binding'

require('knex')({client: 'pg'}).raw(':binding', []).toString()
// should throw an error, since named bindings require an object 

require('knex')({client: 'pg'}).raw(':binding', {}).toString()
// 'DEFAULT'

require('knex')({client: 'pg'}).raw(':binding', {binding: 'tadaa'}).toString()
// '\'tadaa\''
```
This way functionality would be consistent with positional bindings. To me current functionality is really inconsistent.

How does this sound?

wubzz: @elhigu It sounds reasonable for most use cases, such as named bindings for setting or inserting values, but when you start using the named bindings for columns in tablenames it might seem odd to replace it with `DEFAULT`, or really anything at all.

```javascript
knex.raw('SELECT * FROM table WHERE :column: = :value', {column: 'users.name', value: 'knex'}).toString();
///... WHERE "users"."name" = 'knex'
```

```javascript
knex.raw('SELECT * FROM table WHERE :column: = :value', {column: void 0, value: 'knex'}).toString();
//... WHERE DEFAULT = 'knex'
```

I do however agree that completely erasing the binding from the query which we did previously, and currently still are, is very obnoxious and should not be done.

Perhaps if it's a regular binding such as `:val` we can use the `DEFAULT` approach, but if it's `:column:` we leave it the way it is? I don't know.. Can't come up with something that is consistent throughout..

elhigu: @wubzz it should be enough that named bindings are converted to positional bindings with as little magic as possible and then it is job of some other part to decide how to handle `undefined` in positional binding.

I agree that replacing `undefined` with `DEFAULT` is not good functionality... probably throwing an error would be the best way to handle it for positional bindings and for named binding. Converting it to `DEFAULT` is just an accidental undefined behavior... Anyways discussion how to handle `undefined` in case of positional bindings should be done in separate pull request.

wubzz: @elhigu I agree. This function should only replace bindings when it's possible. My proposal is this:

If a binding is present in the query, but is undefined in the bindings object, then simply leave the binding in the sql untouched. That will in itself throw a query error, and thanks to not replacing the binding, it will be obvious what went wrong via the error message. In other words, delete the code that deals with `emptyBindings`.

```javascript
knex.raw('SELECT :val', {val: 1}); //SELECT 1

knex.raw('SELECT :val', {val: void 0}); //SELECT :val -- Throws error
```

You agree and good with merging after this? I expect a few tests to change to accomodate for removing that code, like the references above. I also plan to add this to the docs.

wubzz:
@elhigu I've updated the code, test, and documentation. I feel this is a good solution for now, the processing of undefined bindings like you said is another issue.

elhigu:
@wubzz Looks good thanks for fixing this.

PR #1252, opened 2016-03-07
Code cleanup

jurko-gospodnetic:
Minor code cleanup collected while working on the knex code-base and extracted here so it would not unnecessarily pollute other pull requests.


wubzz:
@jurko-gospodnetic Thanks!

PR #1257, opened 2016-03-08
Fix transaction callback throw handling

jurko-gospodnetic:
Originally if a transaction callback threw an error, `knex`'s & database's transaction tracking got out of sync:
- `knex` considered the transaction completed & rejected its transaction promise
- DB was not told to roll back its transaction so the used database connection got returned to the connection pool without its transaction actually getting rolled back

Effect would be random database deadlocks and/or operations failures later on due to application code running under a transaction when it does not expect to be, e.g. it can cause that transaction to start holding database locks and thus block other regular transactions, and all of it at random depending on the order in which connections get grabbed from the `knex` connection pool.

This changes the behaviour by treating such thrown exceptions the same as if a rejected promise had been returned from them and rolls back the underlying database transaction.

The pull request also does some minor code cleanup to keep the code directly touched by it in sync with the rest of the code.

Initial cleanup:
- 85b780d4fa7c376717b2933d76f7a9c68c12e989 sync test comments about oracle & mssql not reporting BEGIN/ROLLBACK
- 3d8131b5eb5bde8744a169c66ee6135018202a83 add missing tape assertion descriptions

Actual fix:
- 84f85beb73edd040afdc5511e5a6cc939c4cd46f sync knex & DB transaction state on trx callback error throw

Test suite updates:
- 3225ab45bcf5d8ba11b97c9b64b8441ac26cb14a test transaction callback throwing an error
- 96bd74e0c12c3b99831100beb4e0cf1b94c20d55 remove now redundant test case

Doc updates:
- 20843aa8d6cba84d091be0b16ba90364d8cfc67f document handling errors thrown from an exception handler function

### DIFF
--- a/index.html
+++ b/index.html
@@ -1709,6 +1709,12 @@ knex.transaction(function(trx) {
 </pre>
 
   <p>
+    Throwing an error directly from the transaction handler function
+    automatically rolls back the transaction, same as returning a rejected
+    promise.
+  </p>
+
+  <p>
     Notice that if a promise is not returned within the handler, it is up to you to ensure <tt>trx.commit</tt>, or
     <tt>trx.rollback</tt> are called, otherwise the transaction connection will hang.
   </p>
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -32,10 +32,15 @@ function Transaction(client, container, config, outerTx) {
       return makeTransactor(this, connection, trxClient)
     })
     .then((transactor) => {
-      var result = container(transactor)
-
       // If we've returned a "thenable" from the transaction container, assume
       // the rollback and commit are chained to this object's success / failure.
+      // Directly thrown errors are treated as automatic rollbacks.
+      var result
+      try {
+        result = container(transactor)
+      } catch (err) {
+        result = Promise.reject(err)
+      }
       if (result && result.then && typeof result.then === 'function') {
         result.then((val) => {
           transactor.commit(val)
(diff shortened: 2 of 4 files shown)


jurko-gospodnetic:
Could someone run the test suite on `mssql` and `oracle`? I don't see those tested on Travis, and I don't have the environment for them set up. The fix is DB agnostic, so that should not be affected but ti would be great if someone could make sure the test code does not break when used with those systems (their `knex` interface is a bit different than the one for the other systems in that they do not report their `BEGIN` & `ROLLBACK` commands as `knex` queries).

jurko-gospodnetic:
One failed test listed above is completely random and happened only once - due to some unrelated and badly written test code that did not expect to be running on a database system performant enough to finish its query before its 1 ms timeout kicks in.

jurko-gospodnetic:
This pull request supersedes pull request #928.

jurko-gospodnetic:
@wubzz - whom to ask to review/approve/merge this? If it helps, we can do a Google hangout session or something similar and I can help by explaining things directly.

Not being altruistic here, I'd just like to get this in so I can stop maintaining my own project forks for some projects that need these fixes. :stuck_out_tongue_winking_eye:

wubzz:
@jurko-gospodnetic So the gist of this PR is to enforce automatic Rollbacks **if** any type of error *(intentional or not)* is thrown within the transactors container?

Just wanna double check I understand this correctly.

jurko-gospodnetic:
Yup.

We already do that partially by rejecting the transaction promise (the one returned by the `db.transaction()` call) when that happens. Unfortunately we also run the `begin()/savepoint()` operation on that transaction but then forget to run a corresponding rollback one.

wubzz:
@jurko-gospodnetic That makes sense. Thanks!

jurko-gospodnetic:
thanks!!!!

PR #1258, opened 2016-03-08
Code cleanup (trimmed trailing spaces & removed tabs)

jurko-gospodnetic:
Trimmed trailing spaces & converted some tabs to spaces to maintain code consistency.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1251,10 +1251,10 @@ knex('coords').insert([{x: 20}, {y: 30},  {x: 10, y: 20}])
       Only utilized by PostgreSQL databases, the returning method specifies which column should be returned
       by the <a href="#Builder-insert">insert</a> and <a href="#Builder-update">update</a> methods.
       <br />
-	  Passed <tt>column</tt> parameter may be a string or an array of strings. When passed in a string,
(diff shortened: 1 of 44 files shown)


jurko-gospodnetic:
The only tests currently reported as failing are completely random - due to some unrelated and badly written test code that did not expect to be running on a database system performant enough to finish its query before its 1 ms timeout kicked in.

Same effect noticed on pull request #1257.

wubzz:
@jurko-gospodnetic Yea, that test has come back to bite us in the ass more than once. It's a bit tricky to make an integration test across all dialects for it. Currently it seems only postgres is failing from time to time. Will try to adjust the insert query for postgres specifically, unless you have a better proposal.

I've rerun the build on your pull requests for now.

jurko-gospodnetic:
Well, you can use `pg_sleep()` in that query for Postgres to make it sleep a given number of seconds. Not portable, but will get the immediate issue out of the way.

wubzz:
@jurko-gospodnetic See attempts in #1260 .

Thanks for the PR btw, merging this.

PR #1260, opened 2016-03-08
Add .timeout to raw, and attempt to fix the timeout test

wubzz:
It seems that when `.timeout` was added to query builder, it was not added to instances of `.raw`, that's my bad. This will add the same timeout functionality also for running `.raw` as non-bindings.

```javascript
knex.raw('SELECT something')
.timeout(1000)
.then(_.noop);
//Previously did not work. Now works as expected.
```

Also attempting to fix the timeout test in this PR as its previous implementation was very unreliable.

### DIFF
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -256,36 +256,50 @@ module.exports = function(knex) {
     });
 
 
-    it('.timeout() should throw TimeoutError', function() {
+    it('.timeout() should throw TimeoutError', function(done) {
       var dialect = knex.client.config.dialect;
-      var rows = _.times(dialect === 'sqlite3' ? 100 : 3000, function() {
-        return {
-          id: 1,
-          value: 'rO8F8YrFS6uoivuRiVnwrO8F8YrFS6uoivuRiVnwuoivuRiVnw'
-        };
-      });
+      if(dialect === 'sqlite3') { return done(); } //TODO -- No built-in support for sleeps
+      var testQueries = {
+        'postgres': function() {
+          return knex.raw('SELECT pg_sleep(1)');
+        },
+        'mysql': function() {
+          return knex.raw('SELECT SLEEP(1)');
+        },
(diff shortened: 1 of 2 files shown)


test/integration/builder/additional.js
```diff
@@ -256,36 +256,50 @@ module.exports = function(knex) {
...
-            .insert(rows)
-            .timeout(1);
-        })
+      if(!testQueries.hasOwnProperty(dialect)) {
+        throw new Error('Missing test query for dialect: ' + dialect);
+      }
+
+      var query = testQueries[dialect]();
+
+      return query.timeout(1000)
```
jurko-gospodnetic: Perhaps use a much shorter timeout here? Tests are synchronous and having ones that make pauses for timeout detection gets real irritating real fast.

test/integration/builder/additional.js
```diff
@@ -256,36 +256,50 @@ module.exports = function(knex) {
...
+        mssql: function() {
+          return knex.raw('WAITFOR DELAY \'00:00:05\'');
+        },
+        oracle: function() {
+          return knex.raw('dbms_lock.sleep(5)');
+        },
+        'strong-oracle': function() {
+          return knex.raw('dbms_lock.sleep(5)');
+        }
+      };
```
jurko-gospodnetic: Once you have this completely working, assuming this is the way this gets implemented, might be worth extracting this into an engine specific test utility layer so there is one place where such engine specific information needs to be tweaked for each engine instead of hunting all such cases through all the tests.

One other such thing that pops to mind is a way to say whether a specific engine reports its BEGIN & ROLLBACK statements as queries, i.e. using the `query` event, which `oracle` & `mssql` do not seem to do.

wubzz:
I've rerun the build for this 5 times now and it succeeds everytime, so I would say the test is more reliable now.

I'm going to merge this so that future Pull Requests aren't affected by the previously unreliable test.

PR #1261, opened 2016-03-08
Update documentation for table.foreign

raijinsetsu:


index.html
```diff
@@ -2049,9 +2049,10 @@ <h3 id="Schema-Building">Schema Building:</h3>
     </p>
 
     <p id="Schema-foreign">
-      <b class="header">foreign</b><code>table.foreign(column))</code>
+      <b class="header">foreign</b><code>table.foreign(~mixed~)</code>
```
wubzz: Could we change this to plural columns instead of `~mixed~` ? I know it doesn't make much sense seeing as it accepts both Array and Non-Array string literals, but I ask this to keep the consistency of the documentation as much as possible.

IE, `.dropForeign` right below states `columns`

raijinsetsu: Do you want to keep the remaining doc intact? It has both calls
demonstrated.

wubzz: I'm good with the examples that are written in the PR.

wubzz:
@raijinsetsu Thanks!

Issue #1256, opened 2016-03-08
A stack trace with no client information is not useful


jdmarshall:
I'm trying to write my first migration, and I got this error from knex:

```
Knex:warning - migrations failed with error: Cannot read property 'call' of undefined
TypeError: Cannot read property 'call' of undefined
    at TableBuilder.toSQL (/root/api/node_modules/knex/lib/schema/tablebuilder.js:36:11)
    at SchemaCompiler.createTable (/root/api/node_modules/knex/lib/schema/compiler.js:58:19)
    at SchemaCompiler.toSQL (/root/api/node_modules/knex/lib/schema/compiler.js:45:26)
    at SchemaBuilder.toSQL (/root/api/node_modules/knex/lib/schema/builder.js:43:43)
    at /root/api/node_modules/knex/lib/runner.js:35:32
From previous event:
    at Runner.run (/root/api/node_modules/knex/lib/runner.js:30:20)
    at SchemaBuilder.Target.then (/root/api/node_modules/knex/lib/interface.js:27:43)
    at processImmediate [as _immediateCallback] (timers.js:383:17)
```

There is not a single line number, file name or variable name from any of my code anywhere in that error message, so I have no actionable information I can use to fix this problem.  

Aside from using the process of elimination (and having to restore my database repeatedly while I debug the migration) what are my options?

rhys-vdw:
> Aside from using the process of elimination (and having to restore my database repeatedly while I debug the migration) what are my options?

Migrations are run in a transaction, so you shouldn't have to restore your database if something backfires.

You have two options (besides the process of elimination):
 - Share migration code here or on the knex gitter chat and rely on someone in the community to tell you what you're doing wrong.
 - Look into the stack trace you have, work out what you were doing wrong, and then submit a PR improving error handling for that particular misuse.

Obviously the latter would be preferable, but I'll have a look if you post it.

wubzz:
I see this way too often. This is caused by not supplying a valid callback function for the table builder. *(knex.schema.createTable(tableName, undefined), knex.schema.table(tableName, undefined)*

I agree there should be a better error handling in the tablebuilder when the function is omitted from the arguments. Either that, or silently ignore by `noop`, but I would prefer an error being thrown.

PR #1270, opened 2016-03-11
Add TypeError if missing callback in Schema .createTable/.table

wubzz:
Seen quite a few issues caused by missing a callback function, and the stack generated isn't very specific. The documentation states a callback is to be supplied, and while it's easy to forget, this will at least ease up the confusion.

See example #1256


elhigu:
Awesome :)

PR #1274, opened 2016-03-13
Corrects an annoying typo

tbranyen:
Quick fix


wubzz:
@tbranyen Thanks for this, but it has already been merged to master:
https://github.com/tgriesser/knex/blob/master/index.html#L187

tbranyen:
Shouldn't it be in gh-pages so the site is updated?

PR #1277, opened 2016-03-14
Fix a few hyperlinks in the changelog

joepie91:
The hash (#) was missing for two hyperlinks. This fixes that.


blah238:
Thanks @joepie91!

PR #1279, opened 2016-03-15
Update bluebird dependency 2.9.24 -> 3.3.4. Fixes #1216

### LINKED ISSUE #1216: Update bluebird to 3.x
Bookshelf has bluebird 2.x and promises produced by it don't provide bluebird features, such as: 

`.catch() now supports an object predicate as a filter: .catch({code: 'ENOENT'}, e => ...).`
http://bluebirdjs.com/docs/new-in-bluebird-3.html

ref. tgriesser/bookshelf#1137

wubzz:


wubzz:
Note to self: May need to add `reflect` [here](https://github.com/tgriesser/knex/blob/master/src/interface.js#L75).

src/transaction.js
```diff
@@ -229,7 +229,7 @@ function makeTxClient(trx, client, connection) {
     })
   }
   trxClient.acquireConnection = function() {
-    return Promise.settle([trx._previousSibling]).then(function () {
+    return Promise.all([trx._previousSibling].map((p) => { p.reflect() })).then(function () {
```
elhigu: Here probably you should return promise in map?

`.map((p) => { return p.reflect() })` or `.map((p) => p.reflect())` otherwise you will get `Promise.all([undefined])` here.

So maybe 

`return Promise.all([trx._previousSibling].map(p => p.reflect())).then(function () { return connection; });`

Also since code already requires that `trx._previousSibling` is a Promise or otherwise p.reflect() would throw an error, so this could be further simplified:

`return trx._previousSibling.reflect().then(function () { return connection; });`

wubzz: Yeah that's a complete miss, my bad. Thought I had added 'return'

PR #899, opened 2015-07-16
wrap index names

maxcnunes:
The current implementation I have been using at work for multi tenants is based in different tables for each client (tenant). To distinguish each table for each client we use the client id (number) as prefix name. For example I have these tables for client 1 (1_authorization, 1_user) and these tables for client 2 (2_authorization, 2_user).

Due that I have tables starting with number values. And this pull request fixes the problem of executing queries with index names starting with number values.  Since I have been working only with Mysql and Postgres I have applied the changes only in those databases types.

PR #1289, opened 2016-03-17
Wrap index names

maxcnunes:
Pull request already described in this previous one #899.

The only difference of this one is that I have rebased the branch with the master.
And already included wrap index in the new mssql dialect.

I hope is good now for merge.

Thanks


elhigu:
Thanks!

PR #1290, opened 2016-03-17
Allow incrementing by float. Fixes auto-converting to integers.

### LINKED ISSUE #868: Increment rounds to integer when incrementing decimal or float field

avimar:
Fixes #868

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -742,7 +742,7 @@ assign(Builder.prototype, {
 
   // Helper for the incrementing/decrementing queries.
   _counter: function(column, amount, symbol) {
-    var amt = parseInt(amount, 10);
+    var amt = parseFloat(amount);
     if (isNaN(amt)) amt = 1;
     this._method = 'counter';
     this._single.counter = {
--- a/test/integration/builder/updates.js
+++ b/test/integration/builder/updates.js
@@ -75,6 +75,17 @@ module.exports = function(knex) {
       });
     });
 
+    it('should increment a float', function() {
+      return knex('accounts').select('logins').where('id', 1.5).then(function(accounts) {
+        return knex('accounts').where('id', 1).increment('logins').then(function(rowsAffected) {
+          expect(rowsAffected).to.equal(1);
+          return knex('accounts').select('logins').where('id', 1);
+        }).then(function(accounts2) {
+          expect(accounts[0].logins + 1.5).to.equal(accounts2[0].logins);
+        });
+      });
+    });
+
     it('should increment a negative value', function() {
       return knex('accounts').select('logins').where('id', 1).then(function(accounts) {
         return knex('accounts').where('id', 1).increment('logins', -2).then(function(rowsAffected) {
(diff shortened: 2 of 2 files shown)


rhys-vdw:
Hey @avimar, looks good. I was holding off merging #1069 because it did not provide test coverage. I know this is a fairly trivial change, but would you mind adding coverage in [`/test/integration/builder/updates.js`](https://github.com/tgriesser/knex/blob/c7a659a4501a5f8a1cf12e856edbf753bee1d568/test/integration/builder/updates.js#L67-L109)?

avimar:
Gosh. It fails still, but I think because it's tied into the schema that the column is an integer in `test/integration/schema/index.js`

It's set there as an integer several times, programatically and in text. I'm not sure what's going on with that...

wubzz:
@avimar Hey, any further luck on this one? Would be great to get this into 0.11.

avimar:
@wubzz @rhys-vdw I'm not really sure how the schema works -- the columns for increment/decrement tests is an integer.
I'm totally unfamiliar with the test setups.

wubzz:
@avimar I think it should be fine if you add a new `.float()` column to one of the tables in `test/integration/schema/index.js` for this test. Since this change directly affects float values, a float column is needed.

avimar:
@wubzz I'm sorry, I can't figure out how to run the tests on my windows computer and still can't undertsand even where to add it in `test/integration/schema/index.js` -- it looks like then I'll have to update the checks on the generated SQL, too.

I thought I'd "handle this since it's so simple" but it hasn't been...

wubzz:
@avimar Good point, I guess adding stuff to tables and tests that have been around for a long time can be a bit overwhelming, as many tests are affected. Seeing as it's a test suite I don't see the necessity to be delicate here - the important part is that it's being tested - so personally I don't mind even if a new table is used.

Perhaps you can opt for something like this instead? *(Untested, but should work)*

```javascript
it('increment & decrement on float types', function() {
      return knex.transaction((tr) => {
        let insert = () => tr('incrementstest').insert({value: 1}).returning('*');
        let increment = (amount) => tr('incrementstest').increment('value', amount).returning('*');
        let decrement = (amount) => tr('incrementstest').decrement('value', amount).returning('*');
        return tr.schema.dropTableIfExists('incrementstest')
        .createTable('incrementstest', (table) => {
            table.float('value');
          })
        .then(insert)
        .then((rows) => {
            expect(rows[0].value).to.equal(1);
            return increment(0.5)
          })
        .then((rows) => {
            expect(rows[0].value).to.equal(1.5);
            return decrement(1.5);
          })
        .then((rows) => {
            expect(rows[0].value).to.equal(0);
          })
      })
    });
```

avimar:
I'm sorry, I am unable to handle the time and emotions around continuing to make what feels like such a simple change.
Please, someone else deal with this...

elhigu:
@avimar no prob, thanks for the effort! I hope someone takes the ball with this one.If not, I'll be closing this pull request at some point and add info to #868 about initial implementation.

rkaw92:
Just posting a note here to say that sometimes converting to a float isn't an option either (we work with potentially large decimals using `big.js` and don't want precision loss).

elhigu:
@rkaw92 in that case `increment(col, knex.raw('1.234'))` is probably the solution.

PR #1292, opened 2016-03-17
update liftoff dependency so that coffeescript knexfiles and configurations are supported again.

light24bulbs:
I suggest adding coffeescript to test process if continued support is desired.


rhys-vdw:
Thanks for that @light24bulbs. :+1:

PR #1296, opened 2016-03-19
Default pool.ping fn, and silently ignore 'Rollback' errors on a dead connection

wubzz:
This should cover the proposal in #1198 .

This will add a default `ping` function to the default `pool` config issuing a `SELECT 1` query to ensure the connection is still alive. So
1. Run query
2. Kill connection
3. Run query
=== Should be resolved. Previously this was rejected with different errors based on the dialect in use.

@woopstar has however confirmed that this does not solve the mysql error in the issue, so this doesn't cover all cases yet. Will want to cover that case as well before merging this.

A timeout has been added to `rollback` to prevent knex from hanging when rollback is issued on a dead connection. There's no harm in silently ignoring this as the connection is dead and the transaction state cannot be recovered.

**TODO**
- [x] Update docs
- [x] Decide a reasonable timeout for `.rollback`
- [x] ~~Reproduce mysql error, try find a solution~~ The PR solves the issue aside for when it occurs during transactions, at which point recovery is impossible. *(Bound to a single connection)*)
- [ ] Is it possible to write tests for this?

So far I've tested that this works when closing connections in postgres.

CC @rhys-vdw @elhigu @tgriesser if you have any input or thoughts please do tell. I hope these changes makes sense.

### DIFF
--- a/index.html
+++ b/index.html
@@ -506,7 +506,14 @@ pg('table').insert({a: 'b'}).returning('*').toString();
     <h3 id="Installation-pooling">Pooling</h3>
 
     <p>
-      The client created by the configuration initializes a connection pool, using the <a href="https://github.com/myndzi/pool2">Pool2</a> library. This connection pool has a default setting of a <tt>min: 2, max: 10</tt> for the MySQL and PG libraries, and a single connection for sqlite3 (due to issues with utilizing multiple connections on a single file). To change the config settings for the pool, pass a <tt>pool</tt> option as one of the keys in the initialize block. Checkout the <a href="https://github.com/myndzi/pool2">Pool2</a> library for more information.
+      The client created by the configuration initializes a connection pool, using the <a href="https://github.com/myndzi/pool2">Pool2</a> library. This connection pool has a default setting of a <tt>min: 2, max: 10</tt> for the MySQL and PG libraries, and a single connection for sqlite3 (due to issues with utilizing multiple connections on a single file). To change the config settings for the pool, pass a <tt>pool</tt> option as one of the keys in the initialize block.
+    </p>
+    <p>
+      The default connection pool also applies a <tt>ping</tt> function to verify that a connection is still alive by running a <tt>SELECT 1</tt> query.
+      If you don't want this check you may supply your own ping function to overwrite knex's default behavior.
+    </p>
+    <p>
+      Checkout the <a href="https://github.com/myndzi/pool2">Pool2</a> library for more information.
     </p>
 
 <pre>
--- a/src/client.js
+++ b/src/client.js
@@ -200,6 +200,9 @@ assign(Client.prototype, {
         } else if (connection !== void 0) {
           client.destroyRawConnection(connection, callback)
         }
+      },
+      ping: function(resource, callback) {
+        return client.ping(resource, callback);
       }
     }
   },
--- a/src/dialects/maria/index.js
+++ b/src/dialects/maria/index.js
@@ -115,6 +115,10 @@ assign(Client_MariaSQL.prototype, {
       default:
         return response;
     }
+  },
+
+  ping: function(resource, callback) {
+    resource.query('SELECT 1', callback);
   }
 
 })
--- a/src/dialects/mssql/index.js
+++ b/src/dialects/mssql/index.js
@@ -178,6 +178,10 @@ assign(Client_MSSQL.prototype, {
       default:
         return response
     }
+  },
+
+  ping: function(resource, callback) {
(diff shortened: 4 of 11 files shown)


elhigu:
@wubzz 5 second timeout could be slow enough for now... and it could be set configurable if later on if needed (people might want e.g. < 1s timeouts there...).

wubzz:
I've changed it to 5 seconds for now. Also I managed to reproduce the mysql error in a way, and from my perspective, this solves every case except for transactions (which is bound to its single connection). I'll quote from the issue:

> I was finally able to reproduce this error setting connect_timeout, wait_timeout and interactive_timeout to 1 on my mysql server.
>
>COMMIT; - This socket has been ended by the other party
>
>I can confirm the ping solution works when not using transactions.
>
>Transactions are bound to their single connection and therefor ping is not being called on these. Because of this, it eventually throws that error. I don't see how knex can possibly recover from this and "rerun" the entire transaction using a new connection.
>
>From your original post, the error was
>Unhandled rejection Error: BEGIN; - This socket has been ended by the other party
>indicating that you are using a transaction (BEGIN;), however in your code example above I do not see a transaction being used. Can you post your code that causes this issue while inside a transaction?
>
>If you have a timeout of 600 seconds (10 minutes) on your mysql server, there should from my understanding be no way for a transaction to throw this error so long as the transaction is either commited or rolled back within those 10 minutes.

elhigu:
@wubzz Do you think there is any way to add tests for this, e.g. by replacing `_driver()` method of dialect with mock or something like that?

wubzz:
@elhigu This one is really difficult to test, even more so for cross-dialect testing. Right now I can only recreate it with some pretty obnoxious code for mysql:

```javascript
  return Promise.all([
        knex.raw('SET GLOBAL connect_timeout = 2'),
        knex.raw('SET GLOBAL wait_timeout = 1')
      ])
      .then(function() {
          return new Promise(function(resolve, reject) {
            setTimeout(function() {
              return knex.raw('SELECT 1').then(resolve).catch(reject);
            }, 3000); //3 sec > connect_timeout
          });
        })

//Throws
SELECT 1 - This socket has been ended by the other party
```

This test would:
- Only work for mysql _(can't find similar settings for other dialects)_
- Take at least 3 seconds to perform _(minimum value of 2 seconds on connect_timeout)_
- Update GLOBAL settings of the mysql instance _(could maybe reset them after test succeeds/fails)_

Using this test I've verified that the pool changes fixes this, aside for transactions. But I really don't think we can do anything about the transactions, as stated above.

wubzz:
Current hypothesis regarding the mysql2 issue is that #1315 will fix this. Currently errored mysql2 connections dont seem to be disposed at all?

As for tests on this, I recommend against it due to the circumstances posted here earlier.

elhigu:
I got another idea how to possibly test connection closed by the server on postgresql. 

Maybe one could start transaction on one connection on pool, then run separate query outside of the transaction, which commands postgres server to close all connections with query like this http://stackoverflow.com/questions/5108876/kill-a-postgresql-session-connection and then try to do more stuff in the transaction and see how things are failing...

Or exploit some other way functions of explicitly killing connections from DB... also mysql has similar support for killing connections with some magic queries.

wubzz:
Just spent another hour trying to write reliable tests for this -- with no success. Using postgres functions to kill connections is somewhat considered a "gentle" way of doing it, and does not throw the same types of errors, so it feels like it's not really testing what it should be testing. I even tried manipulating the connection's socket in various ways to produce a similar error, but was unable to.

I'm going to merge this the way it is, pretty harmless changes considering. Can't have gotten worse than it already was. :smirk:

AiNoKame:
Hi @wubzz , is there something I'm missing when connecting to a mssql database (using the master branch)? I'm seeing this when making a knex connection:

```
Knex:Error Pool2 - TypeError: resource.query is not a function
Knex:Error Pool2 - TypeError: resource.query is not a function
Knex:Error Pool2 - TypeError: resource.query is not a function
...
```
`resource.query` is only found in this PR (I also changed my connection object key from `server` to `host` and using a url, not a connection string, for that), so I was hoping this was the right place to ask

wubzz:
@AiNoKame Looks like the `ping` function for mssql isn't working. I would have expected tests to find this issue, but just realized that integration tests for mssql are not being run in travis.. Will need to set up an environment for this locally in the future.

For now I've changed the ping function from `resource.query` to `resource.request().query` according to the [mssql library docs](https://www.npmjs.com/package/mssql) which should work. #1343

AiNoKame:
Thanks! :smile_cat:

PR #1308, opened 2016-03-24
DateTime and TimeStamps NULL should remain NULL

### LINKED ISSUE #1276: If it's NULL, let it be NULL
I have NULL data in DATE column. But when fetching using bookshelf (using knex), it's return `Thu Jan 01 1970 07:00:00 GMT+0700 (SE Asia Standard Time)` equal to `new Date(null)` or `new Date(0)`.

After checking knex@0.10.0, I found `parseType` function:

```
// file: .\knex\lib\dialects\maria\index.js
// line: 117
function parseType(value, type) {
  switch (type) {
    case 'DATETIME':
    case 'TIMESTAMP':
      return new Date(value);
    case 'INTEGER':
      return parseInt(value, 10);
    default:
      return value;
  }
}
```

In regards to NULLABLE column type. IMHO, it's should respect the NULL data.

But with `parseType` function above, if the data is NULL will produce:
- start of javascript Date Object for DATETIME and TIMESTAMP
- and `NaN` for Integer

wubzz:
Fixes #1276

### DIFF
--- a/src/dialects/maria/index.js
+++ b/src/dialects/maria/index.js
@@ -120,6 +120,9 @@ assign(Client_MariaSQL.prototype, {
 })
 
 function parseType(value, type) {
+  if(value === null) {
+    return value;
+  }
   switch (type) {
     case 'DATETIME':
     case 'TIMESTAMP':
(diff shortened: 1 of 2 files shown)


kibertoad:
@wubzz Any reason why this was closed without merging?

kibertoad:
Or was it only relevant for maria and mysql never had this issue in the first place?

wubzz:
See original issue for reasons.

PR #1313, opened 2016-03-26
Minor stylistic code & comment touch up + typo fixes

jurko-gospodnetic:
Not a global change - only affects a couple of modules I looked at while tracking down some migration problems.


wubzz:
Thanks for this!

PR #1315, opened 2016-03-29
Mysql2 should also listen to 'error' events.

wubzz:
Briefly noticed in #1198 there's a diff between how `mysql` and `mysql2` handle events, yet they're both almost identical. I can't find an `end` event in `mysql2`, but the `error` is there *(with the appropiate "fatal" flag)* so unless I'm missing something it should be listened to here as well.

https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L96

@elhigu @rhys-vdw Does this make sense to you?


wubzz:
I've debugged this further and it is in fact the case. Connections aren't being disposed using the "MySQL Specific handler" when specified driver is mysql2. This fixes that problem.

Issue #1180, opened 2016-02-02
npm install from master broken


intensite:
Hello,

It appears that the change to remove the lib directory is causing issues when knex is installed from master using npm.

```
PS C:\projets\test_mssql> npm install
npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install"
npm ERR! node v4.2.3
npm ERR! npm  v2.14.7
npm ERR! path C:\projets\test_mssql\node_modules\knex\lib\bin\cli.js
npm ERR! code ENOENT
npm ERR! errno -4058
npm ERR! syscall chmod

npm ERR! enoent ENOENT: no such file or directory, chmod 'C:\projets\test_mssql\node_modules\knex\lib\bin\cli.js'
npm ERR! enoent This is most likely not a problem with npm itself
npm ERR! enoent and is related to npm not being able to find a file.
npm ERR! enoent

npm ERR! Please include the following file with any support request:
npm ERR!     C:\projets\test_mssql\npm-debug.log
```

As you can see this is from a Windows installation.  I am using the version from the master repo as I need MSSQL support.

Can this commit be the cause: #1149

elhigu:
This seems to be the problem https://github.com/tgriesser/knex/blob/master/package.json#L59 npm tries to install knex client to `.bin`before `prepublish` script is ran.

intensite:
Is there a quick fix for the time being?

elhigu:
You could probably fork the repo and remove the `bin` attribute from `package.json` in your own branch. We are resolving if this is problem of npm or should us write work around for it.

rhys-vdw:
Tracking issue:  npm/npm#11395.
Thanks @elhigu

rhys-vdw:
So it looks like the `build` command is not even being run, so it may not be an npm problem after all.

rhys-vdw:
Okay, so I've been looking at this for over three hours and I'm going to leave this for a bit. Seems that `prepublish` is only called when you explicitly run `npm install` in the actual `knex` directory. (eg. `/myProject/node_modules/knex`.)

We can get `build` running automatically in `postinstall`, but definitely don't want this happening when installing from the npm repository, because it will require installing babel.

There does not seem to be a trivial way to do this cleanly.

[...]

PR #1316, opened 2016-03-29
Fix issue tgriesser/knex#1180

maxcnunes:
Fix the problem on installing project from the github.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -1,0 +1,3 @@
+#!/usr/bin/env node
+
+exports = require('../lib/bin/cli');
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "prepublish": "npm run build"
   },
   "bin": {
-    "knex": "./lib/bin/cli.js"
+    "knex": "./bin/cli.js"
   },
   "repository": {
     "type": "git",
@@ -94,6 +94,7 @@
   },
   "files": [
     "README.md",
+    "bin/*",
     "src/*",
     "lib/*",
     "build/*",


maxcnunes:
Actually it still not working. Because it allow install the project without that error of `cli.js` not found. But since the build is only applied during the pre-publish. The `lib` files are not available on installing from github.
I don't like the idea of building the lib during the postinstall. Because it  requires installing "useless" dependencies for production environment such as babel. And it makes the installing slower.
There is a proposal for NPM [include a prepublish for Git](https://github.com/npm/npm/issues/3055). But I guess it will not be implemented.
Seems most projects doesn't provide the build files on installing from git. So I guess the best option for #1180 is just not provide the lib on git. And anyone installing from git, changes their own fork-branch to include the lib files.

elhigu:
I think that this change would still be good. This way one can install from github and then add post install scripts to own project which builds the knex if he really wants to use something from github. To me this fix would be enoguh to close #1880.

elhigu:
This would also need manual testing since nothing in test suite is actually running the `cli.js` client installed by npm scripts.

maxcnunes:
Nowadays installing from github does not include the scripts folder. Which has the `build.sh`. I guess because is not included into the [files](https://github.com/tgriesser/knex/blob/master/package.json#L95) property. So to allow users installing from github and later building the project I think would be good include the scripts folder as well.
We could use `.npmignore` to ignore it and even the `src` folder on installing from npm. I guess the `.npmignore` is only applied on publishing to npm. So would not affect those installing from github.
@elhigu  I can test these ideas and apply them if you think they are a good approach for the problem.

elhigu:
@maxcnunes Sounds good. I don't know much about npm packaging / so I cannot anticipate if that really is viable, but would be great if you can test it :+1:

vellotis:
https://docs.npmjs.com/files/package.json#files

> ### files 
>
> The "files" field is an array of files to include in your project. If you name a folder in the array, then it will also include the files inside that folder. (Unless they would be ignored by another rule.)
> 
> You can also provide a ".npmignore" file in the root of your package or in subdirectories, which will keep files from being included, even if they would be picked up by the files array. The **.npmignore** file works just like a **.gitignore**.

So package.json specification confirms @maxcnunes words.

I was thinking if there could be a way to build (+ babel installing) with `postinstall` if `lib` directory is missing. This way it is not needed to explicitly invoke building from depending project.

#1425 would be a great improvement along with this one.

maxcnunes:
Yeah. The solution I was thinking was exactly this one using `postinstall`. Although I don't have time to implement this right now. Please, feel free anyone to take over.

vellotis:
I will look into it in few days.

tgriesser:
So the latest 0.12 should allow you to install from github, but for now you'll also need to configure a script yourself that goes into the knex directory and runs `npm install` & `npm run build`

vellotis:
@tgriesser 
There is no need for separate script that goes into the knex directory and runs ``npm install & npm run build`. It can be handled the same way as solved in https://github.com/tgriesser/bookshelf/pull/1403

tgriesser:
Yeah, I guess I was just trying to make it possible but discouraged to install from github. But yeah, `buildDependencies` seems like a good compromise. Feel free to open a PR.

PR #1326, opened 2016-04-01
.renameColumn should not drop defaultValue or nullable state.

### LINKED ISSUE #933: .renameColumn drops column schema (default specifically)
Thank you as always for the awesome library!

It looks like .renameColumn ( http://knexjs.org/#Schema-renameColumn ) drops any additional information that a column has when you use it.

I'm assuming that's because the library doesn't have the state of the column while modifying and what is happening is that an `ALTER TABLE MODIFY COLUMN before after` is being done without re-sending the additional data like: `ALTER TABLE MODIFY COLUMN before after TINYINT(1) DEFAULT 1` ?

Example migration:
```javascript
  return knex.schema
    .createTable('throwaway', function(table) {
      table.boolean('has_default').defaultTo(true);
    })
    .then(function() {
      return knex
        .raw('SHOW CREATE TABLE throwaway')
        .then(function(results) {
          // shows that it has a default of `1`
          console.log('results', results);
        });
    })
    .then(function() {
      return knex.schema.table('throwaway', function(table) {
        table.renameColumn('has_default', 'missing_default');

        // default is now NULL after this
      });
    })
    ;
};
```

wubzz:
Currently this happens for mysql. Fixes #933

### DIFF
--- a/src/dialects/mysql/schema/tablecompiler.js
+++ b/src/dialects/mysql/schema/tablecompiler.js
@@ -77,8 +77,17 @@ assign(TableCompiler_MySQL.prototype, {
               if (!refs.length) { return; }
               return compiler.dropFKRefs(runner, refs);
             }).then(function () {
+              let sql = `alter table ${table} change ${wrapped} ${column.Type}`;
+
+              if(String(column.Null).toUpperCase() !== 'YES') {
+                sql += ` NOT NULL`
+              }
+              if(column.Default !== void 0 && column.Default !== null) {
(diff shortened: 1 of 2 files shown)


wubzz:
cc @gabe-alex I'm no mysql expert, would you mind double-checking that I got this right?

gabe-alex:
@wubzz Right, seems to work now.
Although, I had to manually copy the changes to my lib folder, since I'm on Windows and building doesn't seem to work...

wubzz:
Cool, merging this then. It'll be in the next release.

Issue #46, opened 2013-08-23
Modifying columns


ErisDS:
I think it is a reasonable expectation that it would be possible to:
- rename a column
- change the data type
- add or drop the null constraint
- change the default 

And probably other things that I have not thought of. 

I would imagine the syntax for doing this would be something like

```
 knex.Schema.table('table_name', function (t) {
        t.string('my_column').renameTo('something_else');
        t.string('my_column').changeTo('text');
        t.string('my_column').nullable() < adds nullable if not already present
        t.string('my_column').notNullable() < removes nullable if present
        t.string('my_column').defaultTo('whatever') < adds or updates the default
  });
```

Maybe there needs to be something extra in the chain, possibly at the `knex.Schema.table` level to indicate that this is a modify statement not an add.

I realise that this is tricky to implement across various databases, especially in SQLite where you have to create a whole new table, but in a system which requires migrations, without these tools it is going to be necessary to use knex.raw and write all the migrations for each DB supported, which sort of defeats the point of having a nice ORM especially one which is about to support migrations.

dkushner:
@tgriesser, @bendrucker: Has there been any update on this functionality? Writing raw queries in every migration I have that modifies a column (virtually all of them), is getting a bit frustrating.

bendrucker:
No, sorry

ricardograca:
Would it be possible to modify the `renameColumn()` method so that it returns the modified column object so that one can chain any of the column chainable methods to it? Namely I'm trying to update a reference after renaming a column:

    table.renameColumn('stuff_id', 'some_stuff_id').references('some_stuff.id')

This produces an error saying that:

    Object #<TableBuilder_PG> has no method 'references'

If that would be a breaking change (looks like it, although I'm not sure if anyone is using the current return object) then at least provide a way to select a column that already exists.

ErisDS:
Sorry to effectively be nagging on an ancient issue - but I'm desperate to be able to modify the length of text columns. I'd be more than happy to pitch in and try to contribute this if there's a consensus on what the API should be?

olalonde:
Is it yet possible to modify column types or should I use `.raw()`?

ricardograca:
@olalonde Nope, you should use `.raw()`.

tgriesser:
@ErisDS the plan is to make the builder not constrained to normal statements but also add an api for altering the tables:

```js
knex.alterTable(tableName).modifyColumn(colName, type)
[...]

PR #1327, opened 2016-04-02
Implement schema .setNullable/.dropNullable. Fixes #1218

### LINKED ISSUE #1218: How to change nullable constraint in column inside a migration?
I am trying to change a contraint on a column but this just appears to try to re-create the column and fails.

``` js
  return knex.schema.table("images", function(table) {
    table.string("checksum").notNullable();
  });
};

  return knex.schema.table("images", function(table) {
    table.string("checksum").nullable();
  });
};
```

wubzz:
~~This adds `.setNullable(column, [true|false])` to alter a columns nullable constraint.~~
This adds `.setNullable(column)` and `.dropNullable(column)` to alter a columns nullable constraint.

I'm not too happy with the implementation having to call `SELECT 1`, but this was the only way I found to support async actions, which are required in this case. See for example `renameColumn` for a similar case.

This is built around the information gathered in #1218 , so feel free to correct if anything is wrong.

### DIFF
--- a/src/schema/tablecompiler.js
+++ b/src/schema/tablecompiler.js
@@ -205,4 +205,28 @@ TableCompiler.prototype._indexCommand = function (type, tableName, columns) {
   return this.formatter.wrap(indexName);
 };
 
+
+//Default implementation of setNullable. Overwrite on dialect-specific tablecompiler when needed
+//(See postgres/mssql for reference)
+TableCompiler.prototype.setNullable = function(column, nullable) {
+  let tableName = this.tableName();
+  let columnName = this.formatter.columnize(column);
+  return this.pushQuery({
+    sql: 'SELECT 1',
+    output: () => {
+      return this.client.queryBuilder().from(this.tableNameRaw).columnInfo(column)
+        .then((columnInfo) => {
+          if(isEmpty(columnInfo)) {
+            throw new Error(`.setNullable: Column ${columnName} does not exist in table ${tableName}.`)
+          }
+          let nullableType = nullable ? 'null' : 'not null';
+          let columnType = columnInfo.type + (columnInfo.maxLength ? `(${columnInfo.maxLength})` : '');
+          let defaultValue = (columnInfo.defaultValue !== null && columnInfo.defaultValue !== void 0) ? `default '${columnInfo.defaultValue}'` : '';
+          let sql = `alter table ${tableName} modify column ${columnName} ${columnType} ${nullableType} ${defaultValue}`;
+          return this.client.raw(sql);
+        });
+    }
+  });
+};
+
 module.exports = TableCompiler;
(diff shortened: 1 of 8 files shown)


wubzz:
TODO

- [ ] Probably change this to `.dropNullable(column)` / `.setNullable(column)` instead
- [ ] Try find different approach for the promises
- [ ] Research more on sqlite

src/schema/tablecompiler.js
```diff
@@ -205,4 +205,37 @@ TableCompiler.prototype._indexCommand = function (type, tableName, columns) {
...
+    output: () => {
+      return this.client.queryBuilder().from(this.tableNameRaw).columnInfo(column)
+        .then((columnInfo) => {
+          if(isEmpty(columnInfo)) {
+            throw new Error(`.setNullable: Column ${columnName} does not exist in table ${tableName}.`)
+          }
+          let nullableType = nullable ? 'null' : 'not null';
+          let columnType = columnInfo.type + (columnInfo.maxLength ? `(${columnInfo.maxLength})` : '');
+          let defaultValue = (columnInfo.defaultValue !== null && columnInfo.defaultValue !== void 0) ? `default '${columnInfo.defaultValue}'` : '';
+          let sql = `alter table ${tableName} modify column ${columnName} ${columnType} ${nullableType} ${defaultValue}`;
```
elhigu: There is quite a lot copy-paste between this and mssql implementation... could "modify column" string be moved to separate variable and just that variable would be overridden in mssql dialect?

test/integration/schema/index.js
```diff
@@ -467,5 +467,44 @@ module.exports = function(knex) {
       });
     });
 
+    if(knex.client.dialect !== 'sqlite3') {
```
elhigu: Unit test for sqlite3 failure would be good to have.

wubzz:
@elhigu Good points, fixed! :+1: 

@tgriesser I saw #46 earlier today and wish I had seen it earlier. You probably notice this code is very similar to `renameColumn`. Do you want to hold of on this PR for now given your comments in that issue?

rhys-vdw:
@wubzz This is consistent with `renameColumn`.

I could see the appeal of something more chainable:

```js
table.changeColumn('some_column').null(false).default('empty').unique(false);
```

I think we can merge this though. What do you think?

wubzz:
@rhys-vdw Yeah I read #46 after doing this PR, and I agree a lot of what is mentioned in that issue can and would be better chainable. Would also allow for less duplicated code.

I think we should merge it for now, and then move the code out later on if we end up doing some alter-column specific chaining.

nrempel:
Any update on this pull request? This feature would be very nice to have.

mindworker:
@nrempel agree here. Any updates? Looking forward for a merge.

wubzz:
Leaving the decision up to @tgriesser on this one, considering #46 .

nabilfreeman:
Hi guys! any updates here?

elhigu:
I think this was implemented different way by .alter() column syntax

PR #1328, opened 2016-04-02
This allows the user to set their own lock table name

krainboltgreene:


krainboltgreene:
Point me to where I neeed to change docs, if this is accepted.

elhigu:
@krainboltgreene change looks good, what is missing is the documentation to index.html and test case to make sure that feature really works.

krainboltgreene:
@elhigu Any chance you could write the test or at least point me to a recent pull request that deals with this area and has a test?

elhigu:
@krainboltgreene You can start looking to tests here: https://github.com/tgriesser/knex/blob/master/test/integration/suite.js

Thats where migration tests are getting their configuration and knex instance is passed to here: https://github.com/tgriesser/knex/blob/master/test/integration/migrate/index.js

Where you should be able to add tests case for changing the migration lock table name. 

Even better would be if you can find somewhere a test where changing name of the migration table is tested it would be pretty perfect example. I'm just not sure if that has been tested at all :(

Hope that these pointers gets you started.

kibertoad:
@elhigu Would this change be still good to go if I added test and documentation?

elhigu:
Yep, it would be

kibertoad:
why was this deleted? it just missed tests

krainboltgreene:
I don't care anymore.

PR #1342, opened 2016-04-11
Correct connection URL parsing for MSSQL

h0vhannes:
Use connection.server instead of connection.host when connecting to MSSQL

### DIFF
--- a/src/util/parse-connection.js
+++ b/src/util/parse-connection.js
@@ -37,7 +37,11 @@ function connectionObject(parsed) {
     connection.database = db
   }
   if (parsed.hostname) {
-    connection.host = parsed.hostname;
+    if (parsed.protocol.indexOf('mssql') === 0) {
+      connection.server = parsed.hostname;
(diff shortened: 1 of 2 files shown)


rhys-vdw:
Thanks. Would you be able to add test coverage for this?

h0vhannes:
Sure, just don't know where to add tests for this. Is it integration or, maybe, tape ?

rhys-vdw:
Thanks. If there are none already then I tape is preferable for new tests.

h0vhannes:
Ok, there are tests for connection URL parser in tape, so I'll add tests there.

wubzz:
Good catch on this. Just to verify, was this done because it wasn't working properly in 0.10? Given #1239 the initiation should have worked regardless of this. But I do agree this was needed as well.

h0vhannes:
Yes, 0.10 doesn't work with MSSQL connection urls

Issue #948, opened 2015-08-25
Catching stream errors


danielgindi:
I'm trying to use `stream()`, and catch sql errors, but it seems that there's no way to do that.

Say that:

``` javascript
db.select('*').from('unknown_table').stream(function (stream) {

    console.log('This line is reached'); /// <--- Stream is received

    stream.on('error', function (err) {
        console.log('This line is not reached'); /// <--- Seems as a natural place to handle stream errors
    });

}).catch(function (err) {
    console.log('This line is also not reached'); /// <--- Seems as a natural place to handle sql errors
});
```

I do not receive any callback with the error. Instead this is the log:

``` cli
> This line is reached
> Error: ER_NO_SUCH_TABLE: Table 'unknown_table' doesn't exist
```

danielgindi:
@tgriesser any ideas on how to catch these errors?

id0Sch:
so this is still happening, im using mysql syntax.
any idea how to catch these errors?

bertho-zero:
I have the same issue, I have an uncaughtError but my .on('error') isn't triggered.

novemberborn:
I think the problem is in this snippet:

https://github.com/tgriesser/knex/blob/f8989e129fe5863437d8e87fdd6a1e1971841305/src/runner.js#L81:L105

If `ensureConnection()` fails, and no `handler` was provided, the promise rejection is swallowed. Instead it should be emitted on `stream`.

For now, a workaround might look something like this:

```js
let stream
const streamDone = knex()
  // query code
  .stream(streamValue => {
    stream = streamValue
  })

streamDone.catch(err => stream.emit('error', err))

// Now use `stream`
```

novemberborn:
Here's a PR: https://github.com/tgriesser/knex/pull/2199

PR #1345, opened 2016-04-13
propagate pg stream errors

kturney:
fixes https://github.com/tgriesser/knex/issues/948


kturney:
a downside I'm not sure how to address: the error causes Bluebird to log an unhandled exception message if the consuming code is only listening to the stream.

rhys-vdw:
Looks good. These events should be propagated.

> the error causes Bluebird to log an unhandled exception message if the consuming code is only listening to the stream.

Without using the callback argument you mean?

kturney:
@rhys-vdw yes, the log shows up when using the returned stream. You can see an example in the test run: https://travis-ci.org/tgriesser/knex/jobs/122960719#L1630

tgriesser:
Sounds like a good change, if you want to resolve the conflicts I will merge and get this into the next minor bump... If not let me know and I can add this manually

kturney:
build failed because of timeouts, but I'm not seeing a retry/rebuild button

elhigu:
Fixed in https://github.com/tgriesser/knex/pull/2199

PR #1389, opened 2016-05-07
Prevent crash on `seed:make`

nixgadget:
- `template` no longer accepts 3 arguments. Fixed by removing `null`


rhys-vdw:
Thanks @nixgadget

PR #1390, opened 2016-05-08
Support multi-column `whereIn` with query

### LINKED ISSUE #1384: Inconsistency in handling of arguments to `whereIn`
``` js
knex('outer').whereIn('c', knex('inner').select('a')).toString();
// "select * from "outers" where "c" in (select "a" from "inner")"

knex('outer').whereIn(['c', 'd'], [[1, 2], [3, 4]]).toString();
// "select * from "outer" where ("c", "d") in ((1, 2),(3, 4))"

knex('outer').whereIn(['c', 'd'], knex('inner').select('a', 'b')).toString();
// "select * from "outer" where ("c", "d") in (())" <--- Incorrect
```

rhys-vdw:
Add support for supplying a subquery for a `whereIn` statment on multiple columns.

Fixes #1384.


wubzz:
@rhys-vdw Looks good!

Perhaps we can use this new formatter function for non-string literals in `.select()` / `.from()` as well which has been brought up before. Currently doing fancy stuff with those two functions requires some hacky workarounds with `raw`.

src/formatter.js
```diff
@@ -26,7 +27,7 @@ assign(Formatter.prototype, {
   // a "joining" value is specified (e.g. ' and ')
   parameterize(values, notSetValue) {
     if (typeof values === 'function') return this.parameter(values);
-    values  = Array.isArray(values) ? values : [values];
+    values  = isArray(values) ? values : [values];
```
wubzz: Not sure I follow on this change. `const { isArray } = Array;` aren't you then constructing a new Array with values here? Same as `Array(null)` `Array(void 0)` `Array([1,2,3])`. Would mean the ternary operation **cant** fail?

Or am I just missing something obvious?

rhys-vdw: It is actually no change. That translates to:

```js
const { isArray } = Array;
// compiles to:
var isArray = Array.isArray;
```

See [object destructuring](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).

rhys-vdw:
I'm holding off merging this for now. I want to test performance of `Formatter#values` and also use it function in other relevant sections.

elhigu:
Looks like this has been already implemented https://runkit.com/embed/f2wym1fwfrn1 ?

elhigu:
I tried to fix these conflicts in git web editor... needs reviewing again if it still works as intended.

wubzz:
Updated the failing tests here (was due to 'default' old assertion type) and added assertions to remaining dialects.

elhigu:
Thanks!

PR #1392, opened 2016-05-09
BatchInsert remake proposal. Fixes #1391

### LINKED ISSUE #1391: Additions to batchInsert
Two additions that can be useful for BatchInsert and can be implemented without any harm:

- [ ] Supply an already started transaction instead of having BatchInsert create a new one
- [ ] Option to return inserted IDs / rows / Columns via `.returning`

wubzz:
Two things were requested for `batchInsert`:
- [x] Returning columns from inserted rows
- [x] Supply an outside transaction instead of auto-creating

I've added these two things to the current implementation.

I also decided to rewrite the whole procedure and move it out to a seperate file. I feel that it's very likely this function will eventually be ripped out of the knex repository into a seperate module at some point. Like a utility library of sorts. *(We've talked about this before)*

Not entirely satisfied with how complex-looking this simple thing became, and may have missed something. Please have a look  @rhys-vdw @elhigu !

### DIFF
--- a/src/util/batchInsert.js
+++ b/src/util/batchInsert.js
@@ -19,6 +19,7 @@ export default class BatchInsert {
     this._returning       = void 0;
     this._transaction     = null;
     this._autoTransaction = true;
+    
   }
 
   /**
--- a/src/util/batchInsert.js
+++ b/src/util/batchInsert.js
@@ -1,0 +1,84 @@
+'use strict';
+
+import {isNumber, isString, isArray, chunk, flatten} from 'lodash';
+import Promise from '../promise';
+
+export default class BatchInsert {
+	constructor (client, tableName, batch, chunkSize = 1000) {
+		if (!isNumber(chunkSize) || chunkSize < 1) {
+			throw new TypeError(`Invalid chunkSize: ${chunkSize}`);
+		}
+
+		if(!isArray(batch)) {
+			throw new TypeError(`Invalid batch: Expected array, got ${typeof batch}`)
+		}
+
+		this.client       = client;
+		this.tableName    = tableName;
+		this.batch        = chunk(batch, chunkSize);
+		this._returning   = void 0;
+		this._transaction = null;
+		this._autoCommit  = true;
+	}
+
+	/**
+	 * Columns to return from the batch operation.
+	 * @param returning
+	 */
+	returning (returning) {
+		if(isArray(returning) || isString(returning)) {
+			this._returning = returning;
+		}
+		return this;
+	}
+
+	/**
+	 * User may supply their own transaction.
+	 * If this is the case, don't autoCommit their transaction. The responsibility falls on the user.
+	 * @param transaction
+	 */
+	transacting (transaction) {
+		this._transaction = transaction;
+		this._autoCommit = false;
+		return this;
+	}
+
+	then (callback = function() {}) {
+		let transaction;
+		return Promise.resolve()
+		.then(() => {
+				if(this._transaction) {
+					transaction = this._transaction;
+					return Promise.resolve();
+				}
+				return new Promise((resolve) => {
+					this.client.transaction((tr) => {
+						transaction = tr;
+						resolve();
+					});
+				});
+			})
+		.then(() => {
+				return Promise.all(this.batch.map((items) => {
+					let promise = transaction(this.tableName)
+						.insert(items);
+					if(this._returning) {
+						promise.returning(this._returning);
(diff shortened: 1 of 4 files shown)


src/util/batchInsert.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+import {isNumber, isString, isArray, chunk, flatten} from 'lodash';
+import Promise from '../promise';
+
+export default class BatchInsert {
```
elhigu: I think we've been using 2 spaces indentation usually :)

wubzz: Strange, I've been using x2 tabs all along. Does look weird in this PR tho I agree. Changed it now. *(Does not appear updated in the diff view? https://github.com/tgriesser/knex/pull/1392/commits/65af6a8f39b054f2cbf27118349cdc402d2108ac)*

src/util/batchInsert.js
```diff
@@ -0,0 +1,84 @@
...
+	 */
+	transacting(transaction) {
+		this._transaction = transaction;
+		this._autoCommit  = false;
+		return this;
+	}
+
+	then(callback = function() {}) {
+		let transaction;
+		return Promise.resolve()
```
elhigu: Might look more clean if you just wrap getting the transaction to separate promise and resolve it. Something like: 

```
let transactionForBatches = .... promise for getting the transaction ...;

return transactionForBatches.then(trx => {
  // here you can add the rest of promise chain and no need to store transaction to temporary variable
});
```

src/util/batchInsert.js
```diff
@@ -0,0 +1,84 @@
...
+				return new Promise((resolve) => {
+					this.client.transaction((tr) => {
+						transaction = tr;
+						resolve();
+					});
+				});
+			})
+			.then(() => {
+				return Promise.all(this.batch.map((items) => {
+					let promise = transaction(this.tableName)
```
elhigu: This variable is not a promise, but query builder... maybe name could be `insertBatchQuery` or something like that.

elhigu:
Implementation looks correct to me, I don't see any errors in logic / error handling there. I'm not big fan of this feature in its current form, since it relays on input data being in big array. It is not scalable. With this change memory implications will be even bigger that with earlier implementation since now in worst case input *and* results are stored in huge arrays filling the RAM.

However this might be useful for someone who are not working with too big datasets so it doesn't matter much to me.

For relieving the memory pressure would be nice if implementation would take callback to fetch more data to insert / or could work with javascript iterables in a way that everything doesn't have to be in memory at once.

src/util/batchInsert.js
```diff
@@ -0,0 +1,84 @@
...
+        return new Promise((resolve) => {
+          this.client.transaction((tr) => {
+            transaction = tr;
+            resolve();
+          });
+        });
+      })
+      .then(() => {
+        return Promise.all(this.batch.map((items) => {
+          let promise = transaction(this.tableName)
```
elhigu: This variable is not a promise, but query builder... maybe name could be `insertBatchQuery` or something like that.

src/util/batchInsert.js
```diff
@@ -0,0 +1,84 @@
...
+   * @param transaction
+   */
+  transacting(transaction) {
+    this._transaction = transaction;
+    this._autoCommit  = false;
+    return this;
+  }
+
+  then(callback = function() {}) {
+    let transaction;
```
elhigu: Might look more clean if you just wrap getting the transaction to separate promise and resolve it. Something like: 

```
let transactionForBatches = .... promise for getting the transaction ...;

return transactionForBatches.then(trx => {
  // here you can add the rest of promise chain and no need to store transaction to temporary variable
});
```

elhigu:
Looks CI problems are because of jshint is giving some errors

```
src/util/make-knex.js: line 9, col 5, 'Promise' is defined but never used.
src/util/make-knex.js: line 10, col 17, 'isNumber' is defined but never used.
src/util/make-knex.js: line 10, col 27, 'chunk' is defined but never used.
3 errors
```

wubzz:
Not sure what's going on with those CI's..

elhigu:
@wubuzz It was complaining about jshint... try running locally 

```
npm run jshint
```

wubzz:
@elhigu I fixed the JSHint issues, but look at the log from the CI's now. Failing `apt-get install`'s
https://travis-ci.org/tgriesser/knex/jobs/129053667

Issue #1382, opened 2016-05-04
Pooled connections not releasing


devinivy:
I use knex 0.10.0 in my app with postgresql 9.4, and I've been having some trouble with pooling.

When my app gets busy and requires all pooled connections for an extended period of time, I find that connections are not always released back to the pool.  Eventually every connection becomes stuck as being active and the pool's queue grows indefinitely, causing every new query to eventually timeout ("Knex: Timeout acquiring a connection").

Meanwhile, my postgresql database sees those connections as being completely idle (those connections persist, but queries are not being sent on them).  Worth adding I'm not explicitly using `transaction()` or `transacting()` anywhere, so I can't see how it could be the issue described in #1040 / #1177.

I currently do not have a simple reproduction case– I just know that it can eventually happen when my app stays busy and queries start queueing on the pool.

Would love to know if anyone has any hunches or has experienced anything similar!  I'm looking into potential issues with pool2 as well.

wubzz:
@devinivy 
> Worth adding I'm not explicitly using transaction() or transacting()

If not using any transactions at all, getting this error is very strange since all the connections are technically just "shared" between users.

Are you using a single knex instance within your app, or multiple? The only thing I can think of if not using transactions is continiously reaching postgres's connection limit for the duration of `acquireConnectionTimeout` setting, but I find it very unlikely.

devinivy:
It is a single knex instance.  I have verified that the issue is not related to transactions– `DEBUG=knex:tx` didn't turn-up anything at all while the app ran and the issue occurred.  Trying to come-up with a knex-only minimal repro case to share.

devinivy:
I think the issue is that `acquireConnectionTimeout` (default `60000`) is less than `pool.requestTimeout` (default `Infinity`).  Knex's `acquireConnection()` fails due to a timeout, but `pool.acquire()` happily carries on and allocates a connection (to which knex is oblivious).  This connection is never released.

This was very confusing because `acquireConnectionTimeout` and `pool.acquireTimeout` are not analogous, nor are `pool2.acquire()` and `pool.acquireTimeout` directly related.  Turns out `pool2.acquire()` is primarily ruled by `pool.requestTimeout`!

@tjwebb you might be interested in this, since waterline-postgresql can fix this with its own defaults.

wubzz:
[...]

PR #1394, opened 2016-05-09
Abort request for connection from pool upon knex timeout

devinivy:
See #1382 for a more in-depth description of the bug.

This PR aborts the request to pool2 for a connection when knex times out due to `acquireConnectionTimeout`.  The purpose of this PR is to fix the issue without introducing a breaking change.  I think in the future `acquireConnectionTimeout` should be implemented using pool2's `requestTimeout`, or knex should simply set a default for `requestTimeout` (pool2 sets it to `Infinity`).  When that happens, this PR can probably be reverted.

This is my first time working on the knex codebase.  One thing I was unsure of was how the dialects' implementations relate to the base client implementation.  The client's implementation of `acquireConnection()` now returns an object containing a promise, while the dialect/transactional versions of the same method simply return a promise.  I wasn't able to test the affected dialects (mssql and oracle), so I'm not sure if there's anything wrong there.  I _think_ it's all set though.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -210,16 +210,26 @@ assign(Client.prototype, {
   // Acquire a connection from the pool.
   acquireConnection: function() {
     var client = this
-    return new Promise(function(resolver, rejecter) {
+    var request = null
+    var completed = new Promise(function(resolver, rejecter) {
       if (!client.pool) {
         return rejecter(new Error('There is no pool defined on the current client'))
       }
-      client.pool.acquire(function(err, connection) {
+      request = client.pool.acquire(function(err, connection) {
         if (err) return rejecter(err)
-        debug('acquiring connection from pool: %s', connection.__knexUid)
+        debug('acquired connection from pool: %s', connection.__knexUid)
         resolver(connection)
       })
     })
+    var abort = function(reason) {
+      if (request && !request.fulfilled) {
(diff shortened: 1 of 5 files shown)


devinivy:
Honestly can't figure out what the deal is with the failing test– can you make any sense of that stack trace?

wubzz:
@devinivy Could it be related to [transaction `acquireConnection`](https://github.com/tgriesser/knex/blob/master/src/transaction.js#L143) perhaps? The test uses a transaction.

The Stack doesn't make much sense, weird line numbers..

devinivy:
That's it– I'll have to fix this up a little bit.

rhys-vdw:
@wubzz Do you have a good understanding of this issue?

This is a part of Knex of which I have basically zero knowledge. I'd have to do a lot of reading to understand what is going on here.

wubzz:
@rhys-vdw Yeah this one is good to go, reviewed a while ago. Didn't get a notification from the **Edit** in his post for some reason. I see now @devinivy fixed the transaction issue.

rhys-vdw:
Thank you @devinivy!

rhys-vdw:
I'll try to get a release out soon.

elhigu:
I'm doing release right now. It'll be out in couple of hours

PR #1396, opened 2016-05-10
Evaluated multi insert default for undefined already in query compiler

elhigu:
Drivers cannot take knex.raw replacement values as positional replacement values, so they must be evaluated already in query compile time to sql string which is passed to driver.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -22,7 +22,7 @@ var inherits       = require('inherits')
 var EventEmitter   = require('events').EventEmitter
 var SqlString      = require('./query/string')
 
-import {assign, uniqueId, cloneDeep, map} from 'lodash'
+import {assign, uniqueId, cloneDeep} from 'lodash'
 
 var debug          = require('debug')('knex:client')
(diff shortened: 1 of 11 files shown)


elhigu:
Added implementation to work with same principle with every dialect. Now knex.raw('DEFAULT') is not passed until driver, but is already lowered accordingly by query compiler. After this change e.g. invalid use of undefined will be easy to catch in prepBindings (stuff like `.where('id', undefined)`).

elhigu:
Fixes e.g. `Error: Cannot find module 'pg-native'`

elhigu:
Could @wubuzz / @rhys-vdw check this out at some point... This could be released as a part of 0.11.2 maybe? I just did hit this again and had to downgrade back to 0.10 :|

wubzz:
@elhigu I see nothing wrong with the code, :+1: .

I actually kind of prefer `DEFAULT` in the plain sql rather than a binding `?` since technically it was never a used-submitted binding to begin with, so kudos to that. Looks good to me!

PR #1399, opened 2016-05-11
Add option for an array of ignorable migration errors

benf86:
The user can provide an array of error codes, which should be ignored when migrating, in the migrations.ignoreErrors part of the knex config, e.g.:

`"migrations": {
    "ignoreErrors": [
      "ER_DUP_FIELDNAME",
      "ER_CANT_DROP_FIELD_OR_KEY"
    ]
  }`

It was originally inspired by the idea that a migration which fails to e.g. delete a column, which had already been deleted (for whatever reason) isn't necessarily something that should stop the whole migration process from continuing.

The error gets added to the log array returned by the migration function, so the user can still decide how to handle it.

Failed migration errors now also include the failed migration file name for easier locating.

### DIFF
--- a/src/migrate/index.js
+++ b/src/migrate/index.js
@@ -314,6 +314,7 @@ export default class Migrator {
     var directory = this._absoluteConfigDir()
     var current   = Promise.bind({failed: false, failedOn: 0});
     var log       = [];
+    var ignoreErrors = knex.client.config.migrations.ignoreErrors || [];
     each(migrations, (migration) => {
       var name  = migration;
       migration = require(directory + '/' + name);
@@ -337,6 +338,14 @@ export default class Migrator {
         if (direction === 'down') {
           return knex(tableName).where({name: name}).del();
         }
+      })
+      .catch(e => {
+        e.migrationName = name;
+        if (ignoreErrors.indexOf(e.code) > -1) {
+          log.push(e);
+          return Promise.resolve();
+        }
+        throw e;
       });
     })
 


rhys-vdw:
@benf86 Migrations should roll back if any errors occur. If you're worried about deleting a column that already exists then the migration should check for its existance, or use `IF EXISTS` or equivalent.

A more appropriate solution for this would be to add `dropColumnIfExists`.

> Failed migration errors now also include the failed migration file name for easier locating.

That sounds potentially useful.

benf86:
@rhys-vdw I realize that they generally should roll back. However, why not give the developer the option of ignoring a failed migration and continuing instead of rolling back, if they decide it is a non-critical error?

Especially since this way it can also handle knex.raw queries?

rhys-vdw:
knex.raw queries can use `IF EXISTS` also. Or you can manually catch the errors in each migration where you're permitting it to fail.

Anyway, each migration is run in a transaction, so if any error happens in a migration the *entire migration* will fail. So unless you're happy for your migration to be a no-op, but still count as being run, this is not the right approach.

benf86:
> Or you can manually catch the errors in each migration where you're permitting it to fail.

Why didn't I even think of that -_-

While the migration counting even though it wouldn't execute was useful to my use case, you're obviously right that manually catching migrations errors is an option.

Thanks for the explanation and keep up the great work with knex!

rhys-vdw:
@benf86 Thanks for PR, despite it not being the best fit.

If you're interested in contributing you could look at adding some more `ifExists` style methods to help your use case. Just please open a feature request first for discussion with contributors/users.

rhys-vdw:
You'll have to disable migration transactions if you want to catch them
manually. I'd recommend using "if exists" still.

PR #1400, opened 2016-05-11
Fix broken "browser builds" link

mgmcdermott:
Broken link found while browsing through the docs.


rhys-vdw:
Thanks @mgmcdermott!

Issue #1348, opened 2016-04-15
Adding multiple columns in SQL Server migrations fails


nickwhiteley:
The syntax generated for mssql, when adding a column to an existing table in a migration file, prefixes every column with ADD.  Only the first column added should have the ADD clause, others are column separated.

``` js
knex.schema.table('users', function (table) {
  table.string('first_name');
  table.string('last_name');
})
```

Generates:

``` sql
ALTER TABLE [users] ADD  [first_name] nvarchar(255), ADD [last_name] nvarchar(255)
```

But it should generate:

``` sql
ALTER TABLE [users] ADD  [first_name] nvarchar(255), [last_name] nvarchar(255)
```

or:

``` sql
ALTER TABLE [users] ADD  [first_name] nvarchar(255)
ALTER TABLE [users] ADD  [last_name] nvarchar(255)
```

Adding the following to the assign in dialects\mssql\schema\tablecompiler.js fixed it for me.

``` js
  // Add columns to the table
  addColumns : function (columns) {
    var addClause = ' ';
    for (var i = 0, l = columns.sql.length; i < l; i++) {
      this.pushQuery({
        sql: 'ALTER TABLE ' + this.tableName() + ' ADD ' + columns.sql[i],
        bindings: columns.bindings[i]
      })
    };
  },

```

Putting it in a single ALTER statement breaks the rollback (the way I did it, can't see why) hence opting for multiple ALTER statements.

rhys-vdw:
@nickwhiteley Thanks for the report. Could you provide a PR for this change please?

dhensby:
It looks like this issue can be closed as it was fixed in #1348?

kibertoad:
@dhensby Do you mean #1401?

nickwhiteley:
Fixed in PR #1401

PR #1401, opened 2016-05-12
#1348 MSSQL adding and dropping multiple columns fix

nickwhiteley:
This is my first pull request so apologies if something isn't right.
It includes changes to MSSQL for adding and dropping columns.  Previously single column changes worked but multiple columns didn't.
https://github.com/tgriesser/knex/issues/1348

### DIFF
--- a/src/dialects/mssql/schema/tablecompiler.js
+++ b/src/dialects/mssql/schema/tablecompiler.js
@@ -37,6 +37,27 @@ assign(TableCompiler_MSSQL.prototype, {
 
   dropColumnPrefix: 'DROP COLUMN ',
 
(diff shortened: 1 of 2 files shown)


wolfgang42:
I can confirm that the tests pass for MSSQL.

This isn't specifically related to your PR, but it looks like the MSSQL unit tests need to be checked to make sure they are actually correct T-SQL.

nickwhiteley:
Indeed.  I have had a quick look and there are 10 which don't have valid T-SQL.  I don't have time today but will try to investigate further over the next week.

rhys-vdw:
Thanks @nickwhiteley! :+1:

PR #1410, opened 2016-05-14
Make sure 'raw' respects the "not" flag in the query builder. Fixes #…

### LINKED ISSUE #1402: knex.raw in whereNot
Hi,

`knex.raw()` in `whereNot()` function swallows `not` keyword.

```js
// Generates: select * from "table" where "is_active"
// Expected: select * from "table" where not "is_active"

knex
    .select('*')
    .from('table')
    .whereNot( knex.raw('??', 'is_active') );
```

Although I can make another query by hand easily, my code works on an AST generated by a RQL parser. As a result both `a=1&is_active` and `a=1&not(is_active)` RQL queries generates same SQL.

Is this behavior by design or a bug?

Best Regards,

wubzz:
Fixes #1402

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -269,6 +269,7 @@ assign(Builder.prototype, {
       grouping: 'where',
       type: 'whereRaw',
       value: raw,
+      not: this._not(),
       bool: this._bool()
     });
     return this;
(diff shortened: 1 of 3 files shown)


wubzz:
@elhigu @rhys-vdw This makes sense, right?

jurko-gospodnetic:
I guess it's ok. Never encountered the need, but it makes sense when I look at the examples in the tests.

rhys-vdw:
Looks good to me!

PR #1411, opened 2016-05-15
Allow creating datetime in mysql with fractional second part

JWells:
Hi, just a small change to allow setting the fractional second part to datetime columns in mysql.

e.g.
table.datetime('created_at', 6)

will allow

```
mysql> select created_at from users;
+----------------------------+
| created_at                 |
+----------------------------+
| 2016-05-15 16:56:24.831000 |
+----------------------------+
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -1987,22 +1987,22 @@ knex.schema.raw("SET sql_mode='TRADITIONAL'")
     </p>
 
     <p id="Schema-dateTime">
-      <b class="header">dateTime</b><code>table.dateTime(name)</code>
+      <b class="header">dateTime</b><code>table.dateTime(name, [fsp])</code>
       <br />
-      Adds a dateTime column.
+      Adds a dateTime column, with optional fractional seconds part for MySQL.
     </p>
 
     <p id="Schema-time">
-      <b class="header">time</b><code>table.time(name)</code>
+      <b class="header">time</b><code>table.time(name, [fsp])</code>
       <br />
-      Adds a time column.
+      Adds a time column, with optional fractional seconds part for MySQL.
     </p>
 
     <p id="Schema-timestamp">
       <b class="header">timestamp</b><code>table.timestamp(name, [standard])</code>
       <br />
       Adds a timestamp column, defaults to <tt>timestamptz</tt> in PostgreSQL, unless true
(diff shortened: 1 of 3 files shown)


chrisbroome:
`timestamp` columns can also have fractional seconds precision.

JWells:
Good point, I've modified so it applies to time/timestamp and datetime

rhys-vdw:
Hi @JWells. Sorry for the slow response. This has been requested a few times I believe.

However, I'm reluctant to merge this one in without documentation.

I just had a browse and saw that postgresql's column compiler takes a different argument: A boolean to indicate whether the type should be `timestamp` or `timestamptz`. This is despite the fact that PostgreSQL [also supports a precision argument](https://www.postgresql.org/docs/9.5/static/datatype-datetime.html) (with the same syntax).

I think the best choice here would be to keep the API the same for both postgres and mysql (using your proposed argument). To do this we'd have to deprecate passing a boolean to `timestamp` and `datetime` for PostgreSQL, and create method `timestamptz` instead.

So if you're keen to make the above changes @JWells, that would be great. Oherwise just add a note under the docs (`index.html`) specifying that precision can be provided for MySQL and I'll merge it in.

JWells:
No problems. I've updated the docs for this one to get at least the MySQL part done. I agree it should be updated to be consistent across the board. I think that should be separate commits though so left it out of this one. Will try and take a look at updating the postgres ones soon.

jabedhasan21:
Hi @rhys-vdw,
It will be pleasure for me if you merge this pull request. :)
One of my project i am working that's badly needed time/timestamp/datetime field precision.
In every new migration i have to manually alter the table field for precision value. :(

mborst:
So, is there anything that can be done to get this merged? What is missing?

scott-martin:
I could also use this feature. Any news on merging this in?

elhigu:
rebase + comments from  https://github.com/tgriesser/knex/pull/1411#issuecomment-225573162 should be done. Now adays documentation updates should be done to separate repo which is described in CONTRIBUTING.md. After adding pull request for documentation just add link to it, so that they are easy to track together.

elhigu:
Closing older incomplete PRs. Can be reopened if finished.

PR #1412, opened 2016-05-15
Read `version` from `package.json`

rhys-vdw:
Means that it's impossible to have `Knex#VERSION` and the version in `package.json` out of sync. This is one less thing to do for release.

This required the addition of `json-loader` for webpack. I've also added webpack to `devDependencies` rather than installing it manually during the `build` script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "coveralls": "~2.11.1",
     "istanbul": "~0.3.0",
     "jshint": "^2.9.2",
+    "json-loader": "^0.5.4",
     "mariasql": "^0.2.3",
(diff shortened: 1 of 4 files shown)


jurko-gospodnetic:
:+1:

elhigu:
@rhys-vdw nice one 👍 so in addition to running the script one just has to update the changelog before releases?

rhys-vdw:
@elhigu Yes, but I'm hoping to do a HTML -> MD conversion and remove it from the site. Then we can add a PR checklist that requires PRs to actually update the changelog in place. Then it should be as simple as adding a version number and date to the changelog.

Issue #1354, opened 2016-04-18
batchInsert with existing transaction


alexwhitman:
`batchInsert` automatically starts/commits/rollsback a transaction but it would be useful if it could be used with an existing transaction which then isn't automatically completed.

rhys-vdw:
@alexwhitman Does this not work?

```js
knex.transaction(trx => {
  return trx.batchInsert(/* ... */);
});
```

alexwhitman:
I'm not at work to test it now but in theory it should.  However, from looking at the code wouldn't it start a new transaction within the existing?  If so that would seem redundant.

wubzz:
@alexwhitman This has been added to knex 0.11.1, and the documentation has been updated as to how this works.

alexwhitman:
@wubzz Thanks, just tested and it works fine.

Just a thought though, the example that @rhys-vdw gave was to call `batchInsert` on the transaction itself but that still starts a nested transaction.  Instead it would have to be called as
```js
trx.batchInsert(data).transacting(trx);
```
which looks slightly redundant.  Would it be possible to modify so that both
```js
knex.batchInsert(data).transacting(trx);

trx.batchInsert(data);
```
act in the same way?

alexwhitman:
The following patch appears to work but I'm not sure if it's the correct way to implement it:
```diff
diff --git a/src/util/batchInsert.js b/src/util/batchInsert.js
index 4aaaa86..5594c7f 100644
--- a/src/util/batchInsert.js
+++ b/src/util/batchInsert.js
@@ -19,7 +19,10 @@ export default class BatchInsert {
     this._returning       = void 0;
     this._transaction     = null;
     this._autoTransaction = true;
-    
+
+    if (client.transacting) {
+      this.transacting(client);
+    }
   }
 
   /**
```

wubzz:
@alexwhitman I can see the convenience in that, and I think your diff makes perfect sense. Feel free to make a PR on that!

PR #1418, opened 2016-05-16
Use calling transaction with batchInsert

alexwhitman:
Updates `batchInsert` so that it will use the existing transaction which calls it.
```js
// Existing method for using a transaction with batchInsert
knex.batchInsert(data).transacting(trx);

// New method
trx.batchInsert(data);
```
Implements a change discussed in #1354

### DIFF
--- a/test/integration/builder/inserts.js
+++ b/test/integration/builder/inserts.js
@@ -702,6 +702,15 @@ module.exports = function(knex) {
         })
       });
 
+      it('transaction.batchInsert using specified transaction', function() {
+        return knex.transaction(function(tr) {
+          tr.batchInsert('BatchInsert', items, 30)
+          .returning(['Col1', 'Col2'])
+          .then(tr.commit)
+          .catch(tr.rollback);
+        })
+      });
+
     });
 
     it('should validate batchInsert batchSize parameter', function() {
--- a/src/util/batchInsert.js
+++ b/src/util/batchInsert.js
@@ -19,6 +19,10 @@ export default class BatchInsert {
     this._returning = void 0;
     this._transaction = null;
     this._autoTransaction = true;
+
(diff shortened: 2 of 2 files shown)


test/integration/builder/inserts.js
```diff
@@ -702,6 +702,15 @@ module.exports = function(knex) {
         })
       });
 
+      it('transaction.batchInsert using specified transaction', function() {
+        return knex.transaction(function(tr) {
+          tr.batchInsert('BatchInsert', items, 30)
+          .returning(['Col1', 'Col2'])
+          .then(tr.commit)
+          .catch(tr.rollback);
```
jurko-gospodnetic: Any reason why we don't just simplify our transaction using test cases like this:

```
return knex.transaction(functinon(tr) {
    return tr.batchInsert('BatchInsert', items, 30);
});
```
Since we do not do anything special with the result, there seems to be no need to call `returning()` + explicitly calling `tr.commit()` & `tr.rollback()` just seems like potentially confusing noise.

Though I see this is pretty much a copy/paste of an already existing test so we could do that in a separate cleanup pull request as well.

@alexwhitman, @wubzz - what do you think?

jurko-gospodnetic:
:+1:

test/integration/builder/inserts.js
alexwhitman: @jurko-gospodnetic Like you mentioned, I just copied/pasted an existing test.  I'll leave what happens to the tests up to the maintainers.

alexwhitman:
I've rebased this on `master` so that it will now merge after the `eslint` changes.

The Travis build failed for Node 0.12 due to a timeout on an unrelated test.

wubzz:
@alexwhitman Thanks!

PR #1419, opened 2016-05-17
Add SQL Anywhere dialect support.

phil-mitchell:
We have had a request from our users to be able to use knex with SQL Anywhere in Node.  I have written the dialect and the tests are all passing using SQL Anywhere 17.0.0 and the SQL Anywhere Node driver (npm install sqlanywhere).

To test:
- install SQL Anywhere 17 (developer edition download here: [http://scn.sap.com/community/developer-center/sql-anywhere](http://scn.sap.com/community/developer-center/sql-anywhere))
- source /opt/sqlanywhere17/bin64/sa_config.sh
- cd test
- dbinit -dba knex_test,knex_test knex_test.db
- cd ..
- DB='sqlanywhere' npm test

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,7 @@ You can optionally specify which dialects to test using the `DB` environment var
 * maria
 * oracle
 * mssql
+* sqlanywhere
 
 ```bash
 $ DB='postgres mysql' npm test
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Gitter chat
 
 A SQL query builder that is flexible, portable, and fun to use!
 
-A batteries-included, multi-dialect (MySQL, PostgreSQL, SQLite3, WebSQL, Oracle) query builder for
+A batteries-included, multi-dialect (MySQL, PostgreSQL, SQLite3, WebSQL, Oracle, SQL Anywhere) query builder for
 Node.js and the Browser, featuring:
 
 - [transactions](http://knexjs.org/#Transactions)
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "mariadb",
     "sqlite3",
     "oracle",
+    "sqlanywhere",
     "mssql"
   ],
   "author": {
@@ -90,6 +91,7 @@
     "mariasql": false,
     "pg": false,
     "pg-query-stream": false,
+    "sqlanywhere": false,
     "oracle": false,
     "strong-oracle": false,
     "sqlite3": false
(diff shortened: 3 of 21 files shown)


elhigu:
@phil-mitchell could this be implemented as separate client package which is bound to certain kenx version. Since this cannot be easily CI tested by travis and there are not many users using the database, there is big change of this code being broken all the time.

While ago there was feature implemented, which allows to use external custom dialects #1428

@rhys-vdw what do you think about this?

phil-mitchell:
@elhigu external custom dialect should suit my needs.  Although, it will be more maintenance for me to keep up with your versions, even when there are no breaking changes involved.  Deployment would also be more work for end-users.  I do recognize that other non-free servers (Oracle, mssql, ...) which are included in knex directly are more heavily used, so bugs would be more quickly reported.

The custom dialect option was not available when I started looking at this.  I will look at that feature and see if it works for me.

elhigu:
@phil-mitchell We are about to get CI tests working with oracle 👍 If you are able to setup sqlanywhere to run in travis environment the same way that oracle will be setup I wouldn't hesitate to take this dialect in.

phil-mitchell:
It should be possible to set up sqlanywhere in your CI environment.  The development license is free, so it would just be a matter of knowing your requirements.  If you can provide details, I will do the work.

FWIW, I am one of the sqlanywhere core developers, so it should be no trouble for me to resolve any issues we encounter.

elhigu:
@phil-mitchell here is the pull request how support is being added for oracle https://github.com/tgriesser/knex/pull/1503 it uses some helper scripts from https://github.com/cbandy/travis-oracle to install the oracle server instance to CI server.

I suppose that similar workflow should do for sqlanywhere too. 

If you are one of the core developers you may consider releasing own `travis-sqlanywhere` package to give helpers and instructions for other travis users how to run CI against your database.

williamgueiros:
@phil-mitchell if you need help i stay here to help you.

@elhigu i think you are right, in the next version is good idea, putting all dialect external knext.
see that example https://github.com/elixir-ecto/ecto
all databases are a dependency and have maintainer, to developer new features.

Monolithic knext i don't think is a good ideia, for example i don't using mysql,MSSQL and install dialect to that.

elhigu:
@phil-mitchell Also if this is completed I would need someone who is willing to maintain the dialect and answer to issues about it. Closing for now due to inactivity. 

Can be reopened if CI stuff is added to travis configuration, docuementation added and there is someone who is willing to be maintainer for the dialect.

Issue #1361, opened 2016-04-21
`npm install` fails on Windows


wolfgang42:
Windows interprets `./scripts/build.sh` as a call to command `.` with the arguments `/scripts` and `/build.sh`:

```
> knex@0.10.0 build C:\Users\wfaust\Documents\Development\knex
> ./scripts/build.sh

'.' is not recognized as an internal or external command,
operable program or batch file.
```

rhys-vdw:
Can you think of a workaround for this?

wolfgang42:
On second thought, the underlying problem is that this is a UNIX shell script, so the Windows CLI wouldn't understand it anyway. I suppose you could require mingw to be installed, and run it under bash; however that's another dependency.

The least hacky way I can think of to fix the problem is to convert it into `build.js` and run it under Node. Does that sound like a good solution to you?

elhigu:
@wolfgang42 sounds like a good idea. I've been using `shelljs` for that kind of stuff and been pretty happy with it 👍

wolfgang42:
I've done the change [in wolfgang42/knex](https://github.com/wolfgang42/knex/tree/issue-1361-replace-buildsh); however I think I'm going to wait for #1416 to land before submitting a PR.

PR #1425, opened 2016-05-18
Replace scripts/build.sh with build.js.

wolfgang42:
This makes the build script cross-platform, closing #1361.

I also added the new `build.js` to the list of files to be linted.


wolfgang42:
Fixing to work on node < 4...

wolfgang42:
For some reason node=1.10 gets into an (infinite?) loop when `npm run babel` is run from inside `build.js`. I've reproduced the issue but am having trouble diagnosing it.

elhigu:
@wolfgang42 Ah... yep thats probably shelljs which works a bit different on node 0.10. I had problem with it outputting some really verbose messages. I'm not sure if it is the best choice after all if we need to support node 0.10... I had to make it silent with separate flag:

```js
    shelljs.exec('mkdir -p ' + tempPath, {silent: true});
    shelljs.exec('rm -f ' + path.join(tempPath, '*.deleteme'), {silent: true});
    shelljs.exec('rm -f ' + path.join(tempPath, '*.timestamp'), {silent: true});
    shelljs.exec('rm -f ' + path.join(tempPath, '*.m4'), {silent: true});
    shelljs.exec('rm -f ' + path.join(tempPath, '*.conf'), {silent: true});
```

But I'm not sure how those direct rm calls will work on windows :| Maybe you could try to set global silent flag on:

```js
var shell = require('shelljs');
shell.config.silent = true;
```

vellotis:
This issue doesn't reproduce on windows with either v0.10.45 32bit or 64bit.

vellotis:
I think this may be realted shelljs/shelljs#51

wolfgang42:
@elhigu Yep, shelljs/shelljs#427:
> Every command I send to the shell via the shelljs is executed multiple times (random number, usually around 5-10).
> ...
> Unfortunately, we recently dropped support for v0.10 for future development. Since we're getting ready to release v0.7, we probably won't be able to focus on this issue. If you can figure out a fix, please feel free to submit a PR.

The `execSync` library works on 0.10, but breaks in node 5...

wolfgang42:
@vellotis I can reproduce it on Ubuntu with 0.10.45; I haven't tried it with 0.10+Windows.

vellotis:
@wolfgang42 Can you try to implement it as async as mentioned in shelljs/shelljs#51 ?

wolfgang42:
This seems to have been long since obsoleted by #1616 (removed `build.sh`) and #1708 (added `build.js`).

PR #1427, opened 2016-05-19
Expose Knex version also on the constructor

vellotis:
Previously only exposed on knex instance or read/parsed straight from `package.json`.

I suggest it would be more valuable than having it on knex instance. Its an improvement PR. So current API will remain the same.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -38,6 +38,9 @@ export default function Knex(config) {
 // Expose Client on the main Knex namespace.
 Knex.Client = Client
 
+// Expose Knex version on the main Knex namespace.
+Knex.VERSION = require('../package.json').version
+
 // Run a "raw" query, though we can't do anything with it other than put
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -38,6 +38,9 @@ export default function Knex(config) {
 // Expose Client on the main Knex namespace.
 Knex.Client = Client
 
+// Expose Knex version on the main Knex namespace.
+Knex.VERSION = require('../package.json').version
+
```
jurko-gospodnetic: Should we then change setting the version information on a knex instance to get its value from here>

vellotis:
@jurko-gospodnetic The suggestion is reasonable. Updated the solution.

jurko-gospodnetic:
:+1:

rhys-vdw:
Thanks @vellotis!

Issue #1424, opened 2016-05-18
Support custom Client/Dialect


vellotis:
> :information_source: Separate issue for the feature requested in #1280.

It would be great if we could use our own defined dialect which is a instance of base [Client (src/client.js)](https://github.com/tgriesser/knex/blob/e7a42abbb200aaad3f6ac0324c28af75931c0583/src/client.js).

By providing custom dialect in initialization config the suggested fix would be... [src/inrex.js#L17-L22](https://github.com/tgriesser/knex/blob/f37824a11efcb398e11425905fb5bbc864ae8b5a/src/index.js#L17-L22)
```js
/*17*/  if (arguments.length === 0 || (!config.client && !config.dialect)) {
/*18*/    Dialect = makeClient(Client)
/*++*/  } else if (typeof config.client === 'function' && config.client.prototype instanceof Client) {
/*++*/    Dialect = makeClient(config.client)
/*19*/  } else {
/*20*/    var clientName = config.client || config.dialect
/*21*/    Dialect = makeClient(require('./dialects/' + (aliases[clientName] || clientName) + '/index.js'))
/*22*/  }
```
Is the solution acceptable? Test(s) + PR?

BR

rhys-vdw:
@vellotis One thing to point out - you can't rely on `client instanceof Client`, as it may be inheriting from a different version of Knex! (eg. a different node module).

eg.

```js
// my-custom-client.js
var Client = require('knex/lib/client'); // knex 0.10
```

```js
// my-project-in-an-entirely-different-module
var Client = require('knex/lib/client') // knex 0.11
var CustomClient = require('my-custom-client');

assert(CustomClient instanceof Client);
// FAIL! `CustomClient` inherits from the Knex dependency under the `my-custom-client` module.
```

So instead either use [existing methods](https://github.com/tgriesser/knex/blob/e7a42abbb200aaad3f6ac0324c28af75931c0583/src/client.js#L263-L265):

```js
// Check already defined `toString`
if (client.toString() === '[object KnexClient]') {
  makeClient(client);
 }
```

The above method will fail if the new client overrides `toString`. A sentinel prop is a bit more robust:

```js
// knex/src/client.js

// Define some safe property by which we'll identify `Client` types.
const CLIENT_SENTINEL = '__KNEX_CLIENT__';

// Assign it to prototype (so is present on all instances).
Client.prototype[CLIENT_SENTINEL] = true;

// Create helper function on `Client`
[...]

PR #1428, opened 2016-05-19
Allow using custom Client/Dialect

vellotis:
Solve #1424

To oppose @rhys-vdw [note in #1424](https://github.com/tgriesser/knex/issues/1424#issuecomment-220247608).

I stand to my suggested solution. Client API is really much knex version dependent. If some version of knex would change the API, the plugin-client would stop working any way. So there is no point supporting any other soft verification.

Two tests added.
- One for currently present feature
- One for the new feature

As well PR #1427 merging with this PR would be welcomed.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -25,7 +25,7 @@ export default function Knex(config) {
   let Dialect;
   if (arguments.length === 0 || (!config.client && !config.dialect)) {
     Dialect = makeClient(Client)
-  } else if ((typeof config.client === 'object') && (config.client.prototype instanceof Client)) {
+  } else if (typeof config.client === 'object' && config.client.prototype instanceof Client) {
     Dialect = makeClient(config.client)
   } else {
     const clientName = config.client || config.dialect
--- a/src/index.js
+++ b/src/index.js
@@ -25,7 +25,7 @@ export default function Knex(config) {
   let Dialect;
   if (arguments.length === 0 || (!config.client && !config.dialect)) {
     Dialect = makeClient(Client)
-  } else if (typeof config.client === 'object' && config.client.prototype instanceof Client) {
+  } else if (typeof config.client === 'function' && config.client.prototype instanceof Client) {
     Dialect = makeClient(config.client)
   } else {
     const clientName = config.client || config.dialect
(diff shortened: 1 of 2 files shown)


vellotis:
Hmm... test fails for some unknown reason. All ok on my own branch https://travis-ci.org/vellotis/knex/builds/131377631. The same commit 41344a9.

src/index.js
```diff
@@ -25,6 +25,8 @@ export default function Knex(config) {
   let Dialect;
   if (arguments.length === 0 || (!config.client && !config.dialect)) {
     Dialect = makeClient(Client)
+  } else if (typeof config.client === 'function' && config.client.prototype instanceof Client) {
```
rhys-vdw: This can be simplified to:

```js
} else if (config.client instanceof Client)
```

`instanceof` will walk the prototype list. It will return `false` if `config.client` is not a function.

rhys-vdw:
> Hmm... test fails for some unknown reason.

Don't sweat it. Sometimes the tests time out. I can just re-run that individual test (you can see most node versions passed).

If you push a commit addressing my line note the tests will re-run anyway.

src/index.js
vellotis: @rhys-vdw I must say that you are not right. You are not passing initialized Client but the definition of the Client. Lets see it in node cli.
```bash
> Client = require('knex/lib/index').Client
{ [Function: Client] ... }

> MysqlClient = require('knex/lib/dialects/mysql')
{ [Function: Client_MySQL] ... }

> MysqlClient instanceof Client
false

> MysqlClient.prototype instanceof Client
true
```

rhys-vdw: Sorry, you're right.

vellotis:
BR :tada:

rhys-vdw:
👍

elhigu:
I think this would really need documentations since it is new feature... I'll open issue for it.

igorklopov:
FYI i made use of custom dialect feature. The only problem i see is a wierd way of getting ancestors:
https://github.com/igorklopov/firebird-knex/blob/master/src/transaction.js#L5
Can you export Transaction, ColumnCompiler, TableCompiler etc from knex same way as Knex.Client and Knex.Promise:
https://github.com/tgriesser/knex/blob/master/src/index.js#L41

PR #1434, opened 2016-05-20
Raised global test timeout to 10 seconds

elhigu:
There has been too many random test timeout failures.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "plaintest": "mocha --check-leaks -t 10000 -b -R spec test/index.js && npm run tape",
     "prepublish": "npm run build",
     "tape": "node test/tape/index.js | tap-spec",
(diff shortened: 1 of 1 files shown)


jurko-gospodnetic:
There's some other issue at hand there. No tests we have should ever come close to running in 5 seconds.

rhys-vdw:
```
  1) Integration Tests mariadb | mariasql Schema createTable is possible to chain .catch:
     Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
```

wtf

rhys-vdw:
> We could merge this in as a band-aid, but in that case, could we open a separate issue to look into the test failures in more detail as well and get rid of the band-aid?

I'm tempted to say let's not merge this and open an issue to actually fix the issue.

elhigu:
@rhys-vdw Yep... looks like it really is hanging for some reason

PR #1437, opened 2016-05-21
Added support for chaining .insert and .increment/.decrement. Fixes #647

### LINKED ISSUE #647: when chaining update calls, only the last takes effect
Whereas `where` calls can be chained like so:

```
> knex('foo').where('a', 1).where('b', 2).toString()
'select * from "foo" where "a" = \'1\' and "b" = \'2\''
```

`update` calls cannot:

```
> knex('foo').update('a', 1).update('b', 2).toString()
'update "foo" set "b" = \'2\''
```

May not exactly be a bug, but the difference has me wondering whether this is by design.

wubzz:


src/query/builder.js
```diff
@@ -10,7 +10,7 @@ import * as helpers from '../helpers';
 import JoinClause from './joinclause';
 import {
   assign, clone, each, isBoolean, isEmpty, isFunction, isNumber, isObject,
-  isString, isUndefined, tail, toArray
+  isString, isUndefined, tail, toArray, concat, isArray
```
jurko-gospodnetic: these seem to have been ordered alphabetically prior to this change

wubzz: This PR won't be merged anyway. `increment` / `decrement` was no issue, but seems `insert` will be tricky to make work with this. Will have to rethink this before attempting again.

src/query/compiler.js
```diff
@@ -310,9 +310,12 @@ assign(QueryCompiler.prototype, {
   counter() {
     const { counter } = this.single;
     const toUpdate = {};
-    toUpdate[counter.column] = this.client.raw(this.formatter.wrap(counter.column) +
-      ' ' + (counter.symbol || '+') +
-      ' ' + counter.amount);
+    Object.keys(counter).forEach((column) => {
+      let columnCounter = counter[column];
+      toUpdate[column] = this.client.raw(this.formatter.wrap(column) +
+        ' ' + (columnCounter.symbol || '+') +
```
jurko-gospodnetic: Why do we have `|| '+'` here? Seems like the `symbol` value is always set to `+` or `-` in `builder.js`.

src/query/builder.js
```diff
@@ -751,11 +755,23 @@ assign(Builder.prototype, {
...
+
+    if(!counter[column]) {
+      counter[column] = {
+        amount: amt,
+        symbol: (symbol || '+')
+      };
+    } else {
+      //Counter for column already exists. Add/subtract the current column amount by the new amount
+      if(symbol === '-') {
+        counter[column].amount -= amount;
```
jurko-gospodnetic: I think this might actually be buggy. If first do decrement by 10, we set the counter to `{amount: 10, symbol: '-'}` but if we then do another decrement by 5, the value will get changed to `{amount: 5, symbol: '-'}` while it should in fact be changed to `{amount: 15, symbol: '-`}`.

I think we should drop the `symbol` here and simply store the amount as a signed value where positive values represent increments and negative ones represent decrements.

elhigu: To me actually would make more sense that amount is always overridden / failure would be thrown if one tries to set  increment / decrement twice for the same field.

Using builder to calculate values back and forth sounds inefficient and feels to be something that builder shouldn't doing.  Just overriding / throw when same field is set twice is simpler / more robust to implement and easier to test too.

Anyways just my 2-cents, how I feel that feature should work 👍 I don't have any strong opinion about it.

jurko-gospodnetic: :+1:

wubzz: @elhigu In pratical terms that would be pretty annoying, imo. For example if you're doing an application where you need to loop to calculate a new value, and each iteration can contain either an addition or a subtraction, you shouldn't have to store the identifier (+/-) in a seperate variable to know whether to call increment/decrement, nor should you have to store the final value that is to be updated in a seperate variable.

It would be **a lot** easier and far more convenient while in the loop to simply call queryBuilder.increment/decrement with the value of the current iteration, which eventually results in the correct query when executed. That's my 2-cents on this.

In any case I'm not sure why there's a need to review something that I've already closed, this issue needs more thought around the `insert` function as the current implementation won't allow an easy fix.

elhigu: @wubzz Yep. No point of reviewing, just discussing... and this probably is the worst place for that :D 

For that particular use case it would be convenient. But if there are just 0.01% of userbase needing that feature it is hard to justify increase of complexity. 

For example if I'm writing parser that afterwards builds query from AST I probably wouldn't call builder directly from token handlers, but would first create my own internal representation of the query to prevent fight with shortcomings of query builder (also because I would like to normalize the internal presentation before building the final query to make optimization / testing etc. easier).

PR #1441, opened 2016-05-23
Fix bigint support for mssql

sandorfr:


src/dialects/mssql/index.js
```diff
@@ -85,6 +85,15 @@ assign(Client_MSSQL.prototype, {
     })
   },
 
+  prepBindings(bindings) {
+    return map(bindings, (value) => {
+    if ( typeof value === 'number' && value >= 2147483647 || value <= -2147483648){
```
rhys-vdw: Why use these numbers? This is to prevent loss of integer precision right?

The actual valid range is [[Number.MIN_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER), [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)].

sandorfr: the limit for tedious are actually those numbers, kinda hardcoded there : https://github.com/pekim/tedious/blob/50e76efdcd87f687106c722d54c6ea4096f0a127/src/data-type.js#L573

sandorfr: I forgot to mention but yes this is to prevent loss of integer precision. 
so in fact tedious considers js number type are int4 by default. bigints are expected to be strings. I think tedious could be a bit smarter about the way it handles numbers by default. But until they improve that or we work on a PR, we will have to fix this on the knex side I think.

elhigu:
I left comment to #1442. If it makes sense, please close this PR and try another type of implementation.

Issue #1442, opened 2016-05-23
mssql bigint support


sandorfr:
Mssql dialect currently crashes when bigint are used. This is due to the way [tedious handles them](http://pekim.github.io/tedious/api-datatypes.html). 

I'm working on a pull request (#1441) but I'm unable to build the source atm (probably related to #1180).

Looking forward to your feedback and help regarding the build / tests.

rhys-vdw:
#1180 is a problem in installing an this repository as an npm dependency. If you're working on a PR you should build source locally using `npm run babel`.

sandorfr:
@rhys-vdw ok works now, but I had to `rimraf` the node_modules and remove the build script in order to do the package restore.

The build script requires bash (not very windows friendly). I think moving  to plain node script would be easier. I work on a PR for that matter it if you are interested.

rhys-vdw:
> The build script requires bash (not very windows friendly).

See #1425. Perhaps you can help test this PR.

elhigu:
As far as I could figure this out the problem is that we need to tell to mssql driver in which format it should pass javascript variables to DB.

https://github.com/tgriesser/knex/blob/master/src/dialects/mssql/index.js#L107

`req.input()` actually takes more parameters which can be used to define in which format value should be sent. https://github.com/patriksimek/node-mssql

So before calling req.input one could check if value is number and bigger than 32bit and smaller than 53bit and send it as int8 instead of int4. If number is bigger that min/max_safe_integer it should throw an error since value has been clamped to fit 53bit.

Any thoughts? I haven't tested any of this... doing conversion in `prepBindings` didn't feel like appropriate fix, since we still wouldn't tell to driver how value should be passed.

sandorfr:
Thanks for pointing me in the right direction. I'll look into this and update my PR branch accordingly.

sandorfr:
@elhigu I started a new PR with a new implementation. I did not use isSafeInteger because it would probably break on older node version.

sandorfr:
Closing this resolved issue and will open new ones about failing unit tests later.

PR #1445, opened 2016-05-24
add bigint detection and sets mssql req.input when appropriate

sandorfr:
#1442


elhigu:
Implementation looks correct, but this really would need test case too.

src/dialects/mssql/index.js
```diff
@@ -138,6 +140,18 @@ assign(Client_MSSQL.prototype, {
     })
   },
 
+  // sets a request input parameter. Detects bigints and sets type appropriately.
+  _setReqInput(req, i, binding) {
+    if (typeof binding == 'number' && (binding < -2147483648 || binding > 2147483647)) {
+      if (binding < -9007199254740991 || binding > 9007199254740991) {
```
elhigu: Would be nice for the code to give names for the magic numbers used e.g. MIN/MAX_INT4 MIN/MAX_SAFE_INTEGER. Anyways we are using babel to compile sources to ES5 so I suppose Number.MAX_SAFE_INTEGER / Number.isSafeInteger() should work too.

sandorfr: Ok for the constants, isSafeInteger wasn't transpiled or polyfiled (I guess this is a matter of babel profile and plugin) Maybe constants would be the most compatible approach?

elhigu: Yep, sounds good.

sandorfr: OK will do soon and add the corresponding unit tests.

sandorfr:
@elhigu I somewhat failing at writing units test for that cas and I can't find one testing similar issue. I tried to mock the connection but it is still trying to connect to a real database :( ). I you can give me a hint :)

elhigu:
I have no I idea if mocking the client is even possible, do you have mssql installed so that you could just write integration test for it?

sandorfr:
@elhigu ok just pushed them. I have a questions :). Why bail option is activated by default on test script (this bugged me a while)?  And I see 24 tests failing for mssql, they don't appear to be any related to my changes so I guess mssql support is still a WIP?

elhigu:
It is really hard to keep track how mssql / oracle are working since I have never ran mssql tests myself... Probably at some point all tests were passing and I suppose they are broken again. You should try running tests before applying your changes to check if errors are caused by your changes...

I just few hours ago sent email to microsoft people if they could provide me preview of their upcoming mssql for linux version and if there is any possibility to set it up for knex CI... not having high hopes for that though :) maybe some day.

Test bailing is on at least because of integration tests are stateful so if one test fails, it is possible that all the tests will fail after that since DB is not in correct state. Though I don't know what has been real motivation for that.

elhigu:
@sandorfr could you still confirm that other test failures were not caused by this change and maybe add an issue with errors of failing tests if mssql is currently broken?

sandorfr:
Just tested, the errors are on master branch, I think some of them are just outdated tests. I'll open an issue for that.

Regarding mssql tests a basic azure instance is just 5 euros per month so I can provide you an access to one until Microsoft releases mssql for linux (or invites you to join the beta).

elhigu:
@sandorfr thanks for the offer I'll discuss about using azure mssql instance for CI with other collaborators 👍

elhigu:
@sandorfr now that I was thinking how to use external DB from travis CI so that there can be multiple tester instances running against the same DB at the same time, would require some extra coding to implement. If someone is able to make it work I would accept pullrequest for that though :)

sandorfr:
@elhigu Yes concurrency is definitely non trivial. I think one approach would be to use a different schemas for each build (and drop the schema and its objects at the end of the test run). But I think the current implementation for mssql does not support schemas and assumes it is `dbo` (see https://github.com/tgriesser/knex/blob/master/src/dialects/mssql/query/compiler.js#L163 for instance). So the first step would be to enable schema support for mssql.

PR #1447, opened 2016-05-25
Fix webpack build error 'dependency to an entry point is not allowed'

elhigu:


src/util/make-knex.js
```diff
@@ -8,9 +8,8 @@ import QueryInterface from '../query/methods';
 import * as helpers from '../helpers';
 import { assign } from 'lodash'
 import BatchInsert from './batchInsert';
-import Knex from "../index";
 
-export default function makeKnex(client) {
+export default function makeKnex(client, version) {
```
jurko-gospodnetic: Hmmm... Are we sure this will work correctly? I'm quite certain `makeKnex` is called from more places than just `src/index.js` and those places would then need to be updated as well or they'll set the created `knex` instance's `VERSION` attribute to `undefined`.

I'm not quite sure - why is it that we want to add this version parameter anyway? Can we instead just move the whole version code into a separate module and then include it from there to both here and the code that sets it on the `Knex` class?

rhys-vdw: Simplest solution: Just require `package.json` in both locations.

jurko-gospodnetic: Agreed.

Though if it were up to me, I'd add a `version.js` file as an interface layer, have it import the `package.json` file and export just the version information, and then require that `version.js` file from wherever the version is needed. But that's just because importing a '.json' as anything by a localized implementation detail smells bad to me. :imp:

elhigu: Yep... actually we already have that version in external package.json file, which was my second option for implementing this :) I'll change this one to read package json directly.

elhigu:
Still need to add build success test for webpack...

elhigu:
@jurko-gospodnetic I added running webpack build during `npm test`. I had to add one webpack plugin there to make webpack to return exit code 1 when there are errors during execution.

PR #1452, opened 2016-05-26
Any binding that is an object/class containing the function .toSQL mu…

### LINKED ISSUE #1451: Using raw as binding throws module not found / circular json
Given the following example:

`knex(table).whereRaw('LOWER("Login") IN (?)', knex.raw('SELECT LOWER("Login") FROM "users" WHERE "prefix" = ?', [binding]))`

Running the query generates
`TypeError: Converting circular structure to JSON`

Running `.toString()` generates either
`TypeError: Converting circular structure to JSON`
or
`Can't find module 'pg-native'`
*(How can it be random?)*

Running `.toSQL()` generates
`TypeError: Converting circular structure to JSON`

@elhigu Do you know anything about this? It sounds like the stuff with sending bindings down the chain to the driver. If I run `.toString()` on the `raw` as a parameter, the errors go away. (Although the binding gets quoted too..)

Edit: This is tested with latest release.

wubzz:
…st be processed as an Array in Raw's toSQL.

Fixes #1451

### DIFF
--- a/src/raw.js
+++ b/src/raw.js
@@ -25,7 +25,7 @@ assign(Raw.prototype, {
   set(sql, bindings) {
     this._cached = undefined
     this.sql = sql
-    this.bindings = (isObject(bindings) || isUndefined(bindings)) ?  bindings : [bindings]
+    this.bindings = ((isObject(bindings) && !bindings.toSQL) || isUndefined(bindings)) ?  bindings : [bindings]
 
     return this
   },


src/raw.js
```diff
@@ -25,7 +25,7 @@ assign(Raw.prototype, {
   set(sql, bindings) {
     this._cached = undefined
     this.sql = sql
-    this.bindings = (isObject(bindings) || isUndefined(bindings)) ?  bindings : [bindings]
+    this.bindings = ((isObject(bindings) && !bindings.toSQL) || isUndefined(bindings)) ?  bindings : [bindings]
```
jurko-gospodnetic: just noticed there are two whitespace after the `?` :smile:

Issue #707, opened 2015-02-24
Support query cancellation


whitelynx:
Not sure about the other database drivers, but node-postgres has [support for cancelling running queries](https://github.com/brianc/node-postgres/wiki/pg#cancelconfig-client-query). It would be nice to utilize [bluebird's cancellation logic](https://github.com/petkaantonov/bluebird/blob/master/API.md#cancellation) to allow queries to be cancelled via the promise mechanism.

whitelynx:
If that's not feasible at the moment, is there a way to access the node-postgres `client` and `query` objects so the query could be cancelled by user code?

tgriesser:
Curious, what sorts of situations are you looking to cancel queries - just so I understand the use case a little better... would it be based on some sort of timeout (query taking too long) or something else?

whitelynx:
Generally speaking, it'd be in response to a timeout; however, I've been looking at switching [my web-based PostgreSQL client](https://github.com/whitelynx/web-pgq) over to using knex, and in that case it'd be in response to a button press on the client side.

whitelynx:
Just curious... If I were to try and implement this, do you have any guidelines for adding a DB client API function that may only be implemented by some clients? I was thinking of at least implementing this in the PostgreSQL client, and we can work on adding it to other clients later.

elhigu:
There are no guidelines how to do this. 

Would be good to study which dialects could support the feature and then consider if it should be implemented in knex or separate library. 

API should be designed to be generally usable, but if some dialect does not support it using the feature should throw an exception...

whitelynx:
In [the Bluebird docs on Cancellation](http://bluebirdjs.com/docs/api/cancellation.html), where they describe the differences between the old 2.x cancellation semantics and the new 3.x approach, they say:

> The new cancellation has "don't care" semantics while the old cancellation had abort semantics. Cancelling a promise simply means that its handler callbacks will not be called.

_..._

> Note that the `onCancel` hook is really an optional disconnected optimization, there is no real requirement to register any cancellation hooks for cancellation to work. As such, any errors that may occur while inside the `onCancel` callback are not caught and turned into rejections.

_..._

[...]

PR #1454, opened 2016-05-26
`.timeout(ms, {cancel: true})` to cancel query after timeout (only implemented for MySQL)

byronmwong:
Taking a stab at addressing #707 

Adds an optional `options` argument to `.timeout()`.

For MySQL, a `KILL threadId` query will get sent. For all other dialects an error will be thrown immediately from the `.timeout()` call.

While writing tests, I discovered some potentially worrying behavior (comment also added in the test). After a timeout, `knex` releases the connection back to the pool, but the connection may not be usable because the original slow query is still being performed. A subsequent query will acquire the connection (still in-use) and hang until the first query finishes.

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -57,9 +57,16 @@ assign(Builder.prototype, {
     return cloned;
   },
 
-  timeout(ms) {
+  timeout(ms, {cancel} = {}) {
     if(isNumber(ms) && ms > 0) {
       this._timeout = ms;
+      if (cancel) {
+        if (this.client.canCancelQuery) {
+          this._cancelOnTimeout = true;
+        } else {
+          throw new Error("Query cancelling not supported for this dialect");
+        }
+      }
     }
     return this;
   },
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -61,11 +61,8 @@ assign(Builder.prototype, {
     if(isNumber(ms) && ms > 0) {
       this._timeout = ms;
       if (cancel) {
-        if (this.client.canCancelQuery) {
-          this._cancelOnTimeout = true;
-        } else {
-          throw new Error("Query cancelling not supported for this dialect");
-        }
+        this.client.assertCanCancelQuery();
+        this._cancelOnTimeout = true;
       }
     }
     return this;
--- a/src/raw.js
+++ b/src/raw.js
@@ -30,9 +30,16 @@ assign(Raw.prototype, {
     return this
   },
 
-  timeout(ms) {
+  timeout(ms, {cancel} = {}) {
     if(isNumber(ms) && ms > 0) {
       this._timeout = ms;
+      if (cancel) {
+        if (this.client.canCancelQuery) {
+          this._cancelOnTimeout = true;
+        } else {
+          throw new Error("Query cancelling not supported for this dialect");
+        }
+      }
     }
     return this;
   },
@@ -73,6 +80,9 @@ assign(Raw.prototype, {
     this._cached.options = reduce(this._options, assign, {})
     if(this._timeout) {
       this._cached.timeout = this._timeout;
+      if (this._cancelOnTimeout) {
+        this._cached.cancelOnTimeout = this._cancelOnTimeout;
+      }
     }
     if(this.client && this.client.prepBindings) {
       this._cached.bindings = this.client.prepBindings(this._cached.bindings || [], tz);
--- a/src/raw.js
+++ b/src/raw.js
@@ -34,11 +34,8 @@ assign(Raw.prototype, {
     if(isNumber(ms) && ms > 0) {
       this._timeout = ms;
       if (cancel) {
-        if (this.client.canCancelQuery) {
-          this._cancelOnTimeout = true;
-        } else {
-          throw new Error("Query cancelling not supported for this dialect");
-        }
+        this.client.assertCanCancelQuery();
+        this._cancelOnTimeout = true;
       }
     }
     return this;
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -2,9 +2,10 @@
 
 'use strict';
 
-module.exports = function(knex) {
+var Knex   = require('../../../knex');
+var _ = require('lodash');
 
-  var _ = require('lodash');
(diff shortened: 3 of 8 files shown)


src/query/builder.js
```diff
@@ -57,9 +57,16 @@ assign(Builder.prototype, {
     return cloned;
   },
 
-  timeout(ms) {
+  timeout(ms, {cancel} = {}) {
     if(isNumber(ms) && ms > 0) {
       this._timeout = ms;
+      if (cancel) {
```
wubzz: Unnecessary nested if-statements here. `if (cancel && this.client.canCancelQuery)`

src/query/builder.js
```diff
@@ -57,9 +57,16 @@ assign(Builder.prototype, {
...
+        if (this.client.canCancelQuery) {
+          this._cancelOnTimeout = true;
+        } else {
+          throw new Error("Query cancelling not supported for this dialect");
```
wubzz: Is this really needed here given the default client function? I would rather have the second argument ignored for other dialects and make a note of dialect support in documentation than duplicating code.

src/raw.js
```diff
@@ -30,9 +30,16 @@ assign(Raw.prototype, {
     return this
   },
 
-  timeout(ms) {
+  timeout(ms, {cancel} = {}) {
     if(isNumber(ms) && ms > 0) {
       this._timeout = ms;
+      if (cancel) {
```
wubzz: Unnecessary nested if-statements here. `if (cancel && this.client.canCancelQuery)`

src/raw.js
```diff
@@ -30,9 +30,16 @@ assign(Raw.prototype, {
...
+        if (this.client.canCancelQuery) {
+          this._cancelOnTimeout = true;
+        } else {
+          throw new Error("Query cancelling not supported for this dialect");
```
wubzz: Is this really needed here given the default client function? I would rather have the second argument ignored for other dialects and make a note of dialect support in documentation than duplicating code.

test/integration/builder/additional.js
```diff
@@ -299,7 +299,77 @@ module.exports = function(knex) {
...
+        }
+      };
+
+      if(!testQueries.hasOwnProperty(dialect)) {
+        throw new Error('Missing test query for dialect: ' + dialect);
+      }
+
+      var query = testQueries[dialect]();
+
+      var addTimeout = () => query.timeout(1, {cancel: true})
```
wubzz: Unfortunately we can't es6 in the test suite, it's being run for 0.10 and 0.12 as well.

wubzz:
I understand the issue here and don't mind the implementation. Would be great to support other dialects as well later on.

What I'm curious about is what would happen if the `cancelQuery` function is unable to `acquireConnection` due to the pool being full *(or no resources available)*.

byronmwong:
@wubzz thanks for the quick review!

Good point about `cancelQuery` not being to acquire a connection. I could put a timeout on the cancel part as well and throw a different error, e.g. `Unable to cancel query that hit timeout`. What do you think?

src/query/builder.js
byronmwong: The default client function won't throw until the query has been sent and already timed out. I felt it'd give a better dev experience if we threw earlier.

test/integration/builder/additional.js
byronmwong: :+1: will fix

wubzz:
@byronmwong Yeah some form of fallback might be needed. Perhaps the simplest solution would be to have the promise "resolve" itself after timing out. That way the original expected timeout error would still be thrown.

byronmwong:
@wubzz yep that could work, but then we're silently failing on the cancel part.

If a dev calls `.timeout(ms, {cancel: true})` and receives a `TimeoutError`, they'd expect the original query to be cancelled.

test/integration/builder/additional.js
```diff
@@ -256,9 +256,9 @@ module.exports = function(knex) {
     });
 
 
-    it('.timeout() should throw TimeoutError', function(done) {
+    it('.timeout() should throw TimeoutError', function() {
       var dialect = knex.client.config.dialect;
-      if(dialect === 'sqlite3') { return done(); } //TODO -- No built-in support for sleeps
+      if(dialect === 'sqlite3') { return; } //TODO -- No built-in support for sleeps
```
jurko-gospodnetic: you can make sqlite sleep implicitly by using two separate transactions/connections, one locks a database table, and another attempts to access that table, which will cause it to wait for the lock to be released (5 seconds by default, if I recall correctly)

src/query/builder.js
jurko-gospodnetic: Would it be possible to make the client-type-agnostic part of knex perform this check by having it provide a wrapper `timeout()` function that then calls this function only if the check passes? That would kill the duplication and still give us the immediate error feedback.

Haven't checked the relevant knex code though, so might me talking out of my ass here. :smile:

test/integration/builder/additional.js
byronmwong: ohh that's a good idea! @jurko-gospodnetic @wubzz is this a blocker on merging?

byronmwong:
@wubzz: added a timeout on acquiring connection for cancelling queries and got tests to pass. ready for another look!

wubzz:
Looks good to me. Given that the original timeout test was written by me, I don't think the responsibility of adding sqlite support in the test falls on you, so I don't think it's a blocker for this PR.

A quick search shows that `postgres`, `maria` and `oracle` should have support for this action as well. Is this something to look into in this PR, or would you rather a separate issue?

byronmwong:
Turns out the implementation for `maria` is the same as `mysql`, so I've
handled that. I'd prefer a separate issue for `postgres` and `oracle`
support. Thanks!

wubzz:
@byronmwong That's alright with me, I'll make a new issue for postgres/oracle support.

One last thing I almost forgot before I merge this, could you update the docs *(index.html)* of `timeout` to reflect this change? Make sure to specify it only applies to mysql/maria *for now*.

byronmwong:
done! no worries, I should've thought to update the docs myself.

Issue #1448, opened 2016-05-25
Throw error of undefined values in SELECT queries


wubzz:
Per the discussion in #1417:

**Explanation**
Currently `insert` and `update` handle `undefined` values via `valueForUndefined` -- this works great, and should remain the same. The issue arises with `select` queries since one certainly would not want `valueForUndefined` in this case. Throwing an error is the most optimal solution to prevent faulty application code and faulty queries.

The current implementation throws an error message `SQL queries with undefined where clause option` which  definitely gets the point across, but could be more informative. It's also currently only implemented for Postgres.

**Proposal**
I suggest we remove the existing validation check and move it out to the [compiler(Reference)](https://github.com/tgriesser/knex/blob/master/src/query/compiler.js#L37). In this function we have all the information we need: the query method, the bindings, and the generated query. With this we can validate **before** `prepBindings` is called and throw an informative error including information about the query being run. This would also be applied to all dialects since the compiler applies for all. We also have the ability to emit the `query-error` event from this function.

Thoughts? @elhigu @rhys-vdw @mdrmuhaimin

elhigu:
Sounds doable place. As far as I know that place in knex should catch all undefined value errors in bindings.

Though I would like to throw synchronous error already when query builder is called with invalid parameters.

About `query-error` event. In my opinion that should be emitted only for those errors which are happening on query execution time (i.e. after .then() was called), after the `query` event.

I would like to approach this issue by creating bunch of test cases with undefined values in API calls (also in other places than just where statements). In select, from, orderBy, inside sub query function s... and after having good tests coverage of potential use cases, check what is convenient way to solve the issue.

When tests are written we  can also document that undefined handling to documentation so that it won't be "undefined behaviour" anymore.

wubzz:
@elhigu I like that idea, we definitely need to test more with undefined values. I vote for giving it its own suite in both unit and integration tests.

`query-error` I had the same thought, it doesn't really fit in. However, we could have a generic `error` event or `builder-error` / `compiler-error`, something like that. There's no reason not to help the enduser when possible. It could then be emitted wherever we have these checks. Generally these events are good for logging purposes or within a query-builder chain.

wubzz:
Closing this as it has since been solved.

PR #1459, opened 2016-05-28
Check for undefined bindings in WHERE clauses and RAW queries during…

wubzz:
… compile and throw an error.

This is what I had in mind for #1448 . We mentioned an entire suite for testing `undefined` in the issue, but I feel this is a good starting point. The idea is to throw an error during compile if a `SELECT` query contains an undefined binding, or if a `RAW` query contains an undefined binding.

The check is currently recursive. Not sure if this is needed or not.

CC @elhigu @jurko-gospodnetic @rhys-vdw

### DIFF
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -3371,4 +3371,64 @@ describe("QueryBuilder", function() {
     })
   });
 
+
+  it('Any undefined binding in a SELECT query should throw an error', function() {
+    var expectedErrorMessageContains = 'Undefined binding(s) detected when compiling SELECT query:'; //This test is not for asserting correct queries
+    var qbuilders = [
+      qb().from('accounts').where({Login: void 0}).select(),
+      qb().from('accounts').where('Login', void 0).select(),
+      qb().from('accounts').where('Login', '>=', void 0).select(),
+      qb().from('accounts').whereIn('Login', ['test', 'val', void 0]).select(),
+      qb().from('accounts').where({Login: ['1', '2', '3', void 0]}),
+      qb().from('accounts').where({Login: {Test: '123', Value: void 0}}),
+      qb().from('accounts').where({Login: ['1', ['2', [void 0]]]})
+    ];
+    qbuilders.forEach(function(qbuilder) {
+      try {
+        //Must be present, but makes no difference since it throws.
+        testsql(qbuilder, {
+          mysql: {
+            sql: '',
+            bindings: []
+          },
+          oracle: {
+            sql: '',
+            bindings: []
+          },
+          mssql: {
+            sql: '',
+            bindings: []
+          },
+          postgres: {
+            sql: '',
+            bindings: []
+          }
+        });
+        expect(true).to.equal(false, 'Expected to throw error in compilation about undefined bindings.');
+      } catch(error) {
+        expect(error.message).to.contain(expectedErrorMessageContains); //This test is not for asserting correct queries
+      }
+    });
+  });
+
+
+  it('Any undefined binding in a RAW query should throw an error', function() {
+    var expectedErrorMessageContains = 'Undefined binding(s) detected when compiling RAW query:'; //This test is not for asserting correct queries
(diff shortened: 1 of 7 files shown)


test/unit/query/builder.js
```diff
@@ -3371,4 +3371,64 @@ describe("QueryBuilder", function() {
...
+      qb().from('accounts').where({Login: void 0}).select(),
+      qb().from('accounts').where('Login', void 0).select(),
+      qb().from('accounts').where('Login', '>=', void 0).select(),
+      qb().from('accounts').whereIn('Login', ['test', 'val', void 0]).select(),
+      qb().from('accounts').where({Login: ['1', '2', '3', void 0]}),
+      qb().from('accounts').where({Login: {Test: '123', Value: void 0}}),
+      qb().from('accounts').where({Login: ['1', ['2', [void 0]]]})
+    ];
+    qbuilders.forEach(function(qbuilder) {
+      try {
```
elhigu: Probably you could use http://chaijs.com/api/bdd/#method_throw here

```js
expect(function () {
... throwing code ...
}).to.throw(expected error message or regex checking error or something like that);
```

Do you think we should test here also for example if `.select(function () { this.where('id', undefined); })` or `.select(['col', undefined])` etc. works?

elhigu:
Code looks good, but is missing documentation.

wubzz:
@elhigu Yeah about that, I wasn't sure what part of docs to put this in. No existing section is good to place it, imo.

elhigu:
I agree... maybe new subsection somewhere under QueryBuilder (http://knexjs.org/#Builder), which explains about what parameters can be passed and where... E.g. where is ok to pass other query builder instances / subquery functions / knex.raw etc. that place could be ok also for telling that undefined should throw error always except with inserts.

For now it could contain only info about undefined, and when all functionality is well specified also other parameters could be explained there.

wubzz:
@elhigu I kind of feel like these types of heavy validation decisions should be given its own 'big' section so that it's not easy to miss. Essentially what this change does is having the user check their data before calling the library, which is something you'd want to know sooner rather than later.

What about something like

>###  General information
Lorem ipsum...
##### Undefined values in INSERT/Update
Lorem ipsum...
##### Undefined values in SELECT / Raw
Lorem ipsum...


Put it under the *Installation* section or something.

wubzz:
@elhigu So with this we've covered `INSERT`, `SELECT` and `RAW`. Do we do anything for `UPDATE` ?

elhigu:
@wubuzz documentation says that with update undefined value is ignored

wubzz:
@elhigu Yes, that's what it says now, but is there any reason to keep treating it differently?

In my eyes the two scenarios below are both equally dangerous.

1. Run a `SELECT` query to load rows to process programatically, with an undefined `where` binding.

> knex('account').select().where({col: undefined});
//ERROR! Undefined binding(s) detected when compiling SELECT query

2. Run an `UPDATE` query with an undefined `where` binding.

>knex('account').where({login: undefined}).update({password: '123'}).toString();
//update "account" set "password" = '123'

elhigu:
@wubzz Uh I thought you meant that how `.update({col1: 1, col2: undefined })` is handled, where it kind of makes sense to ignore value

But yeah if anywhere in `.where('col1', undefined)` in insert / select / update / delete / etc. then the same error should be thrown. Dropping part of where statement is equally dangerous everywhere.

wubzz:
Then essentially the better place to put this would be in the compiler's `where` function instead. Makes the validation global. I'll try that.

wubzz:
@elhigu Added docs under "where clauses" -- seemed most appropiate. Ready to merge if you're ok with it.

peralmq:
Just wanted to say thanks for adding this! Highlighted a bug in our system that we were then able to track down and fix. :)

aspyatkin:
If you try to save Buffer(0) into a binary column, it will fail since internally it's treated as undefined

Issue #1388, opened 2016-05-06
Warning: a promise was created in a handler but not returned from it


svileng:
I'm trying out knex and migrations for the first time, and I'm getting a few of those after I run `migrate:latest`:

```
Warning: a promise was created in a handler but was not returned from it
    at Runner.ensureConnection (/Users/Svilen/Development/testapp/node_modules/knex/lib/runner.js:156:26)
    at Runner.run (/Users/Svilen/Development/testapp/node_modules/knex/lib/runner.js:29:31)
    at QueryBuilder.Target.then (/Users/Svilen/Development/testapp/node_modules/knex/lib/interface.js:26:43)
    at Migrator._isLocked (/Users/Svilen/Development/testapp/node_modules/knex/lib/migrate/index.js:201:76)
    at /Users/Svilen/Development/testapp/node_modules/knex/lib/migrate/index.js:217:23
    at /Users/Svilen/Development/testapp/node_modules/knex/lib/transaction.js:42:18
From previous event:
    at /Users/Svilen/Development/testapp/node_modules/knex/lib/transaction.js:36:8
From previous event:
    at new Transaction (/Users/Svilen/Development/testapp/node_modules/knex/lib/transaction.js:29:27)
    at Client.transaction (/Users/Svilen/Development/testapp/node_modules/knex/lib/client.js:118:12)
    at Function.transaction (/Users/Svilen/Development/testapp/node_modules/knex/lib/util/make-knex.js:60:21)
    at Migrator._getLock (/Users/Svilen/Development/testapp/node_modules/knex/lib/migrate/index.js:216:24)
    at Migrator._runBatch (/Users/Svilen/Development/testapp/node_modules/knex/lib/migrate/index.js:241:19)
    at /Users/Svilen/Development/testapp/node_modules/knex/lib/migrate/index.js:74:25
    at tryOnImmediate (timers.js:534:15)
    at processImmediate [as _immediateCallback] (timers.js:514:5)
From previous event:
    at /Users/Svilen/Development/testapp/node_modules/knex/lib/migrate/index.js:73:12
From previous event:
    at Migrator.rollback (/Users/Svilen/Development/testapp/node_modules/knex/lib/migrate/index.js:69:28)
    at Command.<anonymous> (/Users/Svilen/Development/testapp/node_modules/knex/lib/bin/cli.js:114:37)
    at Command.listener (/Users/Svilen/Development/testapp/node_modules/commander/index.js:301:8)
[...]

PR #1460, opened 2016-05-30
Fixes "a promise was created but was not returned" warning

agdevbridge:
This would fix Warning: a promise was created in a handler but not returned from it
Issue: #1388

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -52,6 +52,7 @@ function Transaction(client, container, config, outerTx) {
           transactor.rollback(err)
         })
       }
+      return null;
     })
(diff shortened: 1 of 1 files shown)


kolpav:
Hope this will be merged soon

elhigu:
I was thinking if just return would have been enough to keep previous functionality... but this seems harmless though so I'll just merge.

teh-botol:
You forgot to put `return` in here: (line 49 & 52)
```
result.then((val) => {
          return transactor.commit(val)
        })
        .catch((err) => {
          return transactor.rollback(err)
        })
```

PR #1464, opened 2016-06-01
Fixes for mssql dialect columnInfo and binary fields

statyan:


rhys-vdw:
Hey @statyan, thanks for the PR. Sorry to take so long!

Issue #240, opened 2014-04-10
Support for mssql


tybenz:
Wondering if there are any plans to support MSSQL. Could take advantage of a MSSQL library like this one: "https://github.com/patriksimek/node-mssql".

I'd be happy to work on a pull request for this, but I might need a few pointers on how to get started.

tjwebb:
any update on mssql? I'm building https://github.com/cnect/sails-mssql without the benefit of a real MSSQL query builder for node. Would love to use knex under the hood.

bendrucker:
We basically need someone to take the lead on either writing and maintaining the functionality or sponsoring it. Otherwise there are a lot of higher priority items and you shouldn't count on MS SQL support.

tjwebb:
Yea, understand. I'm already invested in the sails mssql adapter, but I'd love to circle back to a knex implementation after the immediate needs of my project are met.

tgriesser:
Yeah absolutely, @tjwebb that'd be great if you could help out with something like this - I'm mapping out a lot of plans for how to clean up a lot of the query composition so it should be much more straightforward on adding new dialects, etc. 

Would also be cool if we could somehow get knex as a backing for sails at some point, I know from that other ticket you were looking to make that a reality - hopefully once the internals of knex are a bit more modular we can figure out something to make that happen.

tjwebb:
@tgriesser Sounds good. If you run across any sponsored dev for mssql
support, let me know.

emanuelhp:
Hi!

I started an implementation of mssql in https://github.com/tlsoftwares/knex, who can and want to help will be very welcome!
If you have any guidance on how to do this work would be helpful.
I started integrating the dialect "mssql" the possibilities of testing and I am already doing the database connection using https://github.com/pekim/tedious.

whitelynx:
@emanuelhp Thanks for working on this! I'll have to check out your fork soon. I did have a question, though... would you be opposed to using https://www.npmjs.com/package/mssql instead of using tedious directly? That library wraps tedious, [the official Microsoft SQL Server client for Node](https://www.npmjs.com/package/msnodesql), and the older [node-tds](https://www.npmjs.com/package/tds) library, which means it'd be easy for users to switch between drivers. (especially since the official MS driver should outperform the rest, but isn't available on Linux)

whitelynx:
[...]

PR #1472, opened 2016-06-06
Add mssql to the list of supported database libraries

alarner:
When looking for MSSQL support it wasn't immediately clear that it was available within Knex. I was happy to find issue #240 but thought it should be listed as a supported database library in the docs.


elhigu:
Actually this adds mssql driver installation instructions, info about mssql being supported platform was added while ago to other parts of documentation, This is good addition anyways, thanks!

PR #1486, opened 2016-06-10
Fix invalid SQL in Oracle dialect ping function

jacobinary:
"SELECT 1" results in the error: ORA-00923: FROM keyword not found where expected

The correct syntax should be "SELECT 1 FROM DUAL".


wubzz:
Makes sense this wasn't caught in CI since there's no runners for Oracle (or mssql) integration, only unit.

Cheers for this @jacobinary !

Issue #1470, opened 2016-06-05
Seed file stub is buggy


elhigu:
Seed file stub is using Promise.join in a wrong way which causes undefined behavior due to race conditions when sending queries... could be that it works when all queries is sent to same connection but actually nothing in Promise.join guarantees order in which promises are resolved...

#1243 Has some more discussion about it.

PR #1493, opened 2016-06-14
:bug: Update seed stubs to run serially.

dannyfritz:
I don't know the other 3 languages very well, so I did my best guess at what their equivalents would be.

I was also unable to run tests because I couldn't get some of the db drivers to install on Windows.

This is a fix for #1470


rhys-vdw:
Thanks mate.

PR #1498, opened 2016-06-14
Made test timeout to be configurable by setting KNEX_TEST_TIMEOUT env…

elhigu:
…ironment variable.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3096,6 +3096,15 @@ $ npm test
       replacing with the path to your config file, and the config file is valid, the test suite should run properly.
     </p>
 
+      <b class="header">My tests are failing because slow DB connection and short test timeouts! How to extend test timeouts?</b><br />
+      Some times e.g. with running CI on travis test suites default timeout 5seconds might be too short. The test suite looks from an environment variable 
+      called <tt>KNEX_TEST_TIMEOUT</tt> for milliseconds how long each test might run before timing out.
+    </p>
+<pre>
+$ export KNEX_TEST_TIMEOUT=30000
+$ npm test
+</pre>
+
     <p id="faq-nonode">
       <b class="header">Can I use Knex outside of Node.js</b><br />
       Yes. While the WebSQL spec is deprecated, there is still an adapter that provides support.
--- a/index.html
+++ b/index.html
@@ -3096,9 +3096,10 @@ $ npm test
       replacing with the path to your config file, and the config file is valid, the test suite should run properly.
     </p>
 
+    <p>
       <b class="header">My tests are failing because slow DB connection and short test timeouts! How to extend test timeouts?</b><br />
-      Some times e.g. with running CI on travis test suites default timeout 5seconds might be too short. The test suite looks from an environment variable 
-      called <tt>KNEX_TEST_TIMEOUT</tt> for milliseconds how long each test might run before timing out.
+      Sometimes, e.g. when running CI on travis, test suite's default timeout of 5 seconds might be too short. In such cases an alternative test timeout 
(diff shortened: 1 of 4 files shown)


elhigu:
✓ accepts the table name, and a "container" function (7075ms) 👍

test/index.js
```diff
@@ -18,6 +18,8 @@ global.d      = new Date();
 Promise.longStackTraces();
 
 describe('Query Building Tests', function() {
+  this.timeout(process.env.KNEX_TEST_TIMEOUT || 5000);
```
jurko-gospodnetic: I have not tested this, but would we be able to kill the duplication by making this call in a global `beforeEach()` call? As I recall, mocha has a top-level implicit anonymous test suite containing everything.

elhigu: I had no idea about that implicit top level. I'll try it out 👍

elhigu: Adding beforeEach() outside of describes was ran before every test, but setting timeout there didn't work:

```js
/*global describe*/

'use strict';

require('source-map-support').install();

global.sinon = require("sinon");

var chai = global.chai = require("chai");

chai.use(require("sinon-chai"));
chai.should();

var Promise   = global.testPromise = require('../lib/promise');
global.expect = chai.expect;
global.d      = new Date();

Promise.longStackTraces();

beforeEach(function () {
  console.log("This log is ran before every test, but setting timeout here has no effect!");
  this.timeout(process.env.KNEX_TEST_TIMEOUT || 5000);
})

describe('Query Building Tests', function() {
  require('./unit/query/builder')
  require('./unit/schema/mysql')('mysql')
  require('./unit/schema/mysql')('maria')
  require('./unit/schema/mysql')('mysql2')
  require('./unit/schema/postgres')
  require('./unit/schema/sqlite3')
  require('./unit/schema/oracle')
  require('./unit/schema/mssql')
})

describe('Integration Tests', function() {
  require('./integration')
})
```

jurko-gospodnetic: Doh... ah well... thanks for trying. :+1:

index.html
```diff
@@ -3096,6 +3096,15 @@ <h2 id="faq">F.A.Q.</h2>
       replacing with the path to your config file, and the config file is valid, the test suite should run properly.
     </p>
 
+      <b class="header">My tests are failing because slow DB connection and short test timeouts! How to extend test timeouts?</b><br />
+      Some times e.g. with running CI on travis test suites default timeout 5seconds might be too short. The test suite looks from an environment variable 
+      called <tt>KNEX_TEST_TIMEOUT</tt> for milliseconds how long each test might run before timing out.
```
jurko-gospodnetic: Suggested doc text update:

> Sometimes, e.g. when running CI on travis, test suite's default timeout of 5 seconds might be too short. In such cases an alternative test timeout value in milliseconds can be specified using the <tt>KNEX_TEST_TIMEOUT</tt> environment variable.

(hmmm, I already wrote up this comment before but it's not here now... must not have confirmed it or something...)

elhigu: done, also fixed one missing paragraph starter tag

jurko-gospodnetic:
:+1:

test/index.js
wubzz: According to [Mocha Docs](http://mochajs.org/#timeouts) calling `.timeout(ms)` in a describe will apply that timeout for all nested tests in that suite. So maybe it'll work by adding `.timeout()` under `describe('Query Building Tests')` / `describe('Integration Tests')` ?

I know it's already merged, just figured I'd mention. :smile:

Issue #1491, opened 2016-06-14
TravisCi add oracle Database


williamgueiros:
hello, I'm  working to add Oracle in TravisCi on the project, for running integration tests.

See the result in link, please add comments if you dislike.

 https://github.com/williamgueiros/knex/pull/1

do you think is possible, add environment variables(ORACLE_LOGIN_ssousername, ORACLE_LOGIN_password) on your ThavisCI?

![captura de tela 2016-06-13 as 19 43 32](https://cloud.githubusercontent.com/assets/19972/16026706/b568da88-31a6-11e6-931f-2d95e0c74103.png)

see more information in:
https://github.com/cbandy/travis-oracle

williamgueiros:
PR after 

https://github.com/tgriesser/knex/pull/990

elhigu:
Looks interesting, I'm very excited of this 👍  

Even though I suppose that anyone can extract those login / password from environment configuration by sending pull request which prints them to build log as plaintext. Which doesn't really matter since it is only a test database and nothing serious.

If this works it would be great reference to achieve the same for mssql.

williamgueiros:
@elhigu ok after that, i will start MSSQL feature.

kibertoad:
PR was merged, issue can be resolved.

PR #990, opened 2015-09-22
add node-oracledb support

atiertant:
this look like working,could someone test?
thanks @KhalilTN

PR #1503, opened 2016-06-15
add travisCi to oracle feature

williamgueiros:
see more information on #1491 and #990

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 # .travis.yml
-sudo: false
+sudo: required
 
 language: node_js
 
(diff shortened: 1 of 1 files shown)


williamgueiros:
@elhigu  please add environment variables

if you prefere add all, using don't display information on console.

elhigu:
@tgriesser @rhys-vdw Looks like I don't have enough rights to set those variables. Could you add them?

vschoettke:
I think pull requests will not be tested if the TravisCI script requires environment variables (because somebody might just add an echo of that variable). The best way would be for Oracle to supply a public url for downloading Oracle XE or any other mean (e.g. Docker image)

williamgueiros:
@vschoettke i 💚  the ideia.
but angry to the oracle license 👎 , i don't remember there i read that, we do not have permission to copy and distribution oracle XE

rhys-vdw:
Hey guys, it seems I am able to modify environment variables, but could somebody guide me through the process? Do I need to create an Oracle account for knex?

rhys-vdw:
> I think pull requests will not be tested if the TravisCI script requires environment variables

Yes, you're correct - thanks for the heads up. Not ideal.

On the bright side a local branch [succeeds](https://travis-ci.org/tgriesser/knex/builds/138286598)!

williamgueiros:
@rhys-vdw  I ❤️ the idea @vschoettke 

but the problem to me, is the oracle license.

@vschoettke do you start this feature ??

rhys-vdw:
Okay, so I want to get this merged even if we don't support Oracle or PR's. A solution like @vschoettke's suggestion would be ideal, but that shouldn't prevent us from running oracle tests on branches of this repository.

In the short term I think we must skip Oracle integration testing in PR's. This means we's need to merge into a local branch to see if PR's pass Oracle integration tests in CI. Still an improvement over the current state of things.

If I understand correctly, this new environment variable `DIALECT` means we end up running all our integration tests twice. We already support listing dialects using the [`DB` environment variable](https://github.com/tgriesser/knex/blob/b21156d92085a859efc2d77f8cc606bf41d781cf/test/knexfile.js#L9). Since this PR hasn't modified `DB` or its default value we are not actually running the tests (including in my the successful branch I linked to [above](https://github.com/tgriesser/knex/pull/1503#issuecomment-226701058)).

It is possible to check whether the [current branch is a pull request](https://docs.travis-ci.com/user/pull-requests#Security-Restrictions-when-testing-Pull-Requests). Perhaps we can use this to conditionally skip some tests?

The logic could be something like this:

1. If we are *not* testing a pull request, include `oracle` in the DB environment variable along with the other dialects.
2. If `oracle` is included in `DB` then install download an install Oracle as per this PR's implementation.

This way we'll have a single build per node version, conditionally testing Oracle if available.

Thoughts?

elhigu:
@williamgueiros Actually oracle license seems really permissive http://www.oracle.com/technetwork/licenses/database-11g-express-license-459621.html It allows package DB with your own application and it allow to redistribute it too if you require your users to agree with the original license.

tgriesser:
This would be awesome to get in here - @williamgueiros @vschoettke what can I do to get this moving forward. Also while I'm here, for those who know oracle well, can anyone tell me what's supposed to be [going on here](https://github.com/tgriesser/knex/blob/53d089bffbde6e16aa79652d258154a67dec8c0a/src/dialects/oracledb/query/compiler.js#L70-L79). 

That `usingClause` value is never actually used anywhere, the newest eslint caught that.

PR #1504, opened 2016-06-16
Fix code blocks.

blainesch:
Before:
knex_js_-_a_sql_query_builder_for_javascript

After:
knex_js_-_a_sql_query_builder_for_javascript


elhigu:
🙇

Issue #1502, opened 2016-06-15
Request: unique hash on query objects for logging


funnylookinhat:
I'm trying to write a simple metrics tool to measure which queries are performing slowly so that I can evaluate my codebase over time.  As a simplified example:

```
var knex = require('knex')({myConfig});

var runningQueries = {};

knex.on('query', function (data) {
  // Generates a unique has using hash using keys like:
  // ['__knexUid', 'method', 'sql', 'bindings', 'options']
  var queryHash = generateHash(data);
  runningQueries[queryHash] = {
    sql: data.sql,
    start: Date.now()
  };
});

knex.on('query-response', function (response, obj, builder) {
  // This hash should match the one in the function above.
  var queryHash = generateHash(obj);
  var runTime = Date.now() - runningQueries[queryHash].start;
  // etc. etc.
  delete runningQueries[queryHash];
});
```

The problem I'm encountering is that multiple queries with the same properties will get jumbled together - and could potentially screw up my timings or cause me to miss queries.  It would be ideal if each query object had a unique hash attached to it (or the Date.now() timestamp it was created at) to identify them uniquely.

I can add the code, but before I submit a PR I'd like your input and to make sure I'm not wasting your time.  :)

elhigu:
Hmm.. isn't __knexUid set for each connection and is unique in that way? Each connection can be running only one query at the time. 

So it should be always that when `query` event comes for certain __knexUid you can attach your own UUID for query in 'query' event and then when next `query-response` or `query-error` event comes you know that it came for the query which was executed last?

My implementation for timing queries was like this:

```js
'use strict';

var timers = {};

function showTime(timerId, opts) {
  var failed = opts && opts.failed;
  var precision = 3;
  var timer = timers[timerId];
  if (!timer) { 
    return; 
  }
  var elapsedHrtime = process.hrtime(timer.start);
  var elapsed = elapsedHrtime[1] / 1000000;
  var milliseconds = elapsedHrtime[0]*1000 + elapsed;
  if (milliseconds > opts.threshold) {
    console.log(failed ? "Slow query FAIL" : "Slow query SUCCESS", milliseconds.toFixed(precision) + " ms - " + timer.query);
  }
  delete timers[timerId];
}

module.exports = function (app, config) {
  app.knex().on('query', function (toSql) {
    var queryString = `${toSql.sql} ${JSON.stringify(toSql.bindings)}`;
    timers[toSql.__knexUid] = {
      start: process.hrtime(),
[...]

PR #1510, opened 2016-06-17
Add query-specific uuid to enable finite metrics.

funnylookinhat:
When trying to profile queries throughout an application, binding to `query`, `query-response`, and `query-error` isn't quite as useful as it could be because there's no way to uniquely identify individual queries that are sent off to the connection.

This adds a (hopefully appropriately) named field of `__knexQueryUid` to both raw and query-builder queries.

For example:

```
knex.on('query', function (obj) {
  // Start profiling obj.__knexQueryUid
});
knex.on('query-response', function (res, obj) {
  // Record metrics for obj.__knexQueryUid
});
knex.on('query-error', function (err, obj) {
  // Record metrics for obj.__knexQueryUid
});
```

This should hopefully resolve this feature request: https://github.com/tgriesser/knex/issues/1502

I didn't see any tests specific to the events that are emitted when running queries - so I'm not sure if/where I should add tests for this PR.


wubzz:
The tests for the events are awkwardly placed [here](https://github.com/tgriesser/knex/blob/master/test/integration/builder/additional.js), in case you wanted to write some.

funnylookinhat:
@wubzz Ah thanks :)  I've added what I think should be necessary now.

funnylookinhat:
@wubzz Let me know if you have any feedback.  Sorry to spam the PR - but I couldn't get the tests working locally so I just abused Travis.  :-P

PR #1526, opened 2016-06-22
Fix #906 Oracle auto increment manual insert id

### LINKED ISSUE #906: oracle id sequence trigger does not handle manual id inserts
Currently increment ids in oracle are working using trigger and a sequence, but if you manually insert a record with an ID, the sequence is not being updated. So if I insert, say, something with ID:4 into an empty table, next insert will be with ID:2, not ID:5. I think the trigger should be changed so that it also sets the lastId in sequence as max id from inserted records.

atiertant:
fix https://github.com/tgriesser/knex/issues/906


elhigu:
Pretty cryptic queries, which I don't understand, but looks fine, since you are the guy who knows oracle stuff :) Just needs rebase. Ping me after rebase so that I'll merge this asap.

atiertant:
this is trigger executed before insert:
it count the number of rows where the primary key is equal to next value of the sequence and redo while this count is not zero.

i'll rebase it on monday

PR #1538, opened 2016-06-29
Transactions Rejecting with Undefined

funnylookinhat:
If you create a transaction and throw an error before committing it, you will get an `Unhandled rejection undefined` error.

I can reproduce it with the following.  You'll notice that the console will output `Unhandled rejection undefined` when running.  In certain instances (e.g. with complex nested promises handled by Bluebird), you can also get a `Warning: a promise was rejected with a non-error: [object Undefined]`

```
Bluebird.resolve()
.then ->
  new Bluebird (resolve, reject) ->
    knex.transaction (trx) ->
      Bluebird.resolve()
      .then ->
        console.log "foo"
      .then ->
        throw new Error "Some error"
      .then ->
        trx.commit()
      .then ->
        resolve "bar"
      .catch (internalErr) ->
        trx.rollback()
        .catch (rollbackErr) ->
          console.log "Rollback Error: #{rollbackErr}"
        .then ->
          reject internalErr
.then (result) ->
  console.log "Done: #{result}"
.catch (err) ->
  console.log "Error: #{err}"
.then ->
  process.exit 0
```

I can confirm that this fixes the issue in the code that I'm working on (proprietary) but also in the sample above.  I can't figure out a way to write a unit test around this as it's being printed from the Bluebird library.

### DIFF
--- a/src/dialects/mysql/transaction.js
+++ b/src/dialects/mysql/transaction.js
@@ -31,6 +31,8 @@ assign(Transaction_MySQL.prototype, {
       })
       .tap(function() {
         if (status === 1) t._resolver(value)
+        if (typeof value === "undefined")
+          value = new Error("Transaction result was undefined.");
         if (status === 2) t._rejecter(value)
       })
     if (status === 1 || status === 2) {


tgriesser:
So when you're working with transactions, you want to either call `trx.commit` / `trx.rollback`, **or** return a promise, not both. Since coffeescript has implicit returns, you're both returning a promise and calling `commit` / `rollback`. Also it seems like you have way more created promises / `.then` / `.catch` than you normally should.
Also

funnylookinhat:
@tgriesser Sorry, I should have been more explicit.  I get the same return even if I don't automatically return the statement... e.g.

```
Bluebird.resolve()
.then ->
  new Bluebird (resolve, reject) ->
    knex.transaction (trx) ->
      Bluebird.resolve()
      .then ->
        console.log "foo"
      .then ->
        throw new Error "Some error"
      .then ->
        trx.commit()
      .then ->
        resolve "bar"
      .catch (internalErr) ->
        trx.rollback()
        .catch (rollbackErr) ->
          console.log "Rollback Error: #{rollbackErr}"
        .then ->
          reject internalErr
    return 'whatever'
.then (result) ->
  console.log "Done: #{result}"
.catch (err) ->
  console.log "Error: #{err}"
.then ->
  process.exit 0
```

I know the use-case seems a bit... unique.  It's my attempt at trying to represent a fairly complex promise-chain I'm using.  I didn't have any warnings about this issue on knex 0.10 - but 0.11 seemed to make them show up all of a sudden.

In either case, I think the PR is still valid - right?  I don't see why you would want to reject with undefined.

tgriesser:
But you are currently rejecting the transaction with `undefined`... which isn't supported behavior, you should be passing an error object to `trx.rollback`. What happens if you change `trx.rollback()` to `trx.rollback(rollbackErr)`. 

Also the indentation in the code there is off, the return value "whatever" you added is being returned to the `new Bluebird` block not the `knex.transaction` block.

Ideally, your code would be something more like this:

```coffee
knex.transaction (trx) ->
   trx.insert(values).into('sometable')
     .then ->
         # some other query here
     .then -> 
        # some other query here, which gives the val below
.then (val) ->
  console.log "Done: with a final resolved value of #{val}"
.catch err -> 
  console.log "Failed: rejected with #{val}" 
```

You shouldn't really need to catch the errors inside of the transaction, since the promise chain will commit / rollback as appropriate.

funnylookinhat:
Ah wow - ok.  You are right - I was passing a blank to .rollback().

Thank you for your patience and help!

PR #1263, opened 2016-03-09
Redshift module

krainboltgreene:
Currently missing tests, still need to remove some things

PR #1572, opened 2016-07-11
Create a new Redshift dialect that inherits from Postgres.

trevorsmith:
I've implemented a dialect for Amazon Redshift inspired by @krainboltgreene's work in #1263. As suggested by @rhys-vdw, this Redshift dialect inherits from Postgres, which results in much less code.

Only a few modifications were necessary:
- Unsupported types have been remapped. I based my choices on several sources, including https://www.flydata.com/resources/flydata-sync/data-type-mapping/, http://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-datatypes.html, http://docs.aws.amazon.com/redshift/latest/dg/c_Supported_data_types.html, and other official documentation.
- SERIAL columns have been replaced with IDENTITY columns.
- The query compiler nullifies the RETURNING clause normally added to Postgres queries.
- A small tweak to the Postgres dialect was made to use `current_schema()` instead of `current_schema`. Redshift treats it as a column name instead of a function if the parentheses aren't added.

The unit tests from Postgres have been copied and modified for Redshift. I attempted to add integration tests, but they seem to be keyed on the driver name (e.g. 'postgresql') instead of the dialect. I don't know know if this can be worked around without changing the structure of the tests.

We're currently using this to initialize new Redshift clusters (new databases/schemas/users/permissions etc.) and issue a range of SELECT, UPDATE, DELETE, and ALTER queries with no trouble.

### DIFF
--- a/test/knexfile.js
+++ b/test/knexfile.js
@@ -117,6 +117,18 @@ var testConfigs = {
     seeds: seeds
   },
 
+  redshift: {
+    dialect: 'redshift',
+    connection: testConfig.redshift || {
+      adapter:  "postgresql",
+      database: "knex_test",
+      user:     "knex_test"
+    },
+    pool: pool,
+    migrations: migrations,
+    seeds: seeds
+  },
+
   sqlite3: {
     dialect: 'sqlite3',
     connection: testConfig.sqlite3 || {
--- a/src/dialects/postgres/query/compiler.js
+++ b/src/dialects/postgres/query/compiler.js
@@ -82,7 +82,7 @@ assign(QueryCompiler_PG.prototype, {
       sql += ' and table_schema = ?';
       bindings.push(this.single.schema);
     } else {
-      sql += ' and table_schema = current_schema';
+      sql += ' and table_schema = current_schema()';
     }
 
     return {
--- a/src/dialects/postgres/schema/compiler.js
+++ b/src/dialects/postgres/schema/compiler.js
@@ -19,7 +19,7 @@ SchemaCompiler_PG.prototype.hasTable = function(tableName) {
     sql += ' and table_schema = ?';
     bindings.push(this.schema);
   } else {
-    sql += ' and table_schema = current_schema';
+    sql += ' and table_schema = current_schema()';
   }
 
   this.pushQuery({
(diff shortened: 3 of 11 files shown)


krainboltgreene:
If knex has the ability to do per-function logs, I would seriously suggest creating warning logs for index, unique, primary key, and special root-only functions/constants.

trevorsmith:
Redshift does not enforce constraints, but it does use them for query planning: http://docs.aws.amazon.com/redshift/latest/dg/t_Defining_constraints.html

As for creating indexes and using leader node functions and constants, Redshift already returns descriptive errors. Is it necessary to replicate this in knex?

krainboltgreene:
@trevorsmith I don't believe I ever remember Redshift giving a descriptive error?

trevorsmith:
Well, I suppose it's not terribly specific, but it does say "<whatever function or statement> is not supported on Redshift tables." I certainly can add a warning in the table compiler when .index() is called.

The only leader node-only function knex appears to implement is `CURRENT_TIMESTAMP` through the `knex.fn.now()` helper, which has no awareness of the current dialect, as well as the hardcoded call in `TableBuilder.timestamps()`. The latter should probably be moved into the ColumnBuilder so the Redshift dialect can use `SYSDATE` or `GETDATE()` instead.

krainboltgreene:
Good point, I thought there were a few more.

trevorsmith:
@rhys-vdw What are your thoughts on moving the creation of timestamp columns from `TableBuilder` to `ColumnBuilder`?

jweinsteincbt:
👍

trevorsmith:
Have any decisions been made on merging this PR? We'd love to move off our forked version of knex.

Our current worker cluster isn't using this adapter for heavy writes (Kinesis is a much better choice for that) but we do use it heavily to initialize/teardown schemas, tables, and to programmatically load and unload data.

test/knexfile.js
```diff
@@ -117,6 +117,18 @@ var testConfigs = {
     seeds: seeds
   },
 
+  redshift: {
+    dialect: 'redshift',
+    connection: testConfig.redshift || {
+      adapter:  "postgresql",
+      database: "knex_test",
+      user:     "knex_test"
```
zachelrath: Shouldn't the user here be `postgres`, if it's using the same user as the postgres configuration?

zachelrath:
+1 This would be very useful for my team as well!

elhigu:
Closing in favor of https://github.com/tgriesser/knex/pull/2233

PR #1574, opened 2016-07-12
fix: allow oracledb connectString (issue #1573)

### LINKED ISSUE #1573: oracledb connectString not possible
As [documented by node-oracledb](https://github.com/oracle/node-oracledb/blob/master/doc/api.md#-81-connection-strings) the oracledb driver supports some "exotic" connection strings.

The simplest ones resemble a URL, like the knex connectionParser stuff, but the more exotic ones do not.

In particular, I need to be able to reference entries in the `tnsnames.ora` file, as [described here](https://github.com/oracle/node-oracledb/blob/master/doc/api.md#-812-net-service-names-for-connection-strings). The result is that the config passed to the oracledb driver should look like:
```
{
  user: 'dfdf',
  password: 'dfdf',
  connectString: 'nameOfConnectionInTnsNames'
}
```

The current implementation (big thanks to @atiertant for his work on this!) does not allow this. In fact, it has hardcoded the `connectString` field to be the concatination of the host and database, see [here](https://github.com/tgriesser/knex/blob/master/src/dialects/oracledb/index.js#L62)

I purpose that users should be able to define the connectString parameter in the `connection` object passed to knex. That should allow even the most exotic oracle connectStrings

I'll make a PR for this.

codyzu:
fixes #1573 

*Question:*

I'm not sure where I can add tests for this... There are tests for the `connectionParser` but my fix is in the oracledb dialect directly.  Should I add a new test file that instantiates this client and tests the parsing? Maybe mocking the actual oracledb package? Seems a little complicated, but let me know what you guys think...

Side note, I've tested this locally and it works, while maintaining backwards compatibility.

ping @rhys-vdw @elhigu because I see you guys were involved in the oracledb merge :-)

### DIFF
--- a/src/dialects/oracledb/index.js
+++ b/src/dialects/oracledb/index.js
@@ -59,7 +59,8 @@ Client_Oracledb.prototype.acquireRawConnection = function() {
     client.driver.getConnection({
       user: client.connectionSettings.user,
       password: client.connectionSettings.password,
-      connectString: client.connectionSettings.host + '/' + client.connectionSettings.database
+      connectString: client.connectionSettings.connectString ||
+        (client.connectionSettings.host + '/' + client.connectionSettings.database)
     }, function(err, connection) {
       if (err)
         return rejecter(err);


codyzu:
ping @atiertant  also

ecowden:
We have a need for this as well. Any update?

Trying this locally, I can verify that it works.

codyzu:
We've been using this in production without any issues since I made the pull request (about a month). I would love to see this merged too because it is the last thing stopping us from using the knex package from npm...

pataltechnologies:
Plus one for getting this merged in as well. We have a need for this in our projects.

elhigu:
To me this looks fine and in #1573 @atiertant did kind of give his blessing for this, so I'll just merge this now 👍  Thanks @codyzy !

PR #1575, opened 2016-07-13
memsql support added

dvv:
Please, consider applying

### DIFF
--- a/src/dialects/memsql/index.js
+++ b/src/dialects/memsql/index.js
@@ -1,0 +1,29 @@
+
+// memsql Client
+// -------
+import inherits from 'inherits'
+import { assign } from 'lodash'
+
+import Client_MySQL2 from '../mysql2'
+import ColumnBuilder from './schema/columnbuilder'
+import ColumnCompiler from './schema/columncompiler'
+import QueryCompiler from './query/compiler'
+
+function Client_MemSQL(config) {
+  Client_MySQL2.call(this, config)
+}
+inherits(Client_MemSQL, Client_MySQL2)
+
+assign(Client_MemSQL.prototype, {
+
+  driverName: 'memsql',
+
+  ColumnBuilder,
+
+  ColumnCompiler,
+
+  QueryCompiler,
+
+})
+
+export default Client_MemSQL
--- a/src/dialects/memsql/query/compiler.js
+++ b/src/dialects/memsql/query/compiler.js
@@ -1,0 +1,21 @@
+
+import inherits from 'inherits'
+import { assign } from 'lodash'
+
+import QueryCompiler from '../../mysql/query/compiler'
+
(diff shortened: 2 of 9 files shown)


elhigu:
Would it be possible to add travis configuration so that this dialect would be CI tested? I wouldn't like to include any more dialects which are not tested automatically (oracle and mssql are so popular that they are currently exceptions).

dvv:
Isn't https://travis-ci.org/tgriesser/knex/builds/144431340 green is what you're asking?

dvv:
I wonder how do I skip tests for primary/foreign constraints (which depend on shard keys in memsql which is never set in tests)

elhigu:
Nope... that is not what I'm asking. One should configure tests to run tests also with memsql dialect. After that travis needs to be configured so that it has memsql server instance and run some integration tests against that.

Currently this dialect has no tests at all. I suppose if the dialect is 100% compatible with mysql extensive testing is not necessary. Minimum to get started would be to add some smoke tests that the dialect actually works. 

And last thing to add would be documentation about the new dialect.

elhigu:
There are some tests enabled/ignored for some dialects by just adding something like

```
if (config.client === 'postgres') {
  // test code here
}
```

elhigu:
btw. the implementation did look nice and clean to me 👍

dvv:
The dialect is special about constraints but it's declared to be 100% compatible in wire protocol.
Now I don't know how to test primaries/foreigns etc.
Eg
```
  604 passing (1s)
  1 failing

  1) Integration Tests mysql | memsql Schema createTable accepts the table name, and a "container" function:
     alter table `test_table_one` add unique `test_table_one_email_unique`(`email`) - The unique key named: 'test_table_
one_email_unique' must contain all columns specified in the primary key when no shard key is declared
```

elhigu:
Maybe you need to require giving shard key when creating table with memsql?  I suppose that would require memsql specific implementation for that schema function. I haven't been touching much schema side of code base so I don't know how those are overridden.

https://docs.memsql.com/docs/how-to-port-your-applications-to-memsql

dvv:
We've no `.shard()` and I don't really want to go that far. All I want is to be able to create/migrate memsql tables with knex )

calebeaires:
This will be the next big feature for us! Waiting for it!

elhigu:
Without complete tests etc. and CI support and dedicated maintainer I would say its better to use as custom dialect, by giving it to knex config like this:

```
const CustomDialect = require('./custom-dialect');

// and in knex config
{
  client: CustomDialect,
  connection: { ... }
}
```

I'm generally really hesitant nowadays of getting more dialects in, since it would require that there is some one who is responsible to figure out bugs related to them and to figure out how certain features are implemented for them. 

Already with current set of dialects it has been hard to figure out how to implement them in a way that all dialect are supported.

PR #1576, opened 2016-07-13
Allow named bindings to be escaped.

trevorsmith:
This PR extends escaping to named parameter bindings.

```javascript
knex.select('*').from('users').whereRaw(":property: = '\\:value' OR \\:property: = :value", {
  property: 'name',
  value: 'Bob'
})
```
Outputs:

```sql
select * from `users` where `name` = ':value' OR :property: = 'Bob'
```

I've included tests and updated the documentation as well.


elhigu:
Looks good to me 👍  Thanks

Issue #1577, opened 2016-07-13
Several bugs with MS SQL schema creation


statyan:
There is an issue with float type column is MSSQL columncompiler. It generates code `ADD COLUMN ColName float(8,2)`, but MSSQL accepts only `float(x)`, where x is size for bits holding mantissa.
MSSQL also have no double type column, float and double from knex should compile to `decimal(x,y)` column type.
Column type `bit`  is always exactly one bit, so call it with `table.column('a').bit(x)'`, where x is any number, leads to error.
Also, the `first` and `after` modifiers are not available for MSSQL.

So, I've fixed this issues and create current Issue just to connect it to the pull request.

PR #1578, opened 2016-07-13
Fixed knex #1577

statyan:
Also fix an issue with `dropPrimary` for MSSQL and fix test for primary key - now column test2 is set to notNullable. Without it primarykey test fails 'cause nullable column can't be set for primary key.


statyan:
Oh, don't matter about 3 commits "Fix for installing from github". It do nothing to code. Have no idea why they are listed for this PR, I thought I've added them later.

test/unit/schema/mssql.js
```diff
@@ -76,11 +76,11 @@ describe("MSSQL SchemaBuilder", function() {
 
   it('test drop primary', function() {
     tableSql = client.schemaBuilder().table('users', function() {
-      this.dropPrimary();
+      this.dropPrimary('testconstraintname');
     }).toSQL();
 
     equal(1, tableSql.length);
-    expect(tableSql[0].sql).to.equal('ALTER TABLE [users] DROP PRIMARY KEY');
+    expect(tableSql[0].sql).to.equal('ALTER TABLE [users] DROP CONSTRAINT [testconstraintname]');
```
rhys-vdw: Is `DROP PRIMARY` not supported by MSSQL?

rhys-vdw:
@statyan Looks good, I don't know much about MSSQL, but I'll cross reference these changes with the docs in future. Would you be able to squash these commits into one please?

test/unit/schema/mssql.js
statyan: Yep, DROP CONSTRAINT only. See this article, section "Using Transact-SQL"
https://msdn.microsoft.com/en-us/library/ms190621.aspx

statyan:
Ok,I'll try do it!

elhigu:
@rhys-vdw when you click "merge pull request" button there is option to do squash merge if one likes.

rhys-vdw:
@elhigu Of course, cheers.

Thanks @statyan!

PR #1585, opened 2016-07-20
MSSQL columnInfo() had hard coded schema name

nickwhiteley:
columnInfo() for MSSQL has the default schema (dbo) hard coded.  It should take the database and schema as parameters like the Postgres version does.

### DIFF
--- a/src/dialects/mssql/query/compiler.js
+++ b/src/dialects/mssql/query/compiler.js
@@ -158,11 +158,19 @@ assign(QueryCompiler_MSSQL.prototype, {
   // Compiles a `columnInfo` query.
   columnInfo() {
     const column = this.single.columnInfo;
-    const sql =
-      `select * from information_schema.columns where table_name = ? and table_schema = 'dbo'`;
+    let sql =`select * from information_schema.columns where table_name = ? and table_catalog = ?`;
+    let bindings = [this.single.table, this.client.database()];
+
+    if (this.single.schema) {
+      sql += ' and table_schema = ?';
+      bindings.push(this.single.schema);
+    } else {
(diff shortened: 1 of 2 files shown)


elhigu:
/home/travis/build/tgriesser/knex/src/dialects/mssql/query/compiler.js
  162:9  error  'bindings' is never modified, use 'const' instead  prefer-const
✖ 1 problem (1 error, 0 warnings)

elhigu:
Not knowing the dialect, changes looks fine I suppose :)

nickwhiteley:
The eslint error looks incorrect as the variable 'bindings' can be modified under the _if_ condition.  Does anyone know if this can be forced through or is it "better" to use const/var to get eslint to pass?

wubzz:
@nickwhiteley It's odd, I know. Short answer is change it to `const`.

Longer answer: An Object of any kind when defined as a constant simply means it cannot be re-assigned to anything else. It can still have properties assigned to itself, but cannot be re-assigned. For instance:
```javascript
const bindings = [1,2,3];
bindings.push(4); //Ok, bindings [1,2,3,4]
bindings = 'String'; //Error thrown.

const bindings = {col1: 'val1'};
bindings.col2 = 'val2'; //Ok, bindings = {col1: 'val1', col2: 'val2'}
bindings = 'String'; //Error thrown.
```

In this case the `bindings` is originally assigned as an Array, and is not re-assigned later on, this is why it should be constant instead. ESLint picks up on this.

nickwhiteley:
I'll change it.
It's counter intuitive (to me) so thanks for the explanation.  Always good to learn.

Issue #716, opened 2015-02-28
Common table expressions in knex.js


ericeslinger:
I couldn't find an issue for this, but I apologize if this is a dupe (for some reason I think I may have asked this already).

I'm using postgres, which has a handy-dandy common table expression syntax. I can do stuff like

    with (select * from documents) as alldocs
    select id from alldocs;

Naturally that's a toy example, but this is a pretty useful tool for doing some of the queries I need to run. Right now I'm just doing Raws, which is a bummer when I want to set up the CTE preamble, but then use json formatted objects to dynamically query the thing in question.

like putting a `.where(opts)` in the middle of the query, which I now have to do manually with string interpolation and counting ?marks.

blah238:
:+1: 

Not sure it helps but I did find an example of using CTEs in Knex (sort of): http://stackoverflow.com/a/27914498/386205

whitelynx:
:+1:

That Stack Overflow example could break if the queries have bound parameters. Instead, I ended up using bindings in `knex.raw()` instead of concatenating strings:
```javascript
var this_cte = knex('this_table').select('this_column');

var that_cte = knex('that_table').select('that_column');

knex.raw('with t1 as (?), t2 as (?) ?', [
    this_cte,
    that_cte,
    knex.select(['this', 'that'])
        .from(['t1', 't2'])
)
.then(function(response) {
    console.dir(response);
});
```

This worked pretty well, but there's some weirdness - especially the fact that it triggers knex's [Raw Queries](http://knexjs.org/#Raw-Queries) feature, which means that:
> [...] the response will be whatever the underlying sql library would typically return on a normal query, so you may need to look at the documentation for the base library the queries are executing against to determine how to handle the response.

It would definitely be preferable to actually have **native** support for CTEs in knex itself, so you don't have to deal with raw query responses in order to use them. I'd envision it working about like this (adapting the above example):
```javascript
knex
.with('t1', this_cte)
.with('t2', that_cte)
.select(['this', 'that'])
    .from(['t1', 't2'])
.then(function(rows) {
    console.dir(rows);
});
```

Or, a shortcut syntax for when you have multiple CTEs:
```javascript
knex.with({
    t1: this_cte,
    t2: that_cte
})
.select(['this', 'that'])
    .from(['t1', 't2'])
.then(function(rows) {
    console.dir(rows);
});
```

whitelynx:
[...]

PR #1599, opened 2016-07-28
add With Clause

atiertant:
fix https://github.com/tgriesser/knex/issues/716
add a with function(alias, callback|raw)
```
knex.with('with_alias', function() {
    this.select('*').from('books').where('author', 'Test')
}).select('*').from('with_alias')
```
output:
`with "with_alias" as (select * from "books" where "author" = 'Test') select * from "with_alias"`


test/unit/query/builder.js
```diff
@@ -3606,6 +3606,151 @@ describe("QueryBuilder", function() {
...
+    });
+  });
+
+  it("wrapped 'with' clause select", function() {
+    testsql(qb().with('withClause', function() {
+      this.select('foo').from('users');
+    }).select('*').from('withClause'), {
+      mssql: 'with [withClause] as (select [foo] from [users]) select * from [withClause]',
+      sqlite3: 'with "withClause" as (select "foo" from "users") select * from "withClause"',
+      postgres: 'with "withClause" as (select "foo" from "users") select * from "withClause"',
```
elhigu: Is it intentional that many of the tests doesn't have oracle / oracledb result checks?

src/query/builder.js
```diff
@@ -68,6 +68,44 @@ assign(Builder.prototype, {
...
+
+  with(alias, statement, bindings) {
+    if (typeof statement === 'function') {
+      return this.withWrapped(alias, statement);
+    }
+    // Allow a raw statement to be passed along to the query.
+    if (statement instanceof Raw && arguments.length >= 2) {
+      return this.withRaw(alias, statement, bindings);
+    }
+
```
elhigu: I suppose that this line should never be run, since only allowed parameters are function and knex.raw. Throwing exception and adding test for that would be good to prevent `undefined does not have property ...` errors.

elhigu:
Looks good. I added couple of small comments. Also would be good to add at least one integration test for this.

src/query/builder.js
atiertant: are you talking about the `return` in line 83 ?
i could replace `return` by `return this` like other functions : 
there is no other function who throw any exception in this file...
would you like me to `return` by:
`throw new Error("invalid arguments for with()");`

elhigu: Yep, could be even more detailed describing that argument must be either function / raw. There are no exceptions thrown in this file currently, but actually many of the function does throw an error if called with invalid parameters.

test/unit/query/builder.js
atiertant: in oracle the with clause is used in a select subquery.
this could be used in insert but could not find a way to use it with knex (raw or not...)

PR #1607, opened 2016-08-04
Updated README with info about multiple references

mccbala:
This issue occurs when multiple references are made to knex. It would be great if there is a way to avoid this or if this occurence is documented. I had to search for a lot of time and finally found this fix here http://stackoverflow.com/a/25854598/1459670

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,9 @@ NOTE: When you are referencing knex in more than one files follow this method in
 
 ```js
 // Initial reference (index.js / app.js)
-var knex = require('knex')(connection_object);
+var Knex = require('knex');
+// Pass connection parameters
+Knex.knex({...});
 
 // Subsequent references in other modules
-var knex = require('knex').knex;
+var Knex = require('knex').knex;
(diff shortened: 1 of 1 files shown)


elhigu:
Either you forgot to document the first part of the trick, where in your main app first you monkey patch `Knex.knex` to initialized instance. 

If it is really found from there already I'm pretty sure that thing you documented is just internal implementation detail, not a tested feature. Doing `require('knex').knex` sounds like anti-pattern which shouldn't be used. 

Instead you can create your own module which creates singleton instance of `knex` with your configuration if you like to achieve functionality above. 

I would just create it once in main application and then pass it to everywhere else that it is used.

mccbala:
@elhigu Yes. I forgot. that

elhigu:
@rhys-vdw whats your opinion on this? Do we want to promote this kind of solution? Probably I'll go to give my second opinion to that stackoverflow article.

mccbala:
@elhigu This may not be the best approach but I required this solution as I was creating data models in ORM style in which I wanted to use knex calls directly instead of writing a singleton instance of knex.

elhigu:
I reviewed this again and I'm sure I don't want to promote this in docs. Closing.

PR #1611, opened 2016-08-06
Added padding to avoid header sticking to the example usage.

karatheodory:
Dear Sirs,

Currently, especially in [some](http://knexjs.org/#Builder-main) [places](http://knexjs.org/#Schema) the header and code seem to be too close to each other. I have taken the liberty to add some padding between them.


tgriesser:
Thanks @karatheodory

PR #1612, opened 2016-08-08
Use config.returning value in PG Queries

nathanhoad:
The config option set up in https://github.com/tgriesser/knex/blob/master/src/dialects/postgres/index.js#L18 wasn't being used. This PR adds a check for the `defaultReturning` option set on the PG Client and uses it if it's there.


tgriesser:
Hmm, I forget when this had gotten added in here but I can't find it referenced in any other tickets and it doesn't look like it's documented as an option anywhere. 

I'm kind of hesitant to add it back in though, I'd rather get a good way to hook into the query construction process. Let me think on this one for a bit and get back to you.

elhigu:
Closing for now, since this is missing tests, documentation and resolving conflict. Can be opened for reconsideration if PR is finished.

Issue #913, opened 2015-07-24
group by problems in union query


unsign3d:
Hi all,
I need to perform 2 group by in the inner query of an union query

this is my code 

```
return bookshelf
        .knex('lists')
        // first half of the query, we just take
        // the list that the user is owner, we need to
        // have the field correctly
        .select('lists.id', 'lists.name', 'lists.public',
          'lists.editable', 'lists.owner_id', 'lists.created_at')
        .count('list_items.id as item_counter')
        .from('lists')
        .leftJoin('list_items', 'list.id', 'list_items.id')
        .where('lists.owner_id', user_id)
        .groupBy('list_members.list_id')
        // we just need to group by list_id
        //.groupBy('list_members.list_id')
        // second half of the query, we need to retrive
        // the list that the user is member, this works because
        // an admin is not a member
        .union(function(){
          this.select('lists.id', 'lists.name', 'lists.public',
            'lists.editable', 'lists.owner_id', 'lists.created_at')
          //.count('list_members.user_id as member_counter')
          .count('list_items.id as item_counter')
          .from('lists')
          .innerJoin('list_members', 'lists.id', 'list_members.list_id')
          .leftJoin('list_items', 'lists_id', 'lists_items.list_id')
          //.where('lists_members.user_id', user_id)
          .groupBy('list_members.list_id');
        });
```

the result query is

```
 select `lists`.`id`, `lists`.`name`, `lists`.`public`, `lists`.`editable`, `lists`.`owner_id`, `lists`.`created_at`, count(`list_items`.`id`) as `item_counter` from `lists` left join `list_items` on `list`.`id` = `list_items`.`id` where `lists`.`owner_id` = 1 union select `lists`.`id`, `lists`.`name`, `lists`.`public`, `lists`.`editable`, `lists`.`owner_id`, `lists`.`created_at`, count(`list_items`.`id`) as `item_counter` from `lists` inner join `list_members` on `lists`.`id` = `list_members`.`list_id` left join `list_items` on `lists_id` = `lists_items`.`list_id` group by `list_members`.`list_id` group by `list_members`.`list_id` - ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'group by `list_members`.`list_id`' at line 1
  Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'group by `list_members`.`list_id`' at line 1

```

[...]

PR #1632, opened 2016-08-17
Reorder components

spruce-bruce:
This is a version of the fix people have been using to address #913. It changes the priority of the union component be be behind group, having, limit and offset.

I had to keep order _after_ union still, so now order is after limit and offset instead of before. I don't fully understand the repercussions of this, but it works fine in my limited testing.


spruce-bruce:
Well that test failing answers that question - this won't work

PR #1644, opened 2016-08-26
Add oracledb stream support

natevecc:
I took a stab at adding support for the Client_Oracledb class to use oracledb#queryStream method to stream results through knex. I added a test but not sure if more is needed so please let me know if I need to make more tests or change how I setup the streaming feature. Finally I wasn't sure how to get the tests working locally with oracledb and an oracle db so I'm hoping the tests will work correctly in travis but if not I'd appreciate some help in getting the tests working locally and I'll update the docs as well.

### DIFF
--- a/src/dialects/oracledb/index.js
+++ b/src/dialects/oracledb/index.js
@@ -345,4 +345,14 @@ Client_Oracledb.prototype.processResponse = function(obj, runner) {
   }
 };
 
+Client_Oracledb.prototype._stream = function(connection, obj, stream, options) {
+  obj.sql = this.positionBindings(obj.sql);
+  return new Promise(function (resolver, rejecter) {
+    stream.on('error', rejecter);
+    stream.on('end', resolver);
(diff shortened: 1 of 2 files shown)


elhigu:
Sorry for the huge delay, I hadn't noticed this one earlier, looks like a pull request that should be merged in after conflicts are fixed. Setting up oracledb locally is a bit painful. If I recall correctly I did it with vagrant + VirtualBox earlier... Would be good to have some vagrant file for setting it up locally, probably I'll do that if I need to set it up again at some point.

vegarringdal:
So knex.js with oracledb streams is not supported atm ?

natevecc:
Sorry I've moved on from the project where I was using this and don't know when I'll have time to get back to it. I'll try fix the conflicts soon but if someone else could handle setting up the tests that'd be much appreciated.

elhigu:
Closing incomplete PR, can be reopened if completed.

PR #1649, opened 2016-09-04
🔧 Ignore stub files in lint task

dannyfritz:
This is a really small PR.

Basically, when running `npm run lint`, the `.stub` files should not be linted because they are not valid JS.


elhigu:
Thanks!

elhigu:
And sorry for the pretty epic delay :)

dannyfritz:
No problem! Thanks for getting to it. Love the library. :+1:

PR #1650, opened 2016-09-04
🚨 Add test for bad client supplied in Knex config

dannyfritz:
When the user provides a bad client in the `knexfile.js`, provide a more useful error than the one thrown by `require`.

I left in some commented out code for testing against `console.log`, but it was ugly so I took it out.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,6 @@
 
 import Raw from './raw';
-import { warn } from './helpers';
+import { error, warn } from './helpers';
 import Client from './client';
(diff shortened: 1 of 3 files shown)


elhigu:
Is this still valid? something similar was implemented in #1822

dannyfritz:
yes, i believe a similar feature was implemented already. feel free to
close :+1:

elhigu:
Sorry about missing this one when the other one was reviewed.

PR #1657, opened 2016-09-10
Add example to Foreign Key

JonathanCMitchell:
Adds an example to Foreign Key method. This method of adding a column and foreign key currently works while the others seem to be broken.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2139,7 +2139,6 @@ knex.table('users')
       <tt>SET NULL</tt>, <tt>NO ACTION</tt>) for the operation.  Note, this is the same as <code>column.references(column)</code>
       but works for existing columns.
     <pre class="display">
-    knex.table('users');
     table.integer('user_id').unsigned();
     table.foreign('user_id').references('Items.user_id_in_items');
     </pre>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2138,8 +2138,14 @@ <h3 id="Schema-Building">Schema Building:</h3>
       You can also chain <tt>onDelete</tt> and/or <tt>onUpdate</tt> to set the reference option (<tt>RESTRICT</tt>, <tt>CASCADE</tt>,
       <tt>SET NULL</tt>, <tt>NO ACTION</tt>) for the operation.  Note, this is the same as <code>column.references(column)</code>
       but works for existing columns.
+    <pre class="display">
+    knex.table('users');
```
elhigu: `knex.table('users')` is a bit misleading, since it should actually be

```
knex.schema.table('users', function (table) {
  ...
})
```

Looks like the other examples are just directly using `table.someBuilderFunction(...)` so would be enough / less misleading to just remove the line `knex.table('users');` from the example.

elhigu:
All documentation updates now seem to go to a different branch :( I can add it to new docs.

elhigu:
btw. we've been creating foreign keys like this: `table.biginteger('siteId').unsigned().notNullable().references('id').inTable('Site').onDelete('CASCADE').index();` which ones are broken?

PR #1659, opened 2016-09-12
1.0 Refactor

tgriesser:
I've commented on several prior tickets about the 1.0 API refactor, and have been working locally on a few iterations of modifying things to get closer to what an ideal 1.0 release should include.

With the number of weird edge cases which currently exist in the internal API, I felt it'd be smoother not to push a big bang rewrite of everything but instead incrementally refactor by:
1. laying out the plan of what should be added
2. prioritizing essential changes 
3. backing in toward other nice-to-have's and better overall structure

This way, anyone interested can comment / give input about the overall ideas direction and contribute fixes, etc. This PR will include any commits related to this restructuring.

Check out the [Refactor](https://github.com/tgriesser/knex/blob/refactor/REFACTOR.md) document for a more in-depth / up-to-date overview about what's changing and why:
##### Knex Structure
- [ ] Dialect: Normalizes sql across languages
- [ ] Adapter: Normalizes db interaction across
- [ ] Client: Adapter + Dialect
- [x] Context: Client + Single Connection + Logging 
- [x] Transaction: Specialized Context
##### API Additions / Changes
- [ ] knex.sql
- [ ] Documented API for custom value interpolation
- [ ] Hooks
- [ ] Standardize / simplify escaping per-dialect
- [ ] Custom logging
- [ ] Customizable pools
- [ ] Schema Column modification API
- [x] Improved A+ behavior on .then with chains
##### Reach goals
- [ ] seed / migration / cli improvement
- [ ] testing refactor
- [x] `asPartial`
- [ ] prepared statements
- [ ] modularization


REFACTOR.md
```diff
@@ -0,0 +1,199 @@
...
+
+Tests are a bit of a mess right now, difficult to follow, test across different drivers etc. This is a big task, which is why it's added as a reach goal - but it's something that should happen 
+
+##### Properly constructed prepared statements
+
+The difficulty with properly constructed prepared statements is the requirement that prepared statements execute on the same connection. The "context" addition should help in getting closer to this, but that still does not help with knowing which connection to checkout from the pool. Something to keep in mind.
+
+##### Modularizing
+
+For knex, this isn't necessarily a goal - most of the pieces of knex aren't really re-usable outside of knex, but it'd be good to design things more modularly for testing purposes, etc.
```
knownasilya: The query builder being it's own module would help in cases like https://www.npmjs.com/package/cartodb-tools

tgriesser: Right, yes. I should clarify that it is a goal, just not the immediate priority. With the number of changes that need to be made, prematurely splitting things out might make things more difficult. But yes I would absolutely like to keep in mind making it simple to eventually split the query builder out.

wubzz:
@tgriesser With the addition of hooks, which I'm guessing are a set of hooks decided by the library, would you oppose the idea of exposing said available Hooks and existing Events as constants through something like this?

```javascript
Knex.HOOKS.beforeStart //'before-start'
Knex.EVENTS.queryError //'query-error'

builder.on(Knex.EVENTS.queryError, () => {...});
builder.select().hook(Knex.HOOKS.beforeStart, () => {...});
```

I was meaning to make a PR for this already relating to the events. Generally working with strings like this instead of constants is just asking to make a mistake somewhere along the line, and there's really no reason that I can see why the library would have the **user** make their own map of this. Preferably libraries should expose available events for programatic use.

tgriesser:
@wubzz yeah that could work. I was also thinking of validating against a list of valid hooks & throwing if the user has provided an incorrect value.

MarkHerhold:
Could splitting the CLI away from the core `knex` dependency be a consideration as well?

tgriesser:
@MarkHerhold Yeah I would like to do that. I haven't thought too much about the cleanest way to approach it but it is something I'd like to see happen.

cjnqt:
Any plans on releasing trx contexts / single connection before the big 1.0 release?

odigity:
If I understand these changes correctly, then I'm also very interested in seeing these merged.  I just attempted to learn how to use transactions with knex.js, and I'm finding the use of .transacting(trx) to be unintuitive and problematic in that it requires modifying the actual code using the DB, which I want to avoid - I want to use transactions in my test framework to rollback after each test by passing in a trx when requiring the module so it uses my transaction-enabled connection without modification.

elhigu:
@odigity I wouldn't suggest of doing that because your tests will end up using just one connection instead of connection pool and it will not represent real world usage of your app.

Also its pretty much impossible to implement unless the application you are testing is running its queries in the same transaction that was started inside test code (you would need to start transaction in test then start your app and pass the created transaction to app so that it makes all the queries to the same transaction and nested transactions might behave starngely...).

Anyways I don't know your exact use-case so I might be wrong. Anyways you don't have to use `.transacting(trx)` currently in knex. You can  create transaction and use trx like main knex instance and do the queries like this `trx('TableName').where('col', 'foo')...`

odigity:
Thanks so much for the response.  I'm kind of in the dark on this one - the knex docs cover the barebones of transactions, but when I google "using knex.js for unit testing" and "ava.js + knex" and "ava.js + database", I couldn't find anything particularly instructive to show me a good way, so I fell back on my Ruby experience where wrapping a unit test in a transaction is a common technique for reseting the DB after each test.

> @odigity I wouldn't suggest of doing that because your tests will end up using just one connection instead of connection pool and it will not represent real world usage of your app.

That did occur to me, but I was willing to accept that if it allowed me to use a simple, write-once method for implementing post-test DB cleanup.  (I did set my pool min/max to 1 to be sure.)  If there's a way to accomplish this while supporting concurrent connections, I will absolute embrace that.
 
> Also its pretty much impossible to implement unless the application you are testing is running its queries in the same transaction that was started inside test code (you would need to start transaction in test then start your app and pass the created transaction to app so that it makes all the queries to the same transaction and nested transactions might behave starngely...).

The way I hoped/expected it to work is:

1) In each unit test, I create a transaction which produces `trx`.

2) I then require the module I want to test and pass the `trx` object to the module constructor so it will be used by the module, thus resulting in all queries happening inside the transaction.

3) After the module method returns (or throws error), I run my assertions on the resulting state of the DB, then call `trx.rollback()` to undo everything from the start to prepare for the next test.

So, that's what I'm trying to achieve and how I originally intended to achieve it.  I'm eager to learn more about:

1) In what whys I'm misunderstanding how Knex.js works and should be used.

2) Best practices for writing atomic unit tests for code that touches the database.

elhigu:
@odigity could you please create new issue for this discussion so that we won't pollute this thread. I will notice new issues when they are created.

tgriesser:
This PR has gotten pretty stale and now that I've had the time to step away from Knex for quite a bit of time and have recently started getting back into it, I have some new perspective/ideas on direction we might want to go here. Will address this in a separate ticket sometime soon(ish), gonna close this but keep the branch around for reference.

wubzz:
@tgriesser @elhigu I've had an eye on this lately because eventhough this specific goal is currently on ice, I believe some of these changes will give quite the lift for this library.

However, I also feel that doing all of this in one swoop is a bit much, and would like to suggest implementing some of these changes in parts instead of targeting one major release. Specifically:

*Transactions / KnexQueryContext*
Totally a needed change to knex, and this PR lays out a really good base.

*knex.sql*
Not only is this a nice feature addition, but the `.escapeId` / `.escapeColumn` functions are also some of the most requested features of knex.

*Better logging*
Speaks for itself. Knex lacks in this department, even using `console.log` at times.

*Tests refactor*
A given.

So my question is if we are in agreement about this and can start implementing it?  @tgriesser

wubzz:
In the spirit of testing and investigating how these changes affect current state of knex I created a branch [_refactor](https://github.com/wubzz/knex/tree/_refactor) with the changes from this PR. Had to be done manually, trying to solve conflicts felt more risky.. Only tested that all `postgres ` tests pass.

What I really like about this is the transactions and connections overhaul. `KnexContext` concept also is a very big lift.

```javascript
app.use((req, res, next) => {
    req.knex = knex.transaction();
    next();
});

//Somewhere else, runs in transaction
req.knex('accounts')
.select();

//ResponseHandler
app.use((req, res) => {
   await req.knex.commit();
});

//ErrorHandler
app.use((error, req, res, next) => {
    await req.knex.rollback(error);
});
```

elhigu:
@wubzz I agree that all of those features would benefit knex without need to rewrite everything. I don't think it is necessary to wait for @tgriesser answer (project development cannot relay on single person) as long as maintainer community agrees on changes (currently pretty much us + some other contributors).

tgriesser:
Yeah, I still want to get back to this. Fortunately, I'm going to be using knex again quite a bit and so I'll have an active interest in getting the library back on track.

I've been investigating the possibility of keeping a pool of "Adapter" objects (with underlying connection) rather than just having a singleton adapter keeping track of/assigning/mutating raw connection objects, and what designs that change potentially allows. I'll comment here or maybe open another ticket with more designs I've been playing around with.

I agree that we really need to simplify the internals though. In getting back into things, I've been wondering what I was thinking when I wrote some of this so long ago 😬. I think maybe simplifying the tests would be a good starting place though.

wubzz:
@tgriesser Great to see you back!

With new design ideas in mind I suppose you'll want to forget the changes of this PR all together and start fresh? 

If so then I should stop development of the [branch](https://github.com/wubzz/knex/tree/_refactor) I made containing this PR's changes. All in all I still think this alone is a great step-up for the internals.

elhigu:
Great to have you back @tgriesser  indeed! I would love to see all future changes as small incremental steps towards to the goal described in OP so that there would be constant advancement. 

About migration improvements I would be really hesitant of deprecating old API anytime soon and I don't see how it could be fixed without breaking all old knex apps (those migrations are already ran to the production databases and then when people run them again on test databases etc., functionality wouldn't be exactly the same anymore and that scares me a lot).

I have been planning about adding another knex-migrations library which could be used side by side with old migrations and using old API would just give warning in the future. Anyways before starting to write it I will create separate issue about design goals (mostly about sane defaults and what kind of features should be included in API to keep it consistent between different dialects) and proposal of the new API so that more people can join in to design it.

We have probably hundreds of migration files in our code repositories so I can also check out from there some most common use cases.

mercmobily:
I broke up the *goal* of this PR into several steps, documented here #6257 I will likely rescue some of the code (or at least the code ideas) from this PR. 

It's a shame this PR didn't land. It would have solved an obscene number of problems.

Issue #1602, opened 2016-07-30
Escaping of bindings in v0.11 with PostgreSQL (or how to get v0.10 behavior?)


moll:
Hey,

I'm upgrading from v0.10 to v0.11 and noticed non-plain object bindings given to `knex.raw` now get escaped by Knex before being passed to the Postgres driver (`node-postgres`) and its `require("pg/lib/utils").prepareValue`. I was depending on them being passed as-is to serialize various types in different ways, thereby expanding what the Pg driver supports independent of any layers above it.

I presume it's something to do with https://github.com/tgriesser/knex/blob/dfca38da339922f8ad8f52087ca5e6c0dcdab081/src/dialects/postgres/utils.js. Off the top of your head, would there be any risk if I [locally] scrapped that entirely and had it pass everything as-is to the Pg module's `prepareValue`? It's used only for bindings and not used for interpolating inside a SQL string, right?

Thanks!

elhigu:
Could you add an example query of your use case?

moll:
Sure:
```
function MyClass() {}
knex.raw("SELECT ?", [new MyClass])
```
That used to be sent directly to Postresql library's `prepareValue`. Now it's stringified beforehand.

elhigu:
@moll that doesn't seem like complete example? how pg driver should have interpreted passing empty instance? Did you use overriding val.toPostgres or something like that? 

Anyways I don't know any way to skip knex's prepare value... I don't have strong opinion if there should be way to do it... I don't know the motivation why knex has separate prepare values implementation. Maybe to have consistent functionality with knexQuery.toString() and with sending query through pg driver.

moll:
In the example above interpreting value or identity objects is Pg's value preparation's responsibility. You could think of it as using `toPostgres`, even though I've got a functional, as opposed to object-oriented, equivalent on Pg's side. I was using Knex merely for its query building capabilities, expecting it to escape only values that can't be handled by placeholders, and those are usually only table or column names.

As far as I can tell now, and correct me if I'm wrong, any custom object serialization, be it of value or identity objects, is not possible in Knex v0.11 because even if Knex is made extensible, Pg _will_ still do its own serialization, thereby escaping serialized strings twice (`"foo"` -> `"\"foo\""`). The proper solution that I can tell is to have Knex pass values untouched to Pg, which will escape or serialize values as necessary. As before. Hence this issue.

Let me know if I need to further elaborate.

elhigu:
[...]

Issue #1548, opened 2016-07-02
Proper string-escapes for Postgres


indeyets:
Currently, knex tries to use backslashes to escape various characters in string literals for Postgres. The problem is, that Postgres doesn't interpret slashes in string literals by-default since version 9.1. It is possible to enable interpretation by setting `standard_conforming_strings` option to `off`, but this is not a recommended way.

Strictly speaking, usage of regular string-literals is just not a predictable option anymore.

**The proper solution is:** use non-standard `E'literal'`. They are defined with a strict [set of rules](https://www.postgresql.org/docs/9.5/static/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE) and should work reliably for any modern Postgres version.

Input: `Hello \"World\"`
Probably wrong output: `'Hello \\"World\\"'` (current knex)
Probably right output: `'Hello \"World\"'`
Definitely right output: `E'Hello \\"World\\"'`

related to #1546, #886, #869, #828, #658

elhigu:
Thanks @indeyets for opening this discussion. I'm not 100% sure if I got the examples right. I mean that how javascript is interpreting string literals has also something to add to this discussion. 

So I'll just open up your example a bit to understand it better... Input `Hello \"World\"` must be written in javascript like `'Hello \\"World\\"'` or otherwise already javascript string literal parsing removes the single  backslash. 
 
Anyways knex is actually not doing the string escaping in most of cases since just positional bindings are passed to node-postgres driver as javascript string which is responsible of escaping them correctly:

```js
> knex('dodo').where('name', 'hell\\"o there').toSQL()
{ method: 'select',
  options: {},
  bindings: [ 'hell\\"o there' ], <-- javascript string containing `hell\o there` is passed to driver (this is good) 
  sql: 'select * from "dodo" where "name" = ?' }
```

But it is true that knex also has its own escaping function, which should work correctly. Which currently indeed replaces \ with \\

```js
> knex('dodo').where('name', 'hell\\"o there').toString()
'select * from "dodo" where "name" = \'hell\\\\"o there\''
```

Which really seem to result invalid string in database (too many backslashes...):

```js
mikaelle=# Hello \\"World\\"';
     ?column?      
-------------------
 Hello \\"World\\"
(1 row)
```

So... we seem to need own string escaping function for postgres or change some other part of string literal interpolation to use E'' literals. 

[...]

PR #1661, opened 2016-09-12
Fix PG string escaping behavior

tgriesser:
Reverts to pre 0.11 behavior for custom parameters, #1602 and adds more consistent escaping for postgresql queries: #1548

Also re-arranges where some of the query parsing/parameterization is happening internally in raw queries for consistency.

This will land in 0.12 release


src/query/string.js
```diff
@@ -1,134 +1,161 @@
...
+  const finalEscapeDate = config.escapeDate || dateToString
+  const finalEscapeArray = config.escapeArray || arrayToList
+  const finalEscapeBuffer = config.escapeBuffer || bufferToString
+  const finalEscapeString = config.escapeString || escapeString
+  const finalEscapeObject = config.escapeObject || escapeObject
+  const finalWrap = config.wrap || wrapEscape
+
+  function escapeFn(val, finalEscape, ctx) {
+    if (val === undefined || val === null) {
+      return 'NULL';
```
wubzz: @tgriesser A lot of changes so hard to follow up, but just to verify, is this part run *before* or *after* the checks run during compile such as https://github.com/tgriesser/knex/pull/1661/files#diff-2fe557044fa519ea01fdf735c4846284R99 ?

If run before, then these checks would never work, so just double-checking. In any case,  :+1: :+1: for this pr.

tgriesser: They run after, basically if `.toString` is used, this is the escaping that happens

PR #1665, opened 2016-09-13
Revert to generic pool

tgriesser:
For 1.0, I'm planning on making the pool optional or easily configurable by allowing folks to easily inject the strategy for managing connections in the configuration. 

In the meantime there's quite a few issues that keep cropping up with Pool2, and as nice as the features like capabilities and backoffs are in that library, I think for now I'm going to keep things a bit simpler with node-pool. 

In looking at node-pool it seems to have developed to be a bit nicer since last time we had it in here as a pooling lib a few years back. So for 0.12 I'm proposing we go with generic-pool and then in 1.0 we'll use the pools that come with the individual adapter libraries, with a documented way to replace with something like generic-pool or pool2.


ErisDS:
> In the meantime there's quite a few issues that keep cropping up with Pool2

Hey @tgriesser, could you possibly share some details on what the known issues with Pool2 are? Trying to assess whether upgrading to knex 0.12 is critical or not.

tgriesser:
Basically this issue https://github.com/TryGhost/Ghost/issues/7346 - if something messes up with the db connection, knex was intercepting the error event emitted by pool2 and just logging it instead of recovering from it properly (or emitting it so it would take down/restart the app). So prior to 0.12 the server can easily get into a bad state.

PR #1676, opened 2016-09-19
Add Cordova-sqlcipher-adapter Support

lvffilho:
Added Cordova-sqlcipher-adapter dialect


3h3c:
Hey,

i also use knex + bookshelf with cordova-sqlite. But now i have to upgrade to a new version of kenx and its now not working with cordova-sqlite. Is your build working?

elhigu:
Without complete tests etc. and CI support and dedicated maintainer I would say its better to use as custom dialect, by giving it to knex config like this:

```
const CordobaSqlite = require('./custom-dialect');
{
  client: CordobaSqlite,
  connection: { ... }
}
```

Issue #1647, opened 2016-09-01
Template tag interface


knownasilya:
Would love to see the following API

```js
let result = await db.raw`select ${column} from ${table} where id = ${id}`;
```

Kind of what https://github.com/devTristan/simple-postgres#template-string-mode does.

tgriesser:
Yep - this is something I'd like to add as well, added to the 1.0 api list at: https://github.com/tgriesser/knex/pull/1659

avimar:
Related: https://github.com/tgriesser/knex/pull/1766

PR #1682, opened 2016-09-22
add analytic functions

atiertant:
add analytic functions (only works on mssql, oracle and postgresql):
- denseRank()
- rank()
- rowNumber()

doc [here](https://github.com/knex/documentation/pull/5)

### DIFF
--- a/src/dialects/mssql/query/compiler.js
+++ b/src/dialects/mssql/query/compiler.js
@@ -112,6 +112,9 @@ assign(QueryCompiler_MSSQL.prototype, {
         if (stmt.type === 'aggregate') {
           sql.push(this.aggregate(stmt))
         }
+        else if (stmt.type === 'analytic') {
+          sql.push(this.analytic(stmt));
+        }
         else if (stmt.value && stmt.value.length > 0) {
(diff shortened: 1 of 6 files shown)


tgriesser:
Thanks for the PR... going to take a look at this one for a bit. I'd like to see if there'd be a simpler way to augment the query building process without adding a ton of extra dialect specific methods to the core builder chaining api.

Perhaps something like #1647 is the best answer for the 5% of queries that don't fit well with the builder, I'd like to get that into a release fairly soon.

atiertant:
@tgriesser hi, have you got some news on it? we needs this feature on our developement, is there something we could do for this PR?

kibertoad:
@tgriesser @elhigu So what was the verdict on this one?

elhigu:
I have no comments or preference if I would like knex to have these or not.

kibertoad:
@atiertant I realize that this was ages ago and probably this feature is no longer relevant for you, but in case you would be willing to rebase it against current master, it can go in.

kibertoad:
Merged as #4188

PR #1685, opened 2016-09-23
Configuration for running integration tests with oracledb client

elhigu:
Testing running integration tests.


elhigu:
I think I finally found the problem... now I'll try out if I can make this work on top of  oracle test fix branch.

PR #1687, opened 2016-09-23
Add browser friendly build with webpack.

ClickSimply:
Noticed that the browser build isn't in place anymore, causing links in the documentation to 404 error to this page:
https://github.com/tgriesser/knex/tree/master/build

This pull request should restore the usability of the link and give folks a drop in browser version of the lib.


tgriesser:
Actually, I removed the browser build on purpose, since it clutters up the repo when it's rebuilt on each commit and anyone who is going to be using the browser build of knex will be better off configuring things as they wish using a build tool like webpack or browserify. 

I've updated the documentation to reflect this change.

PR #1697, opened 2016-09-26
Adding link to ease editing the docs

servel333:
According to https://github.com/jekyll/jekyll-help/issues/5, this type of link will work, however, I don't know how to test to make sure this will produce a meaningful link. Suggestions welcome.


tgriesser:
Thanks for the idea, but the docs have actually been moved over to https://github.com/knex/documentation and since they actually don't use jekyll, I don't believe this shortcut will link as intended.

servel333:
Thanks for the note. I'll make a similar request over at the other repo. I think this illustrates why this is needed though.

PR #1702, opened 2016-09-28
Use Tarn.js as a pool instead of generic-pool. fixes #1694 #1703

### LINKED ISSUE #1694: Connections are not returned to the pool if acquireConnectionTimeout is triggered
This test fails on 0.12.1

```js
const _ = require('lodash');
const Knex = require('knex');
const Promise = require('bluebird');

const knex = Knex({
  client: 'postgres',
  pool: {
    min: 0,
    max: 2
  },
  connection: {
    host: '127.0.0.1',
    database: 'test'
  },
  acquireConnectionTimeout: 100
});

// Start two transactions so that the pool is filled.
_.times(2, (idx) => {
  knex.transaction(() => {
    return Promise.delay(500);
  }).then(() => {
    console.log(`transaction ${idx} ended`);
  });
});

knex.raw('select 1').catch(() => {
  console.log('query failed as expected because of acquireConnectionTimeout');
});

setTimeout(() => {
  // By the time this is called, all the transactions should have been committed and
  // the connections returned to the pool, but the connection for which the acquireConnectionTimeout
  // was called, was never returned to the pool.
  console.log('connections in use:', knex.client.pool._inUseObjects.length, 'should be 0 right?');
}, 1500);
```

koskimas:
This pull request replaces the generic-pool with [Tarn.js](https://github.com/Vincit/tarn.js).

generic-pool doesn't have a built-in timeout for connection acquire nor does
it have a mechanism to cancel a pending acquire. Because of this the issue
#1694 is next to impossible to fix on the knex side.

generic-pool doesn't seem to be well maintained anymore. The last commit is
six monhts old at time of writing this and there are multiple pull requests
waiting, one of which is a fix for the acquire timeout needed here.

Tarn.js is a subset of generic-pool with added timeouts and a comprehensive
test suite. It is designed (by me) to be used with knex and I intend to make
that its only purpose. If you choose to merge this pull request. I'm committed 
to maintaining and bug fixing it to serve knex as well as possible. 

This pull request also fixes an issue #1703 where connections acquired for
transactions never time out, regardless of acquireConnectionTimeout

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -16,7 +16,7 @@ import TableCompiler from './schema/tablecompiler';
 import ColumnBuilder from './schema/columnbuilder';
 import ColumnCompiler from './schema/columncompiler';
 
-import { Pool } from 'generic-pool';
+import { Pool } from 'tarn';
 import inherits from 'inherits';
 import { EventEmitter } from 'events';
 
@@ -172,15 +172,14 @@ assign(Client.prototype, {
           debugPool(level.toUpperCase() + ' pool ' + name + ' - ' + str)
         }
       },
-      create: (callback) => {
-        this.acquireRawConnection()
+      create: () => {
+        return this.acquireRawConnection()
           .tap(function(connection) {
             connection.__knexUid = uniqueId('__knexUid')
             if (poolConfig.afterCreate) {
               return Promise.promisify(poolConfig.afterCreate)(connection)
             }
           })
-          .asCallback(callback)
       },
       destroy: (connection) => {
         if (poolConfig.beforeDestroy) {
@@ -209,7 +208,9 @@ assign(Client.prototype, {
       helpers.warn('The pool has already been initialized')
       return
     }
-    this.pool = new Pool(assign(this.poolDefaults(config.pool || {}), config.pool))
+    this.pool = new Pool(assign(this.poolDefaults(config.pool || {}),
+      {acquireTimeoutMillis: config.acquireConnectionTimeout || 60000},
+      config.pool))
   },
 
   validateConnection(connection) {
@@ -218,25 +219,17 @@ assign(Client.prototype, {
 
   // Acquire a connection from the pool.
   acquireConnection() {
-    let request = null
-    const completed = new Promise((resolver, rejecter) => {
-      if (!this.pool) {
-        return rejecter(new Error('Unable to acquire a connection'))
+    if (!this.pool) {
+      return {
+        completed: Promise.reject(new Error('Unable to acquire a connection'))
       }
-      request = this.pool.acquire(function(err, connection) {
-        if (err) return rejecter(err)
-        debug('acquired connection from pool: %s', connection.__knexUid)
-        resolver(connection)
-      })
-    })
-    const abort = function(reason) {
-      if (request && !request.fulfilled) {
-        request.abort(reason)
+    } else {
+      return {
+        completed: this.pool.acquire().promise.then((connection) => {
+          debug('acquired connection from pool: %s', connection.__knexUid)
+          return connection
+        })
       }
-    }
-    return {
-      completed: completed,
-      abort: abort
     }
   },
 
@@ -252,17 +245,8 @@ assign(Client.prototype, {
 
   // Destroy the current connection pool for the client.
   destroy(callback) {
-    const promise = new Promise((resolver) => {
-      if (!this.pool) {
-        return resolver()
-      }
-      this.pool.drain(() => {
-        this.pool.destroyAllNow(() => {
-          this.pool = undefined
-          resolver()
-        })
-      })
-    })
+    const promise = this.pool ? this.pool.destroy() : Promise.resolve()
+    this.pool = undefined
 
     // Allow either a callback or promise interface for destruction.
     if (typeof callback === 'function') {
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -4,9 +4,9 @@
 import Promise from 'bluebird';
 import { EventEmitter } from 'events';
 import Debug from 'debug'
+import { TimeoutError } from 'tarn';
 
 import makeKnex from './util/make-knex';
-import noop from './util/noop';
 
 const debug = Debug('knex:tx');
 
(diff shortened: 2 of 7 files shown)


src/transaction.js
```diff
@@ -142,6 +142,15 @@ export default class Transaction extends EventEmitter {
   acquireConnection(client, config, txid) {
     const configConnection = config && config.connection
     return Promise.try(() => configConnection || client.acquireConnection().completed)
+    .catch(TimeoutError, (error) => {
```
elhigu: This is nice addition, earlier I was able to fill my pool and stuck waiting new transaction forever...

src/client.js
```diff
@@ -218,26 +219,18 @@ assign(Client.prototype, {
...
-    })
-    const abort = function(reason) {
-      if (request && !request.fulfilled) {
-        request.abort(reason)
+    } else {
+      return {
+        completed: this.pool.acquire().promise.then((connection) => {
+          debug('acquired connection from pool: %s', connection.__knexUid)
+          return connection
+        })
```
elhigu: Why abort is not needed anymore?

src/client.js
```diff
@@ -252,17 +245,8 @@ assign(Client.prototype, {
...
-        return resolver()
-      }
-      this.pool.drain(() => {
-        this.pool.destroyAllNow(() => {
-          this.pool = undefined
-          resolver()
-        })
-      })
-    })
+    const promise = this.pool ? this.pool.destroy() : Promise.resolve()
```
elhigu: This seems to revert the change made 2 weeks ago, that destroying pool waits that all the queries are resolved before ending.

src/client.js
koskimas: Tarn handles the acquire timeout. There is no need to handle it on the knex side anymore. There are tests for this.

src/client.js
koskimas: Yeah, it was unclear to me what drain did. Currently Tarn murders all connections (both used and free) and calls the destroy function for them. 

I'll change destroy so that it waits for all connections to be released before destroying them.

src/client.js
koskimas: This is fixed and tested in Tarn 0.1.5

src/client.js
koskimas: This is now fixed in Tarn 0.1.5. I updated the package.json in the pull request.

elhigu:
@tgriesser @rhys-vdw any any comments on this one? Pull request actually fixes some bugs + our own testsuites did work correctly after this (objection.js + one internal project backend with > 20k lines of typescript and which has 15k lines of tests).

elhigu:
should fix this one too I suppose https://github.com/tgriesser/knex/issues/1709

tgriesser:
Thanks for the PR... I'll try and get to reviewing this tomorrow, I'd personally rather make it so folks can specify their own pooling implementation rather than continue to switch it up. I'd also like to dig in a bit more about #1964 and see what all is happening there, I believe that abort stuff is legacy from pool2, so that probably needs to get fixed.

Not saying this won't get merged, just saying I'd like to investigate a bit further.

koskimas:
I agree with you @tgriesser that switching back and forward with the pool isn't ideal and causes bugs. That is exactly what this pull request is all about. Finally having a pool that is written just for knex. I'd be happy to give you push rights to tarn.js.

Also allowing people to use their own pool is a good feature, but the default pool should serve most use cases.

The problem with #1694 was the legacy abort stuff, yes, but there is no mechanism in generic-pool to implement the same features the abort was solving. And pull requests to add the feature to generic-pool have been pending for six months. I would have just contributed to generic-pool if I thought that would solve this within next few months, but that doesn't seem to be the case.

I do realize that switching to a new library that has been around for a week is not the safest sounding idea 😄. This branch needs to be tested really well before making a release, if you decide to merge it. But I also think that resource pooling is not that difficult a problem. If the pool has good, meaningful tests it should be enough. There is no need for years of battle testing.

koskimas:
And I'm not trying to rush you! Of course you should look into this, and thank you for doing that. Just putting all my thoughts out there 😄

elhigu:
@tgriesser Any updates on this one? Current knex release is still leaking connections (which will fill the pool eventually) when ever aqcuireConnectionTimeout is triggered...

tgriesser:
So 0.12.3 fixes the issue. Thanks for the PR but I think I'm going to just stick with generic-pool for now and instead focus on making it possible to supply the pool externally so folks can use their preferred pooling lib

koskimas:
Can you give me a reason why this was not merged? Tarn has only the features knex uses, it is better tested than generic-pool and you I would have given the ownership for you in addition to committing to work on it myself. generic-pool doesn't seem to be maintained anymore.

tgriesser:
A few reasons: 

- Was able to fix the reason for wanting to switch pools pretty easily in knex with a `setTimeout`
- Didn't have time to go through tarn, but was a bit confused by the recursion limit and the fact that tarn catches errors and logs them rather than forwarding them to the lib consumer. Also seems like there's a lot of promises for something which should generally be pretty simple.
- Not super concerned about the status of generic-pool maintenance since it generally works and is pretty well used across many libs
- Changing a dependency would require a minor version bump

Not ruling out revisiting this in the future if we have more pool issues, it just seemed simpler to fix it in knex land in this case. That said, I would like to make it simple to supply an external pool, will keep you updated on that.

koskimas:
Fair enough. The most important thing is that the bug is now fixed, and knex is stable. Thank you for fixing it.

Recursion limit is there to stop the _asynchronous_ recursion in some cases when the pool is so jammed that it would retry to create a resource too many times. It is probably not needed, but I wanted to be extra paranoid.

Errors are logged only when there is no one to inform about the error. Whenever someone is waiting for resource, the resource promise is rejected. If you found a place where this is not the case, then it's a bug.

Promises are used in places where generic-pool would poll using `setTimeout`. Promises are more elegant than polling don't you think? Also isn't the point of using promises to make the async flow simpler? That is why I used promises instead of callbacks.

elhigu:
IMO, this road where features, which should be responsibility of pool are implemented in knex side will come to an end at some point. Otherwise knex ends up implement all the meaningful functionality of pool. 

The biggest problem with generic pool is that lack of its maintenance forces knex to implement everything on knex side (e.g. maximum lease timeouts and maximum clients waiting to get lease).

Hope that this gets revisited at some point.

koskimas:
@tgriesser Did you see my comment on the test case you added for the bug? It doesn't actually test anything. I used the `done` callback for a reason in the test. I'm aware that mocha accepts promises.

fl0w:
Since this didn't get merged, @koskimas would you be against me publishing a monkey-patch-package, e.g. `knex-tarn`?

koskimas:
Currently tarn doesn't bring any new functionality to knex. Since @tgriesser fixed the bugs only reason to merge tarn is future development. I don't think we should create such a fork. Thanks for the offer though.

tgriesser:
@koskimas yep I updated the test. Also if you want to PR adding tarn against the refactor branch, I'd be up for exploring it a bit further... Specifically would like to see if tarn helps with the `unhandledRejection` I can't seem to figure out the source of in that branch.

PR #1707, opened 2016-09-29
Fix incorrect MSSQL query building with lock used, with clause and Number with decimal part conversion

statyan:
Fixed MSSQL incorect query build when locks are used. MSSQL requires lock hint be placed just after FROM clause. This colud be semi-fix, causeit wolud work for one-table query only. If there is any way to separate FROM and JOIN knex query build phase - I can "upgrade" my fix then.


statyan:
Hmm, why github shows all my commits ? Some of them were already pushed to knex master with previous pull request. Is this ok?

elhigu:
When merging this remember to use squash merge :)
screen shot 2016-11-24 at 14 21 28

statyan:
I'll make a tattoo "Use squash merge, Luke!" )

PR #1711, opened 2016-09-30
Making containsUndefined check for own properties

shimeez:


elhigu:
Hi, what is the motivation for this pull request? Also tests are required also for small changes like this.

dj-foxxy:
I've got a max call stack exceeded error that this pull request might be addressing. The error:
```
... RangeError: Maximum call stack size exceeded
at isFunction (/home/.../meteor/node_modules/lodash/isFunction.js:27:20)
at Object.containsUndefined (/home/.../node_modules/knex/lib/helpers.js:102:41)
...
```
I'm using knex@0.15.6 and lodash@4.17.2 as part of a Meteor application. The error occurs on the Node.js v4.6.2 server (not tested on client).

dj-foxxy:
I just hacked this commit into our code base and it stops it exploding. I have no idea of the wider impact but I'm sticking it in there 'cos I aint getting on where with out it. :)

elhigu:
I read through code where this is used. Seems legit and useful and shouldn't break anything.

Issue #416, opened 2014-08-07
CLI Improvements


rooftopsparrow:
It would be great to have a way to "dry run" a migration to see what would run as well as a flag to print the SQL for a set of migrations.

For example:

``` bash
$ knex --dry-run migrate:latest
Would migrate <x>
Would migrate <y>

$ knex --dry-run --print-sql migrate:latest
ALTER TABLE ...
DROP TABLE ...
```

That way you could easily see what would happen, and also export the sql to file for other purposes.

Let me know what you think.

kabirbaidhya:
I think this is a nice improvement to have. Why was this closed? 

And, is there any recent update on having this feature? If needed contributions for this, I can give a try and do a PR.

elhigu:
Probably due to inactivity or something like that. Looks like consensus has been that it is ok to implement them, but there might be some problems though like mentioned in https://github.com/tgriesser/knex/pull/1717

So PRs are welcome, but no need to reopen this. Also there were 2 different feature requests in this single issue. Two separate ones would be better to treat both separately.

kabirbaidhya:
Makes sense. I'll give it a try and will do a PR, can have further discussion on the PR itself.

unhawkable:
@kabirbaidhya i am looking forward! have you had any success implementing a dry-run option and showing sql queries without executing them?

PR #1717, opened 2016-10-04
feat(migrate): Added option for dry-run.

RWOverdijk:
Dry-run is a useful feature that allows you to _try_ your migration before applying them. This will show you if they'll pass or fail. It's not an exception, it just reverts the changes.

This isn't meant to be the _final_ PR. I assume there's more to it than this, I just wanted to open up a dialog.

**Update:** I just saw someone else misses the same feature: https://github.com/tgriesser/knex/issues/416

### DIFF
--- a/src/migrate/index.js
+++ b/src/migrate/index.js
@@ -25,6 +25,7 @@ const CONFIG_DEFAULT = Object.freeze({
   extension: 'js',
   tableName: 'knex_migrations',
   directory: './migrations',
+  dryRun: false,
   disableTransactions: false
 });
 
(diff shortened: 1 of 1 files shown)


tgriesser:
Seems like a reasonable addition!

RWOverdijk:
I'm pretty sure this won't cut it though. The way I've implemented it now for [wetland](https://github.com/SpoonX/wetland) is by wrapping the migrations and doing the rollback in there.

Adding in this feature would be nice though, I just need to know what I might have missed as I took the shortest route to accomplish what I wanted.

RWOverdijk:
I found out that this won't work for MySQL and oracle dbs. They don't support transactions for DDL queries, as they do an implicit commit. http://dev.mysql.com/doc/refman/5.7/en/implicit-commit.html

PR #1723, opened 2016-10-05
upgrading pg version

vitaly-t:
Upgrading `node-postgres` version to the very last one that's consistent and can be trusted.

See why: [node-postgres fixed @5.1](https://github.com/vitaly-t/pg-promise/issues/206).

And yes, I am the author of [pg-promise](https://github.com/vitaly-t/pg-promise), partially a competing library, but I don't mind contributing to knex anyway :)

Just so the users of both know what's up with the underlying driver ;)


tgriesser:
@vitaly-t thanks for the PR, but I'm a bit confused. If the intention of the PR was to fix it at 5.1, wouldn't you want to remove the `^`?

Also if knex doesn't use pg's built in connection pool, is it safe to assume there aren't any issues?

vitaly-t:
In this case it won't make a difference, as `node-postgres` is now `6.1`, and there won't be any fixes within `5.x` branch, but I am updating it, just in case ;)

chentsulin:
@elhigu This issue is outdated and should be closed.

Issue #1736, opened 2016-10-12
can't install knex 0.12.4


ravshansbox:
ravshan@Ravshans-MacBook-Pro:~/test $ npm install knex  

> knex@0.12.4 postinstall /Users/ravshan/test/node_modules/knex
> node ./scripts/build.js lib "npm run build"

module.js:457
    throw err;
    ^

Error: Cannot find module '/Users/ravshan/test/node_modules/knex/scripts/build.js'
    at Function.Module._resolveFilename (module.js:455:15)
    at Function.Module._load (module.js:403:25)
    at Module.runMain (module.js:590:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:509:3
npm WARN enoent ENOENT: no such file or directory, open '/Users/ravshan/test/package.json'
npm WARN test No description
npm WARN test No repository field.
npm WARN test No README data
npm WARN test No license field.
npm ERR! Darwin 16.0.0
npm ERR! argv "/usr/local/Cellar/node/6.7.0/bin/node" "/usr/local/bin/npm" "install" "knex"
npm ERR! node v6.7.0
npm ERR! npm  v3.10.7
npm ERR! code ELIFECYCLE

npm ERR! knex@0.12.4 postinstall: `node ./scripts/build.js lib "npm run build"`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the knex@0.12.4 postinstall script 'node ./scripts/build.js lib "npm run build"'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the knex package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node ./scripts/build.js lib "npm run build"
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs knex
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls knex
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/ravshan/test/npm-debug.log

yutin1987:
me to, run on `yarn install`

```
liuyutingdeMacBook-Pro-2:server yutin$ yarn install
yarn install v0.15.1
error npm-shrinkwrap.json found. This will not be updated or respected. See [TODO] for more information.
[1/4] 🔍  Resolving packages...
[...]

PR #1737, opened 2016-10-12
include scripts in npm package

calvinmetcalf:
fix for #1736

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,8 @@
     "lib/*",
     "knex.js",
     "LICENSE",
-    "CHANGELOG.md"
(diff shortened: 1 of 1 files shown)


calvinmetcalf:
better longer term solution might be to remove the postinstall

WebReflection:
Please "save" `0.12.x` with this change and flag as deprecated v `0.12.4`, thank you!

```
npm deprecate knex@0.12.4 "Fails the postinstall due missing ./script folder"
```

tgriesser:
Whoops, my mistake

PR #1740, opened 2016-10-13
Adding return's to address warnings mentioned in #1388

gordysc:
This addresses warnings generated because we don't explicitly return a promise as mentioned in Bluebird's API documentation:
http://bluebirdjs.com/docs/warning-explanations.html

### DIFF
--- a/src/runner.js
+++ b/src/runner.js
@@ -183,7 +183,7 @@ assign(Runner.prototype, {
   ensureConnection() {
     return Promise.try(() => {
       return this.connection || new Promise((resolver, rejecter) => {
-        this.client.acquireConnection()
+        return this.client.acquireConnection()
           .then(resolver)
           .catch(Promise.TimeoutError, (error) => {
             if (this.builder) {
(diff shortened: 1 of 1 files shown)


galenandrew:
This should resolve #1388

src/runner.js
```diff
@@ -183,7 +183,7 @@ assign(Runner.prototype, {
   ensureConnection() {
     return Promise.try(() => {
       return this.connection || new Promise((resolver, rejecter) => {
-        this.client.acquireConnection()
+        return this.client.acquireConnection()
```
elhigu: This place needs comment to code, which says something like "need to return promise or null from handler to prevent warning from bluebird"

Otherwise this return statement might be removed by accidents, since it doesn't serve any other purpose here.

src/runner.js
```diff
@@ -195,7 +195,7 @@ assign(Runner.prototype, {
           .catch(rejecter)
       })
     }).disposer(() => {
-      this.client.releaseConnection(this.connection)
+      return this.client.releaseConnection(this.connection)
```
elhigu: Also comment here.

src/runner.js
gordysc: Okay...this seems like a best practice though and we aren't explicitly calling it out elsewhere? I can certainly add the comments

gordysc:
@elhigu Comments have been added as per request.

PR #1748, opened 2016-10-19
Fix: config.pool default min may greater than user's config max

chenhui5416:
use case: 

```
    pool: {max: 1}
  })
```

Now, knex would thorw error, while pool max should greater than min.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -163,9 +163,11 @@ assign(Client.prototype, {
 
   poolDefaults(poolConfig) {
     const name = this.dialect + ':' + this.driverName + ':' + this.__cid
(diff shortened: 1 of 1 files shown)


elhigu:
Hi, thanks for the pull request. There is linting error and the test case for this is missing. 

I think it would be better to just throw error if min > max in that way there wouldn't be any hidden functionality which will hide clearly invalid configuration. Now if you pass `{ max: 2, min: 10 }` to pool, automatic changing min -> 2 will hide this configuration fail, where you probably wanted to pass `{ min: 2, max: 10 }` instead.

elhigu:
Closing for now, can be reopened if implementation is fixed and test case added.

PR #1766, opened 2016-10-31
Add sugar using ES6 template literals

villelahdenvuo:
This allows the following:

```
knex.tpl`
  select * from "sometable"
  join "othertable"
    on "other_field" = ${otherFieldValue}
  where "some_field" = ${someFieldValue}`
```
And the bindings will be automatically separated and passed to `knex.raw`. Not too different from named bindings, but much less code internally.

Any feedback?

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "bluebird": "^3.4.6",
     "chalk": "^1.0.0",
     "commander": "^2.2.0",
+    "common-tags": "^1.3.1",
     "debug": "^2.1.3",
     "generic-pool": "^2.4.2",
     "inherits": "~2.0.1",
--- a/src/client.js
+++ b/src/client.js
@@ -102,6 +102,10 @@ assign(Client.prototype, {
     return new Raw(this).set(...arguments)
   },
 
+  tpl() {
+    return new Raw(this).template(...arguments)
+  },
+
   _formatQuery(sql, bindings, timeZone) {
     bindings = bindings == null ? [] : [].concat(bindings);
     let index = 0;
--- a/src/raw.js
+++ b/src/raw.js
@@ -7,6 +7,7 @@ import { EventEmitter } from 'events';
 import debug from 'debug'
 
 import { assign, reduce, isPlainObject, isObject, isUndefined, isNumber } from 'lodash'
+import { stripIndent } from 'common-tags';
 import Formatter from './formatter'
 
 import uuid from 'node-uuid';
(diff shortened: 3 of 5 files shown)


elhigu:
sadly our test files are not compiled so I suppose you need to call it as a function 🤕

wubzz:
@elhigu @tuhoojabotti With Node 0.12 exiting LTS this month this should no longer be an issue, right?

The syntax does look a lot smoother than standard sql/bindings.

elhigu:
@wubzz it's up to @tgriesser to say what to do with node 0.12 I would be fine with dropping the support for it.

tuhoojabotti:
I see this was planned for the 1.0 refactor, so to me it's fine to wait and get it in 1.0, meanwhile we can use a small helper function.

elhigu:
Node 0.12 support was dropped some time ago, so with documentation + rebase this should be fine. Closing for now and can be reopened if completed.

borekb:
I know this is an old issue but if I'm not If this is ever done, [Sqorn](https://sqorn.org/) can be a great inspiration. Probably the best API I have seen, see examples [here](https://sqorn.org/docs/manual-queries.html).

elhigu:
Also slonik has something like that https://github.com/gajus/slonik to look into

pvider:
I would like to share my implementation using `knex.raw`:
```js
function raw(strings, ...bindings) {
        let sql = ''
        let prevString = ''
        for (let i = 0; i < strings.length; i++) {
            const string = strings[i]
            if (prevString.endsWith('\\$')) {
                sql = sql.slice(0, -2) + '$'
                if (i > 0) sql += '?'
            } else if (prevString.endsWith('$')) {
                sql = sql.slice(0, -1)
                if (i > 0) sql += '??'
            } else {
                if (i > 0) sql += '?'
            }
            sql += string.replace(/\?/g, '\\?')
            prevString = string
        }
        return knex.raw(sql, bindings)
    }```

Examples for PG (double quotes are for identifiers in PG):
```js
// identifiers are prefixed with another '$' without '\\' before it
raw`where $${'username'} is not null`.toString() //==> where "username" is not null
// values are not prefixed with another '$'
raw`where "username" = ${'Nickolas'}`.toString() //==> where "username" = 'Nickolas'
// '$' can be used before values if '$' has '\\' before it
raw`\\$${'abc'}`.toString() //==> $'abc'
// every '?' in string is escaped before passing to knex.raw
raw`???`.toString() //==> ???
```
I have tried to mimic usual `knex.raw` interface where you pass one '?' for values, '??' for identifiers and '?' can be escaped using '\\'. This should be consistent with current knex.raw.

PR #1775, opened 2016-11-08
Mssql does a rollback internally that knex doesn't take into account

RubenSlabbert:
As the title suggests I had a situation in my code where the following code would result in a `TransactionError EABORTED` from the internal mssql driver:

```javascript
knex.transaction(async trx => {
   const id = await trx.do_something_that_errors();
   return await trx.do_something_correct(id);
});
```

The above code works correctly when tested with sqlite (i.e. it returns a rejected promise based on the first function), but fails with the mssql driver. This would happen because the first database call would cause node-mssql to roll back internally, and then when the final function runs the "already aborted" error would be returned.

My patch fixes that by listening for the rollback event on the driver and ensuring that a rollback only runs once. I'm not sure if this is the fix that the maintainers feel most comfortable with (or if it is even correct) since I'm new to the codebase, but it works correctly for my project.

### DIFF
--- a/src/dialects/mssql/transaction.js
+++ b/src/dialects/mssql/transaction.js
@@ -6,6 +6,10 @@ export default class Transaction_MSSQL extends Transaction {
 
   begin(conn) {
     debug('%s: begin', this.txid)
+    conn.tx_.on('rollback', () => {
+      debug('%s: rolling back', this.txid)
+      this._completed = true
+    })
     return conn.tx_.begin()
       .then(this._resolver, this._rejecter)
   }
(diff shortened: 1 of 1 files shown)


elhigu:
I've missed this completely... I was googling a bit, I suppose you have `SET XACT_ABORT ON` it is off by default. Either way this makes sense to support recognising implicit rollback.

This change looks correct, but is missing tests.

@Grimace1975 any comments about this mssql issue?

kibertoad:
@RubenSlabbert Any chance you could write a test for this?

RubenSlabbert:
Apologies for the delay on this. It looks like this has already been fixed (with tests) in a different way in #1885. The fix seems functionally equivalent, however the linked PR makes it more clear what happened. I'll close this PR.

PR #1777, opened 2016-11-09
Mssql update query generates invalid syntax

RubenSlabbert:
The update query for SQL Server requires `from` when doing a join. The join must also be after the `set`. See [here](http://stackoverflow.com/questions/982919/sql-update-query-using-joins).


rhys-vdw:
Hi @RubenSlabbert. These look good. I don't use MSSQL, but have just looked at [this StackOverflow](http://stackoverflow.com/a/9589124/317135) answer to confirm that it's correct.

Thanks for your contribution. Not sure what is going on presently release wise, but I can merge to master.

PR #1784, opened 2016-11-17
allow to use `first` method as aliased select

skarbovskiy:
allow to write something like this

```js
let subselect = knex.first('column').from('table2 as t2')
  .where(knex.raw('t2.ref_id = t1.id'))
  .orderBy('t2.created_at', 'desc').as('column');

let query = knex('table1 as t1')
  .first('t1.*', subselect)
  .where({'t1.id': id, 't1.removed_at': null});
```

without necessity to do `knex.select().limit().as('alias')`

### DIFF
--- a/src/formatter.js
+++ b/src/formatter.js
@@ -142,7 +142,10 @@ export default class Formatter {
   outputQuery(compiled, isParameter) {
     let sql = compiled.sql || '';
     if (sql) {
-      if (compiled.method === 'select' && (isParameter || compiled.as)) {
+      if (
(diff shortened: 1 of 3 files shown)


elhigu:
Hi, i

If I understood correctly this is bugfix to a feature that is already documented?

As far as I can tell the changes seem reasonable, however this PR needs test case to make sure that this feature won't break afterwards. Please check `CONTRIBUTING.md`.

elhigu:
Thanks!

PR #1788, opened 2016-11-17
Use new safe buffer allocation to conform with deprecation of `new Buffer(size)`

### LINKED ISSUE #1763: DeprecationWarning: Using Buffer without `new`...
I'm getting a deprecation warning on node 7.0.0:

`DeprecationWarning: Using Buffer without `new` will soon stop working. Use `new Buffer()`, or preferably `Buffer.from()`, `Buffer.allocUnsafe()` or `Buffer.alloc()` instead.`

AlexTes:
Use of 'new Buffer' has been deprecated in node v6. This swaps the only
deprecated buffer creation in knex with a safe new way of doing allocation. It uses the
safe-buffer module to stay backwards compatible with node < v4. For the full story see [here](https://github.com/AlexTes/buffer-deprecation)

I'd also be totally open to switching out the safe-buffer shim with a ponyfill.

The deprecated method of buffer creation concerns the oracle dialect. I tried to run the tests but failed to do so. If someone who can run the `DB='oracle' npm test` command would try it I'd be very grateful.

To be clear, this doesn't fix #1763. That issue is with knex' dependencies still. In particular node-postgres. However now that `new Buffer(size)` has also been deprecated this should be merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "node-uuid": "^1.4.7",
     "pg-connection-string": "^0.1.3",
     "readable-stream": "^1.1.12",
+    "safe-buffer": "^5.0.1",
     "tildify": "~1.0.0",
     "v8flags": "^2.0.2"
   },
--- a/src/dialects/oracledb/index.js
+++ b/src/dialects/oracledb/index.js
@@ -13,6 +13,7 @@ const helpers = require('../../helpers');
 const Transaction = require('./transaction');
 const Client_Oracle = require('../oracle');
 const Oracle_Formatter = require('../oracle/formatter');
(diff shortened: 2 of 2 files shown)


AlexTes:
Locally the tests pass for me against the default db's. Also for node 0.12.7 . Since it concerns a timeout test (I'll carefully put the blame elsewhere and suggest) it might be the hardware difference?

elhigu:
It was just some random failure. In travis some times DB queries are really slow...

elhigu:
Thanks!

AlexTes:
Happy to help! Thanks for great module!

PR #1796, opened 2016-11-23
Fix broken import in knex.js

timwis:
Unless I've misread, I think this link is broken.

### DIFF
--- a/knex.js
+++ b/knex.js
@@ -5,4 +5,4 @@
 //     For details and documentation:
 //     http://knexjs.org
 
-module.exports = require('./lib/index');
+module.exports = require('./src/index');


elhigu:
Looks like our build system is having some serious problems...

elhigu:
I suppose that this is not a bug. As far as I know this file is not compiled.

timwis:
Huh? `package.json`'s `main` property points to `knex.js`. `knex.js` exports `./lib/index`, which doesn't exist. How is that not broken?

tgriesser:
It exists when the lib is babel compiled, check the node_modules files when it's installed from npm

timwis:
Ah! Thought I was going crazy there.. thanks!

PR #1804, opened 2016-11-24
updated timestamps function

jonstuebe:
adding more options to the timestamps function to allow for:

- auto updated CURRENT_TIMESTAMP for updated_at
- adding a deleted_at column.


elhigu:
Linting fails, no documentation, no tests and I'm not big fan of having 4 true parameters for the function and I wouldn't like to include `deletedAt` here anyways (I suppose it would be some field for soft-delete implementation). 

Though mostly I don't care much how timestamps function works, I never use it myself since it doesn't handle automatic updates for modified timestamp and uses hard coded snake-case naming (overall isnt generic enough). IMO this helper shouldn't even be implemented in knex-level.

Maybe someone who uses this and has stronger opinion how it should work could add some input here?

jonstuebe:
@elhigu ya I can see where you're coming from. I might make a little companion library to knex with some specific migration helpers.

PR #1769, opened 2016-11-03
Support knexfile, migration and seeds in TypeScript

prashaantt:
This will allow using `knex` in a TypeScript project. These commands will now work:

```bash
$ knex init -x ts
$ knex migrate:make migration_name
$ knex seed:make seed_name
```

`knex` will now generate and automatically pick up `knexfile.ts`, `migration_name.ts`, `seed_name.ts`, assuming [`ts-node`](https://github.com/TypeStrong/ts-node) is available to execute from within the project.

If you like this, I'll also update the docs to indicate that this option is now available.

PR #1814, opened 2016-12-01
Update stub examples to not use Promise.all

prashaantt:
Per discussion in https://github.com/tgriesser/knex/pull/1769.


elhigu:
Thanks!

Issue #1628, opened 2016-08-16
knex.schema.createTable does not work


alexey2baranov:
Hello! this code does not create table in my local DB. No error throwened and no console output. What's wrong with it?
```
var knex = require('knex')({
    client: 'pg',
    connection: {
        host     : '127.0.0.1',
        user     : 'postgres',
        password : '***',
        database : 'kp-bookshelf',
        charset  : 'utf8'
    }
});

knex.schema.createTable('user', function(table) {
    table.increments();
    table.string('name');
    table.string('email', 128);
    table.string('role').defaultTo('admin');
    table.string('password');
    table.timestamps();
});
```

wubzz:
@alexey2baranov You need to call `.then` or `.asCallback`

alexey2baranov:
@wubzz 
where ? after createTable()? if so, nothing changed

alexey2baranov:
oh sorry table was created! i was surprised. why should i use then()???

wubzz:
@alexey2baranov Knex is a query-building library, so simply calling `createTable` will do nothing except generate the SQL string.

This will show you the generated SQL:
```javascript
console.log(knex.schema.createTable('user', function(table) {
    table.increments();
    table.string('name');
    table.string('email', 128);
    table.string('role').defaultTo('admin');
    table.string('password');
    table.timestamps();
}).toString());
```

To have knex run the SQL, you need to call `.then` or `.asCallback` as an execution function. This can also be chained with other promises.

alexey2baranov:
yes it works but anyway i can't understand why I should use then()?? as I can see when working with promises Promise body function always executed not matter with then() or without then()

alexey2baranov:
@wubzz 
yes it works but anyway i can't understand why I should use then()?? as I can see when working with promises Promise body function always executed not matter with then() or without then()

tgriesser:
Because in knex a promise isn't actually created internally until `.then` is called, for historical reasons (and which is something I'm looking to change eventually). The reason is that you can chain several schema building calls:

```js
knex.schema
  .createTable('user', function(table) {
    table.increments();
    table.string('name');
    table.string('email', 128);
    table.string('role').defaultTo('admin');
    table.string('password');
    table.timestamps();
  })
  .createTable('account', function(table) {
    table.increments();
    table.string('name');
    table.string('email', 128);
    table.timestamps();
  })
```

And then you call `.then` when you want to execute.

[...]

PR #1822, opened 2016-12-08
Ensure that 'client' is provided in knex config object

wubzz:
The option is documented as required, but it's currently not being validated. This leads to strange side effects such as ones describe in #1628.

I also took the liberty of adding `modifiers` to the base columpcompiler as it currently does not have this property, but it should.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -37,6 +37,14 @@ function clientId() {
 // for a dialect specific client object.
 function Client(config = {}) {
   this.config = config
+
+  //Client is a required field, so throw error if it's not supplied.
+  //If 'this.dialect' is set, then this is a 'super()' call, in which case
+  //'client' does not have to be set as it's already assigned on the client prototype.
+  if(!this.config.client && !this.dialect) {
+    throw new Error(`knex: Required configuration option 'client' is missing.`)
+  }
+
   this.connectionSettings = cloneDeep(config.connection || {})
   if (this.driverName && config.connection) {
     this.initializeDriver()
--- a/src/dialects/oracle/schema/columncompiler.js
+++ b/src/dialects/oracle/schema/columncompiler.js
@@ -9,8 +9,8 @@ import Trigger from './trigger';
 // -------
 
 function ColumnCompiler_Oracle() {
-  this.modifiers = ['defaultTo', 'checkIn', 'nullable', 'comment'];
   ColumnCompiler.apply(this, arguments);
+  this.modifiers = ['defaultTo', 'checkIn', 'nullable', 'comment'];
 }
 inherits(ColumnCompiler_Oracle, ColumnCompiler);
 
--- a/src/dialects/sqlite3/schema/columncompiler.js
+++ b/src/dialects/sqlite3/schema/columncompiler.js
@@ -6,8 +6,8 @@ import ColumnCompiler from '../../../schema/columncompiler';
 // -------
 
 function ColumnCompiler_SQLite3() {
-  this.modifiers = ['nullable', 'defaultTo'];
   ColumnCompiler.apply(this, arguments);
+  this.modifiers = ['nullable', 'defaultTo'];
 }
 inherits(ColumnCompiler_SQLite3, ColumnCompiler);
 
--- a/src/schema/columncompiler.js
+++ b/src/schema/columncompiler.js
@@ -18,6 +18,7 @@ function ColumnCompiler(client, tableCompiler, columnBuilder) {
   this.isIncrements = (this.type.indexOf('increments') !== -1);
   this.formatter = client.formatter();
   this.sequence = [];
+  this.modifiers = [];
 }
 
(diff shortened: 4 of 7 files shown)


eddieajau:
There is a small regression with this. Previously when calling `knex migrate:make` the config option was not required. Now `knex migrate:make` is not correctly configured in the local development environment (but would be configured correctly when knex was actually run in a production environment).

venables:
I disagree with throwing the exception here.  One of the documented features of Knex is:

> You can even use knex without a connection, just for its query building features. Just pass in an empty object when initializing the library. Specify a client if you are interested in a particular flavour of SQL.
> ```var knex = require('knex')({});```
> ```knex('table').insert({a: 'b'}).returning('*').toString();```
> ```// "insert into "table" ("a") values ('b')"```

I think we should revert the Exception bit, and treat the bug as [fixed in the PR](https://github.com/tgriesser/knex/pull/1822/files#diff-ea947a48c758aaa1b0c1482d8091496e) by adding this.modifiers = [].

On a versioning note, this was also a breaking change for us (much of our code doesn't initialize Knex with a database connection until later in app setup), and likely deserved a minor version bump.

elhigu:
Looks like this one actually was a breaking change, since there kind of was documented.  I'm fine with reverting and releasing fixed version as 0.12.8 with explication in changelog. 

I don't like the feature of initializing builder without dialect, but it clearly was intended to work. @wubzz does it sound ok to you? or was there actually documentation somewhere that client was required configuration parameter?

elhigu:
Actually yeah... other issue actually did state the place where it was mentioned.


> Initializing the Library
> 
> The knex module is itself a function which takes a configuration object for Knex, accepting a few parameters. **The client parameter is required and determines which client adapter will be used with the library**.

So I think we should remove the invalid part of the documentation. I would prefer to have client as required, because currently trying to use knex without client is very badly tested and probably will keep on breaking in the future.

But since there was inconsistent documentation about the feature this should have been made with minor version bump instead of patch version...

venables:
Thanks @elhigu - I have found the "client-less" initializer to have been helpful over the years, but I understand your concerns.  

I am happy to help write better test coverage for the feature if that's desired.

Another option would be to allow setting the client like `client: "none"` -- forcing it to be explicitly called out could help prevent user errors and confusion (which seems to be the goal here, right?)

elhigu:
@venables I still don't understand why clientless initialization is more helpful that for example just using `{ client: 'pg' }` initialization? I understand initializing without connection to be able to test queries, but without client the generated SQL is really not runnable anywhere...

Anyways if that feature is really wanted and better tested in that case `{ client : "none" }` sounds like a good way to initialize the clientless builder.

venables:
I won't have all the use-cases for clientless initialization, but the use-cases that I've hit professionally have been:

1. Unit tests. We use this feature heavily in our tests.  The knex project uses/used it in tests as well.
2. Needing to define a Bookshelf model before the database connection is established.  We Initialized Bookshelf with a client-less Knex instance and swapped the Knex instance once Knex was connected.

All things we can work around, but I think it is a really great feature -- even if it were hid behind some sort of advanced `none` setting which uses the base schema class.

This is a feature that Knex was advertising as a good thing, and I agree.

wubzz:
@elhigu I still stand firm that 'client' should always be required. It's unfortunate that the documentation has double standards. The use cases above don't sound too convincing to me. One could just as easily supply `client` in unit tests and the second use case sounds like a 'cheap trick' (no offense) that the library should support in a different way.

elhigu:
@venables In both of those use cases one can initialize `{ client: 'pg' }` or `{ client: 'sqlite3' }`. Only thing which just will be inconvenient is that unit test result checking might need to be fixed a little. However `{ client: 'pg' }` query generation is very near to current clientless output.

I still have not seen any case where `{ client: 'none' }` would be any better choice that `{ client: 'pg' }` or any other real client without giving connection details for query unit testing.

You mentioned about using the base schema class... indeed most of the dialect differences are in schema. Is there any specific reasons why using base schema would be desired instead of some real schema of real dialect?

For now I feel that the best choice would be to revert throwing the error, update changelogs and release 0.12.8. 

Then reapply the fix and release 0.13.0 and mark it as a breaking change and fix the documentation.

venables:
@wubzz @elhigu Thanks for the response.  The unit test use-case was my primary concern, but it looks like the database connection pool is not actually created until a query is executed -- which might be another piece of confusion from the documentation:

> Initializing the library should normally only ever happen once in your application, as it creates a connection pool for the current database, you should use the instance returned from the initialize call throughout your library.

So I think we should be all set by setting our tests to `client: 'pg'` like you suggest.  Thank you again for looking into this.

I'll close the Revert PR -- I'll leave the decision to you guys about 13.0 vs 12.8

TimonLukas:
My use case for this was to use Knex with IBMs DB2. Without specifying a client it actually generated valid queries, which the other client types do not seem to do. Oh well, I just resorted to manually writing my queries.

PR #1834, opened 2016-12-16
Implement query cancellation in sqlite3 dialect

xmirya:
node-sqlite3 provides an interface to sqlite3_interrupt() since 3.1.4, though not mentioned in the documentation; it can be used as-is for Knex query cancellation interface


elhigu:
Interruption test should be added for sqlite dialect.

elhigu:
Closing older incomplete PRs, can be reopened if completed.

PR #1835, opened 2016-12-16
Fix: ES6 Template literals in migration stubs

villelahdenvuo:
Override lodash templateSettings.interpolate because by default it interpolates ${variable} syntax.

For example:
```js
> console.log(_.template('foo(`${DBP}`) <%= d.foo %>', { variable: 'd' }).source)
function(d) {
var __t, __p = '';
__p += 'foo(`' +
((__t = (DBP)) == null ? '' : __t) +
'`) ' +
((__t = ( d.foo )) == null ? '' : __t);
return __p
}
```

But if we don't use the default RegExp instance, but a clone instead, it works: 
```js
> console.log(_.template('foo(`${DBP}`) <%= d.foo %>', { variable: 'd', interpolate: /<%=([\s\S]+?)%>/g }).source)
function(d) {
var __t, __p = '';
__p += 'foo(`${DBP}`) ' +
((__t = ( d.foo )) == null ? '' : __t);
return __p
}
```

Due to:
https://github.com/lodash/lodash/blob/4.17.2/lodash.js#L14754

### DIFF
--- a/src/migrate/index.js
+++ b/src/migrate/index.js
@@ -261,7 +261,7 @@ export default class Migrator {
     const stubPath = this.config.stub ||
       path.join(__dirname, 'stub', this.config.extension + '.stub');
     return Promise.promisify(fs.readFile, {context: fs})(stubPath).then(stub =>
-      template(stub.toString(), {variable: 'd'})
+      template(stub.toString(), {variable: 'd', interpolate: /<%=([\s\S]+?)%>/g})
     );
   }
 


elhigu:
Looks good. Test case would be awesome to have, is there any problems with testability of this?

tuhoojabotti:
@elhigu I didn't have time to look into the tests yet. I had to get my migrations working, but I thought that I should share information I found while debugging the issue. I will look into adding tests, but I'm not sure when I will have time.

elhigu:
Closing older unfinished PRs, can be reopened if completed.

Issue #1798, opened 2016-11-23
Unhandled Exception in `connection.end()` for MySQL Dialect


I00000:
Unfortunately, the fix for https://github.com/tgriesser/knex/issues/1691 does not work because the connection.end continues to crash my node. In my context, I use a pool of connections and I get a timeout from MySQL server due to network problems or server problems (e.g. DB restart). All of these events are considered fatal errors.

In my code, I have tried to catch them like that:
```javascript
pool: {
    min: 2,
    max: 7,
    afterCreate: function (connection, callback) {
        connection.on('error', function (err) {
            console.log('>>>>>>>>>>>>>>>>>>>>>>> err', err);
        });
        callback(null, connection);
    }
}
```
Unfortunately, my handler is never called due to the removeAllListeners call in destroyRawConnection:
```javascript
destroyRawConnection(connection) {
    connection.removeAllListeners()
    connection.end(() => {
        if (err) connection.__knex__disposed = err
    });
}
```
I tried the 2 different patches in destroyRawConnection:
```javascript
destroyRawConnection(connection) {
    connection.end(() => {
       connection.removeAllListeners()
        if (err) connection.__knex__disposed = err
    });
}
```
Unfortunately, this is one does not work but the following work well
```javascript
destroyRawConnection(connection) {
    connection.end(() => {
        if (err) connection.__knex__disposed = err
    });
}
```
For this one, I have changed a lit bit my handler
```javascript
pool: {
    min: 2,
    max: 7,
    afterCreate: function (connection, callback) {
        connection.on('error', function (err) {
           this.removeAllListeners();
            console.log('>>>>>>>>>>>>>>>>>>>>>>> err', err);
        });
        callback(null, connection);
    }
}
```
After some tests, everything seems OK.

elhigu:
Hi, if you provide some kind of test code e.g. against `knex_test` database so I could try to reproduce the problem. @wubzz also knows a lot (more than me) about knex connection issues.

Maybe in that test you could emulate DB restart by killing all connections for example like this http://stackoverflow.com/questions/1903838/how-do-i-kill-all-the-processes-in-mysql-show-processlist

I00000:
This code allows to reproduce the issue:

server.js
```javascript
'use strict';

var express = require('express');
var knex = require('knex');
var port = 7000;
// configuration ===============================================================
var db = knex({
    client: 'mysql2',
    connection: {
        debug: false,
        host: 'localhost',
        port: 3306,
[...]

Issue #1691, opened 2016-09-24
Unhandled Exception in `connection.end()` for MySQL Dialect


brycekbargar:
[There's an open issue on the mysql repo](https://github.com/mysqljs/mysql/issues/1334) for ClearDB not responding the a Quit packet.

This affects knex when destroying an idle connection using [`Client_MySQL.destroyRawConnection()`](https://github.com/tgriesser/knex/blob/master/src/dialects/mysql/index.js#L80).

```js
  destroyRawConnection(connection) {
    // This removes the error handler attached by knex in `aquireRawConnection()`
    // It also removes any error handlers attached by the end user
    connection.removeAllListeners()
    // In the mysql library this takes an error callback, or alternatively emits the 'error' event
    connection.end()
  }
```

The unhandled exception thrown by mysql on `connection.end()` was crashing my application and there is really no way to handle it as knex is currently setup.

I would be willing to open a PR and fix the bug, but I'm not sure which strategy is most appropriate.

The easiest would be to remove the call to `removeAllListeners()`, [according to the NodeJS docs](https://nodejs.org/api/events.html#events_emitter_removealllisteners_eventname)

> "Note that it is bad practice to remove listeners added elsewhere in the code, particularly when the EventEmitter instance was created by some other component or module (e.g. sockets or file streams)."

That said, I'm assuming there is a reason knex is calling this method? Maybe something to do with performance?

An alternative would be to pass in an error callback to `.end()', for example:
```js
  destroyRawConnection(connection) {
    connection.removeAllListeners()
    connection.end(err => {
      connection.__knex__disposed = err
    })
  }
```

Let me know which approach would be best and I'll make the change and open a PR.

As a workaround, for anyone else experiencing issues with knex, heroku, mysql, and cleardb I've done the following and it works for now but is fairly ugly...

```js
config.pool.afterCreate = (conn, cb) => {
  const removeAllListeners = conn.removeAllListeners;
  conn.removeAllListeners = () => {
    removeAllListeners.call(conn);
    conn.on('error', err => conn.__knex__disposed = err);
  };
  cb(null);
};
```

tgriesser:
> The unhandled exception thrown by mysql on connection.end() was crashing my application and there is really no way to handle it as knex is currently setup.

Yikes, sorry about that.

> That said, I'm assuming there is a reason knex is calling this method? Maybe something to do with performance?

[...]

PR #1853, opened 2017-01-07
Unhandled Exception in `connection.end()` for MySQL Dialect

I00000:
fix for https://github.com/tgriesser/knex/issues/1798

@eelhigu: please review?

### DIFF
--- a/src/dialects/mysql/index.js
+++ b/src/dialects/mysql/index.js
@@ -78,7 +78,6 @@ assign(Client_MySQL.prototype, {
   // Used to explicitly close a connection, called internally by the pool
   // when a connection times out or the pool is shutdown.
   destroyRawConnection(connection) {
-    connection.removeAllListeners()
     connection.end(err => {
       if (err) connection.__knex__disposed = err
     })


elhigu:
I'm on my mobile so I cannot be browsing the code right now to verify if this is ok... do you know why that removeAllListeners call is there and will there be event listeners left if its not called anymore? @wubuzz do you know anything about this?

I00000:
hi @tgriesser,

In your following commit https://github.com/tgriesser/knex/commit/7069ce509ed1ed6b1c3a0e9dde8d27a29bfcb30d#diff-b0704a0e57b09b0c3670a2c6b953da1d, you have added that connection.removeAllListeners(), do you know why because in my scenario, it crashes my server and I don't really understand why this layer is responsible for cleaning all listeners.

Thanks in advance for your clarification

wubzz:
@elhigu @I00000 Sorry for the slow response.. I know we've had issues about this in the past, specifically for mysql. I'm not sure why `.removeAllListeners()` was added tbh, I'd think calling `.end()` makes this call redundant. Only @tgriesser would know the answer to this, so I can't really make a decision here.

However, I don't quite understand why the following would not work:
```javascript
connection.end((err) => {
       connection.removeAllListeners();
        if (err) connection.__knex__disposed = err
});
```

What error remains after this change? Since the event listeners aren't removed until *after* `.end()` has been called, surely it should work..

I'm also inclined to agree with #1691 reference to the Node docs:
`Note that it is bad practice to remove listeners added elsewhere in the code`

I00000:
node crahes

novemberborn:
I think the problem is that `mysql2` transitions connections to a broken state, and the dialect in Knex does not recognize this. #2175 seems to fix it for me.

elhigu:
Closing in favor of #2175

PR #1854, opened 2017-01-07
limit allow Infinity

MrQubo:
Consider Infinity as valid value for limit.
It should set no limit at all.


elhigu:
Sounds like a hack to me for trying to circumvent limitation that knex doesn't support removing parts from builder... Passing infinity to limit is usually done by error so I would vote no for this one. 

Other than that, documentation and tests are missing.

elhigu:
Adding info about related issue https://github.com/tgriesser/knex/pull/1907 that has better API for clearing stuff. This should be with same kind of implementation + tests and docuementation.

PR #1874, opened 2017-01-20
ONLY keyword support for PostgreSQL

tomaspinho:
Hello, we're requiring the ability to use the `ONLY` keyword  in PostgreSQL and I have done a crude implementation of it in this PR.
Please discuss and provide any comments that would enable this or other PR to be merged that implements the same functionality.
Thank you!

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -138,11 +138,12 @@ assign(Builder.prototype, {
   // Sets the `tableName` on the query.
   // Alias to "from" for select and "into" for insert statements
   // e.g. builder.insert({a: value}).into('tableName')
-  // `only`: wether the query should use SQL's ONLY to not return inheriting table data.
-  // Defaults to false.
-  table(tableName, only = false) {
+  // `options`: options object containing keys:
+  //   - `only`: whether the query should use SQL's ONLY to not return
+  //           inheriting table data. Defaults to false.
+  table(tableName, options = {}) {
     this._single.table = tableName;
-    this._single.only = only;
+    this._single.only = options.only !== undefined ? options.only : false;
     return this;
   },
 
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -143,7 +143,7 @@ assign(Builder.prototype, {
   //           inheriting table data. Defaults to false.
   table(tableName, options = {}) {
     this._single.table = tableName;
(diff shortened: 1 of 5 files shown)


elhigu:
Sounds like ONLY is actually not postgresql specific functionality, but comes from standard... any ideas if mysql or others support it? 

Generally pull request looks good, change is so simple that I'm fine that there is no integration tests for this since the change is so simple. 

One thing there is that I would like to change. Instead of true/false parameter would be more future proof to pass `{ only: true }` object instead.

Could you make also pull request to documentation repository and add link to it here, so I can check that documentation will be

elhigu:
and thanks for contribution 👍 !

tomaspinho:
Hey @elhigu! Thanks so much for the feedback. From Google searches, I would say Postgres is the only one supporting it, as other engines don't seem to support table inheritance in the same way (but I may be wrong, not a DBA).

I'll see if I can take a look at your request this coming week and update this PR accordingly.

tomaspinho:
Hey @elhigu,

You may find the new docs here: https://github.com/knex/documentation/pull/16
I've already implemented the options object, but the build seems to have failed in an integration test unrelated to this PR. Could you please make it run again? Thanks!

src/query/builder.js
```diff
@@ -138,8 +138,12 @@ assign(Builder.prototype, {
   // Sets the `tableName` on the query.
   // Alias to "from" for select and "into" for insert statements
   // e.g. builder.insert({a: value}).into('tableName')
-  table(tableName) {
+  // `options`: options object containing keys:
+  //   - `only`: whether the query should use SQL's ONLY to not return
+  //           inheriting table data. Defaults to false.
+  table(tableName, options = {}) {
     this._single.table = tableName;
+    this._single.only = options.only !== undefined ? options.only : false;
```
wubzz: Try to prevent negated conditions where possible. Here it would make more sense with explicit checks `this._single.only = options.only === true` as it would also ensure the property `only` is always a boolean. :)

tomaspinho: taken care of ;) thanks!

elhigu:
Thanks @tomaspinho

tomaspinho:
Thank you, @elhigu :smile:

Issue #1884, opened 2017-01-27
MSSQL aborted transactions cause Promise 'unhandled rejection' error


justinwh:
The mssql dialect in knex doesn't handle errors that happen during rollback, so when rollback fails, bluebird prints an "Unhandled rejection" error to the console, and execution hangs.

For background, rollback does fail in MSSQL-- notably when an error causes the transaction to "abort".  When this happens, the server automatically rolls back the transaction, so when the client issues its own ROLLBACK, the server returns an error saying there's no current transaction.

Notes:

- [Microsoft's documentation for the "abort" behavior, which can be turned on, but cannot be completely turned off](https://msdn.microsoft.com/en-us/library/ms188792.aspx)

elhigu:
Thanks for the fixes!

PR #1885, opened 2017-01-27
Pass rollback errors through to transaction promise in mssql dialect …

justinwh:
…(#1884)

Check for rollback failures on mssql and pass them through to the transaction promise-- attaching the underlying cause as an 'originalError' property.  This fixes a hang when mssql server aborts a transaction and the client issues a rollback.


elhigu:
@Grimace1975 could you check this one?

Issue #1887, opened 2017-01-30
PG Raw Binding Bug introduced in 0.11 


Sammons:
I kept running into this error upon upgrading from 0.9.0 to 0.12.6
```
Undefined binding(s) detected when compiling RAW query:
```
with this string going in
```
var s = "\n            org\n            ON provider.id = \n                JSONB_EXTRACT_PATH_TEXT(org.meta, 'settings','credentials', :provider_name )::UUID\n            "
```
and I *knew* my bindings were defined.

So it happens that in the repl, when I test the regex used in the raw.js `replaceKeyBindings` function to match parameters on the aforementioned string, I get a match of

`[ ':provider_name', ':UUID' ]`

The regex in `replaceKeyBindings` in 0.12.x  is `var regex = /\\?(:\w+:?)/g;`

in 0.11.x it appears that the regex stopped checking that the character before the colon was a space/beginning of string. What it probably should have been checking was that the character before was just not a colon.

It makes sense that the whitespace check was removed, because `(:provider_name):UUID` wouldn't work otherwise.

Please note that this issue https://github.com/tgriesser/knex/pull/1661 did not touch on this problem.

elhigu:
Maybe binding replacement could be done in reverse... so we check first, which bindings has been given to raw and after that convert only those occasions from sql string to positional bindings.

So  `knex.raw('select :foobar')` would try to send `select :foobar` to driver, but `knex.raw('select :foobar', { foobar: 1 })` would send `select 1`  and also allow escaping `:` in the same way that `?` can be escaped.

Sammons:
@elhigu that would solve my problem! If a binding is left out or is undefined it would ultimately cause a failed response from PG anyways (as opposed to just throwing an error in knex).

elhigu:
Fixed by https://github.com/tgriesser/knex/pull/1890 .  big thanks to @mathewdgardner

PR #1890, opened 2017-02-02
Allow casting when using named bindings.

mathewdgardner:
I have had lots of trouble using named bindings when writing raw queries in the past but recently I found that most of my trouble occurs if I am casting. This PR expands the regex used for searching for identifies and values and adds a little bit of logic around it.

This may be a naive approach but I'm open to critique.

## Before Examples:

```sql
SELECT foo::TEXT 
FROM <table>
WHERE bar = :bar;
```

Currently, this will yield an error complaining about undefined bindings. What's happening behind the scenes is `::TEXT` is being picked up as a named binding but of course it is missing from the given bindings object.

```sql
SELECT :foo::TEXT 
FROM <table>;
```

Currently, this will result in `:foo:` being thought of as an identifier instead of a value being given and casted.

After some digging I found that this undocumented approach works in the current version:

```sql
SELECT foo:\:TEXT 
FROM <table>
WHERE bar = :bar;
```

While this does not:

```sql
SELECT :foo:\:TEXT 
FROM <table>
```

## After Examples:

These work as expected.

```sql
SELECT foo::TEXT 
FROM <table>
WHERE bar = :bar;

SELECT :foo::TEXT 
FROM <table>;
```

I mostly work on Postgres and so my knowledge of other RDBMS's is very limited. As I mentioned above I am open to suggestions for improvement.

### DIFF
--- a/src/raw.js
+++ b/src/raw.js
@@ -146,40 +146,24 @@ function replaceRawArrBindings(raw, formatter) {
 }
 
 function replaceKeyBindings(raw, formatter) {
-  const values = raw.bindings
-
   let { sql } = raw
 
-  const regex = /\\?(::?\w+:?:?)/g
-  sql = raw.sql.replace(regex, function(full, part) {
-    if (full !== part || part.slice(0, 2) === '::') {
-      return part
+  sql = raw.sql.replace(/\\?(:(\w+):(?!:)|:(\w+):??)/g, function(match, p1, p2, p3) {
+    if (match !== p1) {
+      return p1
     }
 
-    const key = full.trim();
-    const isCasting = key.slice(-2) === '::'
-    const isIdentifier = key[key.length - 1] === ':' && !isCasting
-
-    let value;
-    if (isIdentifier) {
-      value = values[key.slice(1, -1)]
-    } else if (isCasting) {
-      value = values[key.slice(1, -2)]
-    } else {
-      value = values[key.slice(1)]
-    }
+    const key = match.trim();
+    const isIdentifier = key[key.length - 1] === ':'
+    const value = raw.bindings[p2 || p3]
 
     if (value === undefined) {
-      formatter.bindings.push(value);
-      return full;
+      return match
+    } else if (isIdentifier) {
+      return match.replace(p1, formatter.columnize(value))
     }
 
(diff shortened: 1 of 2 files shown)


elhigu:
Hi, thanks for looking into this I'll give more careful look to the implementation a bit later..

On the first glance looks  like you have used template strings in tests... sadly we still cannot use them in tests, since they are ran also with node 0.12. So this causes test failure.

Also if I understood correctly your implementation checks just against `::` operator... so this doesn't solve all the issues with intended colons in queries.

There is a bug report about the same issue, which has also proposal for the solution to the problem that will work in more general way to all colons in the query https://github.com/tgriesser/knex/issues/1887#issuecomment-276220695

mathewdgardner:
@elhigu Thanks for looking into this. I'm not aware of any intended uses of `:` besides using named bindings for identifiers and values, can you provide an example so I can better understand?

elhigu:
@mathewdgardner do you think you could change the implementation to work so that only those named bindings and converted to positional which are found from bindings object?

mathewdgardner:
@elhigu I'll take a stab at it.

mathewdgardner:
@elhigu Let me know if this implementation works you or if I need to squash.

I expanded the regex to either match a named binding identifier that is not followed by a casting or a named binding value otherwise. Casts will still be picked up by the regex but it will only use the given bindings for replacement. If a binding given is undefined, it will still pass it along; that behavior is unchanged.

elhigu:
So... casting `:column:::text` won't work I suppose, but one has to do it like `:comlumn: ::text`? I think that is ok. A bit inconsistent is that now in case of `:var` match eagerly `:var:` match eagerly, but `:var::` does not match eagerly but matches to `:var`.

We could make special cases for `:column:::bigint` and `:value::bigint` cases in addition to only replace stuff that has been given in bindings, but to me current solution is good enough.

Would be great to add a paragraph about how named bindings work in these special cases where syntax is a bit unambiguous.

Code looks good. Consider those last comments if you like to do some more tuning and let me know. If you feel this is ready in current form I can already merge this in.

elhigu:
Awesome thanks a lot!

elhigu:
added issue about documentation

PR #1894, opened 2017-02-03
Added ability to use certain bluebird functions

timhuff:
Added the `.get`, `.mapSeries`, and `.delay` bluebird functions. The most significant omission from the current release would probably be `.get`. If you're selecting a single row from the database, it's reasonable to want to just write `knex.select().from('users').where({username: 'Jim'}).limit(1).get(0)`.


elhigu:
This seems to be done on top of your earlier pull request. Is that necessary?

timhuff:
@elhigu Whoops. I'm sorry - I don't make many PRs. I didn't realize that the PR would change when I modified my repo (thought it would be fixed to the hash that the branch was pointing to at the time of PR creation). I'll put these back in place in my repo.

PR #1895, opened 2017-02-03
.where() can now accept a query builder (#1893)

### LINKED ISSUE #1893: [Proposal] Stand-alone Where Clauses
I wanted to bounce a feature idea off the community here.

I think it'd be nice to have standalone where clauses that you could pass around. My proposal for the API would be:
```javascript
var whereClause = knex.where({state: 'CA'});
knex.select().from('users').where(whereClause);
```

The purpose here would be the ability to pass where clauses into functions. So you could make a function that looks like this:
```javascript
var getUsers = function(whereClause){
    return knex.select().from('users').where(whereClause);
};
```
and then call it with:
```javascript
getUsers(knex.where({state:'CA'}));
```

(Obviously, this would be a useless abstraction - just trying to keep the example simple)

I, personally, have a use case where this functionality would be useful. In the example above, the solution might be to just `getUsers().where({state:'CA'})` but for my use case, the where clause is actually applied to a subquery.

I currently solve the problem by defining the function like so:
```javascript
var getUsers = function(userSelector){
    return knex.select().from('users').whereIn('username', userSelector);
};
```
and then executing it like so:
```javascript
getUsers(knex.select('username').from('users').where({state: 'CA'})
```

(I realize that this solution has potential drawbacks but, for my use case, they're not significant)

timhuff:
This is a solution to #1893

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -218,6 +218,23 @@ assign(Builder.prototype, {
       return this.where(1, '=', column ? 1 : 0)
     }
 
+    // Allow a Query Builder to be passed in
+    if (column instanceof Builder){
+      if (column._single.table){
+        throw new Error('You passed a query builder into appendWhere() that has a table ('+
+          column._single.table+') specified for it');
+      }
+      column._statements.map((function(_this){
+        return function(statement){
+          if(statement.grouping != 'where')
+            throw new Error('You passed a query builder into appendWhere() that had'+
+              ' a non-where clause on it.');
+          _this._statements.push(statement);
+        };
+      })(this));
+      return this;
+    }
+
     // Check if the column is a function, in which case it's
     // a where statement wrapped in parens.
     if (typeof column === 'function') {
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -229,7 +229,7 @@ assign(Builder.prototype, {
           if(statement.grouping != 'where')
             throw new Error('You passed a query builder into appendWhere() that had'+
               ' a non-where clause on it.');
-          _this._statements.push(statement);
+          _this._statements.push(clone(statement));
         };
       })(this));
       return this;
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -300,7 +300,7 @@ assign(Builder.prototype, {
   orWhere: function orWhere() {
     this._bool('or');
     const obj = arguments[0];
-    if(isObject(obj) && !isFunction(obj) && !(obj instanceof Raw)) {
+    if(isObject(obj) && !isFunction(obj) && !(obj instanceof Raw) && !(obj instanceof Builder)) {
       return this.whereWrapped(function() {
         for(const key in obj) {
           this.andWhere(key, obj[key]);
--- a/test/integration/builder/selects.js
+++ b/test/integration/builder/selects.js
@@ -578,7 +578,25 @@ module.exports = function(knex) {
             );
           });
       });
+    });
 
+    it('can have a where clause with complex query builder', function() {
+      var simpleWhereClause = knex.where('id', 4);
+
+      var complexWhereClause = knex
+        .where('logins', '>', 1)
+        .whereIn('id', function () {
+          this.select('id').from('accounts').whereNotNull('logins').where(simpleWhereClause)
+          .where(function () {
+            this.where(simpleWhereClause);
+          });
+        });
+
+      knex('accounts').where(simpleWhereClause).where(complexWhereClause).orWhere(simpleWhereClause)
+      .then(function(results){
+        expect(results[0].id == 4);
+        expect(results[0].logins == 2);
+      });
     });
 
     it('has a "distinct" clause', function() {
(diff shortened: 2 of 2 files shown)


test/integration/builder/selects.js
```diff
@@ -540,6 +540,45 @@ module.exports = function(knex) {
           });
       });
 
+      it.only('allows parameter to be a query builder', function() {
```
elhigu: `.only` should be removed and a bit more complex test cases with multiple conditions and nested subqueries should be added. 

Something like 

```
simpleWhereClause = knex.where('id', 1);

complexWhereClause  = knex
  .where('active', true)
  .whereIn('id', function () { 
    this.from('other').whereExist('imNotNull').where(simpleWhereBuilder).where(function () {
      this.where(simpleWhereBuilder);
    });
  });

knex('accounts').where(simpleWhereClause).where(complexWhereClause).orWhere(simpleWhereClause);
```

src/query/builder.js
```diff
@@ -218,6 +218,23 @@ assign(Builder.prototype, {
       return this.where(1, '=', column ? 1 : 0)
     }
 
+    // Allow a Query Builder to be passed in
+    if (column instanceof Builder){
+      if (column._single.table){
+        throw new Error('You passed a query builder into appendWhere() that has a table ('+
```
elhigu: This error case should have test case.

src/query/builder.js
```diff
@@ -218,6 +218,23 @@ assign(Builder.prototype, {
...
+          column._single.table+') specified for it');
+      }
+      column._statements.map((function(_this){
+        return function(statement){
+          if(statement.grouping != 'where')
+            throw new Error('You passed a query builder into appendWhere() that had'+
```
elhigu: Also a test case for this one is needed

timhuff:
I've gone ahead and made the changes you requested

test/integration/builder/selects.js
timhuff: Good call. Upon writing something similar, I found that this PR would crash on `.orWhere`. It's all fixed up now though.

timhuff:
I attempted to modify the documentation but when I ran `npm run dev`, I received the following error:
```
[webpack-dev-server]  /path/to/knex_documentation/scripts/webpack.config.js:2
[webpack-dev-server]  const {default: settings} = require('./webpack-settings.js')
[webpack-dev-server]        ^
[webpack-dev-server]
[webpack-dev-server]  SyntaxError: Unexpected token {
[webpack-dev-server]      at exports.runInThisContext (vm.js:53:16)
[webpack-dev-server]      at Module._compile (module.js:373:25)
[webpack-dev-server]      at Object.Module._extensions..js (module.js:416:10)
[webpack-dev-server]      at Module.load (module.js:343:32)
[webpack-dev-server]      at Function.Module._load (module.js:300:12)
[webpack-dev-server]      at Module.require (module.js:353:17)
[webpack-dev-server]      at require (internal/module.js:12:17)
[webpack-dev-server]      at module.exports (/path/to/knex_documentation/node_modules/webpack/bin/convert-argv.js:80:13)
[webpack-dev-server]      at Object.<anonymous> (/path/to/knex_documentation/node_modules/webpack-dev-server/bin/webpack-dev-server.js:68:48)
[webpack-dev-server]  at Module._compile (module.js:409:26)
```

test/integration/builder/selects.js
```diff
@@ -540,6 +540,85 @@ module.exports = function(knex) {
...
+      var complexWhereClause = knex
+        .where('logins', '>', 1)
+        .whereIn('id', function () {
+          this.select('id').from('accounts').whereNotNull('logins').where(simpleWhereClause)
+          .where(function () {
+            this.where(simpleWhereClause);
+          });
+        });
+
+      knex('accounts').where(simpleWhereClause).where(complexWhereClause).orWhere(simpleWhereClause)
```
elhigu: Could you please check the generated query from .toSQL() output... like wth tester function does? The query executed here is so insane that parts of it might not be correct but it still would create the same results.

elhigu:
I added one more review note to make sure that output of that more complex query is correct.

timhuff:
Thanks for the input. I'll make the modification.

elhigu:
Also I have no idea how documentation building works... I've always just sent pull request to documentation branch and left building for CI 😬

elhigu:
@timhuff I checked out how document building works... you don't have to worry about it. It is enough to just edit contents and I'll build the new version if necessary. There are scripts in package.json if you want to run documentation locally to see how your changes render.

timhuff:
@elhigu I appreciate that. Unfortunately, I'm very bogged down with work at the moment (and probably for a good time to come). I'll be unable to make the appropriate changes anytime in the near future. I expect to be working with knex for some time to come (really great work you guys have done here), so I'll be sure to get back to this so that my future self can enjoy this new functionality. Thanks.

elhigu:
Closing older unfinished PRs, can be reopened if finished.

PR #1896, opened 2017-02-03
More promise functions

timhuff:
This is #1894 only I created a branch that wasn't my master (that won't be changing) to submit it with.

### DIFF
--- a/src/interface.js
+++ b/src/interface.js
@@ -1,6 +1,6 @@
 
 import * as helpers from './helpers';
-import { isArray, map, clone, each } from 'lodash'
+import { isArray, map, clone, each, identity } from 'lodash'
 
 export default function(Target) {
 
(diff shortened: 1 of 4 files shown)


src/interface.js
```diff
@@ -64,11 +64,11 @@ export default function(Target) {
...
   each(['bind', 'catch', 'finally', 'asCallback',
     'spread', 'map', 'reduce', 'tap', 'thenReturn',
-    'return', 'yield', 'ensure', 'reflect'], function(method) {
+    'return', 'yield', 'ensure', 'reflect',
+    'get', 'mapSeries', 'delay'], function(method) {
     Target.prototype[method] = function() {
-      let then = this.then();
-      then = then[method].apply(then, arguments);
-      return then;
+      const promise = this.then(identity);
```
elhigu: Why identity must be passed here?

timhuff: It doesn't need to be. Before I realize that `.then` doesn't need a parameter, in order to use bluebird's `.get`, I've been using `.then(_.identity).get(0)`. I've pushed a commit to remove it.

elhigu:
Looks good now, thanks!

PR #1902, opened 2017-02-07
Add json support to the query builder for mysql

### LINKED ISSUE #1036: JSON Data Type support in MySQL v5.7.x and SQlite 3.9
Hi,

I was just wondering if the new [JSON data type](https://dev.mysql.com/doc/refman/5.7/en/json.html) present in MySQL v5.7.x is supported?

Thanks

baronfel:
This is a minimal amount of work to make the `json(COLUMN_NAME)` method on the query builder do the right thing for versions of MySQL greater than or equal to 5.7.8, when the json data type first came into being.

It follows most of the patterns put forth for the postgres json support, including retrieving the version from the target server and using that to determine support levels.

I've included tests patterned after the postgres tests which pass on my local mysql instance.

Anything else I should do?

Fixes #1036 for mysql only.

### DIFF
--- a/src/dialects/maria/schema/columncompiler.js
+++ b/src/dialects/maria/schema/columncompiler.js
@@ -1,0 +1,23 @@
+// MariaDB Column Compiler
+// -------
+import inherits from 'inherits';
+import ColumnCompiler_MySQL from '../../mysql/schema/columncompiler';
+
+import { assign } from 'lodash'
+
+function ColumnCompiler_MariaDB() {
+  ColumnCompiler_MySQL.apply(this, arguments);
+  this.modifiers = ['unsigned', 'nullable', 'defaultTo', 'first', 'after', 'comment', 'collate']
+}
+
+inherits(ColumnCompiler_MariaDB, ColumnCompiler_MySQL);
+
+assign(ColumnCompiler_MariaDB.prototype, {
+    // mariadb doesn't support json as of yet. 
+    // see https://jira.mariadb.org/browse/MDEV-9056 for issue status.
+  json () {
+    return 'text';
+  }
+});
+
+export default ColumnCompiler_MariaDB;
--- a/src/dialects/mysql/schema/columncompiler.js
+++ b/src/dialects/mysql/schema/columncompiler.js
@@ -121,8 +121,11 @@ assign(ColumnCompiler_MySQL.prototype, {
   
 })
 
+const MIN_JSON_VERSION = semver.SemVer('5.7.8');
 function jsonColumn(client) {
-  if(!client.version || semver.gte(client.version, '5.7.8')) {
+  if(!client.version) {
+    return 'json'; // this is here jsut for tests
+  } else if(semver.valid(client.version) && semver.gte(client.version, MIN_JSON_VERSION)) {
     return 'json';
   } else {
     return 'text';
--- a/src/dialects/mysql/schema/columncompiler.js
+++ b/src/dialects/mysql/schema/columncompiler.js
@@ -4,6 +4,7 @@
 import inherits from 'inherits';
 import ColumnCompiler from '../../../schema/columncompiler';
 import * as helpers from '../../../helpers';
+import * as semver from 'semver';
 
 import { assign } from 'lodash'
 
@@ -80,6 +81,10 @@ assign(ColumnCompiler_MySQL.prototype, {
     return length ? `varbinary(${this._num(length)})` : 'blob'
   },
 
+  json() {
+    return jsonColumn(this.client);
+  },
+
   // Modifiers
   // ------
 
@@ -113,7 +118,15 @@ assign(ColumnCompiler_MySQL.prototype, {
   collate(collation) {
     return collation && `collate '${collation}'`
   }
-
+  
 })
+
+function jsonColumn(client) {
+  if(!client.version || semver.gte(client.version, '5.7.8')) {
+    return 'json';
+  } else {
+    return 'text';
+  }
+}
 
 export default ColumnCompiler_MySQL;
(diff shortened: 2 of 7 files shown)


baronfel:
I don't know why the tests can't connect to mariadb now. Anyone have any pointers?

elhigu:
Looks like travis has too old mysql server, you need to change .travis.yml file somehow to get newer version installed
```
mysql --version
mysql  Ver 14.14 Distrib 5.5.53, for debian-linux-gnu (x86_64) using readline 6.2
```

src/dialects/maria/schema/columncompiler.js
```diff
@@ -0,0 +1,23 @@
...
+  this.modifiers = ['unsigned', 'nullable', 'defaultTo', 'first', 'after', 'comment', 'collate']
+}
+
+inherits(ColumnCompiler_MariaDB, ColumnCompiler_MySQL);
+
+assign(ColumnCompiler_MariaDB.prototype, {
+    // mariadb doesn't support json as of yet. 
+    // see https://jira.mariadb.org/browse/MDEV-9056 for issue status.
+  json () {
+    return 'text';
```
elhigu: I dont see any value of using text column instead of json, since you won'r be able to run any json queries for those columns anyways. I don't know what has been original reasoning for using json as text columns

baronfel: this is here because maria doesn't have the json data type, and so needs to override the inherited definition of `json` with `text` so that table creation statements don't blow up.

baronfel:
re: travis configuration, how did the mysql connection ever work, then? I didn't touch anything around the configuration at all, and yet what I see is a DB connection issue. Most mysterious.

src/dialects/mysql/schema/columncompiler.js
```diff
@@ -113,7 +118,18 @@ assign(ColumnCompiler_MySQL.prototype, {
...
   }
-
+  
 })
 
+const MIN_JSON_VERSION = semver.SemVer('5.7.8');
+function jsonColumn(client) {
+  if(!client.version) {
+    return 'json'; // this is here jsut for tests
+  } else if(semver.valid(client.version) && semver.gte(client.version, MIN_JSON_VERSION)) {
```
elhigu: I suppose this same check would work also for mariadb removing need for changes in its column compiler...

src/dialects/mysql/schema/columncompiler.js
```diff
@@ -113,7 +118,18 @@ assign(ColumnCompiler_MySQL.prototype, {
...
+    return 'json';
+  } else {
+    return 'text';
```
elhigu: I would just go with using type 'json' always if json column is built and leave it up to DB to throw error if it doesn't support it (IIRC it is how it is done in postgresql dialect).

elhigu: it breaks backwards compatibility though and should be mentioned in CHANGELOG

baronfel: That's not the case, see this snippet from postgres's columns compiler:
```js
function jsonColumn(client, jsonb) {
  if (!client.version || parseFloat(client.version) >= 9.2) return jsonb ? 'jsonb' : 'json';
  return 'text';
}
```

src/dialects/maria/schema/columncompiler.js
elhigu: Shouldn't it blow up if database is asked to create json column, but it doesn't support it?

baronfel: that would be a breaking change with what already exists, but I wouldn't be opposed to it.

elhigu: @wubzz @rhys-vdw @tgriesser any second opinions? This feature is already breaking change for people who has been using mysql >= 5.7 and .json() column type (probably not many or none).

baronfel:
@elhigu thanks for the review!  I'll wait a while for more comments to com in from those that you tagged and then work to incorporate the changes.

elhigu:
Closing and opening to make travis to trigger build again with fresh `.travis.yml` config... there is still not mysql 5.7, but I updated it to trusty with 5.6... maybe later this year travis updates to provide easier installing for 5.7 too.

johannessjoberg:
Any updates on this PR?

baronfel:
Let me summarize and see if I've got it all straight:

- We don't want to do the version-checking logic to determine which storage `json` should choose, instead we'll just always use `json` column type and let the database response blow up if the database doesn't support it
- Therefore, the mariadb changes don't need to happen and can be rolled back\
- Likewise, the version-checking changes in the core mysql driver can be rolled back
- Finally, a CHANGELOG entry should be added to note that mysql will use `json` columns when a column is defined using the `json` builder method, which is breaking for people currently using it and expecting `TEXT`

sourcec0de:
Hey everyone, thanks for your awesome work on knex.
I was hoping to get an update on the support for this feature in MySQL.
The new JSON features are great and it would be cool to be able to set a column as JSON in a migration.

elhigu:
Just to set reminder for me here to be easily found from end of thread. This PR is left open, because it is waiting for travis to support easier installation of mysql 5.7 (or us to write docker based service setup there).

igor-savin-ht:
@elhigu I presume https://github.com/travis-ci/travis-ci/issues/5122 takes care of the mysql version issue? If I create an updated PR against latest master, is this change still good to go?

elhigu:
@igor-savin-ht yes, this would be good to go if you are able to setup travis with mysql 5.7 :+1:

kibertoad:
@elhigu This is redundant now and can be closed.

Issue #1779, opened 2016-11-11
how to reset column selection for query?


helios1138:
IIRC, previously it was possible to call something like `q.select(null)` to reset previously added column names. 
Currently it seems that you can only append new columns to the selections but no way to clear/replace them

alechirsch:
+1

elhigu:
That has never been documented feature, at least I wouldn't like that clearing columns of select would be done by overloading the same call with magic value like `.select()` e.g. `.clearSelect()` would be a lot more robust API... 

But knex doesn't really support of taking out parts of query and I don't think it should go to that road right now since it will add bunch of more surface for errors (there is already very small amount of people maintaining the current code base).

Having possibility to clear only select won't add much value to library without having possibility to modify all parts of the query.

wubzz:
`.clearSelect()` exists, closing.

Issue #854, opened 2015-06-01
Select/Column overriding old columns


tuhoojabotti:
From reading the docs this seems to be the documented behavior, but updating from 0.7.3 to 0.8.6 the behavior changed and broke my tests (thankfully).

Before calling qb.select(col) appended the column, now it replaces the old ones. I couldn't find a way to append a column without overriding the old ones.

Is there a way to append columns?

tgriesser:
Would you be able to provide a simple test case to make sure I match the behavior you're expecting?

tuhoojabotti:
Here's an example:

```javascript
return Model.query(function (qb) {
	// This used to append column, not overwrite.
	// Could be changed to qb.columns(..., { append: true }) or something.
	qb.select(knex.raw('COUNT(*) OVER() AS total_count'));
	if (limit) qb.limit(limit);
	if (offset) qb.offset(offset);
}).fetchAll(options)
	.tap(function (m) {
		if (m.length) res.header('X-Total-Count', m.at(0).get('total_count'));
		else res.header('X-Total-Count', 0); // You went too far.
		m.invoke('set', 'total_count', undefined);
	});
```

Also if you have a better idea for counting all results before limit, that would be nice too. ;)

sjuvonen:
I cannot confirm this. The behaviour seems completely the opposite to me: calling select() will NOT overwrite anything; instead it will APPEND to the old list of columns up for select. According to the documentation, method column() should OVERWRITE the selected columns but it is actually only an alias to select(), hence it doesn't work. 

This is evident by looking at the code lib/query/builder.js#782, as the same method is set for column() as is used with select().

tuhoojabotti:
I think the original problem is fixed now. It was just one broken version after the big internal refactorings. I will close this issue now. @sjuvonen if you think this issue should still be open for discussion, let me know and I'll re-open it. I think that being able to clear selected columns is another issue altogether.

PR #1907, opened 2017-02-11
added clearSelect and clearWhere to query builder

alechirsch:
Solution to #1779, also related to #854

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -976,7 +976,7 @@ assign(Builder.prototype, {
 
   // Helper function for clearing or reseting a grouping type from the builder
   _clearGrouping(grouping){
-    this._statements = reject(this._statements, function(statement){
+    this._statements = reject(this._statements, (statement) => {
       return statement.grouping === grouping;
     });
   }
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -10,7 +10,7 @@ import * as helpers from '../helpers';
 import JoinClause from './joinclause';
 import {
   assign, clone, each, isBoolean, isEmpty, isFunction, isNumber, isObject,
-  isString, isUndefined, tail, toArray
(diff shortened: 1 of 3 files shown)


elhigu:
I hadn't checked this pull request when I made the comment about API naming in the other issue :D nice choice 👍

I'll let someone else to decide if this should be getting in. Code looks good, but I would like to see also tests for different nested subquery builders.

I agree that these 2 are the most common uses for clear variants, but there are much more parts in query builder that people might like to clear, like order by is one of my favorites, since in count queries for it can cause huge performance hit.

src/query/builder.js
```diff
@@ -960,8 +972,14 @@ assign(Builder.prototype, {
       aggregateDistinct: aggregateDistinct || false
     });
     return this;
-  }
+  },
 
+  // Helper function for clearing or reseting a grouping type from the builder
+  _clearGrouping(grouping){
+    this._statements = reject(this._statements, function(statement){
```
elhigu: Arrow functions are preferred in library code if using `function ()` is not necessary.

alechirsch: Loud and clear,  I also prefer store notation, however I saw function syntax in many places in the apiece, was not sure which was preferred.

alechirsch:
@elhigu I agree, we should have the capability to clear any grouping type. I wrote the helper function to support future clearing of different groupings. In fact I did implement it for all of the grouping types that I found in the api, but removed them for now since I did not have the time to write tests for all of them. 

I also was not sure if I caught all of the types, is there a documented list of the types?

src/query/builder.js
alechirsch: Arrow notation*, in the api*

alechirsch:
@elhigu I have added some tests for subqueries

alechirsch:
Closing with preference for #1912

PR #1912, opened 2017-02-14
added clearSelect and clearWhere to query builder

alechirsch:
Copy of #1907, this pull request has my account attached to the commits...

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -10,7 +10,7 @@ import * as helpers from '../helpers';
 import JoinClause from './joinclause';
 import {
   assign, clone, each, isBoolean, isEmpty, isFunction, isNumber, isObject,
-  isString, isUndefined, tail, toArray
+  isString, isUndefined, tail, toArray, reject
 } from 'lodash';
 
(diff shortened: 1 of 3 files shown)


elhigu:
If you have time to implement this for returning, limit, order by, join, group by, having, on,  would be awesome. Still it won't be super useful since one cannot get for example subquery builders from outside to cleanup parts of subquery (like removing order by from some subquery).

Anyways I'm inclining to accept this feature after all. Maybe we could add new section to documentation, which lists all clearing functions, so it would be clear which parts of query builder can be reset... 

Code looks look, only documentations seem to be missing.

alechirsch:
I have added documentation for the feature please let me know if I missed anything. Adding to the docs is slightly confusing.

[https://github.com/knex/documentation/pull/24](https://github.com/knex/documentation/pull/24)

alechirsch:
@elhigu I do not have the time right now to add the other clear methods, however it should be easy to do so with the private method _clearGrouping function that I added. Testing would be the most time consuming.

elhigu:
thanks!

PR #1926, opened 2017-02-20
Revert "Ensure that 'client' is provided in knex config object (#1822)"

venables:
This reverts **some** of commit a4bfee77513d000cf0fddf130a370071c240f93d (PR #1822).  It leaves the fixes to `this.modifiers` for the base schema columncompiler, the oracle columncompiler, and the sqlite3 columncompiler.

The pull request in question fixed a bug in Knex, but also removed a documented (and quite useful) feature of the Knex library, which is to be able to initialize a Knex instance without a client/dialect.

This change leaves the bugfix, and re-enables the ability to initialize Knex without a specific client.

From the documentation:

> You can even use knex without a connection, just for its query building features. Just pass in an empty object when initializing the library. Specify a client if you are interested in a particular flavour of SQL.
```javascript
var knex = require('knex')({});
var pg = require('knex')({client: 'pg'});
knex('table').insert({a: 'b'}).returning('*').toString();
// "insert into "table" ("a") values ('b')"

pg('table').insert({a: 'b'}).returning('*').toString();
// "insert into "table" ("a") values ('b') returning *"
```

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -37,14 +37,6 @@ function clientId() {
 // for a dialect specific client object.
 function Client(config = {}) {
   this.config = config
-
-  //Client is a required field, so throw error if it's not supplied.
-  //If 'this.dialect' is set, then this is a 'super()' call, in which case
-  //'client' does not have to be set as it's already assigned on the client prototype.
-  if(!this.config.client && !this.dialect) {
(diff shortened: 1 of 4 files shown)


venables:
Closing per discussion in #1822

PR #1935, opened 2017-02-23
Properly close PG query stream on error

caseywebdev:
Without calling `close()`, the `queryStream` will continue to execute a
query meaning there was no way to halt a long running query due to an
unrecoverable downstream error. I ran into this issue while streaming
IDs from a very large table as part of an ETL process. An error would
occur downstream in the transformation or loading step and I needed to
cancel the long-running extract stream, but there was no way to do so
and I'd eventually run out of connections in the pool.

This commit also handles the case when the non-callback version of
stream API was used and an exception was thrown. The issue was that the
promise was not reachable and therefore there was no way to add a
rejection handler. This resulted in an unavoidable `Unhandled Promise
Rejection` error.


tgriesser:
Awesome, thanks Casey! Will look to get this out soon

caseywebdev:
👍 you da man. I love this lib 😉

PR #1959, opened 2017-03-10
Fix the bug caused by other database version string

RockyNiu:
Fix the bug caused by other database version string, for example 'Vertica Analytic Database v7.2.1-0'


elhigu:
CI fails because of linter.

Also have you checked if there is discussion in issues about this. I'm not sure if this is the correct fix for the issue.

RockyNiu:
@elhigu I created a issue here https://github.com/tgriesser/knex/issues/1978.
What is your suggestion for the correct fix?

It seems knex doesn't need the version number for serious things.
I have tested this fix in my local and it worked.

wubzz:
I personally think the regex is a bit twisted.. For example in my case `PostgreSQL 9.4.5, compiled by Visual C++ build 1800, 64-bit` yields `9.4.5,` with the comma. I'd have expected `9.4.5`.

I think the suitable solution would be to only pull the version numbers, don't bother matching on `PostgreSQL` at all.

Something like

```javascript
_.first(resp.rows[0].version.match(/[0-9(\.*?)]+/));
```

Which would yield `9.4.5` in my case, `7.2.1` in your case.

Perhaps other scenarios to cover as well, but this should be a more general solution. What do you think?

RockyNiu:
@wubzz Good thoughts! Here is what I think. I assume the default meaning of version is PostgreSQL version, since the file sits in dialects/postgres. But in my case `7.2.1` doesn't mean PostgreSQL 7.2.1, and there may be some difference between `PostgreSQL 7.2.1` and `Vertica Analytic Database v7.2.1-0`. In addition, there might be other companies, besides Vertica, to modify postgre to build their own database, and introduce different version string. To make this function `checkVersion` more generic and meaningful, I suggest to only show serial number of PostgreSQL, but show whole string in other kinds of database.

Another way to resolve this problem is to show whole string from any database. How do you think?

elhigu:
I've been thinking why knex actually needs to parse / validate version string of the connected DB.  Looks like some functionality of schema building depends on it, for example:

```js
function jsonColumn(client, jsonb) {
  if (!client.version || parseFloat(client.version) >= 9.2) return jsonb ? 'jsonb' : 'json';
  return 'text';
}
```

And that is pretty much all I could find.

So I think that instead of trying to parse non standard version strings from forked or similar databases, more general way would be to just be able to give the database version in knex configuration. 

In that way if I make my on mysql 5.7 fork and call it "Awesome DB 1.1337" and connect with knex to it I could say:

```
{
  client: 'mysql',
  version: '5.7'
}
 ```

And knex will consider connected database to be mysql 5.7 compatible. 

It would be simpler than creating perfect regex for every DB fork in the world and I can't think any drawbacks if that info is provided in configuration when one uses non standard DB.

**Comment by RockyNiu** [comment]
@elhigu it seems only dialects/postgre using `jsonColumn` or `client.version`. So I added ```if (config.version) {
  }``` in Client_PG(), and reversed the two commits in checkVersion().

**Comment by elhigu** [comment]
Implementation looks good to me, it still needs test case and documentation.

**Comment by RockyNiu** [comment]
@elhigu will add tests and document it soon. I have no access to our Vertica database, but I will install Communities Edition in a VM locally.

**Comment by elhigu** [comment]
@RockyNiu If you give version in configuration it is not even requested from the database server (after this pull request). 

So knex is not asserting that database returns the correct version, but configuration just overrides the version and tells knex directly that you should consider this database to be postgres 9.5 compatible (in case if you pass version: '9.5' in configuration).

File: test/unit/schema/postgres.js [hunk 1]
```diff
@@ -5,10 +5,18 @@
...
 
 var PG_Client = require('../../../lib/dialects/postgres');
-var client    = new PG_Client({})
+var client    = new PG_Client({});
 
 var equal  = require('assert').equal;
 
+describe("PostgreSQL Config", function(){
+  it('check version', function(){
+    var version = 'v7.2.1-0';
```
elhigu: this should be e.g `8.5` and you can verify that it works by checking what kind of query is generated for request:
 
```
knex.schema.table(t => { 
  t.json('colName');
}).toSQL();
```

RockyNiu:
@elhigu what I wanted to say is that if you set config {version: '9.5'}, and the real database version is '7.0', then it will pass this unit test. However if there is a integration test, it may fail: because the version '9,5' will be used in `jsonColumn` and `jsonColumn` will return `jsonb`, instead of `json`.

test/unit/schema/postgres.js
RockyNiu: Test cases updated. But there were always 3 out of 5 test instances failed. Do you know why?

RockyNiu:
```batchInsert
1) #757 - knex.batchInsert(tableName, bulk, chunkSize)```

elhigu: @RockyNiu I think @wubzz added that test while ago... maybe it random fails... it shouldn't be related to this PR

RockyNiu: @wubzz could you help to look at this? ^

wubzz: @elhigu @RockyNiu Can't reproduce the problem locally, so not sure what's wrong with it. The assertion error itself is bogus. The assertion takes place *after* rows have been inserted, so there should be no way for the table to contain 0 rows, since the promise chain was successful... Seems to just be mariasql too?

wubzz: Actually totally found a timing issue, try pulling latest master.

RockyNiu: @wubzz I did have latest master. what is the sha you have?

RockyNiu: @wubzz no worry. My bad.

RockyNiu: 1) # 1880 - Duplicate keys in batchInsert should not throw unhandled exception

elhigu:
@RockyNiu I believe that if you now rebase your branch to top of master, it should pass all checks.

elhigu:
I cleaned up this PR in #1993 thanks for the contribution @RockyNiu!

Issue #1950, opened 2017-03-07
Concurrent migrations fail instead of waiting


itaysabato:
This is in continuum to a comment I made [here](https://github.com/tgriesser/knex/issues/1217#issuecomment-284764072).

In the [docs](http://knexjs.org/#Notes-about-locks) it says:

> A lock system is there to prevent multiple processes from running the same migration batch in the same time. When a batch of migrations is about to be run, the migration system first tries to get a lock using a SELECT ... FOR UPDATE statement (preventing race conditions from happening). If it can get a lock, the migration batch will run. If it can't, it will wait until the lock is released.

But, in actuality, it is impossible for one process to wait while another is migrating. This can be seen by running two processes at the same time that start by calling `knex.migrate.latest()` - one will succeed and the other will fail with a lock error.

From what I could tell, this is caused by the code for [`_getLock()`](https://github.com/tgriesser/knex/blob/master/src/migrate/index.js#L174) which starts a transaction, selects the migration lock `FOR UPDATE`, sets it to true and commits the transaction before the calling code moves on to perform the migration. This way, any other process that tries to get the lock will fail.

If the transaction will end only after the whole migration, this should not happen.

These are knex logs for running two such processes against PostgreSQL:
```
  knex:client acquired connection from pool: __knexUid1 +1s
  knex:query select * from information_schema.tables where table_name = ? and table_schema = current_schema +3ms
  knex:bindings [ 'knex_migrations' ] +1ms
  knex:pool INFO pool postgresql:pg:client0 - dispense() clients=0 available=1 +32ms
  knex:client acquired connection from pool: __knexUid1 +1s
  knex:query select * from information_schema.tables where table_name = ? and table_schema = current_schema +2ms
  knex:bindings [ 'knex_migrations' ] +1ms
  knex:pool INFO pool postgresql:pg:client0 - dispense() clients=0 available=1 +4ms
  knex:client releasing connection to pool: __knexUid1 +300ms
  knex:pool INFO pool postgresql:pg:client0 - dispense() clients=0 available=2 +1ms
  knex:pool INFO pool postgresql:pg:client0 - dispense() clients=1 available=2 +1ms
  knex:client acquired connection from pool: __knexUid2 +0ms
  knex:query select * from information_schema.tables where table_name = ? and table_schema = current_schema +1ms
  knex:bindings [ 'knex_migrations_lock' ] +0ms
[...]

PR #1962, opened 2017-03-12
Migration transaction

itaysabato:
A suggested fix for #1950.

If no transactions are disabled (neither in the general migration configuration nor in any individual migration file) then the whole migration to the latest version (including the acquisition of the lock and its release) is done in the same transaction which either completely succeeds or completely fails.

Among other benefits, this makes migrating to latest in parallel possible (e.g. in the bootstrap of multiple servers) since another process will automatically wait while another is migrating instead of failing with a lock error and crashing.

I had to fix some existing tests to adjust to this new behavior and I've also added a couple of new ones which test parallel migrations.

Please review :)
Thanks.


elhigu:
Note to myself for the release. This is breaking change.... and thanks @itaysabato  I'll try to review this soon

test/integration/migrate/index.js
```diff
@@ -111,28 +85,27 @@ module.exports = function(knex) {
...
-            return knex('knex_migrations')
-              .where('id', migration1[0])
-              .orWhere('id', migration2[0])
-              .orWhere('id', migration3[0])
-              .del()
+          .spread(function(migration1, migration2, migration3) {
+            return knex.migrate.status({directory: 'test/integration/migrate/test'}).then(function(migrationLevel) {
+              expect(migrationLevel).to.equal(3);
+            })
+              .then(function() {
```
elhigu: here indentation is also drifting

test/integration/migrate/index.js
```diff
@@ -268,12 +221,52 @@ module.exports = function(knex) {
...
+      return knex.migrate.rollback({directory: 'test/integration/migrate/test'});
+    });
+
+    if (knex.client.dialect === 'postgres') {
+      it("is able to run two migrations in parallel in postgres", function () {
+        return Promise.all([
+          knex.migrate.latest({directory: 'test/integration/migrate/test'}),
+          knex.migrate.latest({directory: 'test/integration/migrate/test'})
+        ])
+          .then(function () {
```
elhigu: indentation drifting...

test/integration/migrate/index.js
```diff
@@ -55,41 +54,16 @@ module.exports = function(knex) {
       });
 
       it('should return 0 if code matches DB', function() {
-        // Put in a couple dummy migrations. Needed
-        // since the migrations directory has a couple
-        // bad migrations that don't get loaded. This
-        // will simulate the DB and source being in sync.
-        return Promise.all([
-          knex('knex_migrations').returning('id').insert({
```
elhigu: Lots of code was removed here, is test still checking everything fine (sorry I'm a bit tired and don't understand why code was ok to be removed)?

itaysabato: This code was removed since it added mock data that is no longer required - this migration directory no longer has invalid migrations, they were moved to `test_with_invalid`.

test/integration/migrate/index.js
itaysabato: Right, WebStorm does that sometimes, should I fix it?

test/integration/migrate/index.js
```diff
@@ -268,12 +221,52 @@ module.exports = function(knex) {
...
   });
 
+  describe('knex.migrate.latest in parallel', function() {
+
+    afterEach(function() {
+      return knex.migrate.rollback({directory: 'test/integration/migrate/test'});
+    });
+
+    if (knex.client.dialect === 'postgres') {
+      it("is able to run two migrations in parallel in postgres", function () {
```
itaysabato: The default transaction modes in the other databases either do auto-commit (e.g. MySQL) or don't support `for share` locking (e.g. SQLite). Therefore (by default) concurrent migrations are expected to fail in the same way as before for the non-Postgres databases.

test/integration/migrate/index.js
itaysabato: WebStorm again, fixable.

test/integration/migrate/test_with_invalid/20131019235306_migration_2.js
```diff
@@ -0,0 +1,21 @@
...
+
+
+exports.up = function(knex, promise) {
+  return promise.all([
+    knex.schema
+      .createTable('migration_test_2', function(t) {
+        t.increments();
+        t.string('name');
+      }),
+      knex.schema
```
itaysabato: This whole file was actually copied as-is with the same indentation.

test/integration/migrate/index.js
elhigu: Umm... I don't understand why auto-commit would be a problem here? If each migration is runnings  are done in their own transactions, auto-commit is automatically disabled in mysql? I'm a bit worried about the fact that this functionality ended up working only for postgresql.

itaysabato: From what I learned (and tested) auto-commit is enabled by default in MySQL. If the knex user will change this in his database then I expect it will behave the same as Postgres.

Otherwise, it means that one transaction will see that the other transaction has locked and it will cause it to fail (instead of waiting for it to finish). This is consistent with the current behavior of knex migrations.

itaysabato: Reference: https://dev.mysql.com/doc/refman/5.7/en/innodb-locking-reads.html

> Locking of rows for update using SELECT FOR UPDATE only applies when autocommit is disabled (either by beginning transaction with START TRANSACTION or by setting autocommit to 0. If autocommit is enabled, the rows matching the specification are not locked.

elhigu: @itaysabato As I said in earlier comment and is said in the mysql specs linked: In our case auto-commit is disabled because we are selecting for update inside transaction.

When other transaction sees that row is already locked and tries to select it for update the DB should just wait until lock is freed instead of throwing an errors (the same way that postgresql does).

Since I still don't understand/believe that this won't work in mysql, I need to read more about that and I'll run some tests locally to see how it works and if it really throws an error or wait if two transactions are trying to select for update the same row.

Also another thing is oracle support, which should be also doable... I would like to get this test enabled to every dialect except sqlite. I'll follow up later on when I have some data how stuff really works after playing in mysql shell.

itaysabato: OK great, let me know if there's anything I can do!

elhigu: I just tested this with mysql and it should work the same way that postgres does. So there is no reason why this test wouldn't work for every other DB (except sqlite).

mysql-selectforupdate

itaysabato: Sorry, you're right. I guess I got confused with the whole auto-commit thing.

However, the reason concurrent migrations usually **cannot succeed** in MySQL, MariaDB and Oracle is because DDL statements (e.g. `CREATE TABLE` that is used in the test) cause an implicit commit to the current transaction. See https://dev.mysql.com/doc/refman/5.7/en/implicit-commit.html.

This is what makes the test fail: once the first table in the first migration is created, the transaction is implicitly committed and the awaiting `select for update` statement returns with a value of `1` (i.e. locked) - which causes knex to throw. I've turned on knex query logging where it is seen clearly:
```
 knex:query BEGIN; +0ms
 knex:bindings undefined +0ms
 knex:query BEGIN; +1ms
 knex:bindings undefined +0ms
 knex:query select * from `knex_migrations_lock` for update +2ms
 knex:bindings [] +0ms
 knex:query select * from `knex_migrations_lock` for update +2ms
 knex:bindings [] +0ms
 knex:query update `knex_migrations_lock` set `is_locked` = ? +1ms
 knex:bindings [ 1 ] +0ms
 knex:query show tables like ? +2ms
 knex:bindings [ 'knex_migrations' ] +0ms
 knex:query show tables like ? +1ms
 knex:bindings [ 'knex_migrations_lock' ] +0ms
 knex:query select * from `knex_migrations_lock` +2ms
 knex:bindings [] +0ms
 knex:query select `name` from `knex_migrations` order by `id` asc +1ms
 knex:bindings [] +0ms
 knex:query select max(`batch`) as `max_batch` from `knex_migrations` +2ms
 knex:bindings [] +0ms
 knex:query create table `migration_test_1` (`id` int unsigned not null auto_increment primary key, `name` varchar(255)) default character set utf8 +2ms
 knex:bindings [] +0ms
 Knex:warning - Can't take lock to run migrations: Migration table is already locked
 Knex:warning - If you are sure migrations are not running you can release the lock manually by deleting all the rows from migrations lock table: knex_migrations_lock
 knex:query ROLLBACK; +8ms
```

Unlike the others, in Postgres most DDL statements [do not cause implicit commits](http://stackoverflow.com/a/4736346/2295972), therefore the awaiting transaction will continue only after the lock is released.

I am using Postgres and for my use case this PR would help a great deal, since it is the cleanest way to avoid collisions during concurrent migrations, so I think it is worth it.

elhigu: Fair enough. Looks like only postgres and mssql are not doing implicit commits and oracle + mysql are https://massimotinelli.wordpress.com/2016/06/23/implicit-commit-ddl-oracle-vs-ms-sql-server/. Would be good to document that difference of functionality of different DB engines it is not obvious for many of the users.

elhigu: We are not running integration tests yet for mssql on CI but eventually we will. So for this pull request would be enough to add also `mssql` to be tested in addition to `postgres`.

itaysabato: You mean I should include `mssql` in the `if` condition, though it would have no effect on the current CI test, yes?

elhigu: yes please.

itaysabato: Done. Merge? :)

elhigu:
@itaysabato awesome and thanks !

elhigu:
I think I will release 0.13 pretty soon.

AlSherif:
@elhigu @itaysabato I'm running migrations on a replicated postgres DB system and the migration will fail quite often (on the first migration only) between the creation of the knex lock table and the select on that table, due to both requests going to different DB nodes.  
`     return this._migrationData()
        .tap(validateMigrationList)
        .spread((all, completed) => {
 -        return this._runBatch(difference(all, completed), 'up');
 +        const migrations = difference(all, completed);
 +
 +        const transactionForAll = !this.config.disableTransactions
 +          && isEmpty(filter(migrations, name => {
 +            const migration = require(path.join(this._absoluteConfigDir(), name));
 +            return !this._useTransaction(migration);
 +          }));
 +
 +        if (transactionForAll) {
 +          return this.knex.transaction(trx => this._runBatch(migrations, 'up', trx));
 +        }
 +        else {
 +          return this._runBatch(migrations, 'up');
 +        }`
The check, creation and repeated check for the knex tables are all happening in '_migrationData()' more specifically in '_listCompleted(trx = this.knex)'. This does in our case break a migration that runs over multiple databases (synchronously).
I hacked a fix doing this: 
`  _listCompleted(trx = this.knex) {
    const { tableName } = this.config
    return trx.transaction((rlTrx) => {
      return this._ensureTable(rlTrx)
        .then(() => rlTrx.from(tableName).orderBy('id').select('name'))
        .then((migrations) => map(migrations, 'name'))
    })
  }`
Maybe one of you guys has a better solution?

itaysabato:
@AlSherif An example of how your migration files and config look like might help to understand what is wrong. What exception/error message are you getting?

Did you try disabling transactions?

elhigu:
If error happens before migration_lock table even exist so that 2 different instances are trying to create it at the same time, failing makes sense. I would suggest that you initialize your DB so that migrations / lock table are there from the beginning, before even starting the app that runs migrations.

elhigu:
Another way around the problem might be to create advisory lock before running migrations so that only one transaction tries to create the initial tables https://www.linkedin.com/pulse/postgresql-locking-revealed-petar-partlov

With advisory locks you should be able to create lock that is not related to any table, so it doesn't matter that there is no tables / rows for creating that kind of lock.

AlSherif:
The error that I'm getting is a failed select on the 'knex_migration_lock' table, 'relationship does not exist'.
As I said, I am pretty sure that this is caused due to the 'create knex_migration_lock' and 'select knex_migration_lock' calls going to different nodes in the DB replication pool. Those calls are not in a transaction. I' am not using any asyncronous calls, I just iterate over a list of databases and execute the migration.
This is not a critical problem for us since the migration only has to run once, afterwards the knex tables exist.
Are there any possible issues with my fix? If not, I would tend to with that for the sake of speed and simplicity. I'll check out the advisory locks though.

elhigu:
@AlSherif ah ok. What kind of replication setup are you using? If its postgres native streaming replication is it in async or sync mode? If problem is the one you mentioned advisory locks wont help, since select still can go to node which is out of sync... 

Do you have multiple application instances running migrations at the same time (I just assumed you were since this change fixed that behaviour to block instead of failing)?

elhigu:
If you are running just on application process trying to run migrations at the same time, then you might be able to make all the queries to go to the same connection  by creating Migration instance and passing trx object to it as a constructor argument. Trx object is quite compatible with base knex instance, but it won't use connection pool, but sends all the queries to the same connection. 

https://github.com/tgriesser/knex/blob/master/src/migrate/index.js

If you have multiple processes running migrations at the same time and your replication setup is not synchronous I have no idea how to make it work :)

AlSherif:
@elhigu Yes its only one application process, I just posted it here because I assumed that this fix will run all migration related calls in a transaction, which it doesn't.
I actually use a knex instance and call the migration on it, so it should be possible to use a transaction instance instead. Thanks, that should work then :)

itaysabato:
@elhigu Perhaps I am mistaken, but I think the migrator will have to be rewritten for that to work because it often uses syntax that is not supported on a transaction instance. Specifically, the case of calling `.transaction` on the transaction instance should probably be avoided.

elhigu:
@itaysabato creating nested transactions (transaction inside transactions) is implemented with save points and is supported by knex. But yeah, I don't remember from top of my head, which parts of knex API cannot be used through trx object. There might be some.

AlSherif:
@itaysabato @elhigu Seems to run well so far, calling migrations on a transaction object.

PR #1970, opened 2017-03-14
Transaction should not reject with undefined.

wubzz:


src/transaction.js
```diff
@@ -121,8 +121,15 @@ export default class Transaction extends EventEmitter {
         debug('%s error running transaction query', this.txid)
       })
       .tap(() => {
-        if (status === 1) this._resolver(value)
-        if (status === 2) this._rejecter(value)
+        if (status === 1) {
+          this._resolver(value)
+        }
+        if (status === 2) {
+          if(isUndefined(value)) {
```
joepie91: Is the intention to *only* do this when rejecting with `undefined`? I'm not sure under what circumstances this issue could occur (as I haven't followed the issue closely), but other non-Error rejection values such as numbers or strings would have the same "missing stacktrace" issue.

wubzz: Yes, this is intentional. By definition `undefined` is the only javascript type that can be considered non-existant. Null, Strings, Numbers, Objects, Errors, Buffers, or any other type of value is still valid.

For example some people prefer to use strings (why is beyond me):
````javascript
try {
    throw 'poop'
} catch(e) {
    console.log(e, typeof e);
    //poop string
}
```

So in cases where rollback is called with an explicit value other than `undefined`, that value will be used in the reject handler.

PR #1971, opened 2017-03-14
Upgrade generic-pool to v3.1.7 (Test)

wubzz:
Testing compatibility of upgrading `generic-pool` to 3.1.7. Contains some breaking changes that would require probably 0.12.X -> 0.13.0.


wubzz:
Ah nvm, knex still supporting 0.12 so this won't be possible until support is dropped. https://travis-ci.org/tgriesser/knex/jobs/211110006

PR #1974, opened 2017-03-15
Must always clear the timeout for handling 'acquireConnectionTimeout'…

### LINKED ISSUE #1973: Database connection is not correctly destroyed after knex 0.12.2
Hey 👋

I have recently updated knex from `0.12.2` to `0.12.7` and noticed that the database connection is not closed when calling `connection.destroy`. The result is, that a process doesn't exit.

Here is a snippet:
```
....
this.connection = knex(...);

return this.connection.transaction(function(txn) {
   return doSomething(txn)
       .finally(function() {
           return self.connection.destroy()
                .then(function () {
                    debug('Destroyed connection');
                });
       });
})
```
NOTE: debug('Destroyed connection') get's called.

**I have tested all knex versions and it starts in `0.12.3`.**

The code works for me with `0.12.2`, but does not with `0.12.3`.
Is there anything i can change in my code to get it working with > `0.12.2`?

Otherwise, I went through the changes, maybe related to the pool update https://github.com/tgriesser/knex/pull/1702?

Thank you so much!
Kate

wubzz:
…, even if an error occurs. Fixes #1973

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -239,10 +239,10 @@ assign(Client.prototype, {
         ))
       }, this.config.acquireConnectionTimeout || 60000)
       this.pool.acquire((err, connection) => {
+        clearTimeout(t)
         if (err) {
           return rejecter(err)
         }
-        clearTimeout(t)
         if (wasRejected) {
           this.pool.release(connection)
         } else {


kirrg001:
Thanks!

raijinsetsu:
What version can we expect this fix in? It doesn't look like this project is using milestones.

raijinsetsu:
Ignore my last comment. I found the releases page. Looks like 0.12.8.
Thanks

PR #1987, opened 2017-03-23
Allow PostgreSQL connections to EnterpriseDB

qjroberts:
This PR allows the `pg` client to connect to an EnterpriseDB instance. I've been using a local modification for this for a couple months, and haven't encountered any issues yet. It's been brought up before (#1176), but didn't receive any follow-up.

- [x] Tests - Tried to replicate existing tests. Happy to make suggested changes, as necessary!
- [x] Coding Style - Couldn't find any specific documentation. I can switch the `var` to `const` (`prefer-const`) in tests, but kept it for backwards-compatibility with older Node versions.
- [x] PR on master
- [x] Documentation - No updates, unless EnterpriseDB should be added.
- [x] Passing Tests - Everything, with the exception of `batchInsert` currently failing on `master`. Verified this PR does not affect those tests.

I couldn't figure out any patterns in the tests, so wasn't sure if additional tests are needed. Adding another database (EnterpriseDB) to the testing list seemed excessive, so I've only added a few unit tests to cover the specific functionality.

### DIFF
--- a/src/dialects/postgres/index.js
+++ b/src/dialects/postgres/index.js
@@ -132,7 +132,7 @@ assign(Client_PG.prototype, {
     return new Promise(function(resolver, rejecter) {
       connection.query('select version();', function(err, resp) {
         if (err) return rejecter(err);
-        resolver(/^PostgreSQL (.*?)( |$)/.exec(resp.rows[0].version)[1]);
+        resolver(/^(?:PostgreSQL|EnterpriseDB) (.*?)( |$)/.exec(resp.rows[0].version)[1]);
       });
     });
   },
--- a/test/index.js
+++ b/test/index.js
@@ -39,6 +39,13 @@ if (config.oracle) {
   });
 }
 
+if (config.postgres) {
+  describe('Postgres Tests', function() {
+    this.timeout(process.env.KNEX_TEST_TIMEOUT || 5000);
+    require('./unit/dialects/postgres');
+  });
+}
+
 describe('Integration Tests', function() {
   this.timeout(process.env.KNEX_TEST_TIMEOUT || 5000);
   require('./integration')
--- a/test/unit/dialects/postgres.js
+++ b/test/unit/dialects/postgres.js
@@ -1,0 +1,71 @@
+/* global it, describe */
+
+'use strict';
+var expect = require('chai').expect;
+var knex   = require('../../../knex');
+var knexInstance = knex({
+  client: 'pg',
+  connection: {
+    connectString: 'connect-string',
+    database: 'database',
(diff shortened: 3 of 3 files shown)


elhigu:
@qjroberts Does EnterpriseDB has same version numbers that normal postgres so that EnterpriseDB 9.4 for example is based on the same features that postgres 9.4?

qjroberts:
It looks that way from their documentation, and from the versions I've used. The do seem to add an additional point release (for example, `9.3.13` vs `9.3.13.37` in the PR tests). I forgot to check to see how that might affect the version check elsewhere, but it looks like the only spot is in `src/dialects/postgres/schema/columncompiler.js`, which already uses `parseFloat` to grab the major/minor version. I just tested it, and it looks like `parseFloat` will work the same with the 4-part release numbers as it does with the 3-part ones.

It does look like the tests are failing on the older versions of Node, compared to `master`, so I'll have to take a look at those tomorrow!

wubzz:
@qjroberts Try pulling latest master, timing issue should be fixed, sorry.

qjroberts:
Thanks! I tried looking into it yesterday, but didn't have any luck. It looks like it still exists though, I'm getting the failure locally, both on current `master` and my (rebased) fork.

I'll keep poking around the changes to `batchInsert` to see if I can dig up anything else.

elhigu:
@qjroberts also this PR is somewhat related to this https://github.com/tgriesser/knex/pull/1959#discussion_r107582373

Maybe parsed version info could be limited to first x.y or x.y.z. That way parsed version should be compatible and still provide enough information for knex.

qjroberts:
That might be a better solution. I like what #1959 does as well, but the current code looks like it would just ignore the version check if one was specified. I'm happy to close this one if that's the direction you'd like to go.

The regex solution itself seems a bit weird, especially if there are more Postgres variations that modify that version string. I can take a look if you think it's worth trying to modify the regex into something a bit more generic, in addition to #1959.

elhigu:
@qjroberts I don't believe that we should modify that regex to be much more generic. 

I was fine with this addition though since the EnterpriseDB versions follows to real postgres version and is closely related to the open source postgres project.

So I don't have anything against this, but I believe that 100% of our current user base can be served with the solution in #1959.

One hypothetical usecase for this could be project, which supports some json columns and would like to still support postgres / enterprisedb < 9.2

I leave it up to you if you like to get this to master or to close it.

qjroberts:
Since #1959 seems to satisfy the requirements for this, I'll close this for now. If I find anything that needs an additional solution, I'll open another PR.

Thanks!

PR #1992, opened 2017-03-25
Bugfixes on batchInsert and transactions for mysql/maria

wubzz:
I really don't like the code duplication between the base `Transaction.query` implementation and the ones for `mysql`, `mysql2`, and `maria`. For the sake of "keeping dialects separate", I left it as is. Perhaps a better idea just to listen on the `query-error` event and give the `helpers.warn` a call there instead?

### DIFF
--- a/src/util/batchInsert.js
+++ b/src/util/batchInsert.js
@@ -35,10 +35,11 @@ export default function batchInsert(client, tableName, batch, chunkSize = 1000) 
       .then((tr) => {
         return Promise.mapSeries(chunks, (items) => tr(tableName).insert(items, returning))
           .then((result) => {
-            result = flatten(result);
+            result = flatten(result || []);
 
             if(autoTransaction) {
-              return tr.commit()
+              //TODO: -- Oracle tr.commit() does not return a 'thenable' !? Ugly hack for now.
+              return (tr.commit(result) || Promise.resolve())
                 .then(() => result);
             }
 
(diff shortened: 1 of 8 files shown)


src/util/batchInsert.js
```diff
@@ -35,21 +35,23 @@ export default function batchInsert(client, tableName, batch, chunkSize = 1000)
       .then((tr) => {
         return Promise.mapSeries(chunks, (items) => tr(tableName).insert(items, returning))
           .then((result) => {
-            result = flatten(result);
+            result = flatten(result || []);
 
             if(autoTransaction) {
-              return tr.commit()
+              //TODO: -- Oracle tr.commit() does not return a 'thenable' !? Ugly hack for now.
+              return (tr.commit(result) || Promise.resolve())
```
wubzz: Not too familiar with Oracles code base so not sure. The current state came as a shock to me. If `commit` does not return a promise, how can a standard transaction work at all for oracle?

```javascript
knex.transaction((tr) => tr.raw('select version'))
.then(() => {...})
.catch((error) => {})
```

Really strange and confusing. Currently have no environment for testing oracle.

elhigu: I'm planning to setup docker container with postgres, mysql, oracle and mssql maybe with that it will be easier to test locally all dialects (hopefully I have time to do it before 2018 :p and permission from microsoft to bundle mssql there...).

PR #1998, opened 2017-04-04
Add fetchAsString optional parameter to oracledb dialect

atiertant:
add fetchAsString parameter (see https://github.com/oracle/node-oracledb/blob/master/doc/api.md#propdbfetchasstring)

```
var knex = require('knex')({
  client: 'oracledb',
  connection: {
    host : '127.0.0.1',
    user : 'your_database_user',
    database : 'myapp_test'
  }
  fetchAsString: ['DATE', 'NUMBER', 'CLOB']
});
```


elhigu:
Looks fine except that it should validate input strings and include test and documentation.

test/unit/dialects/oracledb.js
```diff
@@ -39,4 +36,56 @@ describe("OracleDb externalAuth", function() {
...
+    knexClient.schema.createTable('fetchAsStringTable', function (table) {
+      table.increments();
+      table.float('testfloat');
+      table.date('testdate');
+    }).then(function() {
+      knexClient('fetchAsStringTable').insert({
+        testfloat: 7.32,
+        testdate: new Date(2017, 5, 7)
+      }).then(function() {
+        done();
```
elhigu: It would be better to just return promise from function instead of calling done in various places.

test/unit/dialects/oracledb.js
```diff
@@ -39,4 +36,56 @@ describe("OracleDb externalAuth", function() {
...
+      }).catch(done);
+    }).catch(done);
+  });
+
+  it('fetchAsString parameter should return strings on selected types', function(done) {
+    knexClient('fetchAsStringTable').select().then(function(result) {
+      expect(result[0]).to.be.ok;
+      expect(result[0].testfloat).to.be.a('string');
+      expect(result[0].testdate).to.be.a('string');
+      knexClient.destroy(done);
```
elhigu: destroying client here makes all of these tests dependent to each other... sadly thats pretty much how all tests of knex are written.

elhigu:
Generally tests and docs looks good. Only thing is that I don't like using `done()` callback in tests without a good reason, they are more error prone than just returning a promise for test hook.

test/unit/dialects/oracledb.js
atiertant: could make two describe creating client in before and destroy in after.

elhigu: that would be more clean

test/unit/dialects/oracledb.js
```diff
@@ -39,4 +36,72 @@ describe("OracleDb externalAuth", function() {
...
+    after(function() {
+      return knexClient.destroy();
+    });
+
+  });
+
+  describe("without fetchAsString parameter", function() {
+    var knexClient;
+
+    before(function(done) {
```
elhigu: not necessary done parameter

test/unit/dialects/oracledb.js
```diff
@@ -39,4 +36,72 @@ describe("OracleDb externalAuth", function() {
...
+
+});
+
+describe("OracleDb parameters", function() {
+  this.timeout(10000);
+
+  describe("with fetchAsString parameter", function() {
+    var knexClient;
+
+    before(function(done) {
```
elhigu: not necesssary done()

test/unit/dialects/oracledb.js
```diff
@@ -39,4 +36,72 @@ describe("OracleDb externalAuth", function() {
...
   });
+
+  after(function() {
+    sinon.restore(knexInstance.client.driver.getConnection);
+  });
+
+});
+
+describe("OracleDb parameters", function() {
+  this.timeout(10000);
```
elhigu: why this timeout is here?

elhigu:
Sorry for taking ages with this. I found still couple of minor issues from the code. And probably that timeout should be just removed or use some value that is relative to `process.env.KNEX_TEST_TIMEOUT`.

atiertant:
@elhigu timeout was usefull on local only and we removed unnecessary done parameter.

elhigu:
Now it looks perfect, thanks!

elhigu:
I'll try to roll out knex 0.13 during this weekend

atiertant:
@elhigu that's a good news ! thanks

PR #2003, opened 2017-04-10
fix: escapeObject parameter order for Postgres dialect.

santegoeds:
Found this bug when I added support for writing `moment` instances (which in turn I did to fix timezone issues I had when writing `DATE` types).

```javascript
const moment = require('moment')

moment.fn.toPostgres = function () {
  return this.format()
}
```

### DIFF
--- a/src/dialects/postgres/index.js
+++ b/src/dialects/postgres/index.js
@@ -74,7 +74,7 @@ assign(Client_PG.prototype, {
       }
       return escaped
     },
-    escapeObject(val, timezone, prepareValue, seen = []) {
+    escapeObject(val, prepareValue, timezone, seen = []) {
       if (val && typeof val.toPostgres === 'function') {
         seen = seen || [];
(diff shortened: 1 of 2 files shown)


elhigu:
thanks, this needs a test case too.

santegoeds:
Can you suggest a suitable location where I can add a test case?

santegoeds:
I found a location for a test @elhigu.  Please let me know if you'd like me to squash the commits.

elhigu:
Thanks! Fix looks ok, it hides worst bugs from that part of the code. 

I'm pretty sure that actually timezone + seen parameters should be completely removed and replaced with ctx object and then some more changed should be probably done to support seen and timezone parameters as they were intented...

I'm merging this one for now and add an issue about that function...

hulbert:
@elhigu is this `toPostgres` functionality documented anywhere?

elhigu:
@hulbert no and it should not be. It is not part of public API.

PR #2011, opened 2017-04-14
Add option to .increments() to disable use as primary key

### LINKED ISSUE #385: Cannot create multiple PKs (mixin .increments with .primary)
I want to create a table basing on following MySQL dump:

```
CREATE TABLE `my_table` (
  `cmdId` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `deviceId` char(16) NOT NULL,
  `fnNumber` int(10) unsigned DEFAULT NULL,
  `chNumber` int(10) unsigned DEFAULT NULL,
  `cmd` varchar(50) DEFAULT NULL,
  `cmdDate` datetime DEFAULT NULL,
  `delivered` tinyint(1) DEFAULT NULL,
  PRIMARY KEY (`cmdId`,`deviceId`),
  KEY `deviceId` (`deviceId`),
  CONSTRAINT `tblcommands_ibfk_1` FOREIGN KEY (`deviceId`) REFERENCES `second_table` (`deviceId`) ON DELETE NO ACTION ON UPDATE NO ACTION
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

My knex looks like:

```
knex.schema.createTable('my_table', function(t) {
    t.primary(['cmdId', 'deviceId']);
    t.increments('cmdId');
    t.string('deviceId', 16).notNullable().references('second_table.deviceId');
    t.integer('fnNumber').unsigned();
    t.integer('chNumber').unsigned();
    t.string('cmd96', 50);
    t.dateTime('cmdDate');
    t.boolean('delivered');
});
```

However there is a problem with primary keys. I get following error:

```
Error: Multiple primary key defined
```

I assume that `increments` method already creates PK and `primary` method creates another one which causes an error.

Is it possible to achieve this format of table ?

davidfurlong:
Usage: 

`t.increments('id', { primaryKey: false })` => Not primary key
`t.increments('id')` => Primary key as currently

I had some trouble running the `maria` tests

Fixes #385

### DIFF
--- a/src/dialects/mssql/schema/columncompiler.js
+++ b/src/dialects/mssql/schema/columncompiler.js
@@ -18,7 +18,9 @@ inherits(ColumnCompiler_MSSQL, ColumnCompiler);
 
 assign(ColumnCompiler_MSSQL.prototype, {
 
-  increments: 'int identity(1,1) not null primary key',
+  increments(withoutPk){
+    return withoutPk ? 'int identity(1,1) not null' : 'int identity(1,1) not null primary key'
+  },
 
   bigincrements: 'bigint identity(1,1) not null primary key',
 
--- a/src/dialects/mssql/schema/columncompiler.js
+++ b/src/dialects/mssql/schema/columncompiler.js
@@ -18,11 +18,17 @@ inherits(ColumnCompiler_MSSQL, ColumnCompiler);
 
 assign(ColumnCompiler_MSSQL.prototype, {
 
-  increments(withoutPk){
-    return withoutPk ? 'int identity(1,1) not null' : 'int identity(1,1) not null primary key'
+  increments({ primaryKey = true } = {}){
+    return primaryKey ?
+      'int identity(1,1) not null primary key' :
+      'int identity(1,1) not null';
   },
 
-  bigincrements: 'bigint identity(1,1) not null primary key',
+  bigincrements({ primaryKey = true } = {}){
+    return primaryKey ?
+      'bigint identity(1,1) not null primary key' :
+      'bigint identity(1,1) not null';
+  },
 
   bigint: 'bigint',
 
--- a/src/dialects/oracle/schema/columncompiler.js
+++ b/src/dialects/oracle/schema/columncompiler.js
@@ -26,9 +26,9 @@ assign(ColumnCompiler_Oracle.prototype, {
     });
   },
 
-  increments () {
+  increments (withoutPk) {
     this._createAutoIncrementTriggerAndSequence();
-    return 'integer not null primary key';
+    return withoutPk ? 'integer not null' : 'integer not null primary key';
   },
 
   bigincrements () {
--- a/src/dialects/oracle/schema/columncompiler.js
+++ b/src/dialects/oracle/schema/columncompiler.js
@@ -26,14 +26,14 @@ assign(ColumnCompiler_Oracle.prototype, {
     });
   },
 
-  increments (withoutPk) {
+  increments ({ primaryKey = true } = {}) {
     this._createAutoIncrementTriggerAndSequence();
-    return withoutPk ? 'integer not null' : 'integer not null primary key';
+    return primaryKey ? 'integer not null primary key' : 'integer not null';
   },
 
-  bigincrements () {
+  bigincrements ({ primaryKey = true } = {}) {
     this._createAutoIncrementTriggerAndSequence();
-    return 'number(20, 0) not null primary key';
+    return primaryKey ? 'number(20, 0) not null primary key' : 'number(20, 0) not null';
   },
 
   floating(precision) {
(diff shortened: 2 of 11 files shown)


elhigu:
Actually... ignore most of that :) At first I saw only one changed file :o I'll check this out better later.

davidfurlong:
yeah its my understanding that the docs are in a separate repo. I will address failing CI checks

src/dialects/mssql/schema/columncompiler.js
```diff
@@ -18,7 +18,9 @@ inherits(ColumnCompiler_MSSQL, ColumnCompiler);
 
 assign(ColumnCompiler_MSSQL.prototype, {
 
-  increments: 'int identity(1,1) not null primary key',
+  increments(withoutPk){
+    return withoutPk ? 'int identity(1,1) not null' : 'int identity(1,1) not null primary key'
+  },
 
   bigincrements: 'bigint identity(1,1) not null primary key',
```
elhigu: probably implementation of big increments should also take new parameter in and implementations should be refactored to share one implementation.

elhigu:
Yup, historically boolean params has been used, but I haven't approved lately any new ones. Documentation is in another repo so link to docuementation pull request is needed.

davidfurlong:
@elhigu Made the changes 👍

src/dialects/mssql/schema/columncompiler.js
```diff
@@ -18,9 +18,17 @@ inherits(ColumnCompiler_MSSQL, ColumnCompiler);
 
 assign(ColumnCompiler_MSSQL.prototype, {
 
-  increments: 'int identity(1,1) not null primary key',
+  increments({ primaryKey = true } = {}){
+    return primaryKey ?
+      'int identity(1,1) not null primary key' :
+      'int identity(1,1) not null';
```
elhigu: does this really create auto incrementing column?

src/dialects/oracle/schema/columncompiler.js
```diff
@@ -26,14 +26,14 @@ assign(ColumnCompiler_Oracle.prototype, {
     });
   },
 
-  increments () {
+  increments ({ primaryKey = true } = {}) {
     this._createAutoIncrementTriggerAndSequence();
-    return 'integer not null primary key';
+    return primaryKey ? 'integer not null primary key' : 'integer not null';
```
elhigu: this most probably doesnt create auto incrementing column but just integer that is not null

src/dialects/mssql/schema/columncompiler.js
davidfurlong: you've got a point.

elhigu:
Thanks! I spotted some bugs, when I gave it another look though.

src/dialects/mssql/schema/columncompiler.js
davidfurlong: But yes it does https://www.w3schools.com/sql/sql_autoincrement.asp

src/dialects/oracle/schema/columncompiler.js
davidfurlong: You're right about this one

davidfurlong: I'm afraid I'm not quite sure where to start on this one, as I need to create a separate sequence using a namespace and reference it here.

https://www.w3schools.com/sql/sql_autoincrement.asp

davidfurlong:
> Thanks! I spotted some bugs, when I gave it another look though.

Should I fix them? I can try to write some integration tests as well

elhigu:
Yes they should be fixed. You can add here tests which creates a new table with incrementing id column (primary key) and another incrementing columns which is not primary key. Insert two rows there and check that they were incrementing correctly. After test has ran you can drop the test table.

https://github.com/tgriesser/knex/blob/master/test/integration/schema/index.js

elhigu:
This can be reopened if it is completed.

davidfurlong:
I'll try to take another look this week

gaastonsr:
I love this. Looking forward to it.

IMHO this should have been the default from the beginning and be more explicit about our primary keys.

treble-snake:
I found myself in need of this option too, would be very nice to have it.

kibertoad:
Going to give this one a shot.

gpetrov:
@kibertoad I think this should be revisited,  because you can't switch from or to incremental types as well.

Use cases
* going from int primary (no auto inc)  -> increments (with auto inc) 
* going from increments -> bigIncrements

those vases give error as the primary already exists.

So imho the best option is to check if the primary on this field already exists and then do not create it, or add an option as done in this PR

kibertoad:
@gpetrov Help with rebasing would be appreciated :)

OlivierCavadenti:
@gpetrov As I understand, this PR is superseded by https://github.com/knex/knex/pull/4266 ? Or it's another case ?

gpetrov:
@OlivierCavadenti yes indeed, so this can be closed

PR #2020, opened 2017-04-19
Optimization. Quit early if migrations are empty.

vlasenko:


elhigu:
should fix the tests or will be closed in a week or so.

PR #2049, opened 2017-05-06
When no table name specified for Oracle, query builder will use "dual"

erayhanoglu:
knex.select(knex.raw('1')) generetes this sql > "select 1"
This code will work for many dialects except oracle. Oracle uses "dual" table for querying current_date, current_timestamp etc. After this small change, this code works with oracle too.


elhigu:
not working

PR #2055, opened 2017-05-10
Add support for enum columns to SQlite3 dialect

carragom:
Adds support for enum columns in SQLite using a check constraint. Based on the current PostgreSQL implementation. If I missed anything please let me know.


elhigu:
Integration test for enum's check constraint would be great to have to actually verify that created constraints function as expected. Otherwise this looks good.

elhigu:
And thanks :)

carragom:
Ok sorry about that, tests are passing now. Does it look good to you or is there anything else I missed?

elhigu:
This is good. Just need to remember that this is actually breaking change.

Issue #2051, opened 2017-05-09
Named constraints


carragom:
Hi there,

According to the [docs](http://knexjs.org/#Schema-primary), it possible to name you `primary key` and hopefully your `unique` and `foreign key`. But this does not seem to be true at least for **Sqlite**. Is this just lacking on the SQLite implementation?. I could work on a PR to allow this in SQLite.

elhigu:
It should be supported. If the sqlite implementation is missing, PR would be really welcome!

carragom:
Hi,

I'm working on a PR for this but would like to clarify something first. What should be the correct API for naming keys and indexes?. This is how I think the API works

# Indexes
## Fluent API
Only single column indexes can be created like this and the index name is optional. The next example will add the column `colA` to the table and an index named `optional-name`
```
table.string('colA').index('optional-name');
```
## Normal API
### Single column
Single column indexes can be created and the name is optional like the next example that will add an index named `optional-name` with one column `colA`
```
table.index('colA', 'optional-name')
```
### Multiple columns
Multi-column indexes are allowed and the name is optional but if multiple columns are needed **they must be specified in an array as the first parameter** like the next example that will add an index named `optional-name` with two columns `colA` and `colB`
```
table.index(['colA', 'colB'], 'optional-name')
```
Assuming the previous is correct, I think I can do the same for `primary` and `unique` indexes in SQLite. Waiting for your comments to proceed with the PR.

PR #2056, opened 2017-05-11
Add support for named unique, primary and foreign keys to SQLite3

carragom:
Attempt to solve #2051. Allows having named `unique`, `primary` and `foreign keys`. Both fluent and normal API should work except for `foreign keys` which have no fluent counterpart on any dialect AFAIK. This could break things for some user that might have been misusing the API. From my take on [this](https://github.com/tgriesser/knex/blob/0.13.0/test/unit/schema/sqlite3.js#L128-L135) test, in the current version of the SQLite dialect, the following code
```
table.primary('foo', 'bar')
```
Creates a primary key with two columns `foo` and `bar` when it should have created a primary key named `bar` with one column `foo` instead, as seen in other dialects like [MySQL](https://github.com/tgriesser/knex/blob/0.13.0/test/unit/schema/mysql.js#L176-L183).

Let's see if Travis likes it.

### DIFF
--- a/test/integration/schema/index.js
+++ b/test/integration/schema/index.js
@@ -630,5 +630,111 @@ module.exports = function(knex) {
       });
     });
 
+    it('supports named primary keys', function() {
+      var constraintName = 'pk-test';
+      var tableName = 'namedpk';
+      return knex.transaction(function(tr) {
+        return tr.schema.dropTableIfExists(tableName)
+          .then(function() {
+            return tr.schema.createTable(tableName, function(table) {
+              table.string('test').primary(constraintName);
+            })
+          })
+          .then(function() {
+            return tr.schema.dropTableIfExists(tableName);
+          })
+          .then(function() {
+            return tr.schema.createTable(tableName, function(table) {
+              table.string('test');
+              table.string('test2');
+            })
+          })
+          .then(function() {
+            return tr.schema.table(tableName, function(table) {
+              table.primary('test', constraintName)
+            })
+          })
+          .then(function() {
+            return tr.schema.dropTableIfExists(tableName);
+          })
+          .then(function() {
+            return tr.schema.createTable(tableName, function(table) {
+              table.string('test');
+              table.string('test2');
+            })
+          })
+          .then(function() {
+            return tr.schema.table(tableName, function(table) {
+              table.primary(['test', 'test2'], constraintName)
+            })
+          });
+      });
+    });
+
+    it('supports named unique keys', function() {
+      var singleUniqueName = 'uk-single';
+      var multiUniqueName = 'uk-multi';
+      var tableName = 'nameduk';
+      return knex.transaction(function(tr) {
+        return tr.schema.dropTableIfExists(tableName)
+          .then(function() {
+            return tr.schema.createTable(tableName, function(table) {
+              table.string('test').unique(singleUniqueName);
+            })
+          })
+          .then(function() {
+            return tr.schema.dropTableIfExists(tableName);
+          })
+          .then(function() {
+            return tr.schema.createTable(tableName, function(table) {
+              table.string('test');
+              table.string('test2');
+            })
+          })
+          .then(function() {
+            return tr.schema.table(tableName, function(table) {
+              table.unique('test', singleUniqueName);
+              table.unique(['test', 'test2'], multiUniqueName);
+            })
+          });
+      });
+    });
+
+    it('supports named foreign keys', function() {
+      var userTableName = 'nfk_user';
+      var groupTableName = 'nfk_group';
+      var joinTableName = 'nfk_user_group';
+      return knex.transaction(function(tr) {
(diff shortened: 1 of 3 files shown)


elhigu:
Looks good to me.

elhigu:
Thing that has worried me lately with many pull requests that tests are only covering generated queries, but they usually doesn't tests that those queries are actually doing the thing that they try to do correctly.

I would always like to have some integration test for this kind of changes, which actually tries the feature against real database. 

Now I'm worried that changed queries doesn't actually do what they are supposed to do and in worst case they break the code of many end users. 

I'll add requirement of integration tests against real DB to the contributing.md guide, since its better to have library working stable and well than having new features which are not tested well.

carragom:
Well, I'm not sure an integration test would have prevented this specific problem. The integration test expected result would have been wrong just like the expected generated query is wrong but in the end, the integration test would have passed just like the unit test because the wrongly expected result is also valid. I think to prevent this from happening unit and integration test should be required, yes. But also these tests need to be validated extensively to make sure the expectation of the test is correct. Also, having a standard list of tests all with the same name would be really good. Currently, tests with the same intent have slightly different names on most dialects.

elhigu:
> Well, I'm not sure an integration test would have prevented this specific problem. 

That comment of requirement of having integration test, had nothing to do with the test failure of this pull request :) Lack of them is just a thing that has been making feel uneasy already for a long time and that kind of tests makes sure that intended behaviour is not broken in the future by accident.

I would prefer leaving current integration test as is and just add a new simpler case which supports also sqlite. In that case also those alter column method tests will remain intact.

carragom:
Ok, here we go with the integration tests. Let me know what you think.

carragom:
Any idea about why is that test failing?. I see no relation between my changes and that specific test.

elhigu:
@carragom it seems to be one annoying random failure that happens every now and then... I'll run the test again.

test/integration/schema/index.js
```diff
@@ -576,5 +576,111 @@ module.exports = function(knex) {
...
+              table.string('test2');
+            })
+          })
+          .then(function() {
+            return tr.schema.table(tableName, function(table) {
+              table.primary('test', constraintName)
+            })
+          })
+          .then(function() {
+            return tr.schema.dropTableIfExists(tableName);
```
elhigu: Here you should try to drop constraint with raw query or do something else to make sure that given name was used for creating the named constraint.

test/integration/schema/index.js
```diff
@@ -576,5 +576,111 @@ module.exports = function(knex) {
...
+              table.string('test');
+              table.string('test2');
+            })
+          })
+          .then(function() {
+            return tr.schema.table(tableName, function(table) {
+              table.primary(['test', 'test2'], constraintName)
+            })
+          });
+      });
```
elhigu: Here the same thing to check that constraint with name was created.

test/integration/schema/index.js
```diff
@@ -576,5 +576,111 @@ module.exports = function(knex) {
...
+              table.string('test2');
+            })
+          })
+          .then(function() {
+            return tr.schema.table(tableName, function(table) {
+              table.unique('test', singleUniqueName);
+              table.unique(['test', 'test2'], multiUniqueName);
+            })
+          });
+      });
```
elhigu: The same kind of check needed

elhigu: maybe creating other constraint with the same name could also work for testing...

test/integration/schema/index.js
```diff
@@ -576,5 +576,111 @@ module.exports = function(knex) {
...
+          .then(function() {
+            return tr.schema.createTable(joinTableName, function(table) {
+              table.uuid('user').references('id').inTable(userTableName)
+                .withKeyName(['fk', joinTableName, userTableName].join('-'));
+              table.uuid('group');
+              table.primary(['user', 'group']);
+              table.foreign('group', ['fk', joinTableName, groupTableName].join('-'))
+                .references('id').inTable(groupTableName);
+            });
+          });
```
elhigu: and here too.

carragom:
Primary and unique tests should be good to go. Let me know what you think. I'll work on foreign keys sometime next week.

carragom:
Forgot a console.log in there, I'll remove it when I fix the foreign key tests.

carragom:
This PR is ready. I will rebase my branch to master but I'll like to wait for a response to my comment on #2087 first.

elhigu:
@carragom I think, that comment has been answered. Are you still going to finish this?

carragom:
@elhigu Your are right, this is ready to go, I should have time to finish this in the next couple of weeks.

kibertoad:
@carragom Do you need any help with this?

carragom:
I'm sorry for the delay on this. Short on time. @kibertoad if you can take over that would be great.

Issue #1946, opened 2017-03-05
.comment().after() in wrong order for mysql in generated SQL


jeff-kilbride:
I am trying to alter a table and add a column with a comment after an existing column. So:
```
knex.schema.table('myTable', function(t) {
    t.string('new_col_1').notNullable().after('existing_col');
    t.decimal('new_col_2', 3, 2).notNullable().defaultTo(0).comment('In dollars / cents').after('new_col_1');
}
```
However, I am getting an error like the following:
```
Knex:warning - migrations failed with error: alter table `myTable` add `new_col_1` varchar(255) not
null after `existing_col`, add `new_col_2` decimal(3, 2) not null default '0' after `new_col_1`
comment 'In dollars / cents' - ER_PARSE_ERROR: You have an error in your SQL syntax; check the
manual that corresponds to your MySQL server version for the right syntax to use near 'comment 'In
dollars / cents' at line 1
```
Notice, in the generated SQL, the `comment` clause is coming last even though I specified it before the `.after()` method in the chain. This is a problem, because the `comment` clause is part of the column definition, whereas the `after` clause is part of the `alter table` syntax. From the `mysql` docs:
```
alter_specification:
    table_options
  | ADD [COLUMN] col_name column_definition
        [FIRST | AFTER col_name ]
```
So, `after` needs to be the last thing specified in the `add` statement. Just to make sure, I did the following in the `mysql` CLI and it worked:
```
alter table `myTable`
add `new_col_1` varchar(255) not null after `existing_col`,
add `new_col_2` decimal(3, 2) not null default '0' comment 'In dollars / cents' after `new_col_1`
```
I am using `knex 0.12.7`:
```
> knex "--version"

Knex CLI version:  0.12.7
Local Knex version:  0.12.7
```

why2pac:
Solved, #2062

PR #2062, opened 2017-05-13
fixed table alter attributes order. (first, after)

why2pac:
Fixed issue, #1946

Should first and after attributes are placed after comment attribute.


why2pac:
@elhigu I added 2 test cases. At after position with comment and at first position with comment.

elhigu:
Thanks for adding unit tests, but there should be still integration test which uses real DB. Checking only which kind of query is generated is not enough.

https://github.com/tgriesser/knex/blob/master/CONTRIBUTING.md

why2pac:
@elhigu I added real database test cases.

elhigu:
@why2pac awesome, thanks!

why2pac:
@elhigu Thanks!

Issue #1567, opened 2016-07-08
sqlite3 - non existent columns in select or where not throwing error


esalter:
This might be two issues.  Using sqlite3, if I select from a non existent column, or add a non existent column in the where clause, no error is thrown.  The query is being generated correctly.  If I run this same query using mysql then an error is correctly thrown.  Therein lies the use case here - I'd like to use sqlite3 for unit testing failure conditions, but since the behavior is different from the real db I cannot do this currently.

Of note, if the table doesn't exist, that DOES throw an error.  So it's just limited to selecting/filtering by a non existent column.

If this is just a setting flag somewhere I missed, that's great, please just let me know what I need to flip!

````node
          return knex.select(['NonExistentColumn'])
            .from('clients')
            .where('nonexistant', 1)
            .on('query-error', function(error, obj) {
                console.log(error); // never reached
            }).on('query-response', function(response, obj, builder) {
                console.log(obj); // some data including results = []
            })
            .then(function(results) {
                console.log(results); // []
            }).catch(function(err) {
                console.log(err); // never reached
            });
````

elhigu:
There is no way to make sqlite to behave exactly like mysql or postgres. 

Unless there is really good reason using different DB for testing and real application I wouldn't do it. It just limits the feature set of the real DB that you can be using to features compatible with test database and you actually won't know if your code works against real database. 

Since this is not really knex related question, but about generally DB engines, I'm closing this now. Please leave more comments / re-open this issue if this is somehow knex related issue.

esalter:
Thanks for your response, however I do think this is a knex issue of some kind, because if I run the same statement against the underlying driver then it correctly throws an error.  Complete example:


````node
var sqlite3 = require("sqlite3").verbose();
var file = 'error_test.sqlite';

var db = new sqlite3.Database(file);
db.serialize(function() {

  db.run("CREATE TABLE Stuff (thing TEXT)");
  db.each("SELECT * FROM Stuff where nonexist = 1", function(err, row) {
    console.log(err); // { [Error: SQLITE_ERROR: no such column: nonexist] errno: 1, code: 'SQLITE_ERROR' }
    console.log(row); // undefined
  });
});
db.close();

````

I acknowledge the futility of asking for completely uniform behavior across all databases, but the fact remains that the driver does seem to be throwing an error that knex is dropping somewhere along the way (at least, I assume that's what's happening).

[...]

Issue #1048, opened 2015-11-05
Incorrect delimiters used on column names with sqlite3.


rhys-vdw:
I'm not sure of the standards for each DBMS, but this behaviour is surely wrong. Initially reported in tgriesser/bookshelf#948.

``` js
sqlite3('table').select('hello').toString();
// select "hello" from "table"
```

``` sql
create table t(id integer primary key, name text);
insert into t (name) values ('Rhys');

select * from t;
-- 1|Rhys

select "name", "hello" from t;
-- Rhys|hello

select `name`, `hello` from t;
-- Error: no such column: hello
```

The backtick delimiters differentiate between string literal (which is a valid SELECT argument) and column name.

chrisbroome:
So the issue is that sqlite is ambiguous in its column/string delimiters? If at all possible, double quotes should be used for column names and single quotes for string literals (standard behavior).

irsl:
Any news on this? Knex should wrap field names into backticks. The current behaviour is especially bad with where conditions :(

```
console.log(knex("sessions").select("s_id").where({s_id: 123}).toString())
=>
select "s_id" from "sessions" where "s_id" = 123
```

elhigu:
@irsl could you also add example how correct result should be for that query?

irsl:
The problem is the above query will never match any rows, since the constant string "s_id" is never the same as 123.

Column names should be wrapped into backticks both in columnize and in the various where filters too. The above query should look like this:

```
select `s_id` from `sessions` where `s_id` = 123
```

yasir095:
Is this fixed. really need a fix on it. As mysql doesn't support double quoted table names.

elhigu:
@yasir095 With mysql backticks are used:

```js
knex = require('knex')({client: 'mysql'});
console.log(knex("sessions").select("s_id").where({s_id: 123}).toString())
// select `s_id` from `sessions` where `s_id` = 123
```

yasir095:
awesome thanks

Kryten0807:
I'm still seeing this behaviour with sqlite3 version 3.1.8 and knex version 0.12.6:

The query was created with:

    this.database.select('group_id').from('users_to_groups').where('user_id', '=', 1)

The debug output is

    { method: 'select',
      options: {},
      timeout: false,
      cancelOnTimeout: false,
      bindings: [ 1 ],
      __knexQueryUid: '365a3dc6-d107-4abb-9c7f-1ef1dba621db',
[...]

PR #2063, opened 2017-05-14
Field identifiers were not quoted properly in Sqlite3 dialect

irsl:
https://github.com/tgriesser/knex/issues/1048
https://github.com/tgriesser/knex/issues/1567

### DIFF
--- a/src/dialects/sqlite3/index.js
+++ b/src/dialects/sqlite3/index.js
@@ -57,6 +57,10 @@ assign(Client_SQLite3.prototype, {
     return new SQLite3_DDL(this, compiler, pragma, connection)
   },
 
+  wrapIdentifier: function wrapIdentifier(value) {
+    return value !== '*' ? '`' + value.replace(/`/g, '``') + '`' : '*';
+  },
+
   // Get a raw connection from the database, returning a promise with the connection object.
   acquireRawConnection() {
     return new Promise((resolve, reject) => {
--- a/src/dialects/sqlite3/index.js
+++ b/src/dialects/sqlite3/index.js
@@ -57,8 +57,8 @@ assign(Client_SQLite3.prototype, {
     return new SQLite3_DDL(this, compiler, pragma, connection)
   },
 
-  wrapIdentifier: function wrapIdentifier(value) {
-    return value !== '*' ? '`' + value.replace(/`/g, '``') + '`' : '*';
+  wrapIdentifier(value) {
+    return (value !== '*' ? `\`${value.replace(/`/g, '``')}\`` : '*')
   },
 
   // Get a raw connection from the database, returning a promise with the connection object.
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -28,12 +28,25 @@ var clientsWithNullAsDefault = {
   mssql: new MSSQL_Client(useNullAsDefaultConfig),
 }
 
-function qb() {
+function qb_sqlite() {
   return clients.sqlite3.queryBuilder()
 }
-
-function raw(sql, bindings) {
+function raw_sqlite(sql, bindings) {
   return clients.sqlite3.raw(sql, bindings)
+}
+
+function callQueryBuilderCallback(selectedClients, clientType, callback){
+    var client = selectedClients[clientType]
+	
+	return callback(queryBuilderFactory, raw)
+	
+	function queryBuilderFactory(){
+	  return client.queryBuilder()
+	}
+	
+	function raw(sql, bindings) {
+	  return client.raw(sql, bindings)
+	}  
 }
 
 function verifySqlResult(dialect, expectedObj, sqlObj) {
@@ -51,12 +64,11 @@ function verifySqlResult(dialect, expectedObj, sqlObj) {
   });
 }
 
-function testsql(chain, valuesToCheck, selectedClients) {
+function testsql(queryBuilderCallback, valuesToCheck, selectedClients) {
   selectedClients = selectedClients || clients;
   Object.keys(valuesToCheck).forEach(function(key) {
-    var newChain = chain.clone()
-        newChain.client = selectedClients[key]
-    var sqlAndBindings = newChain.toSQL()
+    
+    var sqlAndBindings = callQueryBuilderCallback(selectedClients, key, queryBuilderCallback).toSQL()
 
     var checkValue = valuesToCheck[key]
     if (typeof checkValue === 'string') {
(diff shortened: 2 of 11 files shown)


src/dialects/sqlite3/index.js
```diff
@@ -57,6 +57,10 @@ assign(Client_SQLite3.prototype, {
     return new SQLite3_DDL(this, compiler, pragma, connection)
   },
 
+  wrapIdentifier: function wrapIdentifier(value) {
```
elhigu: the way mysql does it nowadays is:

```js
  wrapIdentifier(value) {
    return (value !== '*' ? `\`${value.replace(/`/g, '``')}\`` : '*')
  },
```
did you copy-paste compiled code? Copy-pasting just fine in this case though :)

irsl:
Yeah, the PR was premature, clearly. In the meanwhile I did the changes you requested, but overriding wrapIdentifier in the Sqlite3 dialect seem to break the others, at least thats what the test harness thinks. I believe this is somehow related to the fact the chain is built with an Sqlite3 client and then it is being cloned. This is the first failing test: 

```
  1) Query Building Tests QueryBuilder allows wrap on raw to wrap in parens and alias:

      oracle: AssertionError: expected 'select "e"."lastname", "e"."salary", (select `avg(salary)` from `employee` where dept_no = e.dept_no) avg_sal_dept from "employee" "e" where "dept_no" = ?' to deeply equal 'select "e"."lastname", "e"."salary", (select "avg(salary)" from "employee" where dept_no = e.dept_no) avg_sal_dept from "employee" "e" where "dept_no" = ?'
      + expected - actual

      -select "e"."lastname", "e"."salary", (select `avg(salary)` from `employee` where dept_no = e.dept_no) avg_sal_dept from "employee" "e" where "dept_no" = ?
      +select "e"."lastname", "e"."salary", (select "avg(salary)" from "employee" where dept_no = e.dept_no) avg_sal_dept from "employee" "e" where "dept_no" = ?
```

As you can see, fields in the subquery were wrapped with backticks, so I believe the chain of the subquery still refers the Sqlite3 client for what the chain was originally constructed for. If I'm right here, this probably also means some queries are verified incorrectly?...

Also, does the `DB=sqlite3 npm test` feature work for you? This seems to be discarded for me and still the complete test set is executed.

Cheers!

irsl:
@elhigu Please re-review.

elhigu:
Install node 0.12 (for example with nvm) and try to run tests locally. You'll find lots of errors.

irsl:
It's all good now on 0.12 as well. Would you please resolve the conflict and merge? I believe this is an important PR as it addresses a *bug* that does break Knex with Sqlite...

test/unit/query/builder.js
```diff
@@ -2,21 +2,31 @@
...
-var PG_Client = require('../../../lib/dialects/postgres')
-var Oracle_Client = require('../../../lib/dialects/oracle')
-var Oracledb_Client = require('../../../lib/dialects/oracledb')
-var SQLite3_Client = require('../../../lib/dialects/sqlite3')
-var MSSQL_Client = require('../../../lib/dialects/mssql')
+var _typeof2 = require('babel-runtime/helpers/typeof');
+
+var _typeof3 = _interopRequireDefault(_typeof2);
+
+var _keys = require('babel-runtime/core-js/object/keys');
```
elhigu: I'm feeling that this is not good way to fix the problems of the PR. Why just not use plain valid ES5 as it has been done for now?

test/unit/query/builder.js
```diff
@@ -25,97 +35,105 @@ var clientsWithNullAsDefault = {
...
+  }
+
+  function raw(sql, bindings) {
+    return client.raw(sql, bindings);
+  }
 }
 
 function verifySqlResult(dialect, expectedObj, sqlObj) {
-  Object.keys(expectedObj).forEach(function (key) {
+  (0, _keys2.default)(expectedObj).forEach(function (key) {
```
elhigu: This shouldn't have been changed and none of the similar changes in the rest of the file either.

test/unit/query/builder.js
```diff
@@ -2,21 +2,31 @@
...
 
 var clients = {
   mysql: new MySQL_Client({}),
   postgres: new PG_Client({}),
   oracle: new Oracle_Client({}),
   oracledb: new Oracledb_Client({}),
   sqlite3: new SQLite3_Client({}),
-  mssql: new MSSQL_Client({}),
-}
+  mssql: new MSSQL_Client({})
```
elhigu: You shouldn't do styling changes in the same pull request that is doing other type of changes. This PR is huge enough without these, It makes reviewing harder since I have to check for every line if it is functional change or not. Specially important with this one, since I've skimmed through this already multiple times.

test/unit/query/builder.js
irsl: I turned the lambdas to regular callback functions using babel. I didn't notice the styling and other side-effects it caused but was just happy that I don't need to do it by hand and unit tests were successful on each platform.
Anyhow, are you happy at all with the way how I refactored the verification logic in the query builder unit test? I wouldn't like to spend additional hours on this and then hearing you disprefer it for any reasons.

test/unit/query/builder.js
```diff
@@ -51,12 +64,11 @@ function verifySqlResult(dialect, expectedObj, sqlObj) {
   });
 }
 
-function testsql(chain, valuesToCheck, selectedClients) {
+function testsql(queryBuilderCallback, valuesToCheck, selectedClients) {
   selectedClients = selectedClients || clients;
   Object.keys(valuesToCheck).forEach(function(key) {
-    var newChain = chain.clone()
-        newChain.client = selectedClients[key]
```
elhigu: Is there some reason why setting the client didn't work? if setting client in chain doesn't affect all subqueries etc, then it is another bug in knex.

test/unit/query/builder.js
elhigu: Good that you mentioned it. I hadn't actually looked that part why all query builder params had been changed to function calls. I now looked it and I still don't understand why it was necessary.

irsl: The answer is here: https://github.com/tgriesser/knex/pull/2063#issuecomment-301566939
Raw subqueries were always generated through Sqlite dialect, so changes in the Sqlite field wrapping actually broke other tests (like the one in the referenced comment)

elhigu: Ok, that sounds like a bug in knex that it doesn't use the same client for all parts in subquery tree. That should be fixed.

For a work around using always pg dialect for raw subqueries could have kept same kind of functionality with old version of sqlite dialects raw.

irsl: Agreed, it should be. 
To avoid further rejections: do you want me to rebase my commits but instead of using callbacks switching the "default" dialect to PG?

elhigu: you could try if it works that way, I think that the whole PR will be a lot less intrusive and I would feel better about it. 

If that doesn't work either, I think that I have to look through query builder code and see if it is easy enough to fix that clients of subqueries are automatically patched to use client of the base query, fix that and then get back to this PR.

irsl: Just checked, switching the query builder tests to Postgres seems to be an acceptable workaround.

Issue #1758, opened 2016-10-26
MSSQL LIMIT(top) causes Query input data to be put in wrong spot


starsoccer:
I was trying to use knex to make a simply query but it seems to be switching up all of the variables in the query.

For example, My code is, 

console.log(knex('transactions').limit(1).where({ "transactionid": 'bcc021e6db2620680e773fd7c5d8726be8f74902728b80397f99fe8445abb403' }).update({ 'currentstatus': '-1' }).toString());

Yet it returns,

update top ('-1') [transactions] set [currentstatus] = 'bcc021e6db2620680e773fd7c5d8726be8f74902728b80397f99fe8445abb403' where [transactionid] = 1;

It seems to be taking what I set as the currentstatus, and using it in the top, and then using the transactionid for the currentstatus, and then the currentstatus as the transactionid.

starsoccer:
On further testing the issue seems to be related to limit. As soon as I remove limit from my query, the query becomes correct

starsoccer:
Issue still not resolved. Select and deleted are not effected, update causes the query to completely break

dirkmc:
I'm seeing the same issue with select
```
knex('MyTable')
.limit(2)
.select('MyID', knex.raw(`MyCol = :someVal as MyCol`, { someVal: 'the val' }))
.toString()

"select top ('the val') [MyID], MyCol = 2 as MyCol from [MyTable]"
```

starsoccer:
Hopefully someone can patch this bug

dhensby:
fixed by #2066?

elhigu:
yep https://runkit.com/embed/qh57udiyjwoi

PR #2066, opened 2017-05-15
Fix mssql binding ordering for queries that combine a limit with select raw or update

dirkmc:
#1758
Note that I also fixed a couple of unit tests that were passing even though they should have failed


elhigu:
Looks good, thanks. I'll rerun the tests until they passes... I really need to fix that oracledb random fail at some point.

dirkmc:
ha yeah I was surprised the tests failed there, I figured it must be something random as I hadn't changed anything related to oracle. Thanks for the quick update!

Issue #2065, opened 2017-05-15
db.schema.hasTable('') returns true


capaj:
This may be some weird behaviour of mysql I don't know about, but shouldn't this method return false?
```javascript
const hasTable = await db.schema.hasTable('')
  if (hasTable) {
    console.log('hasTable: ', hasTable)
```

debug output:
```
  knex:query show tables like ? +2ms
  knex:bindings [ '' ] +0ms
  knex:client releasing connection to pool: __knexUid1 +10ms
  knex:pool INFO pool mysql:mysql:client0 - dispense() clients=0 available=3 +1ms
hasTable:  true

```

On Knex 0.13.0

elhigu:
looks like it is not exact match, that hasTable does on mysql:

```sql
MariaDB [knex_test]> show tables like '';
+----------------------------+
| Tables_in_knex_test ()     |
+----------------------------+
| 10_test_table              |
| BatchInsert                |
| accounts                   |
| batchInsertDuplicateKey    |
| bool_test                  |
| catch_test                 |
| charset_collate_test       |
| composite_key_test         |
| datatype_test              |
| invalid_inTable_param_test |
| knex_migrations            |
| knex_migrations_lock       |
| person                     |
| primarytest                |
| renameColTest              |
| taabl                      |
| test_default_table         |
| test_default_table2        |
| test_table                 |
| test_table_two             |
+----------------------------+
20 rows in set (0.00 sec)

MariaDB [knex_test]> 
```

I would consider this as a knex bug. Knex should check that exact match was found in the returned table list or change query to be something that throws an error if table doesn't exist like shown here  http://stackoverflow.com/a/8829109/360060

elhigu:
Fixed via #2097

PR #2097, opened 2017-06-03
Change the way that MySQL verifies if a table exists (hasTable)

rbadillap:
Hi, on MySQL dialect, the method **hasTable** uses the following query to verify if a table exists:

```sql
SHOW TABLES LIKE ?
```

This, could represent an issue if no parameter is defined ([check this issue](https://github.com/tgriesser/knex/issues/2065)).

I've changed the way to verify the existence of the table, as @elhigu said in the same issue, the best way to accomplish this is checking [MySQL's Information Schema](https://dev.mysql.com/doc/refman/5.7/en/tables-table.html)

I've also added a new test passing an empty parameter on **hasTable** returning *false*.

With this PR, you can close the issue https://github.com/tgriesser/knex/issues/2065.

Edited: The PR has not passed successfully the tests. This is not related with my changes, attached you can find a screenshot of the approved test that I've added.

screen shot 2017-06-02 at 10 44 55 pm

### DIFF
--- a/src/dialects/mysql/schema/compiler.js
+++ b/src/dialects/mysql/schema/compiler.js
@@ -20,9 +20,20 @@ assign(SchemaCompiler_MySQL.prototype, {
 
   // Check whether a table exists on the query.
   hasTable(tableName) {
+    let sql = 'select * from information_schema.tables where table_name = ?';
+    const bindings = [ tableName ];
+
+    if (this.schema) {
+      sql += ' and table_schema = ?';
+      bindings.push(this.schema);
+    } else {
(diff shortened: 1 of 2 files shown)


rbadillap:
BTW, I'm aware of this PR https://github.com/tgriesser/knex/pull/2026, it is more complete than my solution, but I'm not seeing any update on that PR.

Let me know if you want to close this PR in order to give more importance to the PR mentioned above.

elhigu:
Looks good to me. Thanks! To me PR #2026 is kind of separate issue.

PR #2098, opened 2017-06-03
Add unit tests for table comment

rizip1:
I created tests to add table comment. These are for postgresql, mysql and oracle as it seems to me that ```.comment()``` function is not supported for sqlite.
I test setting the comment and setting it back to empty string. I assume that many users will expect that ```.comment(null)``` or simply ```.comment()``` will work the same as ```.comment('')``` I also made tests for these cases. However ```.comment(null)``` and ```.comment()``` does not work that way in mysql so these tests are only for postgresql and oracle.

### DIFF
--- a/test/unit/schema/oracle.js
+++ b/test/unit/schema/oracle.js
@@ -497,6 +497,24 @@ describe("Oracle SchemaBuilder", function() {
     expect(tableSql[0].sql).to.equal('alter table "users" add "foo" decimal(2, 6)');
   });
 
+  it('test set comment', function() {
+    tableSql = client.schemaBuilder().table('users', function(t) {
+      t.comment('Custom comment');
+    }).toSQL();
+
+    equal(1, tableSql.length);
+    expect(tableSql[0].sql).to.equal('comment on table "users" is \'Custom comment\'');
+  });
+
+  it('test set empty comment', function() {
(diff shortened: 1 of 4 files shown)


test/unit/schema/oracle.js
```diff
@@ -497,6 +497,42 @@ describe("Oracle SchemaBuilder", function() {
...
+  it('test set empty comment', function() {
+    tableSql = client.schemaBuilder().table('users', function(t) {
+      t.comment('');
+    }).toSQL();
+
+    equal(1, tableSql.length);
+    expect(tableSql[0].sql).to.equal('comment on table "users" is \'\'');
+  });
+
+  it('test set empty comment using null', function() {
```
elhigu: I wouldn't consider setting comment to be empty string by setting it to null or undefined as a intended behaviour. So better to not add test for that functionality, since better test would be throwing an error because of invalid parameter.

rizip1:
Ok I deleted those non-string args tests and rebased.

rizip1:
I have one question about automatic tests. Usually I must push 3-5 times to get tests passed. The test that fails is always ```renameColumn``` in ```oracle | oracledb``` section (the same test that failed for this push). Should I ignore this test fail when pushing new stuff?

PR #2099, opened 2017-06-03
Add unit test for renaming column in postgresql

rizip1:
There are already unit tests for renaming column but only for oracle.
I figured out that unit testing sqlite, mssql and mysql can not be
done nicely using only ```toSQL``` method but is implemented
in ```test/integration/builder/additional.js```. However test for
postgresql can be done similarly to oracle using no dirty checks
so maybe it can be added.

### DIFF
--- a/test/unit/schema/postgres.js
+++ b/test/unit/schema/postgres.js
@@ -245,6 +245,14 @@ describe("PostgreSQL SchemaBuilder", function() {
     expect(tableSql[0].sql).to.equal('alter table "users" rename to "foo"');
   });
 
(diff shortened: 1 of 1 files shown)


rizip1:
When I run ```npm run test``` locally tests pass without error. However here it fails in ```oracle | oracledb``` part that I have not touched. Does anybody have any idea what is wrong? Thanks for help.

elhigu:
Thanks! Yeah oracle fail is random fail that no one has had time to fix. Sorry about that. I'll try to make test pass and merge after that.

PR #2100, opened 2017-06-04
Update with no values should produce a noop

krizalys:
Update with no values should produce a noop (rather than `update my_table set where ...`, which is malformed SQL).

### DIFF
--- a/src/dialects/mssql/query/compiler.js
+++ b/src/dialects/mssql/query/compiler.js
@@ -78,6 +78,15 @@ assign(QueryCompiler_MSSQL.prototype, {
 
   // Compiles an `update` query, allowing for a return value.
   update() {
+    const updateValues = this.single.update || [];
+    if (Array.isArray(updateValues)) {
+      if (updateValues.length === 0) {
+        return '';
+      }
+    } else if (typeof updateValues === 'object' && isEmpty(updateValues)) {
+      return '';
+    }
+
     const top = this.top();
     const updates = this._prepUpdate(this.single.update);
     const join = this.join();
--- a/src/dialects/mysql/query/compiler.js
+++ b/src/dialects/mysql/query/compiler.js
@@ -4,7 +4,7 @@
 import inherits from 'inherits';
 import QueryCompiler from '../../../query/compiler';
 
-import { assign } from 'lodash'
+import { assign, isEmpty } from 'lodash'
 
 function QueryCompiler_MySQL(client, builder) {
   QueryCompiler.call(this, client, builder)
@@ -17,6 +17,15 @@ assign(QueryCompiler_MySQL.prototype, {
 
   // Update method, including joins, wheres, order & limits.
   update() {
+    const updateValues = this.single.update || [];
+    if (Array.isArray(updateValues)) {
+      if (updateValues.length === 0) {
+        return '';
+      }
+    } else if (typeof updateValues === 'object' && isEmpty(updateValues)) {
+      return '';
+    }
+
     const join = this.join();
     const updates = this._prepUpdate(this.single.update);
     const where = this.where();
--- a/src/dialects/oracle/query/compiler.js
+++ b/src/dialects/oracle/query/compiler.js
@@ -108,6 +108,15 @@ assign(QueryCompiler_Oracle.prototype, {
 
   // Update method, including joins, wheres, order & limits.
   update() {
+    const updateValues = this.single.update || [];
+    if (Array.isArray(updateValues)) {
+      if (updateValues.length === 0) {
+        return '';
+      }
+    } else if (typeof updateValues === 'object' && isEmpty(updateValues)) {
+      return '';
+    }
+
     const updates = this._prepUpdate(this.single.update);
     const where = this.where();
     let { returning } = this.single;
--- a/src/dialects/oracledb/query/compiler.js
+++ b/src/dialects/oracledb/query/compiler.js
@@ -226,6 +226,15 @@ _.assign(Oracledb_Compiler.prototype, {
   },
 
   update: function() {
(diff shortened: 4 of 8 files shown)


wubzz:
@krizalys This is a cool idea and I guess technically prevents potential code issues such as trying to update with an empty object. But in the sense of a "noop", no connection should be created and no query should be run, right?

Right now as far as I can tell this is equal to `raw('').then(() => {...}).catch(() => {...})` which does pull a connection from the pool and does issue an empty query ''.

I'm also honestly not sure if this is the best approach, but I see how it can be an issue. What do you think @elhigu ?

elhigu:
@wubzz I would prefer throwing an error since an empty update sounds like a bug in code.

krizalys:
@wubzz @elhigu I adopted the strategy established by `insert()` (https://github.com/tgriesser/knex/blob/master/test/unit/query/builder.js#L2572-L2620)

Right, the meaning of *noop* could be refined and both empty operations could be improved to not touch the connections at all.

Regarding the idea of throwing an error instead, empty queries are not always the result of a bug: imagine an HTTP `PATCH` request with a simple `application/json` payload `{}`, which is still valid JSON, it may be regarded as a valid RESTful operation, yet would produce a query builder exception.

This, and the fact that a query builder is expected to abstract SQL intricacies such as this one, it should now how to handle this gracefully IMHO.

With that being said, I can see value in throwing an error **in certain situations**, so perhaps what we should do is to make this behavior configurable to allow users choose the most appropriate strategy.

Whatever the decision, `insert()` and `update()` need to have a consistent behavior :)

elhigu:
Missing keys in insert is completely different issue. It is about the fact that insert allows to add multiple rows in one query, which forces knex to take an action if some of the inserted rows has less keys that others. This problem doesn't exist with update, which has only one set of key value pairs to set.

Also with insert adding empty object makes perfectly sense, because columns will be filled with default values and auto-incremented id.

> Regarding the idea of throwing an error instead, empty queries are not always the result of a bug: imagine an HTTP PATCH request with a simple application/json payload {} - which is still valid JSON. It may be regarded as a valid RESTful operation, yet would produce a query builder exception.

It would mean that if you are allowing passing `{}` from your API. You should validate your input and not call query builder at all. If you use it as a feature that knex allows passing an empty object, naturally it is not a problem. It is just about us how we want to define knex's behaviour.

If other option is silently ignore update and not create query at all instead of throwing an error I still stand by the decision of throwing an error instead of silently decide not to execute any query (knex's user should decide if he wants query to be executed).

One case where I can think of need for being able to update row without any data is to force row to run on-update triggers, but if knex doesn't send any query for empty object, then that wouldn't work anyways. To me better way to do that is to do some dummy update for triggering those like `.update({ createdAt: knex.raw('??', ['createdAt])})`.

For configurability, I would like to avoid that. This problem is so easy to provent by the code that is calling knex so I don't see a point in trying to support everything for everyone. Less well tested consistent behaviour is better.

>  Whatever the decision, insert() and update() need to have a consistent behavior :)

Suprisingly I need to disagree with this one too :) In SQL `insert` and `update` are significantly different operations. 

With `insert` empty values makes sense since columns has default values to fill in, but with `update` they doesn't and creating dummy update for that corner case where one likes to run update triggers is easy to implement on the knex's user side who knows what like of table rows one is updating.

elhigu:
@rhys-vdw @wubzz @tgriesser @anyone second opinions would be welcome.

wubzz:
@elhigu I agree with everything that was said. In current version of knex, an error is already technically thrown, since the query that is being run is invalid. 

However - like you said - I also think it would be better to throw a specific error for this scenario, since that will not only make it easier for the user of knex to know what is wrong, but it will also prevent connections from being pulled from the pool, saving resources.

krizalys:
@elhigu Please observe that the current behavior of `insert` is as follows:

```javascript
knex('table').insert().then().catch(console.log);
// => 1st form, empty query generated; no row

knex('table').insert({}).then().catch(console.log);
// => 2nd form, eg: 'INSERT INTO `table`() VALUES()' generated; attempt to create row with default values

knex('table').insert({i: 1}).then().catch(console.log);
// => 3rd form, row created as expected
```

Part of what I'm proposing is to allow `update()` and produce no query, which is consistent with `insert()` (1st form).

If we do not want to allow this, or if we want to throw an error, then better API consistency `insert`/`update` would be achieved by:

* also disallowing/throwing an error for `insert()` (1st form) ;
* **or**, making `insert()` behave as `insert({})` (2nd form).

The other part of what I am proposing is to allow `update({})` (2nd form). For this operation, the best semantic we could achieve would be to produce a query updating each cell of each matching row to their default value, such as:

```sql
UPDATE `table` SET i = DEFAULT;
```

This would also improve API consistency with `insert({})` (2nd form), but as you suggested, it could involve tricky interactions with the database, perhaps be bound to portability limitations. I did not bother with it and stuck with a less ideal noop at this point, but let's see what is the interest over the whole idea.

elhigu:
@krizalys I agree that `.insert()` without any parameters should throw an error. Docs specify currently that it should take `{...}` or `[{...},{...},...]` and we've been adding errors because of invalid parameters to many APIs lately.

> The other part of what I am proposing is to allow update({}) (2nd form). For this operation, the closest semantic would be to produce a query updating each cell of each matching row to its default value, such as:

This is not possible to implement. Knex doesn't have any information about what columns there are in the table. Also it wouldn't make much sense to have magic functionality that overrides all old row's column values with defaults when an empty object is passed in. Update and insert are very different operations with very different syntax and functionality, there is no use to try to abstract them to behave the same way.

elhigu:
Closing, since I don't think this is right thing to do. I'm sorry for wasting the effort put to this.

PR #2101, opened 2017-06-04
Mysql should set empty comment for non-string args

rizip1:
In MySQL If I use ```comment(null)```, ```comment(false)``` or just ```comment()```
I expect that Mysql will set comment to empty string not to ```'null'```, ```'false'``` or ```'undefined'```.

### DIFF
--- a/src/dialects/mysql/schema/tablecompiler.js
+++ b/src/dialects/mysql/schema/tablecompiler.js
@@ -56,7 +56,7 @@ assign(TableCompiler_MySQL.prototype, {
 
   // Compiles the comment on the table.
   comment(comment) {
-    this.pushQuery(`alter table ${this.tableName()} comment = '${comment}'`);
+    this.pushQuery(`alter table ${this.tableName()} comment = '${comment || ''}'`);
   },
 
   changeType() {


elhigu:
Thanks for PR. Anyways, I would like knex to rather throw an error in this case. Non-string parameter should be invalid (we've been working already a while to prevent knex from ignoring invalid usage silently).  Also tests and possibly documentation needs to be updated.

rizip1:
Ok I understand. I made this PR because in sqlite or postgres when you use ```null```, ```false``` or ```undefined``` as ```comment``` param, the comment is set to ```''```. So I though that it would be more consistent if MySQL behaves the same. I also made PR #2098 where I added few unit tests for adding comment to table. As these non-string arguments for postgresql, oracle and so on translate into empty string I also tested these. Maybe it would make sense to add warning into oracle, postgres etc when comment is set to non-string arg as error would be breaking change for those using it that way and throw an error exclusively for MySQL. Would you agree? If yes I could make PR for it.

elhigu:
I would rather have breaking change and throw an error on every dialect instead of warnings on some, this is also not so frequently used feature that it will affect on every knex user. 

There are currently also many other breaking changes lined up for next release so one more won't hurt much.

rizip1:
Ok I made PR for that #2126.

elhigu:
can this be closed, I skimmed through #2126 and it did look goo, but I need to check it out with a bit more thought later on.

PR #2102, opened 2017-06-04
Add unit test for cross-join

rizip1:
Add unit test for cross-join as there was no one currently.
In the first test-case I test composition of two cross-joins.
In the second test-case I test cross-join with ```on``` statement.
I found out that cross-join with ```on``` is only valid for MySQL
and sqlite3 so the test includes only these (it is weird to use
cross-join with ```on``` anyhow).


elhigu:
Thanks!

PR #2104, opened 2017-06-04
integration test for sqlite3, referencing non-existent column in the query

irsl:
Should have been part of the original PR:
https://github.com/tgriesser/knex/pull/2087


elhigu:
Thanks!

PR #2105, opened 2017-06-04
Use the schema when dropping indices in Postgres.

ThePumpingLemma:
Ran into an issue where indices created using `withSchema` could not be dropped correctly since dropping couldn't take the schema into account.

Note that the name can optionally be schema qualified according to the documentation: https://www.postgresql.org/docs/9.6/static/sql-dropindex.html


elhigu:
Seems legit, thanks!

Issue #1979, opened 2017-03-17
table.dropColumns doesn't work with sqlite3


cjwatson:
When using SQLite, `table.dropColumns` doesn't accept multiple column names the way it does with other drivers, per the [documentation](http://knexjs.org/#Schema-dropColumns) (I compared with how `table.dropTimestamps` is implemented just to make sure that I was using it correctly).  Here's how to reproduce this, in a fresh directory:

    $ npm init -y
    $ npm i --save knex sqlite3
    $ node_modules/.bin/knex init
    $ node_modules/.bin/knex migrate:make start
    $ cat >migrations/*_start.js
    exports.up = function(knex, Promise) {
      return Promise.all([
        knex.schema.createTable('user', function(table) {
          table.increments();
        })
      ]);
    };

    exports.down = function(knex, Promise) {
      return Promise.all([
        knex.schema.dropTable('user')
      ]);
    };
    $ node_modules/.bin/knex migrate:make add-columns
    $ cat >migrations/*_add-columns.js
    exports.up = function(knex, Promise) {
      return Promise.all([
        knex.schema.table('user', function(table) {
          table.text('name');
          table.text('login');
        })
      ]);
    };

    exports.down = function(knex, Promise) {
      return Promise.all([
        knex.schema.table('user', function(table) {
          table.dropColumns(['name', 'login']);
        })
      ]);
    };
    $ node_modules/.bin/knex migrate:latest
    $ node_modules/.bin/knex migrate:rollback

Instead of successfully dropping the columns, the output is:

    Using environment: development
    Knex:warning - sqlite does not support inserting default values. Set the `useNullAsDefault` flag to hide this warning. (see docs http://knexjs.org/#Builder-insert).
    Knex:warning - migrations failed with error: The column name,login is not in the user table
    Error: The column name,login is not in the user table
        at SQLite3_DDL.<anonymous> (/home/cjwatson/k/node_modules/knex/lib/dialects/sqlite3/schema/ddl.js:56:28)
        at SQLite3_DDL.tryCatcher (/home/cjwatson/k/node_modules/bluebird/js/release/util.js:16:23)
        at SQLite3_DDL.getColumn (/home/cjwatson/k/node_modules/bluebird/js/release/method.js:15:34)
        at client.transaction.connection (/home/cjwatson/k/node_modules/knex/lib/dialects/sqlite3/schema/ddl.js:223:21)
        at /home/cjwatson/k/node_modules/knex/lib/transaction.js:81:20
[...]

PR #2107, opened 2017-06-05
Fix drop of multiple columns in sqlite3

rizip1:
* Bugfix for issue #1979 
* The problem was that dropColumn method in sqlite3
  Compiler was using only first argument it got from
  apply
* Reworked to use 'arguments' instead

### DIFF
--- a/src/dialects/sqlite3/schema/ddl.js
+++ b/src/dialects/sqlite3/schema/ddl.js
@@ -195,16 +195,19 @@ assign(SQLite3_DDL.prototype, {
     }, {connection: this.connection})
   }),
 
-  dropColumn: Promise.method(function(column) {
+  dropColumn: Promise.method(function(columns) {
     return this.client.transaction(trx => {
       this.trx = trx
-      return this.getColumn(column)
+      return Promise.all(columns.map(column => this.getColumn(column)))
       .bind(this)
       .then(this.getTableSql)
       .then(function(sql) {
         const createTable = sql[0];
-        const a = this.client.wrapIdentifier(column);
-        const newSql = this._doReplace(createTable.sql, a, '');
+        let newSql = createTable.sql;
+        columns.forEach(column => {
+          const a = this.client.wrapIdentifier(column);
+          newSql = this._doReplace(newSql, a, '');
+        })
         if (sql === newSql) {
           throw new Error('Unable to find the column to change');
         }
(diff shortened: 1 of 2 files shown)


elhigu:
Documentation says that `dropColumn` drops just one column and `dropColumns` is for multiple columns. Am I missing something here?

http://knexjs.org/#Schema-dropColumn
http://knexjs.org/#Schema-dropColumns

rizip1:
Yes you are right. The problem in sqlite3 is that ```dropColumns``` will drop only first column. So I made it work for multiple columns. But the way I fixed it  ```dropColumn``` now behaves same as ```dropColumns``` what may not be desired so can rework this.

elhigu:
I have no idea why there are 2 separate methods for dropping columns, but since they are already there, better to keep functionality as it is documented.

rizip1:
I have looked on the behavior of ```dropColumn``` and ```dropColumns``` and I think that they do the same. When you look into ```schema/builder/tablebuilder.js``` file you can see this piece of code
```
AlterMethods.dropColumn =
AlterMethods.dropColumns = function() {
  this._statements.push({
    grouping: 'alterTable',
    method: 'dropColumn',
    args: toArray(arguments)
  });
  return this;
};
```
So ```dropColumn``` is in fact ```dropColumns```. I tried it also in postgresql and you can drop multiple columns just using ```dropColumn``` function. From that point of view ```dropColumn``` is really useless because one can just use ```dropColumns``` also for single column.
So to keep functionality as documented in sqlite3 a would also have to change behavior for other dialects (this function is common for all dialects, so it implies that even now the functionality does now follow the documentation).
Would not it be better to take the fix as it is because anyway this strange behavior is common for other dialects just for sqlite3 it does not work now. And maybe do the refactoring in separate PR?

elhigu:
@rizip1 sounds legit. Would be good to also fix documentation about this to tell that functions are actually the same. Was there already integration test for this functionality (dropColumn / dropColumns)?

rizip1:
Currently there is no integration test for ```dropColumns``` but there are some for ```dropColumn```.  Not sure if changing documentation is good way to go as knowing that ```dropColumn``` will also drop multiple columns could lead to complications later if someone starts to use it that way and the functionality changes in the future.

elhigu:
thanks 👍

Issue #1903, opened 2017-02-08
Add check for column existance when renaming


jehy:
I use mariadb with `mysql` client.

If I try to rename non existent column, I get 

> TypeError: Cannot read property 'Type' of undefined
>     at /web/quest/node_modules/knex/lib/dialects/mysql/schema/tablecompiler.js:104:83
>     at tryCatcher (/web/quest/node_modules/bluebird/js/release/util.js:16:23)
>     at Promise._settlePromiseFromHandler (/web/quest/node_modules/bluebird/js/release/promise.js:510:31)
>     at Promise._settlePromise (/web/quest/node_modules/bluebird/js/release/promise.js:567:18)
>     at Promise._settlePromiseCtx (/web/quest/node_modules/bluebird/js/release/promise.js:604:10)
>     at Async._drainQueue (/web/quest/node_modules/bluebird/js/release/async.js:138:12)
>     at Async._drainQueues (/web/quest/node_modules/bluebird/js/release/async.js:143:10)
>     at Immediate.Async.drainQueues (/web/quest/node_modules/bluebird/js/release/async.js:17:14)
>     at runCallback (timers.js:651:20)
>     at tryOnImmediate (timers.js:624:5)
>     at processImmediate [as _immediateCallback] (timers.js:596:5)

Error is on this line:
```javascript
var sql = 'alter table ' + table + ' change ' + wrapped + ' ' + column.Type;
```

The real reason of error (no column) is very hard to understand.

Simple check for `column===undefined` could be very useful.

elhigu:
Closing in favor of #2155 which has some extra discussion how to fix this.

PR #2110, opened 2017-06-07
Descriptive log when renaming non existing column

rizip1:
Fix for issue #1903 

When renaming column that does not exist migration fails and from the stack trace in console it might not be clear what is the problem.

I added check for column existence, and when the column does not exist it logs meaningful error to console and stack trace also.

There are however more ways to solve this issue. One would probably be to add some check for each dialect (db server) but that is a lot of redundancy. The other approach might be to somehow check column existence in Builder however all builder and compiler stuff is synchronous. The check for column would be asynchronous that results in lot of changes. So in Runner I first compile the query and during the compilation I remember the columns that must exist in array that belongs to Builder. Before running the query I just check If all columns that will be referenced exist. Then the original query is run if no errors were found otherwise error message is logged and exception thrown.

### DIFF
--- a/src/runner.js
+++ b/src/runner.js
@@ -35,11 +35,39 @@ assign(Runner.prototype, {
         helpers.debugLog(sql)
       }
 
-      if (isArray(sql)) {
-        return runner.queryArray(sql);
+      const runOriginalSql = function() {
+        if (isArray(sql)) {
+          return runner.queryArray(sql);
+        }
+        return runner.query(sql);
       }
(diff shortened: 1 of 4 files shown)


rizip1:
I found out that I would not have any time to solve this so maybe it can be closed now, and reopen later if needed.

elhigu:
There was actually alternative approach discussed how to fix this, which would need only single query to get old column info and to create new column... so yep I think closing is ok to do for now. Thanks anyways!

PR #2113, opened 2017-06-08
Fix falsly setting autoTransaction

DerTieran:
When passing a falsy value to transacting a new transaction is created but `autoTransaction` is set to false so the transaction is never committed or rolled back.

This is a problem because this means `batchInsert` works different then `insert` here.
So passing `null` into `transacting` works for `insert`
```js
knex.insert(rows)
    .into('TableName')
    .transacting(null)
```
but it doesn't for `batchInsert`
```js
knex.batchInsert('TableName', rows, chunkSize)
    .transacting(null)
```


wubzz:
Thanks!

PR #2117, opened 2017-06-11
Add unit test for right/rightOuterJoin combination

rizip1:
I found out that there is no unit test for right/rightOuterJoin so here is one that combines them both.
I do not include test for sqlite3 because it currently does not support right join.


elhigu:
thanks!

capaj:
just out of curiosity-if I were to call right/rightOuterJoin with sqlite3 I'd get some error thrown right?

PR #2118, opened 2017-06-11
Add fullOuterJoin unit test

rizip1:
Add unit test for ```fullOuterJoin```. No unitTest for MySQL and sqlite3 because they do not support it. Btw I found out that there is also ```outerJoin``` function which however will not work for any dialect. The issue is that it translates to ```OUTER JOIN ON ...``` which is not supported anywhere but ```FULL JOIN ON``` is supported in mssql, postgres and oracle so that is probably a bug.

### DIFF
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -1922,6 +1922,27 @@ describe("QueryBuilder", function() {
     });
   });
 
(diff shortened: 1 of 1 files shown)


elhigu:
Thanks!

PR #2126, opened 2017-06-16
Throw error on non-string table comment

rizip1:
According to discussion in #2101 

When setting table comment to be empty, knex will throw error on non string args

* add unit tests for all dialects
* removed retyping to empty string

All the unit tests are the same. Do you handle this type of redundancy somehow or is it ok with unit tests? Theoretically it might be used as integration test, but I do not think this will be right because these comment tests test only single unit of functionality.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "babel-preset-es2015": "^6.13.2",
     "chai": "^3.5.0",
     "coveralls": "~2.11.1",
-    "dockerode": "^2.4.3",
+    "dockerode": "2.4.3",
     "eslint": "2.2.0",
     "eslint-plugin-import": "^1.8.0",
     "istanbul": "^0.4.5",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "babel-preset-es2015": "^6.13.2",
     "chai": "^3.5.0",
     "coveralls": "~2.11.1",
-    "dockerode": "2.4.3",
+    "dockerode": "^2.4.3",
     "eslint": "2.2.0",
     "eslint-plugin-import": "^1.8.0",
     "istanbul": "^0.4.5",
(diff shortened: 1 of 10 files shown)


elhigu:
@smulesoft docker tests started failing with node 0.12 any ideas how to get it working, maybe those tests could be at least disabled from 0.12?

smulesoft:
It's a `dockerode` issue (the node docker API). Change the version as shown below and it'll pass
sd

smulesoft:
@rizip1 Did you check this?
I think you should actually set the version to "~.2.4.3" (and if that doesn't work you can check the exact version "2.4.3")

rizip1:
Tried both options but then all tests fail. Is there something else I must do when changing the versions?

smulesoft:
- I checked out your branch and I can run all tests locally both with node `0.12` and `6.10`. What about you? Can you run them locally? 
- Those are all `timeout` errors in the `afterEach` of the same test. Why don't you give it one more trying increasing the timeout in `knexfile.js` for the `postgres` test:

```js
    docker: {
      factory:   './postgres/index.js',
      container: 'knex-test-postgres',
      image:     'postgres:9.6',
      database:  'postgres',
      username:  'postgres',
      password:  '',
      hostPort:  '49152',
      client:    'pg',
      timeout:   10 * 1000 // <-- Here, try 15 or 20 seconds (current value is 10 seconds)
    }
```

elhigu:
Please, rebase, hopefully it work against the latest master.

rizip1:
Done.

package.json
```diff
@@ -34,7 +34,7 @@
     "babel-preset-es2015": "^6.13.2",
     "chai": "^3.5.0",
     "coveralls": "~2.11.1",
-    "dockerode": "^2.4.3",
```
elhigu: I think, this is not necessary anymore, since I dropped node 0.12 support alltogether.

smulesoft: Yeah maybe you can keep using new minor versions, or at least try again with `~2.4.3` before sticking to `2.4.3`.

PR #2145, opened 2017-07-03
Add cascade option to .dropTable() and .dropTableIfExists()

ricardogama:
This PR updates the schema builder's `.dropTable()` and `.dropTableIfExists()` methods to accept the `options` argument.

For now, providing the `cascade` option adds the cascading clause to drop table statements:

```js
knex.schema.dropTable('users', { 
  cascade: true 
});
```

This feature has no breaking changes as no tests needed to be changed, and being this argument an object enables to add more features in the future like `PURGE` for Oracle or `RESTRICT` for PostgreSQL.


elhigu:
This would need also an integration test to make sure that all the queries actually work as they are intended. I'll fix the test bench soon to get test to run again.

ricardogama:
@elhigu As I started digging into the integration tests I found some issues:

* The cascade option has been deprecated since MySQL 5.7, although allowed to facilitate porting;
* SQLite3 has no support for this option;
* Oracle has no integration tests;

Considering the integration suite runs the same tests for each dialect, I'm not sure how should I add tests specifically for PostgreSQL or a previous version of MySQL.

WDYT?

elhigu:
@ricardogama there are many integration tests, which first checks that client is correct and only includes the test for that dialect. Also please rebase, I might have fixed the CI problems for now.

elhigu:
And oracle do have integration tests and they are ran on CI. Mssql is not CI tested yet though. I'm having second thoughts about this, maybe it is not a good idea to implement deprecated API for mysql. Also I'm not sure if it is worth effort to make this as a feature if it works only with postgresql (in that case one could just write it with knex.raw).

Would this work on postgresql, oracle and mssql though?

elhigu:
Closing for now, may be reopened if this is going to be finished.

PR #2164, opened 2017-07-18
Fixing docker CI tests

elhigu:


elhigu:
Looks like I need to setup linux box and try to reproduce the problem locally.

elhigu:
Could not reproduce the problem with node 8 even inside vagrant box. Removing node 8 for now.

Issue #1922, opened 2017-02-18
Including TypeScript files by default in migrations is a breaking change


Seikho:
See #1769 

My existing projects can no longer run migrations when upgrading to `0.12.7` due to now including `*.ts` files by default and expecting `ts-node` to be present.

This behaviour should be opt-in to be backwards compatible.

elhigu:
Are you compiling your javascript files to same directory where your typescript files are? I can see that causes failing..

This seems to be accidentally backwards incompatible change. Though this also is not backwards incompatible in the sense that it is undocumented functionality how running migrations reacts to extra files in migrations directory. 

Nevertheless existence of ts-node should be optional and there should be good error message if it is needed and it is missing. Also it should be possible to configure somehow extensions / glob which migrations are included (with glob / extension list with some defaults this also becomes documented functionality).

Seikho:
Yes, my project's compile output is adjacent to the source.

I'm unsure how to resolve the issue we're in. Do we fix the breaking change by reverting the change and introducing opt-in behaviour then introduce the current `0.12.7` behaviour in `0.13.0` ?

elhigu:
@Seikho I would say that it will be easiest to just add configuration attribute 

```js
var knex = require('knex')({
  client: ...,
  connection: ...,
  migrations: {
    directory: 'migrations',   
    tableName: 'migrations',
    extensions: ['js', 'ts'], // both should be set by default 
  }
});
```

and filter migration file listing with that in `migrate/index.js` simple and straightforward. 

I also noticed that `directory` attribute is not documented so when updating documentation would be awesome if you could add also `directory`attribute to the example.

One should be able to give also longer extensions like `migration.coffee` to extension list so one can use some special naming convention for migrations and allow also other `.js` files to be in the same directory. 

In CONTRIBUTING.md is all the info what should be included in pull request.

elhigu:
And thanks a lot if you are able to pull this off, really appreciated! I will do 0.12.8 update when this is ready.

ravenscar:
just commenting in case anyone else is still experiencing this problem with typescript.

with 0.13.0 and node 8 I simply get the esoteric error:

```
(node:80280) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): SyntaxError: Unexpected token import
```

I eventually tracked this down to knex trying to run the `*.d.ts` generated definition files that TS will generate if you have `"declaration": true` in your `tsconfig.json`.

[...]

PR #2168, opened 2017-07-21
Allows to filter seeds and migrations by extensions

doomsower:
This PR addresses issue #1922 

I added `loadExtensions` parameter where you can specify allowed extensions of seeds/migrations to be executed. Default value is `['js', 'ts']` as was suggested in this issue, but now I think maybe it should be set to `SUPPORTED_EXTENSIONS` so the change is non-breaking?

Couldn't find any tests that deal with ts/coffee/etc. migrations, so I don't know where to start from. I.e. how to to check that ts migrations is executed/not executed based on loadExtensions parameter? But this is small change, I think it's ok as is.

If this PR is acceptable, I will submit PR to docs too.


elhigu:
@doomsower Thanks for the PR, contributions are extremely welcome :) I agree, would be better to have SUPPORTED_EXTENSIONS as default value. Otherwise PR looks clean, but it is missing related PR to documentation and tests.

doomsower:
@elhigu , I changed default extensions to former SUPPORTED_EXTENSIONS. I also added unit tests for `_listAll` in Seeder and Migrator, which are the only modified functions. They are the first unit test for Seeder and Migrator and they should be reviewed for sure.

Actually, I think it is better not to introduce this extra `loadExtensions` parameter, and support only one extension at a time by utilizing existing `extension` parameter. I think it is bad idea to have migrations in different languages at the same, except maybe for some legacy migrations in project that was converted from js to ts, for example. But for such project this would be a breaking change.

If everything is ok, I'll proceed with docs PR

elhigu:
Beautiful, I haven't seen such clean code here for a while 👍 And I agree about the design issue, that it is weird to support multiple language migrations at the same time, but I don't see any strong reason to start limiting it now.

Please proceed and if you see places where things could be made better and more safe for the user it is not out of the question to break backwards compatibility. We have been breaking it quite a lot lately while trying to make lib to work more sane manner, like throwing errors when user does something that is not defined or potentially dangerous behaviour.

doomsower:
Thank you!
Here is PR with docs: https://github.com/knex/documentation/pull/46
Any ETA when I can see this published on NPM?

elhigu:
Thanks! Everything seems to be in order. CI have been broken a while so I haven't been able to get some PRs in which were failing (and I've been very busy lately so I haven't got time to do much of maintenance)... so I cannot say anything about schedule yet. Next version will be 0.14.0 since there are again some breaking changes.

Issue #2169, opened 2017-07-23
MySQL Update Executes In Wrong Order


kurrestahlberg:
MySQL differs from the SQL spec in update by using whatever is the current value in a field at the time of evaluating an expression. More here: https://dev.mysql.com/doc/refman/5.7/en/ansi-diff-update.html

Knex seems to always order all set statements in alphabetical order.

This causes a problem in a statement like the following for example:

```UPDATE table SET col1=col1+1, col2=col1*3;```

In this case let's assume I want col2 to be three times col1's value AFTER the update. The statement above would work correctly for MySQL as col1's new value will be used for calculating col2. Unfortunately at least MariaDB seems to follow the SQL spec and actually use the value of col1 from before the update and therefore store the wrong result.

I could fix this for both databases by doing ```UPDATE table set col2=(col1 + 1)*3,col1=col1+1;``` as then col1 would have the old value with both databases when calculating the value of col2. The problem is that Knex seems to reorder all fields to be updated to alphabetical order and therefore take away the possibility to make sure the correct result is stored with both databases.

elhigu:
Sounds like a good idea to maintain order of keys when creating update... even that as far as I know javascript spec doesn't specify in which order they will be.

absolux:
@kurrestahlberg it's not the fault of knex, it's `Object.keys`, used to construct the `set` clause, which returns the keys ordered in alphabetical order

kurrestahlberg:
@absolux It is actually mostly the fault of MySQL because they deviate from the spec others follow, but Knex could be modified slightly to enable working around the issue. If Knex used a different way to store the update statements (or made sure to store ordering info alongside the other stuff) then keeping to original call order would be possible and the issue would go away.

Would a PR be accepted?

elhigu:
@absolux I actually tried `Object.keys({ col2: 1, col1: 1 })`, `_.keys`and `Object.getOwnPropertyNames` on node 6.10.  All of them did seem to maintain order of keys. I was suspecting that some other part, maybe some object merging or so could have caused that order change in knex. 

@kurrestahlberg I would be happy to accept PR for this (ofcourse well tested one).

absolux:
My mistake.
Actually the issue comes from `compiler._prepInsert(data)` which prepares the data and __*sorts [here](https://github.com/tgriesser/knex/blob/master/src/query/compiler.js#L565)*__ the columns names

kurrestahlberg:
[...]

PR #2171, opened 2017-07-25
Remove sorting of statements from update queries (#2169)

kurrestahlberg:
Consistent updates where a new field value depends on an another field that gets updated in the same query are impossible across different databases unless the original update order is maintained.

The problem is with MySQL not following the SQL spec as described here: https://dev.mysql.com/doc/refman/5.7/en/ansi-diff-update.html. This causes the value used for calculating another value sometimes already be updated and sometimes not, depending on which database is used and what the names of the fields are if the statements are sorted alphabetically.

This PR removes sorting of statements when building an UPDATE query. A possible future issue is that the key order of an object is not defined in the spec while all known JS engines keep non-integer-like keys in insertion order so this should work correctly across all engines. The order might thus be different with an engine in the future. More info on ordering here: https://bugs.chromium.org/p/v8/issues/detail?id=164

### DIFF
--- a/src/query/compiler.js
+++ b/src/query/compiler.js
@@ -592,13 +592,13 @@ assign(QueryCompiler.prototype, {
   _prepUpdate(data) {
     data = omitBy(data, isUndefined)
     const vals = []
-    const sorted = Object.keys(data).sort()
+    const columns = Object.keys(data)
     let i = -1
-    while (++i < sorted.length) {
+    while (++i < columns.length) {
       vals.push(
-        this.formatter.wrap(sorted[i]) +
+        this.formatter.wrap(columns[i]) +
         ' = ' +
-        this.formatter.parameter(data[sorted[i]])
+        this.formatter.parameter(data[columns[i]])
       );
     }
     return vals;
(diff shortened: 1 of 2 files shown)


elhigu:
Should insert be implemented the same way? Even though I don't see the same kind of problems with it so maybe not...

kurrestahlberg:
I was thinking the same but as I couldn't think of any reason why there would be an issue when inserting I figured a smaller change is better.

elhigu:
Yeah, this looks good to me though. Thanks! @wubzz @tgriesser @rhys-vdw anyone, if there are no objections I'll merge this in tomorrow.  I marked this as a breaking change even that this was more like undefined behaviour earlier. Maybe documentation could mention that updates are generated in the Object.key() iteration order of used js-engine. Just to make this functionality documented and official.

wubzz:
No opposition from me, this makes sense. My guess is the `.sort()` originates from a copy-paste of `_prepInsert`. Don't see a real use case for sorting columns for insert either, tho..

elhigu:
I added documentation issue about this. Thanks!

PR #2172, opened 2017-07-25
searchPath should be case-sensitive

### LINKED ISSUE #1751: Using `searchPath` does not respect case-sensitivity
Currently if knex is initialized with

``` javascript
knex({
client: 'pg',
connection: {...},
searchPath: 'TEST'
});
```

The search path will be `test` which is wrong, it should be `TEST`. This is a simple case of missing quotes.
https://github.com/tgriesser/knex/blob/master/src/dialects/postgres/index.js#L160

wubzz:
Previously `{searchPath: 'TEST'}` would result in `test` rather than `TEST`.

Fixes #1751

### DIFF
--- a/src/dialects/postgres/index.js
+++ b/src/dialects/postgres/index.js
@@ -157,7 +157,7 @@ assign(Client_PG.prototype, {
     if (!path) return Promise.resolve(true);
 
     return new Promise(function(resolver, rejecter) {
-      connection.query(`set search_path to ${path}`, function(err) {
+      connection.query(`set search_path to "${path}"`, function(err) {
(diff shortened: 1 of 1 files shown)


smithamax:
This seems to have broken multiple search paths. `'knex,public'` becomes `set search_path to "knex,public"` but should be `set search_path to "knex","public"` I think.

wubzz:
That makes sense, didn't have a single thought relating two multiple search paths. Can you open a new issue?

Also I think the syntax in that case should be `['knex', 'public']` instead of `knex,public`.

wubzz:
Issue #2335

PR #2192, opened 2017-08-17
set long stack traces when debug is enabled in the config

capaj:


elhigu:
I'm not sure if this is a good idea either to explicitly set it always when `{ debug: true }`. I don't feel it necessary, it feels a bit wrong to make this kind of opinionated behaviour in `knex`, but it makes sense and I don't believe it causes problems either.

I have to leave this to other people to decide if this should be merged (after testing that it works + mention in docs about it).

capaj:
@elhigu keep in mind that this does not apply for knex as a library, but for knex as cli tool. When you're running migrations in debug mode you want long stack traces in 99% of cases. People can always force to not use them with the env var set to `0`. I can hardly imagine a case when you're running a migration in debug mode and you'd get annoyed by it being slightly slower. That's the purpose of debug mode-to give you as much logging as possible. Nice long stack trace is quite an important part of it.

rapzo:
I actually find debug mode a bit rugged compared to using `DEBUG=knex*`.
But isn't `JSON.parse` redundant there? Even if the config file is a json, `require`ing takes care of it.

capaj:
@rapzo well this would make debug mode slightly more powerful.
I used JSON.parse because the above line assigns the config from command line arguments, which are always string I presume, so I cast it to Boolean. I could use == but I think this is more readable. Unfortunately there is no better way to parse a string into boolean in JS.

tjwebb:
@capaj doesn't `JSON.parse` need to be inside a try/catch?

kibertoad:
@elhigu What is missing here? try-catch block?

capaj:
I'll remove the parse and change it to explicit condition.

Issue #2194, opened 2017-08-21
Add "NOT ILIKE" for Postgres


rfink:
ILIKE exists, but "NOT ILIKE" is absent. PR coming shortly.

elhigu:
via #2195

PR #2195, opened 2017-08-21
Adding NOT ILIKE for postgres

rfink:
Addresses #2194


rfink:
Thanks! Great point about the docs.

Issue #1633, opened 2016-08-18
Support for ON CONFLICT clause for INSERTs


qcom:
For now I achieved this by constructing a set of parenthesized  `VALUES` terms and calling `knex.raw()` with the built string and array of params:

``` javascript
function insertRecordsOnConflictDoNothing(table, records) {
  const keys = Object.keys(records[0]);

  const getIdentifierSpots = () => keys.map(() => '??').join(', ');
  const getValueSpots = () => keys.map(() => '?').join(', ');

  const columns = `(${getIdentifierSpots()})`;
  const values = records.map(() => `(${getValueSpots()})`).join(', ');

  const valueArgs = records
    .map(record => keys.map(key => record[key]))
    .reduce((array, record) => [...array, ...record], []);

  const rawString = `${columns} VALUES ${values} ON CONFLICT DO NOTHING`;
  const args = [...keys, ...valueArgs];

  return knex(table)
    .insert(knex.raw(rawString, args));
}
```

Obviously this is suboptimal but I'm not too familiar with the knex api. Is there a better method to get this behavior currently?

If this is a desired feature, let me know if I can help implement.

VividVisions:
👍

Industrial:
Would love to see this.

mrdulin:
Two years, any update?

elhigu:
@mrdulin yes. API for this feature was designed year ago https://github.com/tgriesser/knex/pull/2197

Please feel free to implement. Only thing missing for this feature is someone who wants to have this enough to actually make it happen.

master-atul:
I ended up using this raw query like this so : 
```js
const result = await knex('accounts').raw(
      knex('accounts')
      .insert(accounts)
      .toQuery() +
      ` ON CONFLICT (account_number, user_id)
          DO UPDATE SET
          available_credit = excluded.available_credit,
          balance = excluded.balance,
          type = excluded.type,
          name = excluded.name,
          metadata = excluded.metadata,
          updated_at = excluded.updated_at
        RETURNING *;`,
  );
```

elhigu:
@master-atul that is vulnerable to SQL injections. With that approach you should write it like this:

```
const result = await knex.raw(
  `? ON CONFLICT (account_number, user_id)
          DO UPDATE SET
          available_credit = excluded.available_credit,
          balance = excluded.balance,
          type = excluded.type,
          name = excluded.name,
          metadata = excluded.metadata,
          updated_at = excluded.updated_at
[...]

Issue #1351, opened 2016-04-15
Query Builder MERGE aka UPSERT


willfarrell:
The ability to uniformly apply and UPSERT to multiple languages.

Great summary of the different SQL implementations: https://en.wikipedia.org/wiki/Merge_(SQL)

Possible Syntax: `knex().insert({...}).upsert([keys]).update({...})`

rhys-vdw:
I'd like to see a discussion of the API, but `upsert` support would be great.

rhys-vdw:
Duplicate of #54

Issue #1121, opened 2016-01-02
Suggestion on PostgreSQL INSERT ON CONFLICT (Upsert)


ozum:
Hi,

Currently I'm using `knex.raw` for `INSERT ON CONFLICT` queries:

``` js
knex.raw('INSERT INTO table (id) VALUES (?) ON CONFLICT (id) DO UPDATE SET id = EXCLUDED.id RETURNING id', [id])
    .then((result) => {});
```

I want to benefit `knex.insert` and other methods at least part of the query for this type queries. Is there a better way to build this query using any combination of knex.select / knex.raw?

RubenCordeiro:
I think that knex.raw is still the best way to write Upsert statements with KnexJS. Check: the issue https://github.com/tgriesser/knex/issues/54.

hayeah:
I put together a monkey patch to experiment with upsert. See:

https://github.com/tgriesser/knex/issues/54#issuecomment-183838939

It works like this:

```js
knex("oauths").insert(oauth).onConflict(["provider", "uid"],{
  info: profile,
  updated_at: new Date(),
});
```

plurch:
I wrote a relatively simple function which works for my use cases. It uses `knex.insert` and `toString()`

https://gist.github.com/plurch/118721c2216f77640232

ivoneijr:
Nice Suggestion!

duncanfinney:
@plurch Very cool. Thank you.

elhigu:
closing as duplicate https://github.com/tgriesser/knex/issues/54

ron-liu:
I tried the solution @plurch  mentioned. All good except it didn't escape `?` properly.  Check the below case:

*Given the table definition:*
```
create table abc (
	name varchar(100) PRIMARY KEY
)
```

*And the below code:*
```
  const insertSql = knex('abc')
    .insert({ name: '? - \\? - \\? - \\\\? # /? - //? - ///? - ////?' })
    .toString()
  console.log(insertSql)

  await knex.raw(insertSql)
```
*It ends up with the below in the db:* 
`$1 - $2 - $3 - $4 # /$5 - //$6 - ///$7 - ////$8`

My working environment is:
- macbook
- postgres running on docker container: `postgres:12-alpine`

ron-liu:
I end up using the old school way as below: 

```
const main = async () => {
  await knex('abc')
    .insert({ name: 'a' })
    .catch(x => {
      if (!isUniqueViolationError(x)) throw x
[...]

Issue #54, opened 2013-08-30
Adding upsert type capability


tgriesser:
Brought up by @adamscybot in tgriesser/bookshelf#55 - this could be a nice feature to add.

clark800:
:+1:

diegoitaliait:
Hi, 
there are news about this new feature?

Or can someone recommend examples, that show how to simulate this functionality for mysql?

thx

tgriesser:
Right now I'm doing it with `raw`, but I'm working hard on getting this available in here soon.

olalonde:
Postgres just implemented upsert support by the way :+1: 

http://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=168d5805e4c08bed7b95d351bf097cff7c07dd65

https://news.ycombinator.com/item?id=9509870

julianlloyd:
I think this would make a great addition to Knex. :+1:

hotaru355:
I was looking for a way to do a `REPLACE INTO` in MySql and found this feature request. Since `REPLACE` and `INSERT` have exactly the same syntax in MySql I would imagine it to be easier to implement than a `ON DUPLICATE KEY UPDATE`. Are there any plans to implement a `REPLACE` ? Would a PR be something of value?

bradsimantel:
Any updates on this, especially with PostreSQL 9.5?

RubenCordeiro:
I think one important question is whether or not to expose the same ```upsert``` method signature for different dialects, such as PostgreSQL and MySQL. In Sequelize, a issue has been raised regarding the return value of ```upsert``: https://github.com/sequelize/sequelize/issues/3354. 

I realise that some of KnexJS library methods have distinctions regarding the return values in the context of different dialects (such as ```insert```, where an array of the first inserted id is returned for Sqlite and MySQL, while an array of **all** the inserted id's is returned with PostgreSQL).

According to the documentation, the ```INSERT ... ON DUPLICATE KEY UPDATE``` syntax in MySQL has the following behaviour (http://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html):

> With ON DUPLICATE KEY UPDATE, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values.

While in PostgreSQL (http://www.postgresql.org/docs/9.5/static/sql-insert.html):

> On successful completion, an INSERT command returns a command tag of the form

[...]

PR #2197, opened 2017-08-21
[WIP] INSERT ... ON CONFLICT support for postgres

jordansexton:
This is based in part on [this gist](https://gist.github.com/hayeah/1c8d642df5cfeabc2a5b) by @hayeah, and in part on the existing update method implementation.

It would affect #1633, #1351, #1121, and #54. I wrote preliminary tests for it, which I'm using the CI to run, but I'd like feedback on the them (the test files are rather hard to parse).

I don't think any consensus has been reached on a more generalized "upsert" feature, but I think supporting this in a database-specific way is reasonable and makes the API easy to understand.

Asking @elhigu (because you had input on #54), do you think a Postgres-specific approach is acceptable, or would this need to be expanded to support all clients that have some kind of `MERGE`/`ON DUPLICATE KEY UPDATE` functionality?

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -853,6 +853,15 @@ assign(Builder.prototype, {
     return this;
   },
 
+  // Insert on conflict do nothing/update
+  onConflict(columns, updates) {
+    this._single.onConflict = {
+      columns,
+      updates
+    };
+    return this;
+  },
+
   // Delete
   // ------
 
--- a/test/integration/builder/inserts.js
+++ b/test/integration/builder/inserts.js
@@ -648,6 +648,122 @@ module.exports = function(knex) {
       });
     });
 
+    it('should allow insert on conflict do nothing in postgres', function() {
+
+      return knex('accounts')
+        .insert({
+          email:'test7@example.com',
+          first_name: 'User',
+          last_name: 'Test'
+        }, 'id')
+        .onConflict('email', {
+          first_name: 'Test',
+          last_name: 'User'
+        })
+        .testSql(function(tester) {
+          tester(
+            'mysql',
+            'insert into `accounts` (`email`, `first_name`, `last_name`) values (?, ?, ?)',
+            ['test7@example.com','Test','User'],
+            [8]
+          );
+          tester(
+            'postgresql',
+            'insert into "accounts" ("email", "first_name", "last_name") values (?, ?, ?) on conflict ("email") do nothing returning "id"',
+            ['test7@example.com','Test','User'],
+            ['8']
+          );
+          tester(
+            'sqlite3',
+            'insert into `accounts` (`email`, `first_name`, `last_name`) values (?, ?, ?)',
+            ['test7@example.com','Test','User'],
+            [7]
+          );
+          tester(
+            'oracle',
+            "insert into \"accounts\" (\"email\", \"first_name\", \"last_name\") values (?, ?, ?) returning ROWID into ?",
+            ['test7@example.com','Test','User', function (v) {return v.toString() === '[object ReturningHelper:id]';}],
+            [8]
+          );
+          tester(
+            'mssql',
+            'insert into [accounts] ([email], [first_name], [last_name]) output inserted.[id] values (?, ?, ?)',
+            ['test7@example.com','Test','User'],
+            ['8']
+          );
+        });
+
+    });
+
+    it('should allow insert on conflict do update in postgres', function() {
+
+      return knex('accounts')
+        .insert({
+          email:'test8@example.com',
+          first_name: 'User',
+          last_name: 'Test'
+        }, 'id')
+        .onConflict('email', {
+          first_name: 'Test',
+          last_name: 'User'
+        })
+        .testSql(function(tester) {
+          tester(
+            'mysql',
+            'insert into `accounts` (`email`, `first_name`, `last_name`) values (?, ?, ?)',
+            ['test8@example.com','Test','User'],
+            [9]
+          );
+          tester(
+            'postgresql',
+            'insert into "accounts" ("email", "first_name", "last_name") values (?, ?, ?) on conflict ("email") do update set "first_name" = \'Test\', "last_name" = \'User\' returning "id"',
+            ['test8@example.com','Test','User'],
+            ['9']
+          );
+          tester(
+            'sqlite3',
+            'insert into `accounts` (`email`, `first_name`, `last_name`) values (?, ?, ?)',
+            ['test8@example.com','Test','User'],
+            [8]
+          );
+          tester(
+            'oracle',
+            "insert into \"accounts\" (\"email\", \"first_name\", \"last_name\") values (?, ?, ?) returning ROWID into ?",
+            ['test8@example.com','Test','User', function (v) {return v.toString() === '[object ReturningHelper:id]';}],
+            [9]
+          );
+          tester(
+            'mssql',
+            'insert into [accounts] ([email], [first_name], [last_name]) output inserted.[id] values (?, ?, ?)',
+            ['test8@example.com','Test','User'],
+            ['9']
+          );
+        });
+
+    });
+
+    it('should perform update on conflict in postgres and return conflicting row', function() {
+
+      return knex('accounts')
+        .insert({
+          email:'test8@example.com',
+          first_name: 'User',
+          last_name: 'Test'
+        }, 'id')
+        .onConflict('email', {
+          first_name: 'Testing',
+          last_name: 'Users'
+        })
+        .testSql(function(tester) {
+          tester(
+            'postgresql',
+            'insert into "accounts" ("email", "first_name", "last_name") values (?, ?, ?) on conflict ("email") do update set "first_name" = \'Testing\', "last_name" = \'Users\' returning "id"',
+            ['test8@example.com','Test','User'],
+            ['9']
+          );
+        });
+
+    });
 
     describe('batchInsert', function() {
       var dialect = String(knex.client.dialect).toUpperCase();
--- a/test/integration/builder/inserts.js
+++ b/test/integration/builder/inserts.js
@@ -743,26 +743,26 @@ module.exports = function(knex) {
     });
 
     it('should perform update on conflict in postgres and return conflicting row', function() {
-
-      return knex('accounts')
-        .insert({
-          email:'test8@example.com',
-          first_name: 'User',
-          last_name: 'Test'
-        }, 'id')
-        .onConflict('email', {
-          first_name: 'Testing',
-          last_name: 'Users'
-        })
-        .testSql(function(tester) {
-          tester(
-            'postgresql',
-            'insert into "accounts" ("email", "first_name", "last_name") values (?, ?, ?) on conflict ("email") do update set "first_name" = \'Testing\', "last_name" = \'Users\' returning "id"',
-            ['test8@example.com','Test','User'],
-            ['9']
-          );
-        });
-
+      if (knex.client.dialect === 'postgresql') {
+        return knex('accounts')
+          .insert({
+            email:'test8@example.com',
+            first_name: 'User',
+            last_name: 'Test'
+          }, 'id')
+          .onConflict('email', {
+            first_name: 'Testing',
+            last_name: 'Users'
+          })
+          .testSql(function(tester) {
+            tester(
+              'postgresql',
+              'insert into "accounts" ("email", "first_name", "last_name") values (?, ?, ?) on conflict ("email") do update set "first_name" = \'Testing\', "last_name" = \'Users\' returning "id"',
+              ['test8@example.com','Test','User'],
+              ['9']
+            );
+          });
+      }
     });
 
     describe('batchInsert', function() {
--- a/src/dialects/postgres/query/compiler.js
+++ b/src/dialects/postgres/query/compiler.js
@@ -27,9 +27,9 @@ assign(QueryCompiler_PG.prototype, {
   insert() {
     const sql = QueryCompiler.prototype.insert.call(this)
     if (sql === '') return sql;
-    const { returning } = this.single;
+    const { onConflict, returning } = this.single;
     return {
-      sql: sql + this._returning(returning),
+      sql: sql + this._onConflict(onConflict) + this._returning(returning),
       returning
     };
   },
@@ -57,6 +57,22 @@ assign(QueryCompiler_PG.prototype, {
       sql: sql + this._returning(returning),
       returning
     };
+  },
+
+  _onConflict(value) {
+    if (!value) {
+      return '';
+    }
+    const { columns, updates } = value;
+    let sql = ` on conflict (${this.formatter.columnize(columns)}) do `;
+    if (updates) {
+      const updateData = this._prepUpdate(updates);
+      sql += `update set ${updateData.join(', ')}`;
+    }
+    else {
+      sql += 'nothing';
+    }
+    return sql;
   },
 
   _returning(value) {


jordansexton:
Tests appear to be broken on:

```
Selects
      ✓ runs with no conditions
      ✓ returns an array of a single column with `pluck`
      ✓ can pluck a qualified column name, #1619
      ✓ starts selecting at offset
      ✓ returns a single entry with first
      1) allows you to stream
```

Not sure why, nothing relating to streaming would have changed.

elhigu:
I'm running tests again to see if it was some random problem (didn't look like random, since all node versions failed though).

jordansexton:
Thanks! I haven't been able to get the test environment running on my local machine yet, so I haven't been able to debug the failure.

src/query/builder.js
```diff
@@ -853,6 +853,15 @@ assign(Builder.prototype, {
     return this;
   },
 
+  // Insert on conflict do nothing/update
+  onConflict(columns, updates) {
```
elhigu: I think this should also allow passing raw() so that people can also write their own code, what to do in case of conflict (sqlite has quite a different selection than nothin / update set https://sqlite.org/lang_conflict.html).

test/integration/builder/inserts.js
```diff
@@ -648,6 +648,122 @@ module.exports = function(knex) {
...
+          .testSql(function(tester) {
+            tester(
+              'postgresql',
+              'insert into "accounts" ("email", "first_name", "last_name") values (?, ?, ?) on conflict ("email") do update set "first_name" = \'Testing\', "last_name" = \'Users\' returning "id"',
+              ['test8@example.com','Test','User'],
+              ['9']
+            );
+          });
+      }
+    });
```
elhigu: In addition to unit tests which are just generating queries,  there should be real integration test, which actually tries to insert stuff to db and checks that in conflict situation there was correct data added to DB.

elhigu:
@jordansexton I just gave this an initial review how the API and design looks compared to other DBs which may support upsert + basic notes of minimum testing/documentation requirements. Thanks for the pull request though! I'm really happy to see that someone finally started to work on this 😊

elhigu:
> Thanks! I haven't been able to get the test environment running on my local machine yet, so I haven't been able to debug the failure.
--

I will figure out why those tests were failing, I don't believe it has anything to do with this PR.

jordansexton:
Thanks! 👍 All your proposals sound good to me, I'll work on this. Do you have any tips or documentation for how to write unit and integration tests for knex? I just copied the`tester(...)` blocks from elsewhere, but the indirection of the system feels opaque.

jordansexton:
And are you comfortable with calling it `onConflict`? MySQL's version of the same is pretty wordy...

elhigu:
> And are you comfortable with calling it onConflict? MySQL's version of the same is pretty wordy...

Something more generic, which is not exactly the wording of postgresql nor mysql would be better to keep API less opinionated how it is implemented... maybe `onUpsert()` ? sounds a bit weird since end result is `.insert().onUpsert()`, but I can live with that. As far as I know none of the BDs is actually calling their upsert syntax as `upsert` :)

elhigu:
For writing real integration tests, this was pretty clean way to do it (I just browsed back commits until I found this) https://github.com/tgriesser/knex/commit/5fc456cb693a12d789167bb1eeac006edf5f9f17

jordansexton:
Here's a brief rundown of the differing APIs for the PostgreSQL, MySQL, and SQLite:

- Postgres's `ON CONFLICT DO UPDATE` **requires** passing column/index names *if* updates are provided.

> For `ON CONFLICT DO NOTHING`, it is optional to specify a conflict_target; when omitted, conflicts with all usable constraints (and unique indexes) are handled. For `ON CONFLICT DO UPDATE`, a `conflict_target` must be provided.

- MySQL's `ON DUPLICATE KEY UPDATE` doesn't support passing columns at all. The alternative syntax is for doing nothing is `INSERT IGNORE`.

- SQLite's `INSERT OR IGNORE INTO` / `INSERT OR REPLACE INTO` (or just `REPLACE INTO`) also doesn't support passing conflicting columns or indexes.

Arguments against `onUpsert`:

- No supported RDBMS actually calls the operation "upsert", nor is it standard SQL. The SQL standard uses `MERGE` (this is what is implemented by Oracle and MS SQL).
- `insert(...).onUpsert(...)` not only looks confusing, but also ultimately confuses what's happening; makes very little sense in the `IGNORE`/`DO NOTHING` case.
- It sounds like registering an event listener.

Arguments against `onConflict`

- Only Postgres calls it that (technically, SQLite has `ON CONFLICT`, but you just use the shorthand `INSERT OR ...` / `REPLACE INTO` above instead.
- `insert(...).onConflict(...)` itself reads well, but then the parameters passed to `onConflict` aren't consistent between RDBMSs, and even for Postgres, aren't self-documenting. You just have to know what the code you're reading is supposed to do.
- It also sounds like registering an event listener.

In light of the above, I think using a chaining API that better indicates what you want to accomplish might be better.

---

Possible alternatives:

#### 1. `ignore` and `merge`
```javascript
insert({ ...inserts }).ignore([columns]?)
insert({ ...inserts }).merge({ ...updates }, [columns]?)
```
- **Pros**
  - Readable, explicit, no magic.
  - Columns optional for Postgres (or any other DB).
- **Cons**
  - While `MERGE` is in the SQL standard, this _type_ of operation is often referred to as an upsert by devs.
      - `merge` in a chaining API context passing around object literals is questionable, because it most often implies merging objects.
- **Pro or Con**
  - Two methods instead of one. I might consider this a pro, as I tend to prefer an explicit API, even if it means the API surface is larger as a result. Two methods also means it's more easily composed by application code that... builds the query... builder.

#### 2. `ignore` and `replace`
```javascript
insert({ ...inserts }).ignore([columns]?)
insert({ ...inserts }).replace({ ...updates }, [columns]?)
```
- **Pros**
  - Same as above, but `replace` is more contextually meaningful than `merge`.
- **Cons**
  - `REPLACE` itself isn't standard SQL, and is only used by SQLite.

#### 3. `upsert` or `onUpsert`
```javascript
insert({ ...inserts }).upsert({ ...updates }, [columns]?)
insert({ ...inserts }).onUpsert({ ...updates }, [columns]?)
```
- **Pros**
  - Uses `upsert`, which despite not being used by any supported RDBMS, is commonly used by devs.
- **Cons**
  - `insert` followed by `upsert` is pretty weird.
  - `insert` followed by `onUpsert` confuses what is actually happening: you're doing doing something "on upsert" (which itself isn't a thing), but you're doing "an upsert" upon the insert failing.
  - This is what it looks like when you want to `IGNORE`: `insert({ foo: 'bar' }).onUpsert()` or worse, in Postgres, `insert({ foo: 'bar' }).onUpsert(['foo'])`. This is confusing, and I think it's for the reason that when you call `upsert` or `onUpsert` (with or without arguments), it makes it look like something is going to happen, but what you actually want is for nothing to happen.
  - The semantics of `upsert` or `onUpsert` only make sense in the positive case; in the negative case, they look magical (and then we have to document that if the `{ ...updates }` hash isn't passed, you have to pass the columns as the first arg (or we check if it's an array). Ultimately, the semantics of the API don't follow directly from its intentions.

#### 4. `upsert`
```javascript
upsert({ ...inserts }, { ...updates }, [columns]?)
```
- **Pros**
  - One method, one new operation.
  - Semantically makes sense; devs often refer to this operation atomically as such.
- **Cons**
  - A single method hides what's actually going on (in `insert` or an `update`), and doesn't play as nicely with application code that wants to build a query in pieces.
  - Requires replacing existing `inserts` in your code rather than adding calls onto them.
  - With lots of columns, or with raw values, this gets kind of ugly, especially when you have raw values as part of your update (common for incrementing columns on conflict, aggregation, pulling values from other columns, etc.)
  - As with `onUpsert`, magical when you want to `IGNORE`. The API is `upsert({ ...inserts })`, which will do nothing, or in Postgres, `upsert({ ...inserts }, [columns])`.

---

My personal preference is for having two methods, one to ignore/do nothing, and one to perform an update. You would never actually use both, but because you are doing one of two different things, you should invoke methods that semantically match your intent.

Thoughts?

jordansexton:
Hmm, supporting this in SQLite with a simple implementation is dubious.

> When a UNIQUE or PRIMARY KEY constraint violation occurs, the REPLACE algorithm deletes pre-existing rows that are causing the constraint violation prior to inserting or updating the current row and the command continues executing normally. If a NOT NULL constraint violation occurs, the REPLACE conflict resolution replaces the NULL value with the default value for that column, or if the column has no default value, then the ABORT algorithm is used. If a CHECK constraint violation occurs, the REPLACE conflict resolution algorithm always works like ABORT.
~ https://sqlite.org/lang_conflict.html

I understand this to mean that `insert({ ...inserts }).upsert/onUpsert/merge/replace({ ...updates })` is not going to patch with just the updates provided, but is instead going to wipe all the columns not provided to their defaults. This is a pretty big footgun and is not what I would expect as a user of the library.

https://github.com/tgriesser/knex/issues/54#issuecomment-289217529 has some description of it.

This (the fact that it actually deletes the row then inserts a new one, rather than updating the row) suggests to me that a `INSERT OR REPLACE` aka `REPLACE INTO` is a fundamentally different operation, and perhaps should not be supported in the same way.

jordansexton:
#### Problems with SQLite

I think [this SO answer](https://stackoverflow.com/a/418988) is probably the most sound approach, but outside the scope of this feature. I think it's probably best to not support this on SQLite since it's not a safe operation, and the fix options ([1](https://stackoverflow.com/a/7511635), [2](https://stackoverflow.com/a/4330694)) for it would also be very complex/opinionated, and require Knex to know how to select the right row, since SQLite doesn't do this automatically.

The alternative is supporting this in SQLite with just a documentation warning, but that will undoubtedly result in people's data getting deleted if they get their query wrong. I think having the feature poorly or partially implemented is worse than not having it, and it will lead SQLite users to believe that it's equivalent.

#### Problems with `MERGE`

Oracle and MS SQL support the SQL standard `MERGE` operation (or some variant thereof), which takes the form:

```sql
MERGE INTO table_name AS target
USING table_name AS source
    ON target.some_column = source.some_column
    ...
WHEN MATCHED THEN 
    UPDATE SET ...
WHEN NOT MATCHED BY TARGET THEN
    INSERT (...) VALUES (...);
```

The issue here is that the requisite table aliases and condition columns can't be known with any of the API proposals we have so far. We'd have to do something like

```javascript
insert({ ...inserts }).merge({ ...updates }, [conditions]))
```

where `conditions` is maybe one or more of:
- a callback that allows building the merge with a DSL, similar to a join callback (doable, complicated)
- a raw query (problematic because you still need to define the table aliases)
- an array of identifiers to match on (same issue as a raw query)

Alternatively, `merge` needs a `targetAlias` argument as well as `conditions`, which would make raw queries or identifiers doable, but still complicated. There's also the issue of what happens when you want to effectively `IGNORE`/`DO NOTHING`

Ultimately, the API for a SQL standard MERGE needs additional params that can't be provided by this API.

#### Conclusion

I don't think it's useful to implement this in a wholly database-agnostic fashion. The syntax and semantics for MySQL and Postgres are very close; the only difference we care about is the requirement to specify the column(s)/index(es) in Postgres if `DO UPDATE`-ing.

For `MERGE` and `REPLACE INTO`, the functionality is quite different. I think it would ultimately be bad to shoehorn them all into the same API.

elhigu:
Thanks for investigating this throughly!

> Problems with SQLite

I agree. If the implementation would be somewhat hacky better not to support that platform and allow people to decide how to deal with it. 

> My personal preference is for having two methods, one to ignore/do nothing, and one to perform an update. You would never actually use both, but because you are doing one of two different things, you should invoke methods that semantically match your intent.

I think this one is nicer instead of having single method. 

> I don't think it's useful to implement this in a wholly database-agnostic fashion. The syntax and semantics for MySQL and Postgres are very close; the only difference we care about is the requirement to specify the column(s)/index(es) in Postgres if DO UPDATE-ing.

> For MERGE and REPLACE INTO, the functionality is quite different. I think it would ultimately be bad to shoehorn them all into the same API.

I don't believe in it either, but many knex users are actually using knex for abstracting DBs as much as they can, so in that sense it would be convenient. 

It will naturally mean that the DB agnostic way will not be as powerful as separate APIs for each DB. I would like to go with agnostic udpate / ignore way at first and then people who doesn't want to use raw() for having DB specific functionality, they could add also DB specific APIs for certain DBs.

Could simple update on conflict case work somehow like this:
```
knex('blah').insert({id: 1, comment: 'the comment' }).replace({ comment: 'was replaced' }, ['id'])
```

```
MERGE INTO  blah AS target
USING blah AS source
    ON target.id = source.id
WHEN MATCHED THEN 
    UPDATE SET comment = 'was replaced'
WHEN NOT MATCHED BY TARGET THEN
    INSERT (...) VALUES (...);
```

Does this sound viable way to do it with merge? 

You already understand this problem a lot better than I do and all the stuff you've written sounds legit so I trust your judgement on this.  Anyways I would like to emphasis that more generic and less powerful method would probably serve bigger portion of the knex userbase since update/ignore on conflict seems to be the thing that most of the people are looking for.

jordansexton:
@elhigu Sounds good. I will leave SQLite out for now and implement and test on Postgres and Mysql with two methods.

The problem I see with implementing `MERGE` for MS SQL/Oracle is that if we hardcode `source` and `target`, the developer who is doing the operation can't name the aliases, can't have any conflicting aliases for the query, and has to know those are the aliases to refer to them in raw updates.

If this is acceptable to you, I can work on that as well. If not, I will just get PG and MySQL working, and we can revisit this.

elhigu:
@jordansexton good point about the aliases. 

We can hardcode them to something that probably won't collide with users own table names and it can be improved later on to allow overriding default alias names. `mergeSource` and `mergeTarget` could be unique enough?

Its just important to add default names to docuementation with an example usage.

jordansexton:
Okay. I'll finish the implementation for `DO NOTHING/UPDATE` on Postgres and `INSERT IGNORE/ON DUPLICATE KEY UPDATE` on MySQL in this PR, probably using two methods:
```javascript
insert({ ...inserts }).ignore([columns]?)
insert({ ...inserts }).merge({ ...updates }, [columns]?)
```

My reasoning is:

1. `replace` implies full replacement, whereas `merge` implies partial replacement (essentially, the semantics for a generic "object merge" are equivalent to what actually happens in the operation)
2. Only SQLite uses replace. Since this PR won't support SQLite, and replace is a semantically different operation, it's probably better to have a different `replace` method (if supported at all).
3. The SQL standard uses `MERGE`. In the event that other RDBMS's converge upon the standard, `merge` is likely more future-proof than an ad-hoc alternative.
4. A followup PR to support MS SQL and Oracle will be able to use the same method and just work. I don't use either, but I'll take a stab at supporting both when this is settled. The second `columns` argument could be used to provide aliases for those instead.

Druotic:
@jordansexton thanks for your work on this! It'll be great being able to avoid having to use raw queries to get upsert behavior. @elhigu What's left to get an initial iteration of this merged? I see comments about an end-to-end integration test as well as being more db agnostic/supporting additional DBs with this behavior.

elhigu:
@Druotic This initial PR will not be merged, but it needs pretty much complete rework to change the implementation to match APIs described in https://github.com/tgriesser/knex/pull/2197#issuecomment-325934623 and tests  and documentation accordingly.

@jordansexton indeed already did awesome work figuring out good database agnostic APIs to support upserty behavior. 

This is pretty much the last thing that I would like to see finished before releasing knex 0.14

jordansexton:
@elhigu @Druotic Thanks for the ping. I've been pretty busy and spaced on this, but I'll see what I can do about it this weekend. And yes, I plan to implement the API described @ https://github.com/tgriesser/knex/pull/2197#issuecomment-325934623. Any help with the testing is very much appreciated, as I found the test system very inscrutable.

austinkelleher:
I really need this feature. Can't wait to see this get merged. Thanks @jordansexton!

elhigu:
@jordansexton I'm starting to work on release 0.14, if this misses it I can make 0.14.1 when this is ready.

nihiluis:
whats the status? its version 0.14.2 now

elhigu:
@nihiluis Feature is designed, but no one have had time to implement it yet. As soon as its done I will do an additional patch release to get it out (except that I'll be in vacations couple of week before christmas).

tsofist:
Still.. no? 🔍

ivands:
I would really like to see this feature included. Any updates?

elhigu:
Just to keep important info in the end of this thread:

> This initial PR will not be merged, but it needs pretty much complete rework to change the implementation to match APIs described in https://github.com/tgriesser/knex/pull/2197#issuecomment-325934623 and tests  and documentation accordingly.
>
> @jordansexton indeed already did awesome work figuring out good database agnostic APIs to support upserty behavior. 

So if someone is up to implement that API, please do open new PR with implementation! Locking conversation to prevent statusupdateflood, but not closing since this is the only place where DB agnostic upsert API has been discussed throughly.

tgriesser:
Unlocking this as I'd like to take a look at getting this into knex soon. Sorry for the delay everyone

obataku:
how long til it's merged?

elhigu:
@obtaku never, this PR has only plan how the feature should be implemented. There is no code matching that plan... maybe ill just close this and collect all the relevant info to feature request to prevent further confusion.

elhigu:
Created feature request

PR #2199, opened 2017-08-23
Fix connection error propagation when streaming

### LINKED ISSUE #948: Catching stream errors

novemberborn:
If stream() is used without a handler, make sure to emit connection errors on the stream itself.

This might fix #948, but I've only verified the patch when using Postgres.

### DIFF
--- a/src/runner.js
+++ b/src/runner.js
@@ -1,4 +1,4 @@
-import { assign, isArray, noop } from 'lodash'
+import { assign, isArray } from 'lodash'
 import Promise from 'bluebird';
 import * as helpers from './helpers';
 
@@ -80,7 +80,10 @@ assign(Runner.prototype, {
 
     const runner = this;
     const stream = new PassThrough({objectMode: true});
+
+    let hasConnection = false;
     const promise = Promise.using(this.ensureConnection(), function(connection) {
+      hasConnection = true;
       runner.connection = connection;
       const sql = runner.builder.toSQL()
       const err = new Error('The stream may only be used with a single query statement.');
@@ -99,9 +102,13 @@ assign(Runner.prototype, {
       return promise;
     }
 
-    // This promise is unreachable since no handler was given, so noop any
-    // exceptions. Errors should be handled in the stream's 'error' event.
-    promise.catch(noop);
+    // Emit errors on the stream if the error occurred before a connection
+    // could be acquired.
+    // If the connection was acquired, assume the error occured in the client
+    // code and has already been emitted on the stream. Don't emit it twice.
+    promise.catch(function(err) {
+      if (!hasConnection) stream.emit('error', err);
+    });
     return stream;
   },
 
--- a/test/integration/builder/selects.js
+++ b/test/integration/builder/selects.js
@@ -179,19 +179,36 @@ module.exports = function(knex) {
       });
     });
 
-    it('emits error on the stream, if not passed a function, and connecting fails', function(done) {
+    it('emits error on the stream, if not passed a function, and connecting fails', function() {
       var expected = new Error();
       var original = Runner.prototype.ensureConnection;
       Runner.prototype.ensureConnection = function() {
         return Promise.reject(expected);
       };
-      var stream = knex('accounts').stream();
-      stream.on('error', function(actual) {
-        if (actual === expected) {
-          Runner.prototype.ensureConnection = original;
-          done();
-        }
+
+      var restore = () => {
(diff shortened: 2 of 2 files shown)


src/runner.js
```diff
@@ -89,6 +92,17 @@ assign(Runner.prototype, {
         stream.emit('error', err);
       }
       return runner.client.stream(runner.connection, sql, stream, options);
+    }).catch(function(err) {
```
elhigu: if this catch would be in the same position where there was `.catch(noop)` earlier, then the `if (hasHandler) throw err;` line could be removed.

novemberborn:
> package-lock.json probably shouldn't be updated by thiss PR.

Oops! I usually `git add -p` but must have gotten lazy here 😉 

> Tests seems to be failing because of some linting error: 1:27 error 'noop' is defined but never used no-unused-vars

Removed.

---

I've cleaned up the catch and force-pushed to get rid of the `package-lock.json`. I'm not sure where to add a test for this though. I've found the [stream integration tests](https://github.com/tgriesser/knex/blob/d6273ce42f663aa9cff374fef338d35adc25c8f7/test/integration/builder/selects.js#L172:L190) but how would I trigger the connection error?

elhigu:
Maybe connection error could be done by initializing knex instance with bad connection settings? That might work since we need ensureConnection() to throw an error... or maybe monkeypatching ensureConnection() for a moment could work too?

novemberborn:
I must admit the test setup is a bit beyond me. I've pushed a commit which does a terrible monkey-patch. CI passes so… yay? 😄

test/integration/builder/selects.js
```diff
@@ -178,6 +179,21 @@ module.exports = function(knex) {
...
     });
 
+    it('emits error on the stream, if not passed a function, and connecting fails', function(done) {
+      var expected = new Error();
+      var original = Runner.prototype.ensureConnection;
+      Runner.prototype.ensureConnection = function() {
+        return Promise.reject(expected);
+      };
+      var stream = knex('accounts').stream();
+      stream.on('error', function(actual) {
```
elhigu: maybe this could be wrapped to 

```
return new Promise((resolve, reject) => { ... });
```

which will reject after 5ms or so unless error was emitted? That way test won't hang until mocha timeout and cleaning up could also restore the `Runner.prototype.ensureConnection = original;` even when test fails?

novemberborn:
Just wanted to say I haven't forgotten about this. I ran into this in a client project, which I'm busy shipping at the moment. Will try and fix up the PR when that's done.

novemberborn:
I've pushed a commit, hopefully it passes CI!

test/integration/builder/selects.js
```diff
@@ -178,6 +179,38 @@ module.exports = function(knex) {
...
+
+          if (actual === expected) {
+            resolve();
+          } else {
+            reject(new Error('Stream emitted unexpected error'));
+          }
+        });
+      });
+
+      promise.then(restore, restore);
```
elhigu: doesn't this catch rejected promise, so even if promise is rejected test will pass?

novemberborn: No, since the original promise is returned on the next line. This is a more manual [`Promise.prototype.finally`](https://github.com/tc39/proposal-promise-finally).

elhigu: Ok, so test just returns slightly before function is restored. I think it is fine though. Very unlikely to cause any problems.  Is there reason not to use `.finally` ?

novemberborn: The sequencing is such that the restore happens before the test finishes, because Mocha will attach the callbacks to the returned promise *after* the restore callbacks.

`.finally` isn't supported yet in any Node.js version.

elhigu:
All good now, thank you!

Issue #2084, opened 2017-05-27
Allow to customize how identifiers must be wrapped


koistya:
This would unlock scenarios such as automatically converting **camelCase** to **snake_case** and vise verse. See https://medium.com/p/956357872fe4

It could look something like this (`db.js`):

```js
import knex from 'knex'

function wrapIdentifier(type, value) {
  if (type === 'as') return ....;
  if (type === 'select') return ....;
  return ...;
}

export default knex({
  client: 'pg',
  connection: process.env.DATABASE_URL,
  wrapIdentifier,
});
```

Please vote up if you're planning to use this feature, or vote down if you think it shouldn't be in the library.

Ref #929

elhigu:
Hi, could you explain a bit more why type is necessary argument in wrapIdentifier? 

Also I suppose calling this function would be done as a hook as a first step, before  dialect specific wrapIndentifier function.

This would be nice feature to have,  since currently every knex based ORM has been struggling how to modify each queried column names.

stevenvachon:
Or perhaps just a `transform` function. For setting, we could perform a `toSnakeCase` and for getting we could perform a `toCamelCase`. My code looks rather unpleasant having to do this manually.

elhigu:
@stevenvachon yes something like that, I would like to have more specific name that tells to which tells to which part transformation applies. for example `transformIdentifier` and `transformReadColumnNames` (horrible like that), but something like that which tells to which part of knex the function applies :)

stevenvachon:
perhaps a non-flat structure:
```js
transform: {
  identifier: identifier => {},
  readColumnName: name => {}
}
```

stevenvachon:
@elhigu ?

elhigu:
@stevenvachon I don't see anything to comment here that I didn't already say in my last comment. Those names are as good as my examples, cannot say more without reading some code.

stevenvachon:
Ok, what are all the necessary names?

elhigu:
I implemented + documented how to hook to the identifier wrapping part #2217  and https://github.com/knex/documentation/pull/51

elhigu:
This will be released in knex 0.14

sdwrage:
@elhigu Is there an estimated date for knex 0.14 to be released?

elhigu:
before christmas, maybe this month, maybe next

sdwrage:
@elhigu Awesome. Thanks for the update :)

stevenvachon:
[...]

Issue #929, opened 2015-08-05
convert column name format during migration


jamesdixon:
Hi,

In Bookshelf, I see that it's possible to override `parse()` and `format()` if you want to convert from `snake_case` to `camelCase` and vice-versa when fetching or saving. Is there a similar method to do so in Knex? I'm running migrations where my table columns are `snake_case`, but I'm passing in an object where the properties are `camelCase`. I do realize that I can just change the property names to `snake_case` by wrapping it in quotes, but was hoping for something that I may override as can be done in Bookshelf.

Thank you!
James

timruffles:
:+1:

elhigu:
There is currently no support for this in `knex`. Since there are already wrapping methods for column identifiers it might even be quite easy to implement in a way that it would work everywhere, in queries etc.

Though I'm not sure if this would be out of scope... To me I don't mind having that feature. Actually it would be pretty useful also for our project, which is based on `knex`.

prashaantt:
Is there a better solution to this yet, other than manually transforming the object properties between camel and snake case every time? I'm using ES6 classes to model my tables and everything would work beautifully but for this.

elhigu:
@prashaantt no there is currently no other way and it gets even more nasty than just converting them before inserting and after fetching rows. You would also need to convert all the column names passed to query's where clause. 

We ended up just using camelCase column names in our DB.

prashaantt:
> We ended up just using camelCase column names in our DB.

So I suppose now you quote all your column names when writing raw SQL queries, right? Have you found it to be a fair compromise for most practical purposes?

elhigu:
Knex does quoting itself e.g. `knex.raw('?? > ?', ['MyTable.myCol', 1])`

prashaantt:
Yes, raw SQL in the psql CLI is what I meant. Thanks for your insight.

elhigu:
Closing in favor of #2084 which has some suggestion for initial implementation.

joelmukuthu:
Hi all, sorry to plug but I'm working on an ORM for knex which resolves this issue, PTAL https://github.com/joelmukuthu/knorm

elhigu:
@joelmukuthu looks like it wraps pretty big part of knex API and limits it to use only `.where({ obj.. })` style of queries. Does it allow using `raw()` at all?

joelmukuthu:
[...]

PR #2217, opened 2017-09-07
Configuration option to allow override/hook to identifier wrapping code

elhigu:
First part of https://github.com/tgriesser/knex/issues/2084 and https://github.com/tgriesser/knex/issues/929 feature, which allows to write custom identifier wrapper methods. 

So effectively you can describe function how each identifier name are treated, like for example id replacement in `knex.raw('??', ['MyTable.myCol'])` will behave. This is pretty nice feature for ORMs which would like to allow for example different naming scheme for database tables/columns and on javascript side.

Documentation PR is here https://github.com/knex/documentation/pull/51

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -158,6 +158,13 @@ assign(Client.prototype, {
   },
 
   wrapIdentifier(value) {
+    if (this.config.wrapIdentifier) {
+      return this.config.wrapIdentifier(value, this.wrapIdentifierImpl);
+    }
+    return this.wrapIdentifierImpl(value);
+  },
+
+  wrapIdentifierImpl(value) {
     return (value !== '*' ? `"${value.replace(/"/g, '""')}"` : '*')
   },
 
--- a/src/dialects/mssql/index.js
+++ b/src/dialects/mssql/index.js
@@ -66,7 +66,7 @@ assign(Client_MSSQL.prototype, {
     return new ColumnCompiler(this, ...arguments)
   },
 
-  wrapIdentifier(value) {
+  wrapIdentifierImpl(value) {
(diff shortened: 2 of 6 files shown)


stevenvachon:
By "first part", are you referring to the transformation of identifiers with no concept of direction? The code appears to always convert to snake_case, instead of to that *or* camelCase.

elhigu:
@stevenvachon by the "first part" I mean converting identifiers and with "second part" I mean converting column names of the result rows read from database. Did you check the documentation pull request, which explains what the method does? If you have suggestions how to improve that all the comments are appreciated.

stevenvachon:
The documentation is in a script file which doesn't make for the clearest to comprehension. As for the snake_case -> camelCase, I was referring to the test file.

elhigu:
>  As for the snake_case -> camelCase, I was referring to the test file (which instead concatenates).

Yes. That is very much intentional. It just tries that modifying and calling callback actually works.

elhigu:
@wubzz @rhys-vdw @tgriesser @koistya any comments on this one?

elhigu:
I'll still check if I'm able to add info, which part of the identifier is being transformed. Somewhat the same way that @koistya mentioned in issue #2084.

sdwrage:
Is there an estimated timeline for when a pull request will be made to the main knex repo for the last half of this feature? In other words, when will the full implementation be available for use?

elhigu:
@sdwrage do you mean when this one is released or if there is going to be also hook for modifying column names read from database? Anyway, both features will be released in next release knex 0.14.0 hopefully next month. There are some other PRs still which I would like to get merged before it.

sdwrage:
@elhigu both and thank you for the answer :)

Issue #1755, opened 2016-10-24
With Recursive (Common Table Expressions)


coockoo:
I've found `.with` method.
http://knexjs.org/#Builder-with
Can `.withRecursive` method be added?

jordansexton:
I have a work in progress PR for this: https://github.com/tgriesser/knex/pull/2218

Feedback welcome!

PR #2218, opened 2017-09-08
[WIP] WITH RECURSIVE support

jordansexton:
New feature and fix for #1755

I'm testing this out and haven't settled on the API yet, so I'm opening this for feedback.

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -97,6 +97,18 @@ assign(Builder.prototype, {
     return this;
   },
 
+  // Adds a raw `with` clause to the query.
+  withRaw(alias, sql, bindings) {
+    const raw = (sql instanceof Raw ? sql : this.client.raw(sql, bindings));
+    this._statements.push({
+      grouping: 'with',
+      type: 'withRaw',
+      alias: alias,
+      value: raw
+    });
+    return this;
+  },
+
   // Helper for compiling any advanced `with` queries.
   withWrapped(alias, callback) {
     this._statements.push({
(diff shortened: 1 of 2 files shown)


elhigu:
Thanks for the PR!

Btw. there is currently also couple of bugs in the original `.with` statement. For example

```
.with('name', knex('table').update({ a: 1 })); // doesn't work, one has to use knex.raw() for a workaround
.with('name', knex.raw('??', [knex('table').update({ a: 1 })])); // this works
.with('name', builder => builder.table('table').update({ a: 1 })) // creates some invalid select query instead of update
```

I would be super happy if those could be fixed too with this PR 👍 

About this PR I like the API, but with brief look implementation did seem pretty much copy-paste. Could there be more codesharing between normal `.with` and `.withRecursive`? 

And could those dialect, that doesn't support `.withRecursive` throw an error? 

Generally I think this looks good and is very welcome feature!

jordansexton:
Hmm, I'm not sure I'm the best candidate to roll CTE updates into this. I don't use them for anything, and I'm not familiar with support for them or their issues (within Knex or any RDBMS).

A cursory reading of the Postgres docs suggests to me that implementation to support them would be complicated:

> Recursive self-references in data-modifying statements are not allowed.
> ~ https://www.postgresql.org/docs/9.6/static/queries-with.html

I want to limit the scope of this to allowing recursive selects in RDBMSs that support them, so I'll research that.

And, you're right that this PR is totally a copy-paste job at the moment :)
I'll refactor.

One thing that this PR doesn't support yet is the use of recursive functions like
```
WITH RECURSIVE foo(bar, baz) AS (...)
```
because it wraps the CTE alias e.g. `"foo(bar, baz)"` which will throw an error. Depending on how the implementation is changed to support them, it may diverge from plain `WITH` a bit.

elhigu:
I'll be fixing the bugs I mentioned https://github.com/tgriesser/knex/pull/2218#issuecomment-328061292, so only some refactoring and documentation would be needed for this.

elhigu:
@timhuff if you can take over finishing this it would be great 👍

kibertoad:
@timhuff Any chance you could still finish this?

timhuff:
@kibertoad Unfortunately my schedule doesn't currently permit me to work on this. Perhaps here in a few weeks but for the moment I'm booked solid.

kibertoad:
I'll refactor and document this.

kibertoad:
Superseded by https://github.com/tgriesser/knex/pull/2889

Issue #1952, opened 2017-03-08
knex violates promises contract on del()


jmilkiewicz:
Hey

I think I found a pretty nasty bug regarding promises changing its value when using del().
I am not sure if other operations (like update) and/or other database engines are affected, but for sure this problem on  Postgre  kiasaki/alpine-postgres  and knex v0.12.7.

The problem is:
1  insert a singe row to db:
`db('documents').insert({customer_number:1, contract_number:'b',document_number:222,contract_number_prefix:'ab'}).then(console.log)`
2. Delete via del()
`var res = db('documents').del();`
3. resolve returned promise more than once:
`res.then(console.log)` returns 1  (OK)
`res.then(console.log)` returns 0 (Incorrect)
`res.then(console.log)` returns 0 (Incorrect)
`res.then(console.log)` returns 0 (Incorrect)

Why a subsequent resolving promise returns different value - first 1 and then 0 ?

wubzz:
@jmilkiewicz Knex does not create a "single promise" when you use it like in your example. `res` is still just a query builder, not a promise. Everytime you call `.then` on res, a real  DELETE query is being run on the database. The first time you run it the previously inserted row is deleted, so the resolved result is 1, as in one row deleted. The next few times you run it, there are no more rows to be deleted, which is why the resolved result is 0.

`.del()` always returns the number of affected rows. So this is correct behaviour.

jmilkiewicz:
@wubzz 
So `res` having  a promise like interface (`then` method) is not a promise but is only a query builder ?  Do i get it right ?

wubzz:
Yes, that is one way to put it. `.then` executes the query builder.

elhigu:
@jmilkiewicz any object with `.then()` method in Promise A+ spec is called `thenable`. As far as I remember specs doesn't say that calling `.then()` twice for thenable always  must return the same value (would be nice though, but also would prevent to trigger the same query builder multiple times).

jmilkiewicz:
@elhigu 
If you claim that del returns a promise than you are clearly violating promise contract. Look at https://promisesaplus.com/#requirements - it is clearly stated that value/reason MUST not change when promise is fulfilled/rejected. So then/catch shall be free of side-effects - no SQL delete shall be executed if you chain it with "then".
IMHO documentation shall be updated to state that del (i do not know if other methods suffer from the same problem) returns "thenable" which is NOT a promise.

elhigu:
[...]

Issue #1105, opened 2015-12-21
not full promise interface (queries cannot be used in chains)


alex-milanov:
I am trying to chain couple of queries one after another - create table, insert, select ...

I want to create something like this:

```js
createTable()
  .then(insertMockUsers)
  .then(selectUsers)
```

The problem is the .then interface is not acting like a promise interface.

Currently I can do this:
```js
var selectUsers = knex.select().from("users");
selectUsers.then(...)
```
`knex.select().from("users")`  creates knex query object and 
`.then` both executes the query itself and allows us to chain things afterwards. 
Looks like a promise interface, but it's not. 
You can only chain things afterwards but not the action `selectUsers` itself as it is not a promise.

My current solution is a `promisifyKnexQuery` function:
```js
function promisifyKnexQuery(knexQuery){
  return function(){
    var deferred = Promise.pending();

    knexQuery.then(function(result){
      deferred.resolve(result);
    },function(err){
      deferred.reject(err);
    })

    return deferred.promise;
  }
}
```

which can be used like this:
```js
var insertUsers = promisifyKnexQuery( knex("users").insert(mockUsers) );

createTable().then(insertUsers).then(selectUsers)

```

The solution for a proper promise interface is to implement a `toPromise()` method. Which would return a promise function that can be used both to start promise chain and be used inside a promise chain.

```js
var insertUsers = knex("users").insert(mockUsers).toPromise();
```

tgriesser:
Yep, I'm going to fix this before 1.0 by making the backing data for the query chain immutable and throwing if the query is mutated after the promise is executed. Otherwise you'll want to use the `.clone` method to re-use query partials.

tgriesser:
See also #966

rhys-vdw:
Workaround:

```js
var selectUsers = () => knex.select().from("users");

createTable()
  .then(insertMockUsers)
  .then(selectUsers)
```

raijinsetsu:
The promise-like functionality used by transactions is even worse and leads to a hang in my TypeScript application.  Here is an example:

```
let session = {
    commit = (txn: Knex.Transaction) => Promise<void> {
        if (!this._commitPromise) this._commitPromise = txn.table('session').insert(...);
        return this._commitPromise;
    }
};

let token = {
    session: session,
    commit: (txn: Knex.Transaction) => Promise<void> {
        if (!this._commitPromise) {
            this._commitPromise = this.session.commit(txn)
                .then(() => {
[...]

Issue #470, opened 2014-09-09
Transaction API misleading


timruffles:
First off thanks for Knex! :)

The transaction API has a promise-like API (`then()`, `catch()`) that suggests it's exposing a promise for the eventual committal or rollback of the transaction. Instead, however, it's a stateful API with nothing to do with promises - every time you call it [something happens](https://github.com/tgriesser/knex/blob/master/lib/transaction.js#L77) and a different promise is returned!

If you consider the root idea of promises it is that they represent a future value, or a failure to provide that value. Calls to `then()` are _entirely divorced from time_ - e.g you should be able to call `then()` any number of times, before, after, and during the asynchronous operation it represents, and get the same result every time.

I'd be happy to contribute to fix up this aspect of the API. I just wasted a heap of time because something that looked like a promise wasn't one.

timruffles:
@tgriesser yes - that's precisely where I'd see it being a problem. e.g:

```javascript
var active = knex('users').where('deleted_at IS NULL');

exports.active = function() { return active.then() };
exports.online = function() { return active.where('online', true).then() };
```
Also, thinking about it, you'd need a `exec()` to allow reuse of queries via the promise API.  `now` and `later` will be the same value if promises are cached, and no `exec()` API exists.

```javascript
var recentEvents = knex("projects").where("created_at > NOW() - 1000");

var now = recentEvents.then();
var later = timeout(2000).then(function() { return recentEvents.then() });
```

Another option to give builder/result distinction and retain the inline style: change `Result#then()` so if it is resolved with a query builder it will run the query and return a promise for its result.

bruun:
> I just wasted a heap of time because something that looked like a promise wasn't one.

You are not alone. A `.then` in promise-land shouldn't have side-effects, but the knex API does.

baltar:
What's the take on this now? I'm all for an explicit run() method. I use bookshelfjs and avoid direct use of knex, but I just stumbled across a a case where that non-promise then() hit me. I really don't think that preventing a minimal amount of verbosity is enough justification to allow such confusing and in some cases dangerous behaviour.

elhigu:
@baltar  As far as I know consensus was that `QueryBuilder` could be changed to not be `thenable` and has separate method to convert it to promise. 

But this will be really big change so probably not coming any time soon. 

[...]

PR #2226, opened 2017-09-14
Fix multiple thens running query builder multiple times

MellowMelon:
The knex documentation has the following under the `then` method (emphasis mine):
> Coerces the current query builder chain into a promise state, accepting the resolve and reject handlers as specified by the Promises/A+ spec. As stated in the spec, more than one call to the then method for the current query chain will resolve with the same value, in the order they were called; **the query will not be executed multiple times**.

This PR adds a test showing that the bolded statement is not correctly implemented by knex and fixes the bug.

### DIFF
--- a/src/interface.js
+++ b/src/interface.js
@@ -14,8 +14,10 @@ export default function(Target) {
 
   // Create a new instance of the `Runner`, passing in the current object.
   Target.prototype.then = function(/* onFulfilled, onRejected */) {
-    const result = this.client.runner(this).run()
-    return result.then.apply(result, arguments);
+    if (!this._runInstance) {
+      this._runInstance = this.client.runner(this).run();
+    }
+    return this._runInstance.then.apply(this._runInstance, arguments);
   };
 
   // Add additional "options" to the builder. Typically used for client specific
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -19,6 +19,24 @@ module.exports = function(knex) {
       });
     });
 
+    it('should run the query once after calling the .then() function', function() {
+      var p = knex('accounts').insert({
+        first_name: 'Test',
+        last_name: 'Multithen',
+        email:'test@example.com',
+      }, 'id');
+
+      return Promise.resolve().then(function(){
+        return p;
+      }).then(function(){
+        return p;
+      }).then(function(){
+        return knex('accounts').where('last_name', 'Multithen').select('*');
+      }).then(function(multithenAccounts){
+        expect(multithenAccounts.length == 1);
+      });
+    });
+
     it('should forward the .mapSeries() function from bluebird', function() {
       var asyncTask = function(){
         return new Promise(function(resolve, reject){


wubzz:
I never took note to this part of the docs before, but I don't like the sound of it, regardless of Promise spec and whatnot. To me this sounds like the example below will keep returning the same two rows from the initial `.then` call, or am I misunderstanding completely?

```javascript
const queryBuilder = knex('accounts').select();

return queryBuilder
.then((result) => {
    console.log(result.length); // 2
    return queryBuilder.where('userid', '1').select(); // Add where clause
})
.then((result) => {
    console.log(result.length); // 2 again ??
});
```

I'd expect two different queries to be run in this example.

elhigu:
Looks like the documentation is wrong at least and IIRC Promise A+ spec doesn't require that thenable returns always the same result, when its resolved multiple times. Here are some links where this has been discussed earlier:

https://github.com/tgriesser/knex/issues/1952
https://github.com/knex/documentation/issues/27
https://github.com/tgriesser/knex/issues/470
https://github.com/tgriesser/knex/issues/1105

I had also missed this part of docs and indeed it would mean that one couldn't anymore trigger the same query multiple times, by calling `.then` again, but one would have to do something like: query.clone().then() to be able to reuse the same query builder for triggering multiple queries.

It would mean also, that all query builder's query building methods should start to throw an exception if one tries to add stuff to query after it has been triggered once.

This is so fundamental change to how knex is working currently, that it should first be discussed through and see what are up and down sides of each functionality in different use cases.

I'm not completely against this, but for now I would rather just fix the documentation with examples how to trigger the same query multiple times and how to actually just store promise, which will return the same results.

Current way supports for example:

```js
const result = await query;
const refreshedResult await query;
```

With the change one would have to do:
```js
const result = await query.clone();
const result2 = await query.clone();
```

With current system if one likes to store resulted promise and resolve that multiple times without triggering the query again, he can do
```js
// this kind of example could be added to docs
const resultPromise = query.then(res => res); // or any other mapping to returned data
resultPromise.then(res => {
  // query was not triggered again
});
```

ps. Sorry I didn't have much time to write this or think this very clearly, so I might need to edit this later on with better argumentation and language :)

MellowMelon:
No issues with the language, and I understand the reluctance to accept the PR if the change is breaking. Thank you for the quick response. I might prepare a PR to fix the docs instead, since those are clearly broken.

I do want to point out that Bluebird makes no such assumptions about calling then multiple times. The reason we discovered this bug is when we used [Promise.each](http://bluebirdjs.com/docs/api/each.html) ([source code](https://github.com/petkaantonov/bluebird/blob/b26b205f1551e4c89651b279aeb54c6e5ae3e86c/src/each.js#L14)) on some knex queries, which resulted in each query being executed twice. This is extremely surprising behavior, probably more surprising than if the samples in the comments above failed. But if the promises spec really doesn't care about then'ing multiple times, that might be for Bluebird to consider. (I haven't gone to them yet.)

wubzz:
@MellowMelon Actually I just applied this patch to 0.13.0 in a local project, and tested against the example code, and the result is as I suspected.

```javascript
const queryBuilder = knex('Accounts').select();

queryBuilder.on('query', (a) => {
	console.log(a.sql);
});

queryBuilder
.then((rows) => {
	console.log(rows.length);
	return queryBuilder.whereRaw('1 = 0').select(); //Append where clause
})
.then((rows) => {
	console.log(rows.length);
});
```

Console output:
```javascript
select * from "Accounts"
27
27
```

elhigu:
I updated https://github.com/knex/documentation/issues/27 with better description. I think this can be closed for now and if this change is coming to for example 1.0 by @tgriesser then we should discuss it better how everything should work exactly when the behaviour is changed. 

I'll add some tests which makes the current functionality official and which also tests that queryBuilder.then() returns a promise which resolves always to the same result.

MellowMelon:
wubzz: My bad. I misinterpreted the intent of your example..

PR #2233, opened 2017-09-21
Add redshift support without changing cli or package.json

acofer:
Hi. We (at Skuid) are trying to get mssql support by updating to latest knex while keeping the redshift support added in https://github.com/trevorsmith/knex. I noticed that the original pull request from that repo included (probably unwanted) changes to cli.js and package.json. Those are reverted in this PR - is it possible that it could go in now?

The integration tests are not working for me because I don't have local instances of mariadb, oracle, mssql, etc. to test against, and I haven't added docker/dockerode configuration for them to run against containers. If that is the only roadblock to accepting, I can do it?


package.json
```diff
@@ -78,7 +78,7 @@
     "pre_test": "npm run lint && npm run docker_test",
     "tape": "node test/tape/index.js | tap-spec",
     "debug_tape": "node-debug test/tape/index.js",
-    "test": "npm run pre_test && istanbul --config=test/.istanbul.yml cover node_modules/mocha/bin/_mocha -- --check-leaks -t 10000 -b -R spec test/index.js && npm run tape"
+    "test": "node_modules/mocha/bin/mocha --inspect --check-leaks -t 10000 -b -R spec test/index.js"
```
elhigu: Why you have removed coverage and tape tests?

scripts/docker-for-test.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 if [ -n "$(which docker)" ]; then
-    DOCKER_IMAGES=("mysql:5.7" "postgres:9.6")
+    DOCKER_IMAGES=("mysql:5.7" "postgres:9.6" "quay.io/skuid/docker-aws-redshift")
```
elhigu: these docker images are for connection breaking etc tests, not used for general travis CI integration tests.

src/migrate/index.js
```diff
@@ -97,9 +97,15 @@ export default class Migrator {
 
   forceFreeMigrationsLock(config) {
     this.config = this.setConfig(config);
+    console.log("LOOKING FOR LOCKTABLENAME");
```
elhigu: leftover debug prints should be removed

package.json
acofer: This is just a progress commit. I'll ping you when it's ready.

acofer:
It took a while, but all of the integration tests pass against an instance of redshift I set up. I'm not sure what redshift instance Travis CI tests should run against. Thoughts?

elhigu:
@acofer probably it is easiest to start some docker image running the server in travis e.g. with docker-compose.

acofer:
It would be, but there's not a fully-compatible redshift container available, or code from Amazon with which I could build one. I ran all my integration tests against an instance in my personal AWS account.

I tried postgres 8.0.2, but because it is a bit different, the integration tests fail (specifically, redshift has "IF [NOT] EXISTS", but it does *not* have "INSERT...RETURNING").

So, two options:
1. Someone* pays for a redshift instance to keep running for CI tests (about $140/month).
2. Write special logic in the integration tests to modify the redshift dialect when running against postgres 8.0.2 in docker. This feels bad?

* Perhaps my company, but I'd have to go through asking for approval for this case.

elhigu:
Not very attractive options.

There is not much use to keep on running instance in RDS that is actually needed pretty infrequeantly and wouldn't work when travis is running multiple parallel builds e.g. for different node versions.

For second one, I'm not sure if the benefits of having integration tests are enough when they actually are not run on real DB / is not actually the same code that is ran against real DB.

Maybe we just have to settle in the fact that we cannot CI test this one and accept that this dialect works only on with certain verified knex versions and that when ever there is a problem with this dialect, you need to check it out (this dialect really needs its own maintainer, like oracle and mssql dialects has).

acofer:
I'm willing to be the maintainer for the redshift dialect. Tomorrow I can take out the redshift CI testing so that Travis CI will pass for this PR.

acofer:
Tests pass woooo! Sorry it took a few days, we had some fires to put out.

elhigu:
Awesome! I'll give this still one final review and merge it in if everything seems to be in order.

elhigu:
I think that only thing that is still missin is link to documentation PR, which adds info about this dialect to docs repo and to readme.

acofer:
Cool! I'll write and file a docs PR tomorrow or early next week.

acofer:
https://github.com/knex/documentation/pull/61

test/knexfile.js
```diff
@@ -147,6 +147,21 @@ var testConfigs = {
     }
   },
 
+  redshift: {
+    dialect: 'redshift',
+    connection: testConfig.redshift || {
+      adapter:  'postgresql',
+      database: 'dev',
```
elhigu: knex_test has been used as a test DB name in most of the tests

elhigu:
This needs rebase before it can be merged in and would be good to change the test db name to knex_test.

acofer:
Ping?

elhigu:
Well deserved ping. Sorry I've been really busy and I still am and might be until the 10th of January. I hope I manage to give this final review before that though.

test/integration/builder/transaction.js
```diff
@@ -316,15 +327,13 @@ module.exports = function(knex) {
...
-      var db = Knex({
-        client: knex.client.driverName,
-        pool: {
-          min: 0,
-          max: 1
-        },
-        connection: knex.client.connectionSettings,
-        acquireConnectionTimeout: 300
-      })
+      const knexConfig = _.clone(knex.client.config);
```
elhigu: You should be overriding the whole .pool or otherwise you will modify knex.client.config.pool (https://runkit.com/embed/8zngcllrbqta)

test/integration/schema/index.js
```diff
@@ -11,9 +11,11 @@ module.exports = function(knex) {
     describe('dropTable', function() {
 
       it('has a dropTableIfExists method', function() {
+        this.timeout(30000);
```
elhigu: (process.env.KNEX_TEST_TIMEOUT || 5000) + 30000 or Math.max (KNEX_TEST_TIMEOUT, 30000) would be better, because mocha timeout is actually set to 60 seconds in CI

elhigu:
I'm sorry for taking forever with this. Now I just found two small things to change and I'm ready to merge.

elhigu:
Thank you very much

spiderbites:
Very excited to see this land in knex, is there any timeline for a new release that includes the new dialect?

elhigu:
This week

mscheffer:
FYI, a timestamp with a default value will trigger an error when inserting data into it (im talking about redshift)

is there a dedicated topic where we can report details like this one ?

to recap, you need a plain timestamp, don't set any flags about null values, default values etc, i think this applies to other field types as well.

GJ making knex compatible with redshift

elhigu:
Just by reportin new issue with complete, but minimal code example how to reproduce the problem. Please check contributing.md and the issue template.

Issue #1286, opened 2016-03-16
.toSQL() does not position bindings correctly


790:
Version 0.10.0

`> var knex = require('knex')({client:'pg'});`
`> knex.select().from('foo').where('bar', '=', '123').toSQL()`
`{ method: 'select',
  options: {},
  bindings: [ '123' ],
  sql: 'select * from "foo" where "bar" = ?' }`

I would expect the result when using postgres dialect to be `select * from "foo" where "bar" = $1`
knex seems to only perform the binding positioning when it's actually running the queries itself rather than just generating them.

I'd can submit a PR to fix this for pg, assuming it's not intentional, but I'm not familiar with other dialects  and it probably affects them also.

wubzz:
@elhigu Sounds like the same issue as `prepBindings` in #1269 . Maybe we should move `positionBindings` calls to `toSQL()` as well.

[Reference](https://github.com/tgriesser/knex/blob/master/src/dialects/postgres/index.js#L145)

elhigu:
@wubzz I'm not sure if this is an issue... knex has always worked that way that position bindings are replaced just before sending the query. If that is going to be changed the consequences must be investigated throughly...

790:
If it's always worked that way then I'd posit that it's always been broken. A toSQL() method that returns SQL that you can't actually run (at least in postgres) certainly seems broken.

The docs mention that "you can even use knex without a connection, just for its query building features", which is how I'm using it. If this is a supported way of using it then presumably it should produce valid queries whether having a connection or not.

I wouldn't be averse to a new method if compatibility with existing uses of toSQL() is a problem, but I can't see many cases where fixing it would cause a problem. The tests might fail, but they would have been wrong in the first place.

elhigu:
@790 If you like to form query which has all bindings replaced / escaped correctly etc. you can already call `.toQuery()` which returns SQL string where bindings are replaced with values. 

I understand that e.g. if you use `pg` driver you might like to have bindings and postgres compatible SQL separately in hand and I'm not saying that it would be bad thing at all if `.toSQL()` would already give you client specific SQL with in clients SQL binding format... it makes sense.

That being said it is so fundamental change to earlier functionality of `knex` that I don't feel being correct person to make that big decision by myself. 

Probably @tgriesser is the only one who could approve this.

790:
@elhigu Yeah, I have been using toString/toQuery but ideally I want the proper paramaterized queries to pass to pg.

[...]

PR #2237, opened 2017-09-21
Build native sql for a dialect without to string

elhigu:
There has been a problem that knex hasn't been able to be used as a just query builder, except with SQL where all values are interpolated to one string, which has been passed as a string to driver.

This PR adds a way to get toSQL() result also in native SQL dialects format, where bindings are given separately and SQL string with dialects replacements are provided separately. 

For example with postgresql dialalect query would be like:

```js
{
  sql: `select * from "table" where id = $1`,
  bindings: [1]
}
```

Feature is implemented to the result of `toSQL()`as a getter function to prevent memory overhead of storing SQL strings and bindings always in two different format.

Here is also one related issue, where this was first requested https://github.com/tgriesser/knex/issues/1286

### DIFF
--- a/scripts/docker-for-test.sh
+++ b/scripts/docker-for-test.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [ -n "$(which docker)" ]; then
+if [ -n "$(docker info)" ]; then
     DOCKER_IMAGES=("mysql:5.7" "postgres:9.6")
     for image in ${DOCKER_IMAGES[@]}; do
         if [ -z "$(docker images -q ${image})" ]; then
--- a/src/client.js
+++ b/src/client.js
@@ -133,6 +133,7 @@ assign(Client.prototype, {
 
   query(connection, obj) {
     if (typeof obj === 'string') obj = {sql: obj}
+    obj.sql = this.positionBindings(obj.sql);
     obj.bindings = this.prepBindings(obj.bindings)
     debugQuery(obj.sql)
     this.emit('query', assign({__knexUid: connection.__knexUid}, obj))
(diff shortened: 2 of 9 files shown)


elhigu:
@wubzz any comments? To me its a bit hacky to add anonymous function to every toSQL() result, but it felt a better option than changing return type of toSQL() to some new class instance.

wubzz:
@elhigu I agree adding it a new class is overkill. A new function is sufficient enough. However for reading purposes I'd probably suggest something along the lines of:

```javascript
  toSQL(method, tz) {
    this._undefinedInWhereClause = false;

    method = method || this.method
    let val = this[method]()

    const query = {
      method,
      options: reduce(this.options, assign, {}),
      timeout: this.timeout,
      cancelOnTimeout: this.cancelOnTimeout,
      sql: this[method](),
      bindings: this.formatter.bindings || [],
      __knexQueryUid: uuid.v4(),
      toNative: () => ({
       sql: this.client.positionBindings(query.sql),
       bindings: this.client.prepBindings(query.bindings),
     }),
    };

   if ((method === 'select' || method === 'first') && this.single.as) {
       query.as = this.single.as;
    }

    if(this._undefinedInWhereClause) {
      debugBindings(defaults.bindings)
      throw new Error(
        `Undefined binding(s) detected when compiling ` +
        `${method.toUpperCase()} query: ${query.sql}`
      );
    }

    return query;
}
```

But this suggestion isn't really because of your change, the function was already kind of strange to interpret, especially the part where it forces `sql` to become an object only to call `assign` for evidentally no reason at all?

Pardon the strange indentation..

elhigu:
@wubzz Thanks for comments. I agree that original way how that method was written is a bit odd. I'll try to do something to it before merging.

elhigu:
Looks like actually that assign value to query is necessary, since there are more attributes than just `sql` inside the value. I cleaned it up a bit anyways though.

PR #2239, opened 2017-09-22
Updated oprator list and allow operator transformation for e.g. escaping ?

elhigu:
Actually this is slightly breaking change, since now original format how operator was written is not preserved anymore. for example `iLiKe` transforms to `ilike`.

mainly added support for `.where('col', '?', something)`


elhigu:
@wubzz since next release is already breaking one, I consider it ok to make normalization of operator at the same time. I think it was originally an accident to allow preserving case how operator was written in this case.

Issue #2210, opened 2017-09-04
Sets the comment for an increments column


zhongzhi107:
```
knex.schema.createTable('accounts', function(t) {
  // .comment() does not work
  t.increments().comment('AUTO_INCREMENT ID');
});
```

elhigu:
@zhongzhi107 I suppose this is a bug report. In bug reports would be helpful to add also information, what system does right now or what kind of error is thrown, which database is used and what would be expected functionality / query.

Anyways looks like indeed comment is not created for that column at least on mysql:

https://runkit.com/embed/74wcu7p7z5b5

Do you know which would be expected output for your query?

zhongzhi107:
@elhigu Thanks for your reply.

The following result is what I expected:
```
create table `accounts` (`id` int unsigned not null auto_increment primary key comment 'AUTO_INCREMENT ID')
```

This is my system environment information:

```
> sw_vers
ProductName:	Mac OS X
ProductVersion:	10.11.1
BuildVersion:	15B42

> mysql --version
mysql  Ver 14.14 Distrib 5.7.16, for osx10.11 (x86_64) using  EditLine wrapper

> node -v
v7.6.0

> npm list knex
├── UNMET PEER DEPENDENCY pm2@2.x
  └── knex@0.13.0 
```

richraid21:
I dug through the code and found the problem.    

On [line 62](https://github.com/tgriesser/knex/blob/b5d93f10b1a96c47681f5e08a03d79b10a5fb226/src/schema/columncompiler.js#L62) of the common `src/schema/columncompiler` , `getModifiers()` specifically excludes columns of type `increments`. An initial look makes me think this was to avoid the possibility of a user setting a different, disallowed modifier on an incremental column, such as `nullable`.    

Removing this check however does not solve the entire problem. 

If the user does not explicitly define a column name, the generated SQL for adding the comment is incorrect: it renders a `undefined` as the column name. This is due to the default column name being [injected at the generic level](https://github.com/tgriesser/knex/blob/b5d93f10b1a96c47681f5e08a03d79b10a5fb226/src/schema/columncompiler.js#L48), thus, the dialect level `comment()` method is unaware of the column name.

I'd be happy to implement a fix if anyone has suggestions. 

@elhigu ?

elhigu:
[...]

PR #2243, opened 2017-09-25
Fix comments not being applied to increments columns

richraid21:
Addresses #2210 

**Changes**
+ I first had to remove the explicit check in the `getModifiers` function that excluded auto-increment columns.

+ I then abstracted the default column name logic for increments columns to a separate function on the prototype of the generic column compiler. I did this because if a user specified a comment on an increments column with no explicit name, the dialect level `comment` method could not figure out the default column name. It is implemented with a lookup map for `defaults`. In the future, if new defaults are added, developers can add a new default key.

+ Finally, I added a specific assertion in a test case to ensure comments are applied to increments columns in the future

**Notes**
+ I added a `.vscode/` exclusion to the `.gitignore`. Visual Studio Code allows developers to set up settings on a workspace basis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ bin/db.sqlite3
 # Bench stuff
 test/bench/temp
 test/coverage/*
+
+#VsCode workspace
+.vscode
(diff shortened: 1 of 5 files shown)


elhigu:
/home/travis/build/tgriesser/knex/src/schema/columncompiler.js
  30:1  error  Line 30 exceeds the maximum line length of 100  max-len

richraid21:
+ Shortened Line Length
+ Added explicit comment tests for mysql and postgres dialects

elhigu:
This looks great now, thanks!

PR #2252, opened 2017-10-02
add link to objection ORM

capaj:
IMHO objection deserves it's spot right there beside bookshelf

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,9 @@ Node.js and the Browser, featuring:
 
 For support and questions, join the `#bookshelf` channel on freenode IRC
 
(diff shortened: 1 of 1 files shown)


elhigu:
I'm too biased to be deciding if this should be merged or not. For now I've been refraining from promoting objectionjs in knex related discussions, since bookshelf has been *the official knex ORM*.

I think it is up to @tgriesser and other people in community if they feel that Bookshelf should be still exclusively promoted ORM.

tgriesser:
Yep, not opposed to promoting any ORM's that use Knex. Thanks @capaj!

PR #2255, opened 2017-10-03
Fix issue where table can be created with a hyphen

jonathanherring:
New to the library, I created a table in Postgres with a hyphen, ran into (obvious) issues later on when trying to work with the table. Throwing an error on table creation would help prevent folks from running into the same problem.

### DIFF
--- a/src/schema/tablebuilder.js
+++ b/src/schema/tablebuilder.js
@@ -206,6 +206,9 @@ TableBuilder.prototype.foreign = function(column, keyName) {
             if (typeof tableName !== 'string') {
               throw new TypeError(`Expected tableName to be a string, got: ${typeof tableName}`);
             }
+            if (tableName.indexOf('-') > -1) {
+              throw new TypeError(`Table names cannot contain hyphens, got: '-'`);
+            }
             foreignData.inTable = tableName;
             return returnObj;
           },


elhigu:
Would you explain the obvious problems you ran into?

jonathanherring:
Yes, of course. When I used some Knex raw in my seed file, the seed would error, unable to recognize the table name (a very similar command was working on other seeds). After that, migrate:rollback no longer worked because of that same table. One of my Knex commands would work on local with the hyphenated name, but not when deployed to heroku. I had to manually drop all the tables in PSQL and rename to fix the problem.

After some subsequent research, it looks like hyphenated names *can* be used in psql, but every time the table is called in any way, it must have double quotes around it, other query languages require brackets or back tics. 

Perhaps the error message could ask for some sort of escaping to be used with the hyphen.
However, after talking with some other devs, it seems the standard is to not use hyphens so as to avoid the chance at errors all together.

justsml:
Would a solution be to always auto-escape (or quote) the table name?
Maybe only when it's got invalid chars?

elhigu:
Sounds like you have been using knex.raw wrong without using `??`binding for identifiers, which does the quoting automatically.

PR #2261, opened 2017-10-09
Post processing hook for query result

elhigu:
Implement hook for allowing to add post processing hook to modify query results (for example to implement snake_case -> camelCase conversion for returned column names) Implements second part for https://github.com/tgriesser/knex/issues/2084

Docs https://github.com/knex/documentation/pull/58

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -163,6 +163,13 @@ assign(Client.prototype, {
     return sql;
   },
 
+  postProcessResponse(resp) {
+    if (this.config.postProcessResponse) {
+      return this.config.postProcessResponse(resp);
+    }
+    return resp;
+  },
+
   wrapIdentifier(value) {
     if (this.config.wrapIdentifier) {
       return this.config.wrapIdentifier(value, this.wrapIdentifierImpl);
--- a/src/runner.js
+++ b/src/runner.js
@@ -132,19 +132,24 @@ assign(Runner.prototype, {
     return queryPromise
       .then((resp) => {
         const processedResponse = this.client.processResponse(resp, runner);
+        const postProcessedResponse = this.client
+          .postProcessResponse(processedResponse);
+
         this.builder.emit(
           'query-response',
-          processedResponse,
+          postProcessedResponse,
           assign({__knexUid: this.connection.__knexUid}, obj),
           this.builder
         );
+
         this.client.emit(
           'query-response',
-          processedResponse,
+          postProcessedResponse,
           assign({__knexUid: this.connection.__knexUid}, obj),
           this.builder
         );
-        return processedResponse;
+
+        return postProcessedResponse;
       }).catch(Promise.TimeoutError, error => {
         const { timeout, sql, bindings } = obj;
 
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -10,6 +10,42 @@ module.exports = function(knex) {
 
   describe('Additional', function () {
 
+    describe("Custom response processing", () => {
+
+      before('setup custom response handler', () => {
+        knex.client.config.postProcessResponse = (response) => {
+          response.callCount = response.callCount ? (response.callCount + 1) : 1;
+          return response;
+        };
+      });
+
+      after('restore client configuration', () => {
+        knex.client.config.postProcessResponse = null;
+      });
+
+      it('should process normal response', () => {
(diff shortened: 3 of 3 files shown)


richraid21:
Should this be setup as a queue of functions rather than a single assignable property? 

What if the client needs multiple post-processing steps?

elhigu:
> Should this be setup as a queue of functions rather than a single assignable property?
> What if the client needs multiple post-processing steps?

@richraid21 Actually I thought of that, but I wanted to keep knex side simple and robust and leave it up to client to add chaining functionality by themselves if they wish.

joelmukuthu:
@elhigu is it possible to pass the queryBuilder instance that generated the request to this hook, and also for the `wrapIdentifier` hook?

src/client.js
```diff
@@ -163,6 +163,13 @@ assign(Client.prototype, {
     return sql;
   },
 
+  postProcessResponse(resp, method) {
```
joelmukuthu: here `method` is `undefined`. it's not passed here https://github.com/tgriesser/knex/pull/2261/files#diff-b27035181a616c40188ad22a32280732R136

elhigu: I actually decided in the last moment when implementing this, that I don't want to pass `method` to that handler, since it would expose too much internal implementation details of `knex`to end user. If something extra is passed it can be specified and added later (I was planning to use that to allow people to distinct between normal and raw responses).

elhigu: I'll remove this reminder of old initial implementation before merging.

elhigu:
> @elhigu is it possible to pass the queryBuilder instance that generated the request to this hook, and also for the wrapIdentifier hook?

I'm not sure. I would consider that as additional feature though. What kind of use case you had for the query builder? 

Maybe you could add new issue about it and the use cases why it would be good to have query builder reference passed to these hooks (there might be better more general ways to implement support for those use cases).

src/client.js
joelmukuthu: Makes sense, and it probably isn't necessary to know which method generated some data that needs post-processing

src/runner.js
```diff
@@ -132,19 +132,24 @@ assign(Runner.prototype, {
     return queryPromise
       .then((resp) => {
         const processedResponse = this.client.processResponse(resp, runner);
+        const postProcessedResponse = this.client
+          .postProcessResponse(processedResponse);
+
         this.builder.emit(
           'query-response',
-          processedResponse,
+          postProcessedResponse,
```
wubzz: Technically if a `postProcessResponse` function has been supplied in the config, this is no longer the `query-response`. But I'm not sure if it really matters given the context of the event. Just something I took note of.

joelmukuthu:
In both hooks, I need a some context about the identifier or data being processed. In knorm (an ORM sitting on top of knex), I need to know which model the data belongs to in the case of `postProcessResponse`. I thought the easiest thing was to add a reference to the model in the query builder and then I can use a standard hook for processing:

```js
knex.client.config.wrapIdentifier = (identifer, wrap, builder) => {
  if (isField(identifier)) {
    const Model = builder._knormModel;
    const field = Model.fields[identifier];
    return wrap(field.column); // the user is allowed to specify a different column per field
  }
  return wrap(identifier);
};
knex.client.config.postProcessResponse = (data, builder) => {
  const Model = builder._knormModel;
  return new Model(data);
};
```
`postProcessResponse` is not a blocker since I have to overload the CRUD functions anyway. But with `wrapIdentifier` the only other way (I can think of) is to do the pre-processing before sending the identifiers to knex, which means overloading all the query-builder methods. Do you have some suggestions for this? I can as well create a new issue if it's necessary

src/runner.js
elhigu: It hasn't been really query response earlier either, because each `knex` dialect is processing the raw response from driver before passing it to this event.

elhigu:
@joelmukuthu yeah, some kind of method for passing context would be useful. Not sure if that should be the query builder or some other context object. Currently `builder._knormModel` probably wont be copied along if one does `builder.clone()`. Could be better to have some context object there for passing data... that could be useful also for current query events.

src/client.js
tecfu: > I was planning to use that to allow people to distinct between normal and raw responses

Precisely what I was looking for from this.

Issue #2076, opened 2017-05-20
How to write unit tests for methods that use Knex.


odigity:
(Originally posted in #1659, moved here for further discussion.)

I'm kind of in the dark on this one - the knex docs cover the barebones of transactions, but when I google "using knex.js for unit testing" and "ava.js + knex" and "ava.js + database", I couldn't find anything particularly instructive to show me a good way, so I fell back on my Ruby experience where wrapping a unit test in a transaction is a common technique for reseting the DB after each test.

> @odigity I wouldn't suggest of doing that because your tests will end up using just one connection instead of connection pool and it will not represent real world usage of your app.

That did occur to me, but I was willing to accept that if it allowed me to use a simple, write-once method for implementing post-test DB cleanup.  (I did set my pool min/max to 1 to be sure.)  If there's a way to accomplish this while supporting concurrent connections, I will absolute embrace that.
 
> Also its pretty much impossible to implement unless the application you are testing is running its queries in the same transaction that was started inside test code (you would need to start transaction in test then start your app and pass the created transaction to app so that it makes all the queries to the same transaction and nested transactions might behave starngely...).

The way I hoped/expected it to work is:

1) In each unit test, I create a transaction which produces `trx`.

2) I then require the module I want to test and pass the `trx` object to the module constructor so it will be used by the module, thus resulting in all queries happening inside the transaction.

3) After the module method returns (or throws error), I run my assertions on the resulting state of the DB, then call `trx.rollback()` to undo everything from the start to prepare for the next test.

So, that's what I'm trying to achieve and how I originally intended to achieve it.  I'm eager to learn more about:

1) In what whys I'm misunderstanding how Knex.js works and should be used.

2) Best practices for writing atomic unit tests for code that touches the database.

odigity:
I apparently did a poor job googling last night, because I just tried again with "how to write atomic unit tests with knex.js", and am getting some interesting results.  Going to read through them now.  Will post again if I learn something.

odigity:
Nope - got nothing.  Summary in chronological order of sources:

*2016-04-21*  https://medium.com/@jomaora/knex-bookshelf-mocks-and-unit-tests-cca627565d3

Strategy:  Use [mock-knex](https://www.npmjs.com/package/mock-knex).  I don't want to mock the DB - I want to test my methods against an actual MySQL DB to ensure correct behavior - but I took a look at `mock-knex` anyway... it may just be the worst designed library I've ever encountered.  :(

[...]

Issue #285, opened 2014-05-24
Listen/Notify support


StephenGentle:
Is support for listen/notify planned at all? I'm using postgres and currently using a mix of directly using the pg library as well as knex. It would be nice to be able to just use knex.

I am subscribing to the notifications as in this example: http://bjorngylling.com/2011-04-13/postgres-listen-notify-with-node-js.html

rosendi:
@tgriesser Could you please provide a simple example?

farzher:
I can't figure out how to use postgres NOTIFY/LISTEN.
I don't want to make a new `pg` client connection, I want to use an existing knex connection.
I just want to just get access to the raw `pg` client and use their driver's support, but idk how.

drew-r:
@farzher - `.client.acquireRawConnection()`

mikemfleming:
@tgriesser your suggestion to implement this yourself makes sense to me. are you specifically recommending that developers use node-postgres to do that? or were you implying that there is a way to do the same with knex?

here is a relevant example of a node postgres implentation.
http://bjorngylling.com/2011-04-13/postgres-listen-notify-with-node-js.html

Msea:
I am trying to use Listen with knex. Using @farzher 's idea I accessed the raw pg connection on app startup and was able to listen on my channels.
However, because I am using a single raw connection and not the connection pool if my database goes down, I lose the connection. Is there a way around this? Can I use my knex connection to poll for when by db comes back up?

JonnyLinja:
I'm implementing the same thing as @farzher and @Msea. I have it working, but am unsure as to two things.

**1. Separate Connection vs Pool**

Are there any disadvantages to acquiring a connection from the pool, listening to the channel, and then releasing the connection back into the pool? Specifically, I see that maintaining a separate connection via `.client.acquireRawConnection()` was recommended by @drew-r. Was this because of side effects, interference with how knex operates, or performance issues?

Example of what I mean as follows.
```
const connection = await knex.client.acquireConnection();
connection.query('LISTEN foo');
connection.on('notification', (msg) => {
   console.log("got " + msg.channel + " payload " + msg.payload);
});
await knex.client.releaseConnection(connection);
```

**2. Dealing with Disconnects**

For database disconnects, are there any better alternatives to repeatedly calling acquireConnection or acquireRawConnection until it succeeds?

[...]

PR #2273, opened 2017-10-18
enable migrations to run within an external transaction

wenzowski:
After reading #2076 I'm not sure why we can't pass a transaction to `Migrator`?

### DIFF
--- a/docs/src/guide/migrations.md
+++ b/docs/src/guide/migrations.md
@@ -308,6 +308,26 @@ exports.config = { transaction: false };
 
 The same config property can be used for enabling transaction per-migration in case the common configuration has `disableTransactions: false`.
 
+#### Running migrations within an external transaction
+
+The migration methods `latest`, `up`, `down` and `rollback` can be called on a transaction object. When invoked this way, they reuse the surrounding transaction instead of opening a new (nested) one, so the migrations become part of the outer transaction and are committed or rolled back together with it.
+
+This is especially useful for transaction-based testing: run your migrations, exercise your code, then roll back the transaction to leave the database untouched.
+
+```js
+await knex.transaction(async (trx) => {
+  // Migrations run inside `trx`, not in a new nested transaction
+  await trx.migrate.latest();
+
+  // ... run your tests against `trx` ...
+
+  // Rolling back the transaction undoes the migrations as well
+  await trx.rollback();
+});
+```
+
+Note that this is not supported on sqlite3, which does not allow schema changes inside a transaction that is later rolled back without side effects.
+
 #### Sqlite3-specific concerns
 
 Disabling / enabling foreign key checking [has no effect within a transaction](https://sqlite.org/pragma.html#pragma_foreign_keys) on sqlite3, so features such as `.dropColumn()` cannot be emulated correctly inside of a migration-level transaction (and [can even cause data loss](https://github.com/knex/knex/pull/6315)). Knex now throws an error in these cases.
--- a/docs/src/guide/migrations.md
+++ b/docs/src/guide/migrations.md
@@ -315,18 +315,25 @@ The migration methods `latest`, `up`, `down` and `rollback` can be called on a t
 This is especially useful for transaction-based testing: run your migrations, exercise your code, then roll back the transaction to leave the database untouched.
 
 ```js
-await knex.transaction(async (trx) => {
-  // Migrations run inside `trx`, not in a new nested transaction
-  await trx.migrate.latest();
+const Rollback = new Error('rollback');
+try {
+  await knex.transaction(async (trx) => {
+    // Migrations run inside `trx`, not in a new nested transaction
+    await trx.migrate.latest();
 
-  // ... run your tests against `trx` ...
+    // ... run your tests against `trx` ...
 
-  // Rolling back the transaction undoes the migrations as well
-  await trx.rollback();
-});
+    // Throw to roll the whole transaction (migrations included) back
+    throw Rollback;
+  });
+} catch (err) {
+  if (err !== Rollback) throw err;
+}
 ```
 
-Note that this is not supported on sqlite3, which does not allow schema changes inside a transaction that is later rolled back without side effects.
+Use a thrown error to trigger the rollback rather than calling `trx.rollback()` and letting the callback resolve normally — the latter leaves Knex trying to both roll back and commit the same transaction, which hangs on some drivers (e.g. MSSQL).
+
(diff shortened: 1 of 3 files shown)


wenzowski:
```javascript
const rollback = new Error('rollback');
await knex.transaction(async (trx) => {
  await knex.migrate.latest({trx});
  expect(true).toBe(true) // run some tests
  await knex.rollback({trx});
  throw rollback; // never actually commit
}).catch((error) => {
  if (error === rollback) return;
  console.error(error && error.stack);
  throw error;
});
```

elhigu:
This is missing at least tests and documentation. I'm not sure if this is necessary either. Wouldn't it be enough to just pass the transaction instead of knex pool to a constructor of the class? 

```
knex.transaction(async (trx) => {
  let migrator = new Migrator(trx);
  await migrator.xxxxx(yyyy);
});
```

wenzowski:
yes & yes; coming in future commit

I think a sane api would be something like
```javascript
knex.transaction(async (trx) => {
  trx.migrate.latest();
})
```
and the `trx` would stay out of `options`

wenzowski:
oh, and this seems absolutely crucial for transaction-based tests

```javascript
knex.transaction(async (trx) => {
  await knex.migrate.latest({trx});
  const conn = await trx.client.acquireConnection();
  // interact with a raw connection inside transaction context
  throw rollback;
});
```

elhigu:
@wenzowski I don't understand the second example, why you need to acquire connection from clients (I don't remember what that does exactly when called through trx.client... it might just return the same connection all over again that you already have and throwing in the end I suppose is just there if you like to force rollback).

About the earlier example does that currently work?

```
knex.transaction(async (trx) => {
  await trx.migrate.latest(); // await was missing
})
```

or this

```
knex.transaction(trx => trx.migrate.latest())
```

If those work, maybe this PR can be closed?

rapzo:
Ah! Was about to open a PR with the same proposal. Good one @wenzowski

wenzowski:
`knex.transaction(trx => trx.migrate.latest())` does not work. It was the first thing I tried ;)

Regarding the 2nd example, there are all sorts of times it can be useful to have access to the database connection itself. `knex.raw()` doesn't guarantee subsequent calls will be transmitted over the same connection, so if you've got a schema introspection library or an efficient dataloader or anything else that needs to run directly against the driver, that's how you'd get the current connection within the transaction scope.

elhigu:
Ah right, so that was workaround since method trx.raw doesn't exist for some reason... That is one another thing that should be fixed. Does trx.schema or trx.schema.raw exist and does they pass queries to the correct connection? Docs are lacking much info about this stuff and how to do this properly :(

rapzo:
Couldn't we provide the transaction in the constructor as well, via an optional second argument. It is kind of useful since the class is exported and can be extended for project needs?

Also, the `setConfig` needs to be public and use assign from lodash if everything gets transpiled and `Object.assign` is there already?

wenzowski:
Ok, I'll revisit this with something like
```javascript
export default class Migrator {

  constructor(knex, isTrx = false ) {
    this.knex = knex;
    this.isTrx = isTrx;
    this.config = this.setConfig(knex.client.config.migrations);
  }

  // ...
}
```
and
```javascript
knex.transaction(trx => {
  const migrate = new Migrator(trx, true);
  migrate.latest();
})
```
with `knex.transaction(trx => trx.migrate.latest())` to call `new Migrator` as above

elhigu:
I don't like the idea of passing the info if knex instance is transaction or not to the constructor. It would be better to add some property to transaction object to allow checking directly if that is the case (this `knex.isTransaction()` method would be beneficial for other uses too). 

However IMO correct fix for this would be to just make migration API to work through trx instance in other words to make `knex.transaction(trx => trx.migrate.latest())` to work.

Also I'm still wondering this part (I haven't got time to thest this):

> [Is it true that] ... method trx.raw doesn't exist for some reason ... Does trx.schema or trx.schema.raw exist and does they pass queries to the correct connection?

wenzowski:
`knex.transaction(trx => trx.raw())` does exist, ignore comment above about `acquireConnection()`. I still need that for my use case; others should probably use `trx.raw()` unless there's a good reason not to (eg. #285).

Rewrote still needs integration tests but docker takes too long to run so I'm going to let CI handle them.

elhigu:
Also mariadb tests did seem to fail.

src/util/make-knex.js
```diff
@@ -43,6 +43,10 @@ export default function makeKnex(client) {
       return client.transaction(container, config)
     },
 
+    isTransaction() {
```
elhigu: This part needs documentation

wenzowski:
Sorry for the commit noise, but I'm going to keep pushing here and letting CI run the tests. The whole docker test suite is super cool, but it's taking so long to run I can't get any other work done so I need it offloaded to CI. I'm unit testing before committing (nice and fast), but not running the entire suite.

No idea on mariadb, but I'll look into it

zsszatmari:
Hi! Just a thumbs up for this, seems important to me.

kibertoad:
Depends on https://github.com/tgriesser/knex/pull/3084/files as it uses same simplified logic for determining what is a transaction.

elhigu:
Why I cant see any changes in tests now? 🤔

kibertoad:
@elhigu the ones that were there didn't test anything useful. How exactly you would test no second transaction is started?

elhigu:
> @elhigu the ones that were there didn't test anything useful. How exactly you would test no second transaction is started?

By mocking transaction start method and takin call count I suppose 🤔

elhigu:
or counting knex.on('query') events maybe

wenzowski:
I don't know if this PR is helpful in any way? I've long since abandoned use of `knex` in all my projects in favour of raw sql.

kibertoad:
I believe it's a valid case and don't see why not.

bentron2000:
Just a note here that whilst the `isTransaction()` appears in the docs, the typescript defs do not include it. Cheers B

OlivierCavadenti:
Rebased and reimplemented. trx.migrate.latest(), up(), down() and rollback() now reuse the outer transaction instead of starting a nested one, so you can wrap migrations in a transaction for testing.
@kibertoad

OlivierCavadenti:
Hi !
I rebase and fixed up the tests and docs, CI is green now. 
@kibertoad @tgriesser could you take a look when you get a chance?

PR #2281, opened 2017-10-21
Fix: Escape table name in SQLite columnInfo call

### LINKED ISSUE #2184: knex("group").columnInfo() resolves to Error with SQLite3
When using `group` as table name, `knex("group").columnInfo()` resolves to:
```
PRAGMA table_info(group) - SQLITE_ERROR: near "group": syntax error
```

When escaping the table name, all works as expected.
```js
knex("group").columnInfo(); // -> error
knex("'group'").columnInfo(); // -> ok
```

System using: latest `knex`, latest `sqlite3`, latest `node`, latest `npm`, latest `macOS`.

Pinqvin:
Fixes #2184. Not escaping the table name could lead to errors with some table names (group, for example)

### DIFF
--- a/src/dialects/sqlite3/query/compiler.js
+++ b/src/dialects/sqlite3/query/compiler.js
@@ -102,7 +102,7 @@ assign(QueryCompiler_SQLite3.prototype, {
   columnInfo() {
     const column = this.single.columnInfo
     return {
-      sql: `PRAGMA table_info(${this.single.table})`,
+      sql: `PRAGMA table_info(\`${this.single.table}\`)`,
       output(resp) {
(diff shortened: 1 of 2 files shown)


Pinqvin:
Looks like tests are failing with Node 4.x version. Tried to reproduce this locally with 4.8.4 by running just the sqlite3 tests `DB='sqlite3' npm test`, but the tests seemed to run fine with the sqlite3 subset. Is this a random failure or am I missing something here?

elhigu:
Yeah, that was just a random fail. PR seems good, I'll merge this when I see that tests are passing  :) Thank you!

PR #2297, opened 2017-10-27
Update each out-of-date dependency according to david-dm.org

richraid21:
Update the dependencies list to latest stable version of outdated packages.


elhigu:
Thanks!

PR #2300, opened 2017-11-01
Add support for .knex-cli.js

frsechet:
Summary:
- added support for .knex-cli.js file in bin/cli.js.
- if .knex-cli.json contains knexfilePath, `knex` CLI tool uses that as the default path to local knexfile.
- `--knexfile` > `.knex-cli.js` > default configuration
- prevents having to add the same `--knexfile path/to/knexfile` option over and over again in a given project

Example .knex-cli.js:  
```
module.exports = {
    knexfilePath: "./src/myCustomKnexfile.js"
}
```


elhigu:
This is good idea 👍 

I would say that  `.knex-cli.json`should be `.js` file to allow using environment variables and comments there. 

Also I think this is good time to add tests for actually running the migration client from commandline. I've been reluctant of adding features to client because of lack of testing.

frsechet:
@elhigu ok, makes sense. I just added a commit to use .js instead of .json :-)

As for the tests, I won't have time to create a whole set of tests for CLI now (as the boilerplate to get started is a bit too big for my current schedule) but I'll happily contribute if someone gets started on that!

frsechet:
(closed and reopened because travis-ci found this commit to fail for no particular reason)

frsechet:
@elhigu so what's a next step for getting this merged upstream? :-)

elhigu:
> Also I think this is good time to add tests for actually running the migration client from commandline. I've been reluctant of adding features to client because of lack of testing.

Generally tests and documentation.

frsechet:
@elhigu I can do documentation, and I can do tests once there is some barebone for CLI, but definitely not create a whole new set of tests for CLI...

frsechet:
Hey @elhigu, as explained above I won't have time for creating a whole test suite for the CLI. I would be glad to add tests for this feature as soon as there is one, but in the mean time, are you not going to merge this? If not, then let's just close it...
As for documentation, where should it be added?

elhigu:
Yup I still don't want to merge any features without automatic tests. Documentation repo is in https://github.com/knex/documentation. No need to close this yet. I still am going to write some cli test suite if PR is still mergeable by then this can be merged.

frsechet:
https://github.com/knex/documentation/pull/94

elhigu:
Btw. one thing I didn't understand now in this feature is why you are not using `knexfile.js` instead which just requires original knexfile from correct path?

frsechet:
This is an extremely valid remark and one I didn't think about at all in the first place.

This tip would certainly have its place in the official documentation! I have seen that I was not alone wondering (see references above) and nobody in my team has thought about it as well...


I will propose a change in the documentation in another PR and close this one, I guess.

elhigu:
Yeah, I can't believe how I hadn't thought that before either. Closing this and adding just change to documentation sounds like correct way to go 👍

PR #2314, opened 2017-11-08
Add queryContext to schema and query builders

### LINKED ISSUE #2268: Pass context to wrapIdentifier and postProcessResponse hooks
`wrapIdentifier` will be called with a string for every identifier, be it a table-name, column-name, from a query-builder instance or even a schema-builder instance.

`postProcessReponse` will be called with any data coming from the database, be it from a `SELECT` query or queries with `RETURNING`  clauses. 

For both hooks, a little context is needed. Also, ideally the context should be able to be configurable by the user, for example something like:

```js
knex('foo')
  .hookContext({ wrapIdentifier: someObject, postProcessResponse: someOtherObject  })
  .select();

// and the hooks would be of some form like:

knex({
  client: 'pg',
  connection: /* ... */,
  wrapIdentifier(identifier, wrap, someObject) {
  },
  postProcessResponse(data, someOtherObject) {
  }
})
```

For more info check the discussion here https://github.com/tgriesser/knex/pull/2261. 

**NOTE:** These hooks are not yet released at the time of this writing.

joelmukuthu:
Closes #2268 

Adds a `QueryBuilder.prototype.queryContext` method, which allows configuring a context to be passed to `wrapIdentifier` and `postProcessResponse` per query builder.

Documentation: https://github.com/knex/documentation/pull/68


src/query/builder.js
```diff
@@ -53,6 +53,9 @@ assign(Builder.prototype, {
     if (!isUndefined(this._options)) {
       cloned._options = clone(this._options);
     }
+    if (!isUndefined(this._hookContext)) {
+      cloned._hookContext = clone(this._hookContext);
```
joelmukuthu: @elhigu does it make sense to clone the context object as well? I imagine someone would want the same context object to be passed around (to be able to build up on it) but I just added added this to conform with the rest

elhigu: well at least.. you really cannot do deep clone for it, so I would say that it is enough to copy only root element and if someone like some how shared content they just need to add nested structure to context and inner content will be just references to the same data... not sure if that is understandable at all..

joelmukuthu: I understand.. but is there a reason to clone the root context itself?

elhigu: @joelmukuthu yes. to be able to have different contexts for cloned queries. So I can clone query and then change / merge its root context without modifying the original. So that way it be more versatile and user can decide how to use it.

elhigu: Though it will be also caveat that must be documented.

joelmukuthu: Makes perfect sense, thanks for the explanation!

src/formatter.js
```diff
@@ -119,7 +120,11 @@ export default class Formatter {
   }
 
   wrapAsIdentifier(value) {
-    return this.client.wrapIdentifier((value || '').trim());
+    let hookContext;
+    if (this.builder && isFunction(this.builder.hookContext)) {
```
joelmukuthu: @elhigu I'm a bit unsure about these checks (along with [this one](https://github.com/tgriesser/knex/pull/2314/files#diff-b27035181a616c40188ad22a32280732R136)).. they seem a bit hacky but I had to add them for CI to pass. do you know which cases formatter will be instantiated without a builder?

joelmukuthu:
@elhigu / @richraid21 this is ready for review, PTAL. Also, question about the naming of the method: since @elhigu mentioned that the same context could be passed when query events are emitted, should it be called something less specific as `hookContext`, maybe just `context`?

joelmukuthu:
hey @elhigu, have you had a chance to look at this?

elhigu:
Sorry, I haven't got any extra time to put to knex last weeks and looks like I wont be having any until January.

elhigu:
Just FYI, I'm almost back in business and haven't forgotten this, nor the other waiting PRs.

joelmukuthu:
@elhigu thanks and no worries. i'll fix up the merge conflicts in the mean time

test/unit/query/builder.js
```diff
@@ -117,6 +121,30 @@ describe("Custom identifier wrapping", function() {
...
+        mysql: 'select `users_fancy_wrapper_was_here`.`foo_fancy_wrapper_was_here` as `bar_fancy_wrapper_was_here` from `schema_fancy_wrapper_was_here`.`users_fancy_wrapper_was_here`',
+        oracle: 'select "users_fancy_wrapper_was_here"."foo_fancy_wrapper_was_here" "bar_fancy_wrapper_was_here" from "schema_fancy_wrapper_was_here"."users_fancy_wrapper_was_here"',
+        mssql: 'select [users_fancy_wrapper_was_here].[foo_fancy_wrapper_was_here] as [bar_fancy_wrapper_was_here] from [schema_fancy_wrapper_was_here].[users_fancy_wrapper_was_here]',
+        oracledb: 'select "users_fancy_wrapper_was_here"."foo_fancy_wrapper_was_here" "bar_fancy_wrapper_was_here" from "schema_fancy_wrapper_was_here"."users_fancy_wrapper_was_here"',
+        postgres: 'select "users_fancy_wrapper_was_here"."foo_fancy_wrapper_was_here" as "bar_fancy_wrapper_was_here" from "schema_fancy_wrapper_was_here"."users_fancy_wrapper_was_here"',
+        sqlite3: 'select `users_fancy_wrapper_was_here`.`foo_fancy_wrapper_was_here` as `bar_fancy_wrapper_was_here` from `schema_fancy_wrapper_was_here`.`users_fancy_wrapper_was_here`'
+      }, clientsWithCustomIdentifierWrapper);
+    })
+  })
```
elhigu: To make tests a bit more complete should be also tested that 

1. if originalQuery context is changed that context of cloned query will remain intact. 
2. if originalQuery context is `{ objectInCtx: { value: 1 } }` and if value is changed to 2 then in cloned context value also changes (because context is cloned as shallow copy)

test/integration/builder/transaction.js
```diff
@@ -427,7 +428,35 @@ module.exports = function(knex) {
...
+      knex.client.config.wrapIdentifier = (value, wrap, context) => {
+        spy(context);
+        return wrap(value);
+      };
+
+      return knex.transaction(function(trx) {
+        return trx
+          .select()
+          .from('accounts')
+          .hookContext({ foo: 'bar' });
```
elhigu: I'll ask some second and third opinions about this name :)

elhigu: maybe queryContext() would be a bit nicer

elhigu: I actually didn't see the guy with opinions yesterday, so didn't get one. But I kind of like `.queryContext` it is not so specific that `.hookContext` is and it hints that context is bound to single query builder.

joelmukuthu: Agreed, will update :)

joelmukuthu: One thing though, this PR also adds that method to the schema builder. Is it okay to call it `queryContext` there as well or should be perhaps remove it altogether?

src/runner.js
```diff
@@ -132,8 +132,12 @@ assign(Runner.prototype, {
     return queryPromise
       .then((resp) => {
         const processedResponse = this.client.processResponse(resp, runner);
+        let queryContext;
+        if (this.builder && isFunction(this.builder.queryContext)) {
```
elhigu: Is there some case when `this.builder.queryContext` is not a Function? Or when this.builder is not declared? If those checks are here only because some tests are not initialized with builder, then tests should be changed instead (for example by mocking builder).

src/dialects/oracledb/index.js
```diff
@@ -53,7 +53,7 @@ Client_Oracledb.prototype.columnCompiler = function() {
   return new ColumnCompiler(this, ...arguments)
 }
 Client_Oracledb.prototype.formatter = function() {
-  return new Oracledb_Formatter(this)
+  return new Oracledb_Formatter(this, ...arguments)
```
elhigu: Shouldn't all these be 

```
= function(builder) {
   return new Oracledb_Formatter(this, builder);
}
```

Thats more understandable for anyone reading the code.

joelmukuthu: I had it like that at first but then I noticed that the convention was passing all the arguments, for example on line 53 there:https://github.com/tgriesser/knex/blob/59f6cba17839435e3c10d9e1011d5b78e4a970f0/src/dialects/oracledb/index.js#L53

Or is that because those are the generic compiler and transaction classes?

I agree though that it's more readable..

src/runner.js
joelmukuthu: Yeah there were some failing tests and I assumed it meant that some code would also fail. I'll remove the checks and go about fixing those tests :)

src/dialects/oracledb/index.js
elhigu: I suppose it has just been personal preference of the one who originally implemented oracledb dialect.

joelmukuthu: Hmm.. but then it's also the same on the [mssql dialect](https://github.com/joelmukuthu/knex/blob/cdb91b8f8759ae6f40f1007070996cdf31797014/src/dialects/mssql/index.js#L49-L67). I don't mean to prod I'm just a sucker for conventions :)

elhigu: Both dialects are pretty new ones and might have been used as a reference for each other. But sure you are right, since that syntax is used everywhere in those dialects, there is no reason why it couldn't be used also in this PR. So no need to change that.

joelmukuthu:
I fixed all the cases of the missing builder instance for the formatter, and ended up adding `queryContext` to `QueryBuilder`, `Raw`, `SchemaBuilder`, `TableBuilder` and `TableCompiler`.

elhigu:
> queryContext to QueryBuilder, Raw, SchemaBuilder, TableBuilder and TableCompiler

These new places, where it is available will also need tests and documentation. Other than that PR still looks good.

joelmukuthu:
I've tried to add tests for the schema, table and column builders and found some weird cases where the context needed to be passed down from `SchemaCompiler => TableCompiler => ColumnCompiler`. Hoping for a passing build then will update the docs tomorrow.

joelmukuthu:
I added the tests and updated the docs. PTAL :)

joelmukuthu:
Hi @elhigu. This has been ready for review for a while now, is there anything I can do to help out with the reviewing process?

elhigu:
Thanks for the patience with this. Those tests were really great, I'm really happy to merge this in! 👍

joelmukuthu:
Great, thanks!

PR #2329, opened 2017-11-15
allow to pass connection from cli

caub:
it would be great to be able to avoid a knexfile, and just passing a connectionString in options

with this PR, for example `knex --connection ${DB_URL-'pg://tt@localhost/tt'} migrate:latest` would do that

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -34,6 +34,13 @@ function checkLocalModule(env) {
 
 function initKnex(env) {
 
+  var connIdx = process.argv.indexOf('--connection');
+  var knex = require(env.modulePath || 'knex');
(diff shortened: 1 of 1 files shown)


elhigu:
This PR https://github.com/tgriesser/knex/pull/2300 would support also this use case and it enables to use more complex use cases too. Maybe you can finish that one?

caub:
well I saw that PR before, it requires another file, my whole idea was to get rid of files for this because in te rest of code I always initiate knex like this: `const knex = require('knex')(DB_URL);`, I don't really need the knexfile, the connection string can take care of any environment, pooling configuration and anything

PR #2332, opened 2017-11-15
Guarding commit and rollback in case they get called explicitly

bnewtonius:
For mssql dialects, when you call `trx.commit()` explicitly, an error will be thrown later when it tries to first commit against a null `conn.tx_`, which throws an uncaught exception, which later gets caught somewhere in knex, at which point it tries to do a rollback, which again fails because the transaction is null.

If we just check to see if the transaction is already complete first before trying to commit, we shouldn't get these errors.

tl;dr calling `.commit` explicitly on mssql transactions throws an uncaught exception when knex later tries to autocommit.

### DIFF
--- a/src/dialects/mssql/transaction.js
+++ b/src/dialects/mssql/transaction.js
@@ -17,6 +17,7 @@ export default class Transaction_MSSQL extends Transaction {
   }
 
   commit(conn, value) {
+    if ( this._completed ) return;
     this._completed = true
     debug('%s: commit', this.txid)
     return conn.tx_.commit()
@@ -28,6 +29,7 @@ export default class Transaction_MSSQL extends Transaction {
   }
 
   rollback(conn, error) {
+    if ( this._completed ) return;
     this._completed = true
     debug('%s: rolling back', this.txid)
     return conn.tx_.rollback()
@@ -41,6 +43,8 @@ export default class Transaction_MSSQL extends Transaction {
   }
 
   rollbackTo(conn, error) {
+    if ( this._completed ) return;
+    this._completed = true
     debug('%s: rolling backTo', this.txid)
     return Promise.resolve()
       .then(() => this.query(conn, `ROLLBACK TRANSACTION ${this.txid}`, 2, error))


elhigu:
No for this, its better to throw an error than silently ignore things, when people tries to rollback / commit transaction multiple times. (+ missing tests). 

If you have reasonable use case where knex is working wrong, this can be considered again. I was not able to deduce it from the description and there was no tests code to demonstrate the problem either.

bnewtonius:
Here's an example. When you do this against mssql, an uncaught exception is thrown after the promise chain completes (when it tries to autocommit for you).

```
    Promise.all([
        trx('books').insert({title: 'Slaughterhouse Five'}),
        trx('books').insert({title: 'Cat\'s Cradle'})
    ]).then( () => {
        trx.commit();
    });
});
```

An uncaught exception is thrown by the mssql commit:

https://github.com/tgriesser/knex/blob/master/src/dialects/mssql/transaction.js#L22

Which is caught in the promise chain later, which then tries to rollback because there was an error, but can't because `conn._tx` is null, so it throws another exception from the rollback:

https://github.com/tgriesser/knex/blob/master/src/dialects/mssql/transaction.js#L70

If you are using any other dialect, like pg, then you are using the base `Transaction` instead of the `Transaction_MSSQL`, then you will not see this uncaught exception. It is apparently OK with `COMMIT` being sent twice.

The problem is with the autocommit running regardless of whether or not you've called commit explicitly or not. It just throws an uncaught exception for MSSQL, while not throwing an exception for other dialects. This change was not to hide exceptions. It's more to keep it from running commit again after commit has already been called. Another possible option would be to turn off autocommit if commit was called explicitly.

elhigu:
@bnewtonius Is there a problem, when mssql autocommit is not enabled? Or are you referring with `autocommit` the knex doing commit / rollback, when promise is returned from transaction callback function?

If knex is really calling commit multiple times in case above, then knex should be fixed to do the extra commit query. 

So transaction should see if handler returned promise and if it did not, then knex should not try to do the commit / rollback automatically.

bnewtonius:
It's the latter. Knex commits when the promise chain resolves, regardless
of whether the user has called commit explicitly or not. I can look into
checking if the transaction has completed before running the promise
chain's commit after I get back from Thanksgiving break (Monday). Thank you
for responding, and sorry I didn't provide better detail in the original PR.

elhigu:
Sounds good,

> I can look into checking if the transaction has completed before running the promise chain's commit after I get back from Thanksgiving break (Monday).

There shouldn't be need to check if transaction has completed, but only to check if the transaction callback method returned promise or not. If it didn't return promise, then `knex` should not call commit / rollback explicitly.

PR #2340, opened 2017-11-18
Support multiple searchPaths while preserving case-sensitive feature …

### LINKED ISSUE #2335: knex now requires me to specify the schema
# Environment

Knex version: knex@0.14.0
Database + version: postgres (PostgreSQL) 10.1
OS: Archlinux Kernel 4.13.12-1-ARCH #1 SMP PREEMPT

# Bug ?

Not sure if it's a bug since I just did a big upgrade on my machine and switched from Pg 9.6 to 10.1, and from PostGIS 2.3 to 2.4, at the same time, since I found several issues not presented here, I ended upgrading from knex 0.13.0 to 0.14.0.

1. Old code which was working fine with previous mentioned versions of everything involved:

```
// my knex.js module to be required from other files
'use strict';

const config = require('config');
const dbConf = config.get('dbConnection');

  client: 'pg',
  debug: 'debug' === logLevel,
  searchPath: 'public,datasources',
  connection: {
    host: dbConf.host,
    port: dbConf.port,
    user: dbConf.user,
    password: dbConf.password,
    database: dbConf.db,
  },
});

knex.table = (table) => {
  return knex.withSchema(dbConf.schema).table(table);
};

module.exports = knex;
```

Then on a `users` model:
```
const knex = require('./knex');

const findByUserName = (name) => knex.select().from('users').where('name', name).first();
```

Now this was working fine before the upgrade, but now it gives me:

2. Error message

```
error: relation "users" does not exist
    at Connection.parseE (/home/flc/soft/webgis/map-server/node_modules/pg/lib/connection.js:546:11)
    at Connection.parseMessage (/home/flc/soft/webgis/map-server/node_modules/pg/lib/connection.js:371:19)
    at Socket.<anonymous> (/home/flc/soft/webgis/map-server/node_modules/pg/lib/connection.js:114:22)
    at emitOne (events.js:116:13)
    at TCP.onread (net.js:594:20)
  name: 'error',
  length: 104,
  severity: 'ERROR',
  code: '42P01',
  detail: undefined,
  hint: undefined,
  position: '13',
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'parse_relation.c',
  line: '1180',
  routine: 'parserOpenTable' }
```

which seems to be an error from `pg` module rather than `knex`, but still I have to ask here first.

In order to fix it I left my `knex.js` module as is, but changed the `users.js` module like this:

```
const knex = require('./knex');

const k = knex('public.users');

const findByUserName = (name) => k.select().where('name', name).first();
```

So it seems now I have to specify the schema, even the `public` one, in order to use knex.

I included a check in the users module, before defining `k`:

```
knex.raw('show search_path').then(r => console.log(r.rows));
```

which responds: 

```
[ anonymous { search_path: '"public,datasources"' } ]
```

which is fine and is as defined in my `knex.js` module.

So... What's going on?

3. I owe you a simple test case, will try to come up with a simplified version later.

wubzz:
…(by using Array syntax). Fixes #2335 

Docs: https://github.com/knex/documentation/pull/69

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 
 # Master (Unreleased)
+### Bug fixes:
+
+- Fix support for multiple schema names in in postgres `searchPath` #2340
 
 # 0.14.0 - 6 Nov, 2017
 
--- a/src/dialects/postgres/index.js
+++ b/src/dialects/postgres/index.js
@@ -1,10 +1,11 @@
 
 // PostgreSQL
 // -------
-import { assign, map, extend } from 'lodash'
+import { assign, map, extend, isArray, isString, includes } from 'lodash'
 import inherits from 'inherits';
 import Client from '../../client';
 import Promise from 'bluebird';
+import {warn} from '../../helpers';
 
 import QueryCompiler from './query/compiler';
 import ColumnCompiler from './schema/columncompiler';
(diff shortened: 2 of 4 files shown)


wubzz:
Docs: https://github.com/knex/documentation/pull/69

Issue #2312, opened 2017-11-07
truncate() on SQLite does not reset the id value in the sqlite_sequence any longer


lehni:
Upon upgrading to 0.14.0, I've realised that my primary keys don't get reset anymore when I use  `truncate()` with the SQLite dialect.

It looks like the code that's supposed to do the reseting in `truncate()` is actually producing an error:

https://github.com/tgriesser/knex/blob/bf8aa4db100ed34756d1d77268e3b391d5d2b528/src/dialects/sqlite3/query/compiler.js#L89-L99

If you replace `noop` there with `(err) => { console.log(err) }`, you will see the errors logged:

```
{ Error: SQLITE_ERROR: no such column: dummy errno: 1, code: 'SQLITE_ERROR’ }
```

It looks like the table-name should be quoted in that SQL statement. If I manually add quotes, then the ids get reset again:

```js
sql: `delete from sqlite_sequence where name = "${table}"`
```

I'm not able to say if that's the correct fix for this issue.

elhigu:
Thanks for report, looks like there are various places where identifiers are not wrapped properly.

richraid21:
This works on previous versions?

elhigu:
@richraid21 No. Those plain string replacements has been always there.

lehni:
Well... `truncate()` actually did work for me before, as in it was resetting the ids without problems until I upgraded to 0.14.0

elhigu:
Ah, ok it might have something to do with the change where sqlite identifier quoting was changed from `"`to backtick ``` ` ```. Old `"`  (https://github.com/tgriesser/knex/pull/2087).

with " -style quotes, non existent column "iDontExist" was interpret as string, instead of throwing error. (https://github.com/tgriesser/knex/issues/1048).

elhigu:
Did you check the query (sql string) was throwing the ```{ Error: SQLITE_ERROR: no such column: dummy errno: 1, code: 'SQLITE_ERROR’ }``` error? I was not able to reproduce this.

I mean  it really seems to set quotes correctly https://runkit.com/embed/g4idd055djit (I suppose this.tableName already had quotes around it). 

Maybe the code throws an error because some column / table which actually didn't exist earlier, now throws an error (check last comment in #1048)? And if you add `"${tableName}"` then sqlite doesn't throw an error anymore because ``` "`unknownIdentifier`" ``` actually is interpret as string  ``` `unknownIdentifier` ```.

Does this make sense?

[...]

PR #2348, opened 2017-11-23
Quick-fix for truncate() on sqlite3 dialect

lehni:
As discussed in https://github.com/tgriesser/knex/issues/2312#issuecomment-346549966, here the first commit as a  quick fix restoring previous functionality. A full fix for the other observed issues will follow later.

### DIFF
--- a/src/dialects/sqlite3/query/compiler.js
+++ b/src/dialects/sqlite3/query/compiler.js
@@ -87,12 +87,12 @@ assign(QueryCompiler_SQLite3.prototype, {
 
   // Compile a truncate table statement into SQL.
   truncate() {
-    const table = this.tableName
+    const { table } = this.single
     return {
-      sql: `delete from ${table}`,
+      sql: `delete from ${this.tableName}`,
       output() {
         return this.query({
-          sql: `delete from sqlite_sequence where name = ${table}`
+          sql: `delete from sqlite_sequence where name = '${table}'`
         }).catch(noop)
       }
     }
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -97,16 +97,30 @@ module.exports = function(knex) {
           tester('oracle', "truncate table \"test_table_two\"");
           tester('mssql', 'truncate table [test_table_two]');
         })
-        .then(function() {
-
+        .then(() => {
           return knex('test_table_two')
             .select('*')
-            .then(function(resp) {
+            .then(resp => {
               expect(resp).to.have.length(0);
             });
-
+        })
+        .then(() => {
+          // Insert new data after truncate and make sure ids restart at 1.
+          // This doesn't currently work on oracle, where the created sequence
+          // needs to be manually reset.
+          if (knex.client.dialect !== 'oracle') {
+            return knex('test_table_two').insert({ status: 1 })
+              .then(res => {
+                return knex('test_table_two')
+                  .select('id')
+                  .first()
+                  .then(res => {
+                    expect(res).to.be.an('object')
+                    expect(res.id).to.equal(1);
+                  });
+              });
+          }
         });
-
     });
 
     it('should allow raw queries directly with `knex.raw`', function() {


elhigu:
Also testcase for erroneus operation is missing.

lehni:
I am not sure, no. I don't understand Knex enough, and I don't have enough time to get into it at the moment. I also couldn't get the tests to run on my machine, so unfortunately I wasn't able to create a test for it. As it says, this is a quick fix for something that was broken between 0.13 and 0.14 (and there wasn't a test for it then). This is the best I can do currently.

lehni:
Regarding your question about the quotes, I guess single quotes would be better? Both seem to work. I'll adjust it.

lehni:
Regarding tests: If you can explain how to get them to run locally then I'll be happy to try and whip one up.

elhigu:
Single quotes are better because it doesn't have double meaning, where string may be interpreted as identifier or literal. Sqlite's `"` quotes work really strange way.

I always run tests locally with `npm run build; npm test` command, but it requires that one has mysql and postgres running with DB called `knex_test` and to mysql one should be able log in with user `root` without any password... but if you like to run just sqlite tests you can pass DB list in env var like `npm run build; DB="sqlite3" npm test`.

Contributing document has all that info too:

https://github.com/tgriesser/knex/blob/master/CONTRIBUTING.md

lehni:
When I tried to get the tests to run, I saw errors about docker, and couldn't find any information about that. Can I ignore that part?

elhigu:
yep you should not worry about those tests just ignore them. If docker host is not running, then those tests should not be ran either (I always shut down my docker host when im working on knex).

lehni:
I see. I misunderstood this error and though that was the reason why the tests weren't running. Getting into docker to fix a simple issue seemed rather daunting.

I just implemented a unit test for this now by adding to the existing unit test and checking that newly inserted data after truncate() restarts with id 1. I haven't tested this for anything else than sqlite. We'll see how this performs on Travis CI in a moment, I guess. The changed commit is up: 0ed3e659c6df835c88ef2e5b5de089a4f0dbe2c4

lehni:
BTW, it looks like tests are only running on Node up to v7. Shouldn't 8 and 9 be added to .travis.yml?

lehni:
Looks like the empty insert isn't working on oracle:

```
 insert into "test_table_two" ("undefined") values (default) - ORA-00904: "undefined": invalid identifier
``
`
I'll try again with one value.

elhigu:
> BTW, it looks like tests are only running on Node up to v7. Shouldn't 8 and 9 be added to .travis.yml?

Last time I tried there was some problems with node 8...  If it they work now indeed they should be added.

lehni:
Ok, now the tests all run, but again on Oracle, the ids don't restart at 1... Does that mean truncate() is broken there too, or is that just how it is? I don't know Oracle. Not sure what to do about this now?

elhigu:
Sounds like oracle truncate implementation doesn't work correctly. There was also some other strange errors. my new connection pool tests might be random failing :| Gotte checkout oracledb code if its truncate seems reasonable, but now I need to go to sleep :) laters and thanks for putting time to this!

lehni:
Oracle id reset sounds super complicated... I am excluding Oracle from the new test for now, as I have no idea how to implement it there, and no interest in fixing Oracle support:

https://stackoverflow.com/questions/51470/how-do-i-reset-a-sequence-in-oracle

elhigu:
@lehni yeah fixing oracle is clearly out of scope of this PR. @atiertant do you know if oracle's truncate also truncates the sequence?

elhigu:
Thank you @lehni !

lehni:
Cool, thanks for merging!

elhigu:
I'll do the release when I get spare moment.

lehni:
Thanks! That way I can let go of my internal workarounds : )

elhigu:
done

lehni:
thanks!

atiertant:
@elhigu need a test

elhigu:
@atiertant You mean test like this? https://github.com/tgriesser/knex/pull/2348/files#diff-bfeb3d958a56cecbae4f5f8983fdf195R111

elhigu:
oracle is disabled there, because it didn't seem to reset id sequence

atiertant:
@elhigu ok for the test, i'll try to have a look a it later but i don't have time this days...

Issue #2351, opened 2017-11-23
knex has the wrong exit status on bad command


P-Seebauer:
# Environment

Knex version: 0.12.9 and 14.1
Database + version: does not matter
OS: ubuntu

# Bug
1. Explain what kind of behaviour you are getting and how you think it should do

knex (-cli) does exit with 0 if the command supplied was wrong.
It should exit with a non-zero exit code, except you state in your documentation something like 

```knex ⟨anything else⟩ print this help message and exit.```

2. Error Message
None, if you don't count the help text. This is a bug about a lack of errors.

3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real
   database connection to mysql or postgresql, then single file example which initializes
   needed data and demonstrates the problem.

just run 
```
knex foo;
echo $?;
```
It should **not** print 0

elhigu:
Looks like this is only happening on **unknown commands**, not for example in those cases when invalid parameters are given or when command fails. 

I have hard time to believe that anyone is currently relaying on this functionality (just printing knex client help texts and passing also invalid command string in) so dropped the `breaking change` flag from this.

P-Seebauer:
No, I'm not relying on it, I just had a typo in a CI script was looking in the wrong spot because of that. It's probably easily fixable. Maybe I'll give it a try.

PR #2358, opened 2017-11-28
2351 CLI sets exit-code 1 if the command supplied was not parseable

P-Seebauer:
See issue #2351.

Sorry, I haven't found tests for the `bin/cli.js` file.

[Reference to a stack exchange question, that is somewhat related ☺](https://unix.stackexchange.com/questions/405783/why-does-man-print-gimme-gimme-gimme-at-0030).

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -13,9 +13,9 @@ var cliPkg = require('../package');
 
 function exit(text) {
   if (text instanceof Error) {
-    chalk.red(console.error(text.stack));
+    console.error(chalk.red(text.stack));
   } else {
-    chalk.red(console.error(text));
+    console.error(chalk.red(text));
   }
   process.exit(1);
 }
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -186,7 +186,8 @@ function invoke(env) {
   commander.parse(process.argv);
 
   Promise.resolve(pending).then(function() {
-    commander.help();
+    commander.outputHelp();
+    exit('Unknown command-line options, exiting');
   });
 }
 


elhigu:
> Sorry, I haven't found tests for the bin/cli.js file. 

There hasn't been tests for `cli.js` so far. Someone has to be the first one to write them 😅

bin/cli.js
```diff
@@ -186,7 +186,8 @@ function invoke(env) {
   commander.parse(process.argv);
 
   Promise.resolve(pending).then(function() {
-    commander.help();
+    commander.outputHelp();
+    exit('');
```
elhigu: Maybe some more specific error message about command not found could be given.

P-Seebauer: Done. Also realized then that chalk and console.error were called in the wrong order…

Also wasn't sure what your policy is regarding rebasing/amending. 
Sorry if I got that wrong.

elhigu: Eventually PR goes in as a squash merge so rebase/amend/etc. doesn't really matter anything goes.

P-Seebauer:
The failed integration doe snot look like it was caused by my changes (failed to install oracledbinstall)
is there a way to restart the CI?

elhigu:
@P-Seebauer There was a problem that oracle stopped providing binaries to some node versions. After rebase tests should run fine.

P-Seebauer:
I rebased and repushed. 
But it does seem like travis did not want to run the tests…

https://travis-ci.org/tgriesser/knex/builds/329612776

PR #2364, opened 2017-12-01
[WIP] Check constraints

richraid21:
Adds check constraint methods to both column and table level APIs

See [Postgres Check Constraint Docs](https://www.postgresql.org/docs/9.4/static/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS)

Use case: 
```javascript
await knex.schema.createTable('_test', (t) => {
        t.increments()
        t.text('name').notNullable().check('name != NULL') // Stupid example, but shows chaining
        t.date('created')
        t.date('updated')
        
        t.check('updated >= created')
    })
```

**TODO**

- [ ] Determine name scheme for supporting multiple table-level constraints
- [ ] Determine API for named and default-named constraints
- [ ] Postgres Implementation https://www.postgresql.org/docs/9.4/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS
- [x] ~~Mysql Implementation~~ [Unsupported](https://dev.mysql.com/doc/refman/5.7/en/create-table.html) See "Check Constraints" https://dev.mysql.com/doc/refman/8.0/en/create-table-check-constraints.html
- [ ] Maria Implementation
- [ ] SQL Server Implementation https://www.w3schools.com/sql/sql_check.asp
- [ ] Oracle Implementation https://docs.oracle.com/cd/E17952_01/mysql-8.0-en/create-table-check-constraints.html

- [ ] Tests for all above
- [ ] Documentation

Still need to write tests and documentation

- Closes https://github.com/knex/knex/issues/266

### DIFF
--- a/src/dialects/postgres/schema/tablecompiler.js
+++ b/src/dialects/postgres/schema/tablecompiler.js
@@ -90,6 +90,11 @@ TableCompiler_PG.prototype.comment = function(comment) {
   this.pushQuery(`comment on table ${this.tableName()} is '${this.single.comment}'`);
 };
 
+TableCompiler_PG.prototype.check = function(logic) {
+  const name = this.formatter.wrap(`${this.tableNameRaw}_check`);
+  this.pushQuery(`alter table ${this.tableName()} add constraint ${name} check (${logic})`);
+};
+
 // Indexes:
 // -------
 
--- a/src/dialects/postgres/schema/columncompiler.js
+++ b/src/dialects/postgres/schema/columncompiler.js
@@ -10,7 +10,7 @@ import { assign } from 'lodash'
 
 function ColumnCompiler_PG() {
   ColumnCompiler.apply(this, arguments);
-  this.modifiers = ['nullable', 'defaultTo', 'comment']
+  this.modifiers = ['nullable', 'defaultTo', 'comment', 'check']
 }
 inherits(ColumnCompiler_PG, ColumnCompiler);
 
--- a/src/schema/columnbuilder.js
+++ b/src/schema/columnbuilder.js
@@ -22,7 +22,7 @@ export default function ColumnBuilder(client, tableBuilder, type, args) {
 // All of the modifier methods that can be used to modify the current query.
 const modifiers = [
   'default', 'defaultsTo', 'defaultTo', 'unsigned',
-  'nullable', 'first', 'after', 'comment', 'collate'
+  'nullable', 'first', 'after', 'comment', 'collate', 'check'
 ];
 
 // Aliases for convenience.
--- a/src/schema/columncompiler.js
+++ b/src/schema/columncompiler.js
@@ -130,6 +130,11 @@ ColumnCompiler.prototype.specifictype = type => type;
 // Modifiers
 // -------
 
+// Compiles the Check constraint on the table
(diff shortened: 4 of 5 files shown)


src/dialects/postgres/schema/tablecompiler.js
```diff
@@ -90,6 +90,11 @@ TableCompiler_PG.prototype.comment = function(comment) {
   this.pushQuery(`comment on table ${this.tableName()} is '${this.single.comment}'`);
 };
 
+TableCompiler_PG.prototype.check = function(logic) {
+  const name = this.formatter.wrap(`${this.tableNameRaw}_check`);
```
elhigu: I think this naming convention limits number of check constraints per table to be 1.

src/dialects/postgres/schema/tablecompiler.js
```diff
@@ -90,6 +90,11 @@ TableCompiler_PG.prototype.comment = function(comment) {
...
+  this.pushQuery(`alter table ${this.tableName()} add constraint ${name} check (${logic})`);
```
elhigu: needs to support raw() syntax, or otherwise identifier/value bindings cannot be handled properly across the dialects

elhigu:
And there was some indentation error which caused CI to fail :)

richraid21:
@elhigu 
Regarding `raw` syntax, after some testing, it seems like `pg` doesn't support this. 

The following all fail with assorted errors being thrown in `pg`: 

```javascript
await knex.raw('alter table _test3 add constraint ? check (length(name) >= 3)', ['name_check'])
// or
await knex.raw('alter table _test3 add constraint name_check check (length(?) >= 3)', ['name'])
// or
await knex.raw('alter table _test3 add constraint name_check check (length(?) >= ?)', ['name', 3])
// or
await knex.raw('alter table _test3 add constraint ? check (length(?) >= ?)', ['name_check', 'name', 3])
```

elhigu:
Probably all of those examples already throw an error because PG does not use `?` for values bindings. Also identifiers are quoted / wrapped by knex, they are not passed as bindings to drivers. 

e.g. 
```js
raw('?? = ?', ['colName', 'foo'])
``` 

will pass string 

```
"colName" = $1` bindings: ['foo']
``` 
to the driver.

But it is necessary that knex can do its wrapping to make quotes right (and make use of custom identifier wrapping if it is set in knex configuration).

GAumala:
A MySQL implementation should also be included. I'm aware that MySQL parses `CHECK` statements but ignores them, however, [MariaDB doesn't have this limitation as of version 10.2.1](https://mariadb.com/kb/en/library/constraint/#check-constraints). Knex with MySQL driver works seamlessly with MariaDB, so the implementation would still be useful.

I have one question though, is it currently possible to implement the `CHECK` statement with a raw query without having to do the **whole** table declaration as a raw query? Something like:

``` javascript
await knex.schema.createTable('_test', (t) => {
        t.increments()
        t.date('created')
        t.date('updated')
        
        t.raw('check updated >= created')
    })
```

kibertoad:
@richraid21 Are you still interested in finishing this?

OlivierCavadenti:
Superseded by https://github.com/knex/knex/pull/4874

PR #2366, opened 2017-12-03
added editorconfig for consistent indentation

adeyahya:


elhigu:
> added editorconfig for consistent indentation

This PR does more than that. Please elaborate. 

Also there is `package-lock.json` committed, which I would usually prefer to not be included with every PR, because it makes PRs more difficult to review and causes conflicts with other PRs. But since this is the first PR adding that file it is fine.

PR #2372, opened 2017-12-08
Incorrectly set UV_THREADPOOL_SIZE

tomulin1:
Incorrectly set UV_THREADPOOL_SIZE when passed from system variables. It's recognized as string (windows).


elhigu:
Thanks! I believe that code is working wrong on every platform. I'll merge this after getting last tests run.

PR #2376, opened 2017-12-10
fix #2351

### LINKED ISSUE #2351: knex has the wrong exit status on bad command

godfrey3835:


bin/cli.js
```diff
@@ -186,7 +186,9 @@ function invoke(env) {
   commander.parse(process.argv);
 
   Promise.resolve(pending).then(function() {
-    commander.help();
+    commander.outputHelp()
+  }).then(function () {
```
elhigu: `.then` here doesn't really add anything. it would be the same / better to put exit in the same block with help printing.

elhigu:
Duplicate https://github.com/tgriesser/knex/pull/2358/

Issue #2285, opened 2017-10-23
Support for sqlite3-offline


hetsch:
Hi,

while trying to compile sqlite3 for my electron app (and miserably failing as many others ...), I stumbled over an issue where a dude provides prebuilt binaries for sqlite3 and offline app usage. The original thread can be found at https://github.com/mapbox/node-sqlite3/issues/840 and his repository is located at https://github.com/DenisCarriere/sqlite3-offline.

I wanted to ask if there is any chance to support his package or if you can think of any workaround to get it running with knex.

Thank's!

elhigu:
You can hack it like this (in client arg one can actually give the Dialect object that knex should use):

```js
require('sqlite3-offline');

const Knex = require("knex");

const Dialect = require(`knex/lib/dialects/sqlite3/index.js`);
Dialect.prototype._driver = () => require('sqlite3-offline');

const knex = Knex({
  client: Dialect,
  connection: ':memory:'
});

console.log(
  knex.select(knex.raw(1)).toSQL()
);

await knex.schema.createTable('fooobar', (t) => {
  t.bigincrements('id');
  t.string('data');
});
await knex('fooobar').insert({ data: 'nomnom' });

console.log("Gimme all the data:", await knex('fooobar'));
```

https://runkit.com/embed/90b3cpyr4jh2

hetsch:
Yes! Thank you for quick response. This was exactly what I needed!

PR #2381, opened 2017-12-12
[WIP] Added readonly BigQuery dialect

jessezhang91:
Hi, I've added a dialect to send queries to BigQuery using `@google-cloud/bigquery`.

There are a few things I wanted to ask before continuing further with this feature:
- Since my only use case is to use knex as a query builder and executor for SELECT queries, I've really only implemented a way to execute the SELECT queries. From what I understand, the "data manipulation language" for BigQuery isn't really recommended and the schema management/migration isn't SQL but REST API calls. Is this okay?
- There is a way to get `columnInfo` from the BigQuery client but not via some SQL expressions. Are there any recommendations for how to implement this or is this even necessary given the read-only use case?
- From my understanding, the BigQuery library is just a wrapper on top of their REST endpoints. Because of this, reusing a connection from the pool is not really necessary. Is there any way I can simplify the "fake connection" object that I'm doing?
- Are there any tests I should be writing? The SELECT queries are pretty much the same as MySQL. The integration tests would require some private key to BigQuery to execute. I have manually tested it to ensure the query and stream functions work.

Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "v8flags": "^3.0.0"
   },
   "devDependencies": {
-    "@google-cloud/bigquery": "^0.11.1",
+    "@google-cloud/bigquery": "^0.12.0",
     "JSONStream": "^1.0.3",
     "async": "^0.9.0",
     "babel-cli": "^6.11.4",
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "v8flags": "^3.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


elhigu:
Please implement this as a separate npm library and use it as a custom dialect by passing it to knex configuration like this: https://github.com/tgriesser/knex/issues/2285

If dialect is fully implemented and is good quality + well tested (also with CI) it may be included to knex.

jessezhang91:
Ok, I'll aim to do that then. Thanks!

PR #2388, opened 2017-12-16
Set toNative() to be not enumerable

### LINKED ISSUE #2344: Make .toNative() not be enumerable property
# Environment

Knex version: > 0.14.0
Database + version: all
OS: osx

# Bug

When `.toNative()` function was added to `.toSQL()` result, it was not set to be non-enumerable. To prevent it from showing as normal attribute `Object.defineProperties` can be used to make it behave more like class methods.

godfrey3835:
This would fix #2344

### DIFF
--- a/src/query/compiler.js
+++ b/src/query/compiler.js
@@ -52,12 +52,20 @@ assign(QueryCompiler.prototype, {
       timeout: this.timeout,
       cancelOnTimeout: this.cancelOnTimeout,
       bindings: this.formatter.bindings || [],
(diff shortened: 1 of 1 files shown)


elhigu:
Thanks! I'll merge after having successfull test run.

elhigu:
Looks like oracle dropped support for node 5,7 and will drop 9 also soon https://github.com/oracle/node-oracledb/issues/811 so I need to fix test setup to not install oracle stuff on non lts node versions..

PR #2399, opened 2017-12-30
Fixed a bug when using .returning (OUTPUT) in an update query with joins in MSSQL

tankers746:
When using OUTPUT with updates and a join in MSSQL the OUTPUT needs to come before
the joins.

https://stackoverflow.com/questions/4170389/inner-join-within-update-output-in-one-t-sql-query

### DIFF
--- a/src/dialects/mssql/query/compiler.js
+++ b/src/dialects/mssql/query/compiler.js
@@ -87,8 +87,8 @@ assign(QueryCompiler_MSSQL.prototype, {
     return {
       sql: this.with() + `update ${top ? top + ' ' : ''}${this.tableName}` +
         ' set ' + updates.join(', ') +
+        (returning ? ` ${this._returning('update', returning)}` : '') +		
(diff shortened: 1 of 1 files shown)


Grimace1975:
Thabks for this fix, however I do not see a test included to make sure this is caught in the future.

Can you please add a test, also looks like the build is currently breaking for this commit. I am not at a computer rn to figure out why.

Can you make the build not break.

tankers746:
It looks like the master branch is failing to build due to an issue installing oracledb.
Apologies, I'm not sure how I'd go about making a test.

Grimace1975:
Ok I’ll see if I can make one for you and merge this in.

Sent from my iPhone

On Dec 30, 2017, at 11:31 AM, tankers746 <notifications@github.com<mailto:notifications@github.com>> wrote:


It looks like the master branch is failing to build due to an issue installing oracledb.
Apologies, I'm not sure how I'd go about making a test.

PR #2405, opened 2018-01-02
Use wrapIdentifier in columnInfo. fixes #2402

### LINKED ISSUE #2402: columnInfo is not using wrapIdentifier
we are using automatic camelCase transition for tableName and columns as well.
but columnInfo is not changing tableName https://github.com/tgriesser/knex/blob/7de07c966a6ab63087b31cbdd10216d3ddbf8230/src/dialects/postgres/query/compiler.js#L79

koskimas:
Hi,

currently `columnInfo` doesn't respect the `wrapIdentifier` function if one is provided in the configuration. This PR fixes that.

fixes #2402

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -171,10 +171,14 @@ assign(Client.prototype, {
   },
 
   wrapIdentifier(value) {
+    return this.customWrapIdentifier(value, this.wrapIdentifierImpl);
+  },
+
+  customWrapIdentifier(value, origImpl) {
     if (this.config.wrapIdentifier) {
-      return this.config.wrapIdentifier(value, this.wrapIdentifierImpl);
+      return this.config.wrapIdentifier(value, origImpl);
     }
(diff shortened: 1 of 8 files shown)


src/client.js
```diff
@@ -171,10 +171,14 @@ assign(Client.prototype, {
   },
 
   wrapIdentifier(value) {
+    return this.customWrapIdentifier(value, this.wrapIdentifierImpl);
+  },
+
+  customWrapIdentifier(value, origImpl) {
```
elhigu: Kind of hacky, but works at least as long as user doesn not implement their custom quote style too in their wrapIdentifier configuration. Maybe knex would need this approach to be more official and to have separate APIs to override for quoting and for transforming identifiers. Anyways this is good enough for now.

koskimas: Yeah, but there is no other way to do this right now.

PR #2433, opened 2018-01-18
Update dependencies and fix ESLint warnings accordingly

igor-savin-ht:


igor-savin-ht:
@elhigu Would be great if you could take a look at https://github.com/tgriesser/knex/pull/2433/commits/aaa4d71e914faa1c4c1313c26bbdd26067b0b448 -> I've removed it as ESLint demanded me to do so, but I wonder if there was actually a bug there and that param was supposed to be used.

igor-savin-ht:
@elhigu I can restore removed code with eslint ignore and add the ToDo to review it later, if it's not an option to check it now :)

elhigu:
@igor-savin-ht honestly I just haven't got time to read this through with thought. Smaller changes are easier to accept with confidence that they are not breaking anything :) I would prefer eslint ignore for now.

igor-savin-ht:
@elhigu Done!

elhigu:
There were few regex where some escape backslashes were dropped. Are you sure they work exactly the same after the changes?

igor-savin-ht:
@elhigu ESLint was complaining about those escape backslashes being redundant, I tend to trust its judgement on that. Here is an in-depth discussion on this subject: https://github.com/eslint/eslint/issues/7656
As is with unused variable, I'm open to replacing this change with either line-specific ignore or global rule to ignore useless escapes.

elhigu:
I'd rather have that global rule set so I need to verify manually e.g. with runkit that each of those regexps are working as expected after change. In this case I don't trust knex testsuite enough and I believe that eslint has them correctly fixed, but I need to be sure.

src/query/string.js
```diff
@@ -153,7 +153,7 @@ function convertTimezone(tz) {
   if (tz === 'Z') {
     return 0;
   }
-  const m = tz.match(/([\+\-\s])(\d\d):?(\d\d)?/);
+  const m = tz.match(/([+\-\s])(\d\d):?(\d\d)?/);
```
elhigu: Actually I wonder why eslint left `\-`escaping there.

elhigu:
All good https://runkit.com/embed/ygveesp9rad2

PR #2444, opened 2018-01-30
fixed issue when adding "real" column type to table

nickgzzjr:
Before this fix, when I would use the function `table.real(columnName)` node would throw the error `Unhandled rejection error: type "undefined" does not exist` I realized that the `ColumnCompiler` was missing the `real` prototype. After adding it, it worked fine. Test is included. I verified the test failed before my change and passed after my change.

### DIFF
--- a/src/schema/columncompiler.js
+++ b/src/schema/columncompiler.js
@@ -102,6 +102,7 @@ ColumnCompiler.prototype.varchar = function(length) {
 };
 ColumnCompiler.prototype.text = 'text';
 ColumnCompiler.prototype.tinyint = 'tinyint';
+ColumnCompiler.prototype.real = 'real';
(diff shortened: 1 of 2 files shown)


elhigu:
`.float()` creates 4 byte floating point number (real in postgres). Better not to add more types knex which are not compatible between different databases. If one needs to use some database specific type, there is .specificType for that.

Example: https://runkit.com/embed/c2ohhndbb6j7

nickgzzjr:
Yes I am aware of .specificType, and my script will use .specificType when it can't find the type in the table prototype. So perhaps you should consider removing 'real' from columnTypes in tablebuilder.

elhigu:
I would accept PR removing it from the list any time. It is not even breaking change, since it has never produced valid SQL in any dialect.

PR #2449, opened 2018-01-31
feat: countDistinct with multiple columns

### LINKED ISSUE #1776: countDistinct support for multiple columns
countDistinct currently only supports counting the distinct values in a single column, while the distinct method supports finding distinct rows using multiple attributes. It would be great if countDistinct also supported multiple attributes.

With a postgres backend, the following happens

```js
knex('customers')
  .distinct('first_name', 'last_name')
  .select();
// select distinct "first_name", "last_name" from "customers"

knex('customers')
  .countDistinct('first_name', 'last_name');
// select count(distinct "first_name") from "customers"
```

It'd be nice if instead the countDistinct query produced
`select count(distinct("first_name", "last_name")) from "customers"`

joelmukuthu:
Closes #1776


test/unit/query/builder.js
```diff
@@ -2498,6 +2498,31 @@ describe("QueryBuilder", function() {
     });
   });
 
+  it("count distinct with multiple columns", function() {
```
elhigu: +1 for writing test first <3 Also integration smoke test would be needed though to make sure that syntax is correct.

joelmukuthu:
Sorry for the delay on this, but I'm looking to implement it now and found a weird edge-case. At the moment, you can specify an alias for `countDistinct` like `builder.countDistinct('column as alias')` and that get's transformed to `select count(distinct column) as alias`.

This of course won't work with multiple fields so I suggest we change this to the regular object notation like `builder.countDistinct({ alias: 'column' })` or with multiple fields `builder.countDistinct({ alias: ['column1', 'column2'] })`.

I'll leave the current behaviour and only add the object notation as a new feature but maybe we can even add a deprecation notice for it. What do you think @elhigu?

elhigu:
I'm thinking about what usually `{ alias: ['foo', 'bar']}` syntax should do. I don't know if there is any use cases which should be left untouched. Are you planning to add special case how that syntax works only with `.countDistinct` or general change how `{alias:[]}` is interpreted?

@joelmukuthu I'm sorry  I'm not sure if I understood what are you planning to do and what effects it will have to knex backwards compatibility? :)

joelmukuthu:
I could be missing something but as far as I could tell the parts I touched are only dealing with building statements of type `aggregate` (excluding `aggregateRaw`) so the new `{alias: [column1, column2]}` syntax would only be used for those. I think that makes sense since you can't do 
```
select (column1, column2) as alias from table
``` 
(or is it possible?) but you can definitely do 
```
select sum(column1, column2) as alias from table
```

I made sure that the changes don't break the current behaviour, so for instance this test still passes: https://github.com/joelmukuthu/knex/blob/a3a5255ecbb2b4613eab21c237473d407f42591b/test/unit/query/builder.js#L2878-L2905.

FYI I decided to not bother fixing that failing mariasql test since it seems the dialect is being dropped anyway

joelmukuthu:
This is ready for review. Notes:

- Added support for multiple calls for all aggregate methods:
```js
// single column:
builder.count('column1', 'column2');
// multiple columns:
builder.countDistinct('column1', 'column2');
``` 
- Added support for object alias notation for all aggregate methods:
```js
// single column:
builder.count({ alias: 'column' });
// multiple columns:
builder.countDistinct({ alias: ['column1', 'column2'] });
```
- Kept support for string alias notation, though this only works with one column:
```js
builder.count('column as alias');
```

I skipped validating whether or not the dialect supports calling whatever aggregate function with multiple columns i.e. no error is thrown if a user does something they're not supposed to do like `countDistinct` with multiple columns if the database doesn't support it.

I also skipped integration tests for aggregates with multiple columns for these dialects.

joelmukuthu:
Docs: https://github.com/knex/documentation/pull/97

elhigu:
I think this  is good, sorry for taking so much time lookin these. I've been even more busy lately than I was before 😬

joelmukuthu:
@elhigu no worries, thanks for taking a look! We appreciate the hard work you put in :)

kibertoad:
@jrschumacher No, MySQL and PostgreSQL only, judging by tests. Does SQLite support all necessary features for this?

jrschumacher:
@kibertoad after some tests:

```
select count(distinct 'CustomerThing'.'Customer_id', 'CustomerThing'.'thing') from 'CustomerThing' - wrong number of arguments to function count()
```

```
select count(distinct('CustomerThing'.'Customer_id', 'CustomerThing'.'thing')) from 'CustomerThing' - row value misused
```

Works
```
select count(*) from (
  select distinct "CustomerThing"."Customer_id", "CustomerThing".thing from "CustomerThing"
) "CustomerThing"
```

kibertoad:
@jrschumacher Would you consider raising a PR with a fix and expanded test?

jrschumacher:
@kibertoad Absolutely, just wanting to get some input. I'm not familiar with the use of distinct. I'm working on bookshelf/bookshelf#1936

Issue #2442, opened 2018-01-29
Leaked exception if database crashes. (MySQL server has gone away)


conglq:
# Environment

Knex version:0.14.2
Database + version: mariaDb v10.2
OS: Ubuntu
Node: v9.3.0

My connection:
```
const Knex = require('knex')({
  client: 'mariasql',
  connection: {
    host: config.host,
    user: config.user,
    password: config.password,
    db: config.database,
    charset: 'utf8',
    pool: { min: 2, max: 10 },
    acquireConnectionTimeout: 30000
  }
});
```

After period of time my node app crash with error: 
```
events.js:136
      throw er; // Unhandled 'error' event
      ^

Error: MySQL server has gone away
```
Please help me resolve this issue. Thanks all!

elhigu:
https://dev.mysql.com/doc/refman/5.7/en/gone-away.html

elhigu:
if you have pool min size 0 does it still happen? Does it happen if you are not making any queries, but just initialise the kenx instance?

conglq:
@elhigu It still happend when I set pool min size 0. It happen when I am not making any queries after period of time.

elhigu:
@conglq what period of time are we talking to? 1 minute, 1 hour, 1 day?

elhigu:
So should this be enough to reproduce the problem?

```js
const Knex = require('knex')({
  client: 'mariasql',
  connection: {
    host: config.host,
    user: config.user,
    password: config.password,
    db: config.database,
    charset: 'utf8',
    pool: { min: 0, max: 10 },
    acquireConnectionTimeout: 30000
  }
});

async function main() {
    await knex.raw('select 1');
    await periodOfTime();
    await knex.raw('select 1');
}   

main();
```

wubzz:
Just a wild guess, but could this be due to not specifying `evictionRunIntervalMillis` causing connections to never be evicted? Like the pool keeps the connections eventhough the mysql server has closed them.

elhigu:
@wubzz I thought the default pool setting `testOnBorrow: true` should take care of that case that dead connection is not returned from pool (if knex get the event of closed connection from the driver). It would be interesting to see if setting the `evictionRunIntervalMillis` makes a difference 👍  One more thing that could happen is that connection is fetched from pool for example for transaction and never returned. In that case dead connection probably would throw the error mentioned in OP (I don't believe that this is the problem here though).

conglq:
[...]

Issue #2339, opened 2017-11-18
Lets use Tarn.js as a pool instead of generic-pool


elhigu:
# Environment

Knex version: > 0.8
Database + version: all
OS: all

# Feature discussion / request

Knex uses currently `generic-pool`, which has been source of multiple problems and it is not very actively maintained (author is not replying to issues / pull requests). I opened this issue, after I noticed that the `testOnBorrow` feature, which is needed by `knex` didn't have any tests (had one that was just commented out).

Also the `generic-pool` is lacking tests, which means we cannot trust that it works correctly when it is updated. For example `testOnBorrow` feature which should be used to prevent returning dead connections (e.g. closed by database)

I would like to open up discussion https://github.com/tgriesser/knex/pull/1702 again about using own pool component  [Tarn.js](https://github.com/Vincit/tarn.js), which we can update / fix when ever there is need and whose API can be extended to support `knex` usage so that we don't need to write work arounds to support `generic-pool`.

Some related issues: 
https://github.com/tgriesser/knex/issues/2321
https://github.com/tgriesser/knex/issues/1833
https://github.com/tgriesser/knex/issues/2325

wubzz:
From briefly looking at `Tarn.js` docs it seems to cover everything we're currently relying on from `generic-pool`, even _(almost)_ under the same config syntax, which is great. I have to agree that every upgrade of `generic-pool` has resulted in at least one or more major bugs in a release, and for that reason I strongly support this request.

I think for me `generic-pool` crossed the line when the [upgrade guide](https://gist.github.com/sandfox/5ca20648b60a0cb959638c0cd6fcd02d) from version `2` -> `3` failed to mention the complete removal of default `reapIntervalMillis` which heavily breaks existing code... This in turn resulted in a `knex` docs change/warning.

@tgriesser You were against this previously in hopes of instead having the user supply method of pooling. While that is still a good idea, perhaps we should still make one final "swap" of pool libs before that time comes?

richraid21:
If this change is going to occur, it might be a great moment to add support for different read/write endpoints. For high availability systems, this need is crucial.  

Accomplishing this at the pool level would be the most robust solution.

elhigu:
@richraid21 that is a separate issue, better to talk about it in its own ticket https://github.com/tgriesser/knex/issues/2253. Starting do specify there how that master/slave configuration should actually work and what kind of APIs knex would provide for it would be a good start.

tuananh:
[...]

PR #2450, opened 2018-02-02
Use tarn as pool

koskimas:
Hi,

This experimental PR replaces the current pool `generic-pool` with [`Tarn`](https://github.com/vincit/tarn.js). There has been a bunch of discussion about this change in the following issues: https://github.com/tgriesser/knex/issues/2339 https://github.com/tgriesser/knex/pull/1702

At this point the pull request is here mainly so that some brave soul, who is having problems with the current pool, could test this out to see if it solves their problems.

Also fixes https://github.com/tgriesser/knex/issues/2442

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -16,13 +16,12 @@ import TableCompiler from './schema/tablecompiler';
 import ColumnBuilder from './schema/columnbuilder';
 import ColumnCompiler from './schema/columncompiler';
 
-import * as genericPool from 'generic-pool';
-import * as genericPoolErrors from 'generic-pool/lib/errors'
+import { Pool, TimeoutError } from 'tarn';
 import inherits from 'inherits';
 import { EventEmitter } from 'events';
 
 import { makeEscape } from './query/string'
-import { assign, uniqueId, cloneDeep, defaults, get } from 'lodash'
+import { assign, uniqueId, cloneDeep, defaults } from 'lodash'
 
 const debug = require('debug')('knex:client')
 const debugQuery = require('debug')('knex:query')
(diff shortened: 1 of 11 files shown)


src/client.js
```diff
@@ -309,24 +288,38 @@ assign(Client.prototype, {
...
-          this.pool = void 0
-          if(typeof callback === 'function') {
-            callback();
-          }
-        })
-    );
+    let promise = null
+
+    if (this.pool) {
+      promise = this.pool.destroy()
```
koskimas: `destroy` drains and clears the pool just like `drain` + `clear` with the old pool.

koskimas:
Oh, knex still tests agains node 4 and 5. Tarn currently requires 6. That should be easy to fix. There's something else wrong with the tests too. Funny that they all passed on my computer :smile: I'll look into it.

igor-savin-ht:
@koskimas Node 5 is EOL already, Node 4 wil be in 2 months. Do we still want to support them?

koskimas:
@igor-savin-ht It's not up to me. There are only a couple of things in Tarn that require 6. I can easily change those.

igor-savin-ht:
@elhigu Any plans to drop Node 4/5 support in the nearby future? Given that Node.js migration path is really smooth considered to other stacks, I wonder if there are any reasons to support legacy versions for a long time.

koskimas:
There's one test that fails and only on oracle

```
oracledb - acquireConnectionTimeout works
```

I have no idea what that's about. @elhigu Any ideas?

elhigu:
I think we can drop node 4 already () and later on drop support always when node version reach EOL. @koskimas  no idea why that could be failing... it haven't failed before, need to check if test is done wrong or if that timeout is really not working with oracle...

igor-savin-ht:
@elhigu Created a PR for that. Do you want anything in knex/documentation changed as well?

koskimas:
@elhigu I changed that test so I probably broke it, but I did nothing oracle specific in this PR. I'll try to debug that later.

elhigu:
@igor-savin-ht yes please. I also discussed with @koskimas about the changing this PR to work with Node 4 and 5 after all so that there will be at least one version in knex 0.14.x that works and that will be last version supporting node 4.

koskimas:
@elhigu Tests now pass

elhigu:
knex-0 14 2 master

15mins 100MB -> ~250MB

tarn-branch

45mins stable ~105MB

koskimas:
@elhigu I added warnings about the old generic-pool config options

koskimas:
Now that this is merged, who should I give write access to tarn? I'll of course help maintain it too, but I also want to give full access for knex maintainers.

elhigu:
At least for @wubzz and @tgriesser I think. For the rest of the people access may be given on demand when they happen to need them.

wubzz:
I believe we also need to update the *Upgrading* section of the docs:

```
Upgrading 0.13 -> 0.14
generic-pool was upgraded to v3. If you have specified idleTimeoutMillis or softIdleTimeoutMillis in the pool config then you will need to add evictionRunIntervalMillis: 1000 when upgrading to 0.14.

See original issue #2322 for details.
```

Since technically 0.13 -> 0.14 now includes not only `generic-pool` v3 but also `tarn.js` 1.1.2. For better or for worse `tarn.js` uses the old `idleTimeoutMillis` instead of `evictionRunIntervalMillis` *(don't get me wrong, I definitely prefer the old property)*, so the upgrading guide needs to say the reverse of what it's saying right now *(evictionRunIntervalMillis back to idleTimeoutMillis)*.

Just want to double-check with @koskimas @elhigu that I'm right about this before I make an issue in docs repo.

elhigu:
Yup I think we should remove that upgrading section and  add warning about using versions 0.14.0 - 0.14.2

pablopen:
I hope you dont drop support for node 4. I know it only has 2 months left for LTS, but it's really neccessary?
And yes, this will affect me :( I'm using node 4.6 with oracledb right now, and node migration doesn't depend on me,

igor-savin-ht:
@pablopen Given how smooth and painless is node upgrade process is, what is the rationale of whoever is responsible for making the decision not to do it? We can't support obsolete versions forever, so sooner or later update needs to happen anyway.

elhigu:
@pablopen it wont be dropped before 0.15 versions. 0.14.4 will be most stable knex released so far you can still use it after node 4 is deprecated. I’m pretty sure oracledb will drop support for node 4 too around the same time (they already dropped 5 and 7). So if you are not able to upgrade node you need to relay in older package versions.

bertho-zero:
The doc is not updated here: http://knexjs.org/#Installation-pooling

elhigu:
@bertho-zero good catch. Thanks

irsl:
Only partially related: I'm wondering if the Sqlite adapter could now be improved so that read-only queries would not be blocked anymore while a transaction is in progress? In this aspect, knex.js performs worse than using node-sqlite3 directly.

elhigu:
@irsl AFAIK If you add more connections to the pool than just one, there is no blocking in knex side. Please open an issue if you think that knex is blocking multiple connections. IIRC sqlite allows more connections that just one nowadays (except for inmemory DB).

irsl:
@elhigu I just verified, 0.14.6 still holds back the read queries while a transaction is in progress. Note: using the Sqlite library directly this just works fine.

I managed to hook Knex.js 0.12.6 back then to address this, but it was so ugly at the end I decided not to contribute it. My patches obviosly broke with the recent changes in Knex.
Anyways, I just created a small mini-project with my unit tests to see what my expectations are:

https://github.com/irsl/knex-sqlite-transactions/commit/ec22313f84c103875524f10e1cfac2a73e549e86

irsl:
Sorry, I was inattentive regarding your remark on the pools. By increasing the defaults and tweaking a bit in one of my tests everything works as expected. Great!

irsl:
@elhigu I just upgraded to the latest Knex, and started encountering SQLITE_BUSY errors without my custom transaction hooks (when the pool number is higher than 1). I also learned the node-sqlite3 library lacks busy timeout support (https://github.com/mapbox/node-sqlite3/issues/9) and as I see it won't change as it would block the complete applciation due to the way how Node works. 

My question is:

To workaround this, do you think a transaction serialization feature could be added in Knex? My original patch for 0.12.6 looked something like this:

```
        if(!settings.officialSinglePoolMode) {
            const bluePromise = knex.Promise;
            var transactionWaitList = [];
            var originalTransaction = knex.transaction;
            knex.transaction = function(callback){
                if(debug) console.log("!!! transaction was called");

                return new bluePromise(function(resolve,reject){

                    transactionWaitList.push({callback:callback, resolve: resolve, reject: reject});

                    // initial kickoff:
                    if(transactionWaitList.length == 1)
                       fireNextOne();

                });

                function fireNextOne(){
                      // are there any more transactions?...
                      if(transactionWaitList.length <= 0) return;

                      var stuff = transactionWaitList[0];
                      var oex;
                      return originalTransaction(stuff.callback)
                      .catch(ex=>{
                          // console.log("there was an exception!", ex.message)
                          oex = ex;
                      })
                      .then((returnValue)=>{
                         if(debug) console.log("!!! original transaction has finished");
                         transactionWaitList.shift();
                         fireNextOne();

                         if(oex) return stuff.reject(oex)
                         return stuff.resolve(returnValue);
                      })
                }

            }

        }
```

elhigu:
Transactions should be automatically serialized if you have set pool size to be 1, because when transaction is started it tries to acquire connection from pool and if there are no connections available, knex will wait until the previous transaction returns the connection back to the pool.

So I'm failing to see the problem here that you are trying to workaround...

irsl:
Pool size 1 limits the number of concurrent read operations (eg. select) as well. Sqlite is better than that.

elhigu:
@irsl So why are you getting SQLITE_BUSY  errors? You think that only serializing transactions that there are no multiple of them running simultaneously would prevent that once and for all?

I'm still failing to see the problem here that you are trying to workaround...

PR #2469, opened 2018-02-14
Better timestamps options with default values and more flexibility.

paulogdm:
This request only adds three small features:
 
* Rename `created_at`/`updated_at`. (Default to "created_at"/"updated_at".
 * Type of those fields. (Default: "timestamp")
 * Small change on "dropTimestamps".

PS: Please do notice that this is my first PR to a project this big. Be kind to point me in the right direction.

### DIFF
--- a/src/schema/tablebuilder.js
+++ b/src/schema/tablebuilder.js
@@ -167,11 +167,10 @@ each(columnTypes, function(type) {
 });
 
 // The "timestamps" call is really just sets the `created_at` and `updated_at` columns.
-TableBuilder.prototype.timestamps = function timestamps() {
-  const method = (arguments[0] === true) ? 'timestamp' : 'datetime';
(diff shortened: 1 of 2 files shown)


paulogdm:
Tbh I wanted to go ahead and do `function({datatype = 'timestamp', createdatcolname = 'created_at', updatedatcolname = 'updated_at'}){...}` but I think it would break a few things on old node versions...

elhigu:
There is linting error in the code, so tests are failing. Also I don't like at all long parameter list for adding different options. If this is being changed it would require being backwards compatible and new parameters should be passed as options object to method. Try to find old PRs/issues about timestamps columns (sorry I don't have time right now to start searching them). There are lot of discussion about the theme.

I don't think this is correct way to change the functionality.

paulogdm:
@elhigu Ty for the clarification!!!

PR #2473, opened 2018-02-16
Remove usage of create table if not exist from migration api

elhigu:


src/migrate/index.js
```diff
@@ -154,7 +154,7 @@ export default class Migrator {
 
   // Create the migration table, if it doesn't already exist.
```
joelmukuthu: the problem with comments in code :)

elhigu: True 👍 should be few lines above.

elhigu: fixed

PR #2476, opened 2018-02-16
Added knexTxId to query events and debug calls

### LINKED ISSUE #2379: Determine which transaction the query belongs to when using DEBUG
# Feature request

I would like to see to which transaction a query belongs when using `DEBUG`.

```
  knex:tx trx2: Starting top level transaction +4s
  knex:pool INFO pool mysql:mysql:client0 - dispense() clients=1 available=1 +1ms
  knex:query BEGIN; +2ms
  knex:bindings undefined +1ms
  knex:query insert into ... +2ms
  knex:query insert into ... +2ms
  knex:bindings [] +0ms
  knex:query COMMIT; +2ms
  knex:bindings undefined +0ms
  knex:tx trx2: releasing connection +2ms
  knex:client releasing connection to pool: __knexUid1 +0ms
  knex:pool INFO pool mysql:mysql:client0 - dispense() clients=0 available=1 +0ms
```

As you can see `trx2` is the transaction id but it isn't clear to which transaction the query belongs to.

wubzz:
Fixes #2379


wubzz:
@elhigu Any thoughts on this one?

Would have preferred not to bloat the connection object even more, but feel its inevitable for what was requested, and I completely understand why this property is needed.

src/dialects/oracle/transaction.js
```diff
@@ -40,19 +40,24 @@ export default class Oracle_Transaction extends Transaction {
...
-      connection.setAutoCommit(true)
-      if (!config.connection) {
-        t.client.releaseConnection(connection)
-      } else {
-        debugTx('%s: not releasing external connection', t.txid)
-      }
-    })
+    )
+      .then(connection => {
+        connection.__knexTxId = this.txid;
```
wubzz: Not sure why this is not working, unable to test locally..

elhigu: I will setup docker, which runs oracle tests locally in container. node-oracledb driver is pretty painful to install on every platform 😬

wubzz: Yeah I tried installning on Windows, and I can vouch for the difficulty... 😥

elhigu: I don't see anything wrong either. I suppose this.txid is not set at this point. looks like in other places they are referring it with t.txid, but it shouldn't matter. Will be easier to figure out when tests can be ran locally.

elhigu: https://github.com/tgriesser/knex/pull/2491 there is PR for adding npm script to run oracledb tests locally in docker container

wubzz: Took more time than I care to admit to find the issue. Turns out I only changed in `dialects/oracle/transaction.js` assuming that was what test was running, but it was actually running `dialects/oracledb/transaction.js` which I had forgotten to change

Issue #2320, opened 2017-11-12
knex.initialize() calls client.initialize() that has been renamed to client.initializePool()


ignitenet-martynas:
Line 46 of src/util/make-knex.js:
```javascript
    // Typically never needed, initializes the pool for a knex client.
    initialize(config) {
      return client.initialize(config)
    },
```

However the corresponding method in Client has been renamed from initialize() to initializePool(). So the call should be changed to client.initializePool().

It might make sense to rename initialize() to initializePool() for Knex also as that's what it actually does? Maybe keep initialize() with a deprecation warning?

I could do a PR but would like to discuss the proper way to proceed first.

elhigu:
I think it can be just changed to call initializePool. However I am pretty amazed that test suite didn't catch this one. Is it not called at all by any of the tests?

pablopen:
Hi, just came across this bug while trying to reconnect the datapool when the database server is restarted in oracledb. Basically two lines of code:
in src/util/make-knex.js line 46:
```
    // Typically never needed, initializes the pool for a knex client.
    initialize(config) {
      return client.initializePool(config)
    },
```
And in 
src/client.js line 267:
```
  initializePool(config = this.config) {
````
This will allow to call
knex.destroy()
and then:
knex.initialize()
without parameters, and use the previous datapool config.

I'll try to do a PR with the fix.

pablopen:
Just made a PR. And  @elhigu it wasn't called in any test

Issue #106, opened 2013-11-20
using .having()


mistakia:
I'm trying to run something like:

```
knex('books')
    .join('tags_books', function() {
        this.on('books.id', '=', 'tags_books.book_id');
    })
    .join('tags', function() {
        this.on('tags_books.tag_id', '=', 'tags.id');
    })
    .whereIn('tags.name', tags)
    .havingRaw('count(distinct `tags`.`id`) = ' + tags.length)
    .select('books.*')
```

This throws a "Cannot call method 'push' of undefined" at knex/lib/builder.js:339:20

After pulling in the changes from #48, I was able to get it to run.

p.s. just switched from sequelize and I cant be happier. knex/bookshelf are a blast to use.

PR #2477, opened 2018-02-16
Fix knex.initialize() and adds test

pablopen:
Fix knex.initialize() call, needed to call  client.initializePool() instead of client.initialize()
Adds default value for client.initialize(config), it uses its own config if no config provided.

Adds test for initialize in integration.

Refered in issue #2320


elhigu:
Thank you. Basically this pull request finishes new feature, documentation for this method should still be added. I had never heard about this method and I don't know what is the use case for it.

pablopen:
Hi, should I add something anywhere?
I came across this method digging for something to force a reconnection to the database, so looks it's been there a while, but as says the comment in the lines is rarely used.
My use case for this is some error in oracle when the database server is rebooted, and the pool doesn't seems to manage it correctly.(It doesnt try to reconnect).
So I  catch the error with knex.on('query-error',...) and then isolate the ORA's 03113 AND 03114 and try to reboot the pool in those cases with knex.destroy() and knex.initialize().
I may open an issue with more details when I get some time.

elhigu:
@pablopen do you have that problem still with latest 0.14.3 lots of instability issues has been fixed there. I will add also oracle to stress test suite to make sure it reconnects automatically correctly.

Documentation is in another repository. It is mentioned in CONTRIBUTING.md.

pablopen:
@elhigu Yes, I still have those with 0.14.3, the flow goes like this:
Connect knex to oracle,
Make a query
Disconnect oracle (server or docker)
Make another query (It fails as oracle is disconnected)
Connect oracle
Make a query (It fails even when oracle is back online)

I'll check the docs to add mention to this function this evening

elhigu:
Ok, good to know about those problems. I will add oracledb to my stress test setup and try to fix it.

pablopen:
Just created a PR #106 in the documents. This was just a bug-fix of something that used to work.

test/integration/suite.js
```diff
@@ -42,4 +42,11 @@ module.exports = function(knex) {
     });
   });
 
+  describe('knex.initialize', function() {
+    it('should allow initialize the pool with knex.initialize', function() {
+      knex.initialize();
```
elhigu: Isn't this missing pool destroy before trying to reinitialize? Also this is not actually waiting that initialization is done before testing that knex was initialized. 

Also check that before calling init and after calling destroy pool is actually destroyed.

pablopen: Hi, at that time of the test the pool is currently destroyed. I guess that you want the test to check first if the pool is initialized (maybe with !client.pool), and if not destroy it. The waiting until is destroyed is what i don't really know how to implement

elhigu: Looks like initialization is indeed synchronous opeartion, so that actually doesn't have to be waited. Yep, checking client.pool to not exist should be enough for testing that knex was not initialized at start of the test.

pablopen:
I added the to test one line that checks the connection was destroyed before initializing it

elhigu:
Now this is good, thanks!

Issue #2461, opened 2018-02-11
Drop support for mariasql and other dead dialects


elhigu:
That driver is pretty much dead (https://github.com/mscdex/node-mariasql) and during stress testing it with flaky connections it crashes without being able to fix it in knex side 

```
Mikaels-MacBook-Pro-2:stress-test mikaelle$ node reconnect-test-mysql-based-drivers.js | grep -A 3 -- "- STATS "
------------------------ STATS PER SECOND ------------------------
{ mysql: { queries: 0, results: 0, errors: 0 },
  mysql2: { queries: 0, results: 0, errors: 0 },
  maria: { queries: 302, results: 289, errors: 12.5 } }
Assertion failed: (!persistent().IsEmpty()), function Unref, file ../../nan/nan_object_wrap.h, line 106.
Mikaels-MacBook-Pro-2:stress-test mikaelle$ 
```

People should use mysql / mysql2 drivers with mariadb and having this option is misleading.

wubzz:
Honestly no clue how the oracle parts of knex works, but am I correct in assuming [strong-oracle](https://www.npmjs.com/package/strong-oracle) falls under this category?

`websql` ?

elhigu:
Yes both. I see if I can move some alternative dialects to separate packages.

tgriesser:
Yeah let's take out `strong-oracle` and `mariasql`, and make a note that if you want to use MariaDB you can use the `mysql` driver. I could go either way on websql, not sure if people are using it with cordova or anything.

Eventually I'd like to get a better standard for how to author & contribute dialects. Getting back into the codebase after awhile and realizing how much the overall lib structure could use a good cleanup - probably will punt on that as part of some bigger changes after 1.0 though.

wubzz:
Closing as both dialects have been removed from codebase.

PR #2480, opened 2018-02-17
Dropped support for mariasql

wubzz:
Relates to #2461.


elhigu:
I would like to still release one patch version for 0.14.x series during this week before doing next breaking version. Changes looks good.

wubzz:
@elhigu Created a milestone for 0.15 instead since this is breaking.

elhigu:
rebased in #2681

PR #2483, opened 2018-02-18
Changed rm to rimraf in 'npm run dev'

wubzz:
This is primarly for environments where `rm` is not available, such as Windows without Git Bash.

`rimraf` is already a devDependecy so there is no extra cost in that department.


wubzz:
Failing test is not related to this change and only applies to Node 8 (some oracledb test), merging.

PR #2484, opened 2018-02-18
Changed babel preset and use latest node as target when running dev

wubzz:
>🙌 Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update!

Replaced `babel-preset-es2015` with `babel-preset-env` (https://babeljs.io/env/)

I also changed so that running `npm run dev` targets latest node so that debugging code in `lib` during development is easier as not everything gets transpiled, such as classes etc.

### DIFF
--- a/.babelconfig.js
+++ b/.babelconfig.js
@@ -1,0 +1,15 @@
+'use strict';
+
+var isDev = process.env.npm_lifecycle_event === 'dev';
+var presets = [
+  ["env", Object.assign({"loose": true}, isDev ? {"targets": {"node": "current"}} : {})]
+];
+
+module.exports = {
+  "presets": presets,
+  "plugins": [
+    "lodash",
+    "transform-runtime",
+    "add-module-exports"
+  ]
+};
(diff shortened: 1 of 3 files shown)


wubzz:
@elhigu Are you fine with this `dev` change?

I find it very obnoxious to debug the transpiled code in development, for instance when writing tests.

elhigu:
I would love to drop babel all together after node 4 gets to end of line. 

Node 6 already should supports destructuring, ... -operator and default parameters which are the most used more advanced javascript features used.

wubzz:
@elhigu I'd have loved to drop babel as well, but unfortunately it might not be possible even when Node 4 is dropped, since many people still use knex in browser environments for query building purposes.

Unless we at that point compromise and leave the transpiling up to the end user, of cousre. That's an option.

elhigu:
AFAIK knex already needs dome trickery when one tries to use it as query builder on web page, like patching some missing requires etc. 

I would like to know how many people are actually using it on webpage. One way to figure it out would be to drop support and if there are many people angry about it we could add build for web which omits pool / connection stuff and would work out of box when included to webpage.

Need to think more about it after may :)

PR #2491, opened 2018-02-19
Added npm run test:oracledb command that runs oracledb tests in docker

elhigu:
Also cleaned up some old docker configurations from test knexfile

Helps debugging https://github.com/tgriesser/knex/pull/2476


package.json
```diff
@@ -73,12 +73,13 @@
...
     "dev": "rimraf ./lib && babel -w src --out-dir lib --copy-files",
     "lint": "eslint src/**",
-    "plaintest": "mocha --check-leaks -b -R spec test/index.js && npm run tape",
+    "plaintest": "mocha -t 10000 -b -R spec test/index.js && npm run tape",
     "prepublish": "npm run babel",
     "pre_test": "npm run lint",
     "tape": "node test/tape/index.js | tap-spec",
     "debug_tape": "node-debug test/tape/index.js",
     "test": "npm run pre_test && istanbul --config=test/.istanbul.yml cover node_modules/mocha/bin/_mocha -- --check-leaks -t 10000 -b -R spec test/index.js && npm run tape",
+    "test:oracledb": "docker rmi -f --no-prune knex-test-oracledb && docker build -f scripts/oracle-tests-Dockerfile --tag knex-test-oracledb . && docker run -i -t knex-test-oracledb",
```
elhigu: I think I will change this to be named:  `oracledb:test` I'm making PR tomorrow for running mssql tests too (they seem to be failing currently though)

wubzz: This is awesome. 👍

elhigu: I'll update task name in mssql PR

scripts/oracle-tests-Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+ENV KNEX_TEST /oracle-config.js
+ENV DB "sqlite3 oracledb"
+ENV KNEX_TIMEOUT 60000
+
+ADD . knex
+
+CMD /usr/sbin/startup.sh && \
+  cd knex && \
+  npm install && \
+  npm install oracledb && \
```
wubzz: I think `npm install sqlite3` is needed too since `ENV DB "sqlite3 oracledb"`.

Was able to run this eventually *(new to docker)* in Windows environment, but stumbled upon `sqlite3` module not installed when it started running tests.

Strangely enough some sqlite3 tests were failing too on master, investigating..

elhigu: Yeah it might be that sqlite is always required for tests. 

I actually added it because there was some strange segmentation faults happening after running tests. I was able to get rid of them by removing --leak-check flag from plaintest script. I'll try to remove sqlite testing from there or add it to  npm install.

PR #2401, opened 2017-12-30
Implement partial index support

brandonramirez:
Partial/filtered indexes for Postgres, SQLite and Microsoft SQL Server.

Some of the SQL Server integration tests fail, even without this change.  They appear to be existing issues.

I attempted to implement this feature for Oracle too, but getting connectivity to Oracle from my development environment is a nightmare.  Even in the CI environment, Oracle client installation fails intermittently, so I decided to keep my hands out of Oracle.

MySQL and MariaDB do not support this feature, so I did not attempt to implement it there.

There are still outstanding test failures (I think all due to the above issues), so the Travis CI builds keep failing.  I have verified that this feature does not introduce any new test failures.

Fixes #657

PR #2495, opened 2018-02-19
Adding Prettier

tgriesser:
I know this doesn't do much for the major architectural things that need to be changed in the library, but hopefully it'll help with developer experience as we start to tackle those.

@elhigu and @wubzz since you have a better idea of what the outstanding PR's are, let me know when this would be best to merge, so there's less merge conflicts.


package.json
```diff
@@ -66,22 +70,32 @@
     "babel-preset-es2015-loose",
     "rimraf"
   ],
+  "lint-staged": {
+    "*.{js,json}": ["prettier --write", "git add"]
```
wubzz: Nice! 🥇

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5"
```
wubzz: Not sure on the general convention of things but personally I would always opt for `arrowParens: "always"` as well, but of course it's personal preference. Just a thought!

wubzz:
This will help a lot in terms of development! Consistent codebase is key.

As I wrote in gitter I think this can be merged at any time since existing PRs can be fixed using `eslint --fix` *(covers most cases)*.

If we really want to wait then perhaps 0.15 is a good target.

elhigu:
If I understood correctly what @wubzz was suggesting in gitter chat, we could apply prettier straight away and then run prettier to those old PRs too... if that doesn't work, merging code by hand to the master is not so big deal for me either. I can do that manual work for every branch that I'm interested in 👍 

Anyways these PRs are currently on my list:

https://github.com/tgriesser/knex/pull/1390
https://github.com/tgriesser/knex/pull/2401
https://github.com/tgriesser/knex/pull/2449
https://github.com/tgriesser/knex/pull/2476

elhigu:
> If we really want to wait then perhaps 0.15 is a good target.

I believe next version could be 0.15.0. 0.14 series is starting to be stable already (well I still need to add stress testing for mssql and oracle though).

PR #2508, opened 2018-02-23
Added error if chaining update with limit()

capaj:


tgriesser:
Update with limit is valid syntax in MySQL https://dev.mysql.com/doc/refman/5.7/en/update.html

wubzz:
@tgriesser Perhaps #2506 was premature in that case, depending on your view of `.first()` vs `.limit()`. Should I revert that PR?

capaj:
@tgriesser could we throw the error only when the client is `!== Mysql`? That could resolve the situation with #2506 as well. Or do you want to stay away from putting exceptions for certain clients inside query builder?

tgriesser:
Yeah I was thinking of staying away from putting exceptions for certain clients, though I guess since you really shouldn't be using it if you're not able to, I'd be okay with at least warning... but since it's valid in mysql, mssql as well as sqlite3 if compiled with the [`SQLITE_ENABLE_UPDATE_DELETE_LIMIT`](https://sqlite.org/compile.html#enable_update_delete_limit) flag, so I'm not sure what the best option here is, other than to trust users know what they're during when using the limit feature.

@wubzz I'm fine with the change in `.first` because that's explicitly meant to be used in a select, and not with returning, which is accomplished just as easily with `.then(result => result[0])`

Issue #2507, opened 2018-02-22
Expose knex.ident()/knex.ref() function


wubzz:
# Feature discussion / request

1. Explain what is your use case

Generating an identifier such as `table.column` -> `"table"."column"` currently requires `knex.raw('??', 'table.column')` which gets pretty tedious to keep repeating around the codebase.

There have been several issues asking:
- How to do this?
- Is it possible via public API?
- Would a PR adding a way to expose this type of function be accepted?

We have yet to do something about this, so I figured I would revive the discussion with a new feature proposal.

2. Explain what kind of feature would support this

Exposing a new function on the `make-knex.js` object *(returned by calling `knex({})`)* that simply proxies its arguments to `client.formatter(this.raw()).columnize(...arguments)` would allow for syntax such as `knex.ident('table.column')`

3. Give some API proposal, how the feature should work

```javascript
//Old
knex('Table1')
.where('Table1.UserId', knex.raw('??', 'Table2.UserId'))
.select()

//New
knex('Table1')
.where('Table1.UserId', knex.ident('Table2.UserId'))
.select()
```

*(Open for better function name than 'ident'/'identifier')*

I realize that this does not result in much less code, but that is not the goal. It's simply a new public-API-exposed function that serves a specific purpose and can be documented as such.

elhigu:
In objection we have `ref()` for column / json references, `lit()` for these cases and in addition those helpers also supports casting references columns / literals to different types, which is necessary in many cases.

I think this would be good feature to have also in knex.

elhigu:
this ident builder could maybe even have knex.indent('Table.Column').withSchema('Schema') supported.

wubzz:
Won't `.withSchema` add some overhead? Since `Schema.Table.Column` as input should yield the same result, I think. Or maybe add it anyway for consistent syntax with `queryBuilder`?

Casting would also be great, it's another requested feature. Though, we'd also need to expose `knex.castX` somehow since one might want to cast values and not just references. How does Objection handle this? For example `'2018-02-23'::date`

elhigu:
@wubzz I don't see how it would add any significant overhead. Problem with `part1.part2.part3` syntax is currently ambiguity, that one has to interpret it differently every places in knex API.  for exmample `part1.part2` may mean `table.column` or `schema.table` depending where you use it and when used as `part1` it may mean `table`, `column` or `schema`. 

[...]

PR #2509, opened 2018-02-23
Added 'ref' function

wubzz:
This a dead simple implementation of #2507 which could definitely be improved upon. *(Which I'm more than open for)*

Can write more tests and write docs once design and featureset has been finalized.


wubzz:
Maybe support for aliases..?

wubzz:
Added docs https://github.com/knex/documentation/pull/112

elhigu:
thanks!

Issue #2518, opened 2018-02-27
Performance of knex.select().from() is very slow compared to knex.raw()


jeremija:
# Environment

Knex version: 0.14.4
Database + version: SQLite 3.1.13
OS: Arch Linux, Mac High Sierra OS 10.13.3

# Bug

Using `knex.select().from()` causes a big performance degradation when compared to `knex.raw()`. For the same query, the performance can be up to 4x slower:

```
$ node test.js
preparing database...
running benchmarks...
select `id`, `name` from `test` [knex.raw()   : 9999 rows in 49ms]
select `id`, `name` from `test` [knex.select(): 9999 rows in 191ms]
```

I'd expect `knex.select().from()` to only be slightly slower than the raw query. It seems to be worse when querying large tables (`N=100000`):

```
$ node test.js
preparing database...
running benchmarks...
select `id`, `name` from `test` [knex.raw()   : 99999 rows in 493ms]
select `id`, `name` from `test` [knex.select(): 99999 rows in 808ms]
```

I first noticed this with Oracle, but I quickly realized it happens with SQLite3 as well. A full test case is available here: https://github.com/jeremija/knex-bug-report:

```
git clone https://github.com/jeremija/knex-bug-report
npm install
N=100000 node test.js
```

Might be related to #1795.

elhigu:
Here is online version of your benchmark https://runkit.com/embed/8szo1dhyi9wc 

I wouldn't consider this as a bug nor big problem since it is natural that query building and result parsing must take longer time when not using knex.raw ,which basically just sends query to the driver and returns unprocessed result to the caller... however I am a bit interested to see from profiling data what is it actually that is making the biggest difference. This is not related to #1795.

> It seems to be worse when querying large tables (N=100000):

Actually your results show (runkit version) that select version is only ~1.5x compared to raw with large tables and repeated runs and with smaller table difference is bigger ~4x.

elhigu:
Closing since I don't see this being bug. Any extra information about profiling etc. is welcome though.

jeremija:
>> It seems to be worse when querying large tables (N=100000):
>
> Actually your results show (runkit version) that select version is only ~1.5x compared to raw with large tables and repeated runs and with smaller table difference is bigger ~4x.

You're right about 1.5x, but I don't think that's the best way to look at it since the query time difference is not constant. For example (unfortunately I could not run this in runkit since it timed out):

```
$ N=1000000 node test.js
preparing database...
running benchmarks...
[...]

PR #2520, opened 2018-02-28
Testing removal of 'skim'

wubzz:
Temporary, might be closed.

### DIFF
--- a/src/dialects/maria/index.js
+++ b/src/dialects/maria/index.js
@@ -118,9 +118,8 @@ assign(Client_MariaSQL.prototype, {
       case 'select':
       case 'pluck':
       case 'first': {
-        const resp = helpers.skim(rows);
-        if (method === 'pluck') return map(resp, obj.pluck);
-        return method === 'first' ? resp[0] : resp;
+        if (method === 'pluck') return map(rows, obj.pluck);
+        return method === 'first' ? rows[0] : rows;
       }
       case 'insert':
         return [data.insertId];
--- a/src/dialects/maria/index.js
+++ b/src/dialects/maria/index.js
@@ -4,7 +4,6 @@
 import inherits from 'inherits';
 import Client_MySQL from '../mysql';
 import Promise from 'bluebird';
-import * as helpers from '../../helpers';
 import Transaction from './transaction';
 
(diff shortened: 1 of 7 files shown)


wubzz:
@tgriesser This function is 4 years old and hasn't been changed. This is a long shot in the dark, but do you happen to remember the reasoning behind it? 😄 

No tests are failing after removing it. Still not sure if that's reliable enough, tho.

Currently this is a progressive performance bottleneck as described in https://github.com/tgriesser/knex/issues/2518#issuecomment-369145795

elhigu:
Maybe it converts random classes etc. returned from drivers to plain objects?

elhigu:
Anyways I'm pretty if it is wanted then it would be faster to do with plain javascript loop. @koskimas has done lots of micro optimizations to obejction, he probably remembers from top of the head if there is something really inefficient in that implementation.

tgriesser:
> Maybe it converts random classes etc. returned from drivers to plain objects?

Exactly it. I forget which drivers it was now though

wubzz:
That makes sense.

Pressumably the only reason it affects so many dialects is because copy-paste when adding new dialects. I imagine mssql can be one such case.

elhigu:
Looks like this could be used also to copy object own properties to plain object https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

If I didn't do something badly wrong with object assign skimming is ~20x faster https://jsperf.com/knex-skim (on chrome at least)

koskimas:
@elhigu I was about to suggest the same. V8 team has spent some time optimizing Object.assign recently. It will probably get even faster with node 10.

koskimas:
I think you could even skip the whole thing if `obj.constructor === Object`

wubzz:
@tgriesser @elhigu I'm comfortable removing this for both MySQL (from Tim's comment) and Postgres. I can't speak for other dialects, tho.

Anything we can do to speed up resultset is a win-win situation imo.

If you're ok with this I can update the PR.

elhigu:
@wubzz I believe this is good to go, just need to remember write to changelog about that returned row types might not be plain js objects anymore on those dialects, which were skimmed earlier.

PR #2523, opened 2018-03-03
native promises

danielgatis:


kibertoad:
@danielgatis What is the rationale behind this change?

danielgatis:
It was a experiment and I want to run the tests to collect some information. I will close this pull asap.

elhigu:
@kibertoad to get rid of unnecessary dependency mostly. Bluebird is not anymore that indispensable that it was couple years ago. Also after async / await world it is all the time less useful, since for example even if you return bluebird promise from `async` function, it will be wrapped to native promise because of `async` declaration.

kibertoad:
@elhigu Isn't bluebird still much more performant than native promises? (less so on newer Node.js version, but still).

elhigu:
@kibertoad performance difference is nowadays pretty irrelevant. I would be surprised if native implementation would not be faster (if it isn't yet) at some point. Anyways as said it is irrelevant. Both are fast enough.

danielgatis:
The main goal behind this is to make the use of transactions easier with the use of async hooks.

Issue #1528, opened 2016-06-22
knex doesn't parse connection URL query params into connection object


shane-walker:
I've run into an issue with knex not parsing the query params we're appending to our MySQL connection URL, at all.

For example, the following string should be converted to an object that respects the query params it is passed.

**Config URL**
`mysql://root:@localhost/myDatabase?charset=utf8&timezone=UTC`

**Expected Object**

``` javascript
{
  database: 'myDatabase',
  host: 'localhost',
  user: 'root',
  charset: 'utf8',
  timezone: 'UTC'
}
```

**Actual Object**

``` javascript
{
  database: 'myDatabase',
  host: 'localhost',
  user: 'root'
}
```

Tracing the issue, it seems that the function `connectionObject` in the [parse-connection](https://github.com/tgriesser/knex/blob/master/src/util/parse-connection.js#L28) util script isn't set up to handle query params currently.

For reference, this is how the node-mysql driver currently handles query params.
https://github.com/mysqljs/mysql/blob/master/lib/ConnectionConfig.js#L171

sandelius:
And the client is part of the url so we shouldn't need to type both client: 'pg' and 'postgres://'

glenjamin:
I just ran into this with `mssql` as well, where the `?encrypt=true` portion was being ignored.

It'd be good to pass query params down to the dialect driver, there wasn't a mention in the docs that knex would do it's own URL parsing, so it took me a while to track this down.

Perhaps an easy quick improvement would be to error if querystring params are found in the URL, as they're being discarded at the moment.

elhigu:
We should revisit the code of every dialect, how connection string/options are passed to driver.

enapupe:
Maybe this project can help https://github.com/vitaly-t/connection-string

mercmobily:
Thanks for reporting this issue about connection URL query parameters not being parsed properly. This was indeed a limitation in the original implementation where the `connectionObject` function in `parse-connection.js` wasn't handling query parameters at all.

Good news is that this has been fully resolved in the current version of Knex! The connection string parser now properly extracts and applies all query parameters from the URL. Looking at the current implementation in `knex/lib/knex-builder/internal/parse-connection.js:70-84`, you can see that the parser now iterates through `searchParams` and adds each parameter to the connection object:

```javascript
if (parsed.searchParams) {
  for (const [key, value] of parsed.searchParams.entries()) {
[...]

PR #2538, opened 2018-03-23
MSSQL: Support query string options.

khell:
This is a PR to add support to the `mssql` dialect to parse query string parameters when a connection string is provided to knex. It's needed for Azure SQL, where encryption is turned on and the mssql library knex uses requires `encrypt: true`.

The implementation just uses the existing `url.parse` function's [second parameter](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost) to parse query strings into an object. Then it assigns a clone of this object to the options property accepted by the library.

Example connection string: `mssql://user:password@example.com/ExampleDB?encrypt=true`

I also added the Babel plugin to implement a transform for object rest spread. It is supported in the latest Node 8 LTS so I think it should be supported here.

There are some issues with this PR I'd like opinion on:
1. Boolean arguments. The mssql library checks for `encrypt: true` through type coercion. So if `encrypt=false` is passed, then it's still coerced to a boolean `true`. Without parsing the boolean string `'false'` or changing the `mssql` library this is potentially confusing for some users.
2. Not sure if any additional tests are needed. I think this is already covered fairly well by the existing integration tests, which support dynamic configuration of the connection parameters/string.
3. I'm aware there has been discussion around this in the past and #1528 suggests a bigger overhaul may be necessary. If this is not suitable please let me know.

Thanks!

EDIT: I do not know why the Travis CI builds have failed. I cannot view the logs.

### DIFF
--- a/.babelconfig.js
+++ b/.babelconfig.js
@@ -10,6 +10,7 @@ module.exports = {
   "plugins": [
     "lodash",
     "transform-runtime",
-    "add-module-exports"
+    "add-module-exports",
+    ["transform-object-rest-spread", { "useBuiltIns": true }]
   ]
 };
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "babel-eslint": "^8.2.1",
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-lodash": "3.3.2",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-env": "^1.6.1",
     "chai": "^4.1.2",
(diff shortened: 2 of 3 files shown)


elhigu:
Travis did fail because:

```
  Failed Tests: There was 1 failure
    parses mssql connections, aliasing host to server
      ✖ should be equivalent
  total:     257
  passing:   256
  failing:   1
  duration:  3.3s
```

I'm not sure about this PR... it does fix parsing a bit, but it still doesn't support any stranger mssql connection string formats and I'm not sure if this is correct implementation (I know some other dialects also parses query parameters and passes them to driver...). Some more generic way for this standard type of query string would be good and then additional parsers for weirder stuff like mssql other connection string types.

Also test would be needed to check that new code path is executed correctly.

I don't have time in next couple of weeks to investigate further is this correct way to so hopefully someone else will.

mattyweb:
Looks like this commit has a birthday coming up! Any progress on this one? 

Am working on deploying a dockerized node app in Azure at the moment and would love to be able to use this along with the Connection String setting in Azure.

kibertoad:
@mattyweb If you are up for resolving conflicts, writing a test and testing it to work properly, this PR definitely could be accepted :)

kibertoad:
@khell Are you still interested in finishing this PR?

kibertoad:
This seems to be redundant with https://github.com/knex/knex/pull/3547

PR #2542, opened 2018-03-24
 Create timestamp columns with millisecond precision on MySQL 5.6 and newer

### LINKED ISSUE #2524: Timestamp is rounded on MySQL after save
# Environment

Knex version: 0.14.4
Database + version: Mysql 5.7.21
OS: Ubuntu 16.04 x86_64

# Bug

When using the `.timestamps()` schema builder method on a table, the generated datetime columns will round dates on MySQL when inserting data.

For example, inserting `2018-03-04T19:24:53.063Z` will result in `2018-03-04T19:24:53.000Z` when retrieving the same row afterwards.

Obviously this presents some problems when one wants to compare dates.

[This gist](https://gist.github.com/ricardograca/bc26ee92baeeee3f792a5d90050608cc) has all the code necessary to reproduce the issue. Just make sure the database config matches what you have in your system.

I can look into it and provide a PR.

ricardograca:
This changes the `timestamp()` and `timestamps()` schema building methods to output columns that have millisecond precision on MySQL versions 5.6 and newer.

This feature is not enabled by default for now and requires passing a `version: '5.6'` option when initializing the client to enable. The version number should be your actual MySQL server version, although as long as it is 5.6 or greater the feature will be enabled. The version number can be in the form `5.6` or `5.6.10`.

Fixes #2524.

Associated documentation update: https://github.com/knex/documentation/pull/99

### DIFF
--- a/src/dialects/mysql/schema/columncompiler.js
+++ b/src/dialects/mysql/schema/columncompiler.js
@@ -73,11 +73,11 @@ assign(ColumnCompiler_MySQL.prototype, {
   },
 
   datetime() {
-    return supportsPreciseTimestamps(this.client) ? 'datetime(3)' : 'datetime'
+    return supportsPreciseTimestamps(this.client) ? 'datetime(6)' : 'datetime'
   },
 
   timestamp() {
-    return supportsPreciseTimestamps(this.client) ? 'timestamp(3)' : 'timestamp'
+    return supportsPreciseTimestamps(this.client) ? 'timestamp(6)' : 'timestamp'
   },
 
   bit(length) {
--- a/src/dialects/mysql/schema/columncompiler.js
+++ b/src/dialects/mysql/schema/columncompiler.js
@@ -9,9 +9,10 @@ import { assign } from 'lodash'
 
 function supportsPreciseTimestamps(client) {
   if (!client.version) {
-    const message = 'If you are using a MySQL version older than 5.6.4 you should specify the ' +
-                    'version in your client options when initializing Knex. See ' +
-                    'http://knexjs.org/#Schema-timestamps for more information.'
+    const message = 'To get rid of this warning you should specify the mysql dialect version in ' +
+                    'your knex configuration. Currently this defaults to 5.5, but in a future ' +
+                    'release it will default to 5.6 which supports high precision timestamps. ' +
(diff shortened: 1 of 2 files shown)


ricardograca:
@elhigu This is still missing the deprecation warning. Any tips on where I should place it? It's also missing documentation but I guess that goes in the documentation repo.

ricardograca:
No idea why the tests are failing on Node 6 and 8. On Node 6 it seems there's a segmentation fault right after the mocha tests. On 8 an assertion related to migrations is failing. I don't think this is caused by the changes on this PR.

elhigu:
8 failed to some random fail of oracle tests that fails every now and then. That segmentation fault was strange... I just reran those failing builds.

Yes, docs go to other repo and deprecation message inside that supportsPreciseTimestamps function if client.version is not given. Warning should explain that on older mysql versions they need to pass version: '5.5' or something like that or in future that code will break and with newer ones they need to choose 5.5 or older to keep timestamp precision bad. Maybe linking to documentation would be easieast to prevent having that huge warning message.

ricardograca:
@elhigu Updated. I also increased the precision to microsecond since this should help reduce rounding errors and I see no reason not to use the highest available precision.

src/dialects/mysql/schema/columncompiler.js
```diff
@@ -7,6 +7,17 @@ import * as helpers from '../../../helpers';
 
 import { assign } from 'lodash'
 
+function supportsPreciseTimestamps(client) {
+  if (!client.version) {
```
elhigu: I just realised that actually every mysql user needs to set that variable to get rid of the warning 😬 So I think the way to go there is to change a error message a bit to be like:

'To get rid of this warning one should specify mysql dialect flavor in the knex configuration. Currently flavor defaults to 5.5, but in future releases it will default to 5.6 that supports high precision timestamps. See http://knexjs.org/#Schema-timestamps for more information."

And then just remove the warning in some future release and change the default.

ricardograca: Ok, will do.

ricardograca: Updated.

elhigu:
Thanks a lot!

PR #2553, opened 2018-04-01
add clearOrder method fix #2360

### LINKED ISSUE #2360: clear orderBy
To paginate, I'm using query.clone().count('* as count'), but because the query has an ORDER BY, I'm getting an error: "must appear in the GROUP BY clause or be used in an aggregate function".

Is there a way to clear the query orderBy?

Thanks

revolunet:
this adds the `clearOrder` method

is it normal that atm the order clause is reused like in this example ? 

i ran into this for some obscure reason in a sqlite environnement. the order clause was getting a field from another database... 

```js
 const textData = await knex
    .select()
    .from("myTable")
    .where({
      etat: "TOP"
    })
    .orderBy("date", "desc")
    .first()
    .catch(console.log);
```

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -814,6 +814,12 @@ assign(Builder.prototype, {
     return this;
   },
 
+  // Remove everything from select clause
+  clearOrder(){
+    this._clearGrouping('order');
+    return this;
+  },
+
   // Insert & Update
   // ------
 
--- a/src/query/methods.js
+++ b/src/query/methods.js
@@ -77,6 +77,7 @@ export default [
   'pluck',
   'clearSelect',
   'clearWhere',
+  'clearOrder',
   'insert',
   'update',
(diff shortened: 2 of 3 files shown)


revolunet:
doc PR : https://github.com/knex/documentation/pull/100

elhigu:
Beautiful, no comments, thank you!

revolunet:
thanks ;)

do you know a way to fetch a built version from npm until you make a new release ?

elhigu:
I'm just starting to do release

revolunet:
can you please explain why the clear methods are required ?
id like to understand when to use it or not.
thanks

bertho-zero:
In my case I would also need a `clearGroupBy` to make a correct count.

My current result is: `countQuery === [{total: 3}, {total: 2}]`

Issue #2550, opened 2018-03-29
returning is not using wrapIdentifier


kamote:
# Environment

Knex version: `^0.14.4`
Database + version: `oracledb` - npm module version `^2.1.2`
OS: macOS 10.3

# Bug

1. Custom implementation on of `wrapIdentifier` is not being called when using `returning` clause. My custom `wrapIdentifier` just simple value to upper case.

```
knex:query insert into "PERSON" ("NAME", "CREATEDAT", "ID", "SETTINGS", "UPDATEDAT") 
values (:1, :2, :3, :4, :5) returning "id","name" into :6,:7 +0ms
  knex:bindings [ 'JUAN',
  knex:bindings   1522336643820,
  knex:bindings   '99999',
  knex:bindings   '{}',
  knex:bindings   1522336643820,
  knex:bindings   { type: 2001, dir: 3003 },
  knex:bindings   { type: 2001, dir: 3003 } ]
```

```
const knex = require('knex')({
  client: conf.adapter,
  connection: {
    host: conf.host,
    user: conf.user,
    password: conf.password,
    database: conf.database,
  },
  pool: conf.pool,
  wrapIdentifier: (value, origImpl, queryContext) => origImpl(value.toUpperCase())
});
```

@atiertant

PR #2554, opened 2018-04-02
GH-2550 Add support to wrap returning column

kamote:
Returning clause for oracledb client does not support custom wrapper, this PR add a support for that by calling `this.formatter.wrap`

- utilize `this.formatter.wrap` for every columnName in returningClause

### DIFF
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -127,6 +127,107 @@ describe("Custom identifier wrapping", function() {
     }, clientsWithCustomIdentifierWrapper);
   });
 
+  it("should use custom wrapper on multiple inserts with returning", function() {
+    // returning only supported directly by postgres and with workaround with oracle
+    // other databases implicitly return the inserted id
+    testsql(qb().from('users').insert([{email: 'foo', name: 'taylor'}, {email: 'bar', name: 'dayle'}], 'id'), {
+      mysql: {
+        sql: 'insert into `users_wrapper_was_here` (`email_wrapper_was_here`, `name_wrapper_was_here`) values (?, ?), (?, ?)',
+        bindings: ['foo', 'taylor', 'bar', 'dayle']
+      },
+      sqlite3: {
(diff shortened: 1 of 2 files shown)


kamote:
@joelmukuthu good catch thanks! I've updated the unit test at 904080d

test/unit/query/builder.js
```diff
@@ -127,6 +127,107 @@ describe("Custom identifier wrapping", function() {
...
+        sql: 'insert into [users_wrapper_was_here] ([email_wrapper_was_here], [name_wrapper_was_here]) output inserted.[id_wrapper_was_here] values (?, ?), (?, ?)',
+        bindings: ['foo', 'taylor', 'bar', 'dayle']
+      },
+      oracledb: {
+        sql: "begin execute immediate 'insert into \"users_wrapper_was_here\" (\"email_wrapper_was_here\", \"name_wrapper_was_here\") values (:1, :2) returning \"id_wrapper_was_here\" into :3' using ?, ?, out ?; execute immediate 'insert into \"users_wrapper_was_here\" (\"email_wrapper_was_here\", \"name_wrapper_was_here\") values (:1, :2) returning \"id_wrapper_was_here\" into :3' using ?, ?, out ?;end;",
+        bindings: function(bindings) {
+          expect(bindings.length).to.equal(6);
+          expect(bindings[0]).to.equal('foo');
+          expect(bindings[1]).to.equal('taylor');
+          expect(bindings[2].toString()).to.equal('[object ReturningHelper:id]');
```
elhigu: there must be better way to check these than converting binding to string?

kamote: @elhigu Sure I'll find a way for better checks, btw I just copied the whole pattern here 
https://github.com/kamote/knex/blob/904080d437b5426686a192c4d3f35d7982940795/test/unit/query/builder.js#L3544

elhigu: ah right... well If that becomes a problem some day it can be fixed then.

PR #2559, opened 2018-04-04
Support DB schema for migrations - fixes #2499

### LINKED ISSUE #2499: Knex migration attempts to create already existing changelog and fails entire migration
# Environment

Knex  0.14.4
PostgreSQL 9.6.6

# Bug

Since the upgrade to 0.14.4 we've started getting this error in CI:
```
 Knex:warning - migration failed with error: create table "lms"."lms_migrations" ("id" serial primary key, "name" varchar(255), "batch" integer, "migration_time" timestamptz) - relation "lms_migrations" already exists
02:18:00 Error migrating:  { error: relation "lms_migrations" already exists
```

It looks like 7ff766f7a380dbea97b99e4d3bd2077a7b9f8f35 actually wasn't as innocent as expected.
Might be related to the fact that we are using custom migration changelog table via 
> 	tableName: `${SCHEMA}.lms_migrations`
parameter.

kibertoad:


src/migrate/index.js
```diff
@@ -23,6 +23,7 @@ const CONFIG_DEFAULT = Object.freeze({
     '.co', '.coffee', '.eg', '.iced', '.js', '.litcoffee', '.ls', '.ts'
   ],
   tableName: 'knex_migrations',
+  schemaName: null,
```
kibertoad: I've chosen this name to be consistent with currently used "tableName"

.eslintrc.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
     "import/parser": "babel-eslint"
   },
   "env": {
-    "node": true
+    "node": true,
+    "mocha": true
```
kibertoad: Linting is complaining about afterEach usage otherwise.

Issue #360, opened 2014-07-09
List tables


jardakotesovec:
I did not find function that would allow me to get list of tables in database. And it is not obvious to me from doc if its possible to list all columns with 'columnInfo' function.

whitelynx:
For PostgreSQL, we should probably be explicit about the schema, since it's possible to configure it such that `pg_catalog` is not in the search path:
```sql
SELECT tablename FROM pg_catalog.pg_tables WHERE schemaname='public'
```

whitelynx:
Also, this page has a good summary of the ways of querying available databases, schemas, tables, etc. in various databases (though it's missing sqlite): http://troels.arvin.dk/db/rdbms/#cli-list_of_databases

samgranger:
+1 on this

rex:
Big +1 from me! :+1:

zachelrath:
+1

samhagan:
+1

NikolaBorislavovHristov:
+1

tgriesser:
Ok, let's add this as `knex.listTables()`, adding to the 0.14 project

RWOverdijk:
:+1: This is very useful. :)

zh99998:
it's 0.13.0 now but i still can't find it in document

elhigu:
@zh99998 nobody has implemented it yet.

verekia:
For SQLite, you can run this query:
```js
knex('sqlite_master').where('type', 'table')
```
To get an array of just the names inside an `async` function:
```js
(await knex('sqlite_master').where('type', 'table')).map(x => x.name)
```

There might be equivalent metadata tables in other database types, but at least this works well for SQLite and testing things in memory.

GeoffreyPlitt:
Still can't figure out how to do this for Postgres, can we get a working example? knex.raw() doesn't seem to execute SQL, it has to be passed to something else?

elhigu:
For example all these should execute knex.raw query just fine: `knex.raw(...).then(result => ...)` or `result = await knex.raw(...)` or `Promise.resolve(knex.raw(...)).then(result => ...)` or `otherPromise.then(() => knex.raw(...)).then(result => ...)`.

@GeoffreyPlitt do you have some specific case that does not work?

GeoffreyPlitt:
You're right. I was confused by the results being an instance of class Result and couldn't figure out if the query had run or not. I find this returned simpler data:

```
knex('pg_catalog.pg_tables')
[...]

PR #2560, opened 2018-04-08
Implement listTables (#360)

codeclown:
Implements a new method `listTables` which queries and returns a list of table names in the current database. See https://github.com/tgriesser/knex/issues/360.

---

I'd like to keep the test explicit (`expect(resp).to.deep.equal([...])` is better than `expect(resp).to.include.items([...])`). Not sure how to handle that nicely here, because there's a lot of tables created and every time a new one is added, it needs to be added to this `listTables` test.

Thoughts?

### DIFF
--- a/src/dialects/postgres/query/compiler.js
+++ b/src/dialects/postgres/query/compiler.js
@@ -5,7 +5,7 @@ import inherits from 'inherits';
 
 import QueryCompiler from '../../../query/compiler';
 
-import { assign, reduce, identity } from 'lodash'
+import { assign, reduce, identity, map } from 'lodash'
 
 function QueryCompiler_PG(client, builder) {
   QueryCompiler.call(this, client, builder);
@@ -73,6 +73,20 @@ assign(QueryCompiler_PG.prototype, {
 
   forShare() {
     return 'for share';
+  },
+
+  // Compiles a `listTables` query
+  listTables() {
+    return {
+      sql: 'select table_name from information_schema.tables' +
+           'where table_schema = \'public\' and table_catalog = ?',
+      bindings: [this.client.database()],
+      output(resp) {
+        return map(resp.rows, function (table) {
+          return table.table_name
+        })
+      }
+    }
   },
 
   // Compiles a columnInfo query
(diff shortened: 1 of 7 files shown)


src/dialects/postgres/query/compiler.js
```diff
@@ -75,6 +75,20 @@ assign(QueryCompiler_PG.prototype, {
     return 'for share';
   },
 
+  // Compiles a `listTables` query
+  listTables() {
+    return {
+      sql: 'select table_name from information_schema.tables' +
```
wubzz: Here I believe we'll want `schemaName` option somehow, since its pretty common not to be using `public`. A good start might be `where table_schema = current_schema()`, but also have an option.

I'd also recommend `table_catalog = current_catalog()` over the explicit binding.

https://www.postgresql.org/docs/current/static/functions-info.html

gpetrov:
I wonder why this never got accepted - seems like a very good PR to me...

kibertoad:
@gpetrov Because original comment was not addressed, CI is red and there are conflicts.

If you would volunteer to create a refreshed PR out of original one, we would be happy to review and potentially merge it.

gpetrov:
Well @kibertoad  - I will be glad to help, but it might be easier if the original author just finish his tests, he was so close ... @codeclown please do so, it will save us a lot of time and add a very useful feature to knex!

kibertoad:
@gpetrov Considering that this PR is 2 years old and author never addressed previous comments, it is most likely that they are not interested in this PR anymore.

codeclown:
Sorry for leaving the PR hanging, I have not had time to come back to it. I rarely use Knex nowadays. Thanks @gpetrov for creating a new one.

test/integration/builder/additional.js
```diff
@@ -216,6 +216,18 @@ module.exports = function(knex) {
       expect(knex.fn.now().toQuery()).to.equal('CURRENT_TIMESTAMP');
     });
 
+    it('should list tables', () => {
+      return knex().listTables().then(res => {
```
kibertoad, line 220: you can use async-await for simplicity

PR #2561, opened 2018-04-08
Update dependencies

kibertoad:


kibertoad:
@elhigu Nothing particularly scary or important, just a typical version bump to keep everything relevant.

PR #2564, opened 2018-04-11
Removed semicolon from rollback stmt for oracle

### LINKED ISSUE #2563: ROLLBACK; causes error with Oracle
Knex version: ^0.14.2

The ROLLBACK statement [here](https://github.com/tgriesser/knex/blob/397104c3dca2b289b2098309cc7c780e833651c4/src/transaction.js#L100) produces error `Error: ROLLBACK; - ORA-02181: invalid option to ROLLBACK WORK`

Using `ROLLBACK` without the semicolon seems to fix the problem. This is interesting since `ROLLBACK;`  is a valid statement but looks like it treats semicolon as an option here.

Any ideas how we can fix this ?

cemremengu:
Fixes #2563 (hopefully all tests will pass) 🙏

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -97,7 +97,7 @@ export default class Transaction extends EventEmitter {
   }
 
   rollback(conn, error) {
-    return this.query(conn, 'ROLLBACK;', 2, error)
+    return this.query(conn, 'ROLLBACK', 2, error)
       .timeout(5000)
       .catch(Promise.TimeoutError, () => {
         this._resolver();


cemremengu:
Failures seem unrelated 🤔

elhigu:
@atiertant any idea why that broke stuff and if this fix is fine?

cemremengu:
@elhigu Interesting, it was failing for 2 platforms when I submitted (node 4 & 8, I think) but everything seems to pass now. It recovered overnight :)

elhigu:
@cemremengu I saw only node 8 failing and reran those tests. They were only random failures, but my question to @atiertant was about this semicolon issue and if its fix is fine (this PR is missing regression test though).

cemremengu:
Wanted to add tests but wasnt sure where to add them exactly. It would be great if you give me some pointers.

atiertant:
@elhigu not sure that semicolon is the problem, maybe the problem is that the connection is reused after end of transaction (this could explain the random failures with pool).

cemremengu:
@atiertant @elhigu 
I apologize maybe I should provide more context here to be more clear and not take your time:

I was using [this plugin](https://github.com/sunfuze/egg-knex) and it was giving an error on [this line](https://github.com/sunfuze/egg-knex/blob/master/lib/knex.js#L171). Changing it to `ROLLBACK` (I submitted a PR) solved the problem.

Now playing around with the knex itself. Looks like `trx.rollback()` is not producing an error but if I do `trx.raw('ROLLBACK;')` I get the same error. Although they do rollback similarly (at least from what I understand and why I submitted this PR), maybe it was a bug only in this plugin and not knex itself?

atiertant:
@cemremengu 
i think the problem is `raw` method because it is not redefined in transaction, so it should use client's one and then use a connection from pool not the current transaction one.

kibertoad:
@elhigu Is there anything yet to happen with this PR?

tgriesser:
This seems to have broken transactions in mysql.

kibertoad:
@tgriesser We need more tests :(

elhigu:
I'm also not sure if oracledb / mssql transactions are working correctly looking this code: https://github.com/tgriesser/knex/blob/232fe9f1517dba927f6a3a1fb1b8842d7c0a4007/src/util/batchInsert.js#L51 

Also even with hack above batch insert test fails randomly on oracledb.

Transaction implementation in mssql and oracle are quite similar.

tgriesser:
> @tgriesser We need more tests :(

Sure, we also need to simplify a lot of the internals as well. I'll add a failing case in fixing this.

@elhigu Yeah I don't even think that the oracle driver uses this codepath, they seem to have their own transaction implementation internal to the connection? It's all a little tough to follow, seems a lot of this can be simplified.

PR #2565, opened 2018-04-11
Restore 'query' event to its original form (no native sql string)

### LINKED ISSUE #2549: v0.13.0 to v0.14.4 breaking sql string change for on query event
My project is currently on version 0.13.0, and I'm using the `query` event to print the full sql with the bindings injected into the string.

Typescript code:
```
this.knex.client.on('query', (data: {sql: string; bindings: any; }) => {
    var printSql = this.knex.raw(data.sql, data.bindings).toString();
    console.log(printSql);
});
```

This was possible because the `data.sql` string returns the string with `?` and `??` as the placeholder values, so using it in a knex raw object with the bindings would produce the complete query.

However when upgrading to 0.14.4, the `data.sql` string now returns the query string using the `$1, $2, $3...` placeholders which makes it impossible to easily log the complete query.

Examples:
On v0.13.0 it would return something like
`select * from "tableName" where "id" = ?`

Now on 0.14.4 it returns something like
`select * from "tableName" where "id" = $1`

Was this an intended change, and is there an easier way to compute and print the complete query?

wubzz:
Fixes #2549

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -126,7 +126,6 @@ assign(Client.prototype, {
 
   query(connection, obj) {
     if (typeof obj === 'string') obj = {sql: obj}
-    obj.sql = this.positionBindings(obj.sql);
     obj.bindings = this.prepBindings(obj.bindings)
 
     const {__knexUid, __knexTxId} = connection;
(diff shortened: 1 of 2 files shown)


wubzz:
Closing, GitHub won't update with new commit for some reason.

PR #2566, opened 2018-04-11
Bugfix/query event sql string

### LINKED ISSUE #2549: v0.13.0 to v0.14.4 breaking sql string change for on query event

wubzz:
Fixes #2549

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -126,7 +126,6 @@ assign(Client.prototype, {
 
   query(connection, obj) {
     if (typeof obj === 'string') obj = {sql: obj}
-    obj.sql = this.positionBindings(obj.sql);
     obj.bindings = this.prepBindings(obj.bindings)
 
     const {__knexUid, __knexTxId} = connection;
@@ -134,6 +133,8 @@ assign(Client.prototype, {
     this.emit('query', assign({__knexUid, __knexTxId}, obj))
     debugQuery(obj.sql, __knexTxId)
     debugBindings(obj.bindings, __knexTxId)
+
+    obj.sql = this.positionBindings(obj.sql);
 
     return this._query(connection, obj).catch((err) => {
       err.message = this._formatQuery(obj.sql, obj.bindings) + ' - ' + err.message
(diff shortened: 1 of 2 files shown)


elhigu:
Looks good. Should this be marked as breaking change?

Now that there is test for the format I think that should be mentioned in documentation. I just looked those event docs and they seem pretty bad, not telling at all what kind of data is passed to event handler.

wubzz:
I don't believe this is a breaking change, it should be quite the opposite since it restores previous functionality.

I created https://github.com/knex/documentation/issues/103 for docs as a seperate issue. As mentioned in #2549 it needs a complete overhaul, really.

This being the only thing merged since 0.14.5 I'll try to publish 0.14.6 later if I can.

elhigu:
> I don't believe this is a breaking change, it should be quite the opposite since it restores previous functionality.

Yep, this is one of those less clear cases, when functionality is changed, but since it hadn't been documented earlier, its kind of 50%/50% if it should be marked as breaking or not... especially because last time it was changed by accident.

elhigu:
I think I'll make next knex release to be 0.15 and add this to one of breaking changes and explain in log that this has been gone back and forward since 0.13.0 (this has clearly caused issue to someone, so I think reporting this change wont hurt). Maybe there will be some real breaking changes too.

PR #2594, opened 2018-05-04
chore: add Node.js 10

DanielRuf:


kibertoad:
Unfortunately, there are no prebuilt Oracle binaries for Node 10 yet, so this won't pass the CI. Not sure if we want to run tests on incomplete set of dbs.

DanielRuf:
Thanks for the hint, I'll check if CI passes and see if we can find a solution as 10 is the current stable version now.

kibertoad:
@DanielRuf see https://github.com/oracle/node-oracledb/issues/901

DanielRuf:
Ok so this can not be merged yet and I need further info from the maintainers so we can address the Node.js 10 issue(s).

DanielRuf:
Looks better now. Can we retrigger Node.js 6? We finally caught real bugs on Node.js 10 (in our unit tests).

kibertoad:
@DanielRuf close/open pr to rerun.

kibertoad:
@DanielRuf can you explicitly bump oracle driver version in package.json?

kibertoad:
@DanielRuf we didn't commit lock file before, are you sure we want to start doing it now?

DanielRuf:
> we didn't commit lock file before, are you sure we want to start doing it now?

Check the failing builds, `npm ci` is used and needs synced lockfiles ;-)

DanielRuf:
> @DanielRuf can you explicitly bump oracle driver version in package.json?

Which package do you mean in package.json?

kibertoad:
My mistake, I thought we are using it as dev depency, but apparently only travis does.

kibertoad:
@DanielRuf Both failures look like accidents, can you try rerunning the tests?

DanielRuf:
> @DanielRuf Both failures look like accidents, can you try rerunning the tests?

Rerunning them will not fix them. 10 has actual failing tests (probably due to dependencies) and 6 has an issue finding the needed apt packages.

Completely different.

I'll have to do some testing and research, probably by enabling Travis CI on my fork.

kibertoad:
@DanielRuf 
Are you sure about it?
https://travis-ci.org/tgriesser/knex/jobs/389603382 shows one test that I've definitely seen failing randomly before.

DanielRuf:
We should keep an eye on it and try to resolve this with promises or rerunning the tst if it fails.

DanielRuf:
randomely failing things are never good

kibertoad:
Yeah. I think it makes sense to create a separate issue for this.

kibertoad:
@DanielRuf Are you sure about package-lock being necessary, though? It always worked fine before, and all this PR changes is adding one more env.

DanielRuf:
> @DanielRuf Are you sure about package-lock being necessary, though?

Yes, see https://travis-ci.org/tgriesser/knex/jobs/389597882#L817

kibertoad:
I wonder where "out-of-date" package-lock.json comes in if there isn't supposed to be one. Maybe it has something to do with cache used?.. Additional "npm install" somewhere in travis script might help.
@elhigu Are you ok with adding package-lock for compatibility with newer npm, or should we investigate other options?

DanielRuf:
> Package lock has not been necessary earlier, so I don't see why it would suddenly become obligatory for this PR.

Either someone changed it to npm ci on Travis or Travis changed the defaults. I'll change the install step to npm i and revert the change with the lockfile.

kibertoad:
@DanielRuf no, it is most likely due to npm version change in newer node. If everything else fails, you could try locking node version at 10.2.1, I think it has old enough npm to just work.

elhigu:
I dont think that locking node version is a viable solution, it will just break with node 11 and above. Overriding travis scripts nay work or maybe there is some flag for npm wich uses old behavior.

DanielRuf:
It is not the default behaviour of Node.js 10 or npm to use npm ci ;-)

And lockimg the Node.js version is not a solution here.

I will add the `install` part to the Travis config.

.gitignore
```diff
@@ -1,4 +1,5 @@
 yarn.lock
+package-lock.json
```
kibertoad: this change is meaningless, so probably makes sense to restore it to how it previously was to avoid noise in history

DanielRuf: It is not meaningless, please check all changes (I just moved it to the other lockfile ignore rule).

DanielRuf: This is better as the lockfile ignore rules are both at the top.

DanielRuf: Which is more consistent. To make it clear we can also add some comments.

kibertoad: Yeah, I meant the ordering. And yes, it is more consistent, you are right.

elhigu:
it seems to work now 🎉 thank you!

DanielRuf:
@elhigu @kibertoad see https://twitter.com/travisci/status/1006652447660609536

DanielRuf:
So Travis CI changed the default behavior.

PR #2595, opened 2018-05-04
chore: cache node_modules

DanielRuf:


kibertoad:
What are the benefits of this?

DanielRuf:
It caches the `node_modules` and is partly an alternative to `npm ci`. If there is a cache with the same deps it will load the cache instead of installing all deps again = faster builds.

kibertoad:
Hell yeah, difference seems to be quite noticeable.

DanielRuf:
Can be probably better in combination with the other PRs and with following builds / PRs.

Unfortunately the env is not the best for all cases https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System

elhigu:
This seems like a good idea 👍 I have experienced issues with CI npm caches earlier (not updating to new versions to libs properly etc.), but I think we can give this a shot to see if there are any drawbacks this time (I have never used travis cache).

elhigu:
thanks!

DanielRuf:
> This seems like a good idea 👍 I have experienced issues with CI npm caches earlier (not updating to new versions to libs properly etc.), but I think we can give this a shot to see if there are any drawbacks this time (I have never used travis cache).

If this cache strategy creates any problems we can also try `$HOME/.npm` as alternative. So fr I saw no issues with the current cache strategy.

PR #2598, opened 2018-05-07
append error detail if available

capaj:
this makes it easier to see what is wrong with a migration when for example adding a foreign key, but some of the rows have values which aren't in the other table, so instead of:
```
Knex:warning - migration file "20180506221158_missing_refs.js" failed
Knex:warning - migration failed with error: alter table "user_topics" add constraint "user_topics_user_id_foreign" foreign key ("user_id") references "users" ("id") on delete CASCADE - insert or update on table "user_topics" violates foreign key constraint "user_topics_user_id_foreign"
error: insert or update on table "user_topics" violates foreign key constraint "user_topics_user_id_foreign"
    at Connection.parseE (/home/capaj/git_projects/knex-repos/knex/node_modules/pg/lib/connection.js:545:11)
    at Connection.parseMessage (/home/capaj/git_projects/knex-repos/knex/node_modules/pg/lib/connection.js:370:19)
    at Socket.<anonymous> (/home/capaj/git_projects/knex-repos/knex/node_modules/pg/lib/connection.js:113:22)
    at emitOne (events.js:116:13)
    at TCP.onread (net.js:594:20)
```

we get this in a failed migration:
```
Knex:warning - migration file "20180506221158_missing_refs.js" failed
Knex:warning - migration failed with error: alter table "user_topics" add constraint "user_topics_user_id_foreign" foreign key ("user_id") references "users" ("id") on delete CASCADE - insert or update on table "user_topics" violates foreign key constraint "user_topics_user_id_foreign"
Detail: Key (user_id)=(18) is not present in table "users".
error: insert or update on table "user_topics" violates foreign key constraint "user_topics_user_id_foreign"
    at Connection.parseE (/home/capaj/git_projects/knex-repos/knex/node_modules/pg/lib/connection.js:545:11)
    at Connection.parseMessage (/home/capaj/git_projects/knex-repos/knex/node_modules/pg/lib/connection.js:370:19)
    at Socket.<anonymous> (/home/capaj/git_projects/knex-repos/knex/node_modules/pg/lib/connection.js:113:22)
    at emitOne (events.js:116:13)
    at TCP.onread (net.js:594:20)
```

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -143,6 +143,10 @@ assign(Client.prototype, {
 
     return this._query(connection, obj).catch((err) => {
       err.message = this._formatQuery(obj.sql, obj.bindings) + ' - ' + err.message
+      const {detail} = err
+      if (detail) {
+        err.message += '\nDetail: ' + detail
+      }
       this.emit('query-error', err, assign({__knexUid, __knexTxId}, obj))
       throw err
     })
--- a/test/unit/client.js
+++ b/test/unit/client.js
@@ -1,0 +1,34 @@
+'use strict';
+/*global expect, describe, it*/
+const Client = require("../../lib/client");
+const chai = require("chai");
+
+describe('client', function () {
+  it('should append error details when available', async () => {
+    const client = new Client({ client: {} })
+    client._query = () => {
+      const err = new Error('error message')
+      err.detail = 'hmm very tasty details'
+      return Promise.reject(err)
+    }
+    try {
+      await client.query('select * from a', {sql: 'select * from a', bindings: []})
+    } catch (err) {
+      chai.expect(err.message).to.equal('select * from a - error message\nDetail: hmm very tasty details')
+    }
+  })
+
+  it('should not append error details when not available', async () => {
+    const client = new Client({ client: {} })
+    client._query = () => {
+      const err = new Error('error message')
+      return Promise.reject(err)
+    }
+    try {
+      await client.query('select * from a', {sql: 'select * from a', bindings: []})
+    } catch (err) {
+      chai.expect(err.message).to.equal('select * from a - error message')
+    }
+  })
+
+})


src/client.js
```diff
@@ -143,6 +143,10 @@ assign(Client.prototype, {
 
     return this._query(connection, obj).catch((err) => {
       err.message = this._formatQuery(obj.sql, obj.bindings) + ' - ' + err.message
+      let {detail} = err
+      if (detail) {
+        err.message += '\nDetail: ' + detail
```
kibertoad: can you add test for that?

capaj: sure I can. Do you require a functional one or will unit test be enough?

kibertoad: I think that unit test should be perfectly sufficient in this case.

capaj:
@kibertoad unit tests added

test/unit/client.js
```diff
@@ -0,0 +1,34 @@
...
+  it('should append error details when available', async () => {
+    const client = new Client({ client: {} })
+    client._query = () => {
+      const err = new Error('error message')
+      err.detail = 'hmm very tasty details'
+      return Promise.reject(err)
+    }
+    try {
+      await client.query('select * from a', {sql: 'select * from a', bindings: []})
+    } catch (err) {
```
kibertoad: wouldn't this also pass if no error is thrown altogether? See https://stackoverflow.com/questions/26571328/how-do-i-properly-test-promises-with-mocha-and-chai for approaches to test throwing promises (see second answer).

capaj: there is an error thrown in this test always here on line 12: https://github.com/tgriesser/knex/pull/2598/files/fe87405f1bbfd0b561627dceb8c5d63b560726c2#diff-cde1f14853c6f9159fe30b7e09ffdd92R12

If no error is thrown the line of code I added is not executed. Do you want me to write unit tests for code I did not commit? I can do it, but I don't really understand what it has do to with the change I opened.

kibertoad: Fair enough, just pointing out how to decrease amount of assumptions one has to make when reading or maintaining these tests, but it's really a minor nitpick. LGTM
@elhigu Do you think this can be merged?

elhigu:
Is that details attribute specific to a certain dialect? If so, probably would be good to add support for similar attribute of other dialects.

wubzz:
Is the intent of this strictly for migrations? If so, the output could be changed specifically in the [migrations error handler](https://github.com/tgriesser/knex/blob/master/src/migrate/index.js#L260).

Right now this change would also spew out details for all queries that have been run, including those that may be logged in a logservice or similar. If that is the intent then this is all good. Just wanted to mention that.

capaj:
@wubzz I didn't have the need anywhere else but in the migrations. Will move the log there instead of the current place. Thanks for the info! Will reopen when I change it.

Issue #2603, opened 2018-05-14
ORACLE: Connection is not destroyed once database is not reachable


cemremengu:
@atiertant

Knex version: 0.14.6
Database + version: Oracle 11gR2, Client 12cR2
Client: oracledb (official)

Not sure if there are any workarounds for this, since I have seen many resolved issues about it (for other databases though) but I am having this problem for some reason.

If connection is lost while an application is running, minimum acquired connections waiting in the pool do not get destroyed. As a result, even when connection is restored, application will still be getting not connected (`ORA-03114: not connected to ORACLE`) error. The only way to fix at that point is to restart the application.

However, if pool is configured with the option `min:0` this error won't happen. 

What I think is once a connection receives either of

- `ORA-03113: end-of-file on communication channel` 
- `ORA-03114: not connected to ORACLE` 
- ` ORA-03135: connection lost contact` 

it should be evicted from the pool / destroyed no matter what the minimum number of connection configuration is set.

I have tried setting but they did not fix the problem

```
idleTimeoutMillis: 30 * 1000,
createTimeoutMillis: 3 * 1000,
acquireTimeoutMillis: 3 * 1000,
```

cemremengu:
Closing to continue discussion from there

Issue #2514, opened 2018-02-26
Failure to Reconnect to Oracle 12c.


klfgit:
# Environment
node: v8.6.0
oracle db: 2.0.15
Knex version: 0.14.2
Database + version: oracle 12c
OS: REPL 3.10.0-693.17.1.el7.x86_64 #1 SMP Sun Jan 14 10:36:03 EST 2018 x86_64 x86_64 x86_64 GNU/Linux

Select applicable tempalate from below.
If issue is about oracledb support tag @ atiertant. For MSSql tag @ Grimace1975 .
Rest of dialects doesn't need tags.

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do

What I see: Knex does not reconnect after the db bounces.  
What I expect: Knex should reconnect after the db bounces.

2. Error message

> Error: ORA-03114: not connected to ORACLE

3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real
   database connection to mysql or postgresql, then single file example which initializes
   needed data and demonstrates the problem.

My test procedure:
1. Start the DB.
2. Run the code below.
3. Stop the DB.
4. Once I see the 'Error: ORA-03113: end-of-file on communication channel' error.
5. Wait a few seconds then restart the DB.  Once the DB is up I expect to start seeing the SYSDATE again.

```
'use strict';

var knex = require('knex')({
    client : 'oracledb',
    connection: {
        "user": "user",
        "password": "pwd",
        "database": "sid",
        "host": "host:port"
    },
    debug: false,
    pool: {
        "min": 0,
        "max": 3
    }
});

 function  exitIfDBIsDown() {
    knex.raw('select sysdate from dual').then(resp => {
       console.log(resp);
    }).catch(async err => {
        console.log(err);
    });
}

setInterval(exitIfDBIsDown, 2*1000);
```

example output: 

> [ { SYSDATE: 2018-02-27T02:09:24.000Z } ]
> [ { SYSDATE: 2018-02-27T02:09:25.000Z } ]
> [ { SYSDATE: 2018-02-27T02:09:27.000Z } ]
> [ { SYSDATE: 2018-02-27T02:09:29.000Z } ]
> [ { SYSDATE: 2018-02-27T02:09:31.000Z } ]
> [ { SYSDATE: 2018-02-27T02:09:33.000Z } ]
> [ { SYSDATE: 2018-02-27T02:09:35.000Z } ]
[...]

PR #2608, opened 2018-05-15
ORACLEDB: Updated handling of connection errors for disposal

cemremengu:
Solves #2514 and #2603. I tested it on my-machine 😄 I am thinking that simulating this will be tricky and probably take about 15-20 minutes for oracle (or maybe I am exaggerating)

In addition to updating connection error list, we also need to update `oracledb` to check for connection errors.

Disclaimer: I am not good with vanillajs (prototypes etc.) so it would be great if someone reviews the syntax/approach.


elhigu:
To me code looks pretty good, I would need to checkout the connection.__knex__disposed setting if that is the thing that should be done there (probably it is correct though). @atiertant how does tihs looks to you?

src/dialects/oracledb/index.js
```diff
@@ -135,13 +135,19 @@ Client_Oracledb.prototype.acquireRawConnection = function() {
         if (options.resultSet) {
           connection.execute(sql, bindParams || [], options, function(err, result) {
             if (err) {
+              if (Client_Oracle.prototype.isConnectionError(err)) {
```
atiertant: doesn't client.isConnectionError reference the same fonction ?

atiertant:
@cemremengu oracle database has a dead connection detection mechanism but we should call connection.close before set the connection.__knex__disposed.
@elhigu looks great, but didn't tested it

cemremengu:
@atiertant calling `connection.close` seems to throw an error though so I am not sure if it achieves anything
```
 ERROR 8672 nodejs.unhandledRejectionError: NJS-003: invalid connection
```

atiertant:
@cemremengu see node-oracledb examples
https://github.com/oracle/node-oracledb/blob/master/examples/select1.js#L73
should ask @cjbj on a node-oracledb issue which errors needs to close the connection and which other need only to remove the connection object from pool ...
can't find this kind of mechanism in the node-oracledb driver pool, does it has the same bug ?

cemremengu:
Added connection.close() to conform what @cjbj stated

> Closing the connection is the way to remove it from the connection pool, which is important so the pool slot can be reused by a new valid connection. You should always call conn.close(). What you may want to do is ignore errors on the close.

src/dialects/oracle/index.js
```diff
@@ -179,16 +179,22 @@ assign(Client_Oracle.prototype, {
       default:
         return response;
     }
-  }
+  },
 
+  // If the error is any of these, we'll assume we need to
+  // mark the connection as failed
+  isConnectionError(err) {
```
cjbj: See https://github.com/oracle/odpi/blob/master/src/dpiError.c#L65 for a bigger error list used by the node-oracledb connection pool.  Node-oracledb also uses an OCI call at https://github.com/oracle/odpi/blob/29968fc2085941c40fce8f56a42c9920305d2e29/src/dpiConn.c#L204 which does most of the same checks anyway.

src/dialects/oracle/index.js
```diff
@@ -179,16 +179,22 @@ assign(Client_Oracle.prototype, {
...
+    return [
+      'ORA-03114', // not connected to ORACLE
+      'ORA-03113', // end-of-file on communication channel
+      'ORA-03135', // connection lost contact
+      'ORA-12514', // listener does not currently know of service requested in connect descriptor
+      'NJS-040',
+      'NJS-024',
+      'NJS-003',
+      'NJS-024',
```
cjbj: NJS-024 is duplicated a few lines above.  And arguably shouldn't be on the list.

cjbj:
The issue heading mentions oracledb but the files changed are for dialect/oracle, not dialect/oracledb.  Is this what you intended?

Also, my comments about connection pooling are for the node-oracledb connection pool, which isn't used in dialects/oracledb

kibertoad:
@cjbj Does that mean that currently used list of errors that should close connections is invalid in case of not using oracledb connection pool?

cjbj:
@kibertoad the ORA errors that indicate a bad connection are pretty much going to be the same for pooled & non-pooled connections.  You could check for the same numbers that ODPI-C uses, see my earlier link.

kibertoad:
@cemremengu Could you please provide me write permissions to your repo so that I could fix the PR accordingly to cjbj's comments without forking and resubmitting the pr?

kibertoad:
@cjbj Thanks!

cemremengu:
@kibertoad Sorry, snowed under work lately. Updated the error list accordingly

kibertoad:
@cemremengu looks like there are random failures, can you rerun tests by closing and opening pr?

cemremengu:
@kibertoad done. Also given you permission in case you want to tweak stuff

kibertoad:
awesome, thank you! also tests passed, yay :)

elhigu:
I think this could be tested by mocking/monkeypatching client.isConnectionError method and returning the original implementation after tests.

kibertoad:
@elhigu What should such test assert?

elhigu:
@kibertoad that connection state is set to disposed in all of the changed code paths (test should fail before this change).

alanchristensen:
What's the status of this PR?

kibertoad:
@alanchristensen I could resolve the merge conflict, but would really appreciate it if someone else wrote the test :)

pablopen:
Hi, any updates on this?
Looks like it could solve the issues when oracle connections are consuming one pool's connection and failing each request that fall in that connection.

kibertoad:
@pablopen Any chance you would volunteer to write at least some test for it?

pablopen:
@kibertoad I could try. I'll check elhigu suggestion of mocking the errors.

kibertoad:
@pablopen You are my hero :)

elhigu:
I would still have preferred having some tests for this, now this is waiting for regression to happen if anyone ever touches that code.

kibertoad:
@elhigu In ideal world, yes. However, year passed, and noone did, and I lost all hope to ever get to it myself; and this change is both straightforward enough and useful (as evidenced by already reported problems solved by it).

elhigu:
In this particular case I considered that benefits were bigger than risks of it causing problems later on.

PR #2609, opened 2018-05-16
Support postgresql connection uri protocol

cranberyxl:
Postgres supports 2 uri connection protocols:

> The URI scheme designator can be either postgresql:// or postgres://.

https://www.postgresql.org/docs/current/static/libpq-connect.html#id-1.7.3.8.3.6


elhigu:
Thanks!

Issue #2606, opened 2018-05-14
.alter() doesn't call wrapIdentifier


heisian:
Initial migration with `wrapIdentifier` (`knexSnakeCaseMappers` from Objection.js) being used for snake_case to camelCase conversion:
```node
exports.up = (knex) => knex.schema.withSchema('admin').table('policyLogs', (t) => {
  t.string('actionableType')
})
// Results in admin.policy_logs.actionable_type being created
```

column-altering migration:
```node
exports.up = (knex) => knex.schema.withSchema('admin').table('policyLogs', (t) => {
  t.text('actionableType').alter()
})
```

resulting SQL/error:
```
Knex:warning - migration failed with error: alter table "admin"."policy_logs" alter column "actionableType" drop default - column "actionableType" of relation "policy_logs" does not exist
```

Passing in the snake-cased column name works:
```node
exports.up = (knex) => knex.schema.withSchema('admin').table('policyLogs', (t) => {
  t.text('actionable_type').alter()
})
```

elhigu:
https://runkit.com/embed/m9gp3b02aef9

seems to fail at least with postgresql, mysql works, didn't test others right now.

heisian:
@elhigu #2612 think i'm missing anything here?

heisian:
Fixed by #2612

PR #2612, opened 2018-05-16
fix wrapIdentifier not being called in postgres alter column

heisian:
#2606


kibertoad:
@heisian No tests?..

heisian:
Chat would I test? There's an unlimited amount of possibilities for `wrapIdentifier` since it's a user-customizable function.

Should I test camelCase to snake_case conversion? That's what I wanted this fix for. But who's to say that same conversion would be used anywhere else (even though we know it is)? For all we know, someone else out there could be using `wrapIdentifier` to add prefixes to all their column operations.

IMO, and unless it's proved differently, whatever tests already exist to ensure that existing functionality is the same as before is what matters here.

heisian:
OK, how about this? I'm really wondering if there's a way to pass the value from the first time `wrapIdentifier` is called when invoking `table.string('email')` rather than having to call it again explicitly within the postgres `TableCompiler`.

test/unit/schema/postgres.js
```diff
@@ -786,6 +786,19 @@ describe("PostgreSQL SchemaBuilder", function() {
       expect(spy.secondCall.args).to.deep.equal(['email', 'email context']);
       expect(spy.thirdCall.args).to.deep.equal(['users', 'table context']);
     });
+
+    it('TableCompiler calls wrapIdentifier when altering column', function () {
+      client
```
elhigu: How does this test test that wrapIdentifier was called correctly?

src/dialects/postgres/schema/tablecompiler.js
```diff
@@ -46,35 +46,35 @@ TableCompiler_PG.prototype.addColumns = function(columns, prefix, colCompilers)
     // alter columns
     for (const col  of colCompilers) {
       const quotedTableName = this.tableName();
-      const colName = col.getColumnName();
+      const colName = this.client.wrapIdentifier(col.getColumnName(), col.columnBuilder.queryContext());
```
elhigu: maybe this should be something like `this.formatter.wrap(col.getColumnName())`

test/unit/schema/postgres.js
elhigu: ah right, that spy was set on wrapIdentifier function...  original naming of that spy should probably have been done better :) 👍 This is good.

src/dialects/postgres/schema/tablecompiler.js
heisian: I did try that when running tests as well - I'll use that method instead.

heisian:
@elhigu nevermind, I remember why I didn't end up using it, it seems that `wrapAsIdentifier` doesn't get the `queryContext` from `columnBuilder`, and I think the reason is `this.formatter.builder` has the `queryContext` from `table.queryContext` and not `alter().queryContext`.

If I do `process.stdout.write(JSON.stringify(this.formatter.builder.queryContext()))` the output is: `table context`. And not `alter context`.

So any other ideas other than calling `wrapIdentifier` as-is?

heisian:
When modifying the line as you suggested:
screen shot 2018-05-21 at 5 47 47 pm

elhigu:
Maybe allowing to pass the object holding the context to the `this.formatter.wrapAsIdentifier()`? I would be surprised if that same problem you encountered is not accidentally done some elsewhere.

heisian:
Yeah.. the real problem is that the results of `wrapIdentifier` aren't passed into the `alter` block.

What seems to happen is:

1) `wrapIdentifier` is called because of `table.queryContext`
2) `wrapIdentifier` is called because of `table.string('someColumn')`
3) The results of 2) do not get passed into `.alter()`

I couldn't figure out how to pass it through. Within the `alter` block, it is passed `this.formatter` containing the query context of the `table`, instead of the column.

So maybe we need to find where we can pass it the `this.formatter` from 2) instead of 1).

elhigu:
I think adding comment to the place where wrapIdentifier is called instead of formatter is enough for now. We really need to do some bigger refactoring to knex internals at some point.

heisian:
thank you!

PR #2625, opened 2018-05-23
Allow overwriting log functions

wubzz:
The idea is knex has some built-in log functions just as before, but allow the user to overwrite these functions as such:

```javascript
    client: 'pg',
    connection: {...},
    log: {
      warn(message) {
      },
      error(message) {
      },
      deprecate(message) {
      },
      debug(message) {
      },
    }
});
```

This also means seperate clients can use seperate log handlers. Relates to a lot of issues..

There's another [ongoing build](https://github.com/tgriesser/knex/commit/5b6601d1d709d676691dacb2502619ba0f3887ee) with similar feature. I'm not clear on what the differences are between the two.

This may be completely wrong way of implementing this, I'm mostly creating the PR solely to give the topic a rebirth, since it has been requested many times before.


elhigu:
Looks like the difference is that your logger allows to pass different logger handler for every knex instance and the other solution used global logger, which is easier to use by the library (you can just require it in in everywhere without  need to pass it everywhere separately).

Personally I like the less intrusive way of doing it a bit more, but I don't have very strong opinion about it. It is quite easy to change to global logger, if passing the logger instance around gets annoying.

test/tape/harness.js
```diff
@@ -35,6 +35,7 @@ module.exports = function(tableName, knex) {
         var val = cb(t)
         if (val && typeof val.then === 'function') {
           return val.catch(function(err) {
+            console.error(err);
```
tgriesser: Was this intentional related to logging or can we remove it?

tgriesser:
Damn @wubzz this is awesome! Just had one comment, otherwise yeah this is what I was eventually hoping we would get to, with [this branch](https://github.com/tgriesser/knex/commit/5b6601d1d709d676691dacb2502619ba0f3887ee) being an incremental step but this looks much better!

test/tape/harness.js
wubzz: Ah that's my bad. Stacktraces aren't printed when running tape and I had an issue in maria or oracle which I couldn't run locally at the time of making this, so abused Travis. 😄  I'll revert this back.

tgriesser:
Awesome!

wubzz:
Added docs https://github.com/knex/documentation/pull/118

src/logger.js
```diff
@@ -0,0 +1,55 @@
+/* eslint no-console:0 */
+
+import chalk from 'chalk';
+import {isFunction, isNil} from 'lodash';
+
+function log(message, userFn, colorFn) {
+  if(!isNil(userFn) && !isFunction(userFn)) {
```
kibertoad: wouldn't logic be simpler if you didn't have to do the check on each log action and instead filled undefined methods in config with dummy console.log implementation?

src/logger.js
```diff
@@ -0,0 +1,55 @@
...
+      } = {}
+    } = config;
+
+    this._debug     = debug;
+    this._warn      = warn;
+    this._error     = error;
+    this._deprecate = deprecate;
+  }
+
+  debug(message) {
```
kibertoad: this is going to swallow any parameters passed to the logger (e. g. error instance). Using 'arguments' instead is going to be much more robust.

src/migrate/index.js
```diff
@@ -412,7 +413,7 @@ function validateMigrationList(migrations) {
 
 function warnPromise(value, name, fn) {
   if (!value || typeof value.then !== 'function') {
-    helpers.warn(`migration ${name} did not return a promise`);
+    this.knex.client.logger.warn(`migration ${name} did not return a promise`);
```
leontastic: @wubzz This gives a fatal TypeError `Cannot read property 'knex' of undefined`

PR #2626, opened 2018-05-23
Timeout errors shouldn't silently ignore the passed errors

### LINKED ISSUE #2582: Silently ignoring rollback timeout swallows transaction error
Knex version: 0.14.3

# Bug

See old pull request where I believe this was added: https://github.com/tgriesser/knex/pull/1296

The issue is that if a rollback times out, the original transaction error will be swallowed. I believe this is the code:

````js
// src/transaction.js
rollback(conn, error) {
    return this.query(conn, 'ROLLBACK;', 2, error)
      .timeout(5000)
      .catch(Promise.TimeoutError, () => {
        this._resolver();
      });
  }
````

As you can see the original transaction error is forwarded to query, but since the rollback times out the original error is not thrown. There is no harm in swallowing the error from the rollback itself, but that's not what is happening here.

wubzz:
but rather reject with original error. Fixes #2582

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -100,7 +100,7 @@ export default class Transaction extends EventEmitter {
     return this.query(conn, 'ROLLBACK;', 2, error)
       .timeout(5000)
       .catch(Promise.TimeoutError, () => {
-        this._resolver();
+        this._rejecter(error);
       });
   }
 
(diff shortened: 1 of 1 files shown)


elhigu:
Sorry for huge delays of hauling through issues, we've been having few weeks sunny days here :)

elhigu:
i just wrote test to another branch which fails without these changes, so all good and test is on its way.

PR #2632, opened 2018-05-26
Add support for native enums on Postgres

### LINKED ISSUE #394: native enums
postgres has enums, we shouldn't have to manually create one with constraints

willsoto:
This is my initial pass at implementing native enum support for **Postgres only.** 

I am looking for feedback on the implementation and whether or not this is the _best_ way to solve this.

I followed the advice of [this comment](https://github.com/tgriesser/knex/issues/394#issuecomment-315528899) and added the following argument to the `enu` method:

```js
{
  nativeEnum: boolean,
  enumName: string
}
```

If `nativeEnum` is truthy, I use the `enumName` to generate the appropriate `type`. I am not sure if having two fields is entirely necessary (I could check `enumName` alone and use that) but I am not sure of the behavior for other RDBMS's.

Reference https://www.postgresql.org/docs/current/static/sql-createtype.html

Documentation: https://github.com/knex/documentation/pull/117

Closes #394

### DIFF
--- a/src/dialects/postgres/schema/columncompiler.js
+++ b/src/dialects/postgres/schema/columncompiler.js
@@ -30,8 +30,17 @@ assign(ColumnCompiler_PG.prototype, {
 
   // Create the column definition for an enum type.
   // Using method "2" here: http://stackoverflow.com/a/10984951/525714
-  enu(allowed) {
-    return `text check (${this.formatter.wrap(this.args[0])} in ('${allowed.join("', '")}'))`;
+  enu(allowed, options) {
+    options = options || {};
+
+    const values = allowed.join("', '");
+
+    if (options.useNative) {
+      this.tableCompiler.unshiftQuery(`create type "${options.enumName}" as enum ('${values}')`);
+
+      return `"${options.enumName}"`;
+    }
+    return `text check (${this.formatter.wrap(this.args[0])} in ('${values}'))`;
   },
 
   double: 'double precision',
--- a/src/schema/columncompiler.js
+++ b/src/schema/columncompiler.js
@@ -25,6 +25,8 @@ ColumnCompiler.prototype.pushQuery = helpers.pushQuery
 
 ColumnCompiler.prototype.pushAdditional = helpers.pushAdditional
 
+ColumnCompiler.prototype.unshiftQuery = helpers.unshiftQuery
+
 ColumnCompiler.prototype._defaultMap = {
   'columnName': function() {
     if (!this.isIncrements) {
--- a/src/schema/compiler.js
+++ b/src/schema/compiler.js
@@ -1,5 +1,5 @@
 
-import { pushQuery, pushAdditional } from './helpers';
+import { pushQuery, pushAdditional, unshiftQuery } from './helpers';
 
 import { assign, isUndefined } from 'lodash'
 
@@ -19,6 +19,8 @@ assign(SchemaCompiler.prototype, {
   pushQuery: pushQuery,
 
   pushAdditional: pushAdditional,
+
+  unshiftQuery: unshiftQuery,
 
   createTable: buildTable('create'),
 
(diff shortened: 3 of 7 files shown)


willsoto:
@elhigu So does that mean you are okay with the implementation?

src/dialects/postgres/schema/columncompiler.js
```diff
@@ -30,8 +30,19 @@ assign(ColumnCompiler_PG.prototype, {
...
-  enu(allowed) {
-    return `text check (${this.formatter.wrap(this.args[0])} in ('${allowed.join("', '")}'))`;
+  enu(allowed, options) {
+    options = options || {};
+
+    const values = allowed.join("', '");
+
+    if (options.nativeEnum) {
+      this.pushAdditional(function() {
+        this.pushQuery(`create type ${options.enumName} as enum ('${values}')`)
```
elhigu: shouldn't enumName be wrapped to some kind of quotes?

elhigu:
@willsoto yep, implementation looks fine. It's backwards compatible which is the most important part.

willsoto:
Thanks! I will look into the other dialects. Judging from the integration tests, it looks like these are the ones that have support:

https://github.com/tgriesser/knex/blob/db390b11e68a7c023910b6f371fe851f371028f8/test/integration/schema/index.js#L292-L305

Does that seem right?

elhigu:
You need to check out from each SQL dialect's specs which has native support for enum type.

timwis:
Just want to say thanks for working on this! Was just wondering why my enum column wasn't supporting `<=` and realised knex just creates a `text` column with constraints. Coincidental timing you're working on this today.

src/dialects/postgres/schema/columncompiler.js
willsoto: I guess if it is camel cased it should be, just like every other identifier. I will double check though.

willsoto:
@elhigu After doing some research, it seems this change is only required for Postgres. 

- Native `ENUM` for mysql/mysql2/mariadb is already supported [here](https://github.com/tgriesser/knex/blob/master/src/dialects/mysql/schema/columncompiler.js#L79-L81). See [this](https://dev.mysql.com/doc/refman/8.0/en/enum.html)
- SQL Server does not have native enum. See [this](https://stackoverflow.com/a/1434338) and [this](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/c7abe3c1-cd48-4ddf-9ce2-b28d0ffb25f6/enum-in-sql-server?forum=transactsql)
- SQLite 3 does not have native enum. See [this](http://www.sqlite.org/datatype3.html)
- Oracle does not have native enum. See [this](https://stackoverflow.com/a/203547)
- Redshift does not support Postgres's native enum. See [here](https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-datatypes.html)

---

I do have one question though, since I am not entirely sure how to proceed.

In order to have the column be the enum `type`, I need to create it _prior_ to the creation of the table. Looking through the source code, I can't seem to find a way to do that. Any suggestions?

willsoto:
I was trying that out actually. I couldn't quite get it, but it's good to
know I'm on the right track. I'll keep at it.

willsoto:
@elhigu Okay. I think I got it. I added a method called `unshiftQuery` and called that method on the table when creating the column. Let me know if you have any thoughts on `unshiftQuery` - I just copied the `pushQuery` method and `unshift`ed the item.

--- 

Currently 2/3 of the test suites are passing. It isn't clear to me why it is failing on Node 6 though. All tests are passing on there as well, but it seems to exit with a non-zero exit code. Any thoughts?

kibertoad:
@willsoto Have you tried rerunning the tests? Sometimes build just fails randomly.
(I think closing and reopening the PR or any dummy commit would do the trick)

tgriesser:
Cool, was talking to @willsoto this morning and this looks good to me. Thanks!

PR #2634, opened 2018-05-29
Remove babel-plugin-lodash

tgriesser:
While in theory, this may reduce the bundle size, in practice it adds a ton of overhead during startup due to the number of additional requires. Bundle size also shouldn't matter for server side modules.


elhigu:
Some oracledb migration test has been always failing every now and then... Also sometimes for very random segmentation fault when tests are already in the end. 

I don't remember any other random failures.

tgriesser:
> I've been planning to try to drop installing oracledb on normal travis environment, since its installation is a bit unstable sometimes. I have already docker based install for it which can be used also locally for running tests in oracle. That should be possible to run in travis as well.

Cool, sounds like a good plan!

PR #2635, opened 2018-05-29
Add json support to the query builder for mysql

kibertoad:
Minimalistic version of #1902, only adding support for JSON column type for MySQL.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,9 +31,18 @@ after_script:
 notifications:
   email: false
 
+services:
+  - mysql
 addons:
   postgresql: '9.6'
   apt:
+    sources:
+      - mysql-5.7-trusty
     packages:
-    - g++-4.8
-    - gcc-4.8
+      - g++-4.8
+      - gcc-4.8
+      - libmysqlclient-dev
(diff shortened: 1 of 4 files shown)


kibertoad:
@elhigu Any ideas why build may keep failing on Node 6 but works fine on Node 8? Judging by logs, something is wrong with downloading Oracle archive - 
`error [oracle-xe-11.2.0-1.0.x86_64.rpm.zip]:  missing 94371840 bytes in zipfile`
However, why would it fail with one node but not the other? My suspicion is that VM runs out of space due to new libraries added, but that would also probably happen with Node 8.

elhigu:
@kibertoad bad luck I suppose... some times that downloading is really unstable and sometimes it works better. I will do something about it some day...

elhigu:
I had already triggered rebuild for that failing node version... well fingers crossed that now all of them pass :)

kibertoad:
@elhigu is pr itself ok? Probably makes sense to get it within 0.15 window.

kibertoad:
Ah, merged :)

PR #2637, opened 2018-05-30
Fix mssql driver crashing

elhigu:
Since mssql driver has not been maintained, locked driver version and monkey patched it to fix couple of problems of leaking connections / hanging and crashing when connections are closed unexpectedly.

We should eventually change mssql dialect to use directly tedious driver.

I did run this stress test script overnight and it didn't crash and memory consumption didn't grow either (it was stable between 150MB - 400MB)

### DIFF
--- a/scripts/stress-test/knex-stress-test.js
+++ b/scripts/stress-test/knex-stress-test.js
@@ -153,18 +153,18 @@ async function main() {
 
   await recreateProxies();
 
-  /*
   loopQueries('PSQL:', pg.raw('select 1'));
   loopQueries('PSQL TO:', pg.raw('select 1').timeout(20));
 
   loopQueries('MYSQL:', mysql.raw('select 1'));
   loopQueries('MYSQL TO:', mysql.raw('select 1').timeout(20));
 
-  loopQueries('MYSQL2:', mysql2.raw('select 1'));
-  loopQueries('MYSQL2 TO:', mysql2.raw('select 1').timeout(20));
+// mysql2 still crashes app (without connection killer nor timeouts)
+// https://github.com/sidorares/node-mysql2/issues/731
+//  loopQueries('MYSQL2:', mysql2.raw('select 1'));
+//  loopQueries('MYSQL2 TO:', mysql2.raw('select 1').timeout(20));
 
   loopQueries('MSSQL:', mssql.raw('select 1'));
-*/
   loopQueries('MSSQL TO:', mssql.raw('select 1').timeout(20));
 
   setInterval(recreateProxies, 2000);
@@ -173,8 +173,8 @@ async function main() {
     await Bluebird.delay(20); // kill everything every quite often from server side
     try {
       await Promise.all([
-//        killConnectionsPg(),
-//        killConnectionsMyslq(mysql),
+        killConnectionsPg(),
+        killConnectionsMyslq(mysql),
 //        killConnectionsMyslq(mysql2),
         killConnectionsMssql()  
       ]);
--- a/scripts/stress-test/knex-stress-test.js
+++ b/scripts/stress-test/knex-stress-test.js
@@ -24,6 +24,18 @@ const mysql = Knex({
   pool: { max: 50 }
 });
 
+const mssql = Knex({
+  client: 'mssql',
+  connection: {
+    port: 21433,
+    user: "sa",
+    password: "S0meVeryHardPassword",
+    server: "localhost",
+    requestTimeout: 500
+  },
+  pool: { max: 50 }
+});
+
 /* TODO: figure out how to nicely install oracledb node driver on osx
 const mysql = Knex({
   client: 'oracledb',
@@ -50,6 +62,7 @@ function setQueryCounters(instance, name) {
 setQueryCounters(pg, 'pg');
 setQueryCounters(mysql, 'mysql');
 setQueryCounters(mysql2, 'mysql2');
+setQueryCounters(mssql, 'mssql');
 
 const _ = require('lodash');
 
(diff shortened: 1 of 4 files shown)


scripts/stress-test/knex-stress-test.js
```diff
@@ -130,27 +148,35 @@ async function main() {
...
+  setInterval(recreateProxies, 2000);
 
   while(true) {
     await Bluebird.delay(20); // kill everything every quite often from server side
     try {
       await Promise.all([
         killConnectionsPg(),
         killConnectionsMyslq(mysql),
-        killConnectionsMyslq(mysql2),
+//        killConnectionsMyslq(mysql2),
```
kibertoad: Why was this commented out?

scripts/stress-test/knex-stress-test.js
```diff
@@ -24,6 +24,18 @@ const mysql = Knex({
...
+    password: "S0meVeryHardPassword",
+    server: "localhost",
+    requestTimeout: 500
+  },
+  pool: { max: 50 }
+});
+
 /* TODO: figure out how to nicely install oracledb node driver on osx
 const mysql = Knex({
   client: 'oracledb',
```
kibertoad: btw, why mysql is using oracledb client?..

elhigu: just some initial copy-paste for writing oracledb test after mssql is all done.

dhensby:
Is this problem now fixed with the release of 4.2.1? (we got a PR merged that fixed the depletion of connections error - https://github.com/tediousjs/node-mssql/pull/683)

kibertoad:
@elhigu We can probably try updating driver, removing monkey-patching and running existing stress tests?..

dhensby:
It would be very helpful if this limitation was removed because we can't upgrade our knex version as we need the mssql fix in production ;)

kibertoad:
@dhensby Doesn't monkey-patching introduced by @elhigu solve the same issue as well in newer knex version?

dhensby:
@kibertoad TBH, I'm not sure and it's not something we'd be prepared to put time into investigating. We had issues on azure in our production application which we spent a long time debugging and resolving, culminating in getting https://github.com/tediousjs/node-mssql/pull/683 merged.

It's not going to make business sense for us to downgrade from a known working patch "just" so we can upgrade knex.

The reason to upgrade knex would be so that `knex.schema.unique` and `knex.schema.dropUnique` do as expected. The work around to that is far more simple and less risky than rolling back a working patch and risking depleting our connection pool in prod again.

For now we'll use `knex.schema.unique([cols], 'our_own_index_name');` and `knex.schema.dropIndex([cols], 'our_own_index_name');` which works and I'm sure you can appreciate is insignificant in terms of risk to prod vs downgrading the mssql lib.

kibertoad:
@dhensby That makes sense. I'll try to get stress tests working locally on my machine to get the answer.

kibertoad:
@dhensby Wait, was that version ever released? https://www.npmjs.com/package/mssql still shows 4.1.0 as the latest published one.

dhensby:
https://www.npmjs.com/package/mssql/v/4.2.1 - it seems to be `@next` for some reason, but we're definitely using it...

It's not the most maintained package out there ;)

kibertoad:
@dhensby Does stability-wise (other than this specific issue) 4.2.1 perform same or better than the current "stable" 4.1.0 for you? In general it is preferable to avoid "next" dependencies in libraries, I wonder if we should make an exclusion for this particular case.

dhensby:
Yeah, that's a good point, I hadn't realised it was a "next" release, though, anyone wanting to run mssql needs the fix that is in 4.2.1.

In production we haven't had any issues with MSSQL, the library or other symptoms since upgrading to 4.2.1, so we are happy with its stability.

Another option would be to allow users to set some kind of config to opt-out of the 4.1.0 restriction so those that want to run the higher versions can?

This is the diff between 4.1.0 and 4.2.1 (https://github.com/tediousjs/node-mssql/compare/85c14ff5b01b1ae0575591c7e4f1c7327cf82e51...master) and it doesn't look too risky to me.

dhensby:
I've asked that they release 4.2.1 as stable (https://github.com/tediousjs/node-mssql/issues/703) though not sure how likely that is to be actioned :(

elhigu:
I'd still rather get rid of mssql driver and use tedious directly than update knex to support newer mssql. It will be easier to make to work in a robust manner than trying to fix mssql to work well.

I'm open to help / review if anyone is going to change mssql -> tedious, but I really don't want to work with node-mssql's code anymore.

kibertoad:
@elhigu I can absorb the effort of updating the supported mssql version as that doesn't sound like lots of work, but mssql -> tedious is going to require much more elaborate knowledge of SSQL and Tedious, and I don't even use MSSQL myself :D

elhigu:
Not much more plain sql knowledge is needed... node-mssql already uses tedious pretty straightforward manner, but ot just forces to use one additional (and buggy) generic-pool for getting connections which causes most of the problems.

kibertoad:
There is already a PR to replace generic-pool with Tarn.js for node-mssql. Would working on it to make it merged help alleviate the pain?

dhensby:
Whilst I totally understand that and using the mssql package has caused us a lot of pain too, I don't think that larger objective should block derestricting the package currently supported.

For lots of developers (especially ones in our position) throwing out node-mssql is non-starter and simply means we won't be able to upgrade knex, which would be a shame

elhigu:
> There is already a PR to replace generic-pool with Tarn.js for node-mssql. Would working on it to make it merged help alleviate the pain?

I would rather see PR which allows to use node-mssql without any additional pool on top of tedious. Tarn is developed in terms of knex, so it might not be the best choice for mssql driver, but yeah with that it would be definitely easier to fix at least.

elhigu:
> Whilst I totally understand that and using the mssql package has caused us a lot of pain too, I don't think that larger objective should block derestricting the package currently supported.

> For lots of developers (especially ones in our position) throwing out node-mssql is non-starter and simply means we won't be able to upgrade knex, which would be a shame

How are you actually using mssql driver directly? What difference it makes for you if you have mssql wrapping tedious when you are actually using knex to interface it?

dhensby:
We only use knex for our DB migrations / schema building. All our DB interfacing happens through stored procedures and an internal DB SDK that provides an API to those stored procedures.

As knex doesn't support stored procedures we don't use knex for our actual DB SDK and so use the mssql package to bind the params and execute the SPs.

elhigu:
In that case you can still use knex the same way that before and node-mssql for the same stuff you are using it currently. 

Knex just internally wont use node-mssql anymore.

dhensby:
That's a good point ;)

Issue #2601, opened 2018-05-10
ORACLE : node stream is not catching the exception


fvalmeida:
# Environment

Knex version: 0.14.4
Database + version: Oracle 12C
OS: Mas OSX

# Bug 

1. Problem

@atiertant The node stream is not catching the exception

2. Error message

> events.js:183
>       throw er; // Unhandled 'error' event
>       ^
> Error: ORA-00942: table or view does not exist

3. Reduced test code:
```
const knex = require('knex')({
    dialect: 'oracledb',
    connection: {
        host: process.env.HOST,
        user: process.env.USER,
        password: process.env.PASSWORD,
        database: process.env.DATABASE        
    }
});

knex('xXxXxXxXx')
    .stream((stream) => {

        stream.on('data', function (chunk) {
            //
        });
    })
    .catch(function (error) {
        console.error('Ooops!')
    });
```

4. Expected behavior:

It should have printed 'Ooops!' in the console because the table `xXxXxXxXx` doesn't exist.

Proposed fix:

Include these lines into **../dialects/oracle/index.js**

```
_stream: function _stream(connection, obj, stream, options) {
    return new _bluebird2.default(function (resolver, rejecter) {
      stream.on('error', function (err) {
        if (isConnectionError(err)) {
          connection.__knex__disposed = err;
        }
        rejecter(err);
      });
      stream.on('end', resolver);
      var queryStream = connection.queryStream(obj.sql, obj.bindings, options);
      queryStream.pipe(stream);
      // ***** add this line *************
      queryStream.on('error', rejecter);
      // ********************************
    });
  },
```

mercmobily:
Since the reporter didn't mention anything, and since there is a closed PR about it, I closed this issue.

PR #2638, opened 2018-05-31
Make the stream catch errors in the query

### LINKED ISSUE #2109: sql error in streams end in unhandled rejection
When there is an error in a sql query and we return the knex object in order to call *.stream()* on it, we get the following error :
```
Unhandled rejection Error: Expected 2 bindings, saw 1
    at replaceRawArrBindings (/data/node_modules/knex/lib/raw.js:181:11)
    at Raw.toSQL (/data/node_modules/knex/lib/raw.js:123:13)
    at Formatter.unwrapRaw (/data/node_modules/knex/lib/formatter.js:91:21)
    at QueryCompiler_MySQL.whereRaw (/data/node_modules/knex/lib/query/compiler.js:541:54)
    at QueryCompiler_MySQL.where (/data/node_modules/knex/lib/query/compiler.js:314:32)
    at /data/node_modules/knex/lib/query/compiler.js:147:30
    at Array.map (native)
    at QueryCompiler_MySQL.select (/data/node_modules/knex/lib/query/compiler.js:146:33)
    at QueryCompiler_MySQL.toSQL (/data/node_modules/knex/lib/query/compiler.js:108:27)
    at Builder.toSQL (/data/node_modules/knex/lib/query/builder.js:111:44)
    at /data/node_modules/knex/lib/runner.js:104:32
    at tryCatcher (/data/node_modules/bluebird/js/release/util.js:16:23)
    at /data/node_modules/bluebird/js/release/using.js:185:26
    at tryCatcher (/data/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/data/node_modules/bluebird/js/release/promise.js:510:31)
    at Promise._settlePromise (/data/node_modules/bluebird/js/release/promise.js:567:18)
```
This is triggered by any invalid syntax, here we do
```
const query = knex({
        client:     'mysql',
        connection: {
            host:     '...',
            user:     '...',
            password: '...',
            database: '...',
            port:     '...'
        },
    })
.select('*')
.from(SUBSCRIPTION)
.whereRaw(`DAY(${SUBSCRIPTION}.beginDate) in (?)`, ['30', '31']);
```
This is an error because of the second parameter passed to whereRaw, which should be `[['30', '31']]`.

The problem is that there seems to be no easy way to properly catch this error.

I tried passing a callback to the *.stream()* method in order to get a promise back as per the documentation ( http://knexjs.org/#Interfaces-stream ), in hope of getting a rejected promise, but I still get the stream back when doing that, not a promise, so that doesn't help.

This issue might be related to https://github.com/tgriesser/knex/issues/1852 or https://github.com/tgriesser/knex/issues/948 but I'm not really sure about that.

What I ended doing as a workaround this issue is something like the following:
```
try {
        query.toString();
    } catch (e) {
        return {
            error: new Error('There is an error in the cursor SQL syntax',
                { errorMessage: e.message }),
        };
    }
return { query };
```
But that feels rather dirty.

Is there any proper way to catch this kind of error (or any sql syntax error when in stream mode for that matter) ?

fcmatteo:
It should fix #2109 and a case when there is an error running the query.

### DIFF
--- a/src/dialects/mysql/index.js
+++ b/src/dialects/mysql/index.js
@@ -104,7 +104,14 @@ assign(Client_MySQL.prototype, {
     return new Promise((resolver, rejecter) => {
       stream.on('error', rejecter)
       stream.on('end', resolver)
-      connection.query(queryOptions, obj.bindings).stream(options).pipe(stream)
+      const queryStream = connection.query(queryOptions, obj.bindings).stream(options)
+
+      queryStream.on('error', (err) => {
+        rejecter(err);
+        stream.emit('error', err)
+      })
+
+      queryStream.pipe(stream)
     })
   },
 
--- a/src/dialects/postgres/index.js
+++ b/src/dialects/postgres/index.js
@@ -199,15 +199,15 @@ assign(Client_PG.prototype, {
   _stream(connection, obj, stream, options) {
     const PGQueryStream = process.browser ? undefined : require('pg-query-stream');
     const sql = obj.sql;
+
     return new Promise(function(resolver, rejecter) {
       const queryStream = connection.query(new PGQueryStream(sql, obj.bindings, options));
-      queryStream.on('error', function(error) { stream.emit('error', error); });
-      // 'error' is not propagated by .pipe, but it breaks the pipe
-      stream.on('error', function(error) {
-        // Ensure the queryStream is closed so the connection can be released.
-        queryStream.close();
+
+      queryStream.on('error', function(error) {
         rejecter(error);
+        stream.emit('error', error);
       });
+      
       // 'end' IS propagated by .pipe, by default
       stream.on('end', resolver);
       queryStream.pipe(stream);
--- a/src/runner.js
+++ b/src/runner.js
@@ -94,7 +94,7 @@ assign(Runner.prototype, {
         return runner.client.stream(runner.connection, sql, stream, options)
       } catch (e) {
         stream.emit('error', e)
-        throw new Error(e);
+        throw e;
       }
     })
 
--- a/src/runner.js
+++ b/src/runner.js
@@ -84,13 +84,17 @@ assign(Runner.prototype, {
     const promise = Promise.using(this.ensureConnection(), function(connection) {
       hasConnection = true;
       runner.connection = connection;
-      const sql = runner.builder.toSQL()
-      const err = new Error('The stream may only be used with a single query statement.');
-      if (isArray(sql)) {
-        if (hasHandler) throw err;
-        stream.emit('error', err);
+      try {
+        const sql = runner.builder.toSQL()
+
+        if (isArray(sql) && hasHandler) {
+          throw new Error('The stream may only be used with a single query statement.');
+        }
+
+        return runner.client.stream(runner.connection, sql, stream, options)
+      } catch (e) {
+        stream.emit('error', e)
       }
-      return runner.client.stream(runner.connection, sql, stream, options);
     })
 
     // If a function is passed to handle the stream, send the stream
--- a/src/runner.js
+++ b/src/runner.js
@@ -94,6 +94,7 @@ assign(Runner.prototype, {
         return runner.client.stream(runner.connection, sql, stream, options)
       } catch (e) {
         stream.emit('error', e)
+        throw new Error(e);
       }
     })
 
--- a/test/integration/builder/selects.js
+++ b/test/integration/builder/selects.js
@@ -267,10 +267,6 @@ module.exports = function(knex) {
     });
 
     it('emits error on the stream, if not passed a function, and query fails', function(done) {
-      setTimeout(() => {
-        done(new Error('Timeout'));
-      }, 5000)
-
       const stream = knex('accounts').select('invalid_field').stream()
       stream.on('error', function(err) {
         assert(err.code === 'ER_BAD_FIELD_ERROR')
@@ -279,12 +275,7 @@ module.exports = function(knex) {
     })
 
     it('emits error if not passed a function and the query has wrong bindings', function(done) {
-      setTimeout(() => {
-        done(new Error('Timeout'));
-      }, 5000)
-
       const stream = knex('accounts').whereRaw('id = ? and first_name = ?', ['2']).stream()
-
       stream.on('error', function(err) {
         assert(err instanceof Error)
         done()
--- a/test/integration/builder/selects.js
+++ b/test/integration/builder/selects.js
@@ -269,7 +269,7 @@ module.exports = function(knex) {
     it('emits error on the stream, if not passed a function, and query fails', function(done) {
       const stream = knex('accounts').select('invalid_field').stream()
       stream.on('error', function(err) {
-        assert(err.code === 'ER_BAD_FIELD_ERROR')
+        assert(err instanceof Error)
         done()
       })
     })
--- a/test/integration/builder/selects.js
+++ b/test/integration/builder/selects.js
@@ -1,10 +1,10 @@
 /*global describe, expect, it, testPromise, d*/
 'use strict';
 
-var _ = require('lodash')
-var assert  = require('assert')
-var Promise = testPromise;
-var Runner = require('../../../lib/runner');
+const _ = require('lodash')
+const assert  = require('assert')
+const Promise = testPromise;
+const Runner = require('../../../lib/runner');
 
 module.exports = function(knex) {
 
@@ -212,7 +212,7 @@ module.exports = function(knex) {
       return knex('accounts')
         .options({
           typeCast (field, next) {
-            var val
+            let val
             if (field.type === 'VAR_STRING') {
               val = field.string()
               return val == null ? val : val.toUpperCase()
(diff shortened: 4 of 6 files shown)


test/integration/builder/selects.js
```diff
@@ -266,6 +266,31 @@ module.exports = function(knex) {
...
+
+      const stream = knex('accounts').select('invalid_field').stream()
+      stream.on('error', function(err) {
+        assert(err.code === 'ER_BAD_FIELD_ERROR')
+        done()
+      })
+    })
+
+    it('emits error if not passed a function and the query has wrong bindings', function(done) {
+      setTimeout(() => {
```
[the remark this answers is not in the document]
fcmatteo: Right, fixing that.

src/dialects/mysql/index.js
```diff
@@ -104,7 +104,14 @@ assign(Client_MySQL.prototype, {
     return new Promise((resolver, rejecter) => {
       stream.on('error', rejecter)
       stream.on('end', resolver)
-      connection.query(queryOptions, obj.bindings).stream(options).pipe(stream)
+      const queryStream = connection.query(queryOptions, obj.bindings).stream(options)
+
+      queryStream.on('error', (err) => {
```
kibertoad: Is it only relevant for MySQL?

fcmatteo: I'm not sure, I could reproduce the bug only in MySQL because that's the database I am using.

kibertoad:
@fcmatteo Tests are failing on Mariadb, I presume same fix should be applied to it or exclusion added (since I think there were plans to remove maria dialect altogether).

kibertoad:
@fcmatteo Almost there, two failing test on postgres remain :)

fcmatteo:
@kibertoad It's weird. In my machine, all the tests using postgres (10.4) passes.

elhigu:
@fcmatteo have you ran npm run build; before running the tests? You can try with postgres 9.6 too.

src/runner.js
```diff
@@ -84,13 +84,17 @@ assign(Runner.prototype, {
     const promise = Promise.using(this.ensureConnection(), function(connection) {
       hasConnection = true;
       runner.connection = connection;
-      const sql = runner.builder.toSQL()
-      const err = new Error('The stream may only be used with a single query statement.');
-      if (isArray(sql)) {
-        if (hasHandler) throw err;
-        stream.emit('error', err);
+      try {
```
elhigu: This change needs to be checked carefully...

fcmatteo:
@elhigu Yes, I built before running the tests. Now I've tried with PostgreSQL 9.6.9 and passed too. 😕

kibertoad:
@fcmatteo How are you running tests? Using npm?

fcmatteo:
@kibertoad After building with `npm run babel` I run `DB=postgres npm run test`.

kibertoad:
@fcmatteo It does pass locally - I wonder if issue is to CI server being slower to developer workstation or different order of test execution. Looking at the stracktrace, what seems to be happening is that in this block

```
  con.on('parseComplete', function (msg) {
    if (self.activeQuery.name) {
```

self.activeQuery is null. One of the possible causes for this is for 

```
  const connectedErrorHandler = (err) => {
    if (this.activeQuery) {
      var activeQuery = self.activeQuery
      this.activeQuery = null
      return activeQuery.handleError(err, con)
    }
    this.emit('error', err)
  }
```

to get triggered. Hence my suspicion is that connection that has errored out in one of the tests doesn't get this processed before another test begins, effectively resetting activeQuery in the middle of the execution.

kibertoad:
btw, probable #2601 could also be covered? I wonder why it doesn't cause tests to explode for Oracle.

fcmatteo:
I'm trying some things to debug but as I can't reproduce I need to commit. Sorry!

fcmatteo:
@kibertoad OK it seems to be what you say. I destroyed the connection manually and the tests passed. Now we have to look for a real solution to the problem.
BTW, it seems to be a problem that existed before this PR but as there were no test cases where the query fails using streams, it was unnoticed.

kibertoad:
@fcmatteo Now linting fails :)

```
home/travis/build/tgriesser/knex/src/dialects/postgres/index.js
  202:11  error  'client' is assigned a value but never used  no-unused-vars
```

fcmatteo:
It seems that I fixed it but now we it's failing with Oracle in Node 6 and 8. I'll check that later.

kibertoad:
@fcmatteo Yeah, that's to be expected. I presume that #2601 finally kicked in :)

kibertoad:
@fcmatteo Woohoo, it passes! If you gave my write permissions on your fork, I could address any later comments in your absence.

src/runner.js
```diff
@@ -84,13 +84,17 @@ assign(Runner.prototype, {
...
+        const sql = runner.builder.toSQL()
+
+        if (isArray(sql) && hasHandler) {
+          throw new Error('The stream may only be used with a single query statement.');
+        }
+
+        return runner.client.stream(runner.connection, sql, stream, options)
+      } catch (e) {
```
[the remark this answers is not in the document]
fcmatteo: It's the only way I found to emit the error to the stream. We could throw the error again in the `catch`, after emitting to the stream...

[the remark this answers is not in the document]
kibertoad: We probably should rethrow it to preserve the original logic, unless there is a good reason not to interrupt execution here.

fcmatteo:
There's a problem with this test `should remove the record in the lock table once finished`. Any idea?

kibertoad:
@fcmatteo Is it another case of tests leaking? Does the test pass on its own?

kibertoad:
@fcmatteo Given that only node 8 is failing, I wonder if it's just a flaky test. Can you try rerunning tests by closing and reopening PR?

kibertoad:
@fcmatteo Yup, looks like it was a flaky test after all.

src/runner.js
kibertoad: @elhigu The only difference I see now is that it will now emit an error even if _hasHandler_ is truthy. Would you consider that an unwelcome change?

fcmatteo:
I'm back from vacations. Any chance we could merge the PR?

src/runner.js
```diff
@@ -84,13 +84,18 @@ assign(Runner.prototype, {
...
+        stream.emit('error', e)
+        throw new Error(e);
```
elhigu: Why the error is wrapped to new Error before throwing?

elhigu: probably re-throwing e is more compatible backwards + its easier for client to know which kind of db error actually was thrown

src/runner.js
elhigu: it was just a note for my futureself who has time to check this out better.

elhigu:
Thank you!

src/dialects/postgres/index.js
```diff
@@ -199,15 +199,15 @@ assign(Client_PG.prototype, {
...
     const PGQueryStream = process.browser ? undefined : require('pg-query-stream');
     const sql = obj.sql;
+
     return new Promise(function(resolver, rejecter) {
       const queryStream = connection.query(new PGQueryStream(sql, obj.bindings, options));
-      queryStream.on('error', function(error) { stream.emit('error', error); });
-      // 'error' is not propagated by .pipe, but it breaks the pipe
-      stream.on('error', function(error) {
-        // Ensure the queryStream is closed so the connection can be released.
-        queryStream.close();
```
caseywebdev: Why was any of this changed? This query still needs to be closed on `error` or it will continue to execute on the server. See https://github.com/tgriesser/knex/pull/1935

elhigu: Good catch! @fcmatteo Any idea why queryStream.close() was dropped?

kibertoad: Ouch. Can we get a test for this?..

elhigu: Test case for original fix would have catch this that is the reason I'm so insistent that all the fixes has test included. So when this is fixed again, a test would be great to have.

elhigu: @kibertoad you were faster 👍

kibertoad: There are tests in the original PR, actually, I wonder if this particular aspect can be properly tested.

elhigu: Maybe by mocking queryStream.close method or by checking that stream.on('error'... event handler is set.

elhigu: Or with integration test which queries for open connections from database.

fcmatteo: I don't remember. Following the thread of the PR, there were two failing tests in PostgreSQL that apparently were fixed by that.
But testing on local with the previous chunk of code seems to work too.

elhigu: Ok, better to just put it back then and add new test

kibertoad: Followed up with https://github.com/tgriesser/knex/pull/2773

PR #2640, opened 2018-05-31
Remove readable-stream and safe-buffer

### LINKED ISSUE #2639: Remove readable-stream and safe-buffer
Given that we've dropped the Node 4 support, we can start using native Buffer and Readable implementations from Node.

kibertoad:
fixes #2639


package.json
```diff
@@ -15,12 +15,10 @@
     "inherits": "~2.0.3",
     "interpret": "^1.1.0",
     "liftoff": "2.5.0",
-    "lodash": "^4.17.5",
+    "lodash": "^4.17.10",
```
kibertoad: Since we don't commit package-lock, this actually doesn't change anything in proper CI, hence this is too small of a change to warrant separate PR, but this is still useful to force locally checked out sources to update to Lodash with better Node 10 support.

src/runner.js
```diff
@@ -75,7 +75,7 @@ assign(Runner.prototype, {
     const hasHandler = typeof handler === 'function';
 
     // Lazy-load the "PassThrough" dependency.
-    PassThrough = PassThrough || require('readable-stream').PassThrough;
+    PassThrough = PassThrough || require('stream').PassThrough;
```
kibertoad: Do we still need lazy loading?

kibertoad: Then probably makes sense to keep it to be on a safe side.

elhigu:
thanks you very much!

Issue #2630, opened 2018-05-25
returning and wrapIdentifier don't seem to work well together


koskimas:
# Environment

Knex version: 0.14.6
Database + version: postgresql 10
OS: arch linux

# Bug

Hi,

I believe `wrapIdentifier` and/or `postProcessResponse` hooks are not called correctly for identifiers passed to `returning` function on postgresql. Here's a piece of code that reproduces the problem:

```js
const knex = require('knex')({
  client: 'postgres',

  connection: {
    host: '127.0.0.1',
    database: 'test'
  },

  wrapIdentifier: (id, oldImpl) => {
    return oldImpl(id + '_foo');
  },

  postProcessResponse: (result) => {
    // This is never called with the return value of the insert query.
    return result;
  }
});

const main = async () => {
  await knex.schema.dropTableIfExists('t');
  await knex.schema.createTable('t', table => {
    table.increments('id');
    table.string('value');
  });

  const retVal = await knex('t')
    .insert({ value: 'hello' })
    .returning('value')
    .debug();

  console.log(`should return ['hello'] but returns`, retVal);
  await knex.destroy();
};

main().catch(console.error);
```

In the example, `retVal` should be `['hello']` (right?) but it is `[undefined]` instead. `wrapIdentifier` seems to be called correctly for the binding, but the `returning` property is not mapped and a wrong column is picked from the return value (maybe, just quessing here).

Or am I missing something obvious?

elhigu:
Looks like wrapIdentifier is called correclty:

https://runkit.com/embed/9i57f0h6k4te

```js
const knex = require('knex')({
  client: 'postgres',

  wrapIdentifier: (id, oldImpl) => {
    return oldImpl(id + '_foo');
  },
});

knex('t')
  .insert({ value: 'hello' })
  .returning('value').toSQL().sql;

// outputs: insert into "t_foo" ("value_foo") values (?) returning "value_foo"
```
So it may be that postProcess is not called in this case... sounds strange. Btw. shouldn't that insert returning return `[{value_foo: 'hello'}]` instead of `['hello']`?

koskimas:
Yeah, `wrapIdentifier` is called for the actual binding, but the query has the "special" `returning` property shown in the `debug()` output, and it holds an unmapped value `value`. Could that be the reason?

> So it may be that postProcess is not called in this case

`postProcessResponse` is indeed not called.

linkenneth:
[...]

PR #2642, opened 2018-06-02
fixes 2630

koskimas:
This PR fixes the issue #2630 for postgres

### DIFF
--- a/src/dialects/postgres/index.js
+++ b/src/dialects/postgres/index.js
@@ -246,7 +246,8 @@ assign(Client_PG.prototype, {
         if (returning === '*' || Array.isArray(returning)) {
           returns[i] = row;
         } else {
-          returns[i] = row[returning];
+          // Pluck the only column in the row.
+          returns[i] = row[Object.keys(row)[0]];
         }
       }
       return returns;
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -112,6 +112,61 @@ module.exports = function(knex) {
 
     });
 
+    // TODO: This doesn't work on oracle yet.
+    if (['postgresql', 'mssql'].includes(knex.client.dialect)) {
+      describe('returning with wrapIdentifier and postProcessResponse`', () => {
+        let origHooks = {};
+
+        before('setup custom hooks', () => {
+          origHooks.postProcessResponse = knex.client.config.postProcessResponse;
+          origHooks.wrapIdentifier = knex.client.config.wrapIdentifier;
+
+          // Add `_foo` to each identifier.
+          knex.client.config.postProcessResponse = (res) => {
+            if (Array.isArray(res)) {
+              return res.map(it => {
+                if (typeof it === 'object') {
+                  return _.mapKeys(it, (value, key) => {
+                    return key + '_foo';
+                  });
+                } else {
+                  return it;
+                }
+              })
+            } else {
+              return res;
+            }
+          };
+
+          // Remove `_foo` from the end of each identifier.
+          knex.client.config.wrapIdentifier = (id) => {
+            return id.substring(0, id.length - 4);
+          };
+        });
+
+        after('restore hooks', () => {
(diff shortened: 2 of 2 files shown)


src/dialects/postgres/index.js
```diff
@@ -246,7 +246,8 @@ assign(Client_PG.prototype, {
         if (returning === '*' || Array.isArray(returning)) {
           returns[i] = row;
         } else {
-          returns[i] = row[returning];
+          // Pluck the only column in the row.
+          returns[i] = row[Object.keys(row)[0]];
```
koskimas: All other dialects that support `returning` already do it like this.

kibertoad:
@koskimas Oracle seems to be failing.

koskimas:
It would appear that the `accounts` table is only created for postgres :smile:

koskimas:
No that's not it...

koskimas:
The oracle query compiler seems to be borked and it has nothing to do with this PR. I'm not familiar with `oracle` but this cannot be valid SQL:

```
insert into accounts (balance) values (:1) returning balance into :2
```

There are two `into` statements. @elhigu Any ideas? This seems like a separate issue since I only changed postgres dialect. Should I just disable my test for oracle?

kibertoad:
@koskimas What problem do you see with this SQL? Two "into" per se are not invalid.

From what I gather, correct syntax is like this:

```
var tmp number;
BEGIN
  insert into users(name,sur) 
  values('Jane','Doe')
  returning id into :tmp;
END;
```

see https://stackoverflow.com/questions/39797986/what-exactly-does-returning-into-do-in-oracle-sql
You return into local variable in Oracle world.

koskimas:
Ok.. I have no idea what's going on with oracle here, and since I don't have oracle installed on my machine, I have no way to debug it. The last time I tried to install oracle on my arch linux box, I gave up after couple of hours. So if someone else doesn't look into this, I'll have to cancel this PR.

kibertoad:
@koskimas Since you haven't changed anything about Oracle, can't you add ToDo comment into your test and just not run the test on it?

koskimas:
I can do that. I don't know if anyone will merge a partial fix, but let's see :smile:

kibertoad:
Given that PostgreSQL userbase in JS world is much larger than the Oracle one, I would think that rejecting an improvement on such grounds would be insane.
If someone from Oracle world every encounters this issue, he would be definitely much more well-prepared to tackle it himself and improve on this :).

koskimas:
@elhigu What do you think about merging this?

test/integration/builder/additional.js
```diff
@@ -112,6 +112,61 @@ module.exports = function(knex) {
...
+                } else {
+                  return it;
+                }
+              })
+            } else {
+              return res;
+            }
+          };
+
+          // Remove `_foo` from the end of each identifier.
```
elhigu: Symmetric transformation which removes and then re-creates suffix is not ideal to do here, since someone might just break calling wrap / postprocess altogether and test wouldn't know difference or someone may accidentally remove place where handlers are overwritten and still test runs fine. 

Also I had to went to look into handler code after checking the test code and wondering why keys weren't transformed to anything :)

elhigu:
thanks!

test/integration/builder/additional.js
koskimas: Shouldn't there be another test that tests whether the functions are called at all? This test only test if the functions are called correctly provided that the feature works.

koskimas:
Thanks @elhigu. Any plans for the timeline of the next release?

elhigu:
This month. I'm currently fixing mssql + adding it to CI. There were some breaking changes made so I want to get that to 0.15. too. Few tests are still failing so it shouldn't take too long to finish.

elhigu:
@koskimas btw. there are npm scripts nowadays in package.json to be able to run oracle tests locally in docker.

Issue #2644, opened 2018-06-05
Undefined max_batch when using knexSnakeCaseWrappers (Objection)


heisian:
# Bug
When running `knex` migrations configured with Objection's `knexSnakeCaseWrappers`, the promise handler for the `_latestBatchNumber` query **always** returns a default value of `0`:
```node
// knex/src/migrate/index.js, lines 329-332

_latestBatchNumber(trx = this.knex) {
    return trx.from(getTableName(this.config.tableName, this.config.schemaName))
      .max('batch as max_batch').then(obj => obj[0].max_batch || 0);
  }

```

This is because `max_batch` does not exist in `obj`. The actual structure of `obj`, due to `knexSnakeCaseWrappers` is:
```node
obj = {
  maxBatch: 2
}
```

Defaulting to `0` results in a batch number of `1` always being inserted into the `knex_migrations` table when using `knexSnakeCaseWrappers`.

In effect, this makes all of your migrations only use a single batch, which is potentially **_catastrophic_** when performing a rollback.

# Environment

Knex version: 0.14.6
Database + version: Postgres 10.1
OS: macOS/Linux

Select applicable template from below.
If issue is about oracledb support tag @ atiertant. For MSSql tag @ smorey2 .
Rest of dialects doesn't need tags.

koskimas:
@heisian This is a knex issue. The `postProcessResponse` doesn't receive information whether the query is an "internal" one or not. objection's `posProcessResponse` always maps the response and there is nothing we could do here. Knex should probably not use `wrapIdentifier` and `postProcessResponse` in these kind of internal queries.

heisian:
OK, sounds like a slightly deep-rooted issue.. will think this over - am using my own patched branch for the moment.

Would it be prudent to warn about using `knexSnakeCaseWrappers` in Objection, since the use of such will cause only a single batch to be created and thereby roll back the entire db if a rollback is performed?

koskimas:
We should probably add a warning about using `knexSnakeCaseWrappers` in **migrations**. You can safely use it in the actual code.

heisian:
Ok. Do you think it sounds feasible to adjust Knex to use `wrapIdentifier` when running migrations, i.e., `knex.withSchema('mySchema').table('myTable', (t) => { t.increments('myId') }) but **not** when interacting with the metatable `knex_migrations`? Will be looking into it..

kibertoad:
[...]

PR #2645, opened 2018-06-05
Migrations - use postProcessResponse on max_batch query

heisian:
The motivation for this PR is outlined in #2644.

Currently pending tests, however any reviewing parties please advise as to the viability of the proposed solution.

thanks,


elhigu:
Looks feasible, but still one could write some other kind of postprocessing function which breaks this too. 

Probably preventing use of wrapIndentifier and postprocess response in queries made internally by knex like @koskimas mentioned in https://github.com/tgriesser/knex/issues/2644#issuecomment-394596480 is way to go.

heisian:
Okay thanks

heisian:
Not really sure how to approach this.. what I think is ideal is using a cloned instance of the `knex` object passed into `Migrator` without `wrapIdentifier` or `postProcessResponse` in its `client.config` obj.

However, I can't find an easy way to clone the `knex` object within the `Migrator` class.. an error gets thrown when attempting to do so:
```
$ knex --cwd ./ --knexfile ./db/config/knexfile.js migrate:currentVersion
Cannot find module 'pg-native'
/Users/heisian/Web/knex/lib/migrate/index.js:206
    return getSchemaBuilder(trx, schemaName).hasTable(tableName).then(function (exists) {
                                            ^

TypeError: Cannot read property 'hasTable' of undefined
    at Migrator._ensureTable (/Users/heisian/Web/knex/lib/migrate/index.js:206:45)
```
this is the result of doing..
```node
this.knexInternal = _.cloneDeep(knex)
```
and using `knexInternal` to run queries on the metadata tables.

koskimas:
@heisian `cloneDeep` traverses and attempts to clone getters and all kinds of internal class instances. The result would never work, and if it did, it would be a huge hack.

What I'd like to see here is a way to override some configs (like `wrapIdentifer` and `postProcessResponse`) per query. Something like:

```js
knex('foo')
  .config({ wrapIdentifier: customWrapIdentifier })
  .where( ... )
```

I think this would have some other useful use cases. For example in objection, I'd like to be able to configure snake_case to camelCase conversion on knex level *per model class* and not globally. This would make it possible. @elhigu Thoughts?

heisian:
Mm yeah that's actually what I wanted to ask first, if there was a way to override per-query just as you say. That functionality would be great.

elhigu:
@koskimas I like that idea of being able to modify some parts of configuration for individual queries and it would also resolve this problem nicely.

wubzz:
Despite the fact that new functions are documented in the documentation with its very own description, I still think it would be a bit odd to have `.config({..})` alongside existing `.options({...})`.

I realize `.options` is currently dialect-specific as specified by the docs, but maybe it doesn't have to be..

heisian:
@wubzz good point.. I haven't had to use `.options` yet - is it feasible to just append `wrapIdentifier` to available `.options` then?

elhigu:
@wubzz .options() was maybe too vague name for that method allowing to pass DB driver specific parameters to certain query... I wouldn't like to overload that to do 2 separate things. Maybe we can name this one better, which is modifying knex config parts for certain query so that there will not be confusion between those? I'm not sure.

heisian:
what about a second argument for `knex('table', { wrapIdentifier })`, or `knex.clone({ wrapIdentifier })`?  ¯\\\_(ツ)_/¯

wubzz:
Perhaps `.queryConfig({...})` for the sake of consistency. That would be consistent with ie `.queryContext({...})`. At the same time that feels very weird considering these settings are on client level not builder level..

Another solution would be to append these functions to the build chain itself, kind of like how events are supported both on the Client and on the Builder.

```javascript
knex('table')
.wrapIdentifier((value, original, queryContext) => original(value))
.postProcessResponse((result, queryContext) => result)
.select(['a'])
```

I'm not sure either. Just throwing ideas out there. I'm ok with whatever.

elhigu:
Since that method is actually patching / merging config, maybe `.patchConfig` or `.mergeConfig` or `.patchClientConfig` would be even more appropriate. 

Also there could be validation that connection attribute cannot be changed for now, because it would require reinitialization of the knex client which might affect globally to every client created from the same knex instance (unless this will also clone the instance, which would mean separate pools, which might cause other problems, because it would cause max connections to be bigger as planned originally when knex was initialized)

heisian:
Yes, I was thinking another client, but I did not consider your point about higher max connections than expected. I do like `.patchConfig`/`.mergeConfig`. Maybe `mergeConfig` since Objection.js uses `mergeContext` and `mergeEager`, etc.?

heisian:
@elhigu How averse are we to just renaming `max_batch` to `max`? There is no difference in functionality whatsoever, and case converters are not going to break it. IMO this is an incredibly low path to fixing something that would, to do it 'properly', require refactoring much, much, more of the library as suggested above.

thanks

bin/cli.js
```diff
@@ -241,7 +241,7 @@ function invoke(env) {
 }
 
 var cli = new Liftoff({
-  name: 'knex',
+  name: '@heisian/knex',
```
elhigu: Why this ?

elhigu:
I really don't understand how current code changes in this PR are related to discussion we have had earlier about being able to provide different postProcess method for certain queries.

heisian:
The discussion strayed from the original problem:
1) I am using `knexSnakeCaseMappers` in app code
2) I am also using `knexSnakeCaseMappers` in migrations
3) `wrapIdentifier` gets called for all user-invoked query-building methods and migrations.
4) `wrapIdentifier` does NOT get called for knex's internal migration queries.
5) This causes `max_batch` to not get mapped to `maxBatch`, resulting in Knex defaulting to a batch number of `0` for *every* migration.
6) This has the catastrophic effect of rolling back the **entire** database on any rollback, regardless of how many batches the user **thinks** may have run previously.

This code change solves the problem without having to do all the intense refactoring we discussed earlier, simply by making the `max_batch` alias into `max`, which has no consequence outside of the scope of finding the latest batch number.

See also: https://github.com/Vincit/objection.js/pull/949/files

elhigu:
Sounds kind of reasonable, even that it really doesn't resolve the problem of being able to ignore those hooks some times. 

If this issue is fixed by changing max_batch to maxbatch, then new feature request should be at least added with initial design that was discussed here already.

heisian:
Ok, we'll do some investigation. @ekeric13

kibertoad:
Fixed in 0.16.x

PR #2647, opened 2018-06-05
Remove WebSQL dialect

kibertoad:


kibertoad:
@elhigu Is anything missing?

elhigu:
Was there documentation changes PR yet?

kibertoad:
Don't think so, will try to make one when I get an access to computer in non-office env

PR #2648, opened 2018-06-06
Option to disable validation of migrations

melnikaite:
This option is very helpful for deploying different branches to the same environment or for developers sho use multiple branches during the day.

### DIFF
--- a/test/unit/migrate/migrator.js
+++ b/test/unit/migrate/migrator.js
@@ -5,19 +5,20 @@
 var mockFs = require('mock-fs');
 var knex = require('../../../knex');
 
+var config = {
+  client: 'pg',
+  connection: {
+    user: 'postgres',
+    password: '',
+    host: '127.0.0.1',
+    database: 'knex_test'
+  },
+  migrations: {
+    directory: 'test/integration/migrate/migration'
+  }
+};
+
 describe('Migrator.loadExtensions', function () {
-  var config = {
-    client: 'pg',
-    connection: {
-      user: 'postgres',
-      password: '',
-      host: '127.0.0.1',
-      database: 'knex_test'
-    },
-    migrations: {
-      directory: 'test/integration/migrate/migration'
-    }
-  };
   var migrator;
 
   before(function() {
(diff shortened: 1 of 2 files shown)


test/unit/migrate/migrator.js
```diff
@@ -71,3 +71,34 @@ describe('Migrator.loadExtensions', function () {
   });
 
 });
+
+describe('Migrator.validateMigrationList', function () {
+  var config = {
```
kibertoad: since this is identical to the config in another test, maybe that one should be extracted higher and reused?

test/unit/migrate/migrator.js
```diff
@@ -71,3 +71,34 @@ describe('Migrator.loadExtensions', function () {
...
+      database: 'knex_test'
+    }
+  };
+
+  it('should throw error if migrations list is invalid', function () {
+    var migrator = knex(config).migrate;
+    try {
+      migrator.validateMigrationList([[], ['absent']]);
+    } catch (err) {
+      return expect(err).to.be.an.instanceof(Error);
```
kibertoad: this test will not fail if there is no error thrown

elhigu:
So this feature will make it easier to have incompatible database and actual code after changing branch and running migrations from other branch (migrations of other branch stays but code is reverted when changing branch).

If you are doing development like that maybe correct solution would be to rollback before changing the branch so that the DB will be in correct state.

I'm not convinced at all that this feature should be in knex. I would like to see more discussion (from more people) if this is really wanted or not.

melnikaite:
@elhigu, rollback works only for branches I'm working on, but in some cases, e.g. CI, deploy script doesn't know if it should run rollback and how many times, also we don't want to loose some data if rollback contains `dropColumn`. Of course I understand that this option can make code in branch incompatible, that is why I'm going to use this only for test environment.

elhigu:
In CI I would suggest always loading basic data from some SQL dump and then run migrations on top of that. In that way you'll always have stable and correct data in DB.

melnikaite:
@elhigu, yeah, but currently we don't have any seeds

elhigu:
@melnikaite well you clearly have some data in your CI db server, so you can take dump from that and load it to DB before running `knex migrate:latest; npm test` (or what ever you do for testing).

idrabenia:
Hi guys,

This is useful feature. Because sometimes we need to downgrade of application to previous version. But your library does not allow this. It will throw errors and do not provide options for downgrade the app.

I used multiple libraries like liquibase, RoR Active Record, flyway, sequelize. No one does not block downgrades.

Please add this options.

Thanks!
\- Ilya

elhigu:
@idrabenia You re wrong. I'll repeat what I said in https://github.com/tgriesser/knex/pull/2648#issuecomment-395234350

Knex allows downgrading DB. Migration down scripts are for that.

You just do it properly by running migration downgrade script before downgrading the code. If you downgrade code first, you will lose the DB downgrade script so knex will not know how to downgrade DB. Knex is not doing any ORM reflection of DB and models do it cannot do that automatically (and it is bad and dangrous practise to do that automatically anyways).

If you are just downgrading code without downgrading DB your database is not in consistent state with your code and code might do wrong things etc. So knex prevents you from running any more migrations if DB is in inconsistent state.

If you have left your DB to reach inconsistent state you can downgrade it by running the downgrade for the DB with migration system (for example remotely from code branch where correct downfgrade scripts exist) or by restoring DB backup from state where migrations had not been ran yet (safest).

idrabenia:
@elhigu 

"You just do it properly by running migration downgrade "
Are you run anytime on you production code that drop columns? This is user data need to save.

Your validation is wrong. 

See how professionals from Java cominity implemented migration - https://www.liquibase.org

idrabenia:
@tgriesser What do you think?

idrabenia:
"If you have left your DB to reach inconsistent state you can downgrade it by running the downgrade for the DB with migration system (for example remotely from code branch where correct downfgrade scripts exist) or by restoring DB backup from state where migrations had not been ran yet (safest)."

Why with other libraries I may downgrade my DB without unneeded validation, but here it is not possible?

PR #2663, opened 2018-06-18
Allow PG timestamps without timezone

stavros-wb:
The `timestamps` method always translates to timestamp with timezone for PG and derivatives (redshift). If the first argument is set to true, then use timestamps without timezone.

### DIFF
--- a/src/schema/tablebuilder.js
+++ b/src/schema/tablebuilder.js
@@ -168,9 +168,10 @@ each(columnTypes, function(type) {
 
 // The "timestamps" call is really just sets the `created_at` and `updated_at` columns.
 TableBuilder.prototype.timestamps = function timestamps() {
-  const method = (arguments[0] === true) ? 'timestamp' : 'datetime';
-  const createdAt = this[method]('created_at');
-  const updatedAt = this[method]('updated_at');
+  const isTimestamp = arguments[0] === true;
+  const method = isTimestamp ? 'timestamp' : 'datetime';
+  const createdAt = this[method]('created_at', isTimestamp);
+  const updatedAt = this[method]('updated_at', isTimestamp);
   if (arguments[1] === true) {
     const now = this.client.raw('CURRENT_TIMESTAMP');
     createdAt.notNullable().defaultTo(now);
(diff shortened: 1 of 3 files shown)


src/schema/tablebuilder.js
```diff
@@ -168,9 +168,10 @@ each(columnTypes, function(type) {
 
 // The "timestamps" call is really just sets the `created_at` and `updated_at` columns.
 TableBuilder.prototype.timestamps = function timestamps() {
-  const method = (arguments[0] === true) ? 'timestamp' : 'datetime';
-  const createdAt = this[method]('created_at');
-  const updatedAt = this[method]('updated_at');
+  const isTimestamp = arguments[0] === true;
```
elhigu: true / false as a first parameter is already used for selecting between datetime and timestamp type. It cannot be overloaded to mean between timestamp/timestampz for some dialects. Also we should never use this kind of true / false parameters anyways, but some options parameter object. There was some issue or PR where that functionality was actually specified, so to be able to finish this PR that issue should be found first to check out plans, how this should be implemented.

stavros-wb: The main issue is that there's no way to declare a column as timestamp without TZ. from what I see, it should do this when passing `false` as the first argument in either `{datetime, timestamp}` function. Isn't this what the method `timestamps` should do? The docs explicitly state that passing `true` as the first argument would result in a `timestamp` not a `timestamp with timezone`.

>Adds a created_at and updated_at column on the database, setting these each to dateTime types. When true is passed as the first argument a timestamp type is used. Both colums default to being not null and the current timestamp when true is passed as the second argument.

elhigu: Docs says timestamp as in timestamp generally in SQL it is not talking specifically about the postgres timestamp type. It is intentional that it generates timestamp with timezone in postgresql. There is also issue about that... it was fixed maybe 2-3 years ago.

stavros-wb: is this the one https://github.com/tgriesser/knex/issues/184 ?

elhigu: IIRC biggest problem with plain timestamp type is that it didn't take into account timezone info part when one tires to store ISO8601 strings to postgres timetamp columns.

```sql
foo=> SELECT '2011-01-01 00:00:00+03'::TIMESTAMP;
      timestamp      
---------------------
 2011-01-01 00:00:00
(1 row)
```

To be honest I don't know why anyone would even like to use timestamp without timezone type with postgres for anything.

elhigu:
can be reopened if wanted / PR fixed

PR #2666, opened 2018-06-20
Fix collation when renaming column in MySQL dialect

straub:
Ran into an issue today where `.renameColumn()` was failing because the columns had a different collation after the rename and recreating the foreign keys caused an error. The collation was being lost during the `alter change`. This change preserves the collation through the alter.

I'm not familiar enough with the knex test suite to know how to go about adding an integration test for this. If someone can provide some pointers, I'd appreciate it. Didn't want to let that stop me raising the issue and providing this patch.

Thanks!


kibertoad (approved):

kibertoad:
Released in 1.0.2

PR #2673, opened 2018-06-28
Added toWhereClauseSQL() & toWhereClauseString()

### LINKED ISSUE #2526: Ability to generate "just" an expression, without a select

I would like to be able to generate only an expression with knex. I'm using [join-monster](https://www.npmjs.com/package/join-monster) for my GraphQL-postgres API, and in order to filter my data, I need to supply it with a WHERE-clause, and only a WHERE-clause ([see here for an example])(https://join-monster.readthedocs.io/en/latest/where/).

Now, generating the clause can be somewhat complex, because I may have to adjust it depending on the parameters supplied by the client. It would be super nice if I could harness knex for that - join-monster actually already recommends knex for query execution, so I have it in my project already.

However, as far as I can see, there is no way of creating **only** an expression with knex. I can only generate the clause as part of a select. Example:

    console.log(
      knex.queryBuilder()
      .where('a', 1)
      .andWhere('b', '>=', 77)
      .toString()
    );
    // -> "select * where \"a\" = 1 and \"b\" >= 77"
    //                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
    //                        what I actually need

My current ugly workaround is to match out the expression I need:

    console.log(
      knex.queryBuilder()
      .where('a', 1)
      .andWhere('b', '>=', 77)
      .toString()
      .match(/where (.*)?/)[1]
    );
    // -> "\"a\" = 1 and \"b\" >= 77"

Which gives me what I need. But that seems like a very bad approach to me.

Is there a better way to get **just** the expression I need here?

wubzz:
Fixes #2526.

I'll add docs once it's decided if this should be merged or not.

### DIFF
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -98,6 +98,31 @@ function testquery(chain, valuesToCheck, selectedClients) {
   })
 }
 
+function testToWhereClauseSQL(chain, valuesToCheck, selectedClients) {
+  selectedClients = selectedClients || clients;
+  Object.keys(valuesToCheck).forEach(function(key) {
+    var newChain = chain.clone()
+    newChain.client = selectedClients[key]
+    var result = newChain.toWhereClauseSQL()
+    var expectedSql = valuesToCheck[key].sql;
+    var expectedBindings = valuesToCheck[key].bindings;
+    expect(result.sql).to.equal(expectedSql)
+    expect(result.bindings.length).to.equal(expectedBindings.length);
+    expect(result.bindings).to.deep.equal(expectedBindings);
+  })
+}
+
+function testToWhereClauseString(chain, valuesToCheck, selectedClients) {
+  selectedClients = selectedClients || clients;
+  Object.keys(valuesToCheck).forEach(function(key) {
+    var newChain = chain.clone()
+    newChain.client = selectedClients[key]
+    var sqlString  = newChain.toWhereClauseString()
+    var checkValue = valuesToCheck[key]
+    expect(checkValue).to.equal(sqlString)
+  })
+}
+
 describe("Custom identifier wrapping", function() {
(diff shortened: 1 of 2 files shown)


elhigu:
I was thinking while back about naming of this and maybe it should be even more generic like SQL expression related. Maybe this API could be used to build any kind of expressions to pass different parts of queries and not to be limited just to where clause?

test/unit/query/builder.js
```diff
@@ -5887,4 +5912,99 @@ describe("QueryBuilder", function() {
       });
     });
   });
+
+  describe('toWhereClauseSQL()/toWhereClauseString()', function() {
+    var builder1 = qb().table('test').where('id', 1).andWhere('name', 'knex');
+    var builder2 = qb().table('test')
```
elhigu: Builder could have also joins, select and with statements etc. to show that approach works also for more complex queries.

wubzz: This is true. But join is a clause of its own, even if it does contain a where clause within itself.

This kind of depends on the outcome of your other comment:

**if feature should be where-clause only**
In this case I think it's perfectly fine to exclude join/having/on as these are seperate clauses, even if they contain where clause within itself.

**if feature should not be where-clause only**
In this case the feature should add support for remaining types, such as join. This might be a bit more complex, though. And I don't think it would be possible to contain this within a single function call, but rather split per clause type.

elhigu: I agree that removing where parts of joins are perfectly fine. 

Adding those complexities will test that those additional parts in query will not mess up fetching the `where clause` part of the main query. So it will test that this feature works with also more complex queries than just `select * from "foo" where ...`.

Also will this work when group by is used in the main query? I think that expression builder stuff should be separate from this feature after all.

PR #2675, opened 2018-06-28
Feature/knex cast

wubzz:
This is a preliminary build..

### DIFF
--- a/src/cast.js
+++ b/src/cast.js
@@ -1,7 +1,6 @@
 'use strict';
 import Raw from './raw';
 import {keys} from "lodash";
-import Client from "./client";
 
 class Cast extends Raw {
   constructor(client) {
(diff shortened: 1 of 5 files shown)


src/cast.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  as(alias) {
+    this.alias = alias;
+
+    return this;
+  }
+
+  static get TYPES() {
+    return {
```
wubzz: Unfortunately it's not quite this generic across dialects. Will need a mapping per dialect.

kibertoad: Is this the reason why PR was closed?

wubzz: Yes. Will need to rethink structure and research different dialects before-hand. I'm not keen on oracle/mssql syntax.

Will most likely end up having to create a base TYPES map on the client, and make deviations on each individual dialects' client.

PR #2681, opened 2018-07-01
Drop mariadb support 

elhigu:
- Rebased Dropped support for mariasql #2480


elhigu:
Yes

kibertoad:
@priyanka-c There is no official support for maria db anymore as separate dialect that previously existed was removed. You should use mysql driver instead, and it is supposed more or less work (to the extent that mysql driver supports mariadb).

Issue #2591, opened 2018-05-03
hasTable does not work with schema in MSSQL


Fabi2607:
# Environment

Knex version: 0.14.4
Database + version: MSSQL Express 2012 @smorey2 
OS: Windows 7

# Bug

After creating a schema "test" in MSSQL a table can be created with:
```
knex.schema.withSchema("test").createTable("tableName", ...)
```

afterwards the table exists in the database, but
```
knex.schema.withSchema("test").hasTable("tableName")
```
returns false.

The generated SQL also omits the schema name.

## Expected behaviour
`knex.schema.withSchema("test").hasTable("tableName")` returns true.

Fabi2607:
I think the following should work:

```sql
SELECT t.name 
  FROM sys.tables AS t
  INNER JOIN sys.schemas AS s
  ON t.[schema_id] = s.[schema_id]
  WHERE s.name = N'schemaName' AND t.name = N'tableName';
```
I am not really experienced in MSSQL, but this seems to cover my use case.

I did some more testing, this also affects OracleDB, which generates

```sql
select TABLE_NAME from USER_TABLES where TABLE_NAME = 'tableName'
```

 i think the proper query for OracleDB is

```sql
SELECT table_name, owner
  FROM all_tables
  WHERE owner='schemaName' AND table_name = 'tableName'
```

But this is more or less guesswork and some searching on the internet, so do not take my word for it...

Fabi2607:
Any updates on this? If the above sounds reasonable I could also create a pull request with these changes.

elhigu:
no updates, nobody has fixed / implemented this yet. PRs are welcome of course.

Fabi2607:
Our problems are solved with the merged PR, should I close it? Or do you want an issue to document the potential issues with Oracle?

elhigu:
Documentation wouldn't hurt, but that is a bit of a special case, which the person working on oracle with that like of tableownership scheme should know about. So this is good like this.

PR #2688, opened 2018-07-05
Implement missing schema support for mssql dialect

Fabi2607:
- implement schema support in tablecompiler
- add unit tests for hasTable, hasColumn and renameTable

Fixes the mssql part of #2591

### DIFF
--- a/src/dialects/mssql/schema/compiler.js
+++ b/src/dialects/mssql/schema/compiler.js
@@ -23,7 +23,7 @@ assign(SchemaCompiler_MSSQL.prototype, {
   renameTable(tableName, to) {
     this.pushQuery(
       `exec sp_rename ${this.formatter.parameter(
-        tableName
+        prefixedTableName(this.schema, tableName)
       )}, ${this.formatter.parameter(to)}`
     );
   },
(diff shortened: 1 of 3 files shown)


Fabi2607:
Any feedback on this? Should I cleanup the commit history or can it stay this way?

 I also have a small script that tests this against a database, but the integration tests do not seem to cover schema support at all, so I wasn't sure if I should implement the MSSQL specific case there...

tgriesser:
@Fabi2607 I'm not familiar enough with mssql to make a good call here, is there anything backward incompatible about this change / anything that would break the existing use?

Fabi2607:
All of this should be backwards compatible, as this should only affect calls using withSchema syntax. 
Also none of the existing tests needed to be adjusted.

The only thing that could break is, if someone worked around this issue using `withSchema('schema').hasTable('schema.tableName')` which is explicitly discouraged in the documentation.

kibertoad:
@Fabi2607 Prettier was merged, so make sure to pull changes from master and run prettier :)

Fabi2607:
Rebased on master and ran prettier.

Fabi2607:
@tgriesser any idea why the test is failing?

kibertoad:
@Fabi2607 just random flakiness. Try closing and reopen8ng pr.

PR #2691, opened 2018-07-05
chore: Update dependencies. Remove estraverse

kibertoad:
Removes estraverse as it was only needed to run tests in Node 4, which is no longer supported. Also bump dependency versions.


kibertoad:
@elhigu Any ideas why mssql might be consistently failing on Node 6?

elhigu:
Looks like the problem is that mssql server doesn't have enough time to startup. Maybe I'll need to setup longer timeout there. I've seen this happening on some other PRs as well.

kibertoad:
@elhigu Done!

kibertoad:
@elhigu Judging by https://travis-ci.org/tgriesser/knex/jobs/401093618 it looks like it's not just about not having enough time to startup, but about not starting at all :(

elhigu:
@kibertoad looks like it or could be that or it just takes sometimes huge amount of time... sometimes CI systems work really slow. It never seems to fail when I just retrigger single build env :( I need to figure why that happens (maybe print mssql logs after timeout or something like that helps)

elhigu:
Looks like the real problem is that creating testdatabase fails some times https://travis-ci.org/tgriesser/knex/jobs/401116305#L1074 should be pretty easy to fix...

elhigu:
Yeah mssql init now works in a robust manner: https://travis-ci.org/tgriesser/knex/jobs/401325827 in line  977 things would have failed earlier and knex_test database would never have been created.

PR #2692, opened 2018-07-06
Normalized and validated driverNames of test db clients and fixed oracle test setup

elhigu:
* Removed/converted tests referring to old oracle driver to be oracledb tests
* Fixed all tests to use driverName to check which driver is used for test
* Updated oracledb docker script to work again
* Updated travis configuration to run oracledb tests in separate environments with npm run oracledb:test
* Fixed mssql server database initialization to retry if first attempt fails because server not being ready


elhigu:
Looks like actually 50 more tests are failing... will fix + update travis to run oracledb test in completely test instances from others to speedup test env building.

elhigu:
Now this is ready for review

.travis.yml
```diff
@@ -10,30 +10,35 @@ cache:
...
+    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
   - node_js: "7"
-    env: DB="mssql mysql mysql2 postgres sqlite3" CXX=g++-4.8 KNEX_TEST_TIMEOUT=60000
-
-install: npm i
+    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
+  - node_js: "6"
+    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
+  - node_js: "10"
+    env: TEST_ORACLEDB=true NODE_VER=10 KNEX_TEST_TIMEOUT=60000
```
kibertoad: Why split conf for same env into two lines?

test/integration/builder/additional.js
```diff
@@ -113,7 +113,7 @@ module.exports = function(knex) {
     });
 
     // TODO: This doesn't work on oracle yet.
-    if (['postgresql', 'mssql'].includes(knex.client.dialect)) {
+    if (['pg', 'mssql'].includes(knex.client.driverName)) {
```
kibertoad: Probably we should start using enum for this :)

elhigu: Probably, but I won't be doing that in this PR :) Its a bit out of scope to change all of those to enums. Also now enums won't bring much help anymore, because most of the used string values are validated to be one of available driverNames.

.travis.yml
elhigu: To not have to run first all normal tests and then start building oracledb docker image, run source building and run all tests one more time just for oracle. Though I have been thinking of committing prebuilt base image for running oracledb tests to dockerhub so that setting up that container would be lot faster.

elhigu: aa I think now I got what you meant... true.

elhigu: I'll try it out a bit different way.

elhigu: Either of these didn't work:

```
  - node_js: "10"
    env: 
    - DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
    - TEST_ORACLEDB=true NODE_VER=10 KNEX_TEST_TIMEOUT=60000
    - TEST_ORACLEDB=true NODE_VER=8 KNEX_TEST_TIMEOUT=60000
    - TEST_ORACLEDB=true NODE_VER=6 KNEX_TEST_TIMEOUT=60000
```

```
matrix:
  include:
  - node_js: "10"
    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
    env: TEST_ORACLEDB=true NODE_VER=10 KNEX_TEST_TIMEOUT=60000
    env: TEST_ORACLEDB=true NODE_VER=8 KNEX_TEST_TIMEOUT=60000
    env: TEST_ORACLEDB=true NODE_VER=6 KNEX_TEST_TIMEOUT=60000
  - node_js: "8"
    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
  - node_js: "7"
    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
  - node_js: "6"
    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
```

elhigu: that matrix notation works a bit different way than normal travis config, but it is necessary to be able to setup different environment variables for different node versions.

kibertoad: @elhigu
What about
```
  - node_js: "10"
    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
    env: TEST_ORACLEDB=true NODE_VER=10 KNEX_TEST_TIMEOUT=60000
  - node_js: "8"
    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
    env: TEST_ORACLEDB=true NODE_VER=8 KNEX_TEST_TIMEOUT=60000
  - node_js: "7"
    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
  - node_js: "6"
    env: DB="mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
    env: TEST_ORACLEDB=true NODE_VER=6 KNEX_TEST_TIMEOUT=60000
```
?

elhigu: It won't work either, because only the last env under each section is going to be accounted for (last key overrides earlier one with the same name in yaml)

kibertoad: Ah, I see :)

elhigu: Also I think it is easier to read when all oracle envs are one after each other compared to sprinkling them under many sections.

PR #2697, opened 2018-07-09
Add Prettier

tgriesser:
Now that 0.15 is out, re-opening the attempt in #2495

@wubzz added `arrowParens: "always"` as suggested.


wubzz:
Awesome 👍

PR #2698, opened 2018-07-09
TypeScript

tgriesser:
I think I mentioned this before, but this is the beginning of an experimental migration toward TypeScript. This may not get merged, so just consider this an experiment/WIP for now.

For starters, we'll have the TypeScript compiler building instead of Babel and just compile the js files, so we can migrate files one-at-a-time, but eventually some of the goals for this is:

- Full (strict) type coverage for all files, converted from js -> ts
- No `export default` anywhere, making it simpler to use an editor to do full project refactors based on the symbol
- Strong typings for splitting dialects out from the core
- Good use of generics to be able to return types (and possibly column names) in the builder
- Jest instead of mocha/tape - leveraging jest features like snapshot tests
- Simplified tests, a lot of things feel like a hack right now
- Less integration tests - shouldn't need to re-test things that are tested in the various dialect modules, we should assume that if the types check it's good, and we run a few simple queries just to be sure
- Stronger guarantees when doing more ambitious things like moving to async/await
- An overall better foundation for adding more ambitious features to the core

Probably will be broken for a bit, will be welcoming contributions once some of the core config for this is ironed out and we have some examples of the conventions to follow.


JakeGinnivan:
@tgriesser is you need a hand with this let me know :)

JakeGinnivan:
@tgriesser another option is to upgrade to babel 7 (which can parse the TypeScript types and ignore them), then setup tsc with --allow-js. Then you can start incrementally adding types and getting the benefit now.

elhigu:
@JakeGinnivan we just updated to babel 7 two days ago :)

tgriesser:
Closing in favor of https://github.com/tgriesser/knex-next (still super early on, nothing is actually setup from a connection standpoint, so don't look too much into it or expect it to work)

tgriesser:
@JakeGinnivan I'd personally rather have no types internally than partial/loose types. Without `--strict` and usage across the whole codebase I fail to see any benefit over using the definitions from Definitely Typed from an end user perspective. 

Also there's a ton of how the library is structured more generally I'd like to address with this effort and I believe TypeScript and some of the new ideas in the repo linked above will provide a path forward on that.

wubzz:
@tgriesser knex-next looks interesting, big step up. Would like to help when possible. I guess this early on would just be getting in the way, but just FYI.

PR #2700, opened 2018-07-09
Istanbul -> NYC

tgriesser:
Seems the semicolon is needed in mysql to execute the statement.


kibertoad:
@tgriesser if we are using nyc, maybe it's time to enable coverage threshold enforcement?

tgriesser:
This actually doesn't really fix any issues.

What I thought was a bug is actually just a super confusing API for the internals of how `.timeout()` is supposed to work / seems to be interacting with the connection dispose. This might be related to other timeout issues we're seeing in the test suite. Having a hard time following the logic here, looking forward to getting a chance to refactor it with types and possibly async/await internally which will remove the need for `dispose`.

@kibertoad Not yet. Would rather get code/test quality up in general first (refactor for less hacks) and then coverage will follow.

kibertoad:
@tgriesser Main idea behind enforced code coverage is ensuring that new code doesn't reduce it :). So it could actually be enforced at any current state.

tgriesser:
@kibertoad fair, feel free to open a PR :)

PR #2701, opened 2018-07-09
Update mocha

kibertoad:
Main reason why we didn't update mocha before was this: https://github.com/mochajs/mocha/issues/3044
However, these days old mocha is triggering security audit warnings due to old dependency with a critical vulnerability; hence maybe it is finally to move on :). I've added the --exit flag to replicate behaviour of old mocha.


tgriesser:
Great!

kibertoad:
@tgriesser Oracle builds are failing, though; does the error message look harmless enough?

tgriesser:
Yeah I'm personally not as concerned about oracle for now, @elhigu was working on some stuff to stabilize that test process I need to get back to him on.

elhigu:
Yep, oracle fails time to time, I'm suspecting that its because of some problem in transaction implementation.

PR #2711, opened 2018-07-13
Fix issue with select(0)

### LINKED ISSUE #2658: .select(0) should give select 0, not select *
# Environment

Knex version: 0.14.6
Database + version: PostgreSQL 10.4
OS: MacOS 10.13.4 (High Sierra)

# Bug

select(0) should give select 0 in SQL, instead gives select *.

I suspect there's a check somewhere that reads

```
if (!args[0]) { ... }
```

which should probably be

```
if (args[0] !== undefined) { ... }
```

2. Error message

Compare:
```
> knex('user').select(0).toSQL()
{ method: 'select',
  options: {},
  timeout: false,
  cancelOnTimeout: false,
  bindings: [],
  __knexQueryUid: 'a59d89c8-a6e5-4cf1-babc-33603f9a7a9f',
  sql: 'select * from "user"' }
> knex('user').select(knex.fn.now(), 0).toSQL()
{ method: 'select',
  options: {},
  timeout: false,
  cancelOnTimeout: false,
  bindings: [],
  __knexQueryUid: 'a1a63acd-ea77-4a90-af79-fafc8bf1fcb8',
  sql: 'select CURRENT_TIMESTAMP, 0 from "user"' }
```

The second error message illustrates correct behavior in other cases.

EDIT: what is more problematic is this:

```
> knex('user').select(0, knex.fn.now()).toSQL()
{ method: 'select',
  options: {},
  timeout: false,
  cancelOnTimeout: false,
  bindings: [],
  __knexQueryUid: 'f6ced08b-f018-4f5e-81ce-04d9cacedd86',
  sql: 'select * from "user"' }
```

See above.

wubzz:
Fixes #2658.

I looked through other builders to find similar cases but couldn't find any at first glance.

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -120,7 +120,7 @@ assign(Builder.prototype, {
   // Adds a column or columns to the list of "columns"
   // being selected on the query.
   columns(column) {
-    if (!column) return this;
+    if (!column && column !== 0) return this;
     this._statements.push({
       grouping: 'columns',
       value: helpers.normalizeArr.apply(null, arguments),
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -8139,4 +8139,14 @@ describe('QueryBuilder', function() {
       );
     });
   });
+
+  it('Can call knex.select(0)', function() {
+    testquery(qb().select(0), {
+      pg: 'select 0',
+      mysql: 'select 0',
+      mssql: 'select 0',
+      'pg-redshift': 'select 0',
+      oracledb: 'select 0',
+    });
+  });
 });


linkenneth:
Thanks for the fix! Not trying to be pedantic but this seems to add a specific check for the 0 case. Are there other similar cases that should also be fixed, or is this the only one? (for example, should false be a literal? what about null?)

wubzz:
@tgriesser I personally haven't had the opportunity of working on a project using TypeScript yet, but I'm curious to learn. So would you mind clarifying how TypeScript would have helped in this case?

From my albeit very brief understanding is that you'd do something like
```javascript
columns(column: string | number | IRaw) {
    if(!column) {
        return this;
    }
}
```

But TypeScript would not pick up on the if-statement in this case, would it?
```javascript
//Column === 0
if(!column) {
    return this;
}
```

elhigu:
@wubzz just by verifying that only String and raw would have been valid parameters for calling that method.

PR #2713, opened 2018-07-13
Use columnize instead of wrap in using().

### LINKED ISSUE #2136: [Knex 12.6] Using clause generates syntax error
When using knex 12.6
When using Postgres, calling
```js
.using('col');
```
Will produce a using clause in SQL without the brackets like
```sql
using "col"
```
It should be
```sql
using ("col")
```
Calling
```js
.using(['col', 'col2']);
```
Will produce SQL
```sql
using "col,col2"
```
It should be
```sql
using ("col", "col2")
```
Additionally, the using method is completely undocumented on knexjs.org as far as I can see.

wubzz:
This is an attempt to fix #2136. Also added an integration test, couldn't find any existing.

### DIFF
--- a/src/query/compiler.js
+++ b/src/query/compiler.js
@@ -594,7 +594,7 @@ assign(QueryCompiler.prototype, {
   },
 
   onUsing(clause) {
-    return this.formatter.wrap(clause.column);
+    return '(' + this.formatter.columnize(clause.column) + ')';
   },
 
   // Where Clause
--- a/test/integration/builder/joins.js
+++ b/test/integration/builder/joins.js
@@ -1889,5 +1889,59 @@ module.exports = function(knex) {
           );
         });
     });
+
+    it('Can use .using()', () => {
+      let joinName = 'accounts_join_test';
+
+      return knex.schema
+        .dropTableIfExists(joinName)
+        .then(() =>
+          knex.schema.createTable(joinName, (table) => {
+            table.bigint('id');
+            table.string('email');
+            table.integer('testcolumn');
+          })
+        )
+        .then(() =>
+          knex(joinName).insert([
+            {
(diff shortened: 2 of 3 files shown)


wubzz:
Also added issue in docs repo: https://github.com/knex/documentation/issues/124

wubzz:
Oracle doesnt seem to like my insert query using

```javascript
{
    id: 3,
    email: knex('accounts')
        .whereNot({ id: 3 })
        .select('email')
        .limit(1),
    testcolumn: 70,
},
```

>  begin execute immediate 'insert into "accounts_join_test" ("email", "id", "testcolumn") values ((select * from (select "email" from "accounts" where "id" = :1) where rownum <= :2), :3, :4)' using (select * from (select "email" from "accounts" where "id" = :1) where rownum <= :2), :3, :4; execute immediate 'insert into "accounts_join_test" ("email", "id", "testcolumn") values ((select * from (select "email" from "accounts" where not "id" = :1) where rownum <= :2), :3, :4)' using (select * from (select "email" from "accounts" where not "id" = :5) where rownum <= :6), :7, :8;end; - ORA-06550: line 1, column 198:

Does oracle not support subqueries in inserts or something?

wubzz:
Should be safe to merge now.

src/query/compiler.js
```diff
@@ -594,7 +594,7 @@ assign(QueryCompiler.prototype, {
   },
 
   onUsing(clause) {
-    return this.formatter.wrap(clause.column);
+    return '(' + this.formatter.columnize(clause.column) + ')';
```
elhigu: I really don't remember nor how formatter.wrap or formatter.columnize behave, but this seems legit 👍

PR #2719, opened 2018-07-18
Add option of existingType to .enum method to support repeated use of enums

ekeric13:
I was super happy to see knex supporting native enums in postgres, but when I had an enum that I used repeatedly across different columns in different tables, I was running into errors as the `enum` method created the type enum for you each time. The type would already exist so it would error. I only wanted to create the type once, afterwards I just wanted to use that enum.

Could we add a flag in the options param to not run the sql that creates the type enum?

The first time you run it you could do

```
table.enu('current_mood', ['sad', 'ok', 'happy'], { useNative: true, enumName: 'mood' })
```

And then the other times you can run

```
table.enu('mood', ['sad', 'ok', 'happy'], { useNative: true, enumName: 'mood', existingType: true })
```

### DIFF
--- a/src/dialects/postgres/schema/columncompiler.js
+++ b/src/dialects/postgres/schema/columncompiler.js
@@ -30,7 +30,10 @@ assign(ColumnCompiler_PG.prototype, {
   enu(allowed, options) {
     options = options || {};
 
-    const values = allowed.join("', '");
+    const values =
+      options.useNative && options.existingType
+        ? undefined
+        : allowed.join("', '");
 
     if (options.useNative) {
       if (!options.existingType) {
--- a/src/dialects/postgres/schema/columncompiler.js
+++ b/src/dialects/postgres/schema/columncompiler.js
@@ -33,9 +33,11 @@ assign(ColumnCompiler_PG.prototype, {
     const values = allowed.join("', '");
 
     if (options.useNative) {
-      this.tableCompiler.unshiftQuery(
-        `create type "${options.enumName}" as enum ('${values}')`
-      );
+      if (!options.existingType) {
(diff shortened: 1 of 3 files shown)


heisian:
I second this proposition and motion for a move to merge :P

elhigu:
Sounds reasonable, I cannot think any better API for handling this situation.

kibertoad:
@ekeric13 Would you consider finishing this PR?

ekeric13:
Yeah for sure. @heisian 	and I have a working implementation, just need to write tests for it.

kibertoad:
Awesome! Also PR for documentation in https://github.com/knex/documentation would be most appreciated :)

kibertoad:
@ekeric13 Thank you very much! It's 0:30 here already, it was a long day; I'll take a look at it tomorrow and comment ASAP, sorry for the delay.

ekeric13:
So what is the process after this? Just will be merged in for the next release?

kibertoad:
@ekeric13 Probably can be merged within 0.16.0 window, but before that I assume we should consider the alternative improved syntax that you've suggested?..

kibertoad:
@ekeric13 Looked at previous and current syntax and I think that it makes perfect sense, would you do the change?

ekeric13:
Okay, added the proposed ternary operator to make the function type agnostic of the values in the case where the values aren't being used.

Updated the docs as well for further clarification:
https://github.com/knex/documentation/pull/147/commits/e3d6f48bb6f81c4bcb5aa56fd7d813d72dfb1af1

kibertoad:
@ekeric13 Awesome work, thank you!

heisian:
🎊

PR #2724, opened 2018-07-21
Allow timestamp with timezone on mssql databases

caiquecastro:
The SQL Server supports the following date / time data types:

date (Transact-SQL)
datetime (Transact-SQL)
datetime2 (Transact-SQL)
datetimeoffset (Transact-SQL)
smalldatetime (Transact-SQL)
time (Transact-SQL)

In my opinion, timestamp can be represented by (datetime, datetime2, datetimeoffset and smalldatetime).

Regarding the docs and integration tests, I want to wait for the reply about the incompatibility between dialects on the parameter for timestamp function on schema builder.

### DIFF
--- a/src/dialects/mssql/schema/columncompiler.js
+++ b/src/dialects/mssql/schema/columncompiler.js
@@ -62,7 +62,9 @@ assign(ColumnCompiler_MSSQL.prototype, {
 
   datetime: 'datetime',
 
-  timestamp: 'datetime',
+  timestamp(useTz) {
+    return useTz ? 'datetimeoffset' : 'datetime';
+  },
 
   bit(length) {
     if (length > 1) {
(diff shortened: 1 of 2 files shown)


src/dialects/mssql/schema/columncompiler.js
```diff
@@ -62,7 +62,9 @@ assign(ColumnCompiler_MSSQL.prototype, {
 
   datetime: 'datetime',
 
-  timestamp: 'datetime',
+  timestamp(useTz) {
```
elhigu: use options object instead of single boolean parameter.

caiquecastro: Regarding the object parameter on the function, it will be different between the dialects

elhigu:
Code seems to be fine, this is so simple change that I don't require an integration test for it now... However documentation PR is still needed before this can be merged.

caiquecastro:
You mean the PR on knex/documentation repo?

I just did it.

https://github.com/knex/documentation/pull/128

Please check for me.

Thanks

elhigu:
Thanks!

PR #2730, opened 2018-07-25
Fix issues with warnPromise when migration does not return a promise.…

### LINKED ISSUE #2725: Fatal error when running a migration with knex CLI which returns `undefined` instead of a Promise
# Environment

Knex version:
Database + version: PostgreSQL 10.4 + Knex 0.15.2
OS: Mac OS

# Bug

Fatal error when running a migration with knex CLI which returns `undefined` instead of a Promise

2. Error message

`TypeError: Cannot read property 'knex' of undefined`
This originates from this line https://github.com/tgriesser/knex/blame/f55140f4f6684730d8e2cdca5b351b30a99afc81/src/migrate/index.js#L413

I believe this line is a bug.

Example migration run with `knex migrate:latest` or `knex migrate:rollback`:

```node
exports.up = () => {}
exports.down = () => {}
```

wubzz:
… Fixes #2725

### DIFF
--- a/src/migrate/index.js
+++ b/src/migrate/index.js
@@ -354,7 +354,11 @@ export default class Migrator {
           if (!trx && this._useTransaction(migration, disableTransactions)) {
             return this._transaction(migration, direction, name);
           }
-          return warnPromise(migration[direction](trxOrKnex, Promise), name);
+          return warnPromise(
+            this.knex,
+            migration[direction](trxOrKnex, Promise),
+            name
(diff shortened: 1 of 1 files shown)


sanion:
I'm running into the same issue :( Its been 4 hours already :(

kibertoad:
@sanion Try using latest NEXT (0.16) release, I don't think fix is in stable branch yet.

PR #2733, opened 2018-07-27
Compile with before update so that bindings are put in correct order

### LINKED ISSUE #2727: CTE queries binding wrong values
# Environment

Knex version: 0.15.2
Database + version: PostgreSQL 10.4
OS: macOS 10.13.6

# Bug
When using the `with` method to generate data-modifying CTEs, there seems to be a problem with the value binding; the values passed into the query builders are being bound to the wrong placeholders.

I've got a minimal example:
```js
  client: 'pg',
  connection: 'psql://postgres@localhost:5432/<DB_NAME>'
});

const q1 = knex
  .table('user')
  .update({ name: 'foo' })
  .where({ group_id: 1 })

const q2 = knex
  .with('updated_group', (qb) => 
    qb
      .table('group')
      .update({ group_name: 'bar' })
      .where({ group_id: 1 })
      .returning('group_id')
  )
  .table('user')
  .update({ name: 'foo' })
  .where({ group_id: 1 })

console.log('WITHOUT CTE')
console.log(q1.toString())

console.log('WITH CTE')
console.log(q2.toString())
```
This gives me the output:
```
WITHOUT CTE
update "user" set "name" = 'foo' where "group_id" = 1

WITH CTE
with "updated_group" as (update "group" set "group_name" = 'foo' where "group_id" = 1 returning "group_id") update "user" set "name" = 'bar' where "group_id" = 1
```

You can see that in the second example (`WITH CTE`), the bindings of `foo` and `bar` are the wrong way round.

If I use `toSQL` to inspect these queries, the following is the `bindings` array for the second query:
```
bindings: [ 'foo', 1, 'bar', 1 ],
```
So seems like these are being substituted into the query sequentially, and since the `WITH` clause will always come first, the bindings are in the wrong order (should be `[ 'bar', 1, 'foo', 1 ]`). Obviously changing the order of the method chain doesn't affect the final query.

Am I just misunderstanding the `with` API, or is this actually an issue? Thanks.

wubzz:
Fixes #2727

### DIFF
--- a/src/dialects/mssql/query/compiler.js
+++ b/src/dialects/mssql/query/compiler.js
@@ -84,6 +84,7 @@ assign(QueryCompiler_MSSQL.prototype, {
   // Compiles an `update` query, allowing for a return value.
   update() {
     const top = this.top();
+    const withSQL = this.with();
     const updates = this._prepUpdate(this.single.update);
     const join = this.join();
     const where = this.where();
@@ -91,7 +92,7 @@ assign(QueryCompiler_MSSQL.prototype, {
     const { returning } = this.single;
     return {
       sql:
-        this.with() +
+        withSQL +
         `update ${top ? top + ' ' : ''}${this.tableName}` +
         ' set ' +
         updates.join(', ') +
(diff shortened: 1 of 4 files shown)


elhigu:
Looks good I think 👍

PR #2735, opened 2018-07-28
Feature/2690: Multiple migration directories

kibertoad:


kibertoad:
(missing an essential test and some necessary code, closing for now)

kibertoad:
Still broken, need to figure out the black magic of running integration tests locally and fix what's remaining.

src/migrate/Migrator.js
```diff
@@ -42,6 +40,7 @@ const CONFIG_DEFAULT = Object.freeze({
   schemaName: null,
   directory: './migrations',
   disableTransactions: false,
+  sortDirsSeparately: false,
```
kibertoad: The idea behind this flag is that while we are forced to sort all migrations at once for backwards compatibility (so that their execution order does not change), it is typically much more convenient to execute directories sequentially and only worry about migration ordering within the scope of a single directory (where it is very visible).

kibertoad:
@elhigu This ended up being way more exhausting than I originally estimated it to be, but it finally seems to be done. Please review :)

kibertoad:
@tgriesser @elhigu Any chance for a review :)?

JakeGinnivan:
Hi, this is actually quite related to what I would like to do, but it would be great to make the knex migrations plugable. 

The existing could become knex-file-migrations. I would like to have an in memory provider which would allow me to put my migrations into my bundle, solving a bunch of code sharing and build issues for me.

Thoughts?

elhigu:
@JakeGinnivan I wouldn't like to have it separated to different npm package, which would increase overhead for developers. Also I don't understand with that description how exactly you would like to make it plugable. Please open feature request about it with detailed info. This is not good place to talk about it.

JakeGinnivan:
Will just open a pull request and use that as a discussion point :)

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 
 # 0.15.2 - 19 Jul, 2018
 
+- Introduced support for specifying multiple directories for the Migrator
```
elhigu: This actually seems to be in wrong position in changelog and needs link to PR number

src/migrate/Migrator.js
```diff
@@ -295,9 +299,17 @@ export default class Migrator {
   // passing any `variables` given in the config to the template.
   _writeNewMigration(name, tmpl) {
     const { config } = this;
-    const dir = this._absoluteConfigDir();
+    const dirs = this._absoluteConfigDir();
+    if (dirs.length > 1) {
```
elhigu: Maybe here it would be more usable to use the last migration dir where to put new migration and add that to documentation. Otherwise one would always need separate configuration just for being able to create new migration...

kibertoad: Good idea!

CHANGELOG.md
kibertoad: Right, sorry, will move to unreleased part.

src/migrate/Migrator.js
```diff
@@ -97,7 +93,7 @@ export default class Migrator {
         .tap(validateMigrationList)
         .then((val) => this._getLastBatch(val))
         .then((migrations) => {
-          return this._runBatch(map(migrations, 'name'), 'down');
+          return this._runBatch(migrations, 'down');
```
elhigu: Was _runBatch somehow changed to allow to take migration list as an input instead of migration names?

kibertoad: _runBatch per se no, it merely passes migrations param to getNewMigrations method which does expect the migration list.

kibertoad: This change was needed because batch execution needs to be aware of directories as well.

JakeGinnivan:
Hey @kibertoad and @elhigu, don't want to complicate this. But if https://github.com/tgriesser/knex/pull/2775 goes ahead, conceptually the migrator is not running against folders anymore.

The same refactoring would still work but it would be a missed opportunity to simply put this logic inside the `FsMigrations` class. It would also significantly reduce the changes to the Migrator class.

Because the PRs are so related, it would be great to talk about my PR in relation to this one. Thoughts?

JakeGinnivan:
Obviously my preference would be to have this PR rebase on top of my PR. 

Another side note, is this behaviour could be gained without a PR to knex at all if someone wanted a different behaviour. This one makes sense to put in core, but not all custom migration layouts would.

kibertoad:
@elhigu All comments addressed, build pass. Is anything remaining?

elhigu:
@JakeGinnivan thanks for the heads up. I haven't had time to check that your PR yet to see how it reflects to this one to decide which should be on top of which.

elhigu:
Thanks @kibertoad !

unhawkable:
i'm afraid that knex migrate:make command is not covered - where will it put a new migration when mutliple directories are configured?

kibertoad:
@unhawkable See https://github.com/tgriesser/knex/pull/2735/commits/306b8bd0a3fc6515899fe2372438bbce8b404d88

Probably I should update documentation to codify that behaviour.

unhawkable:
ok thanks i see now. It will take the last dir. It would be super-cool though if it would ask for a directory to put the new migration in in an interactive mode with a list to choose from 👍 - just a suggestion ;)

elhigu:
@unhawkable yes, there is room for improvement if someone needs better control on that.

JakeGinnivan:
Ah, I see now. If you store the relative path it may introduce a number of other compatibility issues and breaking changes. Ignore ☝️

elhigu:
yep, and prevent moving migration from one directory to another.

PR #2738, opened 2018-07-30
MSSQL removes support column width/length on data type int

kamote:
- fixes error `Cannot specify a column width on data type int`


kamote:
@elhigu Most of current integration test are now using ```.integer() `` w/o length
https://github.com/tgriesser/knex/blob/master/test/integration/schema/index.js#L546-L547

elhigu:
So if current integration tests are creating integers with specific width how come current integration tests are not throwing that error?

kamote:
@elhigu upon looking, there is no MSSQL integration test that uses specific width which explains integration tests are not throwing.
In our case we found this kind of error since we are deploying same codebase with diff DB clients

elhigu:
@kamote thanks for looking that up. In that case such test should be added (regression test to make sure the same thing will not start failing again).

kibertoad:
@kamote Could you give me write access on your repo so that I could add test?

elhigu:
@kibertoad did you try just to pull it and push changes? I have usually been able to do that... if it doesnt work, you can create your own PR also and link it here 👍

Issue #2262, opened 2017-10-10
Cannot insert explicit value on seed run


kirkor:
I'm using MSSQL server.

I want to insert data with ID's and getting this SQL error:

`Cannot insert explicit value for identity column in table 'AS_DEF' when ID ENTITY_INSERT is set to OFF`

The resolution is to execute `SET IDENTITY_INSERT as_def ON` but I can't do this in the same session. Any solution for this?

elhigu:
You can set `SET IDENTITY_INSERT as_def ON` for each connection in afterConnect hook or you can start transaction, set `SET IDENTITY_INSERT as_def ON` inside transaction and then do the insert.

kirkor:
I'm new to node.js and knex as well. I cant work it out :(

Do you have any working example?

I'm trying transactions currently. And have something like this:
`
exports.seed = function (knex, Promise) {

    const tableName = 'as_def';

    return knex.transaction(function (trx) {
        return knex(tableName)
            .transacting(trx)
            .del()
            .then(function () {
                return trx.raw('SET IDENTITY_INSERT ' + tableName + ' ON').then(function() {
                    return knex
                    .transacting(trx)
                    .insert([{...},])
                    .into(tableName);
                });
            })
            .then(() => {
                return questions(knex, Promise);
            })
            .then(() => {
                return answers(knex, Promise);
            })
            .then(trx.commit)
            .catch(trx.rollback);
    });
};
`

kirkor:
`afterCreate` is not working as well:

`afterCreate: function (conn, done) {
            var request = new sql.Request(conn);
            request.query('SET IDENTITY_INSERT [ey_survey_analysis].[dbo].[assessments_definitions] ON;', function (err, recordset) {
                if(err) {
                    console.log('error');
                }
            });

            done();
        }`

For me looks like a BUG :P

elhigu:
You have some errors in both of your examples.

First one should go somehow like this:

```
exports.seed = function (knex, Promise) {
    return knex.transaction((trx) => {
        return trx('as_def').del()
            .then(() => trx.raw('SET IDENTITY_INSERT as_def ON'))
            .then(() => trx('as_def').insert([{...},]));
    })
    .then(() => console.log('transaction was committed'))
    .catch((err) => console.log('transaction was rolled back'))
    ;
};
```

If it fails, check the error and try to run with DEBUG=knex:* environment variable set to see if some of the sent queries does not make sense.

[...]

PR #2739, opened 2018-07-30
MSSQL Implement `IDENTITY_INSERT ON/OFF` insert query

kamote:
- utilize `_queryContext` to determine if `identityInsert` set to true wrap insert query into `IDENTITY_INSERT ON/OFF`

#Case
- During `integration test` there are some scenario that we need to set values on  primary keys.

## Related issue
https://github.com/tgriesser/knex/issues/2262

## Usage
It's to the user to determine if a table has a identity column
### knex
```
knex('users')
  .queryContext({ identityInsert: true })
  .insert({
    id: 1,
    user: 'foo',
  }, 'id')
```

###  objection.js custom query builder usage
```
insert(args) {
    const knex = this._modelClass.constructor.knex;
    const client = this._modelClass.client;

    if (!this.has(/returning/)) {
      this.returning(this._modelClass.getIdColumn());
    }

    if (client === 'mssql') {
      // eslint-disable-next-line no-unused-vars
      this.onBuildKnex((knexBuilder, objectionBuilder) => {
        knexBuilder.queryContext({
          identityInsert: isIdentityWrapperRequired(
            this.jsonProperties,
            Array.isArray(args) ? args[0] : args
          ),
        });
      });
    }

    return super.insert.apply(this, [args]);
  }
```

### DIFF
--- a/src/dialects/mssql/query/compiler.js
+++ b/src/dialects/mssql/query/compiler.js
@@ -35,6 +35,7 @@ assign(QueryCompiler_MSSQL.prototype, {
   // Compiles an "insert" query, allowing for multiple
   // inserts using a single query statement.
   insert() {
+    const queryContext = this.formatter.builder._queryContext || {};
     const insertValues = this.single.insert || [];
     let sql = this.with() + `insert into ${this.tableName} `;
     const { returning } = this.single;
@@ -75,6 +76,12 @@ assign(QueryCompiler_MSSQL.prototype, {
         sql = '';
       }
     }
+
+    if(queryContext && queryContext.identityInsert) {
+      sql = `SET IDENTITY_INSERT ${this.tableName} ON; ${sql}; `;
+      sql += `SET IDENTITY_INSERT ${this.tableName} OFF;`;
+    }
+
     return {
       sql,
       returning,
--- a/test/integration/builder/inserts.js
+++ b/test/integration/builder/inserts.js
@@ -1300,5 +1300,63 @@ module.exports = function(knex) {
           });
       });
     });
+
+    it('should allow to set primarykey/identitykey value during insert using queryContext', function() {
+      return knex.schema
+        .dropTableIfExists('test_default_table4')
+        .then(function() {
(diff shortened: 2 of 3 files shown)


src/dialects/mssql/query/compiler.js
```diff
@@ -75,6 +76,12 @@ assign(QueryCompiler_MSSQL.prototype, {
         sql = '';
       }
     }
+
+    if(queryContext && queryContext.identityInsert) {
```
elhigu: So this leaves connection to a state it was in the last time and doesn't reset it to on, after query is done? Looks like a bug.

elhigu:
I think that I wouldn't like to include this to knex, but let the poeple who likes to set identity insert on /off to do it manually with knex.raw. It also higly database specific funtionality and breaks the principle where each builder should generate only single query (if mssql driver supports running multiple queries separated with semicolon and with bingings passed nicely it should be a problem).

Also I don't like the idea of insert(), depending on arbitary value in queryContext.

Better to open feature request where the feature can be discussed over and decide if there is any way how it could be included to knex.

src/dialects/mssql/query/compiler.js
kamote: @elhigu Hmmm, no its not a bug, connection/session state gets reset afterwards with 
```
sql = `SET IDENTITY_INSERT ${this.tableName} ON; ${sql}; `;
sql += `SET IDENTITY_INSERT ${this.tableName} OFF;`;
```

kamote:
@elhigu Thanks for the review, actually this idea came from `sequelize` we recently migrated from `sequelize` to `objection.js`.

https://github.com/sequelize/sequelize/blob/master/lib/dialects/mssql/query-generator.js#L366-L370

If we will not use `queryContext` maybe we can modify insert, like adding third parameter 

```
knex('users')
  .insert({
    id: 1,
    user: 'foo',
  }, 'id', true)
```

elhigu:
You can easily implement this kind of functionality with objection's various hooks. That is better place to put this kind of project specific functionality.

Issue #269, opened 2014-05-12
Chained update with increment?


imiric:
Sorry if I'm missing something, but I can't seem to be able to use `.increment()` along with updating other fields on the table.

For instance, I'd want to do:

``` js
knex('products')
  .whereIn('id', ids)
  .update({'updated_at': new Date()})
  .increment('count', 1);
```

But running this I get:

```
Possibly unhandled Error: The query type has already been set to update
```

I understand why it happens, I just think it would be nice to be able to add other fields to the update.

Alternatively, maybe we could remove `.increment()` altogether and somehow in the `.update()` itself be able to reference the table's columns. Maybe something like:

``` js
knex('products')
  .whereIn('id', ids)
  .update({
    'updated_at': new Date(),
    'count': function() {
      return this + 1;
    }
  });
```

Yeah, not sure about the "this + 1" part, but maybe have another way of saying "the existing column value". This would avoid the need for `.increment()` and `.decrement()`.

I realize I could write this using `.raw()`, but it would be nice to have it in the library.

What do you think?

boompig:
I want to reopen this issue, because it makes sense that chaining `.update()` and `.increment()` together should work, but in fact only increment is applied, and update is disregarded silently.

rhys-vdw:
@boompig Agreed. Also looking at the `increment()` definition it seems it's not possible do multiple increments simultaneously.

[`increment` and `decrement`](https://github.com/tgriesser/knex/blob/775d6c2af7106b38d53083bd5185c189f2fcd01f/lib/query/builder.js#L568-L576) both call [`_counter'](https://github.com/tgriesser/knex/blob/775d6c2af7106b38d53083bd5185c189f2fcd01f/lib/query/builder.js#L710-L721), which is then [compiled as `knex.raw`](https://github.com/tgriesser/knex/blob/775d6c2af7106b38d53083bd5185c189f2fcd01f/src/query/compiler.js#L289-L298). This sets `QueryBuilder._method` to `'counter'` instead of `'update'` so the two are treated as different "types" of queries.

The only advantage I can see to using method `'counter'` instead of `'update'` is allowing different dialects to override the `increments`/`decrements` behaviour specifically. However they all just use `raw('columnName + 1')`.

[...]

PR #2740, opened 2018-07-31
Allow chaining of increment, decrement, and update

wubzz:
Relates to #269 

Changes include:
- Chainable `.increment`, `.decrement`, `.update` in the same build-chain
- Can call `.increment` / `.decrement` on same column multiple times
- Uses reduce for final query value
- Object syntax for `.increment` / `.decrement`
- ~Generates queries with `"column" + ?` only~ Reverted!
- Added `.clearCounters()` to clear previous `.increment` / `.decrement` calls in a build chain
- If `.increment` / `.decrement` is called for a column that has already been specified in `.update` then value specified in `.update` will take presedence.

See tests for examples.

May have missed something or completely misinterpreted the issue, in which case please let me know.

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -806,13 +806,36 @@ assign(Builder.prototype, {
   },
 
   // Increments a column's value by the specified amount.
-  increment(column, amount) {
+  increment(column, amount = 1) {
+    if (isObject(column)) {
+      for (const key in column) {
+        this._counter(key, column[key]);
+      }
+
+      return this;
+    }
+
     return this._counter(column, amount);
   },
 
   // Decrements a column's value by the specified amount.
-  decrement(column, amount) {
-    return this._counter(column, amount, '-');
+  decrement(column, amount = 1) {
+    if (isObject(column)) {
+      for (const key in column) {
+        this._counter(key, -column[key]);
+      }
+
+      return this;
+    }
+
+    return this._counter(column, -amount);
+  },
+
+  // Clears increments/decrements
+  clearCounters() {
+    this._single.counter = [];
+
+    return this;
   },
 
   // Sets the values for a `select` query, informing that only the first
@@ -971,15 +994,18 @@ assign(Builder.prototype, {
   // ----------------------------------------------------------------------
 
   // Helper for the incrementing/decrementing queries.
-  _counter(column, amount, symbol) {
-    let amt = parseFloat(amount);
-    if (isNaN(amt)) amt = 1;
-    this._method = 'counter';
-    this._single.counter = {
+  _counter(column, amount) {
+    amount = parseFloat(amount);
+
+    this._method = 'update';
+
+    this._single.counter = this._single.counter || [];
+
+    this._single.counter.push({
       column,
-      amount: amt,
-      symbol: symbol || '+',
-    };
+      amount,
+    });
+
     return this;
   },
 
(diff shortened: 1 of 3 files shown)


src/query/compiler.js
```diff
@@ -736,7 +722,29 @@ assign(QueryCompiler.prototype, {
...
-  _prepUpdate(data) {
+  _prepUpdate(data = {}) {
+    const { counter = [] } = this.single;
+    const grouped = groupBy(counter, 'column');
+
+    for (const column in grouped) {
+      //Skip?
+      if (has(data, column)) {
+        //Needed?
+        this.client.logger.warn(
```
Zolmeister: Most warns have to do with feature support by different SQL engines;
whereas errors are generally thrown for incorrect query construction. I'd say throw an error.
(just glancing at source/tests)

src/query/builder.js
```diff
@@ -806,13 +806,36 @@ assign(Builder.prototype, {
...
+      }
+
+      return this;
+    }
+
+    return this._counter(column, -amount);
+  },
+
+  // Clears increments/decrements
+  clearCounters() {
```
Zolmeister: Could you explain what this endpoint might be used for?
My reaction is that it feels unnecessary

src/query/compiler.js
wubzz: I think its bold to assume calling `.increments` first and then `.update` with the same column to be a user mistake, it may just as likely be perfectly intentional.

No strong opinion on what should happen here, but definitely some form of message must be spit out of knex instead of silently ignoring. I opted for warn over error due to the argument above, but not 100% against changing this.

src/query/builder.js
wubzz: It follows the same principals as `.clearSelect()`, `.clearWhere()`, `.clearOrder()`.

I've yet to find a real life use-case for these functions myself, but for consistency I added one here as well. Suppose someone has found a use for the other clear-functions, else they wouldn't exist.

wubzz:
> Generates queries with "column" + ? only

On second thought this might be a bit confusing since knex internally turns the input value negative. Is it better to preserve the original `+` vs `-` in the query depending on final output value?

Would be
```javascript
const symbol = value < 0 ? '-' : '+';

if(symbol === '-') {
 value = -value
}

data[column] = this.client.raw('?? ? ?', [column, symbol, value])
//"column" = "column" - 500
```

Instead of
```javascript
data[column] = this.client.raw('?? + ?', [column,  value])
//"column" = "column" + -500
```

tgriesser:
> Can call .increment / .decrement on same column multiple times

I think this might be confusing behavior with the rest of the library, where latter calls overwrite the former.

tgriesser:
Otherwise I think it looks good, could you add a test for having multiple different columns called with increment/decrement.

tgriesser:
@wubzz no sorry I just meant the idea of having 
```
.increments('test', 10).increments('test', 20)
``` 
output 
```
test = test + 30
``` 
rather than 
```
test = test + 20
```
is confusing based on the api patterns elsewhere. Does that make sense?

tgriesser:
Yeah let's change it. If people really need that logic they can just precompute the value in the loop and then set once with the final value value rather than having the api handle it.

wubzz:
Updated to reflect review comments.

src/query/builder.js
capaj: @wubzz 
we use it like this with objection.js for example:

image

Issue #2707, opened 2018-07-11
Feature Request: where style clauses within join


alaarihan:
# Environment

Knex version: latest
Database + version: MySQL 5.7.x
OS: Ubuntu 14.04
When I use this:
```
let products = await db.select({
      id: 'p.ID',
      status : 'p.post_status',
      name: 'p.post_title',
      // type: 'terms.name',
      price: 'price.meta_value',
      createdAt: 'p.post_date_gmt',
      updatedAt: 'p.post_modified_gmt'
    })
    .from({p:'wp_posts'})
    .leftJoin({price: 'wp_postmeta'},
    function() {
      this.on('p.id', '=', 'price.post_id').andOn('price.meta_key', '_regular_price')
    })
```

When I execute this query I get this error:
```
select `p`.`ID` as `id`, `p`.`post_status` as `status`, `p`.`post_title` as `name`, `price`.`meta_value` as `price`, `p`.`post_date_gmt` as `createdAt`, `p`.`post_modified_gmt` as `updatedAt` from `wp_posts` as `p` left join `wp_postmeta` as `price` on `p`.`id` = `price`.`post_id` and `price`.`meta_key` = `_regular_price` order by `post_date_gmt` asc limit 2 - Unknown column '_regular_price' in 'on clause'
```

I need to use a string here not a column for join

tgriesser:
The current solution, which admittedly is very ugly is to do:
```
.andOn('price.meta_key', db.raw('"_regular_price"'))
```

We should probably add the concept of a where clause to the join object, so you could do:

```
  this.on('p.id', 'price.post_id').andWhere('price.meta_key', '_regular_price')
```

and it would be output as:

```sql
select .... 
left join `wp_postmeta` as `price` on `p`.`id` = `price`.`post_id` 
  and `price`.`meta_key` = "_regular_price" 
...
```

wubzz:
Wouldn't it make more sense with something like `.onValue` / `.andOnValue` or `.onVal` / `.andOnVal` or `.onEquals` / `.andOnEquals` ? _(for lack of better name)_ 

It just sounds a bit weird to have `.where` / `.andWhere` in a join builder since it doesn't generate a where clause.

elhigu:
> .andOn('price.meta_key', db.raw('"_regular_price"'))

In 0.15.0 there is also helper for being able to write ` db.raw('??', '_regular_price')` easier like `db.ref('_regular_price')` which does quoting automatically.

[...]

PR #2746, opened 2018-08-01
Feature/join on val and on val or on val

wubzz:
Relates to #2707 - See discussion

Allows join using parameters instead of having to rely on `Raw`.

Before
```javascript
this.on('p.id', '=', 'price.post_id').andOn('price.meta_key', knex.raw('"_regular_price"'))
```

After
```javascript
this.on('p.id', '=', 'price.post_id').andOnVal('price.meta_key', '_regular_price')
```

### DIFF
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -8149,4 +8149,59 @@ describe('QueryBuilder', function() {
       oracledb: 'select 0',
     });
   });
+
+  it('join with onVal andOnVal orOnVal', () => {
+    testsql(
+      qb()
+        .select({
+          id: 'p.ID',
+          status: 'p.post_status',
+          name: 'p.post_title',
+          // type: 'terms.name',
+          price: 'price.meta_value',
+          createdAt: 'p.post_date_gmt',
+          updatedAt: 'p.post_modified_gmt',
+        })
+        .from({ p: 'wp_posts' })
+        .leftJoin({ price: 'wp_postmeta' }, function() {
+          this.on('p.id', '=', 'price.post_id')
+            .onVal(function() {
+              this.onVal('price.meta_key', '_regular_price').andOnVal(
+                'price_meta_key',
+                '_regular_price'
+              );
+            })
+            .orOnVal(function() {
+              this.onVal('price_meta.key', '_regular_price');
+            });
+        }),
+      {
+        pg: {
+          sql:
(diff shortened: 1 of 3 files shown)


test/unit/query/builder.js
```diff
@@ -8149,4 +8149,59 @@ describe('QueryBuilder', function() {
...
+        .from({ p: 'wp_posts' })
+        .leftJoin({ price: 'wp_postmeta' }, function() {
+          this.on('p.id', '=', 'price.post_id')
+            .onVal(function() {
+              this.onVal('price.meta_key', '_regular_price').andOnVal(
+                'price_meta_key',
+                '_regular_price'
+              );
+            })
+            .orOnVal(function() {
```
tgriesser: Why does this need the extra closure here?

tgriesser:
Also based on your comment on #2707:

> What I don't really like is having to call .onVal() with an additional .onVal() + .andOnVal() to get it wrapped, but I guess it's the same for normal join functions, so maybe it's ok..

Could we just mirror object syntax from where `and` clauses... 

```
.onVal({
  'price.meta_key': '_regular_price',
  'price_meta_key': '_regular_price'
})
.orOnVal('price_meta.key', '_regular_price')
```

test/unit/query/builder.js
wubzz: It doesn't necessarily. It's just personal preference. I usually prefer keeping conditions within bounds of closures so that future additions don't accidentally end up in the wrong closure. 

Calling without function would also yield valid syntax.

wubzz:
@tgriesser 
> Could we just mirror object syntax from where and clauses...

Object syntax is supported already, but just like where clauses this does not generate a closure. So the closure syntax requires functions.

```javascript
knex('table').where({col1: '1', col2: '2'}).orWhere('col3', '4').toString()
//"select * from `table` where `col1` = '1' and `col2` = '2' or `col3` = '4'"
```

tgriesser:
Oh that's weird, I thought it did. 🤔

Just curious, does the behavior of the join differ with or without the explicit grouping via parens?

wubzz:
I can't speak for other databases but Postgres is 'smart enough' to understand these priorities in most cases. For example the test in this PR Postgres would yield the same join result with or without grouping via parens.

tgriesser:
Cool, I think this looks good. @elhigu any thoughts here? 

Going to wait to see if there's anything else that should go out with 0.15.x before merging this, will hopefully have this out soon though.

elhigu:
Looks like some breaking changes has already been merged in after last release so I don't think that 0.15 will have any additional patch versions. No other thoughts about this than that. I suppose this one is good. 

On objection side this had been resolved by having `lit()` helper function, which is used to tell query builder that passed value should be treated as value instead of identifier.

Like: `.on('column', lit('Im value'))`

We got a bit annoyed about having so many different variants of basically the same method, like `.where()`, `.orWhere()`, `.whereRef()`, `.orWhereRef()` + `.andXXXXXX´ variants so having `ref(indentifier)` and `lit(val)` helpers allowed us to get rid of having to tell parameter type by calling different function name :) 

No extra thoughts about this. PR look fine to me.

PR #2749, opened 2018-08-02
Add tests for multiple union arguments with callbacks and builders

scotttrinh:
I've [been sending pull requests the TypeScript DefinitelyTyped repo](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/27818) when I've found things like this, but I thought it might be nice to also upstream some test changes as I run across cases that aren't being covered already and are ambiguous in the documentation. I'll also open a documentation PR that disambiguates this, too.

Most of these tests are superfluous once/if knex moves to TypeScript, since it is more or less just type assertions.

### DIFF
--- a/test/integration/builder/unions.js
+++ b/test/integration/builder/unions.js
@@ -4,7 +4,7 @@
 
 module.exports = function(knex) {
   describe('unions', function() {
-    it('handles unions', function() {
+    it('handles unions with a callback', function() {
       return knex('accounts')
         .select('*')
         .where('id', '=', 1)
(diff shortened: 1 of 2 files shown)


test/integration/builder/unions.js
```diff
@@ -4,7 +4,7 @@
 
 module.exports = function(knex) {
   describe('unions', function() {
-    it('handles unions', function() {
+    it('handles unions with a callback', function() {
```
wubzz: Since neither of these tests make assertions on the result I think either adding assertions of the resultset or adding corresponding unit tests to ensure generated SQL is what's to be expected.

I see the latter has been done for some of the new tests, but not all. For example raw and pure functions unit tests are missing.

scotttrinh: I assumed since none of these tests have assertions on the resultset that we were just detecting if it threw.

I'm very happy to add the other unit test, though. I'll go ahead and cover each possible case!

scotttrinh: ⚡️ cccae9a

Instead of writing a ton of repetitive tests, I opted to just mix the types for the array and multi-argument cases. Let me know if that feels comprehensive enough or if we ought to add a whole section testing each of the possible cases. I think there are 9 total: single callback, builder, raw; array callback, builder, raw; multi-argument callback, builder, raw.

wubzz: Looks good to me 👍

wubzz:
Thanks!

Issue #2428, opened 2018-01-16
table.bigIncrements is not working with Oracle.


krutiparmar:
# Environment
Windows 8

NodeJs version: 6.3.1
KnexJS version : 0.12.8
Database + version: Oracle 12c

@atiertant. 

I am creating a table with the following code in Oracle using Knex :
 knex.schema.createTable('APIUsers', function(table) {
      table.bigIncrements('UserId').primary().unsigned();
      table.string('UserName', [100]).notNullable().unique();
      table.string('Password', [256]).notNullable();
      table.string('EmailAddress', [100]).notNullable();
      table.string('DisplayName', [100]).nullable();
      table.timestamp('CreatedDate').notNullable();
      table.timestamp('UpdatedDate').notNullable();
      table.boolean('IsDeleted').notNullable().defaultTo('false');
    })

Which gives me the following error :
Unhandled rejection Error: DECLARE PK_NAME VARCHAR(200); BEGIN  EXECUTE IMMEDIATE ('CREATE SEQUENCE "apiusers_seq"');  SELECT cols.column_name INTO PK_NAME  FROM all_constraints cons, all_cons_columns cols  WHERE cons.constraint_type = 'P'  AND cons.constraint_name = cols.constraint_name  AND cons.owner = cols.owner  AND cols.table_name = 'APIUsers';  execute immediate ('create or replace trigger "apiusers_autoinc_trg"  BEFORE INSERT on "APIUsers"  for each row  declare  checking number := 1;  begin    if (:new."' || PK_NAME || '" is null) then      while checking >= 1 loop        select "apiusers_seq".nextval into :new."' || PK_NAME || '" from dual;        select count("' || PK_NAME || '") into checking from "APIUsers"        where "' || PK_NAME || '" = :new."' || PK_NAME || '";      end loop;    end if;  end;'); END; - ORA-01422: exact fetch returns more than requested number of rows
ORA-06512: at line 1
    at Error (native)

The knex code creates a sequence for the increment of BigInteger Datatype by itself and that sequence is unable to function properly.

It works fine with MSSQL.

Can anyone please help?

elhigu:
Looks like the latest knex is producing the same kind of SQL https://runkit.com/embed/yrgl08kfsxpa

You can try to split query

```
[...]

PR #2750, opened 2018-08-02
Added owner constraint for auto increment related queries in Oracle

JaffParker:
Related issue/comment: [#2428](https://github.com/tgriesser/knex/issues/2428#issuecomment-410053616)

This is a very rough fix, I'm not even sure whether it would be fit to be merged or not really, but at least it gives an idea. To fix the problem described in my comment, I simply take the the username from the config and pass it to the functions that generate SQLs.

There's a couple of things that I didn't really take care of, so if you decide to merge, pay attention to them:

* I tried to look for usages of the functions in `trigger.js` file and found a few, but I feel like I still missed one or two
* I forgot to capitalize the owner in the query, but as well did not consider whether the user wants names capitalized or not in the config

These are the 2 points that better be double-checked, but the general idea works (I tested it in my project).

### DIFF
--- a/src/dialects/oracle/schema/columncompiler.js
+++ b/src/dialects/oracle/schema/columncompiler.js
@@ -21,7 +21,8 @@ assign(ColumnCompiler_Oracle.prototype, {
       const tableName = this.tableCompiler.tableNameRaw;
       const createTriggerSQL = Trigger.createAutoIncrementTrigger(
         this.client.logger,
-        tableName
+        tableName,
+        this.client.config.connection.user
       );
       this.pushQuery(createTriggerSQL);
     });
--- a/src/dialects/oracle/schema/trigger.js
+++ b/src/dialects/oracle/schema/trigger.js
@@ -1,7 +1,7 @@
 import * as utils from '../utils';
 
 const trigger = {
-  renameColumnTrigger: function(logger, tableName, columnName, to) {
+  renameColumnTrigger: function(logger, tableName, columnName, to, owner) {
     const triggerName = utils.generateCombinedName(
       logger,
       'autoinc_trg',
@@ -20,6 +20,7 @@ const trigger = {
       `    FROM all_constraints cons, all_cons_columns cols` +
       `    WHERE cons.constraint_type = 'P'` +
       `    AND cons.constraint_name = cols.constraint_name` +
+      `    AND cons.owner = ${owner}` +
       `    AND cons.owner = cols.owner` +
       `    AND cols.table_name = '${tableName}';` +
       `    IF ('${to}' = PK_NAME) THEN` +
@@ -43,7 +44,7 @@ const trigger = {
     );
   },
 
-  createAutoIncrementTrigger: function(logger, tableName) {
+  createAutoIncrementTrigger: function(logger, tableName, owner) {
     const triggerName = utils.generateCombinedName(
       logger,
       'autoinc_trg',
@@ -59,6 +60,7 @@ const trigger = {
       `  FROM all_constraints cons, all_cons_columns cols` +
       `  WHERE cons.constraint_type = 'P'` +
       `  AND cons.constraint_name = cols.constraint_name` +
+      `  AND cons.owner = ${owner}` +
       `  AND cons.owner = cols.owner` +
       `  AND cols.table_name = '${tableName}';` +
       `  execute immediate ('create or replace trigger "${triggerName}"` +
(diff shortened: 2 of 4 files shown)


src/dialects/oracle/schema/columncompiler.js
```diff
@@ -21,7 +21,8 @@ assign(ColumnCompiler_Oracle.prototype, {
       const tableName = this.tableCompiler.tableNameRaw;
       const createTriggerSQL = Trigger.createAutoIncrementTrigger(
         this.client.logger,
-        tableName
+        tableName,
+        this.client.config.connection.user
```
elhigu: Afaik it might be possible that there is no value in `config.connection.user`.

src/dialects/oracle/schema/trigger.js
```diff
@@ -20,6 +20,7 @@ const trigger = {
       `    FROM all_constraints cons, all_cons_columns cols` +
       `    WHERE cons.constraint_type = 'P'` +
       `    AND cons.constraint_name = cols.constraint_name` +
+      `    AND cons.owner = ${owner}` +
```
elhigu: Literal is not quoted at all

src/dialects/oracle/schema/trigger.js
```diff
@@ -59,6 +60,7 @@ const trigger = {
       `  FROM all_constraints cons, all_cons_columns cols` +
       `  WHERE cons.constraint_type = 'P'` +
       `  AND cons.constraint_name = cols.constraint_name` +
+      `  AND cons.owner = ${owner}` +
```
elhigu: literal is not quoted

src/dialects/oracle/schema/trigger.js
```diff
@@ -102,6 +104,7 @@ const trigger = {
       `    FROM all_constraints cons, all_cons_columns cols` +
       `    WHERE cons.constraint_type = 'P'` +
       `    AND cons.constraint_name = cols.constraint_name` +
+      `    AND cons.owner = ${owner}` +
```
elhigu: same here

elhigu:
Also I'm not sure yet if this is wanted functionality.

JaffParker:
Well, I described in the referenced issue the use case. It is indeed very specific, but I could fix the technicalities in this PR and it will stay in the code just as a nice little nod to people who face the same error as me. And it could check for the value being available in the config and only reference it if it is.

On the other hand, if that functionality isn't in place, I doubt you would lose users because this is the only library out there that works well with Oracle, so you kinda have the monopoly.

I'll fix the problems and you decide whether you want it or not.

kibertoad:
@JaffParker Are you still interested in finishing PR?

JaffParker:
Yeah, totally. I kinda got busy with work and eventually forgot the active PRs that I was making 😳 sorry about that. I'll finish it soon

etino:
I run into the same problem with a multi-user schema database on Oracle. 
The `migrate` command fails when there is already a table with another user in the same schema.
The problem is the same explained here https://github.com/knex/knex/issues/3145
I think that this merge request could resolve the issue.
Let me know if I can contribute in some way.

cjbland:
Four years later I'd like to ask that this be re-looked at, please. We are running into this issue as well in a multi-user schema database.

src/dialects/oracle/schema/columncompiler.js
cjbland: I agree with this statement. We use wallets to connect to Oracle via the `externalAuth: true` parameter in our connection.

src/dialects/oracle/schema/trigger.js
```diff
@@ -20,6 +20,7 @@ const trigger = {
       `    FROM all_constraints cons, all_cons_columns cols` +
```
cjbland, line 20: Another suggestion is to query from `user_constraints` and `user_cons_columns` instead of the "all" variants. This will limit the scope to the current user/schema only.

src/dialects/oracle/schema/trigger.js
```diff
@@ -59,6 +60,7 @@ const trigger = {
       `  FROM all_constraints cons, all_cons_columns cols` +
```
cjbland, line 60: Same comment as above, using `user_constraints` and `user_cons_columns` might be a better option.

src/dialects/oracle/schema/trigger.js
```diff
@@ -102,6 +104,7 @@ const trigger = {
       `    FROM all_constraints cons, all_cons_columns cols` +
```
cjbland, line 104: Same comment as above, using `user_constraints` and `user_cons_columns` might be a better option.

PR #2757, opened 2018-08-08
Use Datetime2 for MSSQL datetime + timestamp types

willmorgan:
Datetime2 is now the recommended column type for new date work: https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-2017

### DIFF
--- a/src/dialects/mssql/schema/columncompiler.js
+++ b/src/dialects/mssql/schema/columncompiler.js
@@ -60,9 +60,9 @@ assign(ColumnCompiler_MSSQL.prototype, {
 
   uuid: 'uniqueidentifier',
 
-  datetime: 'datetime',
+  datetime: 'datetime2',
 
-  timestamp: 'datetime',
+  timestamp: 'datetime2',
 
   bit(length) {
     if (length > 1) {
--- a/test/integration/schema/index.js
+++ b/test/integration/schema/index.js
@@ -355,7 +355,7 @@ module.exports = function(knex) {
               'create index "test_table_one_logins_index" on "test_table_one" ("logins")',
             ]);
             tester('mssql', [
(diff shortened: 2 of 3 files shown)


kibertoad:
Is it possible to have an integration test for this?

willmorgan:
> Is it possible to have an integration test for this?

Of course. I see that some tests are failing because of this. Not sure about the other failures though.

kibertoad:
@willmorgan Is this change compatible with older MSSQL versions?

willmorgan:
Hi @kibertoad - it's compatible with SQL server versions 2008 and above. So it's got 10 years of BC, at least 😉 

https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetime2-transact-sql?view=sql-server-2017

kibertoad:
LGTM.
@smorey2, what do you think?

smorey2:
Personally, I still use datetime in all of my applications, working with different versions of sql server, or integrations to other application system. Plus any of my previous knex work will already have datetime as the backing table field.

Lots of my t-sql code, and others in systems we integrate with, will still reference datetime instead of datetime2.

That being said, datetime2 is more efficient, and while not deprecated, it is recommended to use.

I can see it going ether way. I think it comes down how far back does the knex project want to support, SQL server 2005 official support ended, April 12, 2016, and SQL Server 2008 and 2008 R2 will end July 9, 2019.

kibertoad:
@smorey2 Considering that 0.14.x is pretty stable and the major theme of 0.15.x was dropping compatibility for unsupported Node.js versions, I would think that explicitly dropping support for unsupported SQL server versions makes sense.

smorey2:
Sounds good to me.

Sent from my iPhone

On Aug 13, 2018, at 3:16 PM, Igor Savin <notifications@github.com<mailto:notifications@github.com>> wrote:


@smorey2<https://github.com/smorey2> Considering that 0.14.x is pretty stable and the major theme of 0.15.x was dropping compatibility for unsupported Node.js versions, I would think that explicitly dropping support for unsupported SQL server versions makes sense.

PR #2773, opened 2018-08-19
Restore original PostgreSQL stream error logic

kibertoad:


kibertoad:
Fix following the conversation in #2638

I could really use some help writing a test for this: queryStream is not exposed so I can't mock any of its methods, and I don't know how to query for open connections.

kibertoad:
OK, this breaks stuff. Needs more in-depth investigation.

Issue #2756, opened 2018-08-08
Ignore duplicate variants of same migration (e.g. `.js`/`.ts`)


sgoll:
# Environment

Knex version: 0.15.2
Database + version: any
OS: any

# Feature discussion / request

## Use case

Code that implements migrations in TypeScript (or CoffeeScript or some other dialect that gets compiled to plain JavaScript) needs to be aware of different filename extensions in development VS production.

Even with the `loadExtensions` setting introduced in #2168, it is sometimes difficult to tell exactly when 
to use (only) the `.ts` file and when to use (only) the `.js` file.

One particular issue that I experience is this:
- During development, I use `ts-node` to run Knex migrations. I want that to use the `.ts` files.
- During testing, I rely on Jest and Wallaby.js. Jest sees only the `.ts` files. Wallaby.js sees the `.ts` files and `.js` files created just-in-time directly next to them. I want both to use the `.ts` files only.
- During production, I only have the `.js` files available in the migrations directory. Nonetheless, I want to be able to operate on the same database instance that I used during development, i.e. seamlessly run the dev version or the prod version of my application on any database.
- Ideally, I would use the same Knex configuration and not adjust more settings than necessary between dev and prod to make tests more reliable.

## Requested feature

Use `loadExtensions` to only load any given basename once, with the extension that first appears in the `loadExtensions` list.

Example: Suppose the migrations directory contains the following files:
- `001.ts` (A, original TypeScript), `001.js` (A, built file)
- `002.ts` (B, original TypeScript)
- `003.js` (C, built file)

We have three migrations (`001`, `002`, `003`). When `loadExtensions` is set to `[".ts", ".js"]`, Knex should only consider the following files to apply as migrations instead of all four files as is the case now:
- `001.ts` (`.ts` has priority over `.js` as it is listed before the latter in `loadExtensions`)
- `002.ts` (there is no other file for the `002` basename)
- `003.js` (ditto)

## API / Implementation proposal

Function `listAll()` in `src/migrate/migration-list-resolver.js` could be adjusted to only return the first instance of each migration with the same basename (i.e. migrations that only differ in their extension), with priority of extension given by their order in `loadExtensions` as described above.

In addition, it would be necessary to replace calls to lodash's `difference()` with `differenceWith()` so that extensions get ignored when comparing all available migrations to the ones that have already been applied and stored in the database (returned by `listAll()` and `listCompleted()`, respectively).

[...]

PR #2775, opened 2018-08-20
Introduced abstraction for getting migrations

JakeGinnivan:
This would allow a webpack migration source which is compatible with bundling. As an example:

``` typescript
const migrationConfig = config || {
    directory: path.resolve(__dirname, 'migrations'),
    migrationSource: new WebpackMigrationSource(require.context('./migrations', false)),
}

class WebpackMigrationSource {
    constructor(private migrationContext: __WebpackModuleApi.RequireContext) {}
    /**
   * Gets the migration names
   * @returns Promise<string[]>
   */
    getMigrations() {
        return Promise.resolve(this.migrationContext.keys())
    }

    getMigration(name: string) {
        return this.migrationContext(name)
    }
}
```

I can now use webpack and bundle my database migration project.

### DIFF
--- a/src/migrate/Migrator.js
+++ b/src/migrate/Migrator.js
@@ -25,7 +25,7 @@ import {
 } from './table-resolver';
 import { getSchemaBuilder } from './table-creator';
 import * as migrationListResolver from './migration-list-resolver';
-import FsMigrations from './sources/fs-migrations';
+import FsMigrations, { DEFAULT_LOAD_EXTENSIONS } from './sources/fs-migrations';
 
 function LockError(msg) {
   this.name = 'MigrationLocked';
@@ -36,7 +36,7 @@ inherits(LockError, Error);
 
 const CONFIG_DEFAULT = Object.freeze({
   extension: 'js',
-  loadExtensions: migrationListResolver.DEFAULT_LOAD_EXTENSIONS,
+  loadExtensions: DEFAULT_LOAD_EXTENSIONS,
   tableName: 'knex_migrations',
   schemaName: null,
   directory: './migrations',
@@ -114,10 +114,7 @@ export default class Migrator {
       getTable(this.knex, this.config.tableName, this.config.schemaName).select(
         '*'
       ),
-      migrationListResolver.listAll(
-        this.config.migrationSource,
-        this.config.loadExtensions
-      ),
+      migrationListResolver.listAll(this.config.migrationSource),
     ]).spread((db, code) => db.length - code.length);
   }
 
@@ -451,11 +448,14 @@ export default class Migrator {
     if (
       config &&
       !config.migrationSource &&
-      (config.directory || config.sortDirsSeparately !== undefined)
+      (config.directory ||
+        config.sortDirsSeparately !== undefined ||
+        config.loadExtensions)
     ) {
       newConfig.migrationSource = new FsMigrations(
         newConfig.directory,
-        newConfig.sortDirsSeparately
+        newConfig.sortDirsSeparately,
+        newConfig.loadExtensions
       );
     }
 
--- a/src/migrate/Migrator.js
+++ b/src/migrate/Migrator.js
@@ -25,6 +25,7 @@ import {
 } from './table-resolver';
 import { getSchemaBuilder } from './table-creator';
 import * as migrationListResolver from './migration-list-resolver';
+import { FsMigrations } from './sources/fs-migrations';
 
 function LockError(msg) {
   this.name = 'MigrationLocked';
@@ -60,7 +61,7 @@ export default class Migrator {
   latest(config) {
     this.config = this.setConfig(config);
     return migrationListResolver
-      .listAllAndCompleted(this.config, this.knex, this._absoluteConfigDir())
+      .listAllAndCompleted(this.config, this.knex)
       .tap(validateMigrationList)
       .spread((all, completed) => {
         const migrations = getNewMigrations(all, completed);
@@ -68,9 +69,9 @@ export default class Migrator {
         const transactionForAll =
           !this.config.disableTransactions &&
           isEmpty(
-            filter(migrations, (migration) => {
-              const migrationContent = require(resolveMigrationPath(migration));
-              return !this._useTransaction(migrationContent);
+            filter(migrations, (name) => {
+              const migration = this.config.migrationSource.getMigration(name);
+              return !this._useTransaction(migration);
             })
           );
 
@@ -89,7 +90,7 @@ export default class Migrator {
     return Promise.try(() => {
       this.config = this.setConfig(config);
       return migrationListResolver
-        .listAllAndCompleted(this.config, this.knex, this._absoluteConfigDir())
+        .listAllAndCompleted(this.config, this.knex)
         .tap(validateMigrationList)
         .then((val) => this._getLastBatch(val))
         .then((migrations) => {
@@ -106,9 +107,8 @@ export default class Migrator {
         '*'
       ),
       migrationListResolver.listAll(
-        this._absoluteConfigDir(),
-        this.config.loadExtensions,
-        this.config.sortDirsSeparately
+        this.config.migrationSource,
+        this.config.loadExtensions
       ),
     ]).spread((db, code) => db.length - code.length);
   }
@@ -150,7 +150,7 @@ export default class Migrator {
   // Ensures a folder for the migrations exist, dependent on the migration
   // config settings.
   _ensureFolder() {
-    const dirs = this._absoluteConfigDir();
+    const dirs = this._absoluteConfigDirs();
 
     const promises = dirs.map((dir) => {
       return Promise.promisify(fs.stat, { context: fs })(dir).catch(() =>
@@ -277,7 +277,7 @@ export default class Migrator {
     ) {
       throw new Error(
         `Invalid migration: ${
-          migration.file
+          migration
         } must have both an up and down function`
       );
     }
@@ -299,7 +299,7 @@ export default class Migrator {
   // passing any `variables` given in the config to the template.
   _writeNewMigration(name, tmpl) {
     const { config } = this;
-    const dirs = this._absoluteConfigDir();
+    const dirs = this._absoluteConfigDirs();
     const dir = dirs.slice(-1)[0]; // Get last specified directory
 
     if (name[0] === '-') name = name.slice(1);
@@ -313,7 +313,7 @@ export default class Migrator {
 
   // Get the last batch of migrations, by name, ordered by insert id in reverse
   // order.
-  _getLastBatch([allMigrations, doneMigrations]) {
+  _getLastBatch([allMigrations]) {
     const { tableName, schemaName } = this.config;
     return getTable(this.knex, tableName, schemaName)
       .where('batch', function(qb) {
@@ -322,7 +322,7 @@ export default class Migrator {
       .orderBy('id', 'desc')
       .map((migration) => {
         return allMigrations.find((entry) => {
-          return entry.file === migration.name;
+          return entry === migration.name;
         });
       });
   }
@@ -355,10 +355,9 @@ export default class Migrator {
     let current = Promise.bind({ failed: false, failedOn: 0 });
     const log = [];
     each(migrations, (migration) => {
-      const directory = migration.directory;
-      const name = migration.file;
+      const name = migration;
       this._activeMigration.fileName = name;
-      const migrationContent = require(directory + '/' + name);
+      const migrationContent = this.config.migrationSource.getMigration(name);
 
       // We're going to run each of the migrations in the current "up".
       current = current
@@ -376,7 +375,7 @@ export default class Migrator {
           );
         })
         .then(() => {
-          log.push(path.join(directory, name));
+          log.push(name);
           if (direction === 'up') {
             return trxOrKnex.into(getTableName(tableName, schemaName)).insert({
               name,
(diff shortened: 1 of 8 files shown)


JakeGinnivan:
FYI @elhigu

JakeGinnivan:
Am actually not sure how the coverage dropped so much.. Seems odd

JakeGinnivan:
Just while I think of it, some other reason you may want this:

1) Custom folder structure, for example https://github.com/tgriesser/knex/pull/2735 could be implemented as a migration source without a pull request
2) I have subclassed Migrator to FilterableMigrator which allows me to have `_PRODUCT` in a migration, then it only gets run for that product (we have an API which is the same codebase and has different seed data per product)
3) You could actually support the typescript / other language support with specific migration sources which compile the migration then return the compiled output.

All of the above and whatever other requirements someone may have around where migrations come from, how to filter them out, and the order they run in can be controlled by the consumer.

It would be awesome to get an idea if this will be merged. It also makes it easy to work around issues like https://github.com/tgriesser/knex/issues/1922

JakeGinnivan:
Will get it done, just wanted to make sure you were happy with the idea in general.

src/migrate/sources/fs-migrations.js
```diff
@@ -0,0 +1,65 @@
+import fs from 'fs';
+import path from 'path';
+import Promise from 'bluebird';
+import { sortBy } from 'lodash';
+
+const readDirAsync = Promise.promisify(fs.readdir, { context: fs });
+
+export class FsMigrations {
```
kibertoad: Why not export it as a default?

src/migrate/Migrator.js
```diff
@@ -419,15 +438,22 @@ export default class Migrator {
   }
 
   setConfig(config) {
-    return assign({}, CONFIG_DEFAULT, this.config || {}, config);
+    const newConfig = assign({}, CONFIG_DEFAULT, this.config || {}, config);
```
kibertoad: Object.assign could be used instead of a Lodash one here.

src/migrate/Migrator.js
```diff
@@ -419,15 +438,22 @@ export default class Migrator {
...
+    if ((config && !config.migrationSource) || !newConfig.migrationSource) {
```
kibertoad: Why checking just newConfig is not enough?

src/migrate/Migrator.js
kibertoad: Are you sure about naming? There is not much new about it, it is mostly a composite or combined one.

src/migrate/Migrator.js
JakeGinnivan: Because of the way tests use this function, I only want to init the migrationSource if it is not on the current config (if there is one) and it's also not specified in the new config.

Just checking new config meant that different tests calling setConfig would not apply.

src/migrate/Migrator.js
JakeGinnivan: It is going to be the new config was my thinking?

src/migrate/sources/fs-migrations.js
JakeGinnivan: Personal preference coming from TypeScript. https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html

Happy to change if that is the preference.

src/migrate/sources/fs-migrations.js
```diff
@@ -0,0 +1,65 @@
+import fs from 'fs';
+import path from 'path';
+import Promise from 'bluebird';
```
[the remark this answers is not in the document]
JakeGinnivan: When are you considering dropping node 6 support? If we have node 8 only then can just use native promisify too.

Will switch.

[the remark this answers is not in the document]
kibertoad: Unfortunately, Node 6 is still about 25% of all production installations, so chances for dropping it in the nearby future are not that great. It is also officially supported till 2019 April.
(I wish we could drop Node 6 and use async/await freely in tests, but alas)

src/migrate/sources/fs-migrations.js
kibertoad: It result in slightly less redundant imports, 
Quick search on project shows that there are multiple cases of "export default class" and none of "export class", so it does seem to be the project convention.

src/migrate/Migrator.js
kibertoad: Not sure if I understand. After you call `assign({}, CONFIG_DEFAULT, this.config || {}, config);`, doesn't newConfig already include everything that config has?

src/migrate/sources/fs-migrations.js
JakeGinnivan: Will change

src/migrate/Migrator.js
JakeGinnivan: Let's say the first test does not specify a migrationSource, but sets the directory. The migrationSource will capture the directory, so when setConfig is called with a different directory, but no migrationSource the current migrationSource (with the old directory) is on this.config.

Basically, the tests can't change the migration directory.

kibertoad: Looks like we are talking about different things :)
Under which circumstances would "!config.migrationSource" check pass but " !newConfig.migrationSource" would not? (double check seems redundant to me is what I'm saying as one config is derived from the other)

JakeGinnivan: If you were just to check newConfig.migrationSource, this would happen

```
const migrator = new Migrator({ directory: '/foo' })
migrator.config.migrationSource.directory === '/foo'

migrator.setConfig({ directory: '/bar' })
// !config.migrationSource === true
migrator.config.migrationSource.directory === '/foo'
```

migrationSource can be set by the user, but also falls back. If the user set a migration source the first time, we don't want to remove it. Will see if I can clean up somehow. But without the extra logic, heaps of tests fail because they can't change the directory.

JakeGinnivan: Hrrmm, looking at this fresh today. I am not sure exactly.

I *think* I was thinking something along the lines of, if you specify a migration source initially, which is not the default, I don't want it overridden if you call setConfig again.

Either way seems brittle, going to try and figure a better way

JakeGinnivan: Have re-written logic so it's clearer

JakeGinnivan:
I think #2756 and #1922 can be solved by doing this (once this PR lands):

```
return knex.migrate.latest({
  migrationSource: new FsMigrations('./directory', false, ['.js']);
});
```

Or something similar, and it will be simple enough to create a migration source if the ootb ones are not configurable enough.

JakeGinnivan:
Down the track we could also move language compilation into different sources. So you could have a TypeScriptFsMigrationSource which does the compilation.

src/migrate/Migrator.js
```diff
@@ -419,15 +440,42 @@ export default class Migrator {
...
+    ) {
+      newConfig.migrationSource = new FsMigrations(
+        newConfig.directory,
+        newConfig.sortDirsSeparately,
+        newConfig.loadExtensions
+      );
+    }
+
+    // Ensure a migrationSource is set
+    if (!newConfig.migrationSource) {
```
kibertoad: This code looks quite redundant. What's the added benefit of Calling F1 with A, B, C if A || B || C set, or calling F1 with A, B if neither are set, if these are full equivalents, as you would simply be passing undefineds that will get resolved to default value in either case anyway?

src/migrate/Migrator.js
```diff
@@ -419,15 +440,42 @@ export default class Migrator {
...
+    // are used
+    if (
+      config &&
+      !config.migrationSource &&
+      (config.directory ||
+        config.sortDirsSeparately !== undefined ||
+        config.loadExtensions)
+    ) {
+      newConfig.migrationSource = new FsMigrations(
+        newConfig.directory,
```
kibertoad: You are checking for values being set on config and then using values from newConfig. Even if this is still going to achieve same purpose, it reads confusingly. Also I still fail to see benefit of the check, considering that the outcome is always the same, we end up creating FsMigrations once with params from config.

src/migrate/Migrator.js
```diff
@@ -419,15 +440,42 @@ export default class Migrator {
...
   setConfig(config) {
-    return assign({}, CONFIG_DEFAULT, this.config || {}, config);
+    const newConfig = assign({}, CONFIG_DEFAULT, this.config || {}, config);
+
+    // If migrationSource is not specified, but fs related config has been specified
+    // recreate the FsMigration source so the new filesystem related settings
+    // are used
+    if (
+      config &&
+      !config.migrationSource &&
```
kibertoad: I think this is the only useful check and the logic should be as following: "if newConfig.migrationSource -> noop, else newConfig.migrationSource = new FSMigrationSource". Everything else doesn't seem to result in any meaningful branches.

JakeGinnivan: But then this wouldn't work

```
const migrator = new Migrator({ directory: './foo' })
migrator.latest({ directory: './bar' })
```

Because the ctor would initialise `migrationSource`, then the call to latest would check to see if newConfig.migrationSource is set. It is, so no-op. Leaving the FsMigrationSource looking at `./foo` rather than `./bar`.

src/migrate/Migrator.js
JakeGinnivan: config is the values the user has passed in, newConfig is the merged config from existing, defaults and then user values.

The additional checks are ensuring the intention of the user works properly.

src/migrate/Migrator.js
kibertoad: Ah, OK, now your logic makes sense. Could you clarify comment a little bit to make it more explicit that you are accounting for the case where migrationSource configuration has changed between previous config and newly passed one?

JakeGinnivan: I could add some specific tests around it. Was confusing to write too :P

JakeGinnivan: Added some comments to clarify

JakeGinnivan:
Hey @kibertoad anything else?

src/migrate/sources/fs-migrations.js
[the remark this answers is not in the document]
elhigu: we can drop it in next year April I suppose https://github.com/nodejs/Release

JakeGinnivan:
Thanks @elhigu 

After it gets merged, do you think you would be able to publish a tagged `@next` release on NPM or something if the next release will be a while due to other breaking changes? Will just save me building and publishing our own copy into nexus.

src/migrate/Migrator.js
elhigu: This method indeed seems unnecessarily complicated... couldn't we just always do

```js
if (!newConfig.migrationSource) {
    newConfig.migrationSource = new FsMigrations(
        newConfig.directory,
        newConfig.sortDirsSeparately,
        newConfig.loadExtensions
    );
}
```

?

elhigu: Ah yeah, I think now I get it... anyways, maybe it would be just better to do first

```js
if (config && (config.directory ||
        config.sortDirsSeparately !== undefined ||
        config.loadExtensions) {
   newConfig.migrationSource = null;
}
```

Which would signal the intention better that if user has passed file migrator configuration attributes to parameter, then we need to override the migration source.

JakeGinnivan: Ah that is actually quite a nice way of doing it. Invalidate the current if one of the config settings change, then set if not set. Will update!

you-fail-me:
This would be really amazing, I'm struggling to package knex migrations service with webpack to run in aws lambda and the current tight bond with file system is very inconvenient for my case

JakeGinnivan:
@elhigu I have made your changes, sorry about the timing, had a busy week

JakeGinnivan:
Actually, made them in https://github.com/tgriesser/knex/pull/2786

I can backport them if you want, but if you like the initial look of that PR I am happy to fix it up after this merges and land it separately?

kibertoad:
@JakeGinnivan Could you give me write permissions on your fork so that I could backport it myself? Would be easier to have it in one place.

JakeGinnivan:
@kibertoad I have backported and also added you in case there is anything else. I am in Australia so I am unsure how much overlap time wise. But will keep an eye on comments today and push anything up quickly

JakeGinnivan:
Struggling to find why some tests are timing out. Will try to get it all sorted, but any help is welcome :)

elhigu:
Looks like the test fails are coming right now from missing method mergeConfig

kibertoad:
@elhigu That solved it for all Nodes but the 10th one. Huh?..

kibertoad:
Fun fact - it is Node 10.10.0-specific, passes fine with earlier ones.

kibertoad:
@elhigu This block times out on Node 10.10.0:
```
      return readDirAsync(absoluteDir).then(
(files) => ({ //This then block is never entered
        files,
        configDir,
        absoluteDir,
      }));
    });
```
If replaced with fs.readdirSync - works perfectly
Since our usage of readDirAsync looks correct, maybe we should lock CI Node version at 10.9.0 for the time being? This looks like an issue in Node and I suspect that our current master is failing as well.

kibertoad:
Update: seems to be a compatibility issue between mock-fs and Node.js 10.10.0 rather than just a bug in Node 10.10.0.

kibertoad:
@elhigu Aaaaaand after mock-fs have fixed its compatibility problem we are back online. Everything green, everything passes, everything ready to merge!

JakeGinnivan:
Awesome! Let me know if you need anything from me

elhigu:
Thank you very much for your co-operation :D

JakeGinnivan:
No problems, any chance of a @next tagged release or something ? :)

elhigu:
yes, lots of chances 👍

Issue #2758, opened 2018-08-08
Feature Request: Lowercase dialect category inputs from user


AstroGD:
# Environment

Knex version: 0.15.2
Database + version: MySQL
OS: Linux Debian

Select applicable tempalate from below.
If issue is about oracledb support tag @ atiertant. For MSSql tag @ smorey2 .
Rest of dialects doesn't need tags.

# Bug

If you try to install Knex on Linux Debian the following error occurres: Error: Cannot find module './dialects/MySQL/index.js'

This due to the fact, that paths are case-sensitive and the path must be /dialects/MySQL/index.js

I highly reccomend you to either change the path in code or rename the folder, as an installation of knex on debian is currently impossible without changing the folder manually.

AstroGD:
I typed `npm install knex` and then run my program. My program then crashed with an unhandled promise rejection because the file was not found

AstroGD:
Sorry i was a bit misleading there, the error does not occurr on installation but on running code using knex

ricardograca:
No idea how you're getting that error since that directory has a lowercase name on this repository and after installation. Can you either explain step by step how to reproduce this with a clean install or provide the necessary code to do so?

AstroGD:
Another mistake from me: The Folder is named "mysql" and needs to be named "MySQL"

AstroGD:
![grafik](https://user-images.githubusercontent.com/35271072/43949810-e0e71ecc-9c8e-11e8-952d-50f4099c8e8a.png)
This is the error after a clean installation

AstroGD:
Step by Step:

- Go into an empty folder
- type npm install knex
- Wait till finished
- put in any js file using knex
- run the js file
- get the error

AstroGD:
Problem: When installing knex the folder is named "mysql" but it should be "MySQL"

ricardograca:
I'm using knex on a project right now with MySQL and not seeing that error. Can you provide the "any js file using knex" that you're using?

AstroGD:
Ill provide it asap

tgriesser:
The folder name is determined by whatever you provide as the `config.client` or `config.dialect`

https://github.com/tgriesser/knex/blob/2183a278268dc7aedf5edeccecb99176b9827f27/src/index.js#L29-L31

I'm guessing you have that as 'MySQL' instead of 'mysql', just fix that and it'll work.

ricardograca:
[...]

PR #2776, opened 2018-08-20
#2758: Implement fail-fast logic for dialect resolution

kibertoad:
refs https://github.com/tgriesser/knex/issues/2758


src/index.js
```diff
@@ -1,73 +1,3 @@
-import Raw from './raw';
-import Client from './client';
+const Knex = require('./knex');
```
kibertoad: Rationale behind this change: it it very counter-intuitive to write tests for a file named "index.js" which is in general supposed to be a catalogue and do not contain logic on its own.

src/knex.js
```diff
@@ -0,0 +1,83 @@
+import Raw from './raw';
+import Client from './client';
+
+import makeKnex from './util/make-knex';
+import parseConnection from './util/parse-connection';
+import { SUPPORTED_CLIENTS, CLIENT_ALIASES } from './constants';
+
+export default function Knex(config) {
+  // If config is string, try to parse it
```
kibertoad: added comment with explanation for each block, please verify the accuracy of them

src/knex.js
```diff
@@ -0,0 +1,83 @@
...
+  if (typeof config === 'string') {
+    const parsedConfig = Object.assign(parseConnection(config), arguments[2]);
```
kibertoad: _.assign replaced with Object.assign in this file since it is supported since Node 4.

src/knex.js
```diff
@@ -0,0 +1,83 @@
...
+/* eslint no-console:0 */
+
+Object.defineProperties(Knex, {
+  VERSION: {
+    get() {
+      console.warn(
+        'Knex.VERSION is deprecated, you can get the module version' +
+          "by running require('knex/package').version"
+      );
+      return '0.12.6';
```
ricardograca: Looks like this return value hasn't been updated in some time :laughing: Maybe it's time to remove this deprecation warning? I know it's not related to the rest of the work done here, so feel free to ignore this comment.

kibertoad: So you suggest to remove entire
```
VERSION: {
    get() {
      console.warn(
        'Knex.VERSION is deprecated, you can get the module version' +
          "by running require('knex/package').version"
      );
      return '0.12.6';
}
```

block? Sure, I can do that.

ricardograca: Yes. The message has been there for almost 2 years now. I think that's enough time for a deprecation warning.

kibertoad:
@ricardograca Did the change, anything else?..

ricardograca:
Cool! I'll take a look at the rest of changes later today.

src/constants.js
```diff
@@ -0,0 +1,39 @@
...
+  postgresql: 'postgres',
+  sqlite: 'sqlite3',
+});
+
+const SUPPORTED_CLIENTS = Object.freeze(
+  [
+    'mssql',
+    'mysql',
+    'mysql2',
+    'oracle',
```
elhigu: I  think that oracle is not actually supported by knex. It is just used by oracledb, which is supported.

src/client.js
```diff
@@ -38,7 +39,8 @@ function Client(config = {}) {
   //Client is a required field, so throw error if it's not supplied.
   //If 'this.dialect' is set, then this is a 'super()' call, in which case
   //'client' does not have to be set as it's already assigned on the client prototype.
-  if (!this.config.client && !this.dialect) {
+  const dbClient = this.config.client || this.dialect;
```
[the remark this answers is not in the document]
kibertoad: Would you like me to log a deprecation warning when this.dialect is set?

src/constants.js
kibertoad: Thanks, will remove

elhigu: afaik. it is not tested either

src/index.js
```diff
@@ -1,73 +1,3 @@
...
-/* eslint no-console:0 */
-
-Object.defineProperties(Knex, {
-  VERSION: {
-    get() {
-      console.warn(
-        'Knex.VERSION is deprecated, you can get the module version' +
-          "by running require('knex/package').version"
-      );
-      return '0.12.6';
```
elhigu: Was this just dropped in this PR? It's hard to follow all the changes after renaming... I need to diff the files locally. If this was dropped, it must be notified in changelog as breaking change.

kibertoad: I've added comments in the PR for all the changes after renaming; this one was done after @ricardograca comment. If you are ok with this change, I'll add it to changelog.

elhigu: Yep, I'm fine with it. I'll check out diff locally later on.

src/client.js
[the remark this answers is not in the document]
elhigu: That is a good idea, but we should discuss about it with other contributors too if they agree that alias is not wanted. My opinion is based on some previous problems with testing all these aliases and many times tests has been depending on .client and then they have failed when .dialect has been used by some other driver setup...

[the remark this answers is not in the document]
elhigu: @tgriesser @wubzz maybe has an opinion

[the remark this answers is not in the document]
wubzz: I think it's a good idea to remove "dialect" so long as the knex internals are also adjusted to reflect this. For example `src/dialects` -> `src/clients`.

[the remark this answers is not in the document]
kibertoad: Considering that there was no deprecation period with warnings prior to that, should we really just drop it immediately?

kibertoad:
@elhigu Changes addressed.

elhigu:
Looks like tests needs still some fixing so that they stop using .dialect too. Now at least custom logger tests are failing because of unexpected deprecation warnings.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ### Breaking Changes:
 
 - Use datetime2 for MSSQL datetime + timestamp types. This change is incompatible with MSSQL older than 2008 #2757
+- Knex.VERSION() method was removed, run "require('knex/package').version" instead.
```
elhigu: Please add also #2776 in the end of line so that this PR is referenced

kibertoad:
@elhigu Fixed!

kibertoad:
Aaaand it's green again.

kibertoad:
@elhigu Is anything currently missing?

elhigu:
Looks good to me, thanks!

Issue #2014, opened 2017-04-17
Pass an argument to knex.migrate.latest()


mwpastore:
I'm running Knex migrations programmatically and I'd like to pass an argument to `knex.migrate.latest({})` at runtime. I see that the configurable object passed to `#latest()` gets merged into the migration configuration, but this configuration does not seem to be available from within an up/down function. 

- Can we pass the config as an extra argument to up/down?
- Can we bind up/down to the migration's `this` at call time to make `this.config` available?
- Is there another option I'm missing?

amagdas:
I need this missing feature as well.

vincenttheeten:
I'd like to see the same thing.

I was hoping to store some kind of `tenantId` on the config object, so I can run migrations for specific tenants in my multi-tenant setup; e.g. create a table under the right (postgres) schema per tenant.

I saw version 0.14.5 introduced some improvements for schema names and migrations.
Unless I'm missing something, this only uses the schema in `schemaName` to make the migrations table, not scope the `up` or `down` migration calls.

kibertoad:
@vincenttheeten Can't you do pretty much this by using different changelog table names for different tenants and then passing different folders as migration sources?

elhigu:
Isnt it visible in `knex.client.config` either?

bertho-zero:
Personally I use `const { schemas } = knex.client.config` in my migrations.

elhigu:
Fixed by #2802

PR #2778, opened 2018-08-21
Pass migration config object as an argument to the migrations themsel…

dofrisec:
…ves, to allow custom parameters to be set for migrations. (#2014)

### DIFF
--- a/src/migrate/index.js
+++ b/src/migrate/index.js
@@ -352,11 +352,11 @@ export default class Migrator {
       current = current
         .then(() => {
           if (!trx && this._useTransaction(migration, disableTransactions)) {
-            return this._transaction(migration, direction, name);
+            return this._transaction(migration, direction, name, this.config);
           }
           return warnPromise(
             this.knex,
-            migration[direction](trxOrKnex, Promise),
+            migration[direction](trxOrKnex, Promise, this.config),
             name
           );
         })
@@ -381,11 +381,11 @@ export default class Migrator {
     return current.thenReturn([batchNo, log]);
   }
 
-  _transaction(migration, direction, name) {
+  _transaction(migration, direction, name, config) {
     return this.knex.transaction((trx) => {
       return warnPromise(
         this.knex,
-        migration[direction](trx, Promise),
+        migration[direction](trx, Promise, config),
         name,
         () => {
           trx.commit();
--- a/src/migrate/stub/coffee.stub
+++ b/src/migrate/stub/coffee.stub
@@ -1,5 +1,5 @@
 
-exports.up = (knex, Promise) ->
+exports.up = (knex, Promise, config) ->
   <% if (d.tableName) { %>
   knex.schema.createTable "<%= d.tableName %>", (t) ->
     t.increments()
@@ -7,7 +7,7 @@ exports.up = (knex, Promise) ->
   <% } %>
 
 
-exports.down = (knex, Promise) ->
+exports.down = (knex, Promise, config) ->
   <% if (d.tableName) { %>
   knex.schema.dropTable "<%= d.tableName %>"
   <% } %>
--- a/src/migrate/stub/eg.stub
+++ b/src/migrate/stub/eg.stub
@@ -1,6 +1,6 @@
 provide: up, down
 
-up = (knex, Promise) ->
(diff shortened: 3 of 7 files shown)


dofrisec:
Associated documentation update:
https://github.com/knex/documentation/pull/130

dofrisec:
I updated the migration stubs that were obvious, all except the typescript stub.
I'm not sure what to do with the typescript migration stub, since I'm not familiar with typescript.

kibertoad:
This would also be helpful for the https://github.com/tgriesser/knex/issues/2799 use-case. @elhigu Do you approve of this idea in general? There's a minor conflict to be resolved before this could be merged, but at this point it's unclear if it's worth doing so.

kibertoad:
@elhigu I think all use-cases are going to be about parametrizing migration execution from the outside. If these should be set through knex instance itself, I think we should define a recommended way to do it (e. g. have precreated knex.locals empty object for population) and document it. If you think that this is a better way, I can do both the code and the documentation change.

elhigu:
> @elhigu I think all use-cases are going to be about parametrizing migration execution from the outside. If these should be set through knex instance itself, I think we should define a recommended way to do it (e. g. have precreated knex.locals empty object for population) and document it. If you think that this is a better way, I can do both the code and the documentation change.

I like something like knex.locals or even more descriptive knex.userParams more. That would be more versatile solution that could work for more things than just migrations and initial values for those already in knex configuration.

dofrisec:
Wouldn't that have race conditions?

> > @elhigu I think all use-cases are going to be about parametrizing migration execution from the outside. If these should be set through knex instance itself, I think we should define a recommended way to do it (e. g. have precreated knex.locals empty object for population) and document it. If you think that this is a better way, I can do both the code and the documentation change.
> 
> I like something like knex.locals or even more descriptive knex.userParams more. That would be more versatile solution that could work for more things than just migrations. Initial attributes for that object could be given already in knex configuration.

Wouldn't that have race conditions?

elhigu:
> > > @elhigu I think all use-cases are going to be about parametrizing migration execution from the outside. If these should be set through knex instance itself, I think we should define a recommended way to do it (e. g. have precreated knex.locals empty object for population) and document it. If you think that this is a better way, I can do both the code and the documentation change.
> > 
> > 
> > I like something like knex.locals or even more descriptive knex.userParams more. That would be more versatile solution that could work for more things than just migrations. Initial attributes for that object could be given already in knex configuration.
> 
> Wouldn't that have race conditions?

Not that I can think of. Why would it?

kibertoad:
Imagine typicaly multi-tenancy scenario: you have independent schemas that can be migrated totally independently. You run their migrations in the loop, setting params on same knex instance. It is totally possible that you will run start migrations for schema 2 while only 77 out of 100 migrations for schema 1 passed, then migration 78 for schema 1 will suddently be executed with params for schema 2.
You can obviously await till schema 1 is done before starting with schema 2, but this means we have more limitations and gotchas that we could have with better isolation (e. g. passing params to migration itself rather than knex).

elhigu:
> Imagine typical multi-tenancy scenario: you have independent schemas that can be migrated totally independently. You run their migrations in the loop, setting params on same knex instance. It is totally possible that you will run start migrations for schema 2 while only 77 out of 100 migrations for schema 1 passed, then migration 78 for schema 1 will suddenly be executed with params for schema 2.
> You can obviously await till schema 1 is done before starting with schema 2, but this means we have more limitations and gotchas that we could have with better isolation (e. g. passing params to migration itself rather than knex).

Ah right, that use case where migrations are ran like that. Yep, something should be thought to address that. 

I was thinking at some point something like knex.userParams({}) which would return kind of instance of knex, using the same pool, but there would be different user parameter set assigned to that "instance".

I'm not sure at all what would be the best solution and API (I know only that I don't want it to be specific functionlity for migrations only).

kibertoad:
Ah, that might work - will try to update existing PR to use the kind of implementation that you are suggesting.

kennedyvardo:
Is that option of passing parameters already implemented?

elhigu:
Yes. https://knexjs.org/#Installation-parametrization

Issue #1588, opened 2016-07-22
how to configure to use the standard ES6 Promise?


c0b:
when running with node v6, is there a way to configure to use the standard ES6 Promise only (get rid of bluebird dependency)?

for example in promise-queue package, it defaults to use whatever is globally available Promise,

https://www.npmjs.com/package/promise-queue

or user can explicitly configure it by `Queue.configure(require('vow').Promise);`

So can this package implement a similar strategy?

johanneslumpe:
Curious: you want to replace a faster library with a slower built-in? On the client it wouldn't matter but in node speed is a factor. What's your reasoning for this?

elhigu:
@johanneslumpe It is factor in some applications, with `knex` I highly doubt that used `Promise` library has any significant effect to performance. It has been discussed that we should write all `Promise` code to use only A+ APIs so that `bluebird` wouldn't be needed. 

After that it should be easy to override, which Promise library is used.

wubzz:
I agree it feels pointless in a library such as knex. Given that knex's codebase already imports its own [promise.js](https://github.com/tgriesser/knex/blob/master/src/promise.js) file internally, it would technically still be very easy to implement, **provided the API is the same** *(which I'm not sure it is?)*. In this file one could default to global promises, else require configured library, or something like that.

c0b:
it's all about giving users a choice; for Node V6+ users, give a choice of managing 3rd party dependencies as minimum as possible.

> Curious: you want to replace a faster library with a slower built-in?

what you're saying might be true in the past, but how about now or 6 months later? (I did some search and can only find out some 2015 benchmarks, if you have more recent comparison, please post some links)
I believe the whole point of making Promise the ES6 standard is for people to easily use Promise, not having to rely on a 3rd party library, and Node or V8's core team can't be blind of the performance difference for ever, as long as the two projects' open source licenses are compatible, they can even borrow some code; or just give them some time I believe the builtin Promise can be faster and better.

c0b:
see Amazon's AWS-SDK: it also default to use whatever Promise is globally available; while give user a choice to configure the favorite Promise library as well

http://docs.aws.amazon.com/AWSJavaScriptSDK/guide/node-making-requests.html#Support_for_Promises

it's all about **choices**

wubzz:
[...]

PR #2790, opened 2018-08-29
Factors bluebird out of the cli module.

qubyte:
Begins to address #1588.

Also of note:
 - The `readFile` and `writeFile` functions I've defined can be replaced with `util.promisify` wrapped `fs.readFile` and `fs.writeFile` any time after Node 6 is no longer supported.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -1,15 +1,38 @@
 #!/usr/bin/env node
 /* eslint no-console:0, no-var:0 */
 var Liftoff = require('liftoff');
-var Promise = require('bluebird');
 var interpret = require('interpret');
 var path = require('path');
 var chalk = require('chalk');
 var tildify = require('tildify');
 var commander = require('commander');
 var argv = require('minimist')(process.argv.slice(2));
-var fs = Promise.promisifyAll(require('fs'));
+var fs = require('fs');
 var cliPkg = require('../package');
+
+function readFile(path) {
+  return new Promise(function(resolve, reject) {
+    fs.readFile(path, function(err, content) {
+      if (err) {
+        reject(err);
+      } else {
+        resolve(content);
+      }
+    });
+  });
+}
+
+function writeFile(path, content) {
+  return new Promise(function(resolve, reject) {
+    fs.writeFile(path, content, function(err) {
+      if (err) {
+        reject(err);
+      } else {
+        resolve();
+      }
+    });
+  });
+}
 
 function exit(text) {
   if (text instanceof Error) {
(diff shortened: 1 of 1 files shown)


bin/cli.js
```diff
@@ -1,16 +1,39 @@
...
 var chalk = require('chalk');
 var tildify = require('tildify');
 var commander = require('commander');
 var argv = require('minimist')(process.argv.slice(2));
-var fs = Promise.promisifyAll(require('fs'));
+var fs = require('fs');
 var cliPkg = require('../package');
 
+function readFile(path) {
+  return new Promise(function(resolve, reject) {
```
kibertoad: why not just use https://www.npmjs.com/package/es6-promisify?

qubyte: That's an option. I was hoping to avoid adding dependencies, but that package has none of its own so the footprint is small.

kibertoad: Promisifying everything by hand is very noisy. Alternatively we could copy promisification code into some internal util and use that to avoid external dependencies - and remove it after we drop Node 6 support.
@elhigu What do you think?

kibertoad: But if it's left as-is, more work would be needed to start using native promisification, as every single block that is like this would have to be rewritten. If es6-promisify is used, change is as trivial as switching the import.

elhigu: That's other discussion when bluebird promises should be removed completely and how. 

Yeah I'm not promoting making this kind of promisification everywhere if there are many places where it is needed and if complete codebase is converted to not use bluebird (even though I'm not sure if there even is many places where that has been used in codebase).

qubyte: There's only one more instance of `promisifyAll`, but it looks like 13 of `promisify`. I'm not averse to using a library given that number.

qubyte: > That's other discussion when bluebird promises should be removed completely and how.

The module changed in this PR is a special case. Most others will have to retain bluebird for now, since code which consumes knex may assume that promises returned are bluebird ones.

I suggest first factoring out all use of methods which are not in common with vanilla promises, and then both removing bluebird and allowing a custom promise to be parsed in (see #1588). This'll mean a major version bump, but will provide an upgrade path for those who are bound to bluebird.

elhigu:
Now that I think of this dropping bluebird from here is actually harmful instead of helpful.

This change would need another round of cleanup when node 6 support is dropped and since bluebird dependency is already dropped from here it will be harder to remember to clean this up later on.

So benefit of this change is now pretty much zero and drawbacks are that we might forget to clean this up later on.

I'd rather fix this and all other places where bluebird is used at the same time when Bluebird dep is remove or after we drop support for node 6 (next April).

I'll try to open discussion again about schedule when to go forward with dropping bluebird.

qubyte:
Unfortunate, but understood. I will return next April.

elhigu:
Indeed. I'm sorry for your trouble.

Issue #2799, opened 2018-09-04
Migrate into Postgres Database Schema other than 'public'


mashaalmemon:
# Environment

Knex version: 0.15.2
Database + version: postgres 9.5.3
OS: mac

# Questions about how to use knex

Is there a way to run a migration and have it execute inside a postgres SCHEMA other than 'public'? Note I am running the migration programatically, not through the CLI.

I've tried this (https://github.com/tgriesser/knex/issues/1138) by setting tableName config option to be '<schema>.db_migrations' (substituting the schema name in). Did not work.

I've searched the web, stack overflow, and issues. Can't seem to find an answer. Please advise.

mashaalmemon:
@kibertoad, absolutely.

Ultimately we have a multi-tenant system. Each tenant lives in a postgres 'schema' other than the default 'public' schema. So if I have a database with tenant1 and tenant2 in it, then that database would contain 3 postgres schemas:

'public' - nothing stored here
'tenant1' - tenant 1 has tables
'tenant2' - tenant 2 has tables

I want to define a general migration for any tenant.

Then I want to be able to run the migration on each of the tenant's schemas, which would update the tables (the same tables in each schema). Again, I am not using the CLI, but doing this programatically by instantiating a Migrator.

So to be clear re your suggestions:
-  to use [withSchema](https://knexjs.org/#Builder-withSchema), I need to know the schema name ahead of time. *Is there a way to pass variables in dynamically to the migration scripts to be consumed when the script is run?*
- prefix table names with schema, would require me to know the schema name ahead of time, which would not be possible, again, unless they are passed into the migration as a variable.

The approach I found working is:
* steps:
  * set `searchPath` in KnexConfig on database connection.  
  * then when running migration, migration is done to schema specified in the search path.
* I don't love this approach because I need to create a connection per migration, but it is workable if no better solution exists

kibertoad:
> Is there a way to pass variables in dynamically to the migration scripts to be consumed when the script is run?

Elegant solution would be https://github.com/tgriesser/knex/pull/2778. Since it is not currently merged and it's unclear at this point whether it would ever be, you could take advantage of the fact that migrations are executed sequentially and set schema as a "global" variable somewhere that migration script would read from.
Alternatively, since you have access to knex instance in your migrations, probably you could set some variables on it directly and then access them from migration script.

mashaalmemon:
[...]

PR #2802, opened 2018-09-11
Pass migration config object as an argument to the migrations (V 2.0)

kibertoad:
Reuses some tests from https://github.com/tgriesser/knex/pull/2778 but implemented in a completely different way.


kibertoad:
@elhigu Updated existing PR to make it mergeable; please let me know if something is missing, will address ASAP - would like to unblock https://github.com/tgriesser/knex/issues/2799

elhigu:
I still think that knex should have more general solution which can be used to resolve also this issue.

kibertoad:
Yeah, I will update this PR based on our discussion in the old PR.

kibertoad:
https://github.com/knex/documentation/pull/135 -> documentation

kibertoad:
> if withUserParams is called for transaction knex client. Maybe and error should be thrown

Should I check by this instanceof Transactor or there is a better way?

> Tests should also check that when transaction is created that user parameters can be accessed through that trx knex instance.

How is that different from existing `transaction of a copy with userParams retains userparams` test?

> (that could be implemented also later on I think)

I would appreciate that :). Stayed till 5 AM yesterday to finish this thing, need to recharge before I'll be able to do larger changes again.

kibertoad:
@elhigu OK, changes for initial param support were surprisingly straightforward to implement, so I think all of your comments are covered now (as previously mentioned, there was a test for userParams in transaction already).

elhigu:
This looks really promising, but I need to check this out when I'm less tired.

I might try to do some additional manual testing  by setting temporarily all knex test instances to be cloned instances to see that tests will work perfectly with those too :)

I also was worried that this could have been actually a lot harder to implement so that everything works fine after cloning, but I'm glad I was wrong. Great work 👍

kibertoad:
@elhigu Well, admittedly, it did took a bit of black magic to get cloning exactly right, as we have a function which is also an object with getters, which is not that straightforward to clone, but fortunately JS is flexible enough. You are welcome!

elhigu:
I got finally time to do some manual testing of this. 

I changed in suite.js passed knex instances to be cloned ones and some of the tests seem to be failing. Most of them seem to be related to knex.raw and maybe to postProcessResponse hook.

```
    require('./schema')(knex.withUserParams({}));
    require('./migrate')(knex.withUserParams({}));

    require('./seed')(knex.withUserParams({}));
    require('./builder/inserts')(knex.withUserParams({}));
    require('./builder/selects')(knex.withUserParams({}));
    require('./builder/unions')(knex.withUserParams({}));
    require('./builder/joins')(knex.withUserParams({}));
    require('./builder/aggregate')(knex.withUserParams({}));
    require('./builder/updates')(knex.withUserParams({}));
    require('./builder/transaction')(knex.withUserParams({}));
    require('./builder/deletes')(knex.withUserParams({}));
    require('./builder/additional')(knex.withUserParams({}));
    require('./datatype/bigint')(knex.withUserParams({}));
```

kibertoad:
@elhigu thanks, will take a look today!

kibertoad:
@elhigu OK, I've addressed everything that looked like genuine problems - remaining ones look like they are related to some isolation between different copies happening (different transactions?) that doesn't look like something worth spending effort on resolving, as I'm not even sure that it's an invalid behaviour, and user is not really supposed to expect different copies of knex to run in single transaction (you even requested explicitly to forbid transaction copying). If you run

```
const knexCopy = knex.withUserParams({});
require('./schema')(knexCopy);
    require('./migrate')(knexCopy);

    require('./seed')(knexCopy);
    require('./builder/inserts')(knexCopy);
    require('./builder/selects')(knexCopy);
    require('./builder/unions')(knexCopy);
    require('./builder/joins')(knexCopy);
    require('./builder/aggregate')(knexCopy);
    require('./builder/updates')(knexCopy);
    require('./builder/transaction')(knexCopy);
    require('./builder/deletes')(knexCopy);
    require('./builder/additional')(knexCopy);
    require('./datatype/bigint')(knexCopy);
```

it seems to pass fine now.

kibertoad:
Woohoo, the odyssey is over now!

elhigu:
Thanks! I'll make new `@next` release with this soon.

kibertoad:
Thank you! Any chance https://github.com/tgriesser/knex/pull/2817 could also get in if there is nothing wrong with the implementation? (I'll resolve conflict in a couple of minutes)

elhigu:
I think this + babel update together might have broken something

kibertoad:
I can totally look into it right now, what's the problem and how I can reproduce?

elhigu:
This one suggests https://www.npmjs.com/package/babel-preset-eslatest-node6 to use https://babeljs.io/docs/en/babel-polyfill/ for that support. Seems a bit more trustworthy.

kibertoad:
Right. Will add it now

kibertoad:
https://github.com/tgriesser/knex/pull/2821

PR #2810, opened 2018-09-17
Update nyc version

kibertoad:
Latest PRs are showing some strange deviations in code coverage percentage. Hopefully bumping nyc version would make it more consistent across baseline and new PRs.


kibertoad:
@elhigu Right, so force merging this should theoretically update the baseline code coverage to correct value and other outstanding PRs won't fail anymore.

Issue #2793, opened 2018-08-30
Update to Babel 7


kibertoad:
Babel 7 just got released, bringing bags of goodies, among them faster transpilation. Probably it's a good idea for us to upgrade as well.

kibertoad:
@elhigu If you approve, I could work on this.

elhigu:
I suppose it can be updated. I would like more if we could drop it completely, but I don't think that is going to happen at least this year.

kibertoad:
Well, if we migrate to Typescript we could drop it immediately, but since that isn't likely to be completed anytime soon either, April 2019 sounds realistic for when we could drop Babel directly (I don't think we really need any Node 10 features such as async iterators).

kibertoad:
Blocked by https://github.com/59naga/babel-plugin-add-module-exports/issues/65

pgebheim:
Well, this means that knex will break other projects which try to use Babel 7.

Would love to see this upgraded, but also -- why do you require babel-runtime as a non-dev-dep?

kibertoad:
@pgebheim Why would it break them? Other projects consume already transpiled sources so they do not depend on how knex uses Babel.
Babel-runtime is an interesting topic, though. @elhigu Could you comment on why we need it if we transpile anyway?

pgebheim:
@kibertoad Right now its breaking projects because babel-runtime is specified as a dependency, so it gets tracked into every project that includes Babel.

The end result is that there are multiple babel-runtime versions in the project and babel will break.

pgebheim:
Upon further investigation I think the error is a bit of a red herring. Knex definitely shouldn't be exporting babel-runtime as a real dependency, but the issue I am seeing is because electron-compile needs babel6.

It just so happens that since knex provided a real dep for babel-runtime and its version is newer than specified in electron-compile, the loading of the dep resolves to the version specified in Knex.

ricardograca:
Fixed with #2813.

PR #2813, opened 2018-09-19
Use Babel 7

kibertoad:
Implements https://github.com/tgriesser/knex/issues/2793


babel.config.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
```
kibertoad: babel.config.js is the officially recommended way to configure Babel

babel.config.js
```diff
@@ -0,0 +1,22 @@
...
+const isDev = process.env.npm_lifecycle_event === 'dev';
+
+module.exports = function(api) {
+  const presets = [
+    [
+      '@babel/preset-env',
+      Object.assign(
+        {},
+        isDev ? { targets: { node: 'current' } } : { targets: { node: '6' } }
```
kibertoad: Since we do not officially support Node versions below 6, I've explicitly locked transpilation at that version. This results in much more readable (and debuggable) code being generated.

package.json
```diff
@@ -7,7 +7,7 @@
     "node": ">=6"
   },
   "dependencies": {
-    "babel-runtime": "^6.26.0",
+    "@babel/runtime": "^7.0.0",
```
kibertoad: Do we still need this?

package.json
```diff
@@ -33,11 +32,11 @@
   "devDependencies": {
     "JSONStream": "^1.3.4",
     "async": "^2.6.1",
-    "babel-cli": "^6.26.0",
-    "babel-eslint": "^8.2.6",
-    "babel-plugin-add-module-exports": "^0.2.1",
-    "babel-plugin-transform-runtime": "^6.23.0",
```
kibertoad: I've removed transform-runtime since it seems to be mostly useful when building to very old Node versions and using it causes https://github.com/babel/babel-loader/issues/152

babel.config.js
```diff
@@ -0,0 +1,22 @@
...
+
+const isDev = process.env.npm_lifecycle_event === 'dev';
+
+module.exports = function(api) {
+  const presets = [
+    [
+      '@babel/preset-env',
+      Object.assign(
+        {},
+        isDev ? { targets: { node: 'current' } } : { targets: { node: 6 } }
```
kibertoad: Since we do not officially support Node versions below 6, I've explicitly locked transpilation at that version. This results in much more readable (and debuggable) code being generated.

package.json
```diff
@@ -89,7 +85,7 @@
     "pre_test": "npm run lint",
     "tape": "node test/tape/index.js | tap-spec",
     "debug_tape": "node --inspect-brk test/tape/index.js",
-    "test": "npm run pre_test && nyc mocha --exit --check-leaks -t 10000 -R spec test/index.js && npm run tape",
+    "test": "npm run pre_test && nyc mocha --exit --check-leaks --globals __core-js_shared__ -t 10000 -R spec test/index.js && npm run tape",
```
kibertoad: Ignore for "__core-js_shared__" was added since currently there is no way around it, Tedious brings babel-runtime into test execution and that results in an expected leakage (see https://github.com/babel/babel-loader/issues/152). This is not a problem with Knex nor something we should address.

package.json
```diff
@@ -155,9 +151,9 @@
   "tonicExampleFilename": "scripts/runkit-example.js",
   "nyc": {
     "check-coverage": true,
-    "lines": 85,
-    "statements": 83,
+    "lines": 84,
```
kibertoad: We generate more concise code now, this decreased coverage slightly.

src/dialects/sqlite3/schema/tablecompiler.js
```diff
@@ -132,7 +132,7 @@ TableCompiler_SQLite3.prototype.renameColumn = function(from, to) {
 
 TableCompiler_SQLite3.prototype.dropColumn = function() {
   const compiler = this;
-  const columns = Object.values(arguments);
+  const columns = values(arguments);
```
kibertoad: Apparently Object.values is one thing that Babel has no idea how to transpile, and polyfilling just for this one case feels like an overkill.

kibertoad:
@elhigu Is this one missing anything?

elhigu:
Very nice +1

PR #2814, opened 2018-09-20
Add json type support for SQLite

kibertoad:
Implements #1036

### DIFF
--- a/src/dialects/sqlite3/schema/columncompiler.js
+++ b/src/dialects/sqlite3/schema/columncompiler.js
@@ -22,4 +22,6 @@ ColumnCompiler_SQLite3.prototype.enu = function(allowed) {
   )}'))`;
 };
 
+ColumnCompiler_SQLite3.prototype.json = 'json';
+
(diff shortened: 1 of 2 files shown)


elhigu:
This one needs also integration test, which makes sure that the create query works against actual DB.

kibertoad:
@elhigu Looks like there is one already, see `test/integration/schema/index.js`:
```
it('allows adding a field', function() {
        return knex.schema.table('test_table_two', function(t) {
          t.json('json_data', true);
        });
      });
```

It used to pass before when text was passed and it doesn't seem to have started failing now.

bkniffler:
This is awesome, can this be released to @next? Thanks!

kibertoad:
@bkniffler next-3 was released on Sep 27, are you sure it doesn't include it?
(we do plan to release next-4 very soon anyway, though)

bkniffler:
Ah, I had next-2 installed since according to package.json it is the latest, at least on master.

Issue #1982, opened 2017-03-20
Query comment


bernardmo:
Hello,

Is there a way to add _sql comment_ to select query without using _raw queries_? I tried to look for docs on http://knexjs.org/ but can see comment support only for table manipulation. 

Query comments are useful for routing hints on maxscale router (https://mariadb.com/kb/en/mariadb-enterprise/mariadb-maxscale-14/maxscale-hint-syntax/).

VMBindraban:
I also would like to see this as a feature.

rfink:
Also support

kibertoad:
@rfink Typically it's better to use emoticons on original request to voice support for the feature rather than making "+1" comments, as these spam everyone subscribed to the issue.

rfink:
First attempt at adding this: https://github.com/tgriesser/knex/pull/2815

elhigu:
Which databases actually supports this?

elhigu:
> you are referring to query routing by comments

I'm referring any kind of comment with the query. I understand it is really useful for that proxy.

bernardmo:
> I'm referring any kind of comment with the query. I understand it is really useful for that proxy.

Well I know that mysql/mariadb, postgresql and sqlite support query comments. I'm not sure for other DBs which are supported by knex (I didn't worked with them)

elhigu:
Yea, looks like `/* comment */` are standard SQL stuff, I had always seen just line comments earlier :) Other thing to check about this is to see if Node's DB drivers supports them.

KristjanTammekivi:
Can confirm that PG supports them. We've hacked the client & querybuilder prototypes so that you can specify a comment with .options(). Works nicely but a proper support would be nicer.

kibertoad:
@KristjanTammekivi Maybe you would be interested in finishing https://github.com/tgriesser/knex/pull/2815?

lindakung:
Very excited to use this feature. Is there an ETA for when it might be released? Thanks!

GeneBuchite:
I am still deciding on which tools to use. I might not even use Vercel. Right now I am working on the backend.
Thank you for your interest

OlivierCavadenti:
Released in [2.5.0](https://github.com/knex/knex/releases/tag/2.5.0)

PR #2815, opened 2018-09-23
Add ability to prepend query comments

rfink:
Addresses #1982, first attempt.


elhigu:
This feature now just seems to allow to add any string to front of query which sounds pretty dangerous. Also it should be tested with all type of queries and throw an error if database doesn't support it.

I'm not sure how this should be supported... if this feature is named as `.comment()` then it should force comment marks there etc.

If this is just meant to allow to prepend anything to query, then naming should be changed too, but having that kind of feature for adding any strings to query would be more useful in a way that it would allow to add it to multiple places also in the middle of queries (that will be a lot harder to implement and would need some discussion about the design even if that is wanted to be included to knex).

rfink:
I don't think the ability to prepend just any old thing to the query would be needed. If I updated it to force comment marks, would that be more acceptable?

elhigu:
Which databases supports comments? 

I think it should be fine... if someone calls .comment() for database that doesn't support it, method should throw an error. There should be then also options object given as second parameter `.comment('foo', { ignoreOnNonSupportedDatabases: true })` for suppressing that error.

test/unit/query/builder.js
```diff
@@ -7478,6 +7478,36 @@ describe('QueryBuilder', function() {
     );
   });
 
+  it('#1982 - should allow query comments in querybuilder', function() {
+    testsql(
+      qb()
+        .from('testtable')
+        .comment('/* Added comment */'),
```
elhigu: Also multiple comments, error throwing and ignore on non supported databasese options should be tested

elhigu: And integration test for the supported databases

rfink:
I'm fairly certain all databases support comments - I believe it's part of the SQL standard. I could be wrong.

wubzz:
I have two requests if you don't mind.

1. Validation of input to `.comment` as string only. Is there any use-case aside from strings?

2. Would prefer if API is simply `.comment('string')` rather than having to specify `.comment('/* string */')`, its more clean. Or is there a reason behind this?

kibertoad:
@rfink Any chance you could finish it?

rfink:
Okay, this slipped my mind and I just found this in my backlog emails (sorry about the delay). I've added the automatic /* */ around the comment, added a check to ensure the comment is a string, added multiple comments for testing, and added the integration tests. The Travis tests seem largely to have passed, with the exception of the Oracle tests, which look to be failing due to a repository issue. Let me know if you need anything added!

elhigu:
I just realized that this could have been probably done with `query.wrap('/* comment */', '');` :)

src/query/compiler.js
```diff
@@ -211,6 +213,14 @@ assign(QueryCompiler.prototype, {
     );
   },
 
+  // Add comments to the query
+  comments() {
+    if (!this.queryComments.length) return '';
+    return this.queryComments
+      .map((comment) => `/* ${comment.comment} */`)
```
elhigu: Pretty dangerous, since this is not escaped in anyways so */ in middle of string will open query to sql injection... Probably would be good to mention in docs, that this method is very much vulnerable and should never be allowed to be read from user input.

test/integration/builder/selects.js
```diff
@@ -735,6 +735,51 @@ module.exports = function(knex) {
       });
     });
```
elhigu: Looks like there is no real integration tests at all checking that this kind of queries can actually be ran by sql drivers used by knex. So there should be some test, which also sends query until the database. Another thing that I want to see tested is putting ? in the comment string just to verify that those will not mess up parameter bindings.

elhigu:
sorry for huge delay on reviewing

rfink:
Hey, no worries. Sorry for the delay on my part. I'm happy to make these changes - I admit I'm not terribly familiar with the integration tests for running the queries. Can you make a recommendation on that?

Also for checking for /* and */ inside the string, are you okay with just a simple indexOf() ?

elhigu:
indexOf is fine. For example here https://github.com/tgriesser/knex/blob/master/test/tape/crossdb-compatibility.js and here https://github.com/tgriesser/knex/blob/master/test/integration/builder/selects.js#L1093  are some integration tests.

rfink:
Awesome - I've added these things, and I believe I have the integration tests running against the db properly, based on the links you sent. Thank you for your advice, and I'll be happy to add anything additional.

rfink:
Actually, it's not happy with ? in the comments. I'm checking into my options there.

elhigu:
I'm not sure what would be the best option. One option would be to automatically escape them in case if there is unescaped `?` in the comment string or just validate that there is no unescaped question marks there.

src/query/builder.js
```diff
@@ -152,6 +154,20 @@ assign(Builder.prototype, {
     return this;
   },
 
+  // Adds a comment to the query
+  comment(txt) {
+    if (!isString(txt)) {
+      throw new Error('Comment must be a string');
+    }
+    if (txt.indexOf('/*') !== -1 || txt.indexOf('*/') !== -1) {
```
gonnza: you can use [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) whichi already returns a boolean and is a bit more declarative (although we could argue indexOf is known by everyone)

```suggestion
    if (txt.includes('/*') || txt.includes('*/')) {
```

kibertoad:
@rfink Would you consider finishing this?

kibertoad:
@rfink Anything we can help with from our side?

gndelia:
I could try to take this PR if rfink can't

kibertoad:
@gndelia Much appreciated!

kibertoad:
@gndelia Are you still up to finish it?

gndelia:
Sorry, I had some difficulties and I forgot about this. Let anyone to start it, if it keeps untaken once I'm free I will write here. 😞

bywo:
I was able to work around this by subclassing QueryBuilder with

```
  toSQL() {
    const ret = super.toSQL();
    ret.sql = `/* my comment */`.concat(ret.sql);
    return ret;
  }
```

src/query/compiler.js
purefan: isnt this handled in [the proposed change to builder.js line 162](https://github.com/knex/knex/pull/2815/files#diff-17127b37f866fedbb3d1739324bd780bdb159055397cd2b702906fa5c02de974R162) ?

test/integration/builder/selects.js
purefan: Good concerns @elhigu ! Would you agree these were addressed in the tests just under? (lines 738~759)

francois2metz:
This pr can be closed https://github.com/knex/knex/pull/5289

OlivierCavadenti:
Closed, implemented in https://github.com/knex/knex/pull/5289

PR #2816, opened 2018-09-24
Typescript migration stub 

anggiaj:
add typings on typescript migration stub

### DIFF
--- a/src/migrate/stub/ts.stub
+++ b/src/migrate/stub/ts.stub
@@ -1,15 +1,15 @@
 import * as Knex from "knex";
 
-exports.up = function (knex: Knex): Promise<any> {
+export async function up(knex: Knex): Promise<Knex.SchemaBuilder> {
     <% if (d.tableName) { %>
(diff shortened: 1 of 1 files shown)


src/migrate/stub/ts.stub
```diff
@@ -1,15 +1,15 @@
 import * as Knex from "knex";
 
-exports.up = function (knex: Knex): Promise<any> {
+export async function up(knex: Knex): Promise<Knex.SchemaBuilder> {
```
elhigu: I don't thing tat typing is right. AFAIK Migration does not always resolve with Knex.SchemaBuilder instance.

anggiaj: true, isn't always return `Knex.SchemaBuilder`, but the default template use `createTable` and `dropTable` which both returns `Knex.SchemaBuilder`

anggiaj: updated

elhigu:
thanks, now it looks good :) gotta wait for tests to finish though.

PR #2817, opened 2018-09-24
Support passing connection override to a query

kibertoad:
implements #2804

PR #2821, opened 2018-09-26
Fix ES6 support

kibertoad:


kibertoad:
Yup, Node 6 is green now.

elhigu:
Now `@next´ is published with latest master.

kibertoad:
awww, #2817 missed the window of opportunity.

PR #2827, opened 2018-09-28
Extracts length from value wrapping to allow for prefix indexing.

fiferdc:
For an existing implementation, I've needed to add a prefix index.  index() does not support adding a column with a length, so I had to fall back on using raw() instead to add this index.  

I've extracted (_length_) from the value and parsed it to avoid SQL injection.  This length is re-introduced after the value is wrapped.


elhigu:
If you want to have custom  functionality for `wrapIndentifier` you should pass that custom function in configuration of your application https://knexjs.org/#Installation-wrap-identifier . 

Not changing the default implementation stored in main repository of the library.  What is done here is something that knex should never do by default.

PR #2829, opened 2018-09-30
Glob match migrations

iilei:
Following up on the discussion at #2756

### DIFF
--- a/src/migrate/Migrator.js
+++ b/src/migrate/Migrator.js
@@ -39,7 +39,6 @@ const CONFIG_DEFAULT = Object.freeze({
   globOptions: DEFAULT_GLOB_OPTIONS,
   tableName: 'knex_migrations',
   schemaName: null,
-  directory: './migrations',
   disableTransactions: false,
   sortDirsSeparately: false,
 });
@@ -386,7 +385,7 @@ export default class Migrator {
   }
 }
 
-export function getMergedConfig(config, currentConfig) {
+export async function getMergedConfig(config, currentConfig) {
   // config is the user specified config, mergedConfig has defaults and current config
   // applied to it.
   const mergedConfig = assign({}, CONFIG_DEFAULT, currentConfig || {}, config);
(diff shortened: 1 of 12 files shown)


iilei:
was filing this one by mistake 😅 - will have another session fixing the build

iilei:
Changelog Gist:

__Breaking Changes:__
- Options `directory` and `loadExtensions` superseded by glob pattern matching options `globPatterns` handed over to [globby](https://github.com/sindresorhus/globby)

__New features:__
- Option [`globOptions.gitignore=false`](https://github.com/sindresorhus/globby#gitignore) available for configuration of globbing behaviour.

iilei:
The failing tests say `mergedConfig.migrationSource.getMigrationDirs` is not implemented. Is that probably related to some caching of compiled sources?

src/migrate/Migrator.js
```diff
@@ -403,11 +405,17 @@ export function getMergedConfig(config, currentConfig) {
...
-      mergedConfig.directory,
-      mergedConfig.sortDirsSeparately
+      mergedConfig.globPatterns,
+      mergedConfig.sortDirsSeparately,
+      mergedConfig.globOptions
     );
   }
 
+  if (!mergedConfig.directory) {
+    const directory = await mergedConfig.migrationSource.getMigrationDirs();
```
JakeGinnivan: This is another breaking change. What would I return for say the `WebpackMigrationSource` which doesn't have directories?

src/migrate/Migrator.js
```diff
@@ -32,11 +35,10 @@ function LockError(msg) {
 inherits(LockError, Error);
 
 const CONFIG_DEFAULT = Object.freeze({
-  extension: 'js',
-  loadExtensions: DEFAULT_LOAD_EXTENSIONS,
+  globPatterns: DEFAULT_GLOB_PATTERNS,
```
JakeGinnivan: This could also be a non-breaking change. If globPatterns are not specified then the directory is used to build the glob?

iilei: I thought of that possibility but there is [already a breaking change related to the use case addressed](#2756 ) with this patch. Additionally I thought it was better this way in favor of a more concise API.

iilei:
I am pausing the work on this one for now. If someone likes to take over, I'd appreciate it

PR #2837, opened 2018-10-03
whereColumn

tgriesser:
Syntax for getting rid of the problem of doing a where clause with column bindings


elhigu:
Added issue about missing documentation. 

Also IMO `.where('column', '=', knex.ref('othercolumn'))` would have been sufficient syntax to support that use case and to keep number of different API method variants smaller.

tgriesser:
True, but it's such a common use case I think it justifies the inclusion as a top level method. Also with the code I'm working on [here](https://github.com/tgriesser/knex-next) I'd eventually like to get rid of the need for `knex.ref` with better API design.

Issue #2716, opened 2018-07-16
Circular dependencies (prevents proper `rollup`)


danielgindi:
# Environment

Knex version: 0.15.1
Database + version: Unrelated
OS: Unrelated

# Bug

Rollup spits those out:
```
Circular dependency: node_modules\knex\lib\raw.js -> node_modules\knex\lib\formatter.js -> node_modules\knex\lib\query\builder.js -> node_modules\knex\lib\raw.js
Circular dependency: node_modules\knex\lib\raw.js -> node_modules\knex\lib\formatter.js -> node_modules\knex\lib\query\builder.js ->  commonjs-proxy:C:\Users\Daniel.SILVERBOLT\Desktop\git\endpoint-analyzer\node_modules\knex\lib\raw.js -> node_modules\knex\lib\r
aw.js
Circular dependency: node_modules\knex\lib\raw.js -> node_modules\knex\lib\formatter.js -> node_modules\knex\lib\raw.js
Circular dependency: node_modules\knex\lib\schema\helpers.js -> node_modules\knex\lib\schema\columncompiler.js -> node_modules\knex\lib\schema\helpers.js
Circular dependency: node_modules\knex\lib\schema\helpers.js -> node_modules\knex\lib\schema\columncompiler.js ->  commonjs-proxy:C:\Users\Daniel.SILVERBOLT\Desktop\git\endpoint-analyzer\node_modules\knex\lib\schema\helpers.js -> node_modules\knex\lib\schema\he
lpers.js
Circular dependency: node_modules\knex\lib\schema\helpers.js -> node_modules\knex\lib\schema\tablecompiler.js -> node_modules\knex\lib\schema\helpers.js
Circular dependency: node_modules\knex\lib\schema\compiler.js -> node_modules\knex\lib\schema\helpers.js -> node_modules\knex\lib\schema\compiler.js
Circular dependency: node_modules\knex\lib\schema\compiler.js -> node_modules\knex\lib\schema\helpers.js ->  commonjs-proxy:C:\Users\Daniel.SILVERBOLT\Desktop\git\endpoint-analyzer\node_modules\knex\lib\schema\compiler.js -> node_modules\knex\lib\schema\compile
r.js
```

It is a huge problem, and will make code unexecutable after "rolling up".
The rollup process will convert commonjs deps into "static imports", which do not support circular dependencies.
Then it will end up with a flat hierarchy where there's a reference to a (yet) undefined variable.

[...]

PR #2838, opened 2018-10-03
Resolve unnecessary circular dependencies in code (#2716)

danielgindi:


elhigu:
Very interesting! I’m currently on vacations, but I’ll take a look as soon as i can.

danielgindi:
These are actually the simplest and tiniest code changes :-)
The problem originated mainly in old legacy code (`Knex.raw()`) which carried the bad design from the past.

elhigu:
Ok, so this didn't seem to change anything about the fact how formatter and dialect builder IIRC calls each other back and forth. But nevertheless this is a good fix too 👍

danielgindi:
Yes, it's not about the dialects yet, but it allows for a much easier (and robust) rollup.

It can now properly resolve imports for the core of `knex`, and as you know that you are using a specific dialect in your code - you can specify the path to the specific dialect to include as a "dynamic module" (https://github.com/rollup/rollup-plugin-commonjs/pull/331).

So the core of knex gets rolled-up with no problem - and allowing optimizations, tree-shaking etc, and the dialect which is user-configurable anyway stays user configurable.

There is probably more refactoring that can be done to make the hierarchy in the code cleaner

Issue #2743, opened 2018-08-01
Oracle Driver missing - Avoid Application Crash 


Velmuruganms12:
**# Environment**
@atiertant 

Knex version: 0.15.2
Database + version: Oracle 11g/12g
OS: Centos

**Current Scenario:**
When Oracle client library is not installed Application is getting crashed with error message. Unable to handle exception at connection creation using knex.  It works good when oracle client library when installed.

**Expected :**
We should able to capture the driver missing error in catch block and continue with application.

**# Feature discussion / request :**
we use knex for connecting to oracle, when install oracle client on the machine we dont have issues.

But when we dont have oracle client we are not able to catch exeception application is getting crashed.

Suggest us how we can handle exception better for this scenario.

**Client Connection Code:**

           ```
     client = require('knex')({
                    client: 'oracledb',
                    connection: {
                        user: details.user,
                        database: details.database,
                        host: details.host,
                        port: details.port,
                        password: details.password,
                    }
                });

```

**Error Message :**

0|npm      | Knex: run
0|npm      | $ npm install oracledb --save
0|npm      | Error: NJS-045: cannot load the oracledb add-on binary for Node.js 6.11.4 (linux, x64)
0|npm      | Node.js require() error was:
0|npm      |   DPI-1047: 64-bit Oracle Client library cannot be loaded: "libclntsh.so: cannot open shared object file: No such file or directory". See https://oracle.github.io/odpi/doc/installation.html#linux for help
0|npm      | Node.js require() mapped to /home/XXXXX/node_modules/oracledb/build/Release/oracledb.node
0|npm      | Node-oracledb installation instructions: https://oracle.github.io/node-oracledb/INSTALL.html
0|npm      | You must have 64-bit Oracle client libraries in LD_LIBRARY_PATH, or configured with ldconfig.
0|npm      | If you do not have Oracle Database on this computer, then install the Instant Client Basic or Basic Light package from
0|npm      | http://www.oracle.com/technetwork/topics/linuxx86-64soft-092277.html

elhigu:
Every dialect crashes the same way because of missing require. What would it help if you would be able to catch that error? At least migrator client would still crash anyways... 

I'm failing to see the actual use case here.

Velmuruganms12:
My use case here is application should not crash because of driver issue, instead it should throw error saying oralce client missing by catching the error message.

wubzz:
Can you explain how you would handle this error while keeping the application running? 

[...]

PR #2843, opened 2018-10-08
Throw instead of process.exit when client module missing

danderson00:
A couple of things - 

- The test relies on the `oracle` module not being in devDependencies. It will fail if it is ever added.
- I added the test to `test/unit/knex.js`, but I don't think the tests in this file are being run. I can't see the output for any of the tests in this file.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -215,10 +215,9 @@ assign(Client.prototype, {
     try {
       this.driver = this._driver();
     } catch (e) {
-      this.logger.error(
-        `Knex: run\n$ npm install ${this.driverName} --save\n${e.stack}`
-      );
-      process.exit(1);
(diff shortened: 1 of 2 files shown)


danderson00:
* being run when `npm test` is executed from the command line. Waiting for CI to run and see the results.

danderson00:
Resolves #2743. Not sure about failures. Looks like mssql failures from the build log, and not sure why code coverage should fall?

wubzz:
The coverage CI is super wonky. Don't know why, but if I had to guess I would say because of webpack -> output differs everytime.

Same in one of my PRs. I added 5 tests and barely any changes to the source, yet it decreased by like 1%..

So the gist of it is I'm personally ignoring it.

Thanks for this!

PR #2845, opened 2018-10-08
Manage types internally

tgriesser:
Copying the knex [type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/knex/index.d.ts) from DefinitelyTyped into the project. 

Putting these here will be useful since we can update them as new API's are merged.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,20 +1,22 @@
 ## How to contribute to Knex.js
 
-* Make changes in the `/src` directory and run `npm run babel` (runs once and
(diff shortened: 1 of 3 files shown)


tgriesser:
Cool, yeah. I'll merge this and then @JakeGinnivan if you want to open a PR here with the updated types we can merge that.

JakeGinnivan:
@tgriesser will get it done, thanks!

serprex:
Type definitions lack support for `{ alias: column }` in joins & `toSQL` on `Knex.Raw`. I'll open a PR when I have time

PR #2847, opened 2018-10-09
[WIP] Cluster support

kraftman:
Hi,

This is an initial attempt to add support for clusters similar to the poolCluster feature of [node-mysql](https://github.com/mysqljs/mysql#poolcluster), with support for adding multiple pools, getting them by name (with regex), and auto removing them based on error count.

I'm looking for some general feedback to see what your thoughts are on this strategy before I finish it off.

Cheers,
Chris


kraftman:
No worries. I found this: https://github.com/richraid21/knex-multiconnect that does as you say, but doesn't have the cluster auto removal support that node MySQL pool clusters do. I guess it could still be done in an independent module though. 

What weve done for now is switch to using AWS's read and write cluster endpoints and leave the failover to them.

kibertoad:
@kraftman Does this PR still make sense then?

chaffeqa:
@kraftman were you able to test this version much?  I'm thinking of forking and using it since I would much rather reley on `node-msql` for the pooling and management upkeep rather than a 1-off repo.

brcline:
@kraftman - did you ever end up using this?

kraftman:
> @kraftman - did you ever end up using this?

nope, I ended up using AWS which provides an url for the cluster instead

PR #2848, opened 2018-10-10
Remove Node 7. Minor dependency bump

### LINKED ISSUE #2841: Remove Node.js 7 from Travis configuration
Are there any reasons why we are running tests against Node.js 7? As is the case with all odd development releases, it's not officially supported.

kibertoad:
fixes #2841


package.json
```diff
@@ -12,7 +12,7 @@
     "bluebird": "^3.5.2",
     "chalk": "2.4.1",
     "commander": "^2.18.0",
-    "debug": "4.0.1",
+    "debug": "4.1.0",
```
kibertoad: I can split changes in package.json into a separate PR if someone feels strongly about it, but these are so tiny that I personally don't think that would be worth the hassle.

PR #2857, opened 2018-10-13
WIP: Replace MSSQL dialect with Tedious.js implementation

tylerjwatson:
## Tedious.JS MSSQL dialect driver

Hello Knex,

I have been working on deprecating the use of node-mssql in favour of tedious which is the underlying TDS stream driver for talking with Microsoft SQL server for about six months now.  I wanted to take it into production before I made people aware of this patch to make sure it was worth submitting.

## Rationale

I ran into major issues with every ORM or query library that uses `node-mssql` under the hood for MSSQL, which ultimately completely prevented me from running my system in production. 

In a nutshell, the problem is that node-mssql is in itself a high-level driver that contains its own mechanisms which Knex either doesn't need, or conflicts with.  To summarize the problems with using this library for this dialect:

* `Timeout acquiring resource` messages, killing the application,
* Random, huge latency spikes between query and result, making performance unstable,
* Connection timeouts on startup, and
* Inability to recover from error conditions

The SQL dialect experience with Microsoft SQL Server on node.js, and Knex in particular, is very poor because of this driver.

## Patch details

This patchset contains a _working_ implementation of tedious backed by a few months of my internal product running in production without issues.  The patch discards the second-level connection pool provided by node-mssql and simply attaches Tedious connection objects to Knex's pooling mechanism like every other dialect.

With my driver running in production there have been **no** resource timeouts, **no** connection issues, and Knex receives the results of the query as fast as Tedious.js can deserialize the token stream. This implementation has been the most stable my project has ever been on MSSQL.

I don't use some of MSSQL's functions so I am unable to complete some of the driver, but it is good enough for me to roll into production right now.

Thanks for looking! :+1:


tylerjwatson:
Crap! Totally forgot those were in there! I will clean ‘m up ASAP.

.travis.yml
```diff
@@ -10,9 +10,9 @@ cache:
 matrix:
   include:
   - node_js: "8"
-    env: TEST_ORACLEDB=true DB="maria mysql mysql2 postgres sqlite3 oracledb" CXX=g++-4.8 KNEX_TEST_TIMEOUT=60000 ORACLE_HOME=/u01/app/oracle/product/11.2.0/xe ORACLE_SID=XE OCI_LIB_DIR=/u01/app/oracle/product/11.2.0/xe/lib LD_LIBRARY_PATH=/u01/app/oracle/product/11.2.0/xe/lib
+    env: TEST_ORACLEDB=true DB="maria mysql mysql2 postgres sqlite3 oracledb" CXX=g++-4.8 KNEX_TEST_TIMEOUT=60000 ORACLE_HOME=/u01/app/oracle/product/11.2.0/xe ORACLE_SID=XE OCI_LIB_DIR=/u01/app/oracle/product/11.2../../lib LD_LIBRARY_PATH=/u01/app/oracle/product/11.2../../lib
```
kibertoad: Any particular reason for this change?

kibertoad:
@tylerjwatson I would like to contribute to the effort (e. g. sync with master), may I ask for a write permissions on your fork?

.travis.yml
tylerjwatson: I don't remember making this change, no.  It's possible that this diff is against a master which was changed elsewhere

tylerjwatson:
@kibertoad Can do, I just need to be careful for the time being.

The reason for the `lib` trash on this branch is because I needed to point yarn somewhere to the built sources so I can use it in production, and this seemed the least path of resistance to "get it working".

I'll ship my internal software onto a separate branch away from this pull, clean up the lib trash, and we'll take it from there.  

Stand by.

tylerjwatson:
Alright everyone, some cleanup has been done.  The lib trash has been removed and some meta files that I didn't mean to touch have been checked out from master to bring them back to where they should be.

Some things I'll need some help with now:

* Configuration format, I just jammed Knex's connection configuration straight into tedious, this is probably not what we want,
* For the life of me I could never get `npm run tape` to work on `plaintest`, so I just removed it to develop the driver.
* Transaction stuff could use a lookover
* Migration tests aren't too happy

Hopefully now we can get some review happening :+1:

kibertoad:
> I just jammed Knex's connection configuration straight into tedious

Isn't that what knex does for other dialects, though? I think {connection: {}} branch is always passed to driver as-is.

Can I get write access to your repo so that I could start fixing stuff, or you still need to remove some things?

package.json
```diff
@@ -54,6 +53,7 @@
     "sqlite3": "^4.0.0",
     "tap-spec": "^4.1.1",
     "tape": "^4.9.0",
+	"tedious":"^2.6.1",
```
kibertoad: Is indentation correct?

tylerjwatson:
Done mate, knew there was something I forgot. :P

src/dialects/mssql/index.js
```diff
@@ -1,207 +1,399 @@
...
   },
 
   // Used to explicitly close a connection, called internally by the pool
   // when a connection times out or the pool is shutdown.
   destroyRawConnection(connection) {
-    return connection.close()
+    debug("connection::destroy");
+    return new Promise((resolve, reject) => {
+      connection.once("end", () => {
+        resolve();
```
dhensby: You need to be careful to make sure that if the connection has been closed due to a connection error, that this promise still resolves. because the `end` event won't trigger when calling `connection.close()` if it's already closed.

We've had this problem with the `mssql` lib (see https://github.com/tediousjs/node-mssql/blob/v5.0.0/lib/tedious.js#L278-L298)

kibertoad:
@dhensby Do you think that there is still value in this PR? Most connection management issues with node-mssql were resolved in v5 already, and it is likely that v6 will address the remaining ones.

dhensby:
Whilst that's true, it does seem a bit strange to go through the mssql lib to access connections given the pools we create internally.

I think there's value to it, but it's a lot of effort for diminishing returns at the moment.

Given the effort has been put in, it would feel a shame not to get it over the line... But this is a few versions behind now...

It's probably 50/50 at the moment

tylerjwatson:
> I think there's value to it, but it's a lot of effort for diminishing returns at the moment.

Respectfully, the effort was simply because node-mssql is unsuitable for production.  Its use of generic-pool and forcing futures to go through this silly arbitration mechanism that breaks the very concept that futures provide programmers means it can never be the basis of a truly stable driver.  

I suppose it depends on what your values are, but for myself I can't risk the application going down for our customers simply because a dialect driver isn't stable.  The benefits are night and day for us and well worth the time investment.  We can never go back to to the previous driver as the trust has been broken here, and it doesn't look like the clashing high-level pool fundamentals are changing.  Sequelize is also a victim of the same problems and their issue trackers are a mess with these sorts of problems.

That said, upstreaming this driver means I don't have to maintain a fork which is great for me but it relies on community engagement as Knex's surface area on what it requires of a dialect far exceeds my use case and there are going to be people who use functionality that I don't.  I may have broken that for some users, but we'll never know if the community has no need to engage with another driver.

At the end of the day if the community is happy with the way it is now, there is no point in merging this.

Cheers

dhensby:
@tylerjwatson it's great to hear candid feedback and I've only started working on the mssql library in the last few months as we need it in production. The only reason we use it is because we need stored procedure support and that doesn't seem to be available anywhere else (that I know of).

I've tried to make quite a lot of progress around issues with pooling and so on.

> For example, their support has people performing bulk operations with `Bluebird.mapSeries` and other insane stuff like increasing pool timeouts because the driver can't understand that futures and database connections aren't, and should never be explicitly tied together. Future mechanisms like Promises shouldn't be used for queues, and doing so should be regarded as an error.

I *think* this has been removed as of v4 (I'm not familiar with the history of the library or how it worked, but there were issues complaining about queuing being removed when people were upgrading from 3 -> 4) so perhaps some of those issues have gone now?

Regardless, if you have any time or inclination to open some issues against the library, that'd be a big help.

We had some really bad issues in production with pool leakage, connection issues and other problems which, through various contributions and maintenance, are now gone.

I have no attachment or incentive to see Knex use `mssql` and it does make sense to me for it to make use of the tedious library directly but, like you say, it's up to the community / maintainers to take on any of the extra work to make it feature complete.

tylerjwatson:
> Regardless, if you have any time or inclination to open some issues against the library, that'd be a big help.

Sure mate, how can I help?

I last dealt with `node-mssql` around the turn of v3 to v4.  The main impetus for this driver was the fact that it would choose random queries to just... stop.  It was un-reproducible. It would happen at indeterminate times to any query, and knex would just stop responding to requests from the clients, grinding our show to a complete halt.  I never quite figured out what it was, so after many sleepless nights and expletives I simply gave up and did my own.  Sometimes it would cough up a message saying `resource aquisition timeout` and other times it would just stop issuing queries.  The root cause is _somewhere_ in node-mssql because my driver has had no downtime in production since we adopted it, ever.

I just could not produce a test case for any reports to be of use to anyone.  I feel that a report without reproducibility is just whinging.  I'm happy to help, what would you have me do?

dhensby:
Thanks for that @tylerjwatson 

As you've said that this was around the turn of v3 -> v4, I think that it's likely a lot of the behaviour you've experienced has been resolved and there may not be a lot off feedback that is applicable anymore.

It seems like there was quite a fundamental shift in v3 -> v4.

I'm glad that you've managed to resolve your production issues by running tedious directly. We don't use knex in prod, just `mssql` and over the past few months we've got it into a pretty healthy shape for prod compared to when we started - but this is all relatively recent and it sounds like your experience is from quite some time ago.

elhigu:
I would be still happy to change mssql -> tedious, but I don't see it as urgent as before, since the worst of the problems did seem to be resolved and my own stress tests did pass on newer mssql version.  I still consider mssql driver as an unnecessary piece that is not adding any value to knex.

tylerjwatson:
Hi guys, I'm back with a later revision of this patch.  I've just completed a rebase of the tedious MSSQL driver onto Knex 0.20.

Some things of note here:
* All the MSSQL-specific monkey patches have been removed
* All tests pass locally, but travis seems to fail some unrelated MySQL things, and configuration resolver stuff that appears to be new.
* Some tests have been modified: `; select @@rowcount` is redundant as tedious has these counts natively from the stream
* The tape has been modified for MSSQL as tedious itself handles transaction stuff.
* In mssql, `rollback` is analogous to `rollbackTo`.

kibertoad:
@tylerjwatson Looking at https://travis-ci.org/knex/knex/jobs/604778022?utm_medium=notification&utm_source=github_status, I see failures only for MSSQL tests. Could you take a look at them?

kibertoad:
@elhigu Do we still want to replace node-mssql with Tedious, given that node-mssql seem to be finally fixed properly?

kibertoad:
Ah, I see your previous comment on exactly this matter, I guess that stands.

dhensby:
Just to give my thoughts on mssql - it does seem a bit overkill to use mssql as the dependency (though it provides a bit of syntactic sugar on the top of tedious), but as @kibertoad points out a lot of the problems with mssql (especially with pooling) have been fixed and I'm about to release v6, which should prove to be a pretty simple upgrade.

It'll definitely be leaner to use tedious directly but the maintenance effort might not be worth it

tylerjwatson:
I’m confused at how having clashing pooling layers with a high-level driver talking to another high-level driver is less maintenance than simply talking to the stream like all the other dialects. Could you please elaborate?

I still get “unable to acquire connection” errors when using stock mssql dialect implementation so I will still be forced to use my downstream version. I’m happy to maintain the dialect once it’s been upstreamed if need be, or I can publish the dialect driver as a separate downstream package in knex-mssql as another option for people to use.

The general sentiment is _mssql seems to be fixed_ which is not a good place to be. It needs to be stable, and not relying on monkey-patching to get us there.

Cheers

Sent from my iPhone

On 4 Nov 2019, at 4:15 am, Daniel Hensby <notifications@github.com> wrote:

﻿

Just to give my thoughts on mssql - it does seem a bit overkill to use mssql as the dependency (though it provides a bit of syntactic sugar on the top of tedious), but as @kibertoad<https://github.com/kibertoad> points out a lot of the problems with mssql (especially with pooling) have been fixed and I'm about to release v6, which should prove to be a pretty simple upgrade.

It'll definitely be leaner to use tedious directly but the maintenance effort might not be worth it

dhensby:
> I’m confused at how having clashing pooling layers with a high-level driver talking to another high-level driver is less maintenance than simply talking to the stream like all the other dialects. Could you please elaborate?

I don't think it's important for us to get bogged down in that discussion. Whether the maintenance burden is greater or less is for the maintainers of this library to decide, it is just a point I was raising.

> The general sentiment is _mssql seems to be fixed_ which is not a good place to be. It needs to be stable, and not relying on monkey-patching to get us there.

The mssql driver has gone through 2 major versions since the original pooling problems and the pooling library has been replaced with tarn and is much more stable. We should not rely on monkey-patches to keep mssql stable.

kibertoad:
Simpler is better, I guess. So as long as we can make tests pass on this PR, should be fine to merge it.

kibertoad:
@tylerjwatson Any estimates on when you would have time to take a look at test failures? I'd like to bump prettier version in dependencies, but doing so is going to cause lots of conflicts for your changes, so I'd prefer to avoid that until this PR is merged.

tylerjwatson:
Hi mate.

Apologies for going dark; I’m sporting an injury at the moment and won’t be back in the game until next week at the very least.

Feel free to do your changes, and I will look after the downstream forks if need be.

Cheers!

Sent from my iPhone

On 15 Nov 2019, at 5:20 am, Igor Savin <notifications@github.com> wrote:

﻿

@tylerjwatson<https://github.com/tylerjwatson> Any estimates on when you would have time to take a look at test failures? I'd like to bump prettier version in dependencies, but doing so is going to cause lots of conflicts for your changes, so I'd prefer to avoid that until this PR is merged.

kibertoad:
Getting closer!
Hope you are feeling better now.

briandamaged:
This PR looks like it will be very handy!

However, we've also been refactoring a lot of the core `Transaction` / `Dialect` logic lately.  On the plus side, this should make it easier to introduce new Dialects in the future.  But, on the downside, I suspect that there are some very major merge conflicts between this branch and `master` now.

Would you like a hand w/ getting this branch updated?

(On a side-note: it might actually be easier to implement this alongside the existing clients rather than as a wholesale replacement.  This would help to avoid merge conflicts and allow potential end-users to beta-test the new implementation)

kibertoad:
@tylerjwatson Just a heads-up - I'm currently working on syncing your branch up with the master, hopefully this PR can finally be merged after all this time.

kibertoad:
@tylerjwatson Way more tests are failing now, which is to be expected after syncing with such a significat internal refactoring, I guess :-/. Will try to catch breakages now.

kibertoad:
@tylerjwatson There are two large groups of failing tests left: response processing being broken and migration table being unlocked. Also I see that transaction reimplementation is incomplete, e. g. `rollbackTo` is not implemented. Would you be able to pick any of these items?

tylerjwatson:
Wow I thought this PR was abandoned.  Appreciate it, I will take a look.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

tylerjwatson (commented):
I made a grave mistake.  Disregard this and I will fix it.

kibertoad:
That seems to have resolved the majority of problems! Only 6 tests are failing, looks like this will be good to go after transaction support is completed.

tylerjwatson:
Hooray!  The 6 failing tests seem to be the transaction tape and all of them are savepoint stuff that are guarded with statements like

```js
      let expectedTrx1QueryCount =
        knex.client.driverName === 'oracledb' ||
        knex.client.driverName === 'mssql'
          ? 1
          : 3;
```

MSSQL has quite strange support for savepoints.  I'm not sure these tapes are valid.

tylerjwatson:
The tapes assume that transaction statements emit `query` events, which in tedious they do not.  I can either try and emit fake `query` events or modify the tapes to take this into account.

kibertoad:
@tylerjwatson Why did they use to pass in the past then? Does `node-mssql` has its own logic for handling transactions? What would be your suggested path forward?

kibertoad:
@tylerjwatson Do you think that it is likely that someone might have been relying on these events in the past in their production logic? That would imply we should be firing them on our end now.

tylerjwatson:
@kibertoad It's impossible for me to say for certain, and I would struggle to find a decent use case that would need to handle transaction semantics inside events.

```js
knex(table).transaction(async function trx1() {
  trx1.on('query', data => {
    if (/begin/i.test(data.sql) {
       // ... seems farfetched.
    }
  });
});
```

It is not my place to tell others what they do with their production code, and since changing a tape that affects all dialects is much more risky than just faking events, I'll look into emitting query events on the transaction.  Behavioural consistency between all dialects as much as possible is also better.

kibertoad:
@tylerjwatson Yes, that sounds reasonable. Thank you!

test/integration/query/selects.js
```diff
@@ -1347,9 +1372,10 @@ module.exports = function (knex) {
               // mssql fails because it tires to rollback at the same time when update query is running
               // hopefully for share really works though...
               if (knex.client.driverName == 'mssql') {
-                expect(err.message).to.be.contain(
-                  "Can't rollback transaction. There is a request in progress"
-                );
+                // TODO: reconcile this behaviour under tedious
```
kibertoad, line 1375: Is this still relevant?

lib/dialects/mssql/transaction.js
```diff
@@ -26,77 +87,86 @@ module.exports = class Transaction_MSSQL extends Transaction {
...
-      configConnection ||
-      (await this.client.acquireConnection());
-
-    try {
-      conn.__knexTxId = this.txid;
-      if (!this.outerTx) {
-        this.conn = conn;
-        conn.tx_ = conn.transaction();
-      }
+  // acquireConnection(client, config, txid) {
```
kibertoad, line 143: Is the commented part still useful for anything or can be dropped?

package.json
```diff
@@ -117,6 +117,7 @@
     "sqlite3": "^5.0.1",
     "tap-spec": "^5.0.0",
     "tape": "^5.1.1",
+    "tedious": "9.2.3",
```
kibertoad, line 120: Can we bump it to 11? Which version do you use for running in prod currently?

test/integration/query/selects.js
tylerjwatson: I'm ...actually not sure what this is about, do you remember?

tylerjwatson: Uncommenting out that test yields a `Requests can only be made in the LoggedIn state, not the SentClientRequest state` error message which is what tedious gives you when you try to make two requests at once.  Is that going to suffice?

package.json
tylerjwatson: This is the version that we have been running in prod.  Our application stack has thousands of connections to Azure SQL, and Tedious 9 allows for token auth.  It makes sense to be in line with the latest releases

test/integration/query/selects.js
kibertoad: Yes, definitely!

kibertoad:
@tylerjwatson Tests are still green, so I would say we can merge it after updating that one transaction test.

test/integration/query/selects.js
tylerjwatson: @kibertoad Sorry, not quite sure what you mean, and I'm a bit confused as to what's happening with this test.  I can simply update the error message in the test to make it pass, but that feels dirty. 

Although, it seems like a fuzzing test anyway that is checking for the presence of an error?

kibertoad: I mean, since we are still throwing an error that is related to executing two requests at same time, sounds good enough, and just changing test assertion sounds sufficient. If previous error was thrown by node-mssql, I'm not sure if we should/can replicate throwing exact same error. I guess we could try-catch, wrap and rethrow error, but that sounds dirtier.

tylerjwatson: The dialect transaction now throws an appropriate error when attempting to rollback a transaction that has a pending request, bringing it inline with how `node-mssql` handles it.  Tedious cannot ever roll back a deadlocked query.

tylerjwatson:
It should be ready now.  Let me know if you need me to pay attention to other areas.

kibertoad:
This is great, thank you!
Note: while I was syncing with master, I dropped few of previous changes from your branch, to focus on driver change per se, without changing the behaviour.
Notably: passing query context to errors; JSON support; removing explicit select rowcount from queries. Since I assume that you had good reasons for these changes, and you would need them should you return upstream, maybe we could discuss them and restore some or all of them in a separate PR?

tylerjwatson:
Awesome!  Glad to see this has finally landed after all this time.  I am _very_ excited about coming back upstream, it's never nice being left in the dark with new features.  I'm particularly excited about the deprecation of Bluebird for native promises.

1) We're finding that when an SQL query error is generated we have no idea where it came from.  Passing express `req` objects as query context is a small-overhead way to attach meaning to errors in the callbacks.  Since query errors are logically owned by queries which have a query context we found no harm in sinking the query context into error emits.  We can simply log them to the logger attached to the express request and we can get meaningful information like requesting user, response time, etc.

There's little use in the `query-error` event otherwise as there's nothing for our devops team to relate back to.  This change is important and we'll need it before we return upstream unless I can figure out something else.

2)  I can't remember what JSON support entails.

3) Removing `@@rowcount` was erroneous functionality and can be left alone as-is.

kibertoad:
@tylerjwatson Gotcha, I'll make a PR restoring query context and JSON support shortly for your reviewn then.
Another thing missing are the TS type changes. Could you look into syncing current knex mssql config types with what is currently supported?

tylerjwatson:
@kibertoad Nah let's do it properly in separate PRs. 

For JSON support see #4278, 

I'll look at doing query-error support properly shortly

I'll also update the types appropriately

PR #2858, opened 2018-10-13
Adjust typing for forUpdate()/forShare() variant with table names

sgoll:
This is a followup to #2835 that became necessary with the merge of #2845. It adjust the typing of `forUpdate()` and `forShare()` to support the new call variants with table names.

It makes it possible to call `forUpdate()` and `forShare()` in either of the following three ways:

- `forUpdate()`
- `forUpdate('lorem', 'ipsum')`
- `forUpdate(['lorem', 'ipsum'])`


tgriesser:
Awesome, thanks @sgoll

PR #2859, opened 2018-10-14
Fix linter errors in tests

henryjw:
A lot of files were modified, but the changes are mostly changing `var` declarations to `let` or `const`.


tgriesser:
Cool, thanks @henryjw! One question, is `eslint-plugin-react` needed for anything or can that be removed?

henryjw:
No problem.
For the packages, I had to install them because otherwise `eslint` errors out. Does liniting work fine for you without them?

kibertoad:
@henryjw Why would you need react eslint for backend node.js library without any traces of React, though?..

kibertoad:
@henryjw Linting used to work fine for me in the past, so I assume it is something on your local configuration end.

henryjw:
@kibertoad you were right. It was something in my local setup. I've removed the additional dependencies.

package.json
```diff
@@ -82,6 +82,7 @@
     "coveralls": "nyc report --reporter=text-lcov | coveralls",
     "dev": "rimraf ./lib && babel -w src --out-dir lib --copy-files",
     "lint": "eslint src/**",
+    "lint-test": "eslint test/",
```
kibertoad: shouldn't this be also included in travis file to be executed in CI?

kibertoad:
@tgriesser @elhigu Do we want to merge this before https://github.com/tgriesser/knex/pull/2857 has landed? I assume that we really, really want that one to be completed and merged, and rebasing it against these changes is going to take some effort. Probably it's going to be easier to rebase this PR later on instead.

package.json
tgriesser: The line above can just be change from 
```
"lint": "eslint src/**",
```
to
```
```

tgriesser:
> I assume that we really, really want that one to be completed and merged, and rebasing it against these changes is going to take some effort.

Yeah we definitely do want it to be merged, though that might take awhile and the lint errors in the tests always bug the hell out of me, I don't mind taking the time to merge this against those changes. 

Looking through the files changed in #2857 PR I actually don't think it'll be too hard, there's not too many test files changed and those that do don't really overlap too much with the lines changed by this PR.

@henryjw if you fix the line I mentioned above with the `lint` (remove `lint-test`), I'm good to merge this.

package.json
henryjw: Fixed.

henryjw:
@tgriesser I made the change but it looks like it broke the build. Does the build also take into account warnings? If so, then I'll have to resolve those as well.

kibertoad:
@henryjw https://travis-ci.org/tgriesser/knex/jobs/441730589 -> judging by logs, those are pretty much errors.

henryjw:
@kibertoad Yea that's strange. On my local they're warnings. Could it be that Travis treats warnings as errors?

henryjw:
@tgriesser @kibertoad Pushed the fixes.

tgriesser:
Great, thanks @henryjw!

PR #2861, opened 2018-10-17
Support patched MSSQL lib

dhensby:
As discussed in previously (https://github.com/tgriesser/knex/pull/2637#issuecomment-420281476), the node-mssql lib pool depletion bug was patched in 4.2.1 (see https://github.com/tediousjs/node-mssql/pull/683) and locking the version of mssql to 4.1.0 is counter productive.

This change aims to force either 4.1.0 (which is then monkey patched) or a version >= 4.2.1 <5 which is left alone.

Ideally this would make a 0.15.3 knex release, if possible.

I know the long-term goal is to drop the mssql dependency all together, but we quite need 4.2.1 and some of the patches in 0.15 of knex too, so this allows a intermediate resolution until mssql dependency is written out.


dhensby:
So test suite is failing because the code coverage has just dipped below the expected values, there doesn't appear to be an existing test for this feature, is there an appropriate way to test this?

kibertoad:
I don't think that 0.15.3 is feasible as we don't have separate branch for 0.15 point releases and master went quite ahead since last 0.15 release with plenty of breaking changes.

Did you run all the stress tests on 4.2.1 for any meaningful amount of time to confirm that all the patched issues are actually resolved?

I think code coverage could be ignored for this one, it's not like we are going to test with different dependency versions.

kibertoad:
In order to avoid test failure, I would suggest to add istanbul ignore to that part of the code - https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md

kibertoad:
Thanks!
I don't think it's documented, take a look at "stress:init", "stress:test" and "stress:destroy" scripts in package.json, and I think @elhigu used to run it overnight to get meaningful results.

dhensby:
Hmm - interesting, the stress test is forcing mssql down to 0 queries, which is weird... maybe the connection depletion problem isn't completely resolved - I shall investigate

kibertoad:
@dhensby Probably it would be useful to see what knex monkey patching is doing and see if it could be applied directly to node-mssql.

dhensby:
Yep

dhensby:
OK - I'm going to close this for now as it seems the mssql lib hasn't fixed the problem, it seems! I've pulled out just the fix and opened a PR here - https://github.com/tediousjs/node-mssql/pull/719

Once that's in I'll revisit this.

dhensby:
Right, so it seems that this is actually a bug in *tedious* and not in node-mssql per se (well, it could be handled better in mssql) - see https://github.com/tediousjs/tedious/pull/769 which was released in 2.7.1.

kibertoad:
@dhensby Did you try running stress test with an updated tedious version?

kibertoad:
@dhensby node-mssql 4.2.2 with fixed tedious was released, might make sense to give it a shot.

dhensby:
OK - this still fails stress testing now with a new error:

```
events.js:167
      throw er; // Unhandled 'error' event
      ^

Error: No event 'socketError' in state 'LoggedInSendingInitialSql'
    at Connection.dispatchEvent (/Users/dhensby/projects/dhensby/knex/node_modules/tedious/lib/connection.js:1014:28)
    at Connection.socketError (/Users/dhensby/projects/dhensby/knex/node_modules/tedious/lib/connection.js:1030:12)
    at Connection.socketEnd (/Users/dhensby/projects/dhensby/knex/node_modules/tedious/lib/connection.js:1047:14)
    at Socket.<anonymous> (/Users/dhensby/projects/dhensby/knex/node_modules/tedious/lib/connection.js:884:18)
    at Socket.emit (events.js:187:15)
    at endReadableNT (_stream_readable.js:1081:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
Emitted 'error' event at:
    at Connection.tedious.on.err (/Users/dhensby/projects/dhensby/knex/node_modules/mssql/lib/tedious.js:252:14)
    at Connection.emit (events.js:182:13)
    at Connection.dispatchEvent (/Users/dhensby/projects/dhensby/knex/node_modules/tedious/lib/connection.js:1014:14)
    at Connection.socketError (/Users/dhensby/projects/dhensby/knex/node_modules/tedious/lib/connection.js:1030:12)
    [... lines matching original stack trace ...]
    at process._tickCallback (internal/process/next_tick.js:63:19)
```

dhensby:
Fix here: https://github.com/tediousjs/tedious/pull/809

kibertoad:
@dhensby It would be fantastic to finish this in time before 0.16.0 release, it would really make life of all the Node.js/MSSQL users better. Is there anything I can do to help it happen?

kibertoad:
Depends on node-mssql bumping tedious version to fixed one: https://github.com/tediousjs/node-mssql/issues/733

elhigu:
Also I need to check that this will still work as stable as patched version.

kibertoad:
@dhensby PR to bump tedious dependency raised: https://github.com/tediousjs/node-mssql/pull/755

src/dialects/mssql/index.js
```diff
@@ -61,25 +61,32 @@ assign(Client_MSSQL.prototype, {
...
-          '(knex patches bug in its implementation)'
-      );
+    /* istanbul ignore next */
+    if (mssqlVersion === '4.1.0') {
+      mssqlTedious.ConnectionPool.prototype.release = release;
+      mssqlTedious.ConnectionPool.prototype._poolCreate = _poolCreate;
+    } else {
+      const [major, minor, patch] = mssqlVersion.split('.');
+      // if version is not ^4.2.2
+      if (major != 4 || minor < 2 || (minor == 2 && patch < 2)) {
```
kibertoad: this logic should be updated to expect >= 5.0.0

dhensby: Done

dhensby:
YES! it passes, I no longer get the error detailed in https://github.com/tgriesser/knex/pull/2861#issuecomment-430958157

I also tested against 4.3.0 to make sure the tedious upgraded did fix it and it does (4.3.0 failed).

kibertoad:
Woohoo!
@elhigu Would you like to check its stability from a branch or releasing within a "next" version for testing would be more convenient?

dhensby:
@kibertoad what I would say is that an alpha release is subject to breaking changes, so I don't know how wise it would be to merge this as-is.

kibertoad:
@dhensby It's a devDependency anyway, I don't imagine knex needing to change anything in production code after stable version is released.

dhensby:
ah, of course, good point

elhigu:
> 

> Woohoo!
> @elhigu Would you like to check its stability from a branch or releasing within a "next" version for testing would be more convenient?

It will be easier to run tests on this branch direclty. I'm on it.

elhigu:
With fast testing this seems to be even more stable, than current master (I was able to crash that when socket is broken during login). 🎉   Next step would be to add deprecation message for that old mssql version and then in 0.17 we could remove that patching code altogether.

elhigu:
I'll leave that stress test running during my lunch and then merge this if everything looks good.

src/dialects/mssql/index.js
```diff
@@ -61,25 +61,31 @@ assign(Client_MSSQL.prototype, {
...
+    /* istanbul ignore next */
+    if (mssqlVersion === '4.1.0') {
+      mssqlTedious.ConnectionPool.prototype.release = release;
+      mssqlTedious.ConnectionPool.prototype._poolCreate = _poolCreate;
+    } else {
+      const [major] = mssqlVersion.split('.');
+      // if version is not ^5.0.0
+      if (major < 5) {
+        throw new Error(
+          'This knex version does not support mssql versions below 5.0.0'
```
dhensby: This message is wrong, I guess. It should be 4.1.0 or ^5

kibertoad: Updated

elhigu:
I suppose https://github.com/tgriesser/knex/blob/master/.travis.yml#L29 those allow failures travis builds should be still removed?

kibertoad:
@elhigu You mean within the context of oracle fix PR, right?

elhigu:
Yeah, I mixed up issues. In oracledb context indeed :)

PR #2865, opened 2018-10-18
Add test for retrieving null dates

### LINKED ISSUE #1276: If it's NULL, let it be NULL

kibertoad:
Looks like original issue was only with Maria dialect which we no longer support. This adds test from https://github.com/tgriesser/knex/pull/1308/ to make sure that other dialects don't have same problem.

fixes #1276

### DIFF
--- a/test/integration/builder/selects.js
+++ b/test/integration/builder/selects.js
@@ -735,6 +735,31 @@ module.exports = function(knex) {
       });
     });
 
+    it('#1276 - Dates NULL should be returned as NULL, not as new Date(null)', function() {
+      return knex.schema
+        .dropTableIfExists('DatesTest')
+        .createTable('DatesTest', function(table) {
+          table.increments('id').primary();
+          table.dateTime('dateTimeCol');
+          table.timestamp('timeStampCol');
+        })
+        .then(function() {
+          return knex('DatesTest').insert([
+            {
+              dateTimeCol: null,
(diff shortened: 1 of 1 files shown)


kibertoad:
@wubzz Do you think that having these tests in knex to guard about possible regressions in the future makes sense?

wubzz:
@kibertoad Yes, I think it's fine. The more coverage we have, the better. Things like these aren't obvious to add tests for generally speaking, you just expect it to work out of the box. But we've had situations where things we expect simply do not work, either due to knex or even on dialect level, so imho it's good additions.

Issue #1476, opened 2016-06-07
Add support to remove certain single parts of a query (e.g. offset, limit)


jgornick:
Currently, there doesn't seem to be a way to remove a limit from a query. I would expect `.limit(null)` to clear it out, but seems to keep the previously set limit value.

My use case is I have a query where I clone it and remove the limit and offset so that I can capture an entire count row count of the query before sending it back to the client. I clone my original query, then want to remove the limit.

My workaround for now is to set the offset to 0 and a limit of some number greater than all record count.

Thanks!

elhigu:
Sadly Knex doesn't support removing stuff from query builder / asking any metadata about the query (which would be really nice feature).

wubzz:
I would consider this a bug based on implementation comparison of `.limit()` and `.offset()`. From a quick peak I can tell "nullifying" offset after a clone is possible, but trying to do the same for `limit()` will print
>A valid integer must be provided to limit

elhigu:
Can order by be removed too? What does documentation say about these? Is passing null still undefined behavior? 

I'm fine with allowing to remove limit, offset, order by and others (if there are some more) by overriding them wirh null, but I would like to have that kind of functionality to be documented and consistent as well.

(Cant check docs now... Sleeping + answering fron mobile :)

wubzz:
@elhigu `.orderBy()` is even worse, this throws an error when called with null because of a missing safe-guard in the compiler. It trying to `formatter.columnize(null)` which throws `TypeError: Cannot read property 'length' of null`.

At the same time, `.orderBy()` is a bit more tricky. You can call the function multiple times for the same querybuilder and it will append to the sort order, so you get multiple objects. Therefore, resetting by calling with null isn't really valid for this function. That brings us back to your point of removing stuff from the querybuilder, which is something we lack today.

arjun-livquik:
My solution is simple and it works for me.
eg.
```
  var qb = knex.select('*').from('users');
  if(limit){
    qb.limit(limit);
  }
  if(offset){
    qb.offset(offset);
  }
  qb.then(function(res){
    console.log(res);
  }, function(err){
    console.error(err);
  });
```

Similar thing can be done for `.orderBy()` as well I guess.

arjunnayak0705:
My solution is simple and it works for me.
eg.
```
  var qb = knex.select('*').from('users');
  if(limit){
    qb.limit(limit);
  }
  if(offset){
    qb.offset(offset);
  }
  qb.then(function(res){
[...]

PR #2868, opened 2018-10-21
#1476 fix. cleareLimit(), cleareOffset()

Fi1osof:


elhigu:
Can be reopened if fixed.

Fi1osof:
@elhigu 
> the same method name implemented twice
Stupid error(((

Oracle error:
```
  1 failing
  1) oracle | oracledb
       unions
         handles unions with a raw query:
     select * from "accounts" where "id" = :1 union select * from accounts where id = :2 - ORA-00942: table or view does not exist
  Error: ORA-00942: table or view does not exist
```
https://github.com/tgriesser/knex/pull/2869

kibertoad:
@Fi1osof Ignore that error, same failure can be observed in master as well, there are known problems with Oracle.

PR #2869, opened 2018-10-21
#1476 fix. clearLimit(), clearOffset()

Fi1osof:


kibertoad:
No tests?..

src/query/builder.js
```diff
@@ -932,6 +932,18 @@ assign(Builder.prototype, {
     return this;
   },
 
+  // Remove limit
+  clearLimit() {
+    delete this._single.limit;
```
kibertoad: delete is an expensive operation, are you sure that "set to undefined" is not enough in this case?

elhigu: I was thinking the same, that probably setting to null would be better, but then I thought that its premature optimizing and probably doesnt have any difference in the real world.

kibertoad: Considering that in this case it doesn't cost anything, I don't see why settle for second best :). And you are right, probably null is better.

Fi1osof:
> No tests?..

Sorry, no know this package so well.

src/query/builder.js
Fi1osof: > delete is an expensive operation, are you sure that "set to undefined" is not enough in this case?

In starts this._single have not props limit and offset. If set {limit: undefined} this._single.hasOwnProperty("limit") will return true. I do not know whether this will lead to conflicts.

kibertoad:
@Fi1osof Simple unit tests should be enough here - assert that builder constructs query as expected after calling new methods.

Take a look at test\unit\query\builder.js for inspiration

src/query/builder.js
kibertoad: I don't think "hasOwnProperty" is relevant for anything other than cloning in Knex case and cloning null is fine, so I think it shouldn't be a problem.

test/unit/query/builder.js
```diff
@@ -750,6 +750,32 @@ describe('QueryBuilder', function() {
     );
   });
 
+  it('clear limits and offsets', function() {
```
kibertoad: In general it is advisable to have more granular tests; would you consider splitting this into two tests, separately for limit and for offset?

Fi1osof: Sorry, i have no more time...

kibertoad:
@Fi1osof Please raise a PR with documentation for the new feature at https://github.com/knex/documentation/pulls

kibertoad:
@elhigu Can you imagine any real-world use-case when user would reset offset or limit in the middle of query building? Considering that this PR was abandoned unfinished and without documentation, I wonder if it's worth the effort to push over the finish line.

Fi1osof:
@kibertoad @elhigu sorry, i do not know english well and do not familiar with documentation and tests. And i can not complete this PR. But i can give an example. 
```
let countQuery = query.clone();

let objects = await this.request(query).then()

countQuery.clearOrder();
countQuery.clearSelect();
countQuery.select(uniqueColumn);

Object.assign(countQuery._single, {
  limit: undefined,
  offset: undefined,
});

let qQuery = knex.count("* as count").from(countQuery.as("t1"));

let count = await this.request(qQuery)
  .then(r => r && r[0].count || 0));
```
I did clone prepared query with subqueries and wont count rows (without sort and limit). If i did not clear limit and offset, i get incorrect count.

kibertoad:
@Fi1osof OK, fair enough, for cloned knex instances it makes sense. Can you at least split test in two as was requested during code review? It doesn't require elaborate English or testing skills, just do something along the lines of

```
it('clear limits', function() {
    testsql(
      qb()
        .select('*')
        .from('users')
        .offset(5)
        .limit(10)
        .clearLimit()
      {
// update code below so that they still have offset
        mysql: {
          sql: 'select * from `users`',
        },
        mssql: {
          sql: 'select * from [users]',
        },
        pg: {
          sql: 'select * from "users"',
        },
        'pg-redshift': {
          sql: 'select * from "users"',
        },
      }
    );
  });
```

and then same for offsets.
I could write documentation myself then.

mdartic:
Was this PR near the "finish line" ;-) ?

PR #2872, opened 2018-10-23
switch to uuid-random instead of uuid package

jacklrs:
Because usage of synchronous functions inside "uuid" package performance suffers, if knex is used in context of node based web-server. I tested "uuid-random" package which increase performance by 30%. (despite of my commit comment, new package does use synchr. functions, but in optimized way)

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "dev": "rimraf ./lib && babel -w src --out-dir lib --copy-files",
     "lint": "eslint '{src,test}/**/*.js'",
     "plaintest": "mocha --exit -t 10000 -b -R spec test/index.js && npm run tape",
-    "prepublish": "npm run babel",
+    "prepare": "npm run babel",
     "pre_test": "npm run lint",
     "tape": "node test/tape/index.js | tap-spec",
     "debug_tape": "node --inspect-brk test/tape/index.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -83,7 +83,8 @@
     "dev": "rimraf ./lib && babel -w src --out-dir lib --copy-files",
     "lint": "eslint '{src,test}/**/*.js'",
     "plaintest": "mocha --exit -t 10000 -b -R spec test/index.js && npm run tape",
-    "prepublish": "npm run babel",
+    "prepare": "npm run babel",
```
kibertoad: why was this changed?

kibertoad:
I'm not sure about using not very battle-tested library (1k downloads/week vs 10m for uuid). Is this a hot spot and bottleneck?

jacklrs:
I avoided any synchronous function calls so far, and this is the only one alerted by node (WARNING: Detected use of sync API ...) in my app servers log (in dev-mode) so far. The best might be, to use something which create UUIDs without usage of synchronous functions ... but did not find a package that so far ...

package.json
kibertoad: This sounds out of scope for this PR, though. Could you prepare a separate PR with explanation of the change for this/

elhigu:
I dont see any reason to change this. 30% difference in microbenchmarks is not valid reason.

If you are able to show some realworld benchmark, where the difference is significant that would be a good reason.

Jabher:
I think you can probably use https://github.com/ai/nanoid for that (I'm not not affiliated with author)
It gives nearly same performance increase (~25%, according to docs), also it uses less memory for IDs due to the length

Issue #2874, opened 2018-10-24
multiple columns in .orderBy()


HurSungYun:
# Environment

Knex version: 0.15.2

# Feature discussion / request

It would be really great to be able to use a single `.orderBy()` when setting order with multiple columns.

For now, I should write code like
```
.select()
.from('table')
.orderBy('a')
.orderBy('b', 'desc')
.orderBy('c', 'desc')
```

It can be simplified like
```
.select()
.from('table')
.orderBy('a', 'asc', 'b', 'desc', 'c', 'desc')
```

or like this
```
.select()
.from('table')
.orderBy(['a', 'asc'], ['b', 'desc'], ['c','desc'])
```

or some other forms better than this would also be okay.

I can work on this if this feature request is accepted.

HurSungYun:
https://github.com/tgriesser/knex/blob/66acbe7719470d22d4aa65fdc9709f2717e1bca7/src/query/builder.js#L542-L550

This code implies a single `.orderBy()` can only handle one column. So for now, I think there is no way to do it.

Another interface suggestions:
```
.select()
.from('table')
.orderBy(['a','b','c'], ['asc', 'desc', 'desc'])
```

```
.select()
.from('table')
.orderBy({
  a: 'asc',
  b: 'desc',
  c: 'desc',
});
```

In implementation level, I wanna suggest two solutions.

* multiple `_statements.push()` in builder

```
for (const column in columns) {
   this._statements.push({
      grouping: 'order',
      type: 'orderByBasic',
      value: column,
      directions[column],
    });
}
```

* add add `type` in `_statements`

```
this._statements.push({
    grouping: 'order',
    type: 'orderByMultiple',
    value: columns,
    directions,
});
```

sgoll:
My two cents: I find the suggested syntax `orderBy('a', 'asc', 'b', 'desc', 'c', 'desc')` very confusing: at first glance it is not clear at all that every other argument has a special meaning (`asc`/`desc` for direction) and is not just another column name. This becomes more problematic when using a variable for the arguments, e.g. `orderBy(...columnsWithDirections)`.

The same goes for `orderBy(['a','b','c'], ['asc', 'desc', 'desc'])`—it is not clear which of the two arrays is the column names and which is the direction (although, arguably, the same could be said for the regular `orderBy(column, direction)` syntax as well, so …

[...]

PR #2881, opened 2018-10-27
support multiple columns in `.orderBy()`

HurSungYun:
implementation of #2874

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -541,7 +541,7 @@ assign(Builder.prototype, {
   // Adds a `order by` clause to the query.
   orderBy(column, direction) {
     if (Array.isArray(column)) {
-      return this.orderByArray(column);
+      return this._orderByArray(column);
     }
     this._statements.push({
       grouping: 'order',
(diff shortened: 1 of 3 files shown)


src/query/builder.js
```diff
@@ -549,6 +552,28 @@ assign(Builder.prototype, {
...
+    for (let i = 0; i < columnDefs.length; i++) {
+      const columnInfo = columnDefs[i];
+      if (isObject(columnInfo)) {
+        this._statements.push({
+          grouping: 'order',
+          type: 'orderByBasic',
+          value: columnInfo['column'],
+          direction: columnInfo['order'],
+        });
+      } else if (typeof columnInfo === 'string') {
```
kibertoad: why not use _.isString?

src/query/builder.js
```diff
@@ -549,6 +552,28 @@ assign(Builder.prototype, {
     return this;
   },
 
+  // Adds a `order by` with multiple columns to the query.
+  orderByArray(columnDefs) {
```
kibertoad: is this supposed to be a public method?

kibertoad:
Thank you for your contribution! Would you consider creating documentation in https://github.com/knex/documentation for it?

kibertoad:
LGTM! @elhigu what do you think?

elhigu:
Everything seemed fine to me 👍

PR #2884, opened 2018-10-28
Migrations with CLI and without knexfile

### LINKED ISSUE #2819: Create migrations with the CLI without knexfile
### Feature discussion / request

It would be convenient to be able to create migrations without knexfile, one could pass the required arguments to the CLI.

In my case only `client` or `dialect` is necessary but one can also imagine passing the directory of the migrations or other by the arguments.

```
knex migrate:make create_rule_table --client mysql --directory ./migrations
```

aurium:
Allows to create and run migrations.

closes #2819


bin/cli.js
```diff
@@ -35,11 +35,29 @@ function checkLocalModule(env) {
...
+  };
+}
+
+function initKnex(env, commander) {
   checkLocalModule(env);
 
-  if (!env.configPath) {
-    exit('No knexfile found in this directory. Specify a path with --knexfile');
+  if (!env.configuration) {
+    if (commander.client) mkConfigObj(env, commander);
```
kibertoad: did you run prettier formatting before commit?..

bin/cli.js
```diff
@@ -52,7 +70,10 @@ function initKnex(env) {
 
   var environment = commander.env || process.env.NODE_ENV;
   var defaultEnv = 'development';
-  var config = require(env.configPath);
+  var config =
```
kibertoad: Would you consider replacing all vars in this file with let/const?

bin/cli.js
```diff
@@ -52,7 +70,10 @@ function initKnex(env) {
...
+    typeof env.configuration == 'object'
```
kibertoad: since we depend on Lodash anyway, you could just use _.isObject

bin/cli.js
```diff
@@ -35,11 +35,29 @@ function checkLocalModule(env) {
   }
 }
 
-function initKnex(env) {
+function mkConfigObj(env, commander) {
+  env.configuration = {
+    ext: 'js',
+    [commander.env || process.env.NODE_ENV || 'development']: {
```
kibertoad: this is very difficult to read. Could you please split it somewhat? Inline condition inside key name resolution is heavy :)

bin/cli.js
```diff
@@ -35,11 +35,29 @@ function checkLocalModule(env) {
   }
 }
 
-function initKnex(env) {
+function mkConfigObj(env, commander) {
+  env.configuration = {
```
kibertoad: so this is mutating passed params? Can this be done in a more immutable way?

bin/cli.js
```diff
@@ -90,6 +112,12 @@ function invoke(env) {
     .option('--debug', 'Run with debugging.')
     .option('--knexfile [path]', 'Specify the knexfile path.')
     .option('--cwd [path]', 'Specify the working directory.')
+    .option('--client [name]', 'Set DB client whitout a knexfile.')
```
kibertoad: "whitout " -> typo

bin/cli.js
```diff
@@ -90,6 +112,12 @@ function invoke(env) {
...
+    .option('--connection [address]', 'Set DB connection whitout a knexfile.')
```
kibertoad: same

bin/cli.js
```diff
@@ -136,8 +164,12 @@ function invoke(env) {
       'Specify the stub extension (default js)'
     )
     .action(function(name) {
-      var instance = initKnex(env);
-      var ext = (argv.x || env.configPath.split('.').pop()).toLowerCase();
+      var instance = initKnex(env, commander);
```
kibertoad: do we need to pass commander instance for internal calls? can't we pass resolved params instead?

bin/cli.js
```diff
@@ -136,8 +164,12 @@ function invoke(env) {
...
+      var ext = (
```
kibertoad: might make sense to extract to internal function with self-explanatory name or at least add comment with explanation

test/migrate-cli.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash -e
+
```
kibertoad: wouldn't it be simpler to just write a Jake (https://www.npmjs.com/package/jake) script within usual unit test?

kibertoad:
@aurium Thank you for your contribution!

kibertoad:
CI is failing with "test/migrate-cli.sh: line 4: realpath: command not found" error, btw.

bin/cli.js
aurium: No. Is it necessary?

bin/cli.js
aurium: Thanks! Changing...

bin/cli.js
aurium: I could, however there are many "var's" in the same scope and upper. I could do this in another PR to isolate the changes.

bin/cli.js
aurium: There is a method to do that on commander API? Can't find.

bin/cli.js
aurium: Now using `initKnex(env, commander.opts())`. looks more explicit on its intention.

bin/cli.js
kibertoad: Fair enough :)

kibertoad:
@aurium Tests are failing on node 6 and 8 now. I think "fs.promises" is only available since Node 10, so probably you would have to promisify fs methods manually (e. g. using bluebird).

kibertoad:
LGTM! @elhigu, what do you think?

elhigu:
Hey... awesome client tests! 🎉 Now we can start finishing all the other client PRs which were deferred because of missing tests 😊

elhigu:
And great work everyone!

PR #2887, opened 2018-11-02
Add verbose option to CLI commands

avaly:
The following CLI commands get a `--verbose` option, which is used to determine whether to output the resulting log from the command action:

- `migrate:latest`
- `migrate:rollback`
- `seed:run`

### DIFF
--- a/test/jake-util/helpers/migration-test-helper.js
+++ b/test/jake-util/helpers/migration-test-helper.js
@@ -9,12 +9,14 @@ function assertExec(cmd, desc) {
   desc = desc || 'Run ' + cmd;
   return new Promise((resolve, reject) => {
     let stderr = '';
-    console.log(`Executing: ${cmd}`);
+    let stdout = '';
+    // console.log(`Executing: ${cmd}`);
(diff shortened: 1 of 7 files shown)


kibertoad:
Needs test. See https://github.com/tgriesser/knex/pull/2884 for a good example.

avaly:
I'll wait for that PR to land first and then I'll add similar tests.

avaly:
I've added tests for the updated code in this PR.

test/jake-util/helpers/migration-test-helper.js
```diff
@@ -9,12 +9,14 @@ function assertExec(cmd, desc) {
   desc = desc || 'Run ' + cmd;
   return new Promise((resolve, reject) => {
     let stderr = '';
-    console.log(`Executing: ${cmd}`);
+    let stdout = '';
+    // console.log(`Executing: ${cmd}`);
```
kibertoad: Should this be commented out?

kibertoad:
@avaly Thank you! Could you also please create a PR with documentation at https://github.com/knex/documentation ?

PR #2889, opened 2018-11-04
Add withRecursive()

### LINKED ISSUE #1755: With Recursive (Common Table Expressions)

caseywebdev:
Closes  #1755, an alternative to #2218

I needed `WITH RECURSIVE` in a project recently so I monkey patched this in and it works great 👌, figured I'd help out the upstream.

I noticed the other PR tried to handle the `UNION ALL` automatically for the recursive section, but in my case I actually wanted just a `UNION`. I think simply having this method add the `RECURSIVE` keyword is the way to go since it's trivial to do the `.union(...)` or `.unionAll(...)` or whatever one needs in the query section of the `.withRecursive()` method.

Future improvements might include a `raw` version, but I think getting this added in its simplest form is an easy win.

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -47,7 +47,7 @@ inherits(Builder, EventEmitter);
 
 const validateWithArgs = function(alias, statement, method) {
   if (typeof alias !== 'string') {
-    throw new Error(method + '() first argument must be a string');
+    throw new Error(`${method}() first argument must be a string`);
   }
   if (
     typeof statement === 'function' ||
@@ -57,7 +57,7 @@ const validateWithArgs = function(alias, statement, method) {
     return;
   }
   throw new Error(
-    method + '() second argument must be a function / QueryBuilder or a raw'
+    `${method}() second argument must be a function / QueryBuilder or a raw`
   );
 };
 
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -45,6 +45,22 @@ function Builder(client) {
 }
 inherits(Builder, EventEmitter);
 
+const validateWithArgs = function(alias, statement, method) {
+  if (typeof alias !== 'string') {
+    throw new Error(method + '() first argument must be a string');
+  }
+  if (
+    typeof statement === 'function' ||
+    statement instanceof Builder ||
+    statement instanceof Raw
+  ) {
+    return;
+  }
+  throw new Error(
+    method + '() second argument must be a function / QueryBuilder or a raw'
+  );
+};
+
 assign(Builder.prototype, {
   toString() {
     return this.toQuery();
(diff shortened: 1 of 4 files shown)


src/query/builder.js
```diff
@@ -45,6 +45,22 @@ function Builder(client) {
 }
 inherits(Builder, EventEmitter);
 
+const validateWithArgs = function(alias, statement, method) {
+  if (typeof alias !== 'string') {
+    throw new Error(method + '() first argument must be a string');
```
kibertoad: why not use template string?

src/query/builder.js
```diff
@@ -118,6 +123,21 @@ assign(Builder.prototype, {
...
+  // With Recursive
+  // ------
+
+  withRecursive(alias, statement) {
+    validateWithArgs(alias, statement, 'withRecursive');
+    return this.withRecursiveWrapped(alias, statement);
+  },
+
+  // Helper for compiling any advanced `withRecursive` queries.
+  withRecursiveWrapped(alias, query) {
```
kibertoad: why separate this from `withRecursive` if it just passes same params after validation? Also, should it be public?

src/query/compiler.js
```diff
@@ -678,12 +678,14 @@ assign(QueryCompiler.prototype, {
     if (!withs) return;
     const sql = [];
     let i = -1;
+    let recursive = '';
     while (++i < withs.length) {
       const stmt = withs[i];
+      if (stmt.recursive) recursive = 'recursive ';
```
kibertoad: It's preferable to always use {}

kibertoad: This looks a bit redundant, you are resetting string multiple times and constructing string at this time looks kinda premature. Might be cleaner to set boolean flag instead and do something like

```
const withBlock = isRecursive ? : 'with recursive ' : 'with '
```
later on

kibertoad:
Thank you for your contribution! Could you also create a PR in https://github.com/knex/documentation?

src/query/builder.js
caseywebdev: I was simply refactoring this to be used in both `with*` methods. I've noticed template aren't used in most of this codebase and I assumed this was support older JS interpreters. If you're saying template strings are fine I'll update.

src/query/builder.js
caseywebdev: I'm just mirroring the pattern established by `with` and `withWrapped`.

caseywebdev:
Comments addressed and docs PR created https://github.com/knex/documentation/pull/144

tgriesser:
Yoo @caseywebdev good to see you man! Looks good to me, thanks for the patch!

caseywebdev:
Hey Tim! Happy to help 👍🏼 Thanks for the merge 🎆

Issue #1840, opened 2016-12-23
Request: enable "indexType" on MySQL dialect.


musou1500:
I want to create spatial index to specific column by using this feature.
I know `knex.raw` make it possible to do.

douglasjunior:
This is very important, how can I create an index of type FULLTEXT?

elhigu:
@douglasjunior with `knex.schema.raw()` at least. I cannot remember if passing index type was already fixed.

kibertoad:
@douglasjunior According to documentation, it is supported, see https://knexjs.org/#Schema-index
If you try and it doesn't work, please consider raising a PR with failing test for it, then we could address it rapidly.

douglasjunior:
I'm writing a PR right now. 😆

douglasjunior:
Done #2890.

glensc:
@knex: perhaps close, as indexType was added with #2890.

glensc:
```ts
  await knex.schema.alterTable('article_attribute', (table) => {
    table.index('value_text', 'article_attribute_value_text', 'FULLTEXT');
  });
```

PR #2890, opened 2018-11-05
Added option to pass indexType for MySQL dialect

douglasjunior:
I followed the [MySQL Reference](https://dev.mysql.com/doc/refman/8.0/en/create-index.html).

```sql
CREATE [UNIQUE | FULLTEXT | SPATIAL] INDEX index_name
    [index_type]
    ON tbl_name (key_part,...)
    [index_option]
    [algorithm_option | lock_option] ...

key_part: {col_name [(length)] | (expr)} [ASC | DESC]

index_option:
    KEY_BLOCK_SIZE [=] value
  | index_type
  | WITH PARSER parser_name
  | COMMENT 'string'
  | {VISIBLE | INVISIBLE}

index_type:
    USING {BTREE | HASH}

algorithm_option:
    ALGORITHM [=] {DEFAULT | INPLACE | COPY}

lock_option:
    LOCK [=] {DEFAULT | NONE | SHARED | EXCLUSIVE}
```

No changes are required in `types/knex.d.ts`.

I'm not sure if I need to write tests for this, I could not find where the MySQL schema tests are written.

Steps to reproduce:

1. Add my fork as dependency in package.json:
```json
"knex": "git+https://github.com/douglasjunior/knex.git#custom"
```

2. Create a migration file:
```js
exports.up = (knex, Promise) => {
    return  knex.schema.createTable('my_table', table => {
        table.bigIncrements('id');

        table.string('my_long_text_column', 1000).notNullable();

        table.index('my_long_text_column', null, 'FULLTEXT'); // <= here
    });
};

exports.down = (knex, Promise) => {
    return knex.schema.dropTableIfExists('my_table');
};
```

3. Run `npx knex migrate:latest`.

4. See the generated schema:
```sql
CREATE TABLE `my_table` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `my_long_text_column` varchar(1000) NOT NULL,
  PRIMARY KEY (`id`),
  FULLTEXT KEY `my_table_my_long_text_column_index` (`nome_original`)
) 
```

Related to #1840

### DIFF
--- a/src/dialects/mysql/schema/tablecompiler.js
+++ b/src/dialects/mysql/schema/tablecompiler.js
@@ -199,8 +199,9 @@ assign(TableCompiler_MySQL.prototype, {
       ? this.formatter.wrap(indexName)
       : this._indexCommand('index', this.tableNameRaw, columns);
     this.pushQuery(
-      `alter table ${this.tableName()} add ${indexType ||
-        ''} index ${indexName}(${this.formatter.columnize(columns)})`
+      `alter table ${this.tableName()} add${
+        indexType ? ` ${indexType}` : ''
+      } index ${indexName}(${this.formatter.columnize(columns)})`
     );
   },
 
--- a/src/dialects/mysql/schema/tablecompiler.js
+++ b/src/dialects/mysql/schema/tablecompiler.js
@@ -194,14 +194,13 @@ assign(TableCompiler_MySQL.prototype, {
       })
     );
   },
-  index(columns, indexName) {
+  index(columns, indexName, indexType) {
     indexName = indexName
       ? this.formatter.wrap(indexName)
       : this._indexCommand('index', this.tableNameRaw, columns);
     this.pushQuery(
-      `alter table ${this.tableName()} add index ${indexName}(${this.formatter.columnize(
-        columns
-      )})`
+      `alter table ${this.tableName()} add ${indexType ||
+        ''} index ${indexName}(${this.formatter.columnize(columns)})`
     );
   },
 
--- a/test/unit/schema/mysql.js
+++ b/test/unit/schema/mysql.js
@@ -300,7 +300,7 @@ module.exports = function(dialect) {
       );
     });
 
-    it('test adding index with type', function() {
+    it('test adding index with an index type', function() {
       tableSql = client
         .schemaBuilder()
         .table('users', function() {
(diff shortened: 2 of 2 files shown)


src/dialects/mysql/schema/tablecompiler.js
```diff
@@ -194,14 +194,13 @@ assign(TableCompiler_MySQL.prototype, {
       })
     );
   },
-  index(columns, indexName) {
+  index(columns, indexName, indexType) {
```
kibertoad: can you also add test for this? probably just a unit test for query generation would suffice

kibertoad: and there probably are existing tests for pg already

douglasjunior: The format of the tests is confusing to me. Many tests are failing and I do not know why. https://travis-ci.org/tgriesser/knex/jobs/450867352 (The tests fail even before the change)

I'll look at the postgre tests to see if I can understand.

douglasjunior: Done.

kibertoad:
```
    it('test adding index', function() {
      tableSql = client
        .schemaBuilder()
        .table('users', function() {
          this.index(['foo', 'bar'], 'baz');
        })
        .toSQL();

      equal(1, tableSql.length);
        'alter table `users` add index `baz`(`foo`, `bar`)'
      );
    });
```

This is an example of a test failing. I assume structure of an index statement changed even when index type is not specified, which is actually probably not something that we want - maybe excessive space somewhere?..

kibertoad: @douglasjunior Feel free to ignore Oracle tests failing, they've been like that for a while already. But MySQL ones should be passing.

douglasjunior: Now I think I get it, please, could you check it again?

kibertoad:
@douglasjunior Thank you very much! Would you consider creating PR for https://github.com/knex/documentation to mention MySQL support for .index()?

kibertoad:
@tgriesser LGTM, what do you think?

douglasjunior:
Added docs https://github.com/knex/documentation/pull/145

glensc:
Filed a request for actual index_type:
- https://github.com/knex/knex/issues/4637

PR #2894, opened 2018-11-06
Version upgrade guide

kibertoad:
I've tried to include cases that are plausible to be found in real production environments. If someone have their favourite breaking change not mentioned, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,9 @@ Node.js, featuring:
 
 Node.js versions 6+ are supported.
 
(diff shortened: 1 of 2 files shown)


elhigu:
Very nice :+1:

kibertoad:
Merged as-is, we can improve going forward.

willmorgan:
@kibertoad Looks good!

kibertoad:
@willmorgan Thanks for checking it out!

PR #2895, opened 2018-11-06
Add tests for specifying explicit pg version

kibertoad:


kibertoad:
@elhigu Tried to find something like this, but couldn't find anything that would be setting {version} option for the knex. Have I missed it somewhere?

elhigu:
I went to look the code, that implements version support, checked commit where it was written with "Blame" and found this: https://github.com/tgriesser/knex/pull/1993/files

elhigu:
Git archeology :D

kibertoad:
Ahahahaha, yeah.
Aha, so there are tests, but they cover different ground (branching in internal logic based on version) and not version resolution logic per se, so there doesn't seem to be direct overlap, hence makes sense to keep new tests.

Issue #1134, opened 2016-01-12
Cannot read property 'split' of null error


demisx:
``` sh
$ knex -V
Knex CLI version:  0.9.0
Local Knex version:  0.9.0
```

Running `knex migrate:make test` produces this error:

``` sh
$ knex migrate:make test
/usr/local/lib/node_modules/knex/lib/bin/cli.js:104
      var ext = (argv.x || env.configPath.split('.').pop()).toLowerCase();
                                         ^
TypeError: Cannot read property 'split' of null
    at Command.<anonymous> (/usr/local/lib/node_modules/knex/lib/bin/cli.js:104:42)
    at Command.listener (/usr/local/lib/node_modules/knex/node_modules/commander/index.js:301:8)
    at emitTwo (events.js:87:13)
    at Command.emit (events.js:172:7)
    at Command.parseArgs (/usr/local/lib/node_modules/knex/node_modules/commander/index.js:610:12)
    at Command.parse (/usr/local/lib/node_modules/knex/node_modules/commander/index.js:458:21)
    at Liftoff.invoke (/usr/local/lib/node_modules/knex/lib/bin/cli.js:166:13)
    at Liftoff.launch (/usr/local/lib/node_modules/knex/node_modules/liftoff/index.js:163:8)
    at Object.<anonymous> (/usr/local/lib/node_modules/knex/lib/bin/cli.js:186:5)
    at Module._compile (module.js:398:26)
    at Object.Module._extensions..js (module.js:405:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Function.Module.runMain (module.js:430:10)
    at startup (node.js:141:18)
    at node.js:1003:3
```

elhigu:
Can you show your knexfile too? Sounds like the migration path is set wrong there.

demisx:
Here it is:
```js
module.exports = {
  development: {
    client: 'pg',
    connection: {
      database: 'data-api-dev',
      user:     '',
      password: ''
    },
    migrations: {
      tableName: 'knex_migrations'
    },
    debug: true
  },

  test: {
    client: 'pg',
    connection: {
      database: 'data-api-test',
      user:     '',
      password: 'password'
    },
    migrations: {
      tableName: 'knex_migrations'
    }
  },

  staging: {
    client: 'pg',
    connection: {
      database: 'data-api-stg',
      user:     '',
      password: 'password'
    },
    pool: {
      min: 2,
      max: 10
    },
[...]

PR #2905, opened 2018-11-13
CLI cleanup. Do not require client for creating seeds or migrations

kibertoad:
Cleanup and improvements after https://github.com/tgriesser/knex/pull/2884/files
fixes https://github.com/tgriesser/knex/issues/1134


bin/cli.js
```diff
@@ -55,12 +55,18 @@ function mkConfigObj(opts) {
 function initKnex(env, opts) {
   checkLocalModule(env);
 
-  if (!env.configuration) {
-    if (opts.client) env.configuration = mkConfigObj(opts);
-    else
+  if (!env.pathToKnexFile) {
```
kibertoad: I've separated `pathToKnexFile` and `configuration` properties to reduce confusion.

bin/cli.js
```diff
@@ -55,12 +55,18 @@ function mkConfigObj(opts) {
...
+    if (opts.client) {
+      env.configuration = mkConfigObj(opts);
+    } else {
       exit(
-        'No knexfile found in this directory. Specify a path with --knexfile'
+        'No knexfile found in this directory. Specify a path with --knexfile or pass --client and --connection params in commandline'
```
kibertoad: I though this was broken recently, but looks like it was the case for quite a while. Doesn't look like we search for "knexfile in this directory" at all. We probably should.

bin/cli.js
```diff
@@ -165,16 +170,14 @@ function invoke(env) {
       `-x [${filetypes.join('|')}]`,
       'Specify the stub extension (default js)'
     )
-    .action(function(name) {
-      var instance = initKnex(env, commander.opts());
-      var ext = (
-        argv.x ||
-        env.configuration.ext ||
-        env.configuration.split('.').pop()
```
kibertoad: I have no idea why we were trying to parse knexfile param for extensions, it looks wrong. I'm open to restoring it if someone explains why it was there in the first place.

src/util/batchInsert.js
```diff
@@ -1,84 +1,84 @@
...
-        transaction = tr;
-
-        return this;
-      },
-    }
-  );
-
-  return wrapper;
-}
+import { isNumber, isArray, chunk, flatten, assign } from 'lodash';
```
kibertoad: No changes in this file, just reformatted with new Prettier version

test/jake/migrate.js
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env jake
+'use strict';
+/* eslint-disable no-undef */
+
```
kibertoad: added .js extension for proper syntax colouring

test/jake/migrate.js
```diff
@@ -0,0 +1,125 @@
...
+    bin.addListener('stderr', (data) => (stderr += data.toString()));
+    bin.run();
+  });
+}
+
+const taskList = [];
+function test(description, func) {
+  const tmpDirPath = os.tmpdir() + '/knex-test-';
+  let itFails = false;
+  rimrafSync(tmpDirPath);
```
kibertoad: added temp directory deletion before each test

test/jake/migrate.js
```diff
@@ -0,0 +1,125 @@
...
+      )
+    )
+    .then(() =>
+      assertExec(
+        `grep exports.up ${temp}/migrations/*_create_rule_table.js`,
+        'Migration created with boilerplate'
+      )
+    ));
+
+test('Create a migration file without client passed', (temp) =>
```
kibertoad: new test

kibertoad:
@aurium Could you please review?

package.json
```diff
@@ -86,7 +86,7 @@
     "plaintest": "mocha --exit -t 10000 -b -R spec test/index.js && npm run tape",
     "prepublish": "npm run babel",
     "pre_test": "npm run lint",
-    "bin_test": "jake -f test/jake/*",
+    "bin_test": "jake -f test/jake/migrate.js",
```
kibertoad: at least locally it doesn't execute anything for me unless specific file (or directory with file called "Jakefile") is specified.

elhigu: Maybe that is platform dependent what it does (shell might expand those wildcards which may break it). I remember having to put wildcard notations in quotes at least with mocha to make globs to work. Like `-f "test/jake/*"`

src/seed/index.js
```diff
@@ -124,9 +124,11 @@ Seeder.prototype._waterfallBatch = function(seeds) {
     seed = require(name);
 
     // Run each seed file.
-    current = current.then(() => seed.seed(knex, Promise)).then(function() {
-      log.push(name);
-    });
+    current = current
+      .then(() => seed.seed(knex, Promise))
+      .then(function() {
```
elhigu: could be also changed to arrow function

kibertoad:
@elhigu Thanks! It was an insanely bad day of a pretty bad month, really appreciate something positive :D

bin/cli.js
```diff
@@ -37,8 +37,8 @@ function checkLocalModule(env) {
 }
 
 function mkConfigObj(opts) {
-  let envName = opts.env || process.env.NODE_ENV || 'development';
-  let useNullAsDefault = opts.client === 'sqlite3';
+  const envName = opts.env || process.env.NODE_ENV || 'development';
+  const useNullAsDefault = opts.client === 'sqlite3';
```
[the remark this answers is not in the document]
kibertoad: good point, it does.

package.json
kibertoad: You are probably right, I've checked logs for original PR that introduced CLI tests, and they do indeed show
```
> jake -f test/jake/*
☑ Create a migration file
☑ Run migrations
```

So I'll revert this change.

bin/cli.js
kibertoad: But how could it possibly work if `env.configuration` is either path to knexfile.js or resolved configuration object?

elhigu: buggily 😛

Issue #2897, opened 2018-11-08
when passing empty list in `.whereIn()`


HurSungYun:
# Environment

Knex version: 0.15.0
Database + version: mysql 5.7
OS: mac OS X

# Bug

I tried this
```
slots_db.client('table')
  .select()
  .where({ 'name': 'a' })
  .whereIn('release', []);
```

expected
```
select * from `table` where `name` = ?
```

but actually
```
select * from `table` where `name` = ? and 1 = ?
```

I believe it should be either the first one or throwing error.

I'd like to work on this issue.

wubzz:
Honestly I think its perfectly fine the way it is. When calling `.whereIn` you are expecting to return rows based on specific criteria, in this case a list of values. If that list happens to be empty, we shouldn't ignore the fact that it was still a criteria for the query to begin with. Right now knex solves this by doing a `1 = 0` query returning no rows at all.

To me this is the same as saying an undefined value in `.where({column: undefined})` should ignore the `column` criteria in the where clause, which feels very dangerous, especially in dynamic situations. Maybe a bad example since this throws `Undefined Bindings` error, but the point is the same.

HurSungYun:
@wubzz I agree with your point.

thanks for your explanation in detail :)

HurSungYun:
I think that throwing error can be a better solution for this situation.

In my use case, it makes business logic broken when ignoring empty list and returning nothing.

I believe it can be safer when knex throws an error. Users can realize their bugs more quickly, if we throw errors.

How do you think about it? @wubzz

wubzz:
I still think its wrong to throw an error. You could just simply do this:

```
const releaseIds = [];
const builder = slots_db.client('table')
  .select()
  .where({ 'name': 'a' });

if(releaseIds.length) {
 builder.whereIn('release', releaseIds);
}
```

Most use-cases consist around argument to `whereIn` being a dynamically populated array, which may very well be an empty array sometimes. Changing this to throw an error is not only a breaking change, but also a hazzle in practise. Doing so would result in __always__ having to do the example above unless your list of IDs is always __static__.

This is just my personal opinion, maybe better to get some more input.

@elhigu @tgriesser @kibertoad

kibertoad:
@wubzz I wouldn't call it a breaking change since current behaviour doesn't seem to be documented, and I don't think it's the one that most users would expect. 
Can you imagine a real-world use-case when you would actually prefer query to return nothing when some filtering criteria is empty?
[...]

PR #2908, opened 2018-11-15
add validation in `.offset()`

HurSungYun:
I believe `value` in `.offset(value)` must be integer like `.limit(value)`

However, there is no warning message for this.


HurSungYun:
@elhigu

https://github.com/tgriesser/knex/blob/master/test/unit/query/builder.js#L6457

I believe this test case covers my PR now.

Is there anything omitted?

elhigu:
Actually that case looks like that it has been designed so that passing null actually clears the offset, thus it shouldn't show warning. Also in case of warning it should check that warning is actually emitted. It can be done by using custom loggers in test.

Also I would prefer throwing an error instead of warning.

kibertoad:
@HurSungYun @elhigu Actually, passing null to reset sounds like a valid case, no?

HurSungYun:
We discussed something like this #2897 

It concluded that everything like this should be valid, so I believe this one also should be valid case

HurSungYun:
@elhigu 

I believe this test cases are enough for checking whether warning message is emitted.

elhigu:
Now this prevents using knex.raw / query builder as a parameter for offset, which has been working previously.

HurSungYun:
My bad. I'll revise it to work properly.

HurSungYun:
The code of `.limit()` is like this.
```
  // Only allow a single "limit" to be set for the current query.
  limit(value) {
    const val = parseInt(value, 10);
    if (isNaN(val)) {
      this.client.logger.warn('A valid integer must be provided to limit');
    } else {
      this._single.limit = val;
    }
    return this;
  },
```

I tried some tests in Node 8,

```
> client('test').select().limit(client.raw('3')).toSQL()
{ method: 'select',
  options: {},
  timeout: false,
  cancelOnTimeout: false,
  bindings: [ 3 ],
  __knexQueryUid: 'cd175d39-f411-4210-b242-519a0c295d47',
  sql: 'select * from `test` limit ?' }

> client('test').select().limit(client.raw('?', [3])).offset(undefined).toSQL()
{ method: 'select',
  options: {},
  timeout: false,
  cancelOnTimeout: false,
  bindings: [ 3 ],
  __knexQueryUid: 'b90d327a-54de-4e40-9ee4-4d6e03e058cf',
  sql: 'select * from `test` limit ?' }

> parseInt(client.raw('3'), 10)
3
```

So, it does not break using knex.raw as a parameter **if the evaluated result of knex.raw can be parsed to int** and if that cannot be parse to int, it is invalid input anyway. 

I would like to suggest two options for `.offset()` validation

```
offset(value) {
  if (
    value instanceof Builder ||
    value instanceof Raw ||
    isNil(value)
  ) {
    this._single.offset = value;
  } else {
    const val = parseInt(value, 10);
    if (isNaN(val)) {
      this.client.logger.warn('A valid integer must be provided to offset');
    } else {
      this._single.offset = val;
    }
  }
  return this;
},
// check all types of value and value = null, undefined case
```
OR
```
offset(value) {
  if (isNil(value)) {
    this._single.offset = value; // for resetting
  } else {
    const val = parseInt(value, 10);
    if (isNaN(val)) {
      this.client.logger.warn('A valid integer must be provided to offset');
    } else {
      this._single.offset = val;
    }
  }
  return this;
},
// check case of value = null, undefined 
```

I understand all this should be backward compatible, but I think it's weird that there's no validation code in contrast with `.limit()`

NeoyeElf:
when will this changes be releaed?

HurSungYun:
@NeoyeElf I fixed something but i don't know when this PR is merged

test/unit/query/builder.js
```diff
@@ -40,6 +40,32 @@ const clientsWithNullAsDefault = {
...
+    Object.assign({ client: 'mysql' }, customLoggerConfig)
+  ),
+  pg: new PG_Client(Object.assign({ client: 'pg' }, customLoggerConfig)),
+  'pg-redshift': new Redshift_Client(
+    Object.assign({ client: 'redshift' }, customLoggerConfig)
+  ),
+  oracledb: new Oracledb_Client(
+    Object.assign({ client: 'oracledb' }, customLoggerConfig)
+  ),
+  // sqlite3: new SQLite3_Client(
```
kibertoad: Should this be commented out?

test/unit/query/builder.js
```diff
@@ -40,6 +40,32 @@ const clientsWithNullAsDefault = {
   ),
 };
 
+var customLoggerConfig = {
```
kibertoad: would be better to convert all vars in this file to let/const

kibertoad:
Sorry for the delay in reviewing. This can be merged after couple of minor comments are addressed.

test/unit/query/builder.js
HurSungYun: changed to const

test/unit/query/builder.js
HurSungYun: there was no sqlite3 tests using sqlite3 a year ago.

maybe someone have missed test code so I did the same thing.

still some test does not include sqlite3, but it may be fixed in another PR.

anyway I un-commented it.

rjkhan:
what is the default value of offset()

kibertoad:
@rjkhan No offset, I believe.

PR #2914, opened 2018-11-17
Do not postprocess internal queries in Migrator

### LINKED ISSUE #2644: Undefined max_batch when using knexSnakeCaseWrappers (Objection)

kibertoad:
fixes #2644
Also fixes .withUserParams() method to work correctly with .queryBuilder()

### DIFF
--- a/src/migrate/Migrator.js
+++ b/src/migrate/Migrator.js
@@ -46,10 +46,12 @@ const CONFIG_DEFAULT = Object.freeze({
 // the migration.
 export default class Migrator {
   constructor(knex) {
-    this.knex = knex;
-    this.config = getMergedConfig(knex.client.config.migrations);
-    this.generator = new MigrationGenerator(knex.client.config.migrations);
+    // Clone knex instance and remove post-processing that is unnecessary for internal queries from a cloned config
+    this.knex = knex.withUserParams(knex.userParams);
+    this.knex.client.config.wrapIdentifier = null;
+    this.knex.client.config.postProcessResponse = null;
 
+    this.generator = new MigrationGenerator(knex.client.config.migrations);
     this._activeMigration = {
       fileName: null,
     };
--- a/src/migrate/Migrator.js
+++ b/src/migrate/Migrator.js
@@ -7,6 +7,7 @@ import {
   each,
   filter,
   get,
+  isFunction,
   isBoolean,
   isEmpty,
   isUndefined,
@@ -47,11 +48,14 @@ const CONFIG_DEFAULT = Object.freeze({
 export default class Migrator {
   constructor(knex) {
     // Clone knex instance and remove post-processing that is unnecessary for internal queries from a cloned config
-    this.knex = knex.withUserParams(knex.userParams);
+    this.knex = isFunction(knex)
+      ? knex.withUserParams(knex.userParams)
+      : Object.assign({}, knex);
     this.knex.client.config.wrapIdentifier = null;
     this.knex.client.config.postProcessResponse = null;
 
-    this.generator = new MigrationGenerator(knex.client.config.migrations);
+    this.config = getMergedConfig(this.knex.client.config.migrations);
+    this.generator = new MigrationGenerator(this.knex.client.config.migrations);
     this._activeMigration = {
       fileName: null,
     };
--- a/src/migrate/table-creator.js
+++ b/src/migrate/table-creator.js
@@ -10,21 +10,26 @@ export function ensureTable(tableName, schemaName, trxOrKnex) {
   const lockTableWithSchema = getLockTableNameWithSchema(tableName, schemaName);
   return getSchemaBuilder(trxOrKnex, schemaName)
     .hasTable(tableName)
-    .then(
-      (exists) =>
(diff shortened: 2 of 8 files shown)


src/migrate/table-creator.js
```diff
@@ -10,21 +10,26 @@ export function ensureTable(tableName, schemaName, trxOrKnex) {
...
     .hasTable(tableName)
-    .then(
-      (exists) =>
-        !exists && _createMigrationTable(tableName, schemaName, trxOrKnex)
-    )
-    .then(() => getSchemaBuilder(trxOrKnex, schemaName).hasTable(lockTable))
-    .then(
-      (exists) =>
+    .then((exists) => {
+      return !exists && _createMigrationTable(tableName, schemaName, trxOrKnex);
```
kibertoad: it's a huge pain to debug arrow functions without {} and prettier really butchers formatting for them, hence the change

src/util/make-knex.js
```diff
@@ -10,24 +10,23 @@ import batchInsert from './batchInsert';
 export default function makeKnex(client) {
   // The object we're potentially using to kick off an initial chain.
   function knex(tableName, options) {
```
kibertoad: There are more changes here than I am comfortable with, TBH. Main reason - there was actually a creeping bug after withUserParams change - invoking knex('tableName') would silently create a builder from wrong (original) knex instance. While figuring out a solution I've implemented a stricter separation between knex() function (which should be as lightweight as possible) and the actual knex execution context (that knex function proxies to). I'm still not happy about our magical knex entity which is function, object and class instance at the same time, but at least there is clearer separation of what is responsible for what now)

test/unit/migrate/migrations/blank_migration.js
```diff
@@ -0,0 +1,3 @@
+exports.up = function(knex) {};
```
kibertoad: having this kinda violates the rule of unit tests having no dependencies, but mocking out require() call for a migration or restructuring migrator to be able to mock out entire sequences is a huge pain, and we don't really need an integration test that would be going over all dialects here at all as we are not testing any db functionality here.

src/migrate/Migrator.js
```diff
@@ -46,10 +47,15 @@ const CONFIG_DEFAULT = Object.freeze({
...
 export default class Migrator {
   constructor(knex) {
-    this.knex = knex;
-    this.config = getMergedConfig(knex.client.config.migrations);
-    this.generator = new MigrationGenerator(knex.client.config.migrations);
-
+    // Clone knex instance and remove post-processing that is unnecessary for internal queries from a cloned config
+    this.knex = isFunction(knex)
+      ? knex.withUserParams(knex.userParams)
+      : Object.assign({}, knex);
```
heisian: This is great! I attempted to clone knex but unsuccessfully in some prototyping code - I would receive an error (forgot what it was) attempting to make queries with my cloned version. Good to see how it is done here.

kibertoad: Admittedly I'm abusing .withUserParams function, but given that it includes all the necessary black magic for cloning to actually produce a fully usable knex instance, that's the best option we have now. Maybe we should eventually replace .clone function with what .withUserParams currently does and call it under-the-hood.

heisian:
@elhigu When do you think the next tagged release will be?

kibertoad:
@heisian Soon. I'm updating changelog and package.json today, but I also hope to squeeze https://github.com/tgriesser/knex/pull/2861 in - which depends on external project merging and releasing; if it doesn't happen tomorrow, probably we are going to release without it.

heisian:
@kibertoad dope, thank you

dlangerenken:
just to be sure: Is this also including the wrapIdentifier function for migration that also converts knex_migrations_lock and other internal tables? This does not seem to be fixed in 0.16.3

elhigu:
@dlangerenken Sounds like you might be right. Looks like implementation tries to reset wrapIdentifier to default too, but it does not seem to be tested by https://github.com/tgriesser/knex/pull/2914/files#diff-09c603a78f9c37fe1a94f869d4a83405R16

That test case needs to be extended to cover it too.

PR #2918, opened 2018-11-20
Update changelog and version number

kibertoad:


elhigu:
thanks! should I release it now?

PR #2927, opened 2018-11-23
remove useless 'use strict'

bertho-zero:
'use strict' and import/export are not meant to be together.


kibertoad:
@bertho-zero Thanks!

PR #2932, opened 2018-11-27
Add new command to print migrations

FabricioFFC:
This PR adds the `print` option to the `migrate` command. When this option is used the migrations will be printed as SQL, so who don't know the knex interface but know SQL (ex: DBA) can verify the migrations created.

screenshot_20181130_111949

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -26,6 +26,15 @@ function success(text) {
   process.exit(0);
 }
 
+function checkMigrationDirection(direction) {
+  if (!direction) return;
+
+  if (direction !== 'up' && direction !== 'down') {
+    console.log(chalk.red('Direction is invalid'));
+    exit('Try to pass: up or down.');
+  }
+}
+
 function checkLocalModule(env) {
   if (!env.modulePath) {
     console.log(
@@ -239,8 +248,12 @@ function invoke(env) {
   commander
     .command('migrate:print')
     .description('Print all migrations in directory')
-    .option('--direction', 'Specify migration direction to print. It can be "up" or "down"')
+    .option(
+      '--direction',
+      'Specify migration direction to print. It can be "up" or "down"'
+    )
     .action(function() {
+      checkMigrationDirection(argv.direction);
       const direction = argv.direction || 'all';
 
       pending = initKnex(env)
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -237,14 +237,14 @@ function invoke(env) {
     });
 
   commander
-    .command('migrate:printAll')
+    .command('migrate:print')
     .description('Print all migrations on directory')
     .option('--direction', 'Specify migration direction to print')
     .action(function() {
       const direction = argv.direction || 'all';
 
       pending = initKnex(env)
-        .migrate.printAll(direction)
+        .migrate.print(direction)
         .then(function() {
           process.exit(0);
         })
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -238,8 +238,8 @@ function invoke(env) {
 
   commander
     .command('migrate:print')
-    .description('Print all migrations on directory')
-    .option('--direction', 'Specify migration direction to print')
+    .description('Print all migrations in directory')
+    .option('--direction', 'Specify migration direction to print. It can be "up" or "down"')
     .action(function() {
       const direction = argv.direction || 'all';
 
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -237,6 +237,21 @@ function invoke(env) {
     });
 
   commander
+    .command('migrate:printAll')
+    .description('Print all migrations on directory')
+    .option('--direction', 'Specify migration direction to print')
+    .action(function() {
+      const direction = argv.direction || 'all';
+
+      pending = initKnex(env)
+        .migrate.printAll(direction)
+        .then(function() {
+          process.exit(0);
+        })
+        .catch(exit);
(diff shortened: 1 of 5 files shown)


bin/cli.js
```diff
@@ -25,6 +25,15 @@ function success(text) {
   process.exit(0);
 }
 
+function checkMigrationDirection(direction) {
+  if (!direction) return;
```
kibertoad: Please wrap in {}

bin/cli.js
```diff
@@ -25,6 +25,15 @@ function success(text) {
...
+
+  if (direction !== 'up' && direction !== 'down') {
+    console.log(chalk.red('Direction is invalid'));
+    exit('Try to pass: up or down.');
```
kibertoad: "Supported directions: up, down" might be more clear.

bin/cli.js
```diff
@@ -195,6 +204,25 @@ function invoke(env) {
...
 
+  commander
+    .command('migrate:print')
+    .description('Print all migrations in directory')
+    .option(
+      '--direction',
+      'Specify migration direction to print. It can be "up" or "down"'
+    )
+    .action(function() {
+      checkMigrationDirection(argv.direction);
```
kibertoad: I would suggest lowercasing direction before doing anything with it

bin/cli.js
```diff
@@ -195,6 +204,25 @@ function invoke(env) {
...
+      const direction = argv.direction || 'all';
+
+      pending = initKnex(env)
+        .migrate.print(direction)
+        .then(function() {
```
kibertoad: arrow function is preferable

bin/cli.js
```diff
@@ -195,6 +204,25 @@ function invoke(env) {
         .catch(exit);
     });
 
+  commander
+    .command('migrate:print')
```
kibertoad: new command needs test. see https://github.com/tgriesser/knex/tree/master/test/jake for an example (https://github.com/tgriesser/knex/blob/master/test/jake/knexfile.js has a recommended structure, migrate.js will be cleaned up in the future)

src/migrate/index.js
```diff
@@ -150,6 +150,39 @@ export default class Migrator {
...
+  print(direction) {
+    global.print = true;
+
+    return migrationListResolver
+      .listAll(this._absoluteConfigDir(), this.config.loadExtensions)
+      .then((migrationFiles) => {
+        let current = Promise.bind({ failed: false, failedOn: 0 });
+
+        each(migrationFiles, (migrationFile) => {
+          const directory = this._absoluteConfigDir();
```
kibertoad: how would this work when directory is an array?

kibertoad: nvm, this is a root directory

kibertoad:
Please resolve conflicts with the master branch as well :)

bin/cli.js
```diff
@@ -236,6 +247,36 @@ function invoke(env) {
...
+    .description('Print all migrations in directory')
+    .option(
+      '--direction',
+      'Specify migration direction to print. It can be "up" or "down"'
+    )
+<<<<<<< HEAD
```
kibertoad: should this be committed?..

FabricioFFC: sorry @kibertoad this was a wip commit. I will notify you when we finish the changes requested.

bin/cli.js
FabricioFFC: @kibertoad I did the integration tests: https://github.com/tgriesser/knex/pull/2932/files#diff-01d824a84fdd0b90ebaadc424247da50R334

I tried to create a test using the jake structure you are using, but I had some problems (listed below) and with integration tests the assert is better.

* in my local env the `bin_test` is not working I needed to run pointing to a file (ex: `jake -f test/jake/migrate.js`) to make it work
* the tests seem to run parallely and I need to remote the `db` created before to run the test using `print`, because my assert would be assert that the `db` was not created (what is a weak assertion)

FabricioFFC:
@kibertoad now I think is everything good to go :) thank for your review and let me know if we are missing something

src/migrate/MigrationGenerator.js
```diff
@@ -24,6 +24,44 @@ export default class MigrationGenerator {
...
+          );
+          const migration = require(directory + '/' + migrationFile.file);
+
+          if (direction === 'up' || direction === 'all') {
+            current = current.then(() => {
+              /* eslint-disable no-console */
+              console.log(`\n${migrationFile.file}`);
+              console.log('======== UP ==========');
+              /* eslint-enable no-console */
+              return migration['up'](knex, Promise);
```
kibertoad: Wait, I'm missing something. Why are we actually executing migrations?

src/runner.js
```diff
@@ -132,6 +132,13 @@ assign(Runner.prototype, {
     this.builder.emit('query', assign({ __knexUid, __knexTxId }, obj));
 
     const runner = this;
+
+    if (global.print) {
```
kibertoad: this looks quite intrusive and relies on global state. Can't this be rewritten to be some param for .migrate() operation?

kibertoad: By the looks of it, currently client.js instantiates new runner and then passes builder to it. Cleaner way to implement this functionality would be to have a "dry-run" runner that prints out queries instead of executing them; hence this could be a parameter in client configuration based on which client would resolve the kind of runner it needs to instantiate.

PR #2933, opened 2018-11-28
Fix knexfile resolution

MaxEvan:
Knexfile resolution wasn't working for me so I came up with this alternate version which seems to work just fine!

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -56,22 +56,21 @@ function mkConfigObj(opts) {
 function initKnex(env, opts) {
   checkLocalModule(env);
 
-  if (!opts.knexfile) {
-    if (opts.client) {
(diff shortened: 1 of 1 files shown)


kibertoad:
@MaxEvan Thank you for the contribution! Could you please update https://github.com/tgriesser/knex/blob/master/test/jake/knexfile.js to include test that would be failing without your changes?

kibertoad:
(note that if you are going to be running these tests not on a Linux box, you would need to replace wildcard ` "bin_test": "cross-env KNEX_PATH=../knex.js KNEX=bin/cli.js jake -f test/jake/*",` in package.json with path to knexfile.js Jake test directly)

MaxEvan:
Yeah closing pull request now

PR #2934, opened 2018-11-29
Apply postprocessing in migrations

kibertoad:


kibertoad:
huh, I have no idea why this is not failing, it used to locally.

src/dialects/oracledb/query/compiler.js
```diff
@@ -296,10 +296,7 @@ _.assign(Oracledb_Compiler.prototype, {
     let returningClause = '';
     let intoClause = '';
 
-    if (
-      _.isEmpty(updates) &&
-      typeof this.single.update !== 'function'
-    ) {
+    if (_.isEmpty(updates) && typeof this.single.update !== 'function') {
```
kibertoad: prettier picked this one up for some reason

src/util/make-knex.js
```diff
@@ -54,6 +54,26 @@ function initContext(knexFn) {
       return this.client.ref(ref);
     },
 
+    disableProcessing() {
```
elhigu: Would be good to add some comment that these should not be added documentation to be a part to knex public API before we have designed better API names and parameters for this.

test/unit/migrate/Migrator.js
```diff
@@ -31,4 +36,35 @@ describe('Migrator', () => {
...
+      knex.destroy();
+    });
+
+    it('latest', (done) => {
+      knex = Knex({
+        ...sqliteConfig,
+        connection: ':memory:',
+        migrationSource,
+        postProcessResponse: () => {
+          done();
```
elhigu: This test actually calls done, before execution of migrations has ended. Because that is separate inmemory database, it probably won't cause any problems. Except test might have been ended already before that expectation below has finished execution.

src/util/make-knex.js
elhigu: And also to mention why these exist (that it was needed for migrations to be able to run those internal queries, with consistent identifier naming).

PR #2935, opened 2018-12-01
Fix CWD parameter support for CLI

kibertoad:


bin/cli.js
```diff
@@ -11,6 +11,7 @@ const argv = require('minimist')(process.argv.slice(2));
 const fs = Promise.promisifyAll(require('fs'));
 const cliPkg = require('../package');
 const { resolveClientNameWithAliases } = require('../lib/helpers');
```
joselcvarela: @kibertoad 
Hi again,
I just install your branch with npm like this ```npm i tgriesser/knex#fix/cli-cwd``` and when run ```npx knex migrate:latest``` inside knexfile.js folder it brokes on this line saying ```Cannot find module '../lib/helpers'```.
While I debug it, find out that it is trying to resolve this file relative to my project folder instead of ```node_modules/knex```

joselcvarela: @kibertoad 
Sorry, I was wrong because I was trying to run without lib folder. I just forked and change to your branch, ran ```npm run babel``` and find out the problem is in commander package, because I am passing --knexfile as argument like this ```npx knex --knexfile ./database/knexfile.js migrate:latest something``` and opt.knexfile is undefined.
So as a fix I just put './knexfile.js' as default value in commander knexfile argument and now it works

kibertoad:
@joselcvarela Can you try "--knexfile=./database/knexfile.js" and see if it makes any difference? Also try putting it after "migrate:latest". Would appreciate if you could report back if either of them would work.

PR #2944, opened 2018-12-10
Do not reset prior Promise overrides on knex import

### LINKED ISSUE #2942: v0.16.1 messes with global Promise
```js
global.Promise = require('bluebird');
console.log(Promise.map); // [Function]
require('knex');
console.log(Promise.map); // undefined
```

kibertoad:
fixes #2942 
This makes behavior consistent with Knex before 0.16.x

### DIFF
--- a/knex.js
+++ b/knex.js
@@ -5,7 +5,14 @@
 //     For details and documentation:
 //     http://knexjs.org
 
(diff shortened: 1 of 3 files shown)


kibertoad:
Do we consider this a regression that warrants a version release?

PR #2946, opened 2018-12-11
Support for joins in update statements

### LINKED ISSUE #2796: Update operations do not support joins
# Environment

Knex version: 0.15.2

# Bug

```
  await knex('table)
    .innerJoin('table2', function() {
      this.on(
        `table.id`,
        `table2.id`
      ).andOn(`table.field`, 'value')
    })
    .update({
      table.field2: 'newValue,
    })
```

This is going to execute without an error, but entire join part will be ignored and all entries will be silently updated. 

https://github.com/tgriesser/knex/issues/1534 suggests either using raw or subquery, but neither of these solutions is particularly elegant, and API-wise there is no reason why this shouldn't work, so I would consider this a bug.

kibertoad:
fixes #2796 
Work sponsored by Devbridge Group.

### DIFF
--- a/src/dialects/postgres/query/compiler.js
+++ b/src/dialects/postgres/query/compiler.js
@@ -34,6 +34,7 @@ assign(QueryCompiler_PG.prototype, {
 
   // Compiles an `update` query, allowing for a return value.
   update() {
+    const join = this.join();
     const withSQL = this.with();
     const updateData = this._prepUpdate(this.single.update);
     const wheres = this.where();
@@ -43,6 +44,7 @@ assign(QueryCompiler_PG.prototype, {
         withSQL +
         `update ${this.single.only ? 'only ' : ''}${this.tableName} ` +
         `set ${updateData.join(', ')}` +
+        (join ? ` from ${this.tableName} ${join}` : '') +
         (wheres ? ` ${wheres}` : '') +
         this._returning(returning),
       returning,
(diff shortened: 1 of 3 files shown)


kibertoad:
Ideally this needs integration tests to make sure that generated sql actually does what we hope it does, but

a) This doesn't break anything reasonable that worked before;
b) PostgreSQL update with joins were broken in worst possible way previously - silently not doing what is expected, so even if some cases break - I would call it an improvement;
c) Our integration tests are quite messy, difficult to setup and leak most of the state, so before introducing integration tests that rely on a new schema, I would prefer to do at least a little bit of cleanup that is going to take a while.

elhigu:
I'm gonna check this out soon. This has been very much wanted feature during the years! 👍

kibertoad:
@elhigu Finally got around to actually testing this and apparently it is slightly wrong.

In order to work it should be either 
```
  FROM table1 t1
  join table2
  on table2.id = t1.fk
  where t1.somecolumn = ?
```

or

```
  UPDATE table1 SET column = ? FROM table1 
  WHERE table1.id = table2.fk
  AND table1.somecolumn = ?
```


Neither of which are quite straightforward to implement within current code, but I'll try to work it out.
And yeah, we do need integration tests, I've started work on that in a separate branch.

timendez:
@kibertoad @elhigu Is this feature dead or was it implemented in a different PR?

kibertoad:
@timendez "Half-baked" is a better description than "dead", it's not correct for all DBs and there are no integration tests. I want to complete it when I have some time, but if you want to volunteer to finish the deed, that'd tremendously speed it up.

PR #2949, opened 2018-12-11
Include @types/bluebird as dependency

### LINKED ISSUE #2948: Could not find a declaration file for module 'bluebird'
# Environment

Knex version: 0.16.2
Database + version: any
OS: any

# Bug

The 0.16 release of Knex includes TypeScript definitions, so that `@types/knex` is no longer required.

However, `package.json` fails to mention `@types/bluebird` as dependency: `types/knex.d.ts` fails to compile in line 15 without that: `import Bluebird = require('bluebird');`

The fix would be to include `@types/bluebird` along with `bluebird` in the `dependencies` of Knex 0.16.

sgoll:
This fixes #2948 by including `@types/bluebird` as dependency of Knex.

We must use `dependencies` rather than `devDependencies` so that `@types/bluebird` gets installed when `knex` is added as regular dependency of another project, and not just when developing Knex itself—since the type definitions are exported as-is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.0.0",
+    "@types/bluebird": "^3.5.3",
     "bluebird": "^3.5.3",
     "chalk": "2.4.1",
     "commander": "^2.19.0",


kibertoad:
Thanks!

PR #2954, opened 2018-12-13
add test that clearing offset

HurSungYun:
According to https://github.com/tgriesser/knex/pull/2908,

clearing offset is valid case for Knex.js, but there's no test for it.


kibertoad:
Thanks!

PR #2959, opened 2018-12-17
Resolve migrations and seeds relatively to knexfile directory

### LINKED ISSUE #2952: knex migration doesn't find the config file
# Environment

Knex version: 0.16.2
Database + version: postgres 11
OS: mac os 10.13.6

# Bug

the knex config file: `/tools/knex/knex.js`
```
require('dotenv').config()

const connection = {
  host: process.env.PGHOST,
  port: process.env.PGPORT,
  database: process.env.PGDATABASE,
  user: process.env.PGUSER,
  password: process.env.PGPASSWORD
}

const knexConfig = {
  client: 'pg',
  connection,
  migrations: {
    directory: './tools/knex/migrations'
  },
  seeds: {
    directory: './tools/knex/seeds'
  }
}

module.exports = knexConfig
```
when I run `npx knex --knexfile=./tools/knex/config.js migrate:rollback`, 
it makes an error `Error: ENOENT: no such file or directory, scandir '/Volumes/HD/Documents/my-app/migrations'`

from the config file it should look for migrations in `./tools/knex/migrations` and not in `./migrations`

This was working well in previous knex version 0.15.2

kibertoad:
fix #2952

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug fixes:
 
 - @babel/polyfill loaded multiple times #2955
+- Resolve migrations and seeds relatively to knexfile directory when specified (the way it used to be before 0.16.1) #2952
 
 # 0.16.2 - 10 Dec, 2018
 
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -57,6 +57,8 @@ function initKnex(env, opts) {
   // If knexfile is specified
   else {
     const resolvedKnexfilePath = path.resolve(opts.knexfile);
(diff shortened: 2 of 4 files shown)


elhigu:
So does passing `--cwd` still change relative directory where migrations etc. are tried to be found?

kibertoad:
@elhigu CWD param is executed prior to resolving knexfile path, so if knexfile path is relative, it will be applied to specified CWD. Migrations and seeds are relative to knexfile resolved against passed CWD.

elhigu:
Ok... so it actually does logically what one would expect --cwd to do 👍  maybe client docs could explain with detail, what those different switches actually does some day :)

PR #2960, opened 2018-12-17
Fix typings

serprex:
Updated the typings with ways which knex's API seems to allow for given our existing code. These were previously not covered by DefinitelyTyped


AmauriAires:
I reviewed it and made a few changes. It should work properly now.

kibertoad:
Thanks!

AmauriAires:
You're welcome.

Issue #1937, opened 2017-02-24
Implement SKIP LOCKED


marianomerlo:
Since POSTGRESQL 9.5, there is an `SKIP LOCKED` option for `SELECT FOR SHARE/FOR UPDATE` that is extremely helpful.

It would be useful to support this.

Thanks

honestserpent:
Would be really useful

kukiric:
I'll take this one. I have already implemented the functionality for PostgreSQL with some unit + integration tests, but I'll also try to get MySQL/MariaDB support in the PR as well for completeness.

marianomerlo:
Thank you so much for this! Closing issue since you all rock!

PR #2961, opened 2018-12-18
WIP: Implement "skipLocked()" and "noWait()"

kukiric:
This PR aims to resolve issue #1937.

It adds two chainable methods into the query builder: skipLocked() and noWait(). They can only be used in a select-like query, and only after a lock mode has been specified with forShare() or forUpdate(). These methods simply append the requested lock mode onto the query, leaving the actual lock handling to the RDMS.

The main use case for these methods is to create a job queue-type system in the database, where each job can only be taken by a single process or thread. Without them, it's possible to create deadlocks with multiple consumers taking exclusive locks on the same rows, when skipping locked rows would otherwise be more ideal.

Currently, this only works with the PostgreSQL dialect. I wanted to add support for MySQL and MariaDB as well, but I can't get the test suite to connect to the latest version of those databases (the client drivers don't seem to support the new authentication mode added in MySQL 8.0), and older versions don't support these statements.

As a WIP Pull Request, I'm open to comments and suggestions. Help is also appreciated.

TODO:

- [x] PostgreSQL support
- [x] MySQL/MariaDB support (currently failing tests)
- [x] Update knex.d.ts
- [x] Add documentation


src/query/builder.js
```diff
@@ -1061,6 +1061,42 @@ assign(Builder.prototype, {
...
+    if (!includes(['pluck', 'first', 'select'], _method)) {
+      throw new Error(`Cannot chain .skipLocked() on "${_method}" query!`);
+    }
+    if (!includes(['forShare', 'forUpdate'], this._single.lock)) {
+      throw new Error(
+        '.skipLocked() can only be used after a call to .forShare() or .forUpdate()!'
+      );
+    }
+    if (this._single.waitMode) {
+      throw new Error('.skipLocked() cannot be used with .noWait()!');
```
kibertoad: is error still correct if `waitMode === 'skipLocked'`?

src/query/builder.js
```diff
@@ -1061,6 +1061,42 @@ assign(Builder.prototype, {
...
+    }
+    this._single.waitMode = 'skipLocked';
+    return this;
+  },
+
+  // Causes error when acessing a locked row instead of waiting for it to be released.
+  noWait() {
+    const { _method } = this;
+    if (!includes(['pluck', 'first', 'select'], _method)) {
```
kibertoad: any reasons why duplicate code can't be reused? do you envision it branching out eventually?

test/integration/builder/selects.js
```diff
@@ -1225,5 +1225,83 @@ module.exports = function(knex) {
...
+                });
+            })
+            .then((res) => {
+              expect(
+                'The query should have been cancelled when trying to select a locked row'
+              ).to.be.false;
+            })
+            .catch((err) => {
+              // TODO: handle errors from more databases
+              if (knex.client.driverName === 'pg') {
```
kibertoad: wouldn't this swallow assertion errors on all other dbs?

src/query/builder.js
kukiric: I'm actually not sure how to best handle this. I just copied this piece from the first() method, since all databases only allow "skip locked" and "nowait" on select queries. I could, however, just leave it up for the database driver to return the error directly to the calling code, without validating it first.

src/query/builder.js
kukiric: That's a good point, I'll make the check a bit more explicit.

test/integration/builder/selects.js
kukiric: Yes, I should probably make this test fail for unsupported databases until a check is added at the beginning of the method.

kukiric:
The mysql2 driver worked with MariaDB, but it seems something is busted with my system since I still get a "client does not support authentication protocol" error when trying to use either driver with the MySQL 8.0 docker image. MySQL 5.7 still works, but it doesn't support these statements.

Also, I just tested this with MariaDB, and I've found out that it supports "no wait", but not "skip locked" like MySQL. How should the tests be handled in this scenario?

kukiric:
Pinging @kibertoad 

Should I drop MySQL + MariaDB support for this PR? The CI set up for this repo doesn't run a supported database version, and it's really inconvenient that MariaDB doesn't support this feature. PostgreSQL works like a charm, though.

kibertoad:
@kukiric Better to disable tests on these DBs temporary with ToDo to enable again when we use newer DB versions for testing. Extra bonus points if you could open an issue to add MySQL 8.0 to our test set.

kukiric:
@kibertoad I'll open the issue later today. Currently, Travis doesn't support MySQL 8.0 directly, nor a recent enough Ubuntu version that can install it via apt. Also, both the mysql and mysql2 packages for node lack support for the new password algorithm that they added in 8.0.4. So for now, this is Postgres-only, with noWait() also incidentally working on newer versions of MariaDB.

test/integration/builder/selects.js
```diff
@@ -1225,5 +1225,103 @@ module.exports = function(knex) {
...
+              ).to.be.false;
+            })
+            .catch((err) => {
+              switch (knex.client.driverName) {
+                case 'pg':
+                  expect(err.message).to.contain(
+                    'could not obtain lock on row'
+                  );
+                  break;
+                case 'mysql':
```
kibertoad: can't you just return in the beginning of the test in case of incompatible db?

kibertoad:
@kukiric Heya! Any chance this PR could be finished? (skipping all the incompatible DBs)

kukiric:
Sorry, I got busy and forgot to finish the PR. I'll wrap it up with docs + error messages on other DB drivers this weekend.

kibertoad:
Awesome!

test/integration/builder/selects.js
kukiric: Yes, incompatible DBs are now skipped. The re-throw is left in place in case someone adds another database to the compatible list, but forgets to create a specific test case for it.

kukiric:
The removed functionality is just a function that returns the string to append to the SQL, so it can be easily put back in when tests are in. It's fine by me if you want to re-enable it before merging, but I don't want to lead people into "here be dragons" territory when dealing with databases.

kukiric:
I guess I'll leave that up to the documentation. I'm going to write the the first draft and make a PR for it as well.

src/query/compiler.js
```diff
@@ -542,6 +543,23 @@ assign(QueryCompiler.prototype, {
...
+  // Compiles the wait mode on the locks.
+  waitMode() {
+    if (this.single.waitMode) {
+      return this[this.single.waitMode]();
+    }
+  },
+
+  // Fail on unsupported databases
+  skipLocked() {
+    throw new Error('.skipLocked() is currently only supported on PostgreSQL');
```
kibertoad: Should it mention PostgreSQL?

kibertoad: Nvm, I'm tired and misreading stuff.

kukiric:
Documentation PR is up at https://github.com/tgriesser/knex/pull/2961

kibertoad:
@kukiric Would you consider finishing this, now that MySQL 8 is used in CI by default?

kukiric:
Sure! I kinda forgot about it as I solved the original issue with a raw query, and moved onto other things, but I can pick this up again.

Should I update the branch by merging or rebasing master, or leave it alone until it's done?

kibertoad:
@kukiric Either approach works, we squash PRs anyway, so whatever is more convenient for you.

kukiric:
MySQL support is back, but it seems to not be locking rows correctly with a limit clause on the latest (Docker-published) version.

src/query/builder.js
kukiric: @kibertoad handled in commit 18cccab, what do you think?

kukiric: I'm also thinking about moving the "isSelectQuery" checks to the proper "forShare" and "forUpdate" methods

test/integration/builder/selects.js
```diff
@@ -1247,5 +1247,141 @@ module.exports = function(knex) {
...
+            })
+            .then((res) => {
+              expect(res).to.be.empty;
+            });
+        });
+    });
+
+    it('forUpdate().noWait() should reject immediately when a row is locked', function() {
+      if (
+        knex.client.driverName !== 'pg' &&
```
kibertoad: `!['pg', 'mysql].includes(knex.client.driverName)` might be easier to read, but it's a matter of taste :)

test/integration/builder/selects.js
```diff
@@ -1247,5 +1247,141 @@ module.exports = function(knex) {
...
+        return;
+      }
+
+      const rowName = 'row for noWait() test';
+      return knex('test_default_table')
+        .insert([
+          { string: rowName, tinyint: 1 },
+          { string: rowName, tinyint: 2 },
+        ])
+        .then(() => {
```
kibertoad: btw, we dropped support for Node 6, so you may use async/await if you like.

src/query/builder.js
```diff
@@ -1093,6 +1091,40 @@ assign(Builder.prototype, {
...
+  noWait() {
+    if (!this._isSelectQuery()) {
+      throw new Error(`Cannot chain .noWait() on "${this._method}" query!`);
+    }
+    if (!this._hasLockMode()) {
+      throw new Error(
+        '.noWait() can only be used after a call to .forShare() or .forUpdate()!'
+      );
+    }
+    if (this._single.waitMode === 'skipLocked') {
```
kibertoad: probably these could be constants?

kibertoad:
I'm OK with merging it after code review comments are addressed.

test/integration/builder/selects.js
kukiric: It's a bit more symbol-heavy, but it's a trivial change nonetheless. Though every other test already starts like this, with each DB check in a separate line.

test/integration/builder/selects.js
kukiric: That would be great actually, since the callback nesting makes the test unnecessarily harder to read.

src/query/builder.js
kukiric: I agree, but I can't find a good place for query builder constants. Is it worth it to create one now, or would this be more fitting for a future refactor?

test/integration/builder/selects.js
kibertoad: It's very much a matter-of-taste. Feel free to keep the original version if you like it more or prefer to be consistent.

src/query/builder.js
kibertoad: We don't get much refactoring PRs, as people have no good reason for submitting these :D. Would be awesome if you created something for this.

test/integration/builder/selects.js
kukiric: I'll keep it that way, thanks

src/query/builder.js
kukiric: Ok, I'll try

kukiric: Made it a separate module in c0b1cd4, how do you like it?

kibertoad:
Thanks a lot!

kibertoad:
Feature is available in 0.18.4

kibertoad:
@kukiric Apparently there is a problem with tests,

```
      promise.then(() => {
        expect(
          'The query should have been cancelled when trying to select a locked row with .noWait()'
        ).to.be.false;
```

Is never returned or awaited, so promise ends up being an unhandled rejection behind the scenes.
You can see this here: https://travis-ci.org/tgriesser/knex/jobs/572000778

Any chance you could fix either the test or the implementation, whichever is needed to make sure it passes? (simple awaiting is not enough, promise does end up reaching unexpected state)

You can see error here: https://travis-ci.org/tgriesser/knex/jobs/572000778

And if you want rejected promises to fail your tests, you can use #3393 as a base.

kukiric:
Sure, give me a day to submit a PR to fix it.

kibertoad:
@kukiric Much appreciated!

kukiric:
@kibertoad PR submitted: #3398

PR #2963, opened 2018-12-20
Typescript type definition for `.asCallback()`

HurSungYun:
In Knex version 0.16.2,

we are using Knex.js with Typescript + callback style code.

As I upgrade Knex version 0.15.0 -> 0.16.2, now it cannot be compiled typescript files anymore because of no type definition for `.asCallback()`.

Therefore, I added `.asCallback()` type definition in `ChainableInterface`

### DIFF
--- a/types/knex.d.ts
+++ b/types/knex.d.ts
@@ -498,6 +498,7 @@ declare namespace Knex {
       writable: T,
       options?: { [key: string]: any }
     ): stream.PassThrough;
+    asCallback(callback: Function): this;
   }
 
   interface Transaction extends Knex {


elhigu:
So was this removed from typings between 0.15 and 0.16 or how it has gone missing?

HurSungYun:
@elhigu I believe it's missed.

Type support has been imported in this repo after this PR. https://github.com/tgriesser/knex/pull/2845

However, `.asCallback()` type definition is not exists.

elhigu:
I suppose this is fine.

Funnily enough that `ChainableInterface extends Bluebird<any>` and bluebird itself has also method called `.asCallback`.  So is bluebird typings also missing that?

HurSungYun:
@elhigu I have found what is wrong.

In Knex@0.16.2, there's no "@types/bluebird" dependencies.

In Knex@0.16.3, there is type dependency for "@types/bluebird".

I tried this in Knex@0.16.2, so it was a problem. and tried to compile in 0.16.3 it has no problem.

So this PR might be mean nothing :(

PR #2969, opened 2018-12-21
Respect "loadExtensions" configuration

walkerburgin:
Upgrading from `^0.15.2` -> `^0.16.3` caused the `"loadExtensions"` option to no longer be respected. I think this is the cause, I'm not particularly familiar with the codebase.


kibertoad:
Ouch. Good catch!

Issue #2972, opened 2018-12-25
Feature Request: seed:run should print the file name of the failing seed


Wintereise:
# Environment

Knex version: 0.15.2
Database + version: PostgreSQL 10.6
OS: Ubuntu Linux

# Feature discussion / request

1. Why

When `seed:run` is called and something goes wrong (perhaps due to a DB constraint), Knex does NOT print which exact seed file the issue happened in. This leaves you with a wild goose chase.

2. Explain what kind of feature would support this

Simply printing the file the error was generated while running + db + table-name would likely be plenty.

3. Give some API proposal, how the feature should work

See above.

# Demonstration

```
λ  knex seed:run
error: 値は型character varying(64)としては長すぎます
    at Connection.parseE (G:\xen\manager\node_modules\pg\lib\connection.js:554:11)
    at Connection.parseMessage (G:\xen\manager\node_modules\pg\lib\connection.js:379:19)
    at Socket.<anonymous> (G:\xen\manager\node_modules\pg\lib\connection.js:119:22)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
    at Socket.Readable.push (_stream_readable.js:208:10)
    at TCP.onread (net.js:607:20)
```

From the above output, you really can't tell which one of my 20+ seed files this actually happened in (or which field for that matter). All it says is that the length validation failed in one of my many VARCHAR fields with length set to 64.

This issue is largely the same for the MySQL2 driver as well, and feels like something that should be addressed in core knex.

kibertoad:
@Wintereise Could you please try 0.16.3? I think it was addressed there.

kibertoad:
Ah, seeds. It should work properly for migrations, but it could be that it's not the case for seeds. I'll look into it.

Wintereise:
Yep, same behavior on 0.16.3 @kibertoad 

```λ  knex seed:run
error: 値は型character varying(64)としては長すぎます
    at Connection.parseE (G:\xen\manager\node_modules\pg\lib\connection.js:554:11)
    at Connection.parseMessage (G:\xen\manager\node_modules\pg\lib\connection.js:379:19)
    at Socket.<anonymous> (G:\xen\manager\node_modules\pg\lib\connection.js:119:22)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
[...]

PR #2973, opened 2018-12-25
Display name of a failed seed

kibertoad:
fixes https://github.com/tgriesser/knex/issues/2972


kibertoad:
@Wintereise np! Thank you for raising the issue!

PR #2979, opened 2019-01-03
Typescript type ability to select aliases with ColumnName

richrout:
I know it's not documented, nor do I know if this is something you guys "officially" support. But we've been using them with PostgreSQL and knex seems to handle it fine.

Basically allows for something like:
```
knex('users').select({
   'zap': 'users.name',
   'foo': knex.raw('select bar from boz')
});
```

### DIFF
--- a/types/knex.d.ts
+++ b/types/knex.d.ts
@@ -216,7 +216,7 @@ declare namespace Knex {
   }
 
   interface Select extends ColumnNameQueryBuilder {
-    (aliases: { [alias: string]: string }): QueryBuilder;
+    (aliases: { [alias: string]: ColumnName }): QueryBuilder;
   }
 
   interface Table {


kibertoad:
@lorefnon What do you think?

lorefnon:
Current typings do allow for a Record of aliases: 

```
  // $ExpectType { username: string; foo: any; }[]
  await knex<User>('users')
      .select({
        username: 'name'  // Type will be inferred as string through User["name"]
      })
      .select({
        foo: knex.raw('select bar from boz')
      });
```

```
  // $ExpectType (Pick<User, "id" | "age"> & { username: string; foo: any; })[]
  await knex<User>('users')
      .select('age', 'id', {
        username: 'name'  // Type will be inferred as string through User["name"]
      })
      .select({
        foo: knex.raw('select bar from boz')
      });
```

kibertoad:
Closing as it seems not be necessary anymore.

PR #2982, opened 2019-01-05
Fix event listener duplication

### LINKED ISSUE #2967: MaxListenersExceededWarning starting from Knex 0.16.x
# Environment

Knex version: 0.16.1-3
Database + version: Postgres 9.6
OS: MacOS
Node: 10.14.2

# Bug

1. After upgrading my project from 0.15.2 to 0.16, running tests started throwing showing several warnings.

2. Error message
```
(node:62291) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 start listeners added. Use emitter.setMaxListeners() to increase limit
(node:62291) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 query listeners added. Use emitter.setMaxListeners() to increase limit
(node:62291) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 query-error listeners added. Use emitter.setMaxListeners() to increase limit
(node:62291) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 query-response listeners added. Use emitter.setMaxListeners() to increase limit
```

The project is open source and available in https://github.com/edevil/node-webapp-sample-js. If you run `npm test` you should obtain the warning. However, I reduced the test code to:

```
const Knex = require("knex");

const knexconfig = {
  client: "postgresql",
  connection: {
    database: "sample_db_citest",
    host: "localhost",
    user: "postgres",
  },
  migrations: {
    directory: "./src/migrations",
    tableName: "knex_migrations",
  },
};

async function doRollback(migrate) {
  await migrate.forceFreeMigrationsLock();
  while ((await migrate.currentVersion()) !== "none") {
    await migrate.rollback();
  }
}

async function doIt() {
  const pool = Knex(knexconfig);
  await doRollback(pool.migrate);
  await pool.migrate.latest();
  await doRollback(pool.migrate);
  await pool.migrate.latest();
  await doRollback(pool.migrate);
  await pool.migrate.latest();
  await doRollback(pool.migrate);
  await pool.migrate.latest();
  await doRollback(pool.migrate);
  await pool.migrate.latest();
  await pool.destroy();
}

doIt().then(console.log);
```
I think the problem lies in the function `doRollback()`, but I don't think I'm doing anything wrong in my code.

kibertoad:
fixes #2967 
Note that something works very differently in Node 6 and this causes removal of duplicate event listeners to result in also mutating list of event listeners in the original Knex instance, which is a much more serious concern than the problem this PR aims to fix.
This PR doesn't make functionality on Node 6 any more broken than it already was, and for newer Node version it seems to be fixed.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -31,6 +31,8 @@ describe('Query Building Tests', function() {
   require('./unit/schema/oracledb');
   require('./unit/migrate/migration-list-resolver');
   require('./unit/seed/seeder');
+  require('./unit/interface');
+  require('./unit/knex');
 });
 
 describe('Integration Tests', function() {
--- a/test/index.js
+++ b/test/index.js
@@ -31,7 +31,7 @@ describe('Query Building Tests', function() {
   require('./unit/schema/oracledb');
   require('./unit/migrate/migration-list-resolver');
   require('./unit/seed/seeder');
-  require('./unit/interface');
+  // require('./unit/interface'); ToDo Uncomment after fixed
   require('./unit/knex');
 });
 
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -5,6 +5,7 @@
 const Knex = require('../../../knex');
 const _ = require('lodash');
 const Promise = require('bluebird');
+const { isNode6 } = require('../../../lib/util/version-helper');
 
 module.exports = function(knex) {
   describe('Additional', function() {
@@ -45,6 +46,9 @@ module.exports = function(knex) {
       });
 
       it('should pass query context for raw responses', () => {
+        if (isNode6()) {
+          return;
+        }
         return knex
           .raw('select * from ??', ['accounts'])
           .queryContext('the context')
@@ -104,6 +108,9 @@ module.exports = function(knex) {
       });
 
       it('should work using camelCased table name', () => {
+        if (isNode6()) {
+          return;
+        }
         return knex('testTableTwo')
           .columnInfo()
           .then((res) => {
@@ -118,6 +125,9 @@ module.exports = function(knex) {
       });
 
       it('should work using snake_cased table name', () => {
+        if (isNode6()) {
+          return;
+        }
         return knex('test_table_two')
           .columnInfo()
           .then((res) => {
@@ -958,6 +968,9 @@ module.exports = function(knex) {
     });
 
     it('Event: query-response', function() {
+      if (isNode6()) {
+        return;
+      }
       let queryCount = 0;
 
       const onQueryResponse = function(response, obj, builder) {
@@ -987,6 +1000,9 @@ module.exports = function(knex) {
     });
 
     it('Event: preserves listeners on a copy with user params', function() {
+      if (isNode6()) {
+        return;
+      }
       let queryCount = 0;
 
       const onQueryResponse = function(response, obj, builder) {
@@ -1020,6 +1036,9 @@ module.exports = function(knex) {
     });
 
     it('Event: query-error', function() {
+      if (isNode6()) {
+        return;
+      }
       let queryCountKnex = 0;
       let queryCountBuilder = 0;
       const onQueryErrorKnex = function(error, obj) {
@@ -1055,6 +1074,9 @@ module.exports = function(knex) {
     });
 
     it('Event: start', function() {
+      if (isNode6()) {
+        return;
+      }
       return knex('accounts')
         .insert({ last_name: 'Start event test' })
         .then(function() {
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -986,7 +986,7 @@ module.exports = function(knex) {
         });
     });
 
-    it('Event: does not duplicate listeners on a copy with user params', function() {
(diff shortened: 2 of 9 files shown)


test/index.js
```diff
@@ -31,6 +31,8 @@ describe('Query Building Tests', function() {
   require('./unit/schema/oracledb');
   require('./unit/migrate/migration-list-resolver');
   require('./unit/seed/seeder');
+  // require('./unit/interface'); ToDo Uncomment after fixed
```
kibertoad: This one looks like it got broken quite a while ago, I'll open a separate issue for it

test/integration/builder/additional.js
```diff
@@ -986,7 +986,7 @@ module.exports = function(knex) {
         });
     });
 
-    it('Event: does not duplicate listeners on a copy with user params', function() {
+    it('Event: preserves listeners on a copy with user params', function() {
```
kibertoad: Probably users are going to be confused if listeners set on original knex suddenly disappear when we copy it with user params, so this is an improvement.

test/unit/knex.js
```diff
@@ -171,17 +238,18 @@ describe('knex', () => {
...
   });
 
   describe('async stack traces', () => {
     it('should capture stack trace on query builder instantiation', () => {
-      const knex = Knex({
-        ...sqliteConfig,
-        asyncStackTraces: true,
-      });
+      const knex = Knex(
+        Object.assign({}, sqliteConfig, { asyncStackTraces: true })
```
elhigu: Was there a problem with original implementation?

test/unit/knex.js
```diff
@@ -98,10 +99,76 @@ describe('knex', () => {
...
+      client: 'sqlite',
+    });
+
+    expect(knex.client.listeners('start').length).to.equal(1);
+    expect(knex.client.listeners('query').length).to.equal(1);
+    expect(knex.client.listeners('query-error').length).to.equal(1);
+    expect(knex.client.listeners('query-response').length).to.equal(1);
+
+    const knexWithParams = knex.withUserParams();
+    knexWithParams.client.on('query', (obj) => {
```
elhigu: I'm starting to have second thoughts if this was a good idea to allow creating modifiable clones of knex instance which are sharing the same pool (feature came kind of accidentally). I think it is a really nice feature to have, but API could have separated cloning and setting same pools for them and setting user parameters of each instance more clearly... anyways test looks good 👍 Wanted to just mention how this small feature of user parameters has exploded a bit for allowing to set custom event handlers too, which is good.

test/unit/knex.js
```diff
@@ -98,10 +99,76 @@ describe('knex', () => {
...
+    knex.on('query-response', onQueryResponse);
+
+    const knexWithParams = knex.withUserParams();
+
+    expect(knex.listeners('query-response').length).to.equal(1);
+    expect(knexWithParams.listeners('query-response').length).to.equal(1);
+  });
+
+  it('adding listener to copy does not affect base knex', () => {
+    if (isNode6()) {
```
elhigu: I'm worried about this !node6 stuff. I would like to know the root cause why this could not work with it. Is babel doing something strange... maybe that polyfill thing is causing it?

test/unit/knex.js
```diff
@@ -98,10 +99,76 @@ describe('knex', () => {
...
 
     expect(knexWithParams.migrate.knex.userParams).to.deep.equal({
+      isProcessingDisabled: true,
+      postProcessResponse: undefined,
       userParam: '451',
+      wrapIdentifier: undefined,
     });
   });
 
+  it('copying does not result in duplicate listeners', () => {
```
elhigu: kind of duplicate with test "adding listener to copy does not affect base knex"

test/unit/knex.js
kibertoad: Yes, Node 6 doesn't support spread, and we don't transpile tests. It used to work because it was never run in CI :)

kibertoad:
@elhigu Worst case scenario, if we conclude that there is nothing that we can do about Node 6 behaviour in this case - we could start a 0.17.x branch early (April is coming up rapidly anyway) and drop support for Node 6 there already.

elhigu:
> @elhigu So what is currently remaining? I'll review the redundant test case; then you still want to test this thing further? We probably can't remove polyfills at this point.

I was hoping to understand the reason, why that failing on node 6 happens to have some picture about how seriously it is broken and possibly to work around that breaking instead of ignoring the tests.

But I'm not completely against of merging this as is either, with some warnings in changelog about what might go wrong with node 6.

kibertoad:
@elhigu Let me do a quick pass over this pr then

test/unit/knex.js
kibertoad: They are different, actually. One tests for what happens after you copy, another one what happens after you mutate the copy.

kibertoad:
@elhigu Pushed the update to documentation. I don't think that investing too much time and energy into investigation makes much sense, April is coming soon anyway.
And this PR doesn't make Node 6 support any more broken than 0.16.0 already did (for all Node versions).

Issue #2990, opened 2019-01-09
Knexwith oracledb gives error while executing procedure having out parameters


soumyamishra12:
# Environment

Knex version:0.14.6
Database + version:
OS:windows

I get what it seems to be a Bluebird error when I try to execute an Oracle stored procedure using raw() :

```
knex.raw("BEGIN schema.STORED.PROCEDURE(?, ?); END;",
    [
        { "dir": oracledb.BIND_IN, "type": oracledb.NUMBER, "val": 5},
        { "dir": oracledb.BIND_OUT, "type": oracledb.NUMBER }
    ]).then((result) => {
        console.log("Success : " + JSON.stringify(result));
    }).catch((err) => {
        console.log("Error : " + err);
    });

```

This gives error as below:

`Error : TypeError: BEGIN schema.STORED.PROCEDURE(:1, :2); END; - expecting an array or an iterable object but got [object Null]`

soumyamishra12:
> You have defined 3 `?´ parameter bindings and provide only 2 parameters. So when knex is looking for 3rd binding, it gets null I suppose.

@elhigu  That a mistake when i copied my code to modify i am passing 2 parameters only and one is out and one is in. i have scanned through the issue  #1901 but didn't find proper answer to solve it .

If any fix will be available with latest versions it will be helpful.

> if (obj.method == 'raw') {
>     return resolver({
>       response: outBinds
>     });
>   }

After adding the above code this works though but this is change in index.js in dialects which is not applicable when installed in different system again.

Is there any other recommendation.
Please help as my project need is using heavily oracle db procs.

elhigu:
Currently only recommendation in #1901 is to use oracledb driver directly instead of through knex. Better solution would be to send PR to knex which would add support + tests for that.

soumyamishra12:
Thanks  @elhigu will create a PR

Issue #1901, opened 2017-02-07
ORACLE: Stored procedure issue using "raw()"


electrotype:
Knex version : 0.12.6

I get what it seems to be a Bluebird error when I try to execute an Oracle stored procedure using `raw()` :
```
knex.raw("BEGIN MY.STORED.PROCEDURE(?, ?, ?); END;",
    [
        { "dir": oracledb.BIND_IN, "type": oracledb.NUMBER, "val": 13036 },
        { "dir": oracledb.BIND_OUT, "type": oracledb.NUMBER },
        { "dir": oracledb.BIND_OUT, "type": oracledb.NUMBER }
    ]).then((result) => {
        console.log("Success : " + JSON.stringify(result));
    }).catch((err) => {
        console.log("Error : " + err);
    });
```

This results in the error :

> Error : TypeError: BEGIN MY.STORED.PROCEDURE(:1, :2, :3); END; - expecting an array or an iterable object but got [object Null]

If I debug, I see that the request works, Oracle response is correct. Here's a screenshot of a breakpoint in file `node_modules/knex/lib/dialects/oracledb/index.js`, line `162` :

![](http://i.imgur.com/qEHpAcT.png)

The `err` is null, there are no `rows` or `resultSet` in the `results` object but a [outBinds](https://github.com/oracle/node-oracledb/blob/master/doc/api.md#execoutbinds) array, containing the OUT parameters of the stored procedure call.

The error actually occures when line `205` is reached :

`resultResolve(results);`

elhigu:
i have no idea why your query works at all, I suppose oracle driver allows passing objects to value bindings... just out of curiosity what does:

```js
knex.raw("BEGIN MY.STORED.PROCEDURE(?, ?, ?); END;",
    [
        { "dir": oracledb.BIND_IN, "type": oracledb.NUMBER, "val": 13036 },
        { "dir": oracledb.BIND_OUT, "type": oracledb.NUMBER },
        { "dir": oracledb.BIND_OUT, "type": oracledb.NUMBER }
    ]).toSQL() 
```

print out?

electrotype:
`.toQuery()` returns :

`BEGIN MY.STORED.PROCEDURE({"dir":3001,"type":2002,"val":13036}, {"dir":3003,"type":2002}, {"dir":3003,"type":2002}); END;`

elhigu:
[...]

PR #2991, opened 2019-01-09
comitted for issue #1901 #2990

soumyamishra12:
This changes are tried and tested for fix of #1901 #2990 and if anything better is there please suggest.


kibertoad:
@soumyamishra12 Would you consider finishing the PR?

Issue #2999, opened 2019-01-15
@babel/polyfill & @types/bluebird should be declared as dev dependencies


mmouterde:
# Environment

Knex version:0.16.3
Database + version: pg 10
OS: Linux

# Bug

Some dependencies are declared in the `dependencies` scope instead of the `dev-dependencies` scope

Please see : https://github.com/tgriesser/knex/issues/2948#issuecomment-454435008

kibertoad:
Imagine that you are using knex as a dependency. If bluebird typings are not `dependencies`, then you don't get them transitively and you don't have typings even though you are using bluebird interfaces in your code (as knex returns bluebird promises).

Polyfill gets called during knex startup, how can you avoid bundling it?

mmouterde:
Ok, sorry, forget about it... bluebird use typescript 2.8, I use 2.6 and there is a breaking change in syntax :/

kibertoad:
@mmouterde 2.6 is more than year old, are you sure you don't want to update?

mmouterde:
yep, updated. Thah works. sorry.

PR #3000, opened 2019-01-15
Update package.json

mmouterde:
fix https://github.com/tgriesser/knex/issues/2999
move some dependencies to dev-dependencies


kibertoad:
See https://github.com/tgriesser/knex/blob/master/knex.js -> this will break

PR #3004, opened 2019-01-18
Always use well documented pg client query() config argument

Druotic:
For queries that included bindings, I found that arguments were sent in this form:

```
pg.query({ text: 'SELECT ....' }, ['some', 'bindings'], callback);
```

Technically, this works when calling out to the Node.js Postgres client directly. However, it is not documented publicly (see [here](https://node-postgres.com/api/client#-code-client-query-config-queryconfig-callback-err-error-result-result-gt-void-gt-void-code-)). This change just adopts an approach that is public documented and therefore less prone to breaking in the future. After this change, the above becomes:

```
pg.query({ text: 'SELECT ....', values: ['some', 'bindings']}, callback);
```

The motivation behind this change was for other libraries that try to instrument the postgres client in between. Using undocumented inputs can lead to issues if the library didn't consider all possible undocumented input patterns. My particular usecase was an AWS X-Ray instrumentation package, [`aws-sdk-xray-postgres`](https://github.com/aws/aws-xray-sdk-node/tree/master/packages/postgres)

note: I was only able to get a subset of tests to pass on my local machine. I added a unit test for explicit arg checking, but am relying on existing integration tests to ensure no regressions in basic querying functionality. I'm relying on the CI job to run the remainder of tests. If something fails, I may need pointers on how to get the full suite running locally - I wasn't able to get everything running using the existing docs... but perhaps I missed some steps.

### DIFF
--- a/src/dialects/postgres/index.js
+++ b/src/dialects/postgres/index.js
@@ -222,10 +222,17 @@ assign(Client_PG.prototype, {
   // Runs the query on the specified connection, providing the bindings
   // and any other necessary prep work.
   _query(connection, obj) {
-    let sql = obj.sql;
-    if (obj.options) sql = extend({ text: sql }, obj.options);
+    let queryConfig = {
+      text: obj.sql,
+      values: obj.bindings || [],
+    };
+
+    if (obj.opts) {
+      queryConfig = extend(queryConfig, obj.options);
+    }
+
     return new Promise(function(resolver, rejecter) {
-      connection.query(sql, obj.bindings, function(err, response) {
+      connection.query(queryConfig, function(err, response) {
         if (err) return rejecter(err);
         obj.response = response;
         resolver(obj);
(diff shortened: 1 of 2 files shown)


elhigu:
```
1) postgresql | pg
       Joins
         supports joins with overlapping column names:
      AssertionError: expected [ Array(2) ] to deeply equal [ Array(2) ]
      + expected - actual
       [
         {
      -    "email": "test2@example.com"
      +    "0": "test@example.com"
      +    "1": "test2@example.com"
         }
         {
      -    "email": "test3@example.com"
      +    "0": "test@example.com"
      32m+    "1": "test3@example.com"
         }
       ]
 ```

elhigu:
Looks like there is some underlying subtle differences 🤔

Druotic:
Aw, darn. I'll dig into this some more after lunch, I must have not considered some edge case.

Druotic:
Fixed! Silly typo. The main path we utilize doesn't hit the options path. 👍 for integration tests!

PR #3006, opened 2019-01-21
Add "constraintName" arg to TS definition

fmenegossi:
Add `constraintName` optional argument to `TableBuilder.primary()` TypeScript definition.
Currently, if you give the second argument, TS will tell that `primary()` expects only 1 argument, when in fact it supports the second optional argument to define the constraint name into the database.


kibertoad:
Thanks!

PR #3016, opened 2019-01-28
schemaName is missing in MigratorConfig typings

mmouterde:
add a schemaName value to MigratorConfig.

### DIFF
--- a/types/knex.d.ts
+++ b/types/knex.d.ts
@@ -825,11 +825,13 @@ declare namespace Knex {
 
   interface MigratorConfig {
     database?: string;
-    directory?: string;
+    directory?: string | string[];
     extension?: string;
     tableName?: string;
-    disableTransactions?: boolean;
     schemaName?: string;
+    disableTransactions?: boolean;
+    sortDirsSeparately?: boolean;
+    
   }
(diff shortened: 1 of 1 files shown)


types/knex.d.ts
```diff
@@ -829,6 +829,7 @@ declare namespace Knex {
     extension?: string;
     tableName?: string;
     disableTransactions?: boolean;
+    schemaName?: string;
```
kibertoad: since you are changing it anyway, could you please also add

`sortDirsSeparately : optional boolean`
and update `directory` one to support both String and String[]?

mmouterde: Absolutely !

There is other diff with Doc :
- `loadExtensions`, `migrationSource` are in doc, not in code
- `database` is in code, not in doc.

kibertoad:
Thanks a lot! Would be nice to address these discrepancies too.

kibertoad:
Unfortunately, I don't have permissions to publish knex artefacts. @elhigu would be a better person to ask.

elhigu:
I just published 0.16.4-next1 pre-release for testing

PR #3039, opened 2019-02-07
mysql warn `.returning()` does not have any effect.

HurSungYun:
In SQLite3, users can be warned that `.returning()` has no effect.

https://github.com/tgriesser/knex/blob/master/src/dialects/sqlite3/query/compiler.js#L18

For consistency, how about MySQL users to be warned as well?

### DIFF
--- a/src/dialects/mysql/query/compiler.js
+++ b/src/dialects/mysql/query/compiler.js
@@ -7,6 +7,14 @@ import { assign, identity } from 'lodash';
 
 function QueryCompiler_MySQL(client, builder) {
   QueryCompiler.call(this, client, builder);
+
+  const { returning } = this.single;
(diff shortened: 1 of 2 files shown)


elhigu:
Tests are missing and I'm not sure if this should be an exception and thrown earlier than compilation phase.

elhigu:
Also sqlite warning could be moved to happen earlier. Nowadays it is possible to test that error messages are emitted, so test should be added in either way.

HurSungYun:
I think the warning would be enough. In the case of using the same builder code for both postgreSQL and MySQL, it would be a problem.

And added some tests for it.

elhigu:
I think this is fine. Thanks!

PR #3040, opened 2019-02-07
[#3033] fix: sqlite3 drop/renameColumn() breaks with postProcessResponse

### LINKED ISSUE #3033: (0.16.3) Using knexSnakeCaseMappers() with sqlite3 dropColumn() breaks tables
# Environment

Knex version: 0.16.3
Database + version: sqlite3 4.0.6
OS: Windows 7 x64

# Bug

When using Objection's `knexSnakeCaseMappers()`, if `dropColumn()` is used, the original table is not deleted/modified, the temporary `table _knex_temp_alter3` is not deleted, and subsequent migrations that manipulate the table error out.

Dropping columns should not break the database schema when using `knexSnakeCaseMappers`

2. Error message

Varies. If the table is not manipulated after the `dropColumn`/`dropColumns` call, no error is produced, but a bogus/incorrect database structure is left behind.

If the table is manipulated in subsequent migrations (or `knex.schema` promises), various errors are thrown, mostly of type `Unhandled rejection Error: SQLITE_ERROR: table _knex_temp_alter3 has no column named column_five`

Files required to reproduce (using `sqlite3`):

**`knexfile.js`**
```javascript
const { knexSnakeCaseMappers } = require('objection');

module.exports = {
  development: {
    client: 'sqlite3',
    connection: {
      filename: './data/development.sqlite3',
    },
    useNullAsDefault: true,
    ...knexSnakeCaseMappers(),
  },
};
```

**`migrations/1_one.js`**
```javascript
  return knex.schema.createTable('table_one', table => {
    table.string('column_one');
    table.float('column_two');
  });
};

exports.down = function(knex, Promise) {};
```

**`migrations/2_two.js`**
```javascript
  return knex.schema.alterTable('table_one', table => {
    table.dropColumn('column_two');
  });
};

exports.down = function(knex, Promise) {};
```

**`migrations/3_three.js`**
```javascriptexports.up = function(knex, Promise) {
  return knex.schema
    .createTable('table_two', table => {
      table.integer('column_one');
    })
    .table('table_one', table => {
      table.integer('column_three');
    })
};

exports.down = function(knex, Promise) {};
```

/wh

elunic:
This fixes #3033, as described there.


elhigu:
Test is missing and someone should check if that processResponse has been designed to be async method.

elunic:
I added test cases for all the added functionality. That required adding a util, so tell me what you think of it all. Looking forward to it.

src/util/snake-case-mappers.js
```diff
@@ -0,0 +1,34 @@
+import { isArray, isObject, snakeCase, camelCase, mapKeys } from 'lodash';
```
elhigu: I don't think this should be included here. You can use some dummy wrap / postprocess functions in tests which converts identifiers lower / uppercase to demonstrate the problem.

kibertoad: can those be simply moved to tests then?

elhigu: There is no reason, why they should be in at all. Just using dummier implementations that does uppercase / lower case operations for identifiers and results should do the same thing and are simpler.

kibertoad:
@elhigu I addressed your review comments and also fixed table name processing which wasn't working correctly either; please give it another look!

src/dialects/sqlite3/schema/ddl.js
```diff
@@ -38,27 +46,31 @@ assign(SQLite3_DDL.prototype, {
...
     return currentCol;
   }),
 
   getTableSql() {
-    return this.trx.raw(
-      `SELECT name, sql FROM sqlite_master WHERE type="table" AND name="${
-        this.tableName
-      }"`
-    );
+    this.trx.disableProcessing();
```
kibertoad: this is an internal query which breaks if we e. g. uppercase fields name and sql.

elhigu:
Thanks! Looks good I suppose :)

Issue #2998, opened 2019-01-14
knex 0.16 doesn't support knexfiles written in TypeScript


brunolm:
# Environment

Knex version: `0.16.3`
Database + version: docker `postgres:10.4`
OS: Windows 10 Pro & Docker `node:8.14.0`

# Bug

1. `knex migrate:make --knexfile knexfile.ts`
2. Error message: `Unexpected token import`

Works normally on `0.15.x`

mikl:
0.16 now has TypeScript types included in the repo (see #2845 ). You probably need to remove the `@types/knex` from your project for this to work correctly.

brunolm:
I uninstalled `@types/knex` but I still got the error.

```ts
/usr/src/app/src/db/knexfile.ts:1
(function (exports, require, module, __filename, __dirname) { import { database } from '../config'
                                                              ^^^^^^

SyntaxError: Unexpected token import
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:617:28)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at initKnex (/usr/src/app/node_modules/knex/bin/cli.js:62:25)
    at Command.commander.command.description.option.action (/usr/src/app/node_modules/knex/bin/cli.js:172:24)
    at Command.listener (/usr/src/app/node_modules/commander/index.js:315:8)
    at emitTwo (events.js:126:13)
    at Command.emit (events.js:214:7)
    at Command.parseArgs (/usr/src/app/node_modules/commander/index.js:654:12)
    at Command.parse (/usr/src/app/node_modules/commander/index.js:474:21)
```

elhigu:
`knex migrate:make --knexfile knexfile.ts` isn't that trying to run typescript code with plain node? I have no idea how that could have worked with 0.15 either.

How does your knexfile look like? is there any import statements?

mikl:
I have no idea why that would work with 0.15. As the error `Unexpected token import` tells you, the `import * as knex from 'knex'` is not understood by Node.js (not in the configuration you're using, at least).

You'll either need to compile your knexfile to JavaScript, use a Node.js version that understands the import syntax, or revert the file to use the old `require()` syntax.

brunolm:
[...]

PR #3041, opened 2019-02-08
Fix for #2998 - Migrator & TypeScript

pvoisin:
As described in issue #2998, it seems that what we've found as a symptom (TypeScript migrations not working) is caused by the underlying `#listAllAndCompleted` function which doesn't properly relay the `config.loadExtensions` to `#listAll`.
With the little change in that PR things seem back to normal and the tests all pass as before.

Note: as this problem wasn't caught by the automated tests, eventually, the test suites should be augmented.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "pg-query-stream": "^1.1.2",
     "prettier": "^1.16.2",
     "rimraf": "^2.6.3",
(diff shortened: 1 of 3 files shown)


kibertoad:
OK, so the proper fix for this would be to fully remove dependency on knexfile from migration:make (since it not really needs it).

PR #3043, opened 2019-02-10
Update enum arguments to reflect latest signature

willsoto:
Reimplemented from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29946

Looks like this was lost when the typings were transferred to the repo.

### DIFF
--- a/types/knex.d.ts
+++ b/types/knex.d.ts
@@ -851,6 +851,7 @@ declare namespace Knex {
 
   interface EnumOptions {
     useNative: boolean;
+    existingType: boolean;
     enumName: string;
 }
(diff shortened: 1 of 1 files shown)


types/knex.d.ts
```diff
@@ -850,6 +849,11 @@ declare namespace Knex {
     now(): Raw;
   }
 
+  interface EnumOptions {
```
kibertoad: Can you also add boolean `existingType` that PostgreSQL supports?

kibertoad: nvm, added myself :)

PR #3044, opened 2019-02-11
Change default value of reapIntervalMillis pool option

cham11ng:
***Change default value of reapIntervalMillis pool option***

I'm getting this message in my log.
```
Pool config option "evictionRunIntervalMillis" is no longer supported. 
See https://github.com/Vincit/tarn.js for possible pool config options.
```
Newer MSSQL library is no longer using [node-pool](https://github.com/coopernurse/node-pool) library for pooling instead it's using [tarn](https://github.com/vincit/tarn.js). So, `evictionRunIntervalMillis` is not available in `tarn.js` instead it has `reapIntervalMillis`
```js
  // how often to check for idle resources to destroy
  reapIntervalMillis: 1000,
```

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -25,7 +25,7 @@ const POOL_CONFIG_OPTIONS = Object.freeze([
   'fifo',
   'priorityRange',
   'autostart',
-  'reapIntervalMillis',
+  'evictionRunIntervalMillis',
   'numTestsPerRun',
   'softIdleTimeoutMillis',
   'Promise',
--- a/src/constants.js
+++ b/src/constants.js
@@ -25,7 +25,7 @@ const POOL_CONFIG_OPTIONS = Object.freeze([
   'fifo',
   'priorityRange',
   'autostart',
-  'evictionRunIntervalMillis',
+  'reapIntervalMillis',
   'numTestsPerRun',
   'softIdleTimeoutMillis',
   'Promise',
--- a/src/dialects/mssql/index.js
+++ b/src/dialects/mssql/index.js
@@ -32,7 +32,7 @@ function Client_MSSQL(config = {}) {
     min: 1,
     max: 1,
     idleTimeoutMillis: Number.MAX_SAFE_INTEGER,
-    evictionRunIntervalMillis: 0,
+    reapIntervalMillis: 0,
   };
 
   Client.call(this, config);
--- a/src/dialects/mssql/index.js
+++ b/src/dialects/mssql/index.js
@@ -32,7 +32,7 @@ function Client_MSSQL(config = {}) {
     min: 1,
     max: 1,
     idleTimeoutMillis: Number.MAX_SAFE_INTEGER,
-    reapIntervalMillis: 0,
+    reapIntervalMillis: 1000,
   };
 
   Client.call(this, config);
--- a/types/knex.d.ts
+++ b/types/knex.d.ts
@@ -817,7 +817,7 @@ declare namespace Knex {
     acquireTimeoutMillis?: number;
     fifo?: boolean;
     autostart?: boolean;
-    reapIntervalMillis?: number;
+    evictionRunIntervalMillis?: number;
     numTestsPerRun?: number;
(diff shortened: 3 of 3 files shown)


src/dialects/mssql/index.js
```diff
@@ -32,7 +32,7 @@ function Client_MSSQL(config = {}) {
     min: 1,
     max: 1,
     idleTimeoutMillis: Number.MAX_SAFE_INTEGER,
-    evictionRunIntervalMillis: 0,
```
elhigu: This is wrong. Here we are setting parameters of internal pool of mssql driver so reapInterval is correct.

src/constants.js
```diff
@@ -25,7 +25,7 @@ const POOL_CONFIG_OPTIONS = Object.freeze([
   'fifo',
   'priorityRange',
   'autostart',
```
elhigu: Would be good to fix completely that configure option set instead of just fixing that one attribute nema.

elhigu: Also looks like these are only used in one place to emit some warnings... so I think it will be just fine to remove those warnings and let the tarn.js do the complaining if invalid attributes are passed to it.

cham11ng: These options are already in `tarn.js`. Only changed one is `reapIntervalMillis`

```
min,
max,
acquireTimeoutMillis,
idleTimeoutMillis
```

cham11ng: I've just removed `evictionRunIntervalMillis` from here. Since, this is only the source of warning. I think Its bad idea to remove warning.

```
this.mssqlPoolSettings = {
    min: 1,
    max: 1,
    idleTimeoutMillis: Number.MAX_SAFE_INTEGER,
    reapIntervalMillis: 0,
  };
```

src/dialects/mssql/index.js
elhigu: This is still invalid change.

Those parameters are passed to generic-pool (https://github.com/coopernurse/node-pool) that is used inside mssql driver.  That pool has configuration attribute called `evictionRunIntervalMillis`.

https://github.com/tgriesser/knex/blob/b56ff04e610140d9554ec609986831cd7936960a/src/dialects/mssql/index.js#L217

src/constants.js
elhigu: Currently that array is listing bunch of options that actually doesn't exist in tarn.js and allows to pass them and it fails to list other parateters that are supported by tarn. This kind warning of invalid parameters will be easier to maintain in tarn instead and it really shouldn't be any concern of knex.

It has been implemented becuase back in days knex swapped the pool implementation that it was using every other week.

src/dialects/mssql/index.js
kibertoad: After mssql v6 is released that is no longer the case, though - they moved to tarn.js as well

elhigu: Are they also providing a way to get connection without having to go through their pool? If so I don't think I have anything against mssql anymore (well except it always lags a bit behind of tedious, but we lag more 😅 ). IIRC one impressive feature that tedious had was possibility to sent multiple queries though connection without waiting first that result from previous query is returned. That allows to send many small queries to server in transaction without need to always wait for server roundtrip between sending queries. Maybe mssql supports it as well.

elhigu: Anyways... looks like we have to require that everyone uses mssql > 6.0.0?

src/dialects/mssql/index.js
```diff
@@ -32,7 +32,7 @@ function Client_MSSQL(config = {}) {
...
+    reapIntervalMillis: 1000,
```
elhigu: Actually right now this can be removed completely since default value is also 1000 and idleTimeTimeout is so big that nothing will be released ever because of idling anyways.

elhigu: that way pool parameters should be compatible with generic-pool and tarnjs

cham11ng:
Closing this. It was my project implementation issue.

PR #3046, opened 2019-02-12
Add onError config for pool

Minishlink:
To handle idle pool errors
see https://github.com/brianc/node-postgres/issues/1324


elhigu:
Knex is not using pool from pg library and missing tests etc.

Issue #2732, opened 2018-07-26
Async credentials while running migrations


kurthopfer:
# Feature discussion / request
I would like to be able to run migrations as part of AWS codepipline through the buildspec process.
Basically a buildspec file issues some commands (node, python, ruby, etc...) after a build has completed.
I want to be able to run `knex migrate:latest --env dev` which works fine if I have my database connection object hard-coded, either in the knexfile.js or a config file. But since I am storing all of the connection information in AWS Secrets manager I have to make an async call to secrets manger before I can populate the knexfile.

I am not sure how in-demand this feature would be but I believe it should be somewhat simple to implement.

elhigu:
I think this would be good feature to have. I've also stumbled on this and have helped few times someone else to work around this.

elhigu:
Implementation could be as easy as just allow to return promise from knexfile.js instead of synchronous configuration object.

mkiddTempus:
I'm having this exact issue as well. Does anyone have suggestions for a work around?

elhigu:
@mkiddTempus dont use knex commandline client for running migrations and that way you can decide to first fetch all secrets before trying to create DB connection

mkiddTempus:
How do you not use the CLI for running migrations? 

I did just create a second cli-specific knexfile with hard coded values and passed it in via --knexfile and  while that worked locally I don't think my devops team will go for it.

elhigu:
By calling migration API in your app code https://knexjs.org/#Migrations-latest

mkiddTempus:
In the mean time since I posted I did some searching and also found this solution: https://york.io/2016/05/04/knex-reset-migrations.html

This still would be a nice feature to have.

jyrkiput:
I had the almost similar issue with AWS Lambdas, where I wanted to get secrets from Secret Manager. My solution was to write custom client, where most the code comes from Postgres client and AWS examples

```
var lodash = require('lodash')
var Client_PG = require('knex/lib/dialects/postgres/')
var inherits = require('inherits');
var Promise = require('bluebird').Promise;

function SecretsManagerClient(config) {
    Client_PG.apply(this, arguments)
}

inherits(SecretsManagerClient, Client_PG)

function createConnectionSettings(connectionSettings) {
    return new Promise(function(resolve, reject) {
        var AWS = require('aws-sdk'),
            region = connectionSettings.region,
            secretName = connectionSettings.secretName,
            secret;

        var client = new AWS.SecretsManager({
[...]

PR #3052, opened 2019-02-14
Proposal for supporting dynamic connection settings

aivopaas:
Knex has automatic pool management and reconnection, but it doesn't provide an easy way to provide dynamic connection settings. That becomes an issue when the DB server IP changes (would have to restart service or at least handle it manually and create new Knex instance), or when we want the pool to connect to different physical DB instances.

My proposed solution is to add an optional configuration option named `getConnection` which accepts a function that returns either the connection settings object or a promise that resolves to the connection settings.

That function (if provided) is called every time when getting a new raw connection from the dialect/driver.

Sample usage:
```js
  client: 'mysql',
  connection: {
    database: 'db_name'
  },
  getConnection: async (settings) => {
    // Get instance of the DB asynchronously (from Consul for example)
    const {host, port, user, password} = await getDbInstance();
    // `settings` is the `connection` options property provided to Knex
    return { ...settings, host, port, user, password };
  }
});
```

That would probably also solve the problem described in #2732

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Boolean param for rollback() to rollback all migrations #2968
 - seed:run print the file name of the failing seed #2972 #2973
 - verbose option to CLI commands #2887
+- `getConnection` configuration option (function) to provide connection settings dynamically #2732
 
 ### Bug fixes:
 
--- a/src/client.js
+++ b/src/client.js
@@ -55,7 +55,7 @@ function Client(config = {}) {
   }
 
   this.connectionSettings = cloneDeep(config.connection || {});
-  if (this.driverName && config.connection) {
+  if (this.driverName && (config.connection || config.getConnection)) {
     this.initializeDriver();
     if (!config.pool || (config.pool && config.pool.max !== 0)) {
       this.initializePool(config);
@@ -225,6 +225,15 @@ assign(Client.prototype, {
     return { min: 2, max: 10, propagateCreateError: true };
   },
 
+  getConnectionSettings() {
+    if (typeof this.config.getConnection === 'function') {
+      return Promise.resolve(
+        this.config.getConnection(this.connectionSettings)
+      );
+    }
+    return Promise.resolve(this.connectionSettings);
+  },
+
   getPoolSettings(poolConfig) {
     poolConfig = defaults({}, poolConfig, this.poolDefaults());
 
@@ -250,13 +259,15 @@ assign(Client.prototype, {
 
     return Object.assign(poolConfig, {
       create: () => {
-        return this.acquireRawConnection().tap((connection) => {
-          connection.__knexUid = uniqueId('__knexUid');
+        return this.getConnectionSettings()
+          .then((settings) => this.acquireRawConnection(settings))
+          .tap((connection) => {
+            connection.__knexUid = uniqueId('__knexUid');
 
-          if (poolConfig.afterCreate) {
-            return Promise.promisify(poolConfig.afterCreate)(connection);
-          }
-        });
+            if (poolConfig.afterCreate) {
+              return Promise.promisify(poolConfig.afterCreate)(connection);
+            }
+          });
       },
 
       destroy: (connection) => {
--- a/src/dialects/mssql/index.js
+++ b/src/dialects/mssql/index.js
@@ -207,7 +207,7 @@ assign(Client_MSSQL.prototype, {
   // connection needs to be added to the pool.
   acquireRawConnection(settings) {
     return new Promise((resolver, rejecter) => {
-      const connectionSettings = Object.assign({}, connectionSettings);
+      const connectionSettings = Object.assign({}, settings);
       connectionSettings.pool = this.mssqlPoolSettings;
       // #1235 mssql module wants 'server', not 'host'. This is to enforce the same
       // options object across all dialects.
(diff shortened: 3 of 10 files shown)


aivopaas:
There are properties used from config.connection (database name for example) in some places other than where connection is created. That makes it impossible to replace connection with the function without causing braking changes and/or changing the options structure (splitting properties that is not needed for only creating connection out of that object).

The dynamic settings should not be able to change any other settings than what is used to determine and connect to the instance of the DB. Database name, collations and whatever other settings there might be must stay the same. I didn't find a good way to force it though, because it might be too dialect-dependant which properties can and which can not change.

elhigu:
> There are properties used from config.connection (database name for example) in some places other than where connection is created. That makes it impossible to replace connection with the function without causing braking changes and/or changing the options structure (splitting properties that is not needed for only creating connection out of that object).

I don't see why that would cause breaking changes. We would just need to write those parts of code a bit different way, where the values are used (checking if we are dealing with function or an object or string first).

> The dynamic settings should not be able to change any other settings than what is used to determine and connect to the instance of the DB. Database name, collations and whatever other settings there might be must stay the same.

Why it shouldn't be able to change anything else? Sounds like really opinionated feature if it handles only some parameters that you need in your specific use-case.

aivopaas:
> I don't see why that would cause breaking changes. We would just need to write those parts of code a bit different way, where the values are used (checking if we are dealing with function or an object or string first).

Database field is used in `compiler.columInfo()` for example. If I'm not mistaken, using dynamic settings function there would cause async settings fetch for every query if not even multiple times per query. And if database name could be changed dynamically, that code would have to be completely changed to take into account the actual instance in the pool against which the query is executed. So it is not that simple to make everything dynamic without rethinking how query builder is working.

> Why it shouldn't be able to change anything else? Sounds like really opinionated feature if it handles only some parameters that you need in your specific use-case.

It's not about my specific use case. I have to change everything except the database name and I can't think o f a good reason for even being able to change the database name. So for the majority it would be ok to always get the settings dynamically (object vs function) but that would indeed require some major refactoring of where database name is currently used.

elhigu:
> > I don't see why that would cause breaking changes. We would just need to write those parts of code a bit different way, where the values are used (checking if we are dealing with function or an object or string first).
> 
> Database field is used in `compiler.columInfo()` for example. If I'm not mistaken, using dynamic settings function there would cause async settings fetch for every query if not even multiple times per query. And if database name could be changed dynamically, that code would have to be completely changed to take into account the actual instance in the pool against which the query is executed. So it is not that simple to make everything dynamic without rethinking how query builder is working.

Calling that dynamic getter multiple times is not a problem, one can do that in a efficient manner if one decides to use the feature of getting configuration in a dynamic way (usually just once when app is starting). 

I still don't like the proposed API at all. Would be better to open feature request with more discussion why this is needed etc. I don't think that just changing password / host etc. is enough. If you like that you can easily create wrapper to get knex instance with correct parameters (even though in that case there would be multiple connection pools, which really doesn't matter if there are multiple database servers). It also ignores the fact that connection attribute might be just a string. 

> > Why it shouldn't be able to change anything else? Sounds like really opinionated feature if it handles only some parameters that you need in your specific use-case.
> 
> It's not about my specific use case. I have to change everything except the database name and I can't think o f a good reason for even being able to change the database name. 

That second sentence really sounds like very specific use. I'm 100% sure that other people will come up to  think of many reasons why one would like to change also some other connection setting which are not included in this PR.

> So for the majority it would be ok to always get the settings dynamically (object vs function) but that would indeed require some major refactoring of where database name is currently used.

I still don't think that just resolving connection object every time before using the database name is necessary a big refactoring (I cannot say if it really is without digging in to the code of those parts, if current compiler flow is completely synchronous, it might need some pre-compilation pass to be added, which resolves configuration just once before starting). After that its only up to the one who implements the dynamic getter to make sure that it works fast enough. 

Anyways I still stand by my opinion that if this is going to be included to knex core it should allow dynamically to get the whole connection details object / string. Otherwise I would recommend to implement this kind of functionality outside of knex by re-creating knex instance with new settings.

For me it would be better to just have that dynamic setup to be resolved once when knex instance is created to help the #2732 issue to prevent knex from needing to watch those changes and to make sure that old pool is torn down and new created when setup is changed (and having to deal all errors that might happen during that). This thing might introduce some pretty nasty bugs if not done properly.

When I think about this more I'm starting to incline that maybe I wouldn't like this to be included to knex at all and leave this to be handled outside of knex. Maybe by writing a `knex-async-setup` package which would implement this externally (it would not be so convenient to use though, since invalidating old query builders when connection setup is changed doesn't sound so trivial).

aivopaas:
I think you are a bit overthinking what I'm trying to solve here.

The problem with current Knex implementation is that for the connection pool it always uses the same backend. I just need a way to make sure every time a connection is added to the pool it is with fresh configuration (an instance that I know should accept the connection, not an old host that might already be teared down.

Having said that, I solved the problem by extending the mysql client with my own patched 
`acquireRawConnection` which does exactly what I would have been able with that PR here, but with a lot less effort. SO I don't really see a point fighting the stone wall right now trying to explain the problem further if there's no strong interest in getting that fixed in Knex.

elhigu:
> I think you are a bit overthinking what I'm trying to solve here.

Not really. I just did not want that kind of really opiniated feature to knex just to solve that single case. 

Having your own workaround is better solution this time. It will be also easier for you to modify to your future needs later on.

PR #3055, opened 2019-02-14
make unionAll()'s call signature match union()

### LINKED ISSUE #3054: .unionAll does not have the same call signature as .union
# Environment

Knex version: 0.16.3
Database + version: MySQL 5.7
OS: N/A

# Bug

Knex's documentation states that `.unionAll()` has the same call signature as `.union()`.

> unionAll — `.unionAll(query)`
> Creates a union all query, with the same method signature as the union method.

In practice, `.unionAll()` only supports the callback syntax, ie:

```js
knex.select('*').from('users').whereNull('last_name').unionAll(function() {
  this.select('*').from('users').whereNull('first_name');
})
```

However, `.union()` supports two additional call signatures, which are _not_ supported by `.unionAll()`, ie:

```js
knex.select('*').from('users').whereNull('last_name').unionAll([
  knex.raw('select * from users where first_name is null'),
  knex.raw('select * from users where email is null')
])
```

Additionally, `unionAll()` can take the same `wrap` parameter that's supported by `.union()`, but this is not mentioned in the documentation.

**This discrepancy is readily apparent when comparing [the implementations](https://github.com/tgriesser/knex/blob/4db047df02653ec9e70ef004e38f47851231c62d/src/query/builder.js#L609-L643) of `.union()` `.unionAll()`**.

# Error message

tl;dr; it's nothing helpful (which seems like a problem itself, but is probably a topic for a different bug report).  knex tries to invoke every function method on the query builder, and finally throws an exception while trying to start a MySQL transaction.

---

tl;dr; `.unionAll()` should be updated to support the same method signature as `.union()` (there's no reason not to), and the documentation should be updated to show that `.unionAll()` also supports `wrap`

schmod:
Consolidates the logic used by union() and unionAll(), allowing the two to be used interchangeably (which is the current behavior claimed by the docs).

This allows `.unionAll()` to be called on multiple subqueries (rather than having to repeatedly call `.unionAll()` for each one).

Additionally, this adds test-coverage for the undocumented `wrap` argument on `.unionAll()` (which  is already documented for `.union()`).

---

### Before: 

This was the only allowable call-signature (apart from a second undocumented `wrap` parameter):
```js
qb.unionAll(subQuery);
```

If I wanted to write the `.unionAll()` equivalent of `qb.union([sq1, sq2])`, I would have needed to write something like:

```js
[s1, sq2].reduce((qb, sq) => qb.unionAll(sq), qb);
```

### After

Any valid call to `.union()` should now work with `.unionAll()`
```js
qb.unionAll(sq1);
qb.unionAll(sq1, wrap);
qb.unionAll(sq1, sq2);
qb.unionAll(sq1, sq2, wrap);
qb.unionAll([sq1, sq2]);
qb.unionAll([sq1, sq2], wrap);
```

---

Fixes #3054

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -634,12 +634,12 @@ assign(Builder.prototype, {
 
   // Add a union statement to the query.
   union(...args) {
-    return this._union('union', ...args);
+    return this._union('union', args);
   },
 
   // Adds a union all statement to the query.
   unionAll(...args) {
-    return this._union('union all', ...args);
+    return this._union('union all', args);
   },
 
   // Adds a `having` clause to the query.
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -605,7 +605,7 @@ assign(Builder.prototype, {
     return this;
   },
 
-  _union(args, clause) {
+  _union(clause, args) {
     let callbacks = args[0];
     let wrap = args[1];
     if (args.length === 1 || (args.length === 2 && isBoolean(wrap))) {
@@ -627,19 +627,19 @@ assign(Builder.prototype, {
         callbacks.push(wrap);
         wrap = false;
       }
-      this._union([callbacks, wrap], clause);
+      this._union(clause, [callbacks, wrap]);
     }
     return this;
   },
 
   // Add a union statement to the query.
-  union() {
-    return this._union(arguments, 'union');
+  union(...args) {
+    return this._union('union', ...args);
   },
 
   // Adds a union all statement to the query.
-  unionAll(callbacks, wrap) {
-    return this._union(arguments, 'union all');
+  unionAll(...args) {
+    return this._union('union all', ...args);
   },
 
   // Adds a `having` clause to the query.
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -605,41 +605,41 @@ assign(Builder.prototype, {
     return this;
   },
 
-  // Add a union statement to the query.
-  union(callbacks, wrap) {
-    if (arguments.length === 1 || (arguments.length === 2 && isBoolean(wrap))) {
+  _union(args, clause) {
+    let callbacks = args[0];
+    let wrap = args[1];
+    if (args.length === 1 || (args.length === 2 && isBoolean(wrap))) {
       if (!Array.isArray(callbacks)) {
         callbacks = [callbacks];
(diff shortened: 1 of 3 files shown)


elhigu:
Thanks! I’ll check this in near future, with brief look only doc updates are missing...

schmod:
Got it. I can write a separate PR in the other repo.

elhigu:
@schmod thanks, please add a link to that to this PR then, so that it is easy to track when doing release.

src/query/builder.js
```diff
@@ -605,41 +605,41 @@ assign(Builder.prototype, {
...
   // Adds a union all statement to the query.
-  unionAll(callback, wrap) {
-    this._statements.push({
-      grouping: 'union',
-      clause: 'union all',
-      value: callback,
-      wrap: wrap || false,
-    });
-    return this;
+  unionAll(callbacks, wrap) {
```
elhigu: It would be more clear to use `unionAll(...args)` and `union(...args)` instead of using the old magic `arguments` array. Now arguments defined for `unionAll` are not unused.

schmod: I hesitated to use `...args` because knex doesn't use it in many places.  I can update if we're cool with it though.

schmod:
Documentation PR: https://github.com/knex/documentation/pull/180
Also updated this PR to use spread args instead of `arguments`.

elhigu:
I think something went really broken in that last change: 

for example like this: 
```
 6) Query Building Tests
       QueryBuilder
         multiple union alls:
      mysql: AssertionError: expected 'select * from `users` where `id` = ? union all  union all ' to deeply equal 'select * from `users` where `id` = ? union all select * from `users` where `id` = ? union all select * from `users` where `id` = ?'
      + expected - actual
      -select * from `users` where `id` = ? union all  union all 
      +select * from `users` where `id` = ? union all select * from `users` where `id` = ? union all select * from `users` where `id` = ?
```

schmod:
@elhigu My bad.  Just pushed a change that should have fixed it.

I'm still seeing a failure on the Node 8 and Oracle tests, but it looks like those are unrelated to this PR.

elhigu:
I restarted node 8 tests, since it looked like random fail.... will merge after those pass. Thanks!

PR #3074, opened 2019-02-28
Add size parameter to integer method

iamchathu:
Size parameter work when we pass it. But it's not in definition.

### DIFF
--- a/types/knex.d.ts
+++ b/types/knex.d.ts
@@ -552,7 +552,7 @@ declare namespace Knex {
     dropColumn(columnName: string): TableBuilder;
     dropColumns(...columnNames: string[]): TableBuilder;
     renameColumn(from: string, to: string): ColumnBuilder;
-    integer(columnName: string): ColumnBuilder;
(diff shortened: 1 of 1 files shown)


types/knex.d.ts
```diff
@@ -552,7 +552,7 @@ declare namespace Knex {
     dropColumn(columnName: string): TableBuilder;
     dropColumns(...columnNames: string[]): TableBuilder;
     renameColumn(from: string, to: string): ColumnBuilder;
-    integer(columnName: string): ColumnBuilder;
+    integer(columnName: string, size:  number): ColumnBuilder;
```
kibertoad: I'm curious which database you are using. Knex seems to accept this values for MSSQL and Oracle, but I can't find anywhere in their documentation that they support it. Not sure what would be the accurate name here, if there is anything in their documentation - we should use that name, but Knex itself calls it "length", not "size". In MySQL world it is "display width", but that is very MySQL-specific.

iamchathu: I'm using MySQL. I will update the parameter to length.

Issue #2581, opened 2018-04-26
Drop-and-recreate migration scripts do not work correctly with async/await


igor-savin-ht:
# Environment

Knex 0.14.6
PostgreSQL 10.1

# Bug

If migration script relies on async/await to wait for a column to be dropped before being created again, "column already exists" error is thrown. Same code works fine if .then block is used instead.

Code example:

```
exports.up = async (knex) => {
    await knex.schema.withSchema(`${SCHEMA}`).table('users', function (table) {
        table.dropColumn('officeId');
    });

    await knex.schema.withSchema(`${SCHEMA}`).table('users', function (table) {
        table.string('officeId', 255);
    });
};
```

igor-savin-ht:
@elhigu Yup, that works just fine.

elhigu:
That is really weird and doesn't make sense at all... if someone likes to write test case and figure out what is going on please do.

kibertoad:
Closing as cannot reproduce anymore. Added tests that are not failing in master.

laurent22:
Happened to me too with this kind of migration:

```typescript
export async function up(db: DbConnection): Promise<any> {
	await db.schema.alterTable('users', function(table: Knex.CreateTableBuilder) {
		table.dropColumn('max_item_size');
		table.dropColumn('can_share_folder');
		table.dropColumn('can_share_note');
	});

	await db.schema.alterTable('users', function(table: Knex.CreateTableBuilder) {
		table.integer('max_item_size').defaultTo(null).nullable();
		table.specificType('can_share_folder', 'smallint').defaultTo(null).nullable();
		table.specificType('can_share_note', 'smallint').defaultTo(null).nullable();
		table.bigInteger('max_total_item_size').defaultTo(null).nullable();
	});
}
```

My fix was to split it into two migrations, although it would be nice if it could be done without hack.

laurent22:
Additionally, this doesn't work either:

```typescript
await db.schema.alterTable('users', function(table: Knex.CreateTableBuilder) {
	table.dropColumn('max_item_size');
	table.dropColumn('can_share_folder');
	table.dropColumn('can_share_note');

	table.integer('max_item_size').defaultTo(null).nullable();
	table.specificType('can_share_folder', 'smallint').defaultTo(null).nullable();
	table.specificType('can_share_note', 'smallint').defaultTo(null).nullable();
	table.bigInteger('max_total_item_size').defaultTo(null).nullable();
});
```

[...]

PR #3080, opened 2019-03-03
Make TS stubs modern

felixmosh:
Small refactor to the TS stubs, so they will use arrow functions and es6 import / export.

### DIFF
--- a/src/migrate/stub/ts.stub
+++ b/src/migrate/stub/ts.stub
@@ -2,20 +2,20 @@ import * as Knex from "knex";
 
 <% if (d.tableName) { %>
 export async function up(knex: Knex): Promise<Knex.SchemaBuilder> {
-    return knex.schema.createTable("<%= d.tableName %>", function (t: Knex.AlterTableBuilder) {
+    return knex.schema.createTable("<%= d.tableName %>", (t: Knex.AlterTableBuilder) => {
         t.increments();
         t.timestamps();
     });
-};
+}
 <% } else { %>
 export async function up(knex: Knex): Promise<any> {
-};
+}
 <% } %>
 <% if (d.tableName) { %>
 export async function down(knex: Knex): Promise<Knex.SchemaBuilder> {
     return knex.schema.dropTable("<%= d.tableName %>");
-};
+}
 <% } else { %>
 export async function down(knex: Knex): Promise<any> {
-};
+}
 <% } %>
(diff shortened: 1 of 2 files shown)


src/migrate/stub/ts.stub
```diff
@@ -2,20 +2,20 @@ import * as Knex from "knex";
 
 <% if (d.tableName) { %>
 export async function up(knex: Knex): Promise<Knex.SchemaBuilder> {
-    return knex.schema.createTable("<%= d.tableName %>", function (t: Knex.AlterTableBuilder) {
+    return knex.schema.createTable("<%= d.tableName %>", (t: Knex.AlterTableBuilder) => {
         t.increments();
         t.timestamps();
     });
-};
+}
```
kibertoad: Why remove semicolons, though?

src/seed/stub/ts.stub
```diff
@@ -1,9 +1,9 @@
 import * as Knex from "knex";
 
-exports.seed = function (knex: Knex): Promise<any> {
+export async function seed(knex: Knex): Promise<any> {
```
kibertoad: if it returns promise anyway, why make it async?

elhigu: To allow using await inside the function.

src/migrate/stub/ts.stub
elhigu: Normal function declaration shouldn't have semicolon there. With arrow function, it would be obligatory though.

src/seed/stub/ts.stub
kibertoad: Await inside the function is kinda broken, though, not sure we want to encourage it: https://github.com/tgriesser/knex/issues/2581

elhigu: Yes we do :) That is really strange though. It doesn't make sense at all that it works when calling then(), but not with await... Maybe we have missed returning a promise from `.then` somewhere 🤔

kibertoad: I'll try writing a failing test for that case.

src/migrate/stub/ts.stub
kibertoad: @felixmosh So if I understand correctly, semicolons should be restored, since it's an arrow function?..

felixmosh: @kibertoad , no, `up` is not an arrow function :]
The function that we pass to `createTable` is an arrow func, and there are semi-colons there.

kibertoad: Fair enough.

PR #3081, opened 2019-03-03
Declare drivers as optionalDependencies (Fix #3059)

### LINKED ISSUE #3059: yarn plug and play and optional peerDependencies
Hi,

`yarn --pnp` is a really interesting thing to speedup module loading:
https://yarnpkg.com/en/docs/pnp
however it needs all modules to declare what they depend on to be able to work properly.

In the case of modules like knex, where one of the dependencies is declared by a parent module, the plug and play system fails (e.g. knex won't find pg).
pg, mysql2, sqlite3 (...) should be in peerDependencies, but optional.

This is precisely what yarn's peerDependenciesMeta field do:
https://github.com/yarnpkg/yarn/commit/011a634d787611e9c1f2f1b888212525f1fc096d

However i suppose it needs to be available in npm too before it can be added to knex's package.json.

Meanwhile, a simple solution would be to be able to pass knex the driver module itself, like
```
var pg = require('knex')({
  client: 'pg',
  driver: require('pg'),
  connection: process.env.PG_CONNECTION_STRING,
  searchPath: ['knex', 'public'],
});
```

kapouer:
Fix #3059 

This is a simpler approach using optionalDependencies.
It's already supported by npm, and yarn --pnp is happy with it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,30 @@
     "uuid": "^3.3.3",
     "v8flags": "^3.1.3"
   },
+  "peerDependencies": {
+    "mssql": "^5.1.0",
+    "mysql": "^2.17.1",
+    "mysql2": "^1.7.0",
+    "pg": "^7.12.1",
+    "sqlite3": "^4.1.0"
+  },
+  "peerDependenciesMeta": {
+    "mssql": {
+      "optional": true
+    },
+    "mysql": {
+      "optional": true
+    },
+    "mysql2": {
+      "optional": true
+    },
+    "pg": {
+      "optional": true
+    },
+    "sqlite3": {
+      "optional": true
+    }
+  },
   "lint-staged": {
     "*.{js,json}": [
       "prettier --write",


elhigu:
So what does yarn --pnp does in this case? pack all the optional drivers to .pnp.js ? I asked those questions in  #3059 because I really would like to know what actually happens :)

kapouer:
I PR'd this to be more explicit about the changes involved. Also that made me actually check what was happening in practice, as you asked.

The problem with optionalDependencies is that npm or yarn (even in --prod mode) try to install all of them and don't fail if they can't.

That's why the "peerDependencies" + "peerDependenciesMeta" was implemented by yarn: it silence the warnings about uninstallable peer dependencies, and do both the job of not installing the modules that are not already installed, and properly setup .pnp.js with the peerDependencies that are actually available.

kapouer:
Please don't merge this PR, it's work in progress.

kapouer:
The problem with peerDependenciesMeta is that it's not supported by npm, and there is no sign showing it will be soon. However missing peerDependencies only shows warnings, not failures, and i think most npm users just use them as hints when something goes wrong.

vlasenko:
@kapouer Is this PR still work in progress? And if yes, what are the blockers?

kapouer:
It can be merged, however it might make npm users yell about the warnings.

jthn:
Wondering if this could get merged.  I'd like to migrate a project to using Yarn PnP, and I'd love to be able to use it without forking this project.

kapouer:
Update: https://github.com/npm/cli/pull/224 now peerDependenciesMeta is supported by npm !

kibertoad:
Woohoo!
Not sure that declaring drivers as peerDependencies is correct approach, though, because it contradicts the original intent of peerDependencies: stating dependencies that consumer of the library is supposed to satisfy.

kibertoad:
Wait, `peerDependenciesMeta` actually solves that. Then it's OK to merge, after driver versions are brought up-to-date

kapouer:
Done. I supposed no new driver has been added since march 2019.

kibertoad:
Thanks!

kibertoad:
@jthn Released in 0.20.1

kapouer:
It's supported but still prints a WARNing.
If that's okay, it'd be safer to keep it that way until npm is fixed: It's a freshly deployed feature there, it might have been rolled back for any other reason.
I've notified upstream (in https://github.com/npm/cli/pull/224) just in case.

PR #3083, opened 2019-03-03
Tests for drop-and-recreate with async/await

kibertoad:
refs #2581


kibertoad:
Unfortunately I do not have access to the production code that was throwing that error originally so I cannot test it with latest knex version; but it does seem that problem does not manifest in current master.

PR #3084, opened 2019-03-03
Fix transaction support for migrations

### LINKED ISSUE #3063: Migrations can no longer be run from within a transaction
# Environment

Knex version: 0.16.0+
Database + version: Postgres 9.6.12
OS: MacOS 10.13

# Bug

**1. Explain what kind of behaviour you are getting and how you think it should do**

Since version 0.16.0, migrations can no longer be run within a transaction. My team is currently using this behavior to stand up a database for running automated tests, and reverting to a clean database by rolling back the transaction after all testing is complete.

Minimal example:

```javascript
const conn = require('knex')(...);
conn.transaction((txn) => {
  txn.migrate.latest();
});
```

**2. Error message**

`Cannot set user params on a transaction - it can only inherit params from main knex instance`

**3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real
   needed data and demonstrates the problem.**

Example (including docker postgres setup): https://github.com/jeffkenney/knex-txn-issue

kibertoad:
fixes #3063

### DIFF
--- a/src/migrate/Migrator.js
+++ b/src/migrate/Migrator.js
@@ -57,7 +57,7 @@ export default class Migrator {
       } else {
         this.knex = knex;
         this.knex.client.logger.warn(
-          'Running migrations in explicitly passed transaction. Make sure you are not using any custom identifier wrappers on original knex instance - they will not be disabled when executing internal queries and may result in them failing.'
+          'Running migrations in explicitly passed transaction. Make sure you are not using any custom identifier wrappers on original knex instance (wrapIdentifier param) - they will not be disabled when executing internal queries and may result in them failing.'
         );
       }
     } else {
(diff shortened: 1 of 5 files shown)


kibertoad:
This fix makes an assumption that we had a pretty good reason why we don't want to modify or clone transaction instances, hence we are only ensuring that they work as much as they used before. If it is OK to modify or clone transaction instances, better solution could be implemented.

src/migrate/Migrator.js
```diff
@@ -49,12 +49,16 @@ export default class Migrator {
   constructor(knex) {
     // Clone knex instance and remove post-processing that is unnecessary for internal queries from a cloned config
     if (isFunction(knex)) {
-      this.knex = knex.withUserParams({
-        ...knex.userParams,
-      });
-      this.knex.disableProcessing();
```
kibertoad: Moved this inside operations instead to ensure correct state before each of them, regardless of how previous ones ended.

PR #3090, opened 2019-03-07
Adds a beforeCreate pool hook to allow connectionSettings to be updated

lbarratt:


src/client.js
```diff
@@ -250,12 +250,20 @@ assign(Client.prototype, {
...
     return Object.assign(poolConfig, {
       create: () => {
-        return this.acquireRawConnection().tap((connection) => {
-          connection.__knexUid = uniqueId('__knexUid');
-
-          if (poolConfig.afterCreate) {
-            return Promise.promisify(poolConfig.afterCreate)(connection);
-          }
+        const beforePromise = poolConfig.beforeCreate ?
+          Promise.promisify(poolConfig.beforeCreate)(this.connectionSettings) :
```
kibertoad: Could you move `Promise.promisify(poolConfig.beforeCreate)` outside of the call so that it is only executed once?

kibertoad: Actually, since this is not being exposed to the end user, keeping it as a callback is also a viable option.

kibertoad:
New functionality needs tests and documentation in https://github.com/knex/documentation :)

lbarratt:
Apologies. This PR was opened  unintentionally and was my attempt at a spike to solve a problem. I will close this.

For anyone that may find this in the future, the was intended to solve the problem of being able to change the connection configuration before a new connection was acquired.

This is because my project uses [IAM authentication for AWS RDS](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html). This means my database user does not have a fixed password and uses temporary access tokens created by [RDS.Signer](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/RDS/Signer.html#getAuthToken-property). Another challenge is that the access tokens expire after 15 minutes, and you need to get a new token for each RDS cluster endpoint (cluster/read-only).

The eventual solution was to manage multiple `knex` instances with a pool of connections for each RDS endpoint. Then recycle the `knex` instances with new tokens every 10 minutes. Not ideal, but seems to do the job.

Being able to modify the connection configuration with a `beforeCreate` hook didn't really make since since the tokens expire in 15 minutes and managing individual connections in the pools seemed like a fair amount of effort.

elhigu:
> New functionality needs tests and documentation in https://github.com/knex/documentation :)

Also it should be discussed in feature request before implementing solution. But sounds like this was for personal use :)

> For anyone that may find this in the future, the was intended to solve the problem of being able to change the connection configuration before a new connection was acquired.

There was a suggestion for this kind of feature while ago, but its API was also not good for taking in https://github.com/tgriesser/knex/pull/3052

rfink:
Throwing in a similar thought here - also would love to be able to inject parameters dynamically. I'm on AWS, and for similar reasons, want to be able to refresh the credentials periodically. This feature would be great for environments where credentials can't be hardcoded or injected via the environment, as well as apps that use service discovery.

PR #3096, opened 2019-03-08
Update mssql dev dep to v5 stable

### LINKED ISSUE #3093: Upgrade mssql lib
The `mssql` package is now stable v5.0.0 and v6.0.0 is in alpha.

v5 mitigates some issues around `generic-pool` and it's *interesting* `acquire` logic.

v6 removes the `generic-pool` dependency and replaces it with `tarn.js`. v6 also upgrades the underlying `tedious` library to v6.

Both deal a lot better with lost connections.

I don't believe either introduce and breaking changes that would affect this library.

I'm happy to author the upgrade PR (most likely just an upgrade to the package.json) but wanted thoughts on what you'd prefer the baseline to test against?

dhensby:
closes #3093


kibertoad:
Thanks!

PR #3098, opened 2019-03-09
fixes #3097

### LINKED ISSUE #340: HAVING w/sub-query
Is it possible to build a query with sub-query in HAVING? Something like:

```
SELECT ..., COUNT(...) AS `offers_count`
FROM
    `fh_stock_order` AS 'sor',
    `fh_stock_offer` AS `sof`
WHERE ...
GROUP BY ...
HAVING
    `offers_count` = (
        SELECT COUNT(*)
        FROM `fh_stock_offer` AS `sof2`
        WHERE `sof2`.`order_id` = `sor`.`order_id`
    )
;
```

Target dialect is mysql 5.5, if it matters.

### LINKED ISSUE #3097: knex production distribution pulls @types/bluebird and @babel/polyfill
# Environment

Knex version: `-- knex@0.16.3`
Database + version: irrelevant, but well... - mysql 5.7
OS: irrelevant, but well - windows 10 with GNU on Windows (GOW)

# Bug

Having installed `knex` as a dependency - I Expect in `node_modules` only dependencies REQUIRED for `production` env.

### Installing knex

Scenario: 
 - run `npm intall knex`.
 - Look in your node_modules

Expect:
- only production time dependencies.

Found:
 - @types/bluebird and @babel/polyfill

Given:
 - some migrations, and some seeds

### running it without the suspected unecessary dependenceis

Scenario:
 - run `rm -rf node_modules/@types/bluebird node_modules/@babel/polyfill
 - run `knex migrate:latest`
 - run `knex seed:run`

Expect (if to believe the `dependencies` section of the `package.json`):
 - an error

Found:
 - the migrations and seeds ran successfully.

I did not proceed to check the rest of the dependencies - so there could be more irrelevant dependencies. I simply knew that `babel` and `types` are not supposed to be in production dependencies section 😕 and knew it would be super easy to test

Thanks, and keep up the good work!

osher:
fixes #3097

sorry, the number in the git comment is wrong :(

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
     "node": ">=6"
   },
   "dependencies": {
-    "@types/bluebird": "^3.5.26",
     "bluebird": "^3.5.3",
     "colorette": "1.0.7",
     "commander": "^2.19.0",


osher:
I will appreciate if someone internal will review the dependencies, and LMK if there is more dependency we may be able to drop out from the production `dependencies` section - and I'll commit it to this branch too to test it :)

kibertoad:
Imagine that you are using knex as a dependency. If bluebird typings are not dependencies, then you don't get them transitively and you don't have typings even though you are using bluebird interfaces in your code (as knex returns bluebird promises).

We did get rid of polyfill prod dependency though :)

osher:
<puzzled>
Excuse me, I cannot understand the sense of your answer - as far as I understand - your answer refers to **development** time, not **production** time....
I'm sure of that because interfaces belong to the IDE, not to production time.

Can you please help me understand what am I getting wrong?

PR #3099, opened 2019-03-09
Implement support for returning started transaction without using transaction() methods callback function

### LINKED ISSUE #3095: Transactions as a promise

Is there a specific design choice in transaction requiring a callback?

I would love to write my transactions with async/await like below.  And I'm not sure if the trx.commit() should be outside or inside the try/catch.

```js
  const trx = await client.transaction()

  try {
    await client
      .transacting(trx)
      .from('OtherItems')
      .whereIn('itemId', itemIds)
      .del()

    await client
      .transacting(trx)
      .from('Items')
      .whereIn('itemId', itemIds)
      .del()

    await trx.commit()
  } catch (err) {
    await trx.rollback()
    throw err
  }
```

As a work-around I made a utility function that I can await.  Is this going to cause me problems down the road?
```js
function getTransaction (client) {
  return new Promise(res => {
    client.transaction(trx => {
      res(trx)
    })
  })
}

const trx = await getTransaction(client)
```

kibertoad:
Original inspiration behind this change was discussion on implementing DDD pattern in Javascript that we had at our company. Currently there is no convenient way for implementation-level (repositories) entities to generate a transaction instance that could then be passed from domain-level (use cases) entities across all the business logic execution steps without knowing what exact db operations are going to be executed.

fixes #3095


kibertoad:
I'm not sure if it is confusing that meaning of a Promise that .transaction() returns is different based on provided params. It might be, and it may be a better solution to have separate method for this case, but I don't know what a good name that would be less confusing in that case would be.

capaj:
@kibertoad this looks useful. It will surely come in handy on our project at @LeapLabs. Thanks!

kibertoad:
@capaj You are welcome :)

src/util/make-knex.js
```diff
@@ -34,12 +34,21 @@ function initContext(knexFn) {
...
+    // Creates a new transaction.
+    // If container is provided, returns a promise for when the transaction is resolved.
+    // If container is not provided, returns a promise with a transaction that is resolved
+    // when transaction is ready to be used.
     transaction(container, config) {
       const trx = this.client.transaction(container, config);
       trx.userParams = this.userParams;
-      return trx;
+
+      if (container) {
```
elhigu: So is `BEGIN TRANSACTION` query already sent to the DB server at this point?

test/unit/knex.js
```diff
@@ -223,6 +223,26 @@ describe('knex', () => {
     });
   });
 
+  it('supports direct retrieval of a transaction without a callback', (done) => {
+    const knex = Knex(sqliteConfig);
```
elhigu: Integration test that is ran for every dialect would be good to have to make sure that returned `trx` is immediately ready for making queries.

elhigu:
Also might be a good idea to add to documentation something about availability of trx.initPromise for hooking functionality to transaction commit / rollback.

src/util/make-knex.js
kibertoad: Isn't that what `this.client.transaction(container, config);` does?

elhigu: probably I was thinking originally that this PR would `Implement support for starting transactions without executing them` thus I could create transaction object, which would not allocate connection before the first query is ran (very useful in for example rest frameworks to provide transaction to handler which will not use any resources, if no queries are sent to there)... But I can see that this is just returning open and started transaction instead.

I was just confused about title of this PR :)

kibertoad:
@elhigu Great point! There was, indeed, a gap in how errors were handled during transaction acquisition.

test/unit/knex.js
kibertoad: Done

kibertoad:
`Also might be a good idea to add to documentation something about availability of trx.initPromise for hooking functionality to transaction commit / rollback.`

Not really. initPromise is only used, as name implies, to track process of transaction itself getting prepared for usage. After this promise is resolved, it returns the transaction itself, and that's it. I don't think we should consider it a part of public api, as users are only supposed to consume it indirectly via knex.transaction()

elhigu:
> `Also might be a good idea to add to documentation something about availability of trx.initPromise for hooking functionality to transaction commit / rollback.`
> 
> Not really. initPromise is only used, as name implies, to track process of transaction itself getting prepared for usage. After this promise is resolved, it returns the transaction itself, and that's it. I don't think we should consider it a part of public api, as users are only supposed to consume it indirectly via knex.transaction()

Right. I though that promise was the promise that get resolved / rejected when trx.commit() / trx.rollback(err) was called.

kibertoad:
@elhigu Hmmm, I can't find your other comment about your expectation for this not to start the actual transaction until someone actually uses it. I think that is doable, although would require slightly more intrusive changes. Do you think we should go for it within the scope of this PR?

src/util/make-knex.js
kibertoad: right, this is that comment. this is totally feasible. do you think I should try implementing that?

kibertoad: @elhigu Implemented transactionProvider for lazy instantiation of transactions that covers use-case that you've described. Feedback most welcome!

elhigu:
> @elhigu Hmmm, I can't find your other comment about your expectation for this not to start the actual transaction until someone actually uses it. I think that is doable, although would require slightly more intrusive changes. Do you think we should go for it within the scope of this PR?

I don't think that is in the scope of this PR. I'll just rename the issue to be less confusing :)

kibertoad:
@elhigu It was a fairly trivial addition, though, so I ended up including it as a feature on top :)

kibertoad:
@elhigu Phew, build finally passes again. Could you please rereview?

kibertoad:
@capaj It was released as knex@0.17.0-next6, would be awesome if you could give it a try! Most likely final 0.17.0 stable will be published in a couple of days.

test/unit/knex.js
```diff
@@ -310,6 +310,40 @@ describe('knex', () => {
     });
   });
 
+  it('propagates error correctly when all connections are in use', function() {
+    this.timeout(30000);
```
elhigu: Why this needs 30secs timeout? 🤔

test/knexfile.js
```diff
@@ -19,6 +19,16 @@ var pool = {
   },
 };
 
+var poolSqlite = {
+  min: 0,
+  max: 1,
+  acquireTimeoutMillis: 5000,
```
elhigu: If possible we could try to have subsecond timeouts in tests to prevent test runs to take too long just waiting around.

test/unit/knex.js
kibertoad: It is unlikely to use any more than just a couple of seconds, it is going to crash and be handled way earlier than that.

test/knexfile.js
kibertoad: Tried reducing it, but I think I was getting some flaky tests that way. Decreased it, let's see how it goes.

PR #3100, opened 2019-03-09
Support nullable timestamps on MySQL

### LINKED ISSUE #3092: setting date time to null crashes migrations
# Environment

Knex version:
Knex CLI version:  0.16.3
Local Knex version:  0.15.2
Database + version:
MYSQL 5.7
OS: OSX

# Bug
Setting date time to null no longer works as expected. Using null causes error. Documentation dose not reflect this change or provide a defaultTo example for null date times. 

Not Working Example that used to work
```
    return Promise.all([
        knex.schema.createTable('users', function (t) {
            t.increments('id').primary();
            t.string('name');
            t.string('email');
            t.string('pw');
            t.timestamp('deleted_at').defaultTo(null);
            // t.timestamps();
        })
    ]);
}
```

 # Error message
```
Using environment: development
migration file "20181207102033_user.js" failed
migration failed with error: create table `permissions` (`id` int unsigned not null auto_increment primary key, `user_id` int unsigned, `allow_user_id` int unsigned, `enabled` boolean default '1', `start_at` timestamp, `end_at` timestamp, `deleted_at` timestamp, `created_at` datetime, `updated_at` datetime) - ER_INVALID_DEFAULT: Invalid default value for 'end_at'
Error: ER_INVALID_DEFAULT: Invalid default value for 'end_at'
```

kibertoad:
fixes #3092

### DIFF
--- a/src/dialects/mysql/schema/columncompiler.js
+++ b/src/dialects/mysql/schema/columncompiler.js
@@ -117,6 +117,13 @@ Object.assign(ColumnCompiler_MySQL.prototype, {
   // ------
 
   defaultTo(value) {
+    // MySQL defaults to null by default, but breaks down if you pass it explicitly
+    // Note that in MySQL versions up to 5.7, logic related to updating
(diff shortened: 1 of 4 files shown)


elhigu:
Thanks! this is great!

PR #3104, opened 2019-03-12
Include 'string' as accepted constructor type definition

enapupe:
Hi there,
I'm totally new to TS so I'm sorry if this is unexpected/unwanted.
I was wiring TS into my codebase and my config call - which happens via knex constructor) was failing type validations because it's set to only accept `Config` interface but I'm currently using a string (via https://github.com/tgriesser/knex/blob/master/src/util/parse-connection.js)

### DIFF
--- a/types/knex.d.ts
+++ b/types/knex.d.ts
@@ -875,7 +875,7 @@ declare namespace Knex {
   //
 
   class Client extends events.EventEmitter {
-    constructor(config: Config);
+    constructor(config: Config | string);
     config: Config;
     dialect: string;
     driverName: string;


types/knex.d.ts
```diff
@@ -875,7 +875,7 @@ declare namespace Knex {
   //
 
   class Client extends events.EventEmitter {
-    constructor(config: Config);
+    constructor(config: Config | string);
     config: Config;
```
enapupe: Not sure if this call should also accept string or not.

kibertoad: @enapupe Are you sure about this? I checked the code and I'm pretty sure clients (at least base one and PostgreSQL) are going to explode if you pass string to them, because of e. g. `config.pool.max !== 0`
And `parseConnectionString` is not called anywhere in the code, so I assume what you want to do is to call `parseConnectionString` with string, get instance of Config and pass that to client constructor. And you don't need to modify typings for that.

enapupe: Totally not sure. I have no idea if the `config` call should also allow string. 
I'm positive the `constructor` should. That's why I haven't touched the  referred line. I appreciate you checking it

kibertoad: @enapupe config is not a method, it is a field, you don't call it. And please check source code for client constructor (src/client.js) or just try running it the way you are describing. I'm pretty confident it will crash. If it will not, please share entire fragment of a code you are using so that I understand your use-case better, because currently it looks wrong.

kibertoad: Actually, I'm not seeing any typings for parseConnectionString function. I think that's where your problem is. It needs to be typed to return Config instance, then you could pass its results to the constructor of Client.

enapupe: I'm thinking I've touched the wrong line. What I want is to allow `knex(DATABASE_URL)` accept a string. Sorry should have added that context to the PR description.

enapupe: And now I see it seems to accept the following format: `knex({ connection: DATABASE_URL })`, which totally works for me.

kibertoad: Ah, that's whole different story:
```
export default function Knex(config) {
  // If config is a string, try to parse it
  if (typeof config === 'string') {
    const parsedConfig = Object.assign(parseConnection(config), arguments[2]);
    return new Knex(parsedConfig);
  }
```

kibertoad: Would be cool to expand typing, though, as passing just a string is also legit

enapupe: Right! Will try again.
(I've deleted the fork accidentally so will have to put up separate pr as I cannot reopen this one)

enapupe: https://github.com/tgriesser/knex/pull/3105

PR #3105, opened 2019-03-12
Include 'string' as accepted Knex constructor type definition

enapupe:
Hopefully I got it right this time!

Allowing Knex() to accept string type as it does accept string configuration parameter: https://github.com/tgriesser/knex/blob/master/src/knex.js#L12


kibertoad:
Beautiful. Thank you!

enapupe:
Thanks for being so responsive!

kibertoad:
haha, np!

PR #3111, opened 2019-03-17
Fix queryContext not being passed to raw queries

### LINKED ISSUE #3110: queryContext not passed to wrapIdentifier when `raw` is used
# Environment

Knex version: 0.16.3
Database + version: Any
OS: Any

# Bug

Hi!

So I noticed that, at least in some cases, `queryContext` is not passed to `wrapIdentifier` called by raw instances inside queries. Here's a reproduction:

```js
const Knex = require('knex');

const knex = Knex({
  client: 'sqlite3',
  useNullAsDefault: true,
  debug: false,
  connection: {
    filename: ':memory:'
  },

  wrapIdentifier(str, origImpl, queryContext) {
    if (str === 'iAmGoingToBeConvertedToId') {
      str = 'id'

      if (!queryContext) {
        throw Error('We should have queryContext here right?')
      }
    }

    return origImpl(str)
  }
});

async function main() {
  await knex.schema.dropTableIfExists('test')
  await knex.schema.createTable('test', table => {
    table.increments('id')
    table.string('text')
  })

  await knex('test')
    .queryContext({ someStuff: true })
    .select('id')
    .where('id', knex.raw('??', 'iAmGoingToBeConvertedToId'))
}

main()
  .then(() => {
    return knex.destroy()
  })
  .catch(err => {
    console.error(err)
    return knex.destroy()
  })
```

kibertoad:
fixes #3110

### DIFF
--- a/src/formatter.js
+++ b/src/formatter.js
@@ -138,6 +138,12 @@ export default class Formatter {
     }
     if (value instanceof Raw) {
       value.client = this.client;
+      if (this.builder._queryContext) {
+        value.queryContext = () => {
+          return this.builder._queryContext;
+        };
+      }
+
       query = value.toSQL();
       if (query.bindings) {
         this.bindings = this.bindings.concat(query.bindings);
--- a/test/unit/knex.js
+++ b/test/unit/knex.js
@@ -185,6 +185,44 @@ describe('knex', () => {
     ).to.equal(null);
   });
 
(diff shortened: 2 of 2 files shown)


kibertoad:
> Just to be safe would be good to add also a test where feature is used in transaction. 

Aye!

PR #3114, opened 2019-03-21
Use URI mode with sqlite3 dialect

doodlydroll:
When sqlite3 database object is created pass 2nd parameter mode
with value

   OPEN_URI | OPEN_READWRITE | OPEN_CREATE

This will enable URI parsing of filename. If filename is not a valid URI
then things will work as if OPEN_URI was not passed.

This allows following use to be possible with sqlite3:

  In memory db with shared cache (different connections to same URI will
  connect to same in-memory db and will see same content):

     filename: "file:./db-file1.sqlite3?mode=memory&cache=shared"

  In memory db with private cache (different connections to same file will
  not connect to same in-memory db and therefore will see different content):

     filename: "file:./db-file1.sqlite3?mode=memory&cache=private"

Sample test (requires chai-as-promised):
```
knex = require("./lib/knex");
util = require("util");
chai = require("chai");
chai.use(require('chai-as-promised'));

describe("sqlite with uri and mode=memory", () => {

    it("mixed shared/private cache", async () => {
        const conf1 = {
            client: "sqlite3",
            connection: {
                filename: "file:./db-file1.sqlite3?mode=memory&cache=shared"
            },
            useNullAsDefault: true,
        };
        const conf2 = {
            client: "sqlite3",
            connection: {
                filename: "file:./db-file1.sqlite3?mode=memory&cache=private"
            },
            useNullAsDefault: true,
        };
        const client1 = knex(conf1);
        const client2 = knex(conf2);
        try {
            await client1.schema.createTable("test", function (table) {
                table.increments();
                table.string("s");
            });

            await client1("test").insert({ s: "foo" });
            await client1("test").insert({ s: "bar" });

            console.log("-- p1 --");
            const p1_rows = await client1("test").select();
            p1_rows.forEach(x => console.log(`row: ${util.inspect(x)}`));
            chai.expect(p1_rows.length).to.equals(2);

            console.log("-- p2 --");
            const f = async () => (await client2("test").select()).forEach(x => console.log(`row: ${util.inspect(x)}`));
            await chai.expect(f()).to.be.rejectedWith('SQLITE_ERROR: no such table: test');
        } finally {
            client1.destroy();
            client2.destroy();
        }
    })

    it("shared cache", async () => {
        const conf1 = {
            client: "sqlite3",
            connection: {
                filename: "file:./db-file1.sqlite3?mode=memory&cache=shared"
            },
            useNullAsDefault: true,
        };
        const conf2 = {
            client: "sqlite3",
            connection: {
                filename: "file:./db-file1.sqlite3?mode=memory&cache=shared"
            },
            useNullAsDefault: true,
        };
        const client1 = knex(conf1);
        const client2 = knex(conf2);
        try {
            await client1.schema.createTable("test", function (table) {
                table.increments();
                table.string("s");
            });

            await client1("test").insert({ s: "foo" });
            await client1("test").insert({ s: "bar" });

            console.log("-- p1 --");
            const p1_rows = await client1("test").select();
            p1_rows.forEach(x => console.log(`row: ${util.inspect(x)}`));
            chai.expect(p1_rows.length).to.equals(2);

            console.log("-- p2 --");
            const p2_rows = await client2("test").select();
            p2_rows.forEach(x => console.log(`row: ${util.inspect(x)}`));
            chai.expect(p2_rows.length).to.equals(2);
        } finally {
            client1.destroy();
            client2.destroy();
        }
    })
})

```

### DIFF
--- a/src/dialects/sqlite3/index.js
+++ b/src/dialects/sqlite3/index.js
@@ -63,6 +63,9 @@ assign(Client_SQLite3.prototype, {
     return new Promise((resolve, reject) => {
       const db = new this.driver.Database(
         this.connectionSettings.filename,
+        this.driver.OPEN_URI |
+          this.driver.OPEN_READWRITE |
+          this.driver.OPEN_CREATE,
         (err) => {
           if (err) {
             return reject(err);


kibertoad:
@doodlydroll Thank you for your contribution! Ideally this should be possible to enable this behaviour by passing a config option. This also needs documentation and a test.

doodlydroll:
Thanks for you answer @kibertoad. 
I guess there these use cases:

- use connection with URI string, fx, { client: "sqlite3", connection: "file:./foo?mode=memory&cache=shared" }
- use connection object with URI in filename, fx, { client: "sqlite3", connection: { filename: "file:./foo?mode=memory&cache=shared" }
- use connection object, fx, { client: "sqlite3", connection: { filename: "foo", mode: memory, cache: shared}}

But for backward compatibility: one should still support ":memory:" as special file name. If connection.filename is ":memory:" that would imply mode: memory, cache: private (by default).

Are these the use cases you think should be implemented?

Could you give me some feedback on how to implement the tests? If the above use cases should be implemented, that would be obvious thing that should be covered. What about the sample test I put in the comment? I was not sure how to integrate that into the existing tests. Should it be done as integration type of test which is sqlite specific or what would be a good approach?

kibertoad:
@doodlydroll Yup, definitely as sqlite-specific integration tests (skip if db is different).

Use-cases that you've listed definitely sound reasonable!
Feel free to add chai-as-promised as a dependency, but unfortunately we can't use async/await (yet) until Node 6 support is dropped.

Might actually be easier to implement them as unit tests to bypass the whole thing with knex instance being passed from above (as you don't really need it). It's not exactly a unit test, but then again, some of our current unit tests aren't either.

konsumer:
[This code](https://gist.github.com/konsumer/77c182cbe80a411d5f7d8745750fdb65) has been handling URIs (for sqlite and others) in a closer-to-expectation way, so I can share URIs with other systems and add db-specific options.

I'm using jest for testing, but would be happy to use mocha/chai, and make a PR, if there is interest.

```json
mysql://localhost/database?unix_socket=/var/lib/mysql/socket
 
 {
  "client": "mysql",
  "connection":  {
    "database": "database",
    "host": "localhost",
    "password": undefined,
    "port": null,
    "user": undefined,
  },
  "unix_socket": "/var/lib/mysql/socket",
}
 

mysql://username:password@hostname:666/databasename
 
 {
  "client": "mysql",
  "connection":  {
    "database": "databasename",
    "host": "hostname",
    "password": "password",
    "port": "666",
    "user": "username",
  },
}
 

postgres:///var/run/postgresql/test?debug=1
 
 {
  "client": "pg",
  "connection":  {
    "database": "var/run/postgresql/test",
    "host": "",
    "password": undefined,
    "port": null,
    "user": undefined,
  },
  "debug": 1,
}
 

postgres://username@hostname/databasename
 
 {
  "client": "pg",
  "connection":  {
    "database": "databasename",
    "host": "hostname",
    "password": undefined,
    "port": null,
    "user": "username",
  },
}
 

sqlite://:memory:?useNullAsDefault=true
 
 {
  "client": "sqlite3",
  "connection":  {
    "filename": ":memory:",
  },
  "useNullAsDefault": true,
}
 

sqlite://file?useNullAsDefault=true
 
 {
  "client": "sqlite3",
  "connection":  {
    "filename": "file",
  },
  "useNullAsDefault": true,
}
 

sqlite://path/to/file
 
 {
  "client": "sqlite3",
  "connection":  {
    "filename": "path/to/file",
  },
}
```

elhigu:
@konsumer that output doesn't seem right. I think we should not mix up connection string with complete knex configuration. Also connection strings are highly dialect specific and cannot be handled universally like that code seem to try to do (specially oracle and sql server has pretty strange ones). We should actually pass those strings to drivers  so that they can do what ever they like with them.

Of course we could have also knex's own universal connection string format which could be then parsed to connection object for each dialect, but still it should not set up other parameters of knex configuration (parameters outside of connection object) and translating parsed values to connection object should be done separately in each dialect's code. So parsing and converting parsed result to db driver specific configuration should be clearly separated.

konsumer:
@ekeric13 yeh, that one seems a bit off for some of the example URLs. I made a [more robust version](https://gist.github.com/konsumer/406372eb12d534c1f27f4dfffde42492) that uses a few modules, and uses a ton of [test urls](https://github.com/kennethreitz/dj-database-url/blob/master/test_dj_database_url.py), many of which wouldn't work in knex, but I threw them in to see how it would dealt with them.

>  I think we should not mix up connection string with complete knex configuration. 

For me, I use connection-strings exclusively to define all my database config in environment variables, so it's useful for me to take this approach, but I'm fine with just using my own code outside of knex (as I do now.)

> We should actually pass those strings to drivers so that they can do what ever they like with them.

This would be fine with me too, but would require some re-working of how knex is structured, and I think it might lead to every driver handling urls a little differently, instead of a fairly universal agreed-on format like `DRIVER://USER:PASSWORD@HOST:PORT/DATABASE?OPTION1=VALUE&OPTION2=VALUE`, which is my main goal, in the first place. My reason for using my own URL parser is that the current one had issues with some URLs that work elsewhere, and didn't work with options (as it does in many other database libraries, in several languages.)

> Of course we could have also knex's own universal connection string format which could be then parsed to connection object for each dialect

Yeh, I actually don't mind ODBC/JDBC type strings, which aren't URLs at all, but clearly define every param. For me, the standard URL format works fine, though, and I would assume is a bit easier for more people to digest without having to go lookup a special connection-string dialect.

loynoir:
> I think we should not mix up connection string with complete knex configuration.

@konsumer @elhigu 
As browsers **NOT** send `location.hash` to servers.
I think, may be like this?
1. drop connection params in `location.search`
2. drop knex config params in `location.hash`

For example
`sqlite://file?connectionParam1=foo&connectionParam2=bar#useNullAsDefault=true&knexPara1=hello`

PR #3118, opened 2019-03-26
Update builder.js

adailsonm:
inRamdomOrder


kibertoad:
Test needed :)

elhigu:
I don't think this is a good idea anyways, we cannot add own builder method for everything. orderByRaw('RAND()') is less opinionated way to do this with small enough overhead.

PR #3119, opened 2019-03-26
update proposal include inRandomOrder

adailsonm:
makes life easier for those who need to sort their data randomly.


elhigu:
I don't think it is a good idea to start adding this kind of trivial and really rarely needed builder methods. If you still want to push this feature idea with community please open a feature request where wether this is something we like to have can be discussed in greater detail.

PR #3121, opened 2019-03-28
Add boolean as a column name in join

felixmosh:
Added missing type

### DIFF
--- a/types/knex.d.ts
+++ b/types/knex.d.ts
@@ -239,7 +239,7 @@ declare namespace Knex {
     (tableName: TableName | Identifier | QueryCallback, clause: JoinCallback): QueryBuilder;
     (
       tableName: TableName | Identifier | QueryCallback,
-      columns: { [key: string]: string | number | Raw }
+      columns: { [key: string]: string | number | boolean | Raw }
     ): QueryBuilder;
     (tableName: TableName | Identifier | QueryCallback, raw: Raw): QueryBuilder;
     (


types/knex.d.ts
```diff
@@ -239,7 +239,7 @@ declare namespace Knex {
     (tableName: TableName | Identifier | QueryCallback, clause: JoinCallback): QueryBuilder;
     (
       tableName: TableName | Identifier | QueryCallback,
-      columns: { [key: string]: string | number | Raw }
+      columns: { [key: string]: string | number | boolean | Raw }
```
kibertoad: When would it be a boolean?

felixmosh: What do you mean? 
The value of join can be boolean :)

kibertoad: Could you please provide example code that was invalid before this change?
The way I'm reading this binding is this:
```
knex('users')
  .join('contacts', 'users.id', '=', 'contacts.user_id')
  .select('users.id', 'contacts.phone')
```
`columns` are these: ` 'users.id', '=', 'contacts.user_id'`
Why would they ever be boolean?

felixmosh: Pay attention to the interface that I've fixed, it is the one that accepts an object.

```javascript
db('tableName')
    .leftJoin('otherTable', {
      'tableName.isDefault': true
    })
``` 
image

It tries to resolve it using the `Raw` interface because there is no boolean.

kibertoad: Fair enough. And if typing is fixed, DB understands boolean value correctly?

felixmosh: Yeah, it works, it converts it into tinyInt automatically

elhigu: Actually that depends highly on db that is used, but something like that :) I had big problems trying to write cross db compatible code that had used boolean column type.

Issue #2445, opened 2018-01-30
Transactions hanging in AWS Lambda


samin:
# Environment
AWS Lambda NodeJS 5.6 with AWS RDS
Knex version: 0.14.2
Database + version: MySQL 5.7.19
OS: Linux

# Bug
Currently any query being made inside `knex.transaction` does not get executed, waiting to fail in timeout. In the example below, logs show `WITH KNEX TRANSACTION` but not `KNEX TRANSACTION BEGIN`. If `knex.transaction` is removed, code works as expected, writing into the `application` table. Am I missing something?

```
const knex = require('knex')({
  client: 'mysql',
  connection: {
    host: process.env.DB_HOST,
    port: process.env.DB_PORT,
    user: process.env.DB_USER,
    password: process.env.DB_PASSWORD,
    database: process.env.DB_NAME,
    ssl: 'Amazon RDS',
    charset: 'utf8mb4'
  },
  pool: false,
  debug: true
});

exports.handler = (event, context, callback) => {
  console.log('WITH KNEX TRANSACTION');

  knex.transaction((trx) => {
    console.log('KNEX TRANSACTION BEGIN');
    return knex('application').transacting(trx).insert({
      name: 'Hello World', user_id: 4, client_id: 'abcknex', client_secret: 'truknex', created: '2017-01-01'
    })
      .then(trx.commit)
      .catch(trx.rollback);
  })
    .then((resp) => {
      console.log('Transaction complete.');
      callback(null, 'Some success message');
    })
    .catch((err) => {
      console.error(err);
    });
};
```

samin:
More details from CloudWatch, with `DEBUG=knex:*` turned on:

![image](https://user-images.githubusercontent.com/263049/35579020-c99ec61a-05cc-11e8-80ae-68375bfed43d.png)

elhigu:
at least you are using trasactions wrong way. *If you use* implicit commit/rollback (return promise from transaction handler) *you must not* call trx.commit/trx.rollback manually. 

Anyways those problems would not cause the functionality you just described it would occur later on. Where did you find out `pool: false` parameter?

samin:
Seems like an option to disable pool: https://github.com/tgriesser/knex/blob/45f5ffbad023e30eeb39162b4b743473f0188052/src/client.js#L51

Anyways, updated the code with your observations and using pool instead, still no luck, begins transaction but does not write or commits.

[...]

Issue #2302, opened 2017-11-02
getting "Knex: Timeout acquiring a connection. The pool is probably full." error


Suhel182:
I am facing "Unhandled rejection TimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?" issue. There are tons of running queries so i'm not able to find the query causing it. This problem seems to get serious when postgres size increases. Is there any solution?

skumar568:
I am also facing this issue.

wubzz:
The solution is to isolate the source and fix it. This is not a "query" error, but rather an application architecture issue. The error occurs when knex is unable to acquire a connection within (default) 60 seconds, and most likely the reason for that is that the pool is full.

In other words there are many things that can cause the error. Two of the most common are:
1. You forget to commit transactions / forget calling `.transacing(trx)` on query builders which causes the transaction to idle
2. The pool is continiously kept busy at max capacity for at least (default) 60 seconds non-stop, for instance due to heavy traffic or long running procedures.

You should look over your database state when this error occurs - are there a lot of idle connections? are there too many connections? etc.

rsfernandes:
I am also facing this error.

While we are working to dig our architecture connections, Theres a way to prevent this error to crash the application? Or at least the error could point out the stack ? 

`rejecter(new _bluebird2.default.TimeoutError('Knex: Timeout acquiring a connection. The pool is probably full. ' + 'Are you missing a .transacting(trx) call?'));
                 ^
TimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?`

wubzz:
@rsfernandes As long as you use `catch` handlers on your builders I don't see how this error would crash the entire application.

As for the stack, I recommend `longStackTraces`. There's nothing built into knex specifically for this error stack.

mifi:
FWIW I got this error with an AWS Aurora database service, and it was caused by auto shutdown feature which terminates the database after inactivity (auto-startup of the database takes like a minute or so, and that caused knex to time out)

elhigu:
https://github.com/tgriesser/knex/issues/3159 There is list of some work around suggestions for aurora.

booboothefool:
@wubzz 
> 2. The pool is continiously kept busy at max capacity for at least (default) 60 seconds non-stop, for instance due to heavy traffic or long running procedures.

[...]

Issue #2141, opened 2017-06-29
Knex: Timeout acquiring a connection. The pool is probably full.


himanshuoodles:
Hi @tgriesser ,
I am facing "Unhandled rejection TimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?" issue while making insert query to postgres database from 2 same application servers. How did you solve this problem ?
Please suggest some solution.

elhigu:
There is probably something wrong how you are using knex. Its impossible to say without seeing the failing code. Please add minimal test code, which can be used to reproduce the problem. 

Also this issue tracker is mainly for development, not for user support, so stackoverflow would be better medium for this.

atiertant:
@elhigu got the same problem using knex@0.12.9 on oracledb dialect via the stack [offshore](https://github.com/Atlantis-Software/offshore) / [offshore-sql](https://github.com/Atlantis-Software/offshore-sql)

Glavin001:
@atiertant What did you do to fix this problem with knex@0.12.9 and OracleDB?

atiertant:
@Glavin001 you should have a look at this: 
```
$ sqlplus / as sysdba
select resource_name, current_utilization, max_utilization, limit_value from v$resource_limit where resource_name in ('processes', 'sessions', 'transactions');
```

Issue #1909, opened 2017-02-13
TimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?.


CodisRedding:
This modules function (module.exports near the bottom) is called as new messages are found on the a queue, so for instance if RabbitMQ has 200 messages in it when this node process starts than it's going to iterate over all 200 immediately. My understanding of how pooling worked within knex was that as long as each function call, in this case, is made with a transaction that it should handle them based on the min and max pool settings.

```js
'use strict';

const bluebird = require('bluebird');
const lab = require('./models/lab');
const labBookshelf = require('./models/base');
const lodash = require('lodash');
const moment = require('moment');

function getOrder(data) {
  const labData = data.lab;

  return lab.order.forge({
    Comments: labData.Comments ? labData.Comments : null,
    EHRId: data.ehrId,
    Enabled: Boolean(labData.Enabled),
    ExternalEncounterId: labData.ExternalEncounterId ? labData.ExternalEncounterId : null,
    ExternalModifiedDateTime: moment(labData.ExternalModifiedDateTime, moment.ISO_8601).toDate(),
    ExternalOrderId: labData.ExternalOrderId,
    ExternalOrderStatus: labData.ExternalOrderStatus ? labData.ExternalOrderStatus : null,
    ExternalPatientId: labData.ExternalPatientId,
    ExternalProviderId: labData.ExternalProviderId ? labData.ExternalProviderId : null,
    ExternalWorkflowStatus: labData.ExternalWorkflowStatus ? labData.ExternalWorkflowStatus : null,
    ProviderSignoffDateTime: labData.SignoffDateTime ? moment(labData.SignoffDateTime, moment.ISO_8601).toDate() : null
  });
}

function getTests(data) {
  const labData = data.lab;
  let tests = [];

  labData.Tests.forEach(function(test) {
    tests.push({
      Comments: test.Comments ? test.Comments : null,
      Description: test.Description ? test.Description : null,
      EHRId: data.ehrId,
      Enabled: Boolean(test.Enabled),
      ExternalModifiedDateTime: moment(test.ExternalModifiedDateTime, moment.ISO_8601).toDate(),
      ExternalOrderId: test.ExternalOrderId,
      ExternalTestId: test.ExternalTestId,
      Name: test.Name,
      OrderingProviderName: test.OrderingProviderName ? test.OrderingProviderName : null,
[...]

PR #3132, opened 2019-04-10
Disable propagateCreateError by default

guerrerocarlos:
The attribute **_propagateCreateError_** should be set to **_false_** by default to prevent unexpected `Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?` error.

### Explanation: 

The **propagateCreateError** is [set to _true_ by default in **Knex**](https://github.com/tgriesser/knex/blob/1948c3d423b812e33c38da435bc449d3d898f5d7/src/client.js#L225) and throws a [TimeoutError](https://github.com/tgriesser/knex/blob/1948c3d423b812e33c38da435bc449d3d898f5d7/src/client.js#L312) if the first _create connection_ to the database fails, [preventing **tarn**](https://github.com/Vincit/tarn.js/blame/d2553fe92d07616fbaa0fd2cc4913a100f445597/README.md#L71) (the connection pool manager) from re-connecting automatically. 

The solution is to set **_propagateCreateError_** to **_false_** thus enabling **knex** to automatically reconnect on _create connection_ failure, instead of throwing the error, and only throwing it when the **_acquireTimeoutMillis_** has passed, [as defined by tarn](https://github.com/Vincit/tarn.js/blame/master/README.md#L73)

#### Related issues:
#1381 #1382 #1655 #2445 https://github.com/strapi/strapi/issues/2790 #1875 #1909 #2302 #2141 #1264

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -222,7 +222,7 @@ assign(Client.prototype, {
   },
 
   poolDefaults() {
-    return { min: 2, max: 10, propagateCreateError: true };
+    return { min: 2, max: 10, propagateCreateError: false };
   },
 
   getPoolSettings(poolConfig) {


elhigu:
Propagating the error is the correct and intentional functionality. If you think there is a problem, please add bug report with reproduction case.

PR #3144, opened 2019-04-15
adding types 

MichaelBitard:
for `havingBetween` `orHavingBetween` `havingNotBetween` `orHavingNotBetween`

### DIFF
--- a/types/knex.d.ts
+++ b/types/knex.d.ts
@@ -150,6 +150,10 @@ declare namespace Knex {
     orHaving: Having;
     orHavingRaw: RawQueryBuilder;
     havingIn: HavingIn;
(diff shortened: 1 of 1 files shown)


types/knex.d.ts
```diff
@@ -432,6 +436,23 @@ declare namespace Knex {
     (columnName: string, values: Value[]): QueryBuilder;
   }
 
+  interface OrHavingNotBetween {
+    (columenName: string, values: Value[]): QueryBuilder;
```
kibertoad: Should be "columnName"

MichaelBitard: thanks!

PR #3149, opened 2019-04-17
Add error detail log to cli.js

risseraka:
Hi there,

A really small PR to add error details when using cli, instead of only having a stack trace, which makes it really hard to know where the issue is when seeding (e.g. `error: null value in column "date" violates not-null constraint`).

LMKWYT,
Cheers.


kibertoad:
Thanks!

PR #3157, opened 2019-04-22
Docker based test dbs

elhigu:
Probably will still fail, because travis might not have oracle client libs installed yet.

Also found out that:
- current test suite is not closing down all started databases (so mocha must be ran with --exit switch)
- acquireConnectionTimeout doesn't work for some reason with oracledb


scripts/docker-compose.yml
```diff
@@ -0,0 +1,81 @@
...
+  postgres:
+    image: postgres:10-alpine
+    ports:
+      - '25432:5432'
+    environment:
+      - POSTGRES_USER=testuser
+      - POSTGRES_PASSWORD=knextest
+      - POSTGRES_DB=knex_test
+  waitpostgres:
+    image: postgres:10-alpine
```
kibertoad: considering that both Amazon and Google clouds currently have Postgres 11, doesn't that version make sense?

test/unit/dialects/oracledb.js
```diff
@@ -7,7 +7,7 @@ const knex = require('../../../knex');
 const config = require('../../knexfile');
 const sinon = require('sinon');
 
-describe('OracleDb externalAuth', function() {
+describe.skip('OracleDb externalAuth', function() {
```
kibertoad: if this should be skipped, would be useful to have comment with explanation

scripts/docker-compose.yml
elhigu: sure :)

test/unit/dialects/oracledb.js
elhigu: yep I actually forgot to try to fix that. It caused all tests just hanging. I'll try to fix that.

.travis.yml
```diff
@@ -5,68 +5,33 @@ sudo: required
...
-  allow_failures:
-  - node_js: "10"
-    env: TEST_ORACLEDB=true NODE_VER=11 KNEX_TEST_TIMEOUT=60000
-  - node_js: "10"
-    env: TEST_ORACLEDB=true NODE_VER=10 KNEX_TEST_TIMEOUT=60000
-  - node_js: "10"
-    env: TEST_ORACLEDB=true NODE_VER=8 KNEX_TEST_TIMEOUT=60000
-  - node_js: "10"
-    env: TEST_ORACLEDB=true NODE_VER=6 KNEX_TEST_TIMEOUT=60000
+    #    - node_js: '11'
```
kibertoad: Were all versions other than 10 supposed to be commented out?

test/unit/knex.js
```diff
@@ -186,6 +187,10 @@ describe('knex', () => {
   });
 
   it('passes queryContext to wrapIdentifier in raw query', () => {
+    if (!sqliteConfig) {
```
kibertoad: why could it be not available?

elhigu: when someone passes for example `DB=oracledb` environment variable. In that case sqliteConfig is undefined.

.travis.yml
elhigu: While trying to fix setup to work in travis it is a bit faster to run tests only with one environment. When tests are working fine, then I'll enable the rest of the node versions.

scripts/docker-compose.yml
```diff
@@ -0,0 +1,106 @@
...
+      - oracledbxe
+    depends_on:
+      - oracledbxe
+    environment:
+      - ORACLE_HOME=/opt/oracle/product/18c/dbhomeXE
+    entrypoint:
+      - bash
+      - -c
+      - 'until /opt/oracle/product/18c/dbhomeXE/bin/sqlplus -s sys/Oracle18@oracledbxe/XE as sysdba <<< "SELECT 13376411 FROM DUAL; exit;" | grep "13376411"; do echo "Could not connect to oracle... sleep for a while"; sleep 5; done'
+#  testrunner:
```
elhigu: I was able to run oracle tests with this locally by copying needed client libs from oracledb container, so if nothing else works I'll add one more container, which actually runs the tests instead of travis trusty.

elhigu:
Only potential problem still here is that mysql server startup hangs some times... I'll try to rerun these to check out if that is still a problem.

elhigu:
Yep, still various tests are unstable with oracledb... I'll skip them on oracle and add an issue about them. Still seems like the problem is that the way how knex does transactions with oracle is not completely stable...

test/integration/builder/updates.js
```diff
@@ -70,6 +70,80 @@ module.exports = function(knex) {
         });
     });
 
+    it('should immediately return updated value for other connections when updating row to DB returns', function() {
```
elhigu: This is the new test case... if we can setup knex oracledb dialect to work in a way that this tests passes, then probably most of the other failing oracle tests will start to work

elhigu:
And still gotta fix those linting issues on sunday.

scripts/docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+  postgres:
+    image: postgres:alpine
+    ports:
+      - '25432:5432'
+    environment:
+      - POSTGRES_USER=testuser
+      - POSTGRES_PASSWORD=knextest
+      - POSTGRES_DB=knex_test
+  waitpostgres:
+    image: postgres:10-alpine
```
kibertoad: why do we need two different postgres images?

kibertoad:
@elhigu Merged changes from master that enabled running tests on Node 12. However, Oracle driver still doesn't support Node 12. Can we not execute Oracle tests on it?

package.json
```diff
@@ -20,11 +20,13 @@
     "liftoff": "3.1.0",
     "lodash": "^4.17.11",
     "mkdirp": "^0.5.1",
+    "oracledb": "^3.1.2",
```
kibertoad: should it be prodDependency?

elhigu: nope, we don't have any of the drivers as knex deps to be able to install only one that is used in the project.

scripts/docker-compose.yml
elhigu: oops

elhigu:
> @elhigu Merged changes from master that enabled running tests on Node 12. However, Oracle driver still doesn't support Node 12. Can we not execute Oracle tests on it?

No, oracle tests cannot be ran with node 12 yet. Also looks like that something else came with you merge here, since there is new failing test (TypeError: qb(...).table(...).having(...).clearHaving is not a function)...

kibertoad:
@elhigu Pull changes from master, I would expect that to fix clearHaving problem

package.json
kibertoad: should be moved then

elhigu:
> @elhigu Pull changes from master, I would expect that to fix clearHaving problem

Yep... I'm rebasing back to top of master... now there are just bunch of conflict to fix 😬

kibertoad:
@elhigu I can resolve conflicts if you want.

elhigu:
Looks like all of them was for package.json + travis (but had to fix them on every commit). You could check afterwards that they look fine after I'll update this branch (I might have lost some of the changes / package version there).

package.json
```diff
@@ -80,21 +83,22 @@
     "rimraf"
   ],
   "scripts": {
-    "babel": "rimraf ./lib && babel src --out-dir lib --copy-files",
+    "format": "prettier --write \"{src,bin,scripts,test}/**/*.js\"",
     "build": "npm run babel",
-    "debug:test": "node --inspect-brk ./node_modules/.bin/_mocha -- --exit -t 0 test/index.js",
+    "debug_test": "mocha --inspect-brk --exit -t 0 test/index.js",
```
kibertoad: duplicate, there's already "debug:test"

package.json
```diff
@@ -80,21 +83,22 @@
...
     "lint": "eslint '{src,test}/**/*.js'",
     "lint:types": "dtslint types",
-    "plaintest": "mocha --exit -t 10000 -b -R spec test/index.js && npm run tape",
+    "plaintest": "mocha --exit -t 10000 test/index.js && npm run tape && npm run test:cli",
     "plaintest:sqlite": "cross-env DB=sqlite3 npm run plaintest",
     "prepare": "npm run babel",
     "prepublishOnly": "npm run babel",
     "pretest": "npm run lint && npm run lint:types && npm run babel",
-    "test": "nyc mocha --exit --check-leaks --globals __core-js_shared__ -t 10000 -R spec test/index.js && npm run test:tape && npm run test:cli",
+    "test": "nyc mocha --exit --check-leaks --globals __core-js_shared__ -t 10000 test/index.js && npm run tape && npm run bin_test",
```
kibertoad: Should be "npm run test:tape && npm run test:cli""

package.json
```diff
@@ -103,6 +107,13 @@
     "mssql:test": "DB=mssql npm run plaintest",
     "mssql:destroy": "docker-compose -f scripts/mssql-docker-compose.yml stop",
     "postmssql:init": "node scripts/wait-for-mssql-server.js && npm run mssql:logs || (npm run mssql:logs;false)",
+    "prepublish": "npm run babel",
+    "pre_test": "npm run lint && npm run lint_types",
+    "bin_test": "cross-env KNEX_PATH=../knex.js KNEX=bin/cli.js jake -f test/jake/Jakefile",
+    "tape": "node test/tape/index.js | tap-spec",
+    "debug_tape": "node --inspect-brk test/tape/index.js",
```
kibertoad: duplicate, there's already "debug:tape"

package.json
```diff
@@ -103,6 +107,13 @@
     "mssql:test": "DB=mssql npm run plaintest",
     "mssql:destroy": "docker-compose -f scripts/mssql-docker-compose.yml stop",
     "postmssql:init": "node scripts/wait-for-mssql-server.js && npm run mssql:logs || (npm run mssql:logs;false)",
+    "prepublish": "npm run babel",
+    "pre_test": "npm run lint && npm run lint_types",
```
kibertoad: duplicate, there's pretest

package.json
```diff
@@ -103,6 +107,13 @@
...
+    "bin_test": "cross-env KNEX_PATH=../knex.js KNEX=bin/cli.js jake -f test/jake/Jakefile",
```
kibertoad: duplicate with test:cli

package.json
```diff
@@ -103,6 +107,13 @@
...
+    "tape": "node test/tape/index.js | tap-spec",
```
kibertoad: duplicate with test:tape

package.json
elhigu: is pretest run automatically before publish?

elhigu: aa you meant pre_test 👍

package.json
elhigu: how about that "prepublishOnly", when that is ran?

kibertoad: npm is throwing deprecation warning about prepublish now. Now there are two separate hooks: `prepublishOnly` for publish and `prepare` for after install. prepublish should actually be removed, next major npm version won't support it.

elhigu: thanks, I'll remove prepublish already then. Its really unlikely that we would publish anything without running babel first, since its pretty much included in every hook there :)

package.json
kibertoad: oracledb is still in prod dependencies :D

elhigu:
Again something really strange happens in travis that doesn't happen locally... created another simpler project to explore that https://github.com/elhigu/travis-node-oracle-18-xe 

Basically locally installed libs are found in /usr/lib/oracle... inside docker container, but in travis that directory doesn't seem to be created, even that both uses the same docker image where to install them.

kibertoad:
Finally green! Should we merge this?

package.json
```diff
@@ -24,7 +24,8 @@
     "tarn": "^1.1.5",
     "tildify": "1.2.0",
     "uuid": "^3.3.2",
-    "v8flags": "^3.1.3"
+    "v8flags": "^3.1.3",
+    "webpack": "^4.30.0"
```
kibertoad: wait, webpack is a runtime dependency?..

elhigu: it appeared there master merge / rebase mess. I thought it was some, maybe typescript related stuff that other people had contributed lately... but anyways it shouldn't be in this PR.

elhigu:
huh finally its over.... I saw one random hanging after tests, which might mean that some connection is left open in some cases. or pool cannot tear down properly. If that happens we could add flag that forces mocha to exit even if there are connections left open and see if stuff works better when tests are written in a more clean manner with async / await .

Issue #3131, opened 2019-04-09
rollback all running in wrong order


ctrimble:
# Environment

Knex version: 0.16.4-next2
Database + version: Aurora PostgreSQL
OS:

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do

We ran `knex.migrate.rollback(migrationsConfig, true)` in an effort to rollback a set of legacy migration files.  We expected this to execute the scripts in reverse chronological order, so that it would have the same effect as repeated calls to `knex.migrate.rollback(migrationsConfig)`.  Instead, it appears to be calling down on the migration files in chronological order.  This caused our rollback to fail.  As a workaround, we made repeated calls to rollback, without the `all` flag, and this worked.

2. Error message

migration failed with error: drop table if exists "client" - cannot drop table client because other objects depend on it

filbertteo:
I encountered the same issue too.

kibertoad:
0.16.6 is out which should address this problem. Please reopen the issue if you are still experiencing the problem after update.

Issue #2907, opened 2018-11-14
Knex migrations refresh


theprobugmaker:
I was wondering if it's possible to refresh the migrations from the database, cleaning everything and calling migrations again.

Things that would be nice:

- `refresh` - rollback everything and call migrations again
- `reset` - rollback all the migrations

Also if it had a `--seed` so it would call `seed` automatically as well, like `knex migrate:refresh --seed`

kibertoad:
@zefexdeveloper This is similar to https://github.com/tgriesser/knex/issues/2844
After that one is implemented, one can have this functionality by simply chaining rollback + migrations + seeds, why have redundant operations for that?

zefexdeveloper:
@kibertoad It was just a suggestion, it was nice that the similar issue mentioned Laravel because that was the main reason I suggested `--seed` and the other functionalities, but chaining would work as well.

bentesha:
Anyone working on this? Will have some free time next week.

ffxsam:
Just started using knex and realized `migration:reset` does not exist like it does in Adonis. Seems like a pretty major thing to not include.

elhigu:
@ffxsam pretty major or pretty trivial and rarely needed. It is usually faster to drop old db and re-create from ground up.

ffxsam:
@elhigu "rarely needed" - not during the development phase, when one is still working on the schema. With Adonis, we've run `migration:reset` several times in the beginning.

elhigu:
@ffxsam if you really run migrations down for resetting your DB I suppose that for you it would be frequently needed operation. Still trivial to to implement in your own npm scripts.

I could never recommend resetting DB like that. It is slow as hell and if there is single poorly written down script it can leave state to database so that next time when you run migrations initial state is different than when running migrations on clean database. 

That being said, I'm totally fine if you like to implement it to knex. I never use knex's command line client to anything anyways.

MrToxy:
Any news on this?

kibertoad:
@MrToxy Is this https://github.com/knex/knex/pull/3187 what you are looking for?

MrToxy:
> @MrToxy Is this #3187 what you are looking for?

Not really, I don't think  #3187  is actually correct in the "refresh" context.
Taking a look at laravel's artisan, running `php artisan migrate:refresh` will re-run all of the migrations creating fresh tables with no data, that's exactly what I'm looking for.

kibertoad:
[...]

Issue #1550, opened 2016-07-04
Migrate to the desired migration.


Claud:
Please add this functionality. Like this:

`knex migrate:20160701152123_name`

MatthewAry:
It's been over a year, I would like to see this too.

olingern:
I think this can be closed as of #3205

ivanovych666:
PR https://github.com/tgriesser/knex/pull/3416 now is merged and I think that this issue can be closed. Now we can migrate up/down the specified migration like this:
```
knex migrate:up 001_migration_name.js
knex migrate:down 001_migration_name.js
```

Issue #783, opened 2015-04-19
Feature request: migrate:reset


light24bulbs:
I frequently change my migrations in my application since it has no production data to maintain yet.  I'd like to be able to quickly reset it(rollback all and then migrate latest) instead of running rollback until all of the migrations have rolled back to the base migration, then running migrate latest.

bendrucker:
Hey there,

I'm working on the rewrite of migrations (https://github.com/knex/arctic)

I'm willing to consider this but I'm not sure about it quite yet. Are you primarily using this locally? If so wouldn't a script to drop the DB, create it, and then migrate to the latest be a better solution?

light24bulbs:
It would certainly work.  I'll throw one together.  I was just thinking that it's probably a common enough use case to get put in the lib, but I defer to your judgement.  Maybe a `migrate:rollback --all` would help as well.  This library has been absolutely outstanding.  Thanks for all the hard work!

bendrucker:
Totally with you on an `--all` flag for migrations. The current API basically just supports `latest` and `rollback` but I'm looking to offer much finer control. Will keep you posted.

robertd:
@bendrucker What's the status of arctic?

mbavio:
@bendrucker Is Arctic going to happen soon?

justincy:
> wouldn't a script to drop the DB, create it, and then migrate to the latest be a better solution?

I ran into multiple issues doing this locally with Postgres such as the difficulty of running commands via the CLI, system permissions, db permissions, etc. I did manage to get it to work, but then realized I wanted to run the script both from the CLI and before tests, including re-running migrations and eventually seeds, so I examined the source of [Migrate](https://github.com/tgriesser/knex/blob/8a303f1a964acc98205575b1fcb594f5d6b0ea90/test/integration/migrate/index.js) and camp up with this simple routine:

```js
var migrate = knex.migrate;

// Force unlock in case of bad state
migrate.forceFreeMigrationsLock()

// Get completed migrations
.then(function(){
  return migrate._listCompleted();
})

// Rollback migrations
.then(function(completedMigrations){
  return migrate._waterfallBatch(0, completedMigrations, 'down');
})

// Migrate to latest
.then(function(){
  return migrate.latest();
});
```

It should be fairly simple to turn that into a `reset()` method in Migrate.

light24bulbs:
That's an improvement over what I do to clean before my tests, which is just 
```coffeescript
[...]

PR #3169, opened 2019-04-28
WIP : Implement more migration features, fix a rollback all bug

leeallen337:
This pull request aims to address a few issues (e.g. feature request and a bug) that I will elaborate on below.

1. ~~https://github.com/tgriesser/knex/issues/3131 - This bug looks like when trying to rollback all migrations it's running them in chronological order and, I think, attempting to rollback all migrations and not just the ones completed. From [this](https://github.com/tgriesser/knex/blob/master/src/migrate/Migrator.js#L118) line in the `Migrator`, `val[0]` I think is a list of all migrations in chronological order rather than only attempting to rollback the completed migrations in reverse chronological order.~~ 
See https://github.com/tgriesser/knex/pull/3172 for a pull request which address this specific bugfix

1. https://github.com/tgriesser/knex/issues/352, https://github.com/tgriesser/knex/issues/1550 and https://github.com/tgriesser/knex/issues/666 - These issues are requests to support migrating or down to specific versions and just migrating up one migration at a time or rolling back one migration at a time. I've implemented functionality to run `migrate:up` or `migrate:down`, which will migrate only the next migration or rollback only the last migration, respectively. There is also an optional `--to` flag which you can specify a specific the filename of the migration to migrate to (inclusive) or rollback to (exclusive) which I think addresses the issue.

1. ~~https://github.com/tgriesser/knex/issues/783 - I also add in the `--all` flag to the `migrate:rollback` cli command which I think would address this issue~~
See pull request https://github.com/tgriesser/knex/pull/3187

1. ~~https://github.com/tgriesser/knex/issues/2907 - This issue is kind of addressed. There's not a command to rollback everything and run latest in one command but a combination of `migrate:rollback --all` and `migrate:latest` would address this issue~~
See pull request https://github.com/tgriesser/knex/pull/3187

I've marked this issue as still a work in progress for a few reasons.
1. I wanted to get some feedback/comments/suggestions etc. from the maintainers about the functionality and implementation in this pull request before proceeding any further.
1. There are probably some things missing like tests, documentation, etc. I see several test files, is there a specific file tests for these should live in?

Possible ToDos:
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -203,6 +203,31 @@ function invoke(env) {
     });
 
   commander
+    .command('migrate:up')
+    .description('        Run the next migration.')
+    .option(
+      '--to <migration>',
+      'The migration file name to run up to and include'
+    )
+    .action((cmd) => {
+      pending = initKnex(env, commander.opts())
+        .migrate.up(null, cmd.to)
+        .spread((batchNo, log) => {
+          if (log.length === 0) {
+            success(color.cyan('Already up to date'));
+          }
+          success(
+            color.green(
+              `Batch ${batchNo} ran the following migrations:\n${log.join(
+                '\n'
+              )}`
+            )
+          );
+        })
+        .catch(exit);
+    });
+
+  commander
     .command('migrate:rollback')
     .description('        Rollback the last set of migrations performed.')
     .option('--verbose', 'verbose')
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -207,7 +207,7 @@ function invoke(env) {
     .description('        Run the next migration.')
     .option(
       '--to <migration>',
-      'The migration file name to run up to and include'
+      'The migration file name to run up to (inclusive)'
     )
     .action((cmd) => {
       pending = initKnex(env, commander.opts())
@@ -245,6 +245,29 @@ function invoke(env) {
           );
         })
         .catch(exit);
+    });
+
+  commander
+    .command('migrate:down')
+    .description('        Rollback the last migration performed.')
+    .option(
+      '--to <migration>',
+      'The migration file name to migrate down to (exclusive)'
+    )
+    .action((cmd) => {
+      pending = initKnex(env, commander.opts())
+        .migrate.down(null, cmd.to)
+        .spread((batchNo, log) => {
+          if (log.length === 0) {
+            success(color.cyan('Already at the base migration'));
+          }
+
+          success(
+            color.green(
+              `The following migrations were rolled back:\n${log.join('\n')}`
+            )
+          );
+        });
     });
 
   commander
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -230,10 +230,11 @@ function invoke(env) {
   commander
     .command('migrate:rollback')
     .description('        Rollback the last set of migrations performed.')
+    .option('--all', 'rollback all migrations')
     .option('--verbose', 'verbose')
-    .action(() => {
+    .action((cmd) => {
       pending = initKnex(env, commander.opts())
-        .migrate.rollback()
+        .migrate.rollback(null, cmd.all)
         .spread((batchNo, log) => {
           if (log.length === 0) {
             success(color.cyan('Already at the base migration'));
--- a/src/migrate/Migrator.js
+++ b/src/migrate/Migrator.js
@@ -105,6 +105,7 @@ export default class Migrator {
       });
   }
 
+  // Run only the next migration or the next migrations up to a specific migration
   up(config, toFile) {
     this._disableProcessing();
     this.config = getMergedConfig(config, this.config);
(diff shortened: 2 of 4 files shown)


kibertoad:
@leeallen337 Could you please split this in multiple PRs, each of which would address just one thing?
Speaking of CLI tests - check out https://github.com/tgriesser/knex/tree/master/test/jake

bin/cli.js
```diff
@@ -202,13 +202,39 @@ function invoke(env) {
         .catch(exit);
     });
 
+  commander
+    .command('migrate:up')
+    .description('        Run the next migration.')
```
kibertoad: Description seems inaccurate.

bin/cli.js
```diff
@@ -222,6 +248,29 @@ function invoke(env) {
         .catch(exit);
     });
 
+  commander
+    .command('migrate:down')
+    .description('        Rollback the last migration performed.')
```
kibertoad: Description seems inaccurate.

kibertoad:
Bugfix for rollback order is very much welcome and would probably warrant one more release in 0.16.x line just for that change alone.

--up and --down commands make sense to me.

Could you please explain how differently --all is supposed to work from just a simple rollback?

kibertoad:
Oh, and documentation should be updated here: https://github.com/knex/documentation

leeallen337:
@kibertoad Happy to split these up. I'll make some new smaller individual pull requests to break these up. I'll tackle the bugfix rollback order one first since that's a bugfix rather than a new feature.

> Could you please explain how differently --all is supposed to work from just a simple rollback?

If you had batch 1, batch 2, and batch 3, you would have to run `migrate:rollback` three separate times to get to your initial state since, on its own, it just rolls back the latest batch. In the same scenario, you could run `migrate:rollback --all` just once and you would achieve the same thing which would rollback all batches. Hope that makes sense

bin/cli.js
leeallen337: What are your thoughts on `Run the next chronological migration that has not yet been run.`?

bin/cli.js
leeallen337: What are your thoughts on `Undo the last previous migration performed.`?

kibertoad:
Appreciate it!
Thank you for the explanation, yeah, I think that feature indeed makes sense.

bin/cli.js
kibertoad: I'm more concerned about 
a) usage of --to flag making description completely incorrect;
b) difference between last migration and last migration batch being not emphasized enough;

leeallen337: Oh I see what you mean. That makes complete sense! When working on a pull request for this I'll try and think of a more accurate and all encompassing description 👍

leeallen337:
@kibertoad I created a separate pull request for the `rollback all` bug. The pull request can be found at https://github.com/tgriesser/knex/pull/3172

I'm thinking of leaving this pull request open to serve as kind of a table of contents and documentation for these smaller ones I'm making. I can close this one after all these smaller ones have been finished. How does that sound? I also don't want to clog up the list of pull requests either so I could just go ahead and close this now too. Let me know what you think

kibertoad:
@leeallen337 Sure, no problem! Appreciate the contribution :)

kibertoad:
Is anything still left?

leeallen337:
@kibertoad Just a couple more things. The `migrate:down` functionality and then adding a `--to <specifiy migration>` flag to `migrate:up` and `migrate:down` so you can run specific migrations in a batch or roll back a group of specified migrations

src/migrate/Migrator.js
```diff
@@ -459,6 +528,34 @@ function validateMigrationList(migrationSource, migrations) {
...
+function getMigrationSubset(migrations, direction, toFile) {
+  const migrationCopy =
+    direction === 'up' ? [...migrations] : [...migrations].reverse();
+
+  if (!toFile) {
+    return migrationCopy.slice(0, 1);
+  }
+
+  const indexOfToMigration = findIndex(migrationCopy, (migration) => {
+    return migration.file === toFile;
```
andrewvsk: Assume we have following migrations:
- 20001101000001_initial.js
- 20001102000002_migration1.js
- 20001103000003_migration2.js

Here is we have three possible migrate 'up/down to' cases:
1. to filename - 20001102000002_migration1.js
2. to timestamp (equal to) - 20001102000002
3. to timestamp (lower / upper bound)
    - up to 20001103000000 = migrationsToRun: [20001101000001_initial.js, 20001102000002_migration1.js]
    - down to 20001102000000 = migrationsToRun: [20001103000003_migration2.js, 20001102000002_migration1.js]
    - use short form 20001103, or pattern 2000??03, or regexp: 20001103.*.js

Actually you are running migration to exactly specific file name. 
What do you think about other cases ?

PR #3171, opened 2019-04-28
Execute tests on Node 12

kibertoad:


kibertoad:
Currently blocked by sqlite support: https://github.com/mapbox/node-sqlite3/issues/1151

DRoet:
FYI the upstream issue has been fixed in sqlite `4.0.7`

kibertoad:
@elhigu Do you prefer to wait with merging this one until Docker tests one lands since this one would be easier to resolve conflicts for?

.travis.yml
```diff
@@ -60,13 +60,13 @@ services:
 addons:
   postgresql: '9.6'
   apt:
-    sources:
-      - mysql-5.7-trusty
```
kibertoad: MySQL 5.7 is default on Xenial, no explicit sources are needed.

.travis.yml
```diff
@@ -60,13 +60,13 @@ services:
...
     packages:
       - g++-4.8
       - gcc-4.8
+      - g++-5
```
kibertoad: new SQLite version requires gcc 5

kibertoad:
End changes are not that intrusive (since xenial actually has needed gcc deps by default), so conflict resolution should be in either case quite trivial.

PR #3177, opened 2019-05-02
Persistent connection when pool min 1 and max 1

mathieumagalhaes:
In case pool config max and min is set to one, acquire a connection once and never release it.
Re-acquiring a connection just returns the existing one.

This might not be the correct logic to put this in place when pool max/min is set to 1.
If that needs to change I would however still love to see this implemented with a 'keepAlive' flag or something like that.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -35,6 +35,7 @@ const { POOL_CONFIG_OPTIONS } = require('./constants');
 function Client(config = {}) {
   this.config = config;
   this.logger = new Logger(config);
+  this.connection = null;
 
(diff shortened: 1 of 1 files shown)


elhigu:
This just wont work any better than having that single connection in pool. Actually it would work a lot worse. Please open bug issue with reproduction of your problem.

elhigu:
Pool does keepalive connections automatically if connection in pool is broken for some reason.

elhigu:
If you are not able to write reproduction case for your problem or not able to even describe it im sure that bug is somewhere else than in knex 😕

PR #3181, opened 2019-05-05
Unify object options handling for datetime/timestamp across dialects and update type definitions

lorefnon:


package.json
```diff
@@ -85,7 +85,7 @@
     "babel": "rimraf ./lib && babel src --out-dir lib --copy-files",
     "coveralls": "nyc report --reporter=text-lcov | coveralls",
     "dev": "rimraf ./lib && babel -w src --out-dir lib --copy-files",
-    "lint": "eslint '{src,test}/**/*.js'",
+    "lint": "eslint src/**/*.js test/**/*.js",
```
elhigu: Probably some quoting is needed to prevent shell from expanding those wildcards on some platforms

src/dialects/oracledb/schema/columncompiler.js
```diff
@@ -12,12 +12,24 @@ inherits(ColumnCompiler_Oracledb, ColumnCompiler_Oracle);
...
+      ({ useTz } = withoutTz);
+    } else {
+      useTz = !withoutTz;
+    }
+    return useTz ? 'timestamp with local time zone' : 'timestamp';
   },
 
-  timestamp: function(without) {
-    return without ? 'timestamp' : 'timestamp with local time zone';
+  timestamp: function(withoutTz) {
```
elhigu: This will need to wait until oracle tests are back online + testcase

test/unit/schema/oracledb.js
```diff
@@ -669,6 +669,16 @@ describe('OracleDb SchemaBuilder', function() {
     expect(tableSql[0].sql).to.equal(
       'alter table "users" add "foo" timestamp with local time zone'
     );
```
elhigu: Cool, there was a test case 👍

package.json
lorefnon: Yes you are right. Updated this to use double quotes - which prevent globbing in sh. Single quotes are problematic in cmd.

src/dialects/oracledb/schema/columncompiler.js
lorefnon: Given that the outcome is identical I don't see why it will need to wait, but its up to you.

lorefnon:
Thanks for taking a look so soon.

test/unit/schema/oracledb.js
kibertoad: Considering that there is a unit test for this, and output is same, do we really need to wait before merging?

elhigu: Yep, I don't believe that test work unless I can see that it has been ran successfully and I don't wan't to check that manually right now.

kibertoad:
@lorefnon Thanks! Great work as usual.

PR #3183, opened 2019-05-06
Treat offset and limit input the same way

AndreasClausen:
Currently, if you give offset a string, it will throw an error like this:
```
code: 'ER_PARSE_ERROR',
  errno: 1064,
  sqlMessage:
   'You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near \'\'0\'\' at line 1',
  sqlState: '42000',
  index: 0,
  sql:
   'select * from `Alert` inner join `AlertType` on `Alert`.`AlertTypeID` = `AlertType`.`AlertTypeID` order by `AlertDate` desc limit 2 offset \'0\'' }
```
This is problematic, seeing how offset is likely used in pagination using a query in the url, which is usually treated as a string. It's weird to me that it's not like this already since `limit`, a relatively similar method, is implemented this way.

### DIFF
--- a/src/query/builder.js
+++ b/src/query/builder.js
@@ -851,7 +851,12 @@ assign(Builder.prototype, {
 
   // Only allow a single "offset" to be set for the current query.
   offset(value) {
-    this._single.offset = value;
+    const val = parseInt(value, 10);
+    if (isNaN(val)) {
+      this.client.logger.warn('A valid integer must be provided to offset');
+    } else {
+      this._single.offset = val;
+    }
     return this;
   },
 


AndreasClausen:
Alright, I'll take a look at tests later. Haven't really looked through the source code at all, I just went straight for this.

On another note, I found this: https://github.com/tgriesser/knex/pull/2908
What happened here? Why did this not go all the way?

elhigu:
> Alright, I'll take a look at tests later. Haven't really looked through the source code at all, I just went straight for this.
> 
> On another note, I found this: #2908
> What happened here? Why did this not go all the way?

Good point. Actually this PR has the same problem that .offset() should allow passing knex.Raw as input too and maybe even query builder (not sure if SQL spec allows subqueries in offset...)

AndreasClausen:
Yeah, probably it would be easier to reopen/duplicate that one and take it from there.

HurSungYun:
@AndreasClausen FYI #2908 has been merged.

AndreasClausen:
Right, thank you

PR #3187, opened 2019-05-11
WIP: Feature/add rollback all to cli

leeallen337:
This pull request adds support for rolling back all completed migrations to the CLI.

This would, in part, address the feature requests in the following issues: https://github.com/tgriesser/knex/issues/783 and https://github.com/tgriesser/knex/issues/2907

I've labeled this still a work in progress because I'm having difficulty getting the test to work the way I want. Perhaps that's me just not understanding how the testing is set up but here's what I'm trying to accomplish:
1. Create migration 001
1. Run `migration:latest`
1. Create migration 002
1. Run `migration:latest`
1. Create migration 003
1. Run the cli for `migrate:rollback --all` to ensure that migrations 002 and 001 were run in that order and it did not run 003.

The problem is that it only runs the first migration (i.e. migration 001) and no additional migrations.

The reason behind having to create and run them incrementally like that is because running `migrate:latest` would just create one batch and rolling back all wouldn't really be testing it correctly since we want to make sure it rolls back all batches. I've left some commented out code (e.g. console.logs) in there in case it might help troubleshoot.

@kibertoad any instruction you might be able to provide around the testing would be appreciated!

### DIFF
--- a/test/jake/jakelib/migrate-test.js
+++ b/test/jake/jakelib/migrate-test.js
@@ -151,6 +151,123 @@ test('migrate:rollback prints verbose logs', (temp) => {
   });
 });
 
+test('migrate:rollback --all rolls back all completed migrations', (temp) => {
+  return (
+    new Promise((resolve, reject) => {
+      fs.writeFile(
+        `${temp}/migrations/000_create_users_table.js`,
+        `
+        exports.up = (knex) => knex.schema
+          .createTable('users', (table) => {
+            table.integer('id');
+          });
+
+        exports.down = (knex) => knex.schema.dropTable('users');
+      `,
+        (err) => (err ? reject(err) : resolve())
+      );
+    })
+      .then(() => {
+        return assertExec(
+          `${KNEX} migrate:latest \
+      --client=sqlite3 \
+      --connection=${temp}/db \
+      --migrations-directory=${temp}/migrations`,
+          'create_users_table'
+        );
+      })
+
+      // .then(() => {
+      //   console.log(fs.readdirSync(`${temp}/migrations`));
+
+      //   return new sqlite3.Database(temp + '/db')
+      // })
+      // .then((db) => {
+      //   return new Promise((resolve, reject) => {
+      //     db.get('SELECT * FROM knex_migrations', function(err, row) {
+      //       console.log(err);
+      //       console.log(row)
+
+      //       return err ? reject(err) : resolve(row);
+      //     })
+      //   });
+      // })
+
+      .then(() => {
+        return new Promise((resolve, reject) => {
+          fs.writeFile(
+            `${temp}/migrations/001_add_columns_to_user_table.js`,
+            `
+          exports.up = (knex) => knex.schema
+            .table('users', (table) => {
+              table.string('name');
+            });
+
+          exports.down = (knex) => knex.schema
+            .table('users', (table) => {
+              table.dropColumn('name');
+            });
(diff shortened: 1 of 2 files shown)


kibertoad:
@leeallen337 Well yeah, it obviously runs all migrations that it can find. Probably the best approach to address that would be creating migrations programmatically and then deleting them after test. Then you could incrementally add up new migrations.

test/jake/jakelib/migrate-test.js
```diff
@@ -151,6 +151,123 @@ test('migrate:rollback prints verbose logs', (temp) => {
   });
 });
 
+test('migrate:rollback --all rolls back all completed migrations', (temp) => {
+  return (
+    new Promise((resolve, reject) => {
+      fs.writeFile(
```
kibertoad: I would recommend using writeFileSync for simplicity, we don't care about performance in tests much.

test/jake/jakelib/migrate-test.js
```diff
@@ -151,6 +151,123 @@ test('migrate:rollback prints verbose logs', (temp) => {
...
+      //       console.log(row)
+
+      //       return err ? reject(err) : resolve(row);
+      //     })
+      //   });
+      // })
+
+      .then(() => {
+        return new Promise((resolve, reject) => {
+          fs.writeFile(
```
kibertoad: if after using writeFileSync and ensuring that files indeed exist at the moment you are running migrations, you still are only running the very first migration, I would recommend checking if knexfile.js has correct path to migrations directory, could be that it's looking in a different place.

leeallen337: Maybe I'm way off base here but why would it only run the very first migration even if there are more migrations in that directory? Shouldn't `migration:latest` run all migrations that haven't been run yet?

kibertoad: It should run all of them, which is why I'm slightly suspicious about async write not completing when expected or knex reading from wrong directory.

leeallen337: That makes sense. I'll investigate

kibertoad:
@leeallen337 I'm planning to cut a new version on this weekend. Should I wait for this PR to be ready first, or you don't expect it completed until next week?

leeallen337:
@kibertoad Don't wait up for this PR. I don't want to hold up a new version release. I don't know if I'll be able to get it completed until next week

kibertoad:
@leeallen337 OK! I guess we could do a follow-up one later on.

leeallen337:
@kibertoad I actually think I fixed the tests. It was a mix of me thinking that I needed to use the [`sqlite3 get`](https://github.com/mapbox/node-sqlite3/wiki/API#databasegetsql-param--callback) rather than [`sqlite3 all`](https://github.com/mapbox/node-sqlite3/wiki/API#databaseallsql-param--callback) so I _think_ all the migrations were actually running correctly. However, when I was querying the database I wasn't getting all the rows since I was using `sqlite3#get` so I thought not all of them were running. However, I think my `rollback --all` at the end was initially incorrect too so I fixed that as well.

Let me know what you think of the changes. I need to step away to get some food but if you want me to make any changes I should be able to make them tonight or tomorrow

kibertoad:
@leeallen337 Thank you, looks great!
One thing left - could you please create a PR for https://github.com/knex/documentation to update documentation for CLI rollback command?

PR #3189, opened 2019-05-12
Fix backward-incompatibility in type definitions

lorefnon:
Changes in https://github.com/tgriesser/knex/pull/3168 introduces  backward incompatibility causing some other type definitions on DefinitivelyTyped to break when used with latest Knex. 

This PR (tested with all the external type definitions in DefinitivelyTyped having knex as dependency) addresses them by: 

- Adding defaults for parameters of all exposed types which are now generic
- Allowing query builders to be used as select arguments (to accomodate as usage).

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -27,10 +27,8 @@ type Value =
   | Buffer
   | Knex.Raw;
 
-type AnyQueryBuilder = Knex.QueryBuilder<any>;
-
 interface ValueMap {
-  [key: string]: Value | AnyQueryBuilder;
+  [key: string]: Value | Knex.QueryBuilder;
 }
 
 type ColumnDescriptor<TRecord, TResult> =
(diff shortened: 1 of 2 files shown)


kibertoad:
Shoud dtslint tests be updated?

lorefnon:
@kibertoad I have added specs for `as` usage, which was usage. I am not sure if something needs to be added for the default type args. 

Do you have any suggestions here ?

kibertoad:
That should suffice! CI is failing now, though.

kibertoad:
@lorefnon 
```
ERROR: 188:3   expect  Expected type to be:
  Pick<Partial<any>, string>[]
got:
  any
ERROR: 188:30  expect  TypeScript@next compile error: 
Argument of type 'QueryBuilder<any, DeferredKeySelection<any, "bar">[]>' is not assignable to parameter of type 'ColumnDescriptor<any, unknown[]>[]'.
  Type 'QueryBuilder<any, DeferredKeySelection<any, "bar">[]>' is missing the following properties from type 'ColumnDescriptor<any, unknown[]>[]': length, pop, push, concat, and 21 more.
```

Looks like problem is with new test.

lorefnon:
Yes, typescript master has some changes around type inference which caused it to fail with typescript@next. It is fixed now.

kibertoad:
Thanks!

kibertoad:
@lorefnon 0.16.7 is out. Would appreciate if you could confirm if it resolves broken backwards compatibility!

lorefnon:
Thanks for releasing. I have tried this in a sample application and verified that with both `typescript@latest` and `typescript@next` the errors are resolved now.

PR #3191, opened 2019-05-12
Use cli-testlab for testing CLI

kibertoad:
Here is a bit of mad science that I was busy with lately: a better, more seamless and way less verbose way to test CLI (still using Jake, but in a more embedded way).
This is to serve as a PoC of sorts. Should this be merged, I'll migrate all remaining cli tests to cli-testlab and remove knex internal CLI testing microframework


test/cli/knexfile-test.spec.js
```diff
@@ -0,0 +1,82 @@
...
+const KNEX = path.normalize(__dirname + '/../../bin/cli.js');
+
+describe('knexfile resolution', () => {
+  /**
+   * @type FileTestHelper
+   */
+  let fileHelper;
+  beforeEach(() => {
+    fileHelper = new FileTestHelper(path.resolve(__dirname, '../jake-util'));
+    fileHelper.deleteFile('test.sqlite3');
```
kibertoad: this is to prevent preexisting file from messing up tests. could be moved to before block, but that would require some juggling around fileHelper instantiation

elhigu:
Would be nice to include also tape / jake tests to be run with coverage to have real coverage numbers

kibertoad:
@elhigu Will look into tape one, but with Jake don't think it's that simple, as it is executed in a separate process still.

PR #3195, opened 2019-05-14
[WIP] Resolve type incompatibilities wrt. current usage patterns

lorefnon:
- [X] Ensure that parameterized types are always assignable to unparameterized types.  (This not being the case is the primary reason https://github.com/tgriesser/knex/pull/3189 was not adequate)
- [X] Address usage of object specific type operators (eg. `Partial`) that don't work well with `any` or `unknown`. (eg. `unknown` is assignable to `any` but not to `Partial<any>`)
- [ ] Update tests to safeguard against variance errors in all functions that accept callbacks.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -8,9 +8,9 @@
 
 /// <reference types="node" />
 
-import events = require('events');
-import stream = require('stream');
-import Bluebird = require('bluebird');
+import * as events from 'events';
+import * as stream from 'stream';
+import * as Bluebird from 'bluebird';
 
 type Callback = Function;
 type Client = Function;
@@ -27,9 +27,11 @@ type Value =
   | Buffer
   | Knex.Raw;
 
-interface ValueMap {
-  [key: string]: Value | Knex.QueryBuilder;
+interface Dict<T = any> {
+  [k: string]: T;
 }
+
+interface ValueDict extends Dict<Value | Knex.QueryBuilder> {}
 
 type ColumnDescriptor<TRecord, TResult> =
   | string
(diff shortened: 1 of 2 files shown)


kibertoad:
Do you think it might make sense to merge current PR as-is and release as next-2 so that people could play with it (and add tests later), or you still plan to do some non-test changes in it?

lorefnon:
Yeah, I think that should be ok. I'll most likely be able to wrap on the tests by monday.

kibertoad:
CI is failing; could you take a look?

kibertoad:
BTW, is it possible to somehow limit list of TS versions we are testing against? "ts-next" feels like volatile target that may break often.

lorefnon:
> BTW, is it possible to somehow limit list of TS versions we are testing against? "ts-next" feels like volatile target that may break often.

dtslint currently doesn't allow configuring this currently, although this has been [requested](https://github.com/microsoft/dtslint/issues/159).

I have found ts-next sometimes helpful in that it makes migration smoother esp. when strict mode becomes stricter over versions. But I agree this can make test cases flaky. 

For now we might adopt a strategy similar to https://github.com/cypress-io/cypress/issues/3527. Would that be ok ?

kibertoad:
Yeah, sounds fine. How difficult would that be?

lorefnon:
It has been a few years since I last configured travis. I'll check on this and get back in some time.

kibertoad:
It's not urgent by any means :). My only concern is that next time when tests randomly break, there might not be a TS expert of your caliber around to quickly fix them ;)

lorefnon:
lol, not sure about the expert part, but I intend to stick around for a while. I'll create a separate pr for the travis related change.

Issue #666, opened 2015-02-06
It would be great to be able to run an individual migration up/down


simonexmachina:
When I'm developing migrations I often want to just run an individual migration. We have a script that resets the database and runs the migrations, but then if I rollback it rolls back all of the migrations, but I want it to just rollback the last migration so that I can run it again (with changes).

It would be good to be able to run `knex migrate:rollback [migration name]`.

chrisbroome:
:+1: Agreed.

I was definitely surprised by the default behavior of the `knex migrate:rollback` command. The batch behavior is not obvious at all. What's the use case for it?

This is the behavior that I was expecting from rollbacks:

* `knex migrate:rollback` without any arguments should roll back only the last migration that was run
* `knex migrate:rollback [migration name]` should roll back everything up to and including the specified migration

And batch support is still desired, I think there should be an explicit command for it:

* `knex migrate:rollback:batch` should roll back the latest batch of migrations &mdash; in other words, it should do exactly what `knex migrate:rollback` currently does
* `knex migrate:rollback:batch [batch number]` should roll bak to the specified batch number

sheerun:
You can find this feature in [knex-migrate](https://github.com/sheerun/knex-migrate) under `--only` flag.

mrdulin:
Need this feature to rollback db to a specific migration version.

The command maybe like: `knex migrate:rollback --version 20190330113112`

sheerun:
@mrdulin It's `knex-migrate down --to 20190330113112 `

ivan-kleshnin:
What's the difference between `rollback` and `down`? Just another API for the same stuff?
Docs are unclear about that.

sheerun:
rollback rollbacks previous action. down migrates down

Issue #352, opened 2014-07-02
Migrate to version


NathanRSmith:
Hi, I was wondering if there is interest in implementing a "migrate:to" method that migrates up/down to the specified version?  Rollback and latest are nice, it only undoes the latest batch of migrations.  Sometimes it is useful to go a specific version to fix a bad migration script (discovered after initial development), verify specific migration behavior, collapse multiple migrations for clarity before deploying to QA & Prod, etc.  Coming from the Django/South world, this is a feature I have appreciated on occasion.

Lacking this feature, is there a recommended way to call the up/down of a specific migration?

sheerun:
At [knex-migrate](https://github.com/sheerun/knex-migrate) I wrote, you can use `knex-migrate up <version>` and `knex-migrate down <version>` to migrate to specific version.

mercmobily:
Closing it in favour of #6232 which has a little more detail

PR #3205, opened 2019-05-16
Feature/add up for migrations

leeallen337:
This pull request attempts to add a new feature requested in https://github.com/tgriesser/knex/issues/666 and briefly in https://github.com/tgriesser/knex/issues/352

The feature is adding a `migrate:up` functionality which would only migrate the next migration that has not been run yet to the Migrator API and the CLI

Documentation PR: https://github.com/knex/documentation/pull/194

Some other features that have been requested are a `migrate:down` and also being able to migrate up and migrate down to a specific version noted https://github.com/tgriesser/knex/issues/1550, https://github.com/tgriesser/knex/issues/352. I've started on these but don't want to put too much into one pull request

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -205,6 +205,28 @@ function invoke(env) {
     });
 
   commander
+    .command('migrate:up')
+    .description('        Run the next migration that has not yet been run.')
+    .action(() => {
+      pending = initKnex(env, commander.opts())
+        .migrate.up()
+        .spread((batchNo, log) => {
+          if (log.length === 0) {
+            success(color.cyan('Already up to date'));
+          }
+
+          success(
+            color.green(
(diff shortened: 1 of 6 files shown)


elhigu:
tests seem to be failing because node 12 doesn't run oracledb tests and it still checks coverage... I'll fix travis to run coverage only on oldest node version

kibertoad:
@elhigu Oldest version is node 6 and that skips some tests. I would suggest latest LTS instead.

elhigu:
I meant node oldes LTS so node 8, we cant know if latest will have always all drivers ready.

elhigu:
https://github.com/tgriesser/knex/pull/3207

kibertoad:
Thanks!

PR #3207, opened 2019-05-17
Set non coverage tests to be default and set travis to run coverage with only node 8

elhigu:
Different node versions gave different coverages, because skipped tests etc.


package.json
```diff
@@ -91,12 +91,11 @@
     "dev": "rimraf ./lib && babel -w src --out-dir lib --copy-files",
     "lint": "eslint '{src,test}/**/*.js'",
     "lint:types": "dtslint types",
-    "plaintest": "mocha --exit -t 10000 test/index.js && npm run test:tape && npm run test:cli",
-    "plaintest:sqlite": "cross-env DB=sqlite3 npm run plaintest",
```
kibertoad: this one was handy for quick local smoketest, can we keep it?

elhigu: I though it would be as easy to just write `DB=sqlite3 npm test` I do that all the time.

elhigu: but I really don't care if it is in there or not, so I can leave it there as test:sqlite

kibertoad: It's less easy on Windows :D
Appreciate it!

elhigu: doesn't those tests work in git bash? I actually work nowadays mostly on windows, but I haven't done anything for knex on it...

elhigu:
I need to go now and I come back maybe on sunday... @kibertoad if tests works now, could you merge this afterwards at some point?

PR #3212, opened 2019-05-21
typings: Make QueryInterface extend PromiseLike<TResult>

tylerjwatson:
Good afternoon Knex community,

Firstly, I am loving the in-built type declarations for Knex which are a lot more succinct than what was traditionally available in `@types/knex`.  

Since instances of QueryInterfaces in knex are monkey-patched promises, this patch introduces support for the symbols in `PromiseLike<T>` from TypeScript's built-in libs into `Knex.QueryInterface<T, TResult>`. 

It brings `.then`, `.catch` and friends into the scope of `QueryInterface` allowing for the type of `TResult` to flow through to promise chains, including support for `await` syntaxes.

Consider the following basic use-case:

image

Here, the result of `currentLinkRows` is of type Knex.QueryBuilder<T, TResult>.

After the patch:

image

Cheers,
T.

### DIFF
--- a/lib/formatter.js
+++ b/lib/formatter.js
@@ -171,8 +171,8 @@ class Formatter {
 
   // Puts the appropriate wrapper around a value depending on the database
   // engine, unless it's a knex.raw value, in which case it's left alone.
-  wrap(value) {
-    const raw = this.unwrapRaw(value);
+  wrap(value, isParameter) {
+    const raw = this.unwrapRaw(value, isParameter);
     if (raw) return raw;
     switch (typeof value) {
       case 'function':
--- a/lib/query/compiler.js
+++ b/lib/query/compiler.js
@@ -865,7 +865,10 @@ Object.defineProperty(QueryCompiler.prototype, 'tableName', {
 
       if (tableName && schemaName) tableName = `${schemaName}.${tableName}`;
 
-      this._tableName = tableName ? this.formatter.wrap(tableName) : '';
+      this._tableName = tableName
+        ? // Wrap subQuery with parenthesis, #3485
+          this.formatter.wrap(tableName, tableName instanceof QueryBuilder)
+        : '';
     }
     return this._tableName;
   },
--- a/test/integration/builder/selects.js
+++ b/test/integration/builder/selects.js
@@ -1380,5 +1380,38 @@ module.exports = function(knex) {
         }
       }
     });
+
+    it('select from subquery', async function() {
+      const subquery = knex.from('accounts').whereBetween('id', [3, 5]);
+      return knex
+        .pluck('id')
+        .orderBy('id')
+        .from(subquery)
+        .then(
+          (rows) => {
+            expect(rows).to.deep.equal([3, 4, 5]);
+            expect(knex.client.driverName).to.oneOf(['sqlite3', 'oracledb']);
+          },
+          (e) => {
+            switch (knex.client.driverName) {
+              case 'mysql':
+              case 'mysql2':
+                expect(e.errno).to.equal(1248);
+                break;
+              case 'pg':
+                expect(e.message).to.contain('must have an alias');
+                break;
+
+              case 'mssql':
+                expect(e.message).to.contain(
+                  "Incorrect syntax near the keyword 'order'"
+                );
+                break;
+              default:
+                throw e;
+            }
+          }
+        );
+    });
   });
 };
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -8179,6 +8179,37 @@ describe('QueryBuilder', () => {
     );
   });
 
+  it('should always wrap subquery with parenthesis', () => {
+    const subquery = qb().select(raw('?', ['inner raw select']), 'bar');
+    testsql(
+      qb()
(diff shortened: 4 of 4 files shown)


kibertoad:
Thank you for your contribution! We have tests for typings now, any chance you could expand them for this chase

lorefnon:
@tylerjwatson Thanks for your interest in improving the typings. Can you please provide an isolated example (in text format) of typescript usage which you are trying to fix along with your tsconfig.json ? 

Basically, where returns a  QueryBuilder, which already extends ChainableInterface, which extends Bluebird's [Promise](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bluebird/v2/index.d.ts#L370) which is a PromiseLike, so I am curious why this is necessary.

tylerjwatson:
> Basically, where returns a QueryBuilder, which already extends ChainableInterface, which extends Bluebird's Promise which is a PromiseLike, so I am curious why this is necessary.

You are dead right! My apologies, I didn't realise that until later.  In my project we use a lot of function composition to compose queries together based on a set of structured business logic:

```ts
type BuilderFunc = (query: knex.QueryInterface) => knex.QueryInterface
```

There are hundreds of wrapper curries that make up the bulk of our query generation system.  An excerpt of such function is as follows:

```ts
function whereDirectlyLinkedTo(
      list: List,
      directlyLinkedTo: ObjectRef,
      alias: string = TABLE_IDENTIFIER,
    ): BuilderFunc {
      return query => {
        let hasFields = false;
        const Principals = PrincipalList(site);

        query = query.andWhere(c => {
          let generator = allListLookups(site, list);

          if (directlyLinkedTo.list === Principals.ListName) {
            generator = allUserLookups(site, list);
          }

          // const superLists = allSuperLists(site, list, true);

          generator = wu(generator).filter(
            i =>
              isUserLookup(i) ||
              (isListLookup(i) &&
                i.Settings.LookupList === directlyLinkedTo.list),
            // wu(superLists).some(x => x.ListName === i.Settings.LookupList)),
          );

          for (const field of generator) {
            hasFields = true;

            c.orWhere(
              `${alias}.${field.ColumnName}`,
              directlyLinkedTo.id as number,
            );
          }

          /* NOTE: If there are no lookup fields on the object, where false is
             required here.  Otherwise this function returns no where clause, and
             returns every row in the table. */
          if (!hasFields) {
            return c.where(false);
          }
        });

        return query;
      };
    },
```

These provide a powerful mechanism for business logic writers to use public APIs that compose entire queries:

```ts
export function queryList(
  knex: Knex,
  site: Site,
  list: List,
  dataSource: DataSource,
  options: Partial<QueryListOptions> = {},
): BuilderFunc {
  options = _.defaults<Partial<QueryListOptions>, Partial<QueryListOptions>>(
    options,
    {
      includeHistory: false,
    },
  );

  return query => {
    query = fromDataSource(knex, site, list, dataSource)(query).select('T.*');

    if (options.linkedTo) {
      const linkedTo = options.linkedTo;
      const linkedToList = getListByName(site, linkedTo.list);
      const UniversalLinks = UniversalLinkList(site);

      if (!canUniversallyLink(list, linkedToList)) {
        query = UniversalLinks.whereDirectlyLinkedTo(list, linkedTo)(query);
      } else {
        query = UniversalLinks.whereUniversallyLinkedTo(knex, list, linkedTo)(
          query,
        ).union(u =>
          UniversalLinks.whereDirectlyLinkedTo(list, linkedTo)(
            fromDataSource(knex, site, list, dataSource)(u),
          ).select('T.*'),
        );
      }
    }

    return query;
  };
}
```

Now my problem with the new Knex typings stems from the contravariance in PromiseLike<> in this particular implementation.  As QueryInterface<> is the most generic interface (ie. all Builder interfaces extend from this interface), the function composition type _BuilderFunc_ above really needs to accept and produce `QueryInterface` curries, which are currently not `PromiseLike<>`able, which means I need to forcibly cast the results of these curries to Knex.QueryBuilder or deal with loose typings.

Conceptually, as all `QueryInterface` instances produced by Knex are indeed promisable, it would _really_ help me out if the PromiseLike<> extension was placed on Knex.QueryInterface itself, that way the Promise interface can disregard the covariance of the different builder methods altogether and it will work unanimously.

You're dead right about Bluebird<T>, I'll fix that asap.

lorefnon:
I think your BuilderFunc should map QueryBuilder to QueryBuilder and not QueryInterface to QueryInterface . The only other entity that implements QueryInterface is the Knex interface which is not a Thenable  where as your change will mark it as such.

tylerjwatson:
The incorrectness and inflexibility of the typings for this project has forced us to remove it completely in our downstream fork.   It doesn’t make sense to argue for months over use-cases that the untyped JavaScript clearly supports — I’ll do you the honours.

kibertoad:
@tylerjwatson I'm sorry it ended up being not a good fit for your project. From my understanding what happened is that one of our core maintainers (who is the TS lead for the project) suggested a way forward for the use-case described, and we never heard back, so it was unclear whether or not solution was satisfactory. Judging by your reaction it was not. If you could respond with more details regarding what was missing or wrong from your experience, that would be beneficial for whole community.

PR #3214, opened 2019-05-21
Allow users to specify the migrations "tableName" parameter via the CLI.

zbmott:
I am using Knex to migrate several projects from MongoDB to PostgreSQL. Since these projects already have a means of storing connection configuration, the `knexfile` is redundant. Instead, I can read my existing configuration and generate the appropriate Knex migration command dynamically. 

However, because my projects share the same PostgreSQL database, I need to namespace their migration histories so that they don't conflict. In order to do that, I need to be able to specify the `tableName` configuration option via the command line, which this PR introduces support for.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -124,6 +124,10 @@ function invoke(env) {
       'Set migrations directory without a knexfile.'
     )
     .option(
+      '--migrations-table-name [path]',
+      'Set migrations table name without a knexfile.'
+    )
+    .option(
       '--env [name]',
       'environment, default: process.env.NODE_ENV || development'
     );
--- a/bin/utils/cli-config-utils.js
+++ b/bin/utils/cli-config-utils.js
@@ -1,4 +1,4 @@
-const { DEFAULT_EXT } = require('./constants');
+const { DEFAULT_EXT, DEFAULT_TABLE_NAME } = require('./constants');
 const { resolveClientNameWithAliases } = require('../../lib/helpers');
 const fs = require('fs');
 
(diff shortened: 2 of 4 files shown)


kibertoad:
Thank you for your contribution! Could you please add test for this case in https://github.com/tgriesser/knex/tree/master/test/cli as well as create a documentation PR in https://github.com/knex/documentation?

zbmott:
I notice that the project doesn't have any unit tests in place for `mkConfigObj`. Since JS isn't my first language, can I ask what approach you'd take to testing this new functionality?

kibertoad:
See https://github.com/tgriesser/knex/tree/master/test/cli
You can make an actual CLI call and then use knex instance to query db to see if correct table exists in db. Using sqlite for testing should be fine.

zbmott:
Added a test for `--migration-table-name` CLI option. Opened PR for documentation update (PR 195 for `knex/documentation`.)

kibertoad:
Awesome, thank you!

zbmott:
Thank you for your help preparing this feature for acceptance. Can I ask what you release schedule looks like? As I said in my initial comment, I built this feature out because it's something I need for a project I'm working on. I'd like to upgrade my version of `knex` and start using this feature as soon as possible.

kibertoad:
I've just released 0.17.0-next5. Enjoy!

kibertoad:
@zbmott Please let me know if something doesn't work as expected :)

zbmott:
Thank you!

PR #3221, opened 2019-05-25
Update Knex.Config types to include log

### LINKED ISSUE #3148: Type definitions missing Config.log
# Environment

Knex version: 0.16.5
Database + version: n/a
OS: n/a

# Bug

1. Explain what kind of behaviour you are getting and how you think it should work

The type definitions here:

    https://github.com/tgriesser/knex/blob/master/types/knex.d.ts

don't define the log configuration options documented here:

    https://knexjs.org/#Installation-log

From what I understand, this should be defined as part of the `Config` interface.

2. Error message

I get a type error like this:

```
Error:(80, 5) TS2345: Argument of type '{ client: string; connection: any; useNullAsDefault: true; log: { warn: (msg: string) => void; error: (msg: string) => void; }; }' is not assignable to parameter of type 'string | Config'.
  Object literal may only specify known properties, and 'log' does not exist in type 'Config'.
```

3. Reduced test code:

```typescript
import * as knex from "knex";

const connection = knex({
    client: "mysql",
    log: {
        error: (msg: string) => {
            console.error(red(msg));
        },
    },
});
```

## Give some API proposal, how the feature should work

I tried to add a workaround but couldn't get it to merge correctly with existing definitions, however might be useful as a starting point:

```typescript
type LogFn = (msg: string) => void;
type DeprecateFn = (method: string, alternative: string) => void;

interface Config {
    log?: Partial<{
        warn: LogFn;
        error: LogFn;
        debug: LogFn;
        deprecate: DeprecateFn;
    }>;
}
```

lorefnon:
Fixes #3148

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1564,6 +1564,7 @@ declare namespace Knex {
     useNullAsDefault?: boolean;
     searchPath?: string | string[];
     asyncStackTraces?: boolean;
+    log?: Logger;
   }
 
   interface ConnectionConfig {
@@ -1694,6 +1695,15 @@ declare namespace Knex {
     numTestsPerRun?: number;
     softIdleTimeoutMillis?: number;
     Promise?: any;
+  }
+
+  type LogFn = (message: string) => void;
+
+  interface Logger {
+    warn?: LogFn;
+    error?: LogFn;
+    debug?: LogFn;
+    deprecate?: (method: string, alternative: string) => void;
   }
 
   interface MigratorConfig {
--- a/types/test.ts
+++ b/types/test.ts
@@ -4,11 +4,20 @@ import * as Knex from 'knex';
 // import Knex from 'knex'
 // when "esModuleInterop": true
 
-const knex = Knex({
+const clientConfig = {
   client: 'sqlite3',
   connection: {
     filename: './mydb.sqlite',
   },
+};
+
+const knex = Knex(clientConfig);
+
+const knex2 = Knex({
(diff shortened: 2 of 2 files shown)


types/index.d.ts
```diff
@@ -1696,6 +1697,15 @@ declare namespace Knex {
     Promise?: any;
   }
 
+  type LogFn = (message: string) => void;
+
+  interface LoggerConfig {
```
kibertoad: Should it be called Config if signature appears to be that of a full-blown logger?

lorefnon: Hmm ... I can simply rename it to Logger ?

lorefnon:
On a somewhat unrelated note, what is the ideal place (in documentation) to add notes about typescript usage ? Can it be a part of the primary documentation site or a separate page somewhere ?

kibertoad:
Good question! We don't have well-defined place yet. What do you think makes most sense? We could use a new section on a website documentation for sure.

lorefnon:
Similar to SQL dialect dropdown, should we have a language dropdown with javascript (default) & typescript ? Selecting typescript in the dropdown shows ts specifics notes alongside the docs for individual functions.

kibertoad:
Great idea!

elhigu:
I think, if there is need to have typescript docs, they can be shown in the same page. Those should be helpful also for people using plain javascript. Actually all dialect specific notes are also in the same page, dialect dropdown only selects type output of generated queries inside documentation.

PR #3228, opened 2019-05-29
Add migrate:down functionality

leeallen337:
This pull request attemps to add a new feature requested in https://github.com/tgriesser/knex/issues/666 and briefly in https://github.com/tgriesser/knex/issues/352

The feature is adding a `migrate:down` functionality which would rollback only the last run migration to the Migrator API and the CLI

Documentation PR: https://github.com/knex/documentation/pull/197


kibertoad:
ThanksĄ

kibertoad:
@leeallen337 Do you use Node 8+? Next major release 0.18 is going to drop Node 6 support, so if you need it on an older node, would make sense to merge this PR to branch 0.17.

leeallen337:
@kibertoad I do use Node 8+. Maybe it should get merged into 0.17? I saw that `migrate:up` made it into 0.16 so it might be beneficial for users still on Node 6 to have `migrate:down` in addition to `migrate:up`?

I went ahead and created a pull request for it since I'm about to leave for work and won't be available until later

https://github.com/tgriesser/knex/pull/3232

PR #3235, opened 2019-05-30
Do not reject promise on transaction rollback

### LINKED ISSUE #3246: Transaction rollback cause unhandled rejection
Example code:

```js
const knex = require('knex')(/* config */);

process.on('unhandledRejection', err => {
    console.log('unhandled!');
    console.log(err);
});

async function test() {
    await trx.rollback();
    console.log('OK');
}

test();
```

Output:

```log
03:15:36.907 info: OK
03:15:36.915 error: Transaction rejected with non-error: undefined
03:15:36.915 info: unhandled!
03:15:36.917 info: Transaction rejected with non-error: undefined
```

I can not define error handler cause `.transaction()` returns `trx.initPromise`.
https://github.com/tgriesser/knex/blob/master/src/util/make-knex.js#L41-L52

My temporary workaround:
```js
const trx = await new Promise(resolve => 
    knex.transaction(trx => 
        resolve(trx)
    ).catch(() => {})
);
```

kibertoad:
fixes #3246


kibertoad:
@elhigu I've updated implementation to have explicit checks that we actually were rollbacking ourselves. What do you think?

elhigu:
Sounds like a good idea! I didn’t see any problems with code either. 👍

I should write out some tests which actually tests what happens in knex  In different dialects when query throws e.g. conflict error inside transaction which is not leaked out of handler and not causing knex side rollback. Currently I know only more or less what pg dialect does in that case.

kibertoad:
Does that need to happen in this pr?

elhigu:
> Does that need to happen in this pr?

no, thats why I said `I should...`, I just have personal interest to know how knex really behaves in those cases with other drivers.

kibertoad:
@elhigu Thought so, but clarified just in case, as you didn't approve PR either :-D

kibertoad:
Ahaha, yeah, mobile version is a mess. Thanks!

Issue #3229, opened 2019-05-29
Incorrect type definitions for single row queries


spacejack:
Knex version: 0.17.0

Example:

```typescript
const result = await knex('foo').first('id')
```

`result` then has the type `any[]` which breaks previously working code. Eg:

```typescript
const {id} = await knex('foo').first('id')
```
Now results in the error: `Property 'id' does not exist on type 'any[]'`

The following also does not work:

```typescript
const result: {id: number} = await knex('foo').first('id')
```

As a typescript user I appreciate continued efforts to improve types but I feel that unless a distinction can be made between single/multi-row selects then result types should revert back to `any`.

kibertoad:
@lorefnon

spacejack:
Wow, I see you've already tackled the issue. Thanks for the quick response!

I should have added that it would be fair to return a type of `T | undefined` for a single row query. But I'm not sure how precise of a type you're implementing. `any` is fine as well.

kibertoad:
@lorefnon Any reason why we might prefer `any` over `T | undefined?`

spacejack:
No, `T | undefined` is preferable. I'm not sure how to get a type for `T` though?

lorefnon:
Yes, you are right. This should include undefined.

kibertoad:
@spacejack Could you please try knex@0.17.1-next and see if it solves your problem in satisfactory way?

spacejack:
@kibertoad thanks for the update. With this version, the example I showed above now returns an 'any' type, so I can apply my own type as before:
```typescript
const result: {id: number} | undefined = await knex('foo').first('id')
```
Also works with 'where' clauses:
```typescript
const result: {id: number} | undefined = await knex('foo').first('id').where({bar})
```
However if I chain a `count()`, `sum()` or other aggregate functions then the result type switches back to an array. Unfortunately I use `first()` a lot when doing aggregates.
```typescript
const result: {count: number} = await knex('foo').first('id').count({count: '*'})
```

kibertoad:
@spacejack 0.17.1 with fixes for this issue is out!

spacejack:
Thank you, it seems more correct now for single-row queries.

It is still a bit awkward to declare the response types for both single and multi-row queries however. These no longer work:
```typescript
const result: {count: number} = await knex('foo').first().count({count: '*'})
const results: {count: number}[] = await knex('foo').count({count: '*'})
```
[...]

PR #3237, opened 2019-05-31
Update type of aggregation results to not be arrays when first has been invoked before

lorefnon:
Fixes: https://github.com/tgriesser/knex/issues/3229#issuecomment-497560577


ravishivt:
There's still an issue with `first('*')` as it still returns `any[]`.  Workaround is to not specify any columns which defaults to `*`, i.e. `first()`.

PR #3242, opened 2019-06-01
Use migration.extension from config when generating migration

crallen:
In the knex documentation, the config setting "migration.extension" is documented as such:

"extension: the file extension used for the generated migration files (default js)"
source: [https://knexjs.org/#Migrations-API](https://knexjs.org/#Migrations-API)

However, this did not actually work as documented. If you were to set this setting to "ts" and then run the "migrate:make" command without the "-x" argument, the migration tool would still generate a "js" migration file.

Upon digging into the CLI code, I discovered that the CLI was not looking for this setting at all. Instead, it would look for an undocumented "ext" setting at the root level of the configuration. Sure enough, setting this in my knexfile (shown below) produced the result I expected.

```js
// knexfile.js - pulls from a knex configuration section in my app
if (process.env.NODE_ENV === 'production') {
  module.exports = require('./dist/config').default.knex;
} else {
  require('ts-node').register();
  const config = require('./src/config').default.knex;
  module.exports = {
    ...config,
    ext: config.migrations.extension
  };
}
```

This pull request contains a change to the CLI that will add the "migration.extension" setting to the places it looks to determine the extension of the generated migration file, just before checking for this "ext" setting.


kibertoad:
@crallen Thank you for your contribution! Could you please add a test for this case?
https://github.com/tgriesser/knex/blob/master/test/cli/knexfile-test.spec.js
is a good example.

kibertoad:
@crallen There's an even better example now: https://github.com/tgriesser/knex/blob/master/test/cli/migrate-make.spec.js
Should be pretty straightforward to expand existing tests to cover your case.

kibertoad:
@crallen This actually doesn't work. Did you test it? Also documentation is not wrong, `extension` param is part of Migration API documentation, not CLI one. There is a gap in CLI documentation, however, that is correct.
Fortunately I'm doing changes in this part of the code anyway, so I'll fix your PR along the way.

kibertoad:
Released in 0.18.0-next1

lookfirst:
Thanks for adding this, not sure why seed generation was excluded? I'd love to generate my seed files with a .ts extension automatically.

PR #3250, opened 2019-06-03
[WIP] Partially replace Bluebird with native Promises.

kibertoad:
Remove deprecated global methods. Remove Bluebird from typings.

Does not remove Bluebird as a dependency completely, only replaces Bluebird with native Promises where it was straightforward to do so without additional code changes. Idea is to grab low-hanging fruits first and then split remaining (non-trivial) work in chunks and do it incrementally (there were some devs in the community who offered their help as well).


kibertoad:
@lorefnon Removal of bluebird from typings causes major problems with our typings. Our genericified types seem to rely heavily on usage of `.map()` and `.reduce()` methods. How would you recommend to proceed? Simply drop part of the tests that cover that part and rely on users doing mapping on their side from `any` returned by knex?

package.json
```diff
@@ -79,7 +78,8 @@
     "coveralls": "nyc report --reporter=text-lcov | coveralls",
     "lint": "eslint \"src/**/*.js\" \"test/**/*.js\"",
     "lint:types": "dtslint types",
-    "pretest": "npm run lint && npm run lint:types",
+    "old:pretest": "npm run lint && npm run lint:types",
```
kibertoad: this is temporary, I just don't want CI to fail due to an already known problem with typings, will be restored after we figure out how to proceed.

lorefnon:
> Our genericified types seem to rely heavily on usage of .map() and .reduce() methods. How would you recommend to proceed? Simply drop part of the tests that cover that part and rely on users doing mapping on their side from any returned by knex?

I just moved the map/reduce etc. outside of await in https://github.com/tgriesser/knex/pull/3251 which should fix this.

test/tape/pool.js
```diff
@@ -1,12 +1,12 @@
 'use strict';
 
 const test = require('tape');
+const Promise = require('bluebird');
```
elhigu: Is this right?

elhigu:
So far did seem like just removing bluebird requires and fixing types. I hope fixing tests and docs will be as straightforward.

elhigu:
maybe we could stop passing Promise to migrations and seeds.

test/tape/pool.js
kibertoad: yeah, removed it at first and then realized it makes tests fail. should revert change altogether

package.json
```diff
@@ -19,6 +18,7 @@
     "liftoff": "3.1.0",
     "lodash": "^4.17.11",
     "mkdirp": "^0.5.1",
+    "oracledb": "^3.1.2",
```
caiquecastro: Is it correct?

kibertoad: nope. good catch! will remove.

kibertoad:
Useful parts of this were merged separately, closing this one.

PR #3277, opened 2019-06-10
Merge 0.17 into master

kibertoad:


kibertoad:
@lorefnon After this is merged, it might make sense to start first merging changes that are relevant to 0.17 into 0.17 branch, and then to merge with master via direct `0.17 -> master` pull requests. Still gonna cause conflicts in some cases (e. g. when touching typings that had their type converted from Bluebird to Promise), but hopefully will simplify usual cases.

dgadelha:
Can we expect the `ts` extension fix to be included in this release or just the next one?

kibertoad:
@dgadelha How urgently do you need it? I can cut a 0.17 release with it now, or including the migration extension fix in a couple of days.

dgadelha:
Not so urgent, no problem if it takes a few days.

PR #3279, opened 2019-06-11
Adds stub type definition

joelgallant:
Is this excluded intentionally? For my projects, I share this type between knexfile and service level.


kibertoad:
Thanks! Could you please also create a PR in https://github.com/knex/documentation to add it to the list of migrator parameters?

kibertoad:
Released in 0.18.0-next1

PR #3287, opened 2019-06-17
refactor: replace usage of Bluebird Promise.tap with native Promise.then

### LINKED ISSUE #3255: [Bluebird] Remove .tap usage for promises
Remove usage of `tap()` method provided by Bluebird promises but not available for native ones.

benrki:
fixes #3255

Alternative solution: implement a version of .tap in a Promise wrapper around the native Promise (or a providable A+ compliant library) instead of Bluebird.js was done with [src/promise.js](https://github.com/tgriesser/knex/blob/3846b7182ce0eba7a43237cc8e498e5391d33102/src/promise.js) previously.

I opted to always return the resolved promise value even in cases where it was not used in case it might be extended later in the promise chain.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -251,12 +251,13 @@ assign(Client.prototype, {
 
     return Object.assign(poolConfig, {
       create: () => {
-        return this.acquireRawConnection().tap((connection) => {
+        return this.acquireRawConnection().then((connection) => {
           connection.__knexUid = uniqueId('__knexUid');
 
           if (poolConfig.afterCreate) {
             return Promise.promisify(poolConfig.afterCreate)(connection);
           }
+          return connection;
         });
       },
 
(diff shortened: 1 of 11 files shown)


kibertoad:
@benrki Personally I would prefer to avoid extending standard Promise functionality as much as possible, since otherwise we would be moving from one custom Promise solution to another, except less tested and polished one, so I think that this is indeed a good approach :)

src/client.js
```diff
@@ -251,12 +251,13 @@ assign(Client.prototype, {
 
     return Object.assign(poolConfig, {
       create: () => {
-        return this.acquireRawConnection().tap((connection) => {
+        return this.acquireRawConnection().then((connection) => {
           connection.__knexUid = uniqueId('__knexUid');
 
           if (poolConfig.afterCreate) {
             return Bluebird.promisify(poolConfig.afterCreate)(connection);
```
kibertoad: Isn't this part going to mess up with return functionality? Probably return call should be replace with await here.

kibertoad:
@benrki Thank you for your contribution! In two cases there were missing awaits for interim promises, but I believe I've got them all now. Let's see if CI passes :D

PR #3292, opened 2019-06-17
Update maintainer-related info

kibertoad:
@wubzz @elhigu @lorefnon Please let me know if you would like not to be mentioned in package.json.
If someone feels that all maintainers, not only recently active ones should be mentioned, also please let me know.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,9 +4,10 @@ Knex version:
 Database + version:
 OS:
 
-Select applicable tempalate from below.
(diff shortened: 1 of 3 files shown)


lorefnon:
Thanks for adding :) 

Also, beautiful blog. You may want to update the footer links pointing to Ghost's fb & twitter.

kibertoad:
@lorefnon No, thank you for all the hard work! (I've pinged Tim re: permissions again, but I assume he is still swamped atm, sorry again for the process taking forever). Haha, good catch, will fix!

PR #3296, opened 2019-06-19
adds stub to seeds type

ksmithut:
Looks like there was a missing type for `config.seeds.stub` used here: https://github.com/tgriesser/knex/blob/master/src/seed/Seeder.js#L102.


kibertoad:
Thanks!

kibertoad:
Released in 0.18.0-next2

PR #3301, opened 2019-06-21
add clickhouse dialect to knex library

chirag-madlani:
"Clickhouse" connection & basic SQL queries are working fine now.
"preWhere" clause and other clickhouse specific clauses are in-progress.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "bluebird": "^3.5.5",
+    "clickhouse": "^1.2.20",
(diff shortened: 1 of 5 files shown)


kibertoad:
Appreciate the effort! However, current project policy is that new dialects should be introduced as external libraries, installed on top of knex. After you publish it, we can link to it from our documentation. Does that make sense :)?

chirag-madlani:
Sure I wasn't aware of it. Let me close this PR

PR #3308, opened 2019-06-26
Fix comments that were modified by a find & replace

citycide:
The find & replace of ES module import syntax to cjs `require`s in #3227 modified some comments.


kibertoad:
Thanks!

PR #3316, opened 2019-06-28
CLI/add stub option

### LINKED ISSUE #3219: [Feature] knex CLI migrate:make add --stub option (implementation inside)

- be able to designate a stub file from the CLI when creating migration files with `knex migrate:make`

```sh
$ knex migrate:make migration-name --stub /path/to/stub-name.stub
```

- path is relative to `knexfile` location

```sh
$ knex migrate:make migration-name --stub stub-name.stub
```

- `stub-name.stub` in `config.migrations.directory`

- allows users to designate a stub file to use (like `config.stub`) but dynamically through the CLI
- allows users to create / share migration stubs for common operations

the implementations below are small changes only in `/bin/cli.js`. didnt see tests to run for the CLI but the implementations can handle these cases:

## neutral

- no `--stub` passed: uses current implementation (looks for `config.stub` or uses knex default for the given extension)
- `--stub` missing arg: commander describes missing arg error

## success: 

- `$ knex migrate:make migration-name ...`:
  - `--stub /path/to/stub`: loads stub at location relative to `knexfile` location
  - `--stub name.stub`: loads stub at `config.migrations.directory`/`name.stub`

```sh
Using environment: development
Attemping to use stub: <path/to/stub> | <config.migrations.directory/name.stub>
Created Migration: <path/to/migration>
```

## failure: 

- `$knex migrate:make migration-name ...`:
  - `--stub /path/to/stub`: does not exist
  - `--stub name.stub`: does not exist

```sh
Using environment: development
Attemping to use stub: <path/to/stub> | <config.migrations.directory/name.stub>
Error: ENOENT: no such file or directory, open '<path/to/stub> | <config.migrations.directory/name.stub>'
```

- `--stub name.stub`: while `config.migrations.directory` is not defined in `knexfile`

```sh
Using environment: development
Failed to load stub <name.stub>
config.migrations.directory must be defined in knexfile
```

## implementations

- two implementation options, in-line or declarative with external function
- both implementations require one (non-conflicting) addition made in `initKnex()`
  - adds a `workingConfig` property to the `invoke()` scoped `env` object:
- changes designated by `// +` comment

```js
function initKnex(env, opts) {
  // clipped section above for brevity
  let config = env.configuration;

  if (!environment && typeof config[defaultEnv] === "object") {
    environment = defaultEnv;
  }

  if (environment) {
    console.log("Using environment:", color.magenta(environment));
    config = config[environment] || config;
  }

  if (!config) {
    console.log(color.red("Warning: unable to read knexfile config"));
    process.exit(1);
  }

  if (argv.debug !== undefined) {
    config.debug = argv.debug;
  }


// + 
  // attach working config after it is populated
  env.workingConfig = config;

  const knex = require(env.modulePath);
  return knex(config);
}
```

## in-line solution

```js
commander
    .command("migrate:make <name>")
    .description("        Create a named migration file.")
    .option(
      `-x [${filetypes.join("|")}]`,
      "Specify the stub extension (default js)",
    )
// +
    .option(
      `--stub [<relative/path/from/knexfile>|<name>]`,
      "Specify the migration stub to use. Stub <name> must be located in config.migrations.directory",
    )
// +
    .action(name => {
      // clipped for brevity
// +
      let stub = argv.stub;
      if (stub) {
        if (!stub.includes("/")) {
          const stubDirectory = env.workingConfig.migrations.directory;
          if (!stubDirectory) {
            console.log(color.red("Failed to load stub"), color.magenta(stub));
            exit("config.migrations.directory in knexfile must be defined");
          }
          stub = path.join(stubDirectory, stub);
        }
        console.log("Attemping to use stub:", color.magenta(stub));
      }
// +
      pending = instance.migrate
        .make(name, { extension: ext, stub }) // + 
        .then(name => {
          success(color.green(`Created Migration: ${name}`));
        })
        .catch(exit);
    });
```

## declarative solution

```js
// +
function extractStub(env, argv) {
  let stub = argv.stub;

  if (!stub) {
    return null;
  }

  if (!stub.includes("/")) {
    const stubDirectory = env.workingConfig.migrations.directory;
    if (!stubDirectory) {
      console.log(color.red("Failed to load stub"), color.magenta(stub));
      exit("config.migrations.directory in knexfile must be defined");
    }

    stub = path.join(stubDirectory, stub);
  }

  console.log("Attemping to use stub:", color.magenta(stub));
  return stub;
}
// +

// clipped for brevity

commander
    .command("migrate:make <name>")
    .description("        Create a named migration file.")
    .option(
      `-x [${filetypes.join("|")}]`,
      "Specify the stub extension (default js)",
    )
// +
    .option(
      `--stub [<relative/path/from/knexfile>|<name>]`,
      "Specify the migration stub to use. Stub <name> must be located in config.migrations.directory",
    )
// +
    .action(name => {
     // clipped for brevity
// + 
      const stub = extractStub(env, argv);
      pending = instance.migrate
        .make(name, { extension: ext, stub }) // +
        .then(name => {
          success(color.green(`Created Migration: ${name}`));
        })
        .catch(exit);
    });
```

the-vampiire:
closes #3219 

adds `--stub` option to `migration:make` CLI command

**usage**
- `--stub <name>` looks for stub file by name in `config.migrations.directory`
- `--stub </file/path>` looks for stub file by path relative to `knexfile.*`

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -13,69 +13,14 @@ const cliPkg = require('../package');
 const {
   mkConfigObj,
   resolveKnexFilePath,
+  resolveEnvironmentConfig,
+  exit,
+  success,
+  checkLocalModule,
+  getMigrationExtension,
+  getStubPath,
 } = require('./utils/cli-config-utils');
 const { DEFAULT_EXT } = require('./utils/constants');
-
-function exit(text) {
-  if (text instanceof Error) {
-    console.error(
-      color.red(`${text.detail ? `${text.detail}\n` : ''}${text.stack}`)
-    );
-  } else {
-    console.error(color.red(text));
-  }
-  process.exit(1);
-}
-
-function success(text) {
-  console.log(text);
-  process.exit(0);
-}
-
-function checkLocalModule(env) {
-  if (!env.modulePath) {
-    console.log(
-      color.red('No local knex install found in:'),
-      color.magenta(tildify(env.cwd))
-    );
-    exit('Try running: npm install knex');
-  }
-}
-
-function getMigrationExtension(env, opts) {
-  const config = resolveEnvironmentConfig(opts, env.configuration);
-
-  let ext = DEFAULT_EXT;
-  if (argv.x) {
-    ext = argv.x;
-  } else if (config.migrations && config.migrations.extension) {
-    ext = config.migrations.extension;
-  } else if (config.ext) {
-    ext = config.ext;
-  }
-  return ext.toLowerCase();
-}
-
-function getStubPath(env, opts) {
-  const config = resolveEnvironmentConfig(opts, env.configuration);
-  const stubDirectory = config.migrations && config.migrations.directory;
-
-  const { stub } = argv;
-  if (!stub) {
-    return null;
-  } else if (stub.includes('/')) {
-    // relative path to stub
-    return stub;
-  }
-
-  // using stub <name> must have config.migrations.directory defined
-  if (!stubDirectory) {
-    console.log(color.red('Failed to load stub'), color.magenta(stub));
-    exit('config.migrations.directory in knexfile must be defined');
-  }
-
-  return path.join(stubDirectory, stub);
-}
 
 function initKnex(env, opts) {
   checkLocalModule(env);
@@ -120,29 +65,7 @@ function initKnex(env, opts) {
 
   const resolvedConfig = resolveEnvironmentConfig(opts, env.configuration);
   const knex = require(env.modulePath);
-  // attach working config after it is populated
-  env.workingConfig = resolvedConfig;
   return knex(resolvedConfig);
-}
-
-function resolveEnvironmentConfig(opts, allConfigs) {
-  const environment = opts.env || process.env.NODE_ENV || 'development';
-  const result = allConfigs[environment] || allConfigs;
-
-  if (allConfigs[environment]) {
-    console.log('Using environment:', color.magenta(environment));
-  }
-
-  if (!result) {
-    console.log(color.red('Warning: unable to read knexfile config'));
-    process.exit(1);
-  }
-
-  if (argv.debug !== undefined) {
-    result.debug = argv.debug;
-  }
-
-  return result;
 }
 
 function invoke(env) {
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -56,6 +56,27 @@ function getMigrationExtension(env, opts) {
   return ext.toLowerCase();
 }
 
+function getStubPath(env, opts) {
+  const config = resolveEnvironmentConfig(opts, env.configuration);
+  const stubDirectory = config.migrations && config.migrations.directory;
+
+  const { stub } = argv;
+  if (!stub) {
+    return null;
+  } else if (stub.includes('/')) {
+    // relative path to stub
+    return stub;
+  }
+
+  // using stub <name> must have config.migrations.directory defined
+  if (!stubDirectory) {
+    console.log(color.red('Failed to load stub'), color.magenta(stub));
+    exit('config.migrations.directory in knexfile must be defined');
+  }
+
+  return path.join(stubDirectory, stub);
+}
+
 function initKnex(env, opts) {
   checkLocalModule(env);
   if (process.cwd() !== env.cwd) {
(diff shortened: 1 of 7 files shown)


test/cli/migrate-make.spec.js
```diff
@@ -1,11 +1,57 @@
...
+          },
+        }
+      };
+      `,
+    { isPathAbsolute: true }
+  );
+};
+
+const getMigrationFileContents = (fileHelper) => {
+  const [migrationFileName] = fs
```
kibertoad: You don't need this. As I've said, there's a new method in cli-testlab 1.7.0 (just bump the version) called `getFileGlobTextContent` specifically for this use-case.

bin/cli.js
```diff
@@ -197,13 +199,30 @@ function invoke(env) {
...
+      `--stub [<relative/path/from/knexfile> | <name>]`,
+      'Specify the migration stub to use. If using <name> the file must be located in config.migrations.directory'
+    )
     .action((name) => {
       const opts = commander.opts();
       opts.client = opts.client || 'sqlite3'; // We don't really care about client when creating migrations
       const instance = initKnex(env, opts);
       const ext = getMigrationExtension(env, opts);
+      let { stub } = argv;
+      if (stub) {
```
kibertoad: I would prefer to encapsulate this logic inside a separate function, akin to `const ext = getMigrationExtension(env, opts);`.
And, TBH, probably both of them could be moved to `bin/utils/cli-config-utils.js`.

bin/cli.js
```diff
@@ -99,6 +99,8 @@ function initKnex(env, opts) {
 
   const resolvedConfig = resolveEnvironmentConfig(opts, env.configuration);
   const knex = require(env.modulePath);
+  // attach working config after it is populated
+  env.workingConfig = resolvedConfig;
```
kibertoad: Do we need yet another config on top of existing one? We can't just pass it where we need it?

test/cli/migrate-make.spec.js
```diff
@@ -213,4 +259,94 @@ development: {
     );
     expect(fileCount).to.equal(1);
   });
+
+  it('Create a new migration using --stub </file/path> relative to the knexfile', async () => {
```
kibertoad: I would extract all of them into a separate `describe()` block; probably need to add one for existing ones as well then.

test/cli/migrate-make.spec.js
```diff
@@ -1,11 +1,57 @@
...
 
+const fs = require('fs');
 const path = require('path');
 const { FileTestHelper, execCommand } = require('cli-testlab');
 const expect = require('chai').expect;
 
 const KNEX = path.normalize(__dirname + '/../../bin/cli.js');
 
+// start: utils for --stub option tests //
+const migrationStubSetup = (fileHelper) => {
```
kibertoad: Could you please use `function migrationStubSetup (fileHelper) {` style instead?

bin/cli.js
the-vampiire: no i dont think so either. we should definitely pass it in. do you know how i can access the `config.migrations.directory` from the knex instance returned on line 204?

```js
      const instance = initKnex(env, opts);
```

i wasnt able to figure that out myself before which resulted in this little surgery 😂 . if i can access it from the instance i can remove this duplication

bin/cli.js
the-vampiire: sure thing. i wrote it both ways in the original issue so ill push it up with the util function from there for you to take a look.

bin/utils/cli-config-utils.js
```diff
@@ -55,7 +59,94 @@ function resolveDefaultKnexfilePath(extension) {
...
+    ext = config.migrations.extension;
+  } else if (config.ext) {
+    ext = config.ext;
+  }
+  return ext.toLowerCase();
+}
+
+function getStubPath(env, opts) {
+  const config = resolveEnvironmentConfig(opts, env.configuration);
+  const stubDirectory = config.migrations && config.migrations.directory;
```
the-vampiire: this could be moved down. doesnt need to be assigned unless func reaches line 133

kibertoad:
Thanks!

kibertoad:
Released in 0.18.3

Issue #3309, opened 2019-06-26
Issues in last changes to transactions


vonagam:
@kibertoad, i see two issues with pull requests #3235 and #3297 which were published in 0.18:

In first one, you check sql equality to `"ROLLBACK"` string, but what about nested transactions which use `"ROLLBACK TO SAVEPOINT ..."` for rollback?

In second one, you did not just expose `executionPromise`, you created it. It means that now everybody should handle this new promise (especially people who use a container function), otherwise you will get unhandled rejections for previously fine working code. So now there are two separate promises that behave the same but are handled separately, i don't think that really what you wanted.

kibertoad:
thanks for raising the issue! I will take a look

fruuf:
having a similar error here, if you throw an exception within the knex.transaction handler an unhandled rejection bubbles up all the way to process.on

elhigu:
@fruuf AFAIK it is normal behavior and has always been like if transaction handler throws an exception, it is leaked... you should handle those exceptions by your handler. However if you make handler to be async function, then it implicitly return a promise, which triggers knex's implicit commit / rollback functionality.

I would suggest always using `knex.transaction(asyc (trx) => {...})` type of handlers.

If you think you are experiencing regression, open new issue with reproduction code to verify the issue.

optii:
I seem to be having similar problems since the 0.18 update with transactions. I will try and open an issue with reproduction code if I get time

vonagam:
Just to be clear, you can find reproduction code for this issue in knex tests. For example [second one for transaction](https://github.com/tgriesser/knex/blob/ae6bb64cbe7787e6ddfe883b3b2a4fc6c27d653e/test/integration/builder/transaction.js#L25), you can see unhandled rejection warning in [travis log for second PR](https://travis-ci.org/tgriesser/knex/jobs/547885316#L2266).

To not get any warnings you need to write code like this with double catch:
```js
await knex.transaction(async (trx) => {
  trx.executionPromise.catch(() => {} /* catch new */);
  throw undefined;
})
.catch(() => {} /* catch old */);
```

Ideally you should add this snippet to knex mocha setup to treat unhandled rejections as tests failure and not just warnings:
```js
process.on('unhandledRejection', (reason) => {
  console.error('Unhandled Rejection:', reason);
  process.exit(1);
});
```

(But i see that you have other unrelated unhandled rejections, so that may wait.)

kibertoad:
Created #3320 for test improvement.

[...]

PR #3319, opened 2019-06-30
Fix remove duplicate transaction promise

vonagam:
Fixes second issue from #3309.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -65,16 +65,17 @@ class Transaction extends EventEmitter {
         const init = client.transacting
           ? this.savepoint(connection)
           : this.begin(connection);
+        const executionPromise = new Bluebird((resolver, rejecter) => {
+          this._resolver = resolver;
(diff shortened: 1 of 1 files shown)


elhigu:
Would it be possible to add test case for the bug it fixes?

vonagam:
As i said in the issue - your tests are already sufficient. You can see it in tests logs.

Look: [logs](https://travis-ci.org/tgriesser/knex/jobs/547885625#L2278) for PR that introduced the problem (there are 260 `UnhandledPromiseRejectionWarning`), [logs](https://travis-ci.org/tgriesser/knex/jobs/552343867#L1844) for this PR (there are none of those).

If you want tests to show rejections more clearly you should force tests to exit with non-zero status on unhandled rejections. How to do it i wrote in the issue. But you need to eliminate other unhandled rejections (there are around ten of them) before you do so.

kibertoad:
Thanks! I'll release a fixed version shortly.

PR #3326, opened 2019-07-03
add mariadb connector

knoxcard:
add mariadb connector


elhigu:
it doesnt work like that.

Issue #2071, opened 2017-05-18
Add modify method for reuse of Schema modification snippets


zachelrath:
The [`.modify()` method on QueryBuilder](http://knexjs.org/#Builder-modify) is a very useful way to re-use common query modification snippets. It would be great to be able to have the same capability for reuse of Schema Builder snippets. Adding `modify()` as a new prototype method on tablebuilder and columnbuilder's prototypes would enable this. I would have the logic be virtually identical to the logic for [QueryBuilder's modify method](https://github.com/tgriesser/knex/blob/master/src/query/builder.js#L912).

elhigu:
Sounds kind of useful, how that would work when schema builder creates multiple queries? Any real world examples how you would be using this?

zachelrath:
Okay, real world examples:

We have a series of common migration commands that we want to reuse throughout our migrations, to maintain consistency in our database structure and minimize the complexity of individual migrations. We store these in utility classes that we reference in our migration files. So for example, we have a module called "migrationUtils.js" that exposes reusable, composable schema modification snippets such as the following:

```
// MigrationUtils.js

const siteIdForeignKey = function(table) {
	table.uuid("site_id").unsigned().notNullable();
	table.foreign("site_id").references("site.id");
	return table;
};

const primaryKey = function(table, knex) {
	return table.uuid("id").primary().defaultTo(knex.raw("uuid_generate_v4()"));
};

const foreignKey = function(sourceTable, targetTableName, columnName, notNullable = true) {
	if (!columnName) columnName = targetTableName + "_id";
	const column = sourceTable.uuid(columnName);
	column.unsigned();
	if (notNullable) column.notNullable();
	sourceTable.foreign(columnName).references(targetTableName + ".id");
	return column;
};

const requiredForeignKey = function(sourceTable, targetTableName, columnName) {
	return sourceTable.modify(foreignKey, targetTableName, columnName, true);
};

const modstamps = function(table) {
	table.uuid("created_by_id").unsigned().notNullable();
	table.uuid("updated_by_id").unsigned().notNullable();
	table.foreign("created_by_id").references("user.id");
	table.foreign("updated_by_id").references("user.id");
	return table;
};

const defaultColumns = function(table, knex) {
	table.modify(primaryKey, knex);
	table.modify(siteIdForeignKey,);
	table.modify(modstamps);
	table.timestamps();
	return table;
};

module.exports = {
	defaultColumns: defaultColumns,
	foreignKey: foreignKey,
[...]

Issue #1579, opened 2016-07-15
Custom Builder Functions


mtom55:
I'd like to add some custom functions to knex which I have to use all the time but i'm finding it hard to get them to work.

They are:
   knex.selectOne
   knex.selectZeroOrOne

   Instead of doing first, selectOne does not apply a limit and instead throws an error if it gets more than one record back.  This allows the author to catch particular errors particular early.

I started with:

```
knex.client.QueryBuilder.prototype.selectOne = function (columns) { 
    const args = new Array(arguments.length);
    for (let i = 0; i < args.length; i++) {
      args[i] = arguments[i];
    }
    this.select.apply(this, args);
    this._method = 'first';
    //this.limit(1);
    return this;
}
```

but obviously the check on how many records get returned has to happen after the results have come back 

Would you be able to point me in the right direction?

tgriesser:
Just FYI This code won't work in 0.12, you'll need to require `knex/lib/query/builder` directly to modify that prototype. 

For 1.0 I'll plan to define a standard way to customize the query builder chain like you're trying to do above, and with the addition of "hooks", you'll be able to easily setup post-result transforms.

thetutlage:
@tgriesser I have waiting for this feature for while and start modifying the `QueryBuilder` prototype. It will be so great to have an interface to extend the query builder.

👍

sidewalksalsa:
@tgriesser - How far are we away from 1.0?

mercmobily:
This has been resolved. Knex now provides `QueryBuilder.extend()` method that allows developers to add custom methods to the query builder.

Issue #1456, opened 2016-05-27
Make knex easier to extend with new operators


mnpenner:
Currently I'm hacking around the API to add new operators:

``` js
import QueryBuilder from 'knex/lib/query/builder';

const origWhere = QueryBuilder.prototype.where;

function escapeLike(term) {
    return term.replace(/[_%]/g, '\\$&');
}

Object.assign(QueryBuilder.prototype, {
    where: function(column, operator, value) {
        if(arguments.length === 3) {
            switch(('' + operator).toLowerCase().trim()) {
                case 'contains':
                    operator = 'like';
                    value = `%${escapeLike(value)}%`;
                    break;
                case 'startswith':
                    operator = 'like';
                    value = `${escapeLike(value)}%`;
                    break;
                case 'endswith':
                    operator = 'like';
                    value = `%${escapeLike(value)}`;
                    break;
            }
        }
        return origWhere.call(this,column,operator,value);
    }
});
```

It would be nice if there were an 'official' way to do this.

Maybe something like:

``` js
knex.addOperator('contains', function(column, value) {
    return this.where(column, 'like', `%${escapeLike(value)}%`);
});
```

Which would allow usage such as:

``` js
knex.from('t').select('*').where('foo','contains','b_r').orWhere('baz','StartsWith','q%x');
```

``` sql
select * from `t` where `foo` like '%b\_r%' or `baz` like 'q\%x%'
```

rhys-vdw:
I'd be more inclined to give the tooling by which a user could implement `addOperator` as a plugin.

There's talk of some re-architecting of the project, so now might not be the best time for considering this change.

mnpenner:
Fair enough. As long as the new plugin architecture is flexible enough to inject changes into the existing `where` methods, that should be fine.

PR #3334, opened 2019-07-06
Feat: Allow to extend knex query builder

### LINKED ISSUE #1158: Extending knex
Is there any way to extend knex ( have not found one ) to add my own methods.

Below will not work because after the first method call knex returns an instance of queryBuilder.

``` javascript
var knex = require('knex')(config)
knex.paginate = function () {}
```

I also have an idea to have an `extend` method, which can be used to add/extend new methods, but before working on it, i want to know are you guys even interested in something like this or not.

felixmosh:
This PR allows to extend query builder with custom functions such as `paginate` or `onDuplicateUpdate` that will allow add custom behaviour or specific dialect methods over the basic query builder functions.

``` javascript
knex('products')
    .where('price', '<', 20)
    .paginate(15, 1, true)
    .then(paginator => {
        console.log(paginator.current_page);
        console.log(paginator.data);
    });
```

```javascript
knex()
  .insert({ id: 1, name: 'foo' })
  .into(customers)
  .onDuplicateUpdate('name');

// which will generate
```
```sql
Insert  into customers (id, name) values (1, 'foo') on duplicate key update name=Values(name)
```
Currently there is no way to properly extend the QueryBuilder without monkey-patching the `QueryBuilder.prototype`.

 closes #1158


elhigu:
This feature is wanted and it has been discussed before. I want to research earlier discussions and review the code, since for this I want knex to have carefully designed api and functionality.

This fundamental features should have design discussions before doing PRs or to be prepared to redo the whole implementation in a different manner.

felixmosh:
Since my implementation is small one i would happy to redo it.
I'm open to discussion 😊

kibertoad:
TBH, this seems a simple and non-intrusive solution, so I would vote for including this implementation.
My only concern is about TypeScript part of the equation - my assumption is that since QueryBuilder is part of Knex namespace, it should be possible to augment its interface locally and have additional methods on it. I wonder if it's possible to add test for this in https://github.com/tgriesser/knex/blob/master/types/test.ts

One additional suggestion: it might make sense to throw a warning when you are overriding existing method using .extend()

kibertoad:
Related issues: 
https://github.com/tgriesser/knex/issues/2071 
https://github.com/tgriesser/knex/issues/1579
https://github.com/tgriesser/knex/issues/1456
https://github.com/tgriesser/knex/issues/1158

kibertoad:
There doesn't seem to be many valuable proposals in prior issues, TBH. `I'd be more inclined to give the tooling by which a user could implement addOperator as a plugin.` from #1456 doesn't seem feasible int he nearby future, and, frankly, sounds like an over-engineering for simple cases where this particular three-lines long solution would do just fine.

felixmosh:
> TBH, this seems a simple and non-intrusive solution, so I would vote for including this implementation.
> My only concern is about TypeScript part of the equation - my assumption is that since QueryBuilder is part of Knex namespace, it should be possible to augment its interface locally and have additional methods on it. I wonder if it's possible to add test for this in https://github.com/tgriesser/knex/blob/master/types/test.ts
> 
> One additional suggestion: it might make sense to throw a warning when you are overriding existing method using .extend()

I can check, TS by design allows to extend interfaces, but there some case it won't work.
I'm already using that at https://github.com/felixmosh/knex-on-duplicate-update/blob/master/types.d.ts
I will add a test for this.

felixmosh:
> There doesn't seem to be many valuable proposals in prior issues, TBH. `I'd be more inclined to give the tooling by which a user could implement addOperator as a plugin.` from #1456 doesn't seem feasible in he nearby future, and, frankly, sounds like an over-engineering for simple cases where this particular three-lines long solution would do just fine.

Agree, it can be a base for more pluggable solution (if needed).

What do you think regarding exposing `QueryBuilder` constructor on `knex`?

test/unit/knex.js
```diff
@@ -468,4 +468,30 @@ describe('knex', () => {
...
+    beforeEach(() => {
+      connection = new sqlite3.Database(':memory:');
+    });
+
+    afterEach(() => {
+      connection.close();
+    });
+
+    it('should extend default queryBuilder', (done) => {
+      Knex.QueryBuilder.extend('customSelect', function(value) {
```
kibertoad: This leaks extended querybuilder outside of thus test, I think. Would be good to delete custom metode in after() block.

felixmosh: So you suggest to add remove method as well? cause with the current api, it always adds the method for the entire app.

kibertoad: wouldn't simple delete `Knex.Querybuilder.prototype.customMethod` work? I don't think we should expose deletion method via API.

felixmosh: It will work, I will update my PR soon

felixmosh:
@kibertoad I'm not sure how to test the type augmentation of the QueryBuilder extension, can you give me some pointers?

test/unit/knex.js
```diff
@@ -468,4 +469,37 @@ describe('knex', () => {
...
+      const knex = Knex({ client: 'sqlite3' });
+      knex
+        .connection(connection)
+        .customSelect(42)
+        .then((result) => {
+          expect(result[0].value).to.equal(42);
+          done();
+        });
+    });
+
```
elhigu: There should be also tests that custom method is available for example in transaction and maybe in some cases one likes to add it to join builder or some nested subquery builder.

felixmosh: Not sure I'm following, the extend is done on the Builder prototype only once in the app.
There is no difference if you are using the custom method with sub-query or any other chain.

I'm missing something?

test/unit/knex.js
kibertoad: Transaction is a different entity, not an original knex builder, but user would be justified to expect custom methods on it as well. It is highly likely we'd need to copy them over, but test would show that. Same with join builder, it's a different thing.

test/unit/knex.js
kibertoad: Transaction is a different entity, not an original knex builder, but user would be justified to expect custom methods on it as well. It is highly likely we'd need to copy them over, but test would show that. Same with join builder, it's a different thing.

felixmosh: So, should I add `extend` to them as well? should it be separate method for each?

kibertoad: No, it would be better if they were inherited automatically. User needs not to concern himself with difference between all three.

felixmosh: Agree, do Transaction & Join Builder are inherites from Builder? If so, the requirement is to check that it extends the prototype, meaning that is available at the child class?

kibertoad: Nope:
src/transaction.js
src/query/joinclause.js (TBH, join builder is a farily different beast, I'm not sure querybuilder methods make much sense there. @elhigu, what use-case did you have in mind?)

kibertoad: I'm not sure what is going to be safest and cleanest solution here, TBH. Maybe QueryBuilder.extend could also extend prototype of Transaction, or overall .extend method could be moved to general Knex.extend that covers both QueryBuilder and Transaction.

felixmosh: What is the use case of extending join & transaction?

kibertoad: You can use transaction pretty much same way you can use knex instance. See example from documentation:
```
// Using trx as a transaction object:

const books = [
  {title: 'Canterbury Tales'},
  {title: 'Moby Dick'},
  {title: 'Hamlet'}
];

trx('catalogues')
  .insert({name: 'Old Books'}, 'id')
  .then(function(ids) {
    books.forEach((book) => book.catalogue_id = ids[0]);
    return trx('books').insert(books);
  })
  .then(trx.commit)
  .catch(trx.rollback);
})
```

Considering that currently transaction supports same query-building methods that knex.QueryBuilder does, as a user I would be very surprised if I'm supposed to set those up separately.

kibertoad: Although looking at the code, I suspect this is going to just work (test to prove that would be most welcome, though). The way transactor is created is this:
```
  const transactor = makeKnex(trxClient);
```
So what gets called is this:
```
function makeKnex(client) {
  // The object we're potentially using to kick off an initial chain.
  function knex(tableName, options) {
    return createQueryBuilder(knex.context, tableName, options);
  }

  redefineProperties(knex, client);
  return knex;
}
```

Unless I'm missing something, this is going to call same QueryBuilder constructor in the end.

felixmosh: Ha, I'm not using transaction in that way ;)
```javascript
knex.transaction((trx) => {
  knex(t1).transacting(trx)...
})
```
If it has the same query methods, it make sense to extend it as well.

kibertoad: `trx` var in this context is the same transactor you would get from `await knex.transaction()`, and see above, I suspect it already works.

felixmosh: Let me add a test for it to assure it, if it not works we will need to think how extend it as well

kibertoad:
@felixmosh Regarding testing: not sure, never tested for this case myself, but from my understanding it is possible to assert literal types for specific variables

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v8/node-tests.ts
e. g. `// $ExpectType <T extends Function>(fn: T, message: string) => T`
Not sure if it's going to work, but worth a try.

elhigu:
Sounds like only the issue #1158 had some really meaningful discussion around the area. Right now I think that the only drawback with this extension method is to be able to nicely override old methods with extended functionality. Being able to call original implementation of rewritten method sounds like something that I would like to have (I've used that quite a lot with objection). 

I think that for now we could add this way that just allows to add stuff to prototype and pass it around to other builders as experimental extension support (to be able to deprecate it fast if it is not sufficient). From that experience we can decide later on if we actually want to make extending to work with real inheritance and setting then query builder instances to knex in a bit same way that it was suggested here https://github.com/tgriesser/knex/issues/1158#issuecomment-176093904 and how it was implemented for objection).

elhigu:
> Nope:
> src/transaction.js
> src/query/joinclause.js (TBH, join builder is a farily different beast, I'm not sure querybuilder methods make much sense there. @elhigu, what use-case did you have in mind?)

For transaction all the same methods should be found that are found from main query builder... so all use cases I suppose :)

For join builder and subquery builder probably doesn't need to have all the extensions... I suppose extender method could have `opts : { queryBuilder: true, joinBuilder: true, subBuilder: true }` object to tell if what we are extending (I don't remember if there really is a separate subquery builder though).

elhigu:
Then another thing which needs testing is to check that instance created with `knex.withUserParams({customUserParam: 'table1'});` works fine with extended functionlality

kibertoad:
@elhigu Given that joinBuilder has methods along the line of 
```
  – onIn
  – onNotIn
  – onNull
  – onNotNull
  – onExists
  – onNotExists
  – onBetween
  – onNotBetween
```
, is that really related to extending QueryBuilder per se? Probably there could be separate method exposed for extending just the JoinBuilder, but I would say that would be a separate PR altogether.

elhigu:
Separate PR sounds fine if that even is ever needed. I agree that it doesn't sound really important.

felixmosh:
Sorry for the confusion, what are the action items? :)

kibertoad:
@felixmosh 

1) Add test for using custom methods from transaction (`const trx = await knex.transaction; trx.customMethod()`)
2) Add test for using custom methods from knex copy with user params (`const knewWithParams = knex.withUserParam({}); knewWithParams .customMethod()`)
3) Add documentation in https://github.com/knex/documentation
4) Try adding a test in `https://github.com/tgriesser/knex/blob/master/types/test.ts` that would assert interface augmentation to work.
5) This one was suggested by @elhigu -> possibility to override existing methods while preserving reference to original methods somehow (I guess those could be reattached to original object using some prefix or container). Not sure if `this` works correctly for custom methods attached to QueryBuilder, needs a try. I would imagine having something like
```
function where() {
console.log('log stuff')
this.original.where(...args) // or this.originalWhere(...args)
}

knex.QueryBuilder.extend('where', where)

**Comment by felixmosh** [comment]
> @felixmosh
> 
> 1. Add test for using custom methods from transaction (`const trx = await knex.transaction; trx.customMethod()`)
> 2. Add test for using custom methods from knex copy with user params (`const knewWithParams = knex.withUserParam({}); knewWithParams .customMethod()`)
> 3. Add documentation in https://github.com/knex/documentation
> 4. Try adding a test in `https://github.com/tgriesser/knex/blob/master/types/test.ts` that would assert interface augmentation to work.
> 5. This one was suggested by @elhigu -> possibility to override existing methods while preserving reference to original methods somehow (I guess those could be reattached to original object using some prefix or container). Not sure if `this` works correctly for custom methods attached to QueryBuilder, needs a try. I would imagine having something like
> 
> ```
> function where() {
> console.log('log stuff')
> this.original.where(...args) // or this.originalWhere(...args)
> }
> 
> knex.QueryBuilder.extend('where', where)
> ```

Thanx, will work on it at the evening.
Regarding point #5, I think that this should be a separate feature, currently it is not allowed to override existing methods.

**Comment by kibertoad** [comment]
@felixmosh Sounds good, as long as you think that could be introduced later in a non-breaking way.

**Comment by elhigu** [comment]
Lets mark this support experimental, so it will be ok to break when we have more user experience about how it should work.

**Comment by felixmosh** [comment]
Agree, how do i mark it as experimental?

**Comment by elhigu** [comment]
Just by mentioning it in documentation

**Comment by kibertoad** [comment]
Unfortunately, this needs to be updated to accomodate for /src -> /lib movement of knex files. Should be fairly easy fix, though.

**Comment by felixmosh** [comment]
No problem, i will fix that at the weekend.

File: types/test.ts [hunk 3]
```diff
@@ -4,6 +4,12 @@ import * as Knex from 'knex';
 // import Knex from 'knex'
 // when "esModuleInterop": true
 
+declare module 'knex' {
+  interface QueryBuilder {
+    onDuplicateUpdate(...columnNames: string[]): QueryBuilder;
```
kibertoad: Are any assertions possible on existence of this method?

kibertoad:
LGTM!
@elhigu @lorefnon What do you think?

types/test.ts
felixmosh: I've improved it & added a usage. 
:]

felixmosh:
A documentation PR https://github.com/knex/documentation/pull/218

lorefnon:
> @lorefnon What do you think?

I am personally not a huge fan of multiple libraries extending shared context, and in general, one import potentially changing the outcome of another import (across files). 

I think this will also cause jump to definition to not work (even in a pure typescript codebase - it will just jump to the declaration which would have to be separately defined).

But given that people see value in this, and also because the current knex codebase is already not very statically explorable (because of not using ES6 classes in some places, extensive aliasing etc. features like jump-to-definition, ide outline etc. often don't work), hence I don't have a very strong opinion here.

kibertoad:
Released in 0.19.1

PR #3336, opened 2019-07-07
Fix mssql import not being ignored due to change in browser field in package.json

lorefnon:
Fixes: https://github.com/knex/documentation/issues/212


lorefnon:
I am assuming that this was result of a quick find/replace in editor. 
If there was a reason for this change I can close this PR and check further.

kibertoad:
Nope, definitely careless mass replace after removal of babel. Thank you!

PR #3337, opened 2019-07-07
fix: Respect the knexfile stub option while generating a migration

### LINKED ISSUE #3333: Custom migration stubs are not used
I'm trying to define my own migration stubs, to avoid having to create the same boilerplate code manually.

This is the stub file (project/migration.stub.js):
```js
//@ts-check

/**
 * @param {import("knex")} knex
 */
exports.up = async knex => {}

/**
 * @param {import("knex")} knex
 */
exports.down = async knex => {}
```
And this is my knexfile (project/knexfile.js):
```js
// Update with your config settings.
require("dotenv/config")
const { DB_HOST, DB_NAME, DB_USER, DB_PASS } = process.env

module.exports = {
  development: {
    client: "sqlite3",
    connection: {
      filename: "./dev.sqlite3"
    },
    migrations: {
      tableName: "knex_migrations",
      stub: "migration.stub.js"
    }
  },

  production: {
    client: "pg",
    connection: {
      host: DB_HOST,
      user: DB_USER,
      password: DB_PASS,
      database: DB_NAME
    },
    migrations: {
      tableName: "knex_migrations"
    }
  }
}
```

However, when I run `knex migrate:make something` it gives me this:
```js
exports.up = function(knex) {
  
};

exports.down = function(knex) {
  
};
```

According to the documentation, this is correctly configured. Am I missing something?
Knex version: 0.18.3

bohdan-shulha:
fixes #3333

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -154,9 +154,15 @@ function invoke(env) {
       opts.client = opts.client || 'sqlite3'; // We don't really care about client when creating migrations
       const instance = initKnex(env, opts);
       const ext = getMigrationExtension(env, opts);
+      const configOverrides = { extension: ext };
+
       const stub = getStubPath(env, opts);
+      if (stub) {
+        configOverrides.stub = stub;
+      }
+
       pending = instance.migrate
-        .make(name, { extension: ext, stub })
+        .make(name, configOverrides)
         .then((name) => {
           success(color.green(`Created Migration: ${name}`));
         })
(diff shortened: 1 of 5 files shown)


kibertoad:
Thanks! Could you please add test in https://github.com/tgriesser/knex/blob/master/test/cli/migrate-make.spec.js for this case?

bohdan-shulha:
@kibertoad , thanks for the tests and the release, I had no time this week to implement test cases

PR #3345, opened 2019-07-10
Update tarn.js

kibertoad:


kibertoad:
@elhigu @koskimas  `.afterCreate()` hook explodes as unsupported now. We don't support and don't want to support that anymore? Listening to creation events is the way to go now?

elhigu:
afterCreate is implemented in knex side and looks like it is passed to pool paameters by accident.

test/unit/seed/seeder.js
```diff
@@ -90,7 +90,7 @@ describe('Seeder._waterfallBatch', function() {
   it('should throw an error with correct file name', (done) => {
     seeder._waterfallBatch(['1-first.js', '2-second.js']).catch((error) => {
       expect(error.message).to.match(
-        /^Error while executing "(\/\w+)+\/1-first\.js" seed: throwing in first file$/
+        /^Error while executing .*1-first.js" seed: throwing in first file$/
```
kibertoad: Original assertion was failing on Windows

lib/client.js
```diff
@@ -297,7 +297,15 @@ Object.assign(Client.prototype, {
       return;
     }
 
-    this.pool = new Pool(this.getPoolSettings(config.pool));
+    const tarnPoolConfig = {
+      ...this.getPoolSettings(config.pool),
+    };
+    // afterCreate is an internal knex param, tarn.js does not support it
+    if (tarnPoolConfig.afterCreate) {
+      delete tarnPoolConfig.afterCreate;
```
alexandrebodin: You will have the same issue with `beforeDestroy` I think.

elhigu: `beforeDestroy` is deprecated and is not really useful hook anyways (it has been deprecated already for a long time). I think we can drop it already.

alexandrebodin: Sure the PR should simply delete it from the code so there is no way it is passed by error then :)

kibertoad: Removed `beforeDestroy` from typings and implementation, and mentioned in migration guide that it's no longer supported. Users should still update their code not to pass it, though, as silently ignoring configuration options is confusing.

elhigu: When people start complaining that their hacky code (*) doesn't work without it we can tell them to use pool's event handlers instead.

*)  code relaying to that hook cannot work in a  robust manner unless it is counting more or less how many connections were discarded from pool... so far all the people who has been using it and with whom I've been talking to have made invalid assumptions when that is called and have tried to do some DB disconnected kind recognitions and knex re-inits there.

kibertoad: Yeah, migration guide already mentions pool event handlers.

PR #3349, opened 2019-07-13
Fix #3341: Update Knex.raw type to be any by default because the actual type is dialect specific

### LINKED ISSUE #3341: Raw return type seems off
# Environment

Knex version: 0.18.3
Database + version: postgres
OS: macOS 10.14.4
VSCode: 1.36.0
TS: 3.5.3

# Bug

When making a raw query, the results returns `any[]`, however the actual object seems to be a results object.

Previously I was returning `results.rows`, however now the bindings now say that the return type of raw(...) is `any[]`, however, printing indicated that it is a results object. I just upgraded from `"knex": "^0.16.3",` and saw this behavior.

It _could_ be related to this change - https://github.com/tgriesser/knex/commit/9c6ab963f6f6af135ef72b6d5fdfb14b6f3707c6#diff-88bec6beae84369c0376b56b3bb88fe1R268

```js
const queryDb = () => {
  const results = await db.query.raw(rawQueryBuilder.join(" "), bindings);
  console.log(results);

  // Use of ts-ignore allows it to work as it did before
  // @ts-ignore
  return results.rows;
}

// Results of log
Result {
  command: 'SELECT',
  rowCount: 8,
  oid: null,
  rows: [...],
  fields: [...],
  parsers: [...],
  _types: TypeOverrides {
    _types: {
      getTypeParser: [Function: getTypeParser],
      setTypeParser: [Function: setTypeParser],
      arrayParser: [Object]
    },
    text: {},
    binary: {}
  },
  RowCtor: null,
  rowAsArray: false
}
```

Screenshot of IDE indicating the return type

Screen Shot 2019-07-10 at 5 32 01 PM

Screen Shot 2019-07-10 at 11 27 38 AM

# Expect

I would expect the return type of a `raw` query to return something like this (or at least

```
interface RawQueryResult<T>{
  command: string;
  rowCount: number;
  rows: T[]; // Or even just rows: any[];
  ...
}

```

lorefnon:


kibertoad:
Thanks!

PR #3350, opened 2019-07-13
[WIP] Add test for enum on mssql

caiquecastro:
Mssql does not have a enum data type. We can use check constraints.


kibertoad:
@caiquecastro Emmm, adjusted test is failing now. Were you planning to adjust mssql implementation as well?

caiquecastro:
Yeah! I'm checking how can i do it. Does the function receive any parameters about the column name?

PR #3354, opened 2019-07-16
Minor enhancements around aliasing of aggregates

### LINKED ISSUE #2871: count with multiple aliased columns
# Environment

Knex version: 0.15.2

It would be nice to be able to use a single `.count()` call with an object to set multiple aliased count columns instead of having to call it multiple times.

It seems most of the work for this was already done in #2449 so it should be very simple to add support for this. I can provide a PR if there's any interest in having this feature.

Currently if one tries to use this notation only the first key is being used, and the rest is just ignored. 

Instead of:

```js
// ...
.count({ something: 'something.id' })
.count({ somethingElse: 'something_else.id' })
// ...
```

we could just:

```js
// ...
.count({
  something: 'something.id',
  somethingElse: 'something_else.id'
})
// ...
```

### LINKED ISSUE #3315: count() returns something like [ { '': 1 }]
# Environment

Knex version: 0.18.0
Database + version: MS SQL 2017
OS: Mac 10.14.5

# Bug
The docs say that `.count()` should return `number | string`, but I'm getting `[ { '': 1 } ]`. My code is simply:

```js
const results = await db(TABLE_NAME).count()
console.log(results) // [ { '': 1 } ]
```

Perhaps this is a MS SQL specific issue?

lorefnon:
- Allows an options object to specify alias (as suggested in https://github.com/tgriesser/knex/issues/3315#issuecomment-507439215) (Closes #3315)
- Allows multiple aliases when using object syntax (Closes #2871)

### DIFF
--- a/lib/dialects/mssql/query/compiler.js
+++ b/lib/dialects/mssql/query/compiler.js
@@ -136,7 +136,7 @@ Object.assign(QueryCompiler_MSSQL.prototype, {
         const stmt = columns[i];
         if (stmt.distinct) distinct = true;
         if (stmt.type === 'aggregate') {
-          sql.push(this.aggregate(stmt));
+          sql.push(...this.aggregate(stmt));
         } else if (stmt.type === 'aggregateRaw') {
           sql.push(this.aggregateRaw(stmt));
         } else if (stmt.value && stmt.value.length > 0) {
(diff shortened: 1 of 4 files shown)


kibertoad:
One test is failing on mssql.

lorefnon:
@kibertoad this has now been addressed.

kibertoad:
I assume this also needsdocumentation update?

elhigu:
Added issue for it.

lorefnon:
Yes, I'll take care of that.

PR #3357, opened 2019-07-16
Seeds now support multiple seed directories, with a sorting option.

candyapplecorn:
### This pull request enables Knex's seed feature to:

- Create seeds in specified subdirectories.
- Run seeds from multiple subdirectories, with a sorting option.

Link to **documentation** branch's pull request: https://github.com/knex/documentation/pull/220

### DIFF
--- a/test/jake-util/seeds-in-subdirs/production/first.js
+++ b/test/jake-util/seeds-in-subdirs/production/first.js
@@ -1,0 +1,1 @@
+exports.seed = (knex, Promise) => {};
--- a/test/jake/jakelib/seed-test.js
+++ b/test/jake/jakelib/seed-test.js
@@ -15,6 +15,75 @@ const KNEX = path.normalize(__dirname + '/../../../bin/cli.js');
 const taskList = [];
 /* * * TESTS * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
+test(
+  taskList,
+  "seed:make's --subdirectory specifies the destination directory",
+  (temp) => {
+    return assertExec(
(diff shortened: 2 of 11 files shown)


candyapplecorn:
Closed because actual manual testing revealed this code to not work. Thought the jake cli tests would take care of that for me but nope, there's more to it.

test/jake/jakelib/seed-test.js
```diff
@@ -15,6 +15,75 @@ const KNEX = path.normalize(__dirname + '/../../../bin/cli.js');
 const taskList = [];
 /* * * TESTS * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
+test(
```
kibertoad: I would say that for new tests `cli-testlab` tests would be preferable (and easier to write) than using jake directly. See test/cli/migrate-make.spec.js for an example.

Also async/await is preferred over using promise chains.

test/jake-util/seeds-in-subdirs/production/first.js
```diff
@@ -0,0 +1 @@
+exports.seed = (knex, Promise) => {};
```
kibertoad: Promise is not passed to seeds anymore, I think

test/jake/jakelib/seed-test.js
candyapplecorn: @kibertoad Thanks for the tips. I will use cli-testlab instead. Also, yeah, I definitely didn't want to be using promise chains and Bluebird. I just figured the rest of the codebase was using it, so I should too. So, cool! async await valhalla awaits me

kibertoad:
Released in 0.21.3

PR #3361, opened 2019-07-18
Update configuration typings to allow for oracle db connectionstring

evansimonross:
I am working on a typescript project with an oracle database using SID and we currently have it configured with a custom connectString. We were hoping to move to knex and ran into a typescript issue. Here is the resolution we used to get it working.


kibertoad:
Thanks!

kibertoad:
Released in 0.19.1

evansimonross:
Thanks very much! :)

Issue #3340, opened 2019-07-09
Any way we can configure the connection through an async function?


Ali-Dalal:
So, we have the db config stored in AWS secret manager. and we use knex with serverless labmda.

so the question is is it possible to load the config as an async function which calls the secret manager rather than setting a simple config object?

```
    production: {
        client: config.db.production.client,
        connection: async function () {
            // load secrets from secret manager as await function
           // then feed the required values to the return statement.
            return {
                host: secret.db.production.host,
                database:
                secret.db.production.name,
                user:
                secret.db.production.user,
                password:
                secret.db.production.password
            };
        },
        pool: {
            min: Number(config.db.production.min_pool) || 0,
            max: Number(config.db.staging.max_pool) || 1
        },
        migrations: {
            tableName: 'knex_migrations'
        }
    }
```

kibertoad:
This is not a currently supported feature, similar (but somewhat different) feature was requested here: https://github.com/tgriesser/knex/issues/2732

API proposal and (after proposal is approved) PR for this would be welcome.

elhigu:
I think this API is good (function can be also non async) and it could be specified in a way that as long as it returns the same configuration object `===` knex will not try to reinitialize anything and if configuration object is different, then knex internally tries to create new connections to new destination.. (at this phase if configuration object changes we may just throw an error).

Later on if it is needed we could also allow fetching the whole knex configuration object in async manner.

Ali-Dalal:
@elhigu Thanks for the explanation.

I read some of the comments you and other people mention and suggest.

for example, to load the config to a file via prescript and then load that file into knex.

The problem with that approach is security wise. the entire idea of loading the config via AWS secret manager is to keep the variables in memory. so no direct access via code files.

I tried to do async module export and load but the code gets really ugly.

what cause the issue to be really painful is I use bookshelf in the project and if there is a relation between two tables and we need to reflect the relation via bookshelf models then it can't be done via async require.

I will try to dig in knex code base and I will be really happy If I can implement that feature.

I already stuck into this problem since 4 days :(

elhigu:
Lets leave this opened until someone takes this over :)

Ali-Dalal:
Hi @elhigu ,
hope you are doing well.

I try to modify some part of the code and I come up with a solution.

The solution allows the connection to be entered as async function

```
config: {
connection: async function(){/*...*/}
/*rest of config*/
}
```
The files I modifed are:
1. client.js
2. dialects/postgres/index.js

[...]

PR #3364, opened 2019-07-20
#3340 Connection settings can be an async function only for PG client

Ali-Dalal:


lib/dialects/postgres/index.js
```diff
@@ -105,7 +105,9 @@ Object.assign(Client_PG.prototype, {
   // connection needs to be added to the pool.
   acquireRawConnection() {
     const client = this;
-    return new Bluebird(function(resolver, rejecter) {
+    return new Bluebird(async function(resolver, rejecter) {
+      if (typeof client.connectionSettings === 'function')
+        client.connectionSettings = await client.connectionSettings();
```
kibertoad: Does it make sense to resolve settings on every acquisition? Why not store them after lazy loading originally?

Ali-Dalal: @kibertoad pls correct me if I am wrong. To my understanding, Node runtime will execute this function only one time and store the result in memory.
Is this what you asking for?

```
let tmpSettings;
if (typeof client.connectionSettings === 'function')
  if(! tmpSettings)
        tmpSettings = await client.connectionSettings();
  client.connectionSettings = tmpSettings;
```

elhigu:
This relates to issue https://github.com/tgriesser/knex/issues/3340

lib/dialects/postgres/index.js
elhigu: @Ali-Dalal node runtime will not automatically store anything to memory. Each call to `client.connectionSettings()` creates new promise that gets executed.

Ali-Dalal: @elhigu What about the above snippet?

elhigu: maybe connection settings should not be set before this point... is there any reason why they would need to be cloned already in client.js?

elhigu: > @elhigu What about the above snippet?

That doesn't work either, because you are creating new tmsSettings on every call to aquireRawConnection.

Ali-Dalal: I tried some examples and if you can have a look at the following output

The code:
```
//dialects/pg/index.js
  acquireRawConnection() {
    const client = this;
    return new Bluebird(async function(resolver, rejecter) {
      console.log("Executing function");
      if (typeof client.connectionSettings === 'function') {
        client.connectionSettings = await client.connectionSettings();
      }
//... rest of the code

//knexfile.js
//...settings
    production: {
        client: 'pg',
        connection: async function () {
            console.log("executing function in connection");
            if (!dbConfig) {
                let secret = await getSecretFromSecretManager(`KEY_NAME`);
                let connection = JSON.parse(secret.SecretString);
                dbConfig = {
                    host: connection.host,
                    user: connection.username,
                    password: connection.password,
                    database: connection.db_name
                };
            }
            return dbConfig;
        },
//...rest of settings
```

log:
for the first time:
```
executing function in connection
getting secret from secret manager
  knex:query select **** undefined +0ms
  knex:bindings [ 'name', 1 ] undefined +0ms
  knex:client releasing connection to pool: __knexUid1 +403ms

```
for the 2nd time:
```
Executing function
  knex:client acquired connection from pool: __knexUid2 +2m
  knex:query select ***undefined +2m
  knex:bindings [ 'name', 1 ] undefined +2m
  knex:client releasing connection to pool: __knexUid2 +408ms
```
From the output above, the production connection config function is executed only 1 time and this mean fetching the db config from secret manager or any other provider will be called only 1 time.


To answer your question,

> maybe connection settings should not be set before this point... is there any reason why they would need to be cloned already in client.js?

the reason is if we switch cleint.js to export an async function then every module that needs to use client.js has to require it like this.

```
require('client').then(function(){
//...rest of the code
})
```
This is not freidnly code :(

can you please give a resolution on the output above and tell me your thoughts?

I tried the code of this pull request on staging environment and it works like a charm.

I've seen around 6 people asking for this feature.

If you are ok wtih it then I can write integration test and setup the rest of dialects

Thanks.

Ali-Dalal: > > @elhigu What about the above snippet?
> 
> That doesn't work either, because you are creating new tmsSettings on every call to aquireRawConnection.

tmpSettings should be defined outside the scope

elhigu: > maybe connection settings should not be set before this point... is there any reason why they would need to be cloned already in client.js?

I meant with that to use directly `config.connection` from `acquireRawConnection`, but I suppose it is not exposed or something. This was just something that really looks horrible:

```
client.connectionSettings = await client.connectionSettings();
```

So to prevent that + moving resolving real configuration to base implementation should be done by adding new async function to base client, which resolves configuration and which can be shared between every dialect.

Actually I don't thing there should be anything dialect specific in this feature (except for calling some base implementation function to get the config).

Ali-Dalal: @elhigu Thanks for the information.

> So to prevent that + moving resolving real configuration to base implementation should be done by adding new async function to base client, which resolves configuration and which can be shared between every dialect.

Can you guide me more? 
```
//client.js
function Client(config = {}) {
  this.config = config;
// How to resolve config.connectionSettings if it is async function?
  this.logger = new Logger(config);
//...rest of client code
module.exports = Client;
```
for example, if the client is a sync function and we need to convert it to an async function then everytime we need to use the client, we have to require in the following style
```
require('client').then(()=>{})
```

elhigu: By adding a new async method to the `Client` class. The same way that other methods of that class are implemented. https://github.com/tgriesser/knex/pull/3364/files#diff-50cfa59973c04321b5da0c6da0fdf4feL73

kibertoad:
@Ali-Dalal Any updates?

Ali-Dalal:
@kibertoad I haven't gotten the chance to modify the PR yet. for now, I am using Knex from my account's repo which I completely don't agree with.

I will try to modify the PR tomorrow and ask for the feed back again

Ali-Dalal:
closing this PR as it is a duplication of #3403 
https://github.com/tgriesser/knex/pull/3403

PR #3368, opened 2019-07-23
Add .isCompleted() to transaction

thekuom:
Adds .isCompleted() to a transactor, which just calls the underlying .isCompleted() from the transaction. I needed it because in my application, I pass a transaction around as part of a request and as a safeguard, I am rolling back the transaction. However, sometimes the transaction is already complete. This just harmlessly exposes whether I am allowed to commit/rollback or not.

### DIFF
--- a/test/unit/knex.js
+++ b/test/unit/knex.js
@@ -445,6 +445,15 @@ describe('knex', () => {
     expect(completed).to.be.true;
   });
 
+  it('returns false when calling isCompleted within a transaction handler', async () => {
+    const knex = Knex(sqliteConfig);
+    await knex.transaction((trx) => {
+      expect(trx.isCompleted()).to.be.false;
+
+      return trx.select(trx.raw('1 as result'));
+    });
+  });
+
   it('creating transaction copy with user params should throw an error', () => {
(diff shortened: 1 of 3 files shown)


kibertoad:
@thekuom Thank you, this is really useful! Could you please add documentation for it at https://github.com/knex/documentation?

thekuom:
Here you go: https://github.com/knex/documentation/pull/223

test/unit/knex.js
```diff
@@ -414,6 +414,37 @@ describe('knex', () => {
...
+    const knex = Knex(sqliteConfig);
+    const trxProvider = knex.transactionProvider();
+    const trx = await trxProvider();
+
+    await trx.rollback();
+
+    const completed = trx.isCompleted();
+    expect(completed).to.be.true;
+  });
+
```
elhigu: looks like test is missing where that API is used inside trx handler function.

Like this:

```
knex.transaction(trx => {
   trx.isCompleted();  
});
````

thekuom: Done

kibertoad:
Released in 0.19.1

PR #3372, opened 2019-07-26
#3367 Add interface for PG Connection object

### LINKED ISSUE #3367: Port missing in 'ConnectionConfig' TypeScript definition?
# Environment

Knex version: 0.18.4
Database + version: Postgres 9.5
OS: macOS 10.14

If issue is about TypeScript definitions, tag @lorefnon.

# Bug

(I'm opening this issue under the assumption that I'm supposed to use 'ConnectionConfig' to type my PG connection config object. If my assumption is wrong, then there seems to be a missing type definition for Postgres connection config?) 

---

The `pg` library [accepts a `port` key](https://github.com/brianc/node-postgres-docs/blob/master/content/features/1-connecting.mdx#programmatic), and passing the `port` key to knex indeed works: 
```
import Knex from 'knex';

// TS compiler complaining about 'port' not being in the type definition
const pgConnectionConfig: Knex.ConnectionConfig = {
    port: 1234,
    host: 'myhost'
    user: 'myuser'
    password: 'mypw'
    database: 'mydb'
};

// Attempts to connect to 'myhost' on port 1234
const k = Knex({
  client: 'pg',
  connection: pgConnectionConfig,
  ...
});
```
but this key is missing in the [`ConnectionConfig` interface](https://github.com/tgriesser/knex/blob/master/types/index.d.ts#L1618).

timorthi:
Closes #3367


types/index.d.ts
```diff
@@ -1737,6 +1739,21 @@ declare namespace Knex {
...
+    password?: string;
+    port?: number;
+    host?: string;
+    connectionString?: string;
+    keepAlive?: boolean;
+    stream?: stream.Duplex;
+    statement_timeout?: false | number;
+  }
+
+  type RedshiftConnectionConfig = PgConnectionConfig;
```
timorthi: Figured I'd sneak this in since they use the same driver anyway.

kibertoad:
Could you please update the PR to make sure build passes?

kibertoad:
Thanks!

PR #3382, opened 2019-08-04
Fix MSSQL escaping

kibertoad:


lib/dialects/mssql/index.js
```diff
@@ -207,7 +207,15 @@ Object.assign(Client_MSSQL.prototype, {
   },
 
   wrapIdentifierImpl(value) {
-    return value !== '*' ? `[${value.replace(/\[/g, '[')}]` : '*';
```
kibertoad: `.replace(/\[/g, '[')}]` path was removed because it doesn't seem to do anything. Please correct me if that's wrong.

test/unit/dialects/mssql.js
```diff
@@ -0,0 +1,37 @@
...
+    },
+  });
+
+  it('should escape statements with [] correctly', async () => {
+    const sql = knexInstance('projects')
+      .where('id] = 1 UNION SELECT 1, @@version -- --', 1)
+      .toSQL();
+    console.log(sql);
+    expect(sql.sql).to.equal(
+      'select * from [projects] where "id] = 1 UNION SELECT 1, @@version -- --" = ?'
```
elhigu: I would like this to be verified with integration test + to test identifier that has both `"` and `[]` chars in it.

elhigu: (I had no idea that mssql supports `"` quoting for identifiers.

kibertoad: [] are illegal symbols as per MSSQL documentation, apparently, so we don't need to test that. As advised by snyk people, I replaced our escaping logic with the one that `sequelize` uses, which removes `[]` in the first place.
And MSSQL does not support arrays.

Issue #3320, opened 2019-06-30
 Treat unhandled rejections as tests failure


kibertoad:
Ideally we should add this snippet to knex mocha setup to treat unhandled rejections as tests failure and not just warnings:

```
process.on('unhandledRejection', (reason) => {
  console.error('Unhandled Rejection:', reason);
  process.exit(1);
});
```

But first we need to fix all our existing unhandled rejections in tests.

vonagam:
Here is a package that you can use instead of custom snippet - [`loud-rejection`](https://github.com/sindresorhus/loud-rejection). Just require `loud-rejection/register` through mocha `--require` option or in setup file. It will allow tests to finish running instead of abrupt exit on an unhandled rejection.

elhigu:
I'd rather have that snipped instead of extra dev dependency just for that.

vonagam:
Ok, here is a snippet that won't abrupt exit tests:

```js
let count = 0;

 process.on('unhandledRejection', (reason) => {
  console.error('Unhandled rejection:', reason);
  count++;
});

 process.on('exit', (code) => {
  if (count !== 0) {
    console.error(`Total count of unhandled rejections: ${count}`);
    process.exitCode = code || 1;
  }
});
```

PR #3393, opened 2019-08-13
More consistent handling of nested transactions

### LINKED ISSUE #3389: Using nested transactions with save points
# Environment

Knex version: `0.19.1`
Database + version: `PostgreSQL 11`
OS: `MacOS Mojave`

# Bug
I am trying to use nested transactions with save points and it doesn't seems to be working with promises API.

Here's some code (I am sharing the minimum required code to reproduce the problem)

**Following doesn't work**

```js
async function run () {
  const trx2 = await trx.transaction()
  console.log('never reached here')
}
```

**Following works**

```js
function runWithCallbacks () {
  knex.transaction((trx) => {
   trx.transaction((trx2) => {
    console.log('reached here>')
   })
 })
}
```

The first example doesn't work since the promise doesn't resolve unless `trx2` commit/rollback and it cannot, since the code never moves to the next line.

Is this the intended behavior?

kibertoad:
fixes #3389 #3320

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 # .travis.yml
 language: node_js
-dist: xenial
 sudo: required
 
 cache:
@@ -18,7 +17,7 @@ matrix:
       env: TESTSCRIPT=test:nyc DB="oracledb mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
 install:
   - npm i
-  - (echo $DB | grep oracledb) && npm install oracledb || true
+  - (echo $DB | grep oracledb) && npm install oracledb@3.1.2 || true
 
 before_script:
   - npm run db:start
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -18,7 +17,7 @@ matrix:
       env: TESTSCRIPT=test:nyc DB="oracledb mssql mysql mysql2 postgres sqlite3" KNEX_TEST_TIMEOUT=60000
 install:
   - npm i
-  - (echo $DB | grep oracledb) && npm install oracledb || true
+  - (echo $DB | grep oracledb) && npm install oracledb@3.1.2 || true
```
kibertoad: See #3396 for the reason of this change.

elhigu:
Is there need to add more info about nested transactions and how rejections are happening to documentation?

kibertoad:
@elhigu I'll review it just in case.

PR #3394, opened 2019-08-13
Fallback to JSON when using JSONB in MySQL

### LINKED ISSUE #3386: jsonb function in schema builder appears to be broken in mysql
# Environment

Knex version: 0.19.0
Database + version: mysql 8.0.17
OS: [debian](https://github.com/docker-library/mysql/blob/master/8.0/Dockerfile#L1)

# Bug

   `jsonb` column builder in mysql should fallback to json: 

   As per the docs: 

   > jsonb — table.jsonb(name)
   > Adds a jsonb column. Works similar to table.json(), but uses native jsonb type if possible.

2. Error message

   `ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'undefined)' at line 1`

```
const config = { client: 'mysql', connection: process.env.DB_CONNECTION_STRING }
const knex =  require('knex')(config);
> knex.schema.createTable('users', t => { t.increments('id'); t.jsonb('metadata') }).toSQL()
[ { sql:
     'create table `users` (`id` int unsigned not null auto_increment primary key, `metadata` undefined)',
    bindings: [] } ]
```

*Note:* `undefined` is used as type of `metadata`.

kibertoad:
fixes #3386

### DIFF
--- a/lib/dialects/mysql/schema/columncompiler.js
+++ b/lib/dialects/mysql/schema/columncompiler.js
@@ -113,6 +113,10 @@ Object.assign(ColumnCompiler_MySQL.prototype, {
     return 'json';
   },
 
+  jsonb() {
+    return 'json';
+  },
+
   // Modifiers
   // ------
 
--- a/test/unit/schema/mysql.js
+++ b/test/unit/schema/mysql.js
@@ -66,19 +66,29 @@ module.exports = function(dialect) {
       );
     });
 
-    if (dialect !== 'maria') {
-      it('adding json', function() {
-        tableSql = client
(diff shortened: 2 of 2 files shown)


kibertoad:
@lorefnon Could you take a look?

kibertoad:
Please ignore Oracle test failures, this is due to newly released Oracle driver, I've pinned its version in https://github.com/tgriesser/knex/pull/3393

lorefnon:
Thanks :)

Issue #2002, opened 2017-04-07
CockroachDB Support


ansarizafar:
Please include support for CockroachDB [https://www.cockroachlabs.com/](https://www.cockroachlabs.com/)

absolux:
@ansarizafar 
CockroachDB can be used with the `pg` driver [look at](https://www.cockroachlabs.com/docs/build-a-nodejs-app-with-cockroachdb.html)

elhigu:
Support is implemented in https://github.com/tgriesser/knex/pull/1993

nicck:
hm... I would say passing version of connected db is implemented, not CockroachDB support.

elhigu:
@nicck is there some CockroachDB queries which are not compatible with postgresql? If so I'll reopen this as a feature request.

nicck:
It's the opposite, PostgreSQL queries not supported by CockroachDB. I mentioned some of them in https://github.com/cockroachdb/cockroach/issues/15299#issuecomment-301061458. CockroachDB is not PostgreSQL, protocol is compatible but [CockroachDB support only subset of SQL queries](https://www.cockroachlabs.com/docs/sql-feature-support.html) which are valid in PostgreSQL.

elhigu:
@nicck since cockroach db queries is subset of postgres queries, you should be able to use cockroach db just fine with postgres dialect. Just don't use those features that are not supported by it.

nicck:
I can avoid using `uuid` and `json` types, but some sql queries, like [`select * for update`](https://github.com/tgriesser/knex/blob/a5f4c395297e4262aebf85534ec523509d2294e9/src/migrate/index.js#L175) and [`... table_schema = current_schema`](https://github.com/tgriesser/knex/blob/2e1a459a9e740f24b9a4647bd4da427854e551dd/src/dialects/postgres/schema/compiler.js#L22) are parts of knex as I understand and it's not possible to avoid them if you use `pg` dialect.

elhigu:
@nicck ah right... those queries done by migration system requires some fixing. At least that select for update could be avoided by some flag which allows migration locking to throw an error instead of blocking (like it did work while ago).

So currently migration system is unusable with CockroachDB. Do you know if there are any other problems?

thomasdavis:
I was trailing Cockroach DB yesterday and also got stuck on the migrations. Commenting as a plus one and so I can watch the thread.

elhigu:
[...]

PR #3395, opened 2019-08-13
migrate: Refactor _lockMigrations to avoid forUpdate

bdarnell:
Since _isLocked is only called just before _lockMigrations, it is redundant
and we can accomplish the same  thing by checking the rowCount returned
by the update (verified by testing with postgres, mysql, sqlite, and
mssql).

The benefit of this change is improving compatbility with CockroachDB,
which does not support FOR UPDATE.

Updates #2002

### DIFF
--- a/lib/migrate/Migrator.js
+++ b/lib/migrate/Migrator.js
@@ -240,32 +240,23 @@ class Migrator {
     }
   }
 
-  _isLocked(trx) {
-    const tableName = getLockTableName(this.config.tableName);
-    return getTable(this.knex, tableName, this.config.schemaName)
-      .transacting(trx)
-      .forUpdate()
-      .select('*')
-      .then((data) => data[0].is_locked);
-  }
-
   _lockMigrations(trx) {
     const tableName = getLockTableName(this.config.tableName);
     return getTable(this.knex, tableName, this.config.schemaName)
       .transacting(trx)
-      .update({ is_locked: 1 });
+      .where('is_locked', '=', 0)
+      .update({ is_locked: 1 })
+      .then((rowCount) => {
+        if (rowCount != 1) {
+          throw new Error('Migration table is already locked');
+        }
+      });
   }
 
   _getLock(trx) {
     const transact = trx ? (fn) => fn(trx) : (fn) => this.knex.transaction(fn);
     return transact((trx) => {
-      return this._isLocked(trx)
-        .then((isLocked) => {
-          if (isLocked) {
-            throw new Error('Migration table is already locked');
-          }
-        })
-        .then(() => this._lockMigrations(trx));
+      return this._lockMigrations(trx);
     }).catch((err) => {
       throw new LockError(err.message);
     });
--- a/test/integration/migrate/index.js
+++ b/test/integration/migrate/index.js
@@ -261,6 +261,57 @@ module.exports = function(knex) {
           });
       });
 
+      it('should work with concurent calls to _lockMigrations', async function() {
+        if (knex.client.driverName == 'sqlite3') {
+          // sqlite doesn't support concurrency
+          this.skip();
+          return;
+        }
+
+        const migrator = knex.migrate;
+        try {
+          // Start two transactions and call _lockMigrations in each of them.
+          // Simulate a race condition by waiting until both are started before
+          // attempting to commit either one. Exactly one should succeed.
+          //
+          // Both orderings are legitimate, but in practice the first transaction
+          // to start will be the one that succeeds in all currently supported
+          // databases (CockroachDB 1.x is an example of a database where the
+          // second transaction would win, but this changed in 2.0). This test
+          // assumes the first transaction wins, but could be refactored to support
+          // both orderings if desired.
+          const trx1 = await knex.transaction();
+          await migrator._lockMigrations(trx1);
+          const trx2 = await knex.transaction();
+          // trx1 has a pending write lock, so the second call to _lockMigrations
+          // will block (unless we're on a DB that resolves the transaction in
+          // the other order as mentioned above).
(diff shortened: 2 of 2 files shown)


lib/migrate/Migrator.js
```diff
@@ -240,32 +240,23 @@ class Migrator {
...
-      .select('*')
-      .then((data) => data[0].is_locked);
-  }
-
   _lockMigrations(trx) {
     const tableName = getLockTableName(this.config.tableName);
     return getTable(this.knex, tableName, this.config.schemaName)
       .transacting(trx)
-      .update({ is_locked: 1 });
+      .where('is_locked', '=', 0)
```
elhigu: This does not work. There can be two processes at the same time arriving to this row and both will return one row and both then updates the lock value to 1.

elhigu:
This approach doesn't work.

lib/migrate/Migrator.js
elhigu: forUpdate exactly causes DB to prevent that from happening.

kibertoad:
@elhigu Couldn't this be an alternate mechanism that one can opt-in for? Considering that usually migrations are not being executed in same environment concurrently from multiple places, this particular approach is unlikely to cause many real issues in actual production environments for users that may rely on it.

elhigu:
That would mean basically option for omitting is locked check completely. I'm not against it, but still I would prefer solution, which actually works. I'm pretty sure there must be a way to do something like this that actually works also on CockroachDB

kibertoad:
@elhigu Option to disable lock check might actually be useful.

elhigu:
In that case those changes that were made to _lockMigrations could be implemented anyways as a secondary safety measure.

bdarnell:
> This does not work. There can be two processes at the same time arriving to this row and both will return one row and both then updates the lock value to 1.

Can you please elaborate on when exactly this fails? I still believe it works, although reasoning about sub-serializable isolation levels is always tricky and under-documented. 

My informal argument is that in any database that distinguishes between regular `SELECT` and `FOR UPDATE`, the `WHERE` clause of an `UPDATE` statement should be evaluated in the `FOR UPDATE` mode (the [mysql docs](https://dev.mysql.com/doc/refman/8.0/en/innodb-locks-set.html) are explicit that `UPDATE WHERE` gets an exclusive lock on every record it encounters). If there are two concurrent invocations of `UPDATE t SET is_locked=1 WHERE is_locked=0`, the first one will grab a write lock on the row when it reads the zero value, then the second one will have to wait until the first transaction commits (so it will see the new value of 1). Therefore only the transaction that causes a transition from 0 to 1 will return "1 row affected".

There are databases that don't work this way, for example ones that are asynchronously replicated and eventually consistent. My proposal would indeed be incorrect for these systems. However, I've never seen a system like that that has a useful implementation of `SELECT FOR UPDATE`, so the original implementation would also fail in this case.

> I'm pretty sure there must be a way to do something like this that actually works also on CockroachDB

One alternative is for the CockroachDB dialect to turn `forUpdate()` into a no-op (or for the database itself to accept and ignore the `FOR UPDATE` clause). CockroachDB always runs with `SERIALIZABLE` isolation and most uses of `FOR UPDATE` are unnecessary in this mode. However, I'd prefer to avoid unnecessary uses of `FOR UPDATE` rather than just ignoring them (and eventually implement it in CockroachDB), since there are (rare) cases where it's still useful in `SERIALIZABLE` transactions (mainly to control lock ordering and minimize deadlocks).

elhigu:
> Can you please elaborate on when exactly this fails? I still believe it works, although reasoning about sub-serializable isolation levels is always tricky and under-documented.

Both processes will get the lock and start running migrations parallel.

> My informal argument is that in any database that distinguishes between regular SELECT and FOR UPDATE, the WHERE clause of an UPDATE statement should be evaluated in the FOR UPDATE mode 

Hmm... that actually might be true +1 I have never tried that if it really works that way. Only with plain selects and then updating... I'll try it out how it behaves.

elhigu:
It seems to work at least on postgres so that indeed is a lot better approach 👍 We should add integration test for this to also see in which databases that is supported. 

update-locking-test

bdarnell:
I've added a test, which passes on postgres, mysql, and mssql (it doesn't work on sqlite because it doesn't support concurrent transactions anyway). I followed the existing promises-and-callback style but it would be simpler if it's OK to introduce async/await into this file.

kibertoad:
@bdarnell By all means feel free to use async-await! Promise chains are just legacy.

kibertoad:
@bdarnell Can you take a look at plethora of failures in https://travis-ci.org/tgriesser/knex/jobs/576541045? Something fishy is going on there.

bdarnell:
```
  1) Integration Tests
       oracle | oracledb
         knex.migrate
           knex.migrate.latest
             should work with concurent calls to _lockMigrations:
     TypeError: knex.transaction(...).then(...).then(...).then(...).then(...).then(...).then(...).finally is not a function
      at Context.<anonymous> (test/integration/migrate/index.js:327:19)
```

This is the root cause, and then because we failed to clean up in the `finally` callback, everything afterwards fails with some variation of "migration table already locked" or migrations just silently fail to run. (There's a global `before` function that's supposed to reset this lock, but it doesn't seem to be doing its job and I don't understand why)

The failures are all on the Node 8 job; that version of node must be using a version of Promise that doesn't have the `finally` method. Hopefully this just goes away if I change everything to async/await.

kibertoad:
@bdarnell Yup, that seems to be it.
I've started preparing 0.19.3 release, but I can delay it until this PR is merged :)

kibertoad:
@bdarnell Appreciate the time spent on this!

kibertoad:
Will cut a new release today.

PR #3398, opened 2019-08-16
Fix broken noWait test

kukiric:
This fixes one of the tests from #2961 that got broken recently, and throws a very minor refactoring of the skipLocked tests to make them a little bit cleaner.

I had to re-write most of the test to use try/catch to have it resolve properly (otherwise the catch handler wouldn't fully handle the error when using await), but after doing so, it also seems the test is easier to read this way.

I also made the involved tests clear the related rows before starting, as I was getting test failures after running them multiple times because they were designed with only a single run in mind (oops).


kibertoad:
Thanks a lot!

PR #3399, opened 2019-08-16
fix: make transaction rejection consistent across dialects

thetutlage:
Making `mysql` and `mysql2` dialects use the same `sql` check as the base transaction class.

### DIFF
--- a/lib/dialects/mysql/transaction.js
+++ b/lib/dialects/mysql/transaction.js
@@ -30,11 +30,7 @@ Object.assign(Transaction_MySQL.prototype, {
         if (status === 1) t._resolver(value);
         if (status === 2) {
           if (isUndefined(value)) {
-            if (
-              sql &&
-              sql.toUpperCase() === 'ROLLBACK' &&
-              t.doNotRejectOnRollback
-            ) {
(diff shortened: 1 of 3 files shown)


lib/dialects/mysql/transaction.js
```diff
@@ -30,11 +30,7 @@ Object.assign(Transaction_MySQL.prototype, {
         if (status === 1) t._resolver(value);
         if (status === 2) {
           if (isUndefined(value)) {
-            if (
-              sql &&
-              sql.toUpperCase() === 'ROLLBACK' &&
-              t.doNotRejectOnRollback
-            ) {
+            if (t.doNotRejectOnRollback && /^ROLLBACK\b/i.test(sql)) {
```
kibertoad: What happens if sql is nil here? It just resolves to false or throw an error?

kibertoad:
You can pull latest changes from master to resolve oracle test failures.

lib/dialects/mysql/transaction.js
thetutlage: Returns `false`

kibertoad:
Thanks! I'll try to publish new release tonight.

thetutlage:
Wow, thanks 😊

Issue #3307, opened 2019-06-25
Schema qualifier for enums


parm-rm:
# Environment

Knex version: 0.17.6
Database + version: PostgreSQL 11.4
OS: Manjaro Linux 18.0.4

# Feature discussion / request

While creating a table with an explicit schema and I'm looking to use a native enum as one of the columns. I haven't found any documentation about how to qualify an enum declaration with schema name.

I've tried the following
```
table.enum('bazz_type', [ 'foo', 'bar' ], {
	useNative : true,
	enumName  : 'other_schema.bazz_types',
});
```
but it lead to creation of `other_schema.bazz_types` under `public` schema
```
Debugging output:
'create type ' +
'"other_schema.bazz_types" ' +
"as enum ('foo', 'bar')
```
One possible solution could be to add a `schema` string option under `EnumOptions`.

elhigu:
Yep, looks like a bug that enum type creation does not respect withSchema.

https://runkit.com/embed/e0rde4cgbloq

for a workaround this should work though:

```
knex.schema.withSchema('foo').table('table', t => {
  t.enum('enum', ['foo', 'bar'], { useNative: true, enumName: knex.ref('foo.enu_type') });
})```

**Comment by elhigu** [comment]
Do you want table and enum type to live in different schemas?

**Comment by parm-rm** [comment]
> Do you want table and enum type to live in different schemas?

No, just wanted it to respect table schema context. Having an schema override for enums too is "nice-to-have".

**Comment by elhigu** [comment]
Ok, I'll label this as a bug then :)

**Comment by parm-rm** [comment]
> Yep, looks like a bug that enum type creation does not respect withSchema.
> 
> https://runkit.com/embed/e0rde4cgbloq
> 
> for a workaround this should work though:
> 
> ```
> knex.schema.withSchema('foo').table('table', t => {
>   t.enum('enum', ['foo', 'bar'], { useNative: true, enumName: knex.ref('foo.enu_type') });
> })
> ```

That workaround doesn't seem to work. Instead of generating the following SQL
```
create type "foo"."enu_type" as enum ('foo', 'bar')
```
it generates
```
create type ""foo"."enu_type"" as enum ('foo', 'bar')
```
which is invalid and leads to `zero-length delimited identifier at or near """"` error.

elhigu:
[...]

PR #3400, opened 2019-08-16
Fix native enum with specified schema (#3307)

Saka7:
# Fixes issue: #3307

## Current behaviour:

```javascript
knex
      .schemaBuilder()
      .withSchema('test')
      .table('users', function(table) {
        table
          .enu('foo', ['bar', 'baz'], {
            useNative: true,
            enumName: 'foo_bar',
          })
      })
      .toSQL();
```

## Outputs:
```sql
-- ...
create type "foo_bar" as enum ('bar', 'baz');
alter table "test"."users" add column "foo" "foo_bar";
-- ...
```

Schema builder doesn't take into account `withSchema` expression and always creates a type in `public` schema.

## API proposal:
- `options.schema = '<schema name>'` - use specified schema name;
- `options.schema = true` -  use schema from `withSchema` expression (for backward compatibility if needed);

### Example 1: `options.schema = '<schema name>'` 

```javascript
knex
      .schemaBuilder()
      .withSchema('test')
      .table('users', function(table) {
        table
          .enu('foo', ['bar', 'baz'], {
            useNative: true,
            enumName: 'foo_bar',
            schema: 'test_2',
          })
      })
      .toSQL();
```

Outputs:
```sql
-- ...
create type "test_2"."foo_bar" as enum ('bar', 'baz');
alter table "test"."users" add column "foo" "test_2"."foo_bar";
-- ...
```

### Example 2: `options.schema = true`

```javascript
knex
      .schemaBuilder()
      .withSchema('test')
      .table('users', function(table) {
        table
          .enu('foo', ['bar', 'baz'], {
            useNative: true,
            enumName: 'foo_bar',
            schema: true,
          })
      })
      .toSQL();
```

Outputs:
```sql
-- ...
create type "test"."foo_bar" as enum ('bar', 'baz');
alter table "test"."users" add column "foo" "test"."foo_bar";
-- ...
```

### DIFF
--- a/lib/dialects/postgres/schema/columncompiler.js
+++ b/lib/dialects/postgres/schema/columncompiler.js
@@ -36,11 +36,10 @@ Object.assign(ColumnCompiler_PG.prototype, {
 
     if (options.useNative) {
       let enumName = '';
+      const schemaName = this.tableCompiler.schemaNameRaw;
 
-      if (options.schema === true && this.tableCompiler.schemaNameRaw) {
-        enumName += `"${this.tableCompiler.schemaNameRaw}".`;
-      } else if (options.schema && typeof options.schema === 'string') {
-        enumName += `"${options.schema}".`;
+      if (schemaName) {
+        enumName += `"${schemaName}".`;
       }
 
       enumName += `"${options.enumName}"`;
--- a/lib/dialects/postgres/schema/columncompiler.js
+++ b/lib/dialects/postgres/schema/columncompiler.js
@@ -35,13 +35,23 @@ Object.assign(ColumnCompiler_PG.prototype, {
         : allowed.join("', '");
 
     if (options.useNative) {
+      let enumName = '';
+
+      if (options.schema === true && this.tableCompiler.schemaNameRaw) {
+        enumName += `"${this.tableCompiler.schemaNameRaw}".`;
+      } else if (options.schema && typeof options.schema === 'string') {
+        enumName += `"${options.schema}".`;
+      }
+
+      enumName += `"${options.enumName}"`;
+
       if (!options.existingType) {
         this.tableCompiler.unshiftQuery(
-          `create type "${options.enumName}" as enum ('${values}')`
+          `create type ${enumName} as enum ('${values}')`
         );
       }
 
-      return `"${options.enumName}"`;
+      return enumName;
     }
     return `text check (${this.formatter.wrap(this.args[0])} in ('${values}'))`;
   },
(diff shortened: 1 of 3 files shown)


Saka7:
Some integration test for `oracledb` have been failed, it doesn't look related to changes added in the PR. Can somebody take a look, please?

Saka7:
Another topic: please let me know if we need to keep backward compatibility for native enum (always create enum in the `public` schema). If no, we could remove this ugly `schema: true` options and always use specified schema from `withSchema` expression.

kibertoad:
@Saka7 You can pull latest changes from master to fix Oracle failures. Let me read the proposal and the PR in more details and I'll get back to you on it :)

Saka7:
@kibertoad Did you get a chance to look at this?

kibertoad:
@Saka7 In this particular case I would consider ignoring `.withSchema` a bug and not intended behaviour, so maintaining backwards incompatibility doesn't make sense; if user was explicitly specifying schema, he shouldn't be surprised when it is actually used.

kibertoad:
Fix by itself is great, thank you for your contribution!

Saka7:
@kibertoad ok, so I'll remove `schema: boolean` option. A schema for enum will be used from `.withSchema()`. And it can be overridden by the value from `schema: string` option.

kibertoad:
@Saka7 Do you see benefit from option.schema? Isn't it redundant and inconsistent with other operations? What are th euse-cases for it?

Saka7:
@kibertoad In case somebody by some reason needs to create enum in different schema than the table. But, I guess it's a very rare case, so we might remove `schema` option at all.

kibertoad:
@Saka7 Then we would need to add same for all operations :)

Saka7:
@kibertoad ok, I'll remove `schema` option.

kibertoad:
Thank you!

Saka7:
@kibertoad thank you for the review!

ivanovych666:
Sorry, but, previously we had a problem that for ENUM we can't use any schema except "public" and now we have a problem that we can't use existing types from other schemas or even from "public" schema. This is ridiculous. Maybe creating new tables in one schema and creating new types in another is not well used, but how to create a table with ENUM column where we want to use existing TYPE from a different schema? Please don't forget that we have the "existingType" option and add the option to manually define the schema name back.

kibertoad:
@ivanovych666 That is a valid case, however, I'm not sure if having options.schema is the best solution for that, as if it would be affecting both source and target enum, you still won't be able to use enums from one schema inside another schema. Can't you just append `${schemaName}.` to source enum manually right now?
I'm OK with adding `existingTypeSchema` in a new PR since that provides an explicit way to achieve this specific goal, though. Feel free to submit one.

ivanovych666:
@kibertoad I have created a simple solution, please check here https://github.com/tgriesser/knex/pull/3413

PR #3403, opened 2019-08-19
Connection config provider

Ali-Dalal:
As requested in this PR https://github.com/tgriesser/knex/pull/3364
@elhigu @kibertoad Can you please check? if the changes are ok then I will add unit test and update the rest of the dialects


lib/client.js
```diff
@@ -53,8 +54,12 @@ function Client(config = {}) {
   if (config.version) {
     this.version = config.version;
   }
-
-  this.connectionSettings = cloneDeep(config.connection || {});
+  // if the type of connection settings is a function, pass it to the next layer
+  if (typeof config.connection === 'function') {
```
kibertoad: _.isFunction is preferable

lib/dialects/postgres/index.js
```diff
@@ -105,7 +106,8 @@ Object.assign(Client_PG.prototype, {
   // connection needs to be added to the pool.
   acquireRawConnection() {
     const client = this;
-    return new Bluebird(function(resolver, rejecter) {
+    return new Bluebird(async function(resolver, rejecter) {
+      await client.connection().resolveConnectionConfig();
```
kibertoad: We can simplify this, I think, instead of 
```
await client.connection().resolveConnectionConfig();
const connection = new client.driver.Client(client.connectionSettings);
```

We can do just
```
const connection = await client.connection().resolveConnectionConfig();
```

lib/connection.js
```diff
@@ -0,0 +1,18 @@
+class Connection {
```
kibertoad: Considering that this is not a connection per se, but just connection config resolver, I would recommend naming it as such - ConnectionConfigResolver

lib/client.js
kibertoad: I'd make this whole thing more encapsulated:
```
this.connectionSettingsProvider = new ConnectionSettingsProvider(config.connection)
```
And move the isFunction branching inside provider itself.
You don't even need dependency on the client, I think.

lib/dialects/postgres/index.js
kibertoad: In case we need connectionSettings anywhere else, we could also add `client.connectionSettings = connectionConfig` here, but it shouldn't be provider's responsibility to mutate that state.

lib/client.js
Ali-Dalal: https://github.com/tgriesser/knex/pull/3403/files#diff-cbc1ae6887774cf34c8d4bc73dece12fR11

lib/dialects/postgres/index.js
Ali-Dalal: As requested
```
const connectionSettings = await client
        .connectionConfigProvider()
        .resolveConnectionConfig();
      const connection = new client.driver.Client(connectionSettings);
```
https://github.com/tgriesser/knex/pull/3403/files#diff-0fdffe855b98a6c04d603ff6c442c163R110

I agree with you that we shouldn't mutate the state of Client. but if `client.connectionSettings ` is used somewhere else, then this might cause a problem.

What I have done is I mutated the state of the client in `connectionProvider.js`
https://github.com/tgriesser/knex/pull/3403/files#diff-cbc1ae6887774cf34c8d4bc73dece12fR14

lib/connection.js
Ali-Dalal: https://github.com/tgriesser/knex/pull/3403/files#diff-cbc1ae6887774cf34c8d4bc73dece12fR14

lib/client.js
Ali-Dalal: yes, you are right. I've removed connection handling in client.js and moved it to connectionProvider.js

lib/client.js
```diff
@@ -53,8 +54,12 @@ function Client(config = {}) {
   if (config.version) {
     this.version = config.version;
   }
-
-  this.connectionSettings = cloneDeep(config.connection || {});
+  // if the type of connection settings is a function, pass it to the next layer
```
kibertoad: Why is this commented out?

lib/connectionProvider.js
```diff
@@ -0,0 +1,25 @@
+const { isFunction, cloneDeep } = require('lodash');
+
+class ConnectionConfigProvider {
+  constructor(client) {
+    this.client = client;
+    this.resolvedConnectionSettings = null;
```
kibertoad: Why not resolve settings at this step already? At the very least we can definitely do that if it's a static config.

lib/connectionProvider.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  async resolveConnectionConfig() {
+    if (this.resolvedConnectionSettings) return this.resolvedConnectionSettings;
+    if (isFunction(this.client.config.connection)) {
```
kibertoad: We can do the check once in constructor and set some flag after that.

lib/dialects/mssql/index.js
```diff
@@ -213,8 +213,12 @@ Object.assign(Client_MSSQL.prototype, {
   // Get a raw connection, called by the `pool` whenever a new
   // connection needs to be added to the pool.
   acquireRawConnection() {
-    return new Bluebird((resolver, rejecter) => {
-      const settings = Object.assign({}, this.connectionSettings);
+    const client = this;
+    return new Bluebird(async (resolver, rejecter) => {
+      const settings = await client
```
kibertoad: Is there a reason why we might want to set connectionConfigProvider each time for a new connection anew and not just set it once on a client when it is created?

lib/dialects/mssql/index.js
```diff
@@ -213,8 +213,12 @@ Object.assign(Client_MSSQL.prototype, {
...
+        .connectionConfigProvider()
+        .resolveConnectionConfig();
+      // const settings = Object.assign({}, this.connectionSettings);
```
kibertoad: is this needed?

lib/client.js
Ali-Dalal: conniectionProvider takes the responsibility to manage the connection settings.

https://github.com/tgriesser/knex/pull/3403/files#diff-cbc1ae6887774cf34c8d4bc73dece12fR16

lib/connectionProvider.js
Ali-Dalal: Can you give more details about this, please?

if we need to manage this in `ConnectionConfigProvider` constructor, the how-to manipulate the promise as the type of the function is async?

lib/dialects/mssql/index.js
Ali-Dalal: in `ConnectionConfigProvider` class I processed `config.connection` resolve the connection settings for only one time.  So calling 
```
const settings = await client
        .connectionConfigProvider()
        .resolveConnectionConfig();
```
 will return the processed object direclty without processing anything anymore.

lib/dialects/mssql/index.js
Ali-Dalal: Yes, it is needed. if the config is passed as an async function, then this will cause a problem without resolving the config.

Also,  this line 
```
const settings = Object.assign({}, this.connectionSettings);
```
is not needed any more as `ConnectionConfigProvider` is taking the responsibility for this.

I will clarify the need for this PR again in case I miss any clarification.

The main purpose is we need to fetch the config of the DB from an async store (AWS SSM as an example) and this requires passing the config as an async function in order to wait for the config to be loaded from the promise.

Travis 1st stage passes but the other stages failed and I believe it is because of oracledb dialect which fails only in one of the test cases. Can you have a look please and help?

lib/client.js
kibertoad: yeah, but why
```
  // if (typeof config.connection === 'function') {
  //   this.connectionSettings = config.connection;
  // } else {
  //   this.connectionSettings = cloneDeep(config.connection || {});
  // }
```
is committed to the repo?

kibertoad: and the comment `if the type of connection settings is a function, pass it to the next layer` doesn't make sense anymore either

lib/connectionProvider.js
kibertoad: What I mean is this:

1) In constructor check if it's a function. If it is - set `this.isConfigSourceFunction = true` and do nothing else. If it is not - set `this.isConfigSourceFunction = false` and `this.resolvedConnectionSettings = cloneDeep <...>`

2) in `resolveConnectionConfig ` check if `this.resolvedConnectionSettings` is set, if it is, return, if it is not and `isConfigSourceFunction`, resolve from function, if neither, probably throw an error.

lib/dialects/mssql/index.js
kibertoad: Will it?

```
  connectionConfigProvider() {
    return new ConnectionConfigProvider(this);
  },
```

lib/dialects/mssql/index.js
kibertoad: Can you please delete the unneeded commented out code from PR?

kibertoad: re: Oracle

this seems directly related:
```
1) Oracledb driver tests
       OracleDb externalAuth
         externalAuth and connectString should be sent to the getConnection:
     AssertionError: expected getConnection to have been called exactly once, but it was called 0 times
      at Context.<anonymous> (test/unit/dialects/oracledb.js:36:25)
```

I believe you'd need to update some of mocks in oracle tests, as call hierarchy has obviously changed.

lib/client.js
Ali-Dalal: I've removed the comments

lib/connectionProvider.js
Ali-Dalal: I think this line in `ConnectionConfigProvider` class does the same, right?
```
  async resolveConnectionConfig() {
    if (this.resolvedConnectionSettings) return this.resolvedConnectionSettings;
```
the lines after check for the type of the connection. if it is a function resolve it, other wise, deepclone it.

lib/dialects/mssql/index.js
Ali-Dalal: any recommendation?

lib/connectionProvider.js
kibertoad: Partially. What I'm saying is that everything that can be moved from `resolveConnectionConfig ` to constructor should ideally be moved, `resolveConnectionConfig ` should be as simple as possible.
Current implementation is different from what I've described.

lib/dialects/mssql/index.js
Ali-Dalal: re: Oracledb

I believe the issue is because of switching the bludbird callback to async in line 79

```
const asyncConnection = new Bluebird(async function(resolver, rejecter) {
//...res of the code
}
```
where the linter is showing an error at line 105
```
    client.driver.fetchAsString = client.fetchAsString;
```
error:
`possible race condition client.driver.fetchAsString might be reassinged based on an outdated value of client.driver.fetchAsString`

need to debug the code but the issue I can't get the tests working on my machine due to insufficient storge  issue (mac 128 GB)

any ideas?

lib/client.js
Ali-Dalal: removed

lib/dialects/mssql/index.js
kibertoad: As I've said, you can call `connectionConfigProvider ` once after client is created and configuration is known. And probably it should be renamed to `initConnectionConfigProvider` to emphasise it mutates the state.

kibertoad:
@Ali-Dalal Do you need any help for finishing the PR?

Ali-Dalal:
@kibertoad Yes please. 

I need clarification on this point 

> in resolveConnectionConfig check if this.resolvedConnectionSettings is set, if it is, return, if it is not and isConfigSourceFunction, resolve from function, if neither, probably throw an error.

Do we need to resolve the config if it is a function inside the constructor?

kibertoad:
@Ali-Dalal As I've said before, we can't resolve config from function since it could be async. We could do it if it's a static config, though, but disregard this part, we can refactor it later. Important part at this point is to finalize the implementation :).

lib/dialects/mssql/index.js
kibertoad: I assume this is still relevant?

Ali-Dalal: @kibertoad I've checked the tests. only Oracle dialect fails. and honestly, I have no idea why. can you please help? I will do one more round to try to find the problem

kibertoad:
@maximelkin Any chance you could take a look at the Oracle problem to see what is wrong here? I don't have readily available instance nor configured Docker to test easily myself, unfortunately.

kibertoad:
@Ali-Dalal Can you pull changes from master, btw? I wonder if latest fixes in master would resolve the issue.

lib/dialects/oracledb/index.js
```diff
@@ -76,27 +76,30 @@ Client_Oracledb.prototype.prepBindings = function(bindings) {
 // connection needs to be added to the pool.
 Client_Oracledb.prototype.acquireRawConnection = function() {
   const client = this;
-  const asyncConnection = new Bluebird(function(resolver, rejecter) {
+  const asyncConnection = new Bluebird(async function(resolver, rejecter) {
     // If external authentication dont have to worry about username/password and
     // if not need to set the username and password
-    const oracleDbConfig = client.connectionSettings.externalAuth
-      ? { externalAuth: client.connectionSettings.externalAuth }
+    const connectionSettings = await client
```
maximelkin: do not use async inside new promise callback
it will not catch errors, so you get unhandledRejection

lib/dialects/mssql/index.js
maximelkin: >externalAuth and connectString should be sent to the getConnection
add await in test 
```
-     knexInstance.client
+    await knexInstance.client
      .acquireRawConnection()
      .then(function(resolve) {}, function(reject) {});
```

Ali-Dalal: @maximelkin Which test in which file?

maximelkin:
https://github.com/eslint/eslint/issues/11899
just mark rule require-atomic-updates as ignored, it will be removed shortly due to false-positives

Ali-Dalal:
@kibertoad @maximelkin guys can you help in the last error from eslint?

check travis for more information

`106:5  error  Possible race condition: `client.driver.fetchAsString` might be reassigned based on an outdated value of `client.driver.fetchAsString`  require-atomic-updates`

kibertoad:
@Ali-Dalal as @maximelkin has mentioned, "just mark rule require-atomic-updates as ignored, it will be removed shortly due to false-positives*

Ali-Dalal:
@kibertoad @maximelkin done.

can you guide me where is the best place to write a unit test for testing async config provider?

Ali-Dalal:
@kibertoad I will write unit tests under `unit/dialects` for every one

lib/dialects/mssql/index.js
```diff
@@ -217,8 +217,11 @@ Object.assign(Client_MSSQL.prototype, {
   // Get a raw connection, called by the `pool` whenever a new
   // connection needs to be added to the pool.
   acquireRawConnection() {
-    return new Bluebird((resolver, rejecter) => {
-      const settings = Object.assign({}, this.connectionSettings);
+    const client = this;
+    return new Bluebird(async (resolver, rejecter) => {
+      const settings = await client
+        .connectionConfigProvider()
```
kibertoad: So in this line we are still always initializing a new ConnectionConfigProvider, meaning we will resolve credentials again on each connection.
Can we adjust `connectionConfigProvider` method to lazy-instantiate and not reset each time?

lib/dialects/oracledb/index.js
kibertoad: @Ali-Dalal This needs to be addressed

lib/dialects/mssql/index.js
Ali-Dalal: When a new `knex` instance defined, a new `client` is attached and a new `ConnectionConfoProvider` as well and once we call resolve connection, for the first time it will execute the function if the connection is a function then it will mutate the state of the client and set the connection to the resolved one.

I tried it on my machine and the connection is resolved only one time and for the 2nd time it loads the config from 
```
async resolveConnectionConfig() {
    if (this.resolvedConnectionSettings) return this.resolvedConnectionSettings;
```
https://github.com/knex/knex/pull/3403/files#diff-cbc1ae6887774cf34c8d4bc73dece12fR10

> Can we adjust connectionConfigProvider() method to lazy-instantiate and not reset each time?

can you give some examples how I can solve this?

kibertoad: I think you are seeing connection being resolved only once because it is being reused from the connection pool and not recreated on every request - for new connections it is likely that same `acquireRawConnection` will be called again and you will have to resolve credentials again.

I'll take a swing at the branch sometime this week to see how to improve the situation.

elhigu: Easy way to check with multiple connections is to create transactions and never commit / rollback them. In that way the old connections cannot be reused from pool.

oranoran: Guys, sorry for coming in late in the process, but I noticed this PR when I was considering opening a similar issue, with the use case in mind being support of [Amazon RDS IAM authentication](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.AWSCLI.PostgreSQL.html).

For this use case though, the resolved configuration can't be cached indefinitely, because the password needs to be changed every 15 minutes.

Can we please add an optional field to the returned configuration object, determining whether it should be cached or not?

```javascript
// Something like:
{
    host : '127.0.0.1',
    user : 'your_database_user',
    database : 'myapp_test'
    cacheResolvedConfig: true
}

// Or even more powerful, so we'll be able to cache for exactly 15 minutes:
{
    host : '127.0.0.1',
    user : 'your_database_user',
    database : 'myapp_test'
    cacheResolvedConfig: (config) => config.awsTokenEnd >= Date.now()
}
```

What do you think?

kibertoad:
Superseded by #3497

PR #3406, opened 2019-08-22
Fix return all columns when use returning('*') or returning('users.*)

danielhuynh9051:


test/integration/builder/additional.js
```diff
@@ -190,6 +190,23 @@ module.exports = function(knex) {
             ]);
           });
       });
+
+      // It was needed auto increment ID
+      // it('should return the all columns when single properties as * are given to returning', () => {
```
kibertoad: should this be commented out?

kibertoad:
@danielhuynh9051 Could you please describe problem that you are having and why you think this change is necessary?

test/integration/builder/additional.js
danielhuynh9051: This feature will display all columns and values. In case, unit test required exactly response data include auto increasement ID. I can't guess this ID so I commented my case.

danielhuynh9051:
@kibertoad current, I can't use returning('*') on PostgreSQL (DigitalOcean)
`wrapIdentifier` function replaced '*' by "" so PostgreSQL was occurred error.

`error: zero-length delimited identifier at or near """"`

danielhuynh9051:
Oops. I think it's my mistake. When I convert to snakeCase lodash removed *.
`
wrapIdentifier: (value, origImpl, queryContext) => {
        return origImpl(_.snakeCase(value));
}
`
 So I will close this pull request.

erkinsergey:
Work on PostgreSQL and knex 0.95.6:
```
knex(table)
  .insert()
  .returning(knex.raw('*'))
```

Issue #3405, opened 2019-08-21
TypeScript with strict mode and null defaults


DataTables:
# Environment

Knex version: 0.19.1
Database + version: N/A
OS: N/A

# Feature discussion / request

1 - Explain what is your use case
`ColumnBuilder.defaultTo()` typing does not accept `null` as a valid value.

I'm using TypeScript with `strict: true` and `table.text('styling').defaultTo(null);` gives a TypeScript error:

> Argument of type 'null' is not assignable to parameter of type 'Value'.ts(2345)

2 - Explain what kind of feature would support this
Using TypeScript in strict mode.

3 - Give some API proposal, how the feature should work
Can we just add `null` to the [`Value` type](https://github.com/tgriesser/knex/blob/4125855190cc1fb3a0dc2cbb0b258cde40d6d312/types/index.d.ts#L367)?

Happy to send a PR with that if thought suitable.

DataTables:
Workaround is currently to use `.defaultTo(null as any);`

lorefnon:
Yes, taking a closer look our current usage of `null` is inconsistent.  In some places we do `Value | null` and in some places we don't. 

We can add `null` in `Value` type, but we may need to remove it in some cases like [whereBetween](https://github.com/tgriesser/knex/blob/4125855190cc1fb3a0dc2cbb0b258cde40d6d312/types/index.d.ts#L1085).

DataTables:
Good point! I'd forgotten about things like `whereBetween`. Without introducing a `ValueOrNull` type its hard to see how it could be done other than `Value | null` since excluding types isn't the easiest thing in TS.

I've created #3407 to just use `Value | null` for `defaultTo()` at the moment.

PR #3407, opened 2019-08-22
#3405 Allow `ColumnBuilder.defaultTo()` to be `null`

DataTables:
Augment the `defaultTo` method's type to allow `null` which is required when setting a column's default to be `null` with TypeScript's strict mode enabled.


kibertoad:
Thanks!

Issue #3390, opened 2019-08-08
CLI: list all pending migrations


Saka7:
# Feature request: Knex CLI list all pending migrations

Are there any plans for `list` command? The idea is to print all migrations which haven't been applied yet.

API Example: 
```sh
knex list

20190809115505_migration10
20190809115506_migrations11
20190809115507_migrations12
```
References:
- [sqitch](https://github.com/sqitchers/sqitch) - `sqitch status`
- [node-migrate](https://github.com/tj/node-migrate) - `migrate list`

kibertoad:
@Saka7 Sounds useful! We would gladly accept PR for this.

kibertoad:
Released in 0.19.3

MohamedLamineAllal:
it's no more available ! Why was it removed !

PR #3409, opened 2019-08-22
knex cli list all migrations with status (#3390)

vithalreddy:
Add's `list` or `status` command to knex cli to show both completed and pending migrations.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -19,9 +19,9 @@ const {
   checkLocalModule,
   getMigrationExtension,
   getStubPath,
-  print_migrations_list,
 } = require('./utils/cli-config-utils');
 const { DEFAULT_EXT } = require('./utils/constants');
+const { migrationsLister } = require('./utils/migrationsLister');
 
 function initKnex(env, opts) {
   checkLocalModule(env);
@@ -276,7 +276,7 @@ function invoke(env) {
       pending = initKnex(env, commander.opts())
         .migrate.list()
         .then(([completed, newMigrations]) => {
-          print_migrations_list(completed, newMigrations);
+          migrationsLister(completed, newMigrations);
         })
         .catch(exit);
     });
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -19,6 +19,7 @@ const {
   checkLocalModule,
   getMigrationExtension,
   getStubPath,
+  print_migrations_list,
 } = require('./utils/cli-config-utils');
 const { DEFAULT_EXT } = require('./utils/constants');
 
@@ -275,35 +276,7 @@ function invoke(env) {
       pending = initKnex(env, commander.opts())
         .migrate.list()
         .then(([completed, newMigrations]) => {
-          let message = '';
-
-          if (completed.length === 0) {
-            message += color.red('No Completed Migration files Found. \n');
-          } else {
-            message = color.green(
-              `Found ${completed.length} Completed Migration file/files.\n`
-            );
-
-            for (let i = 0; i < completed.length; i++) {
-              const file = completed[i];
-              message += color.cyan(`${file} \n`);
-            }
-          }
-
-          if (newMigrations.length === 0) {
-            message += color.red('No Pending Migration files Found.\n');
-          } else {
-            message += color.green(
-              `Found ${newMigrations.length} Pending Migration file/files.\n`
-            );
-
-            for (let i = 0; i < newMigrations.length; i++) {
-              const file = newMigrations[i];
-              message += color.cyan(`${file.file} \n`);
-            }
-          }
-
-          success(message);
+          print_migrations_list(completed, newMigrations);
         })
         .catch(exit);
     });
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -269,8 +269,8 @@ function invoke(env) {
     });
 
   commander
-    .command('list')
-    .alias('status')
+    .command('migrate:list')
+    .alias('migrate:status')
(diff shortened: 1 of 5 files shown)


bin/cli.js
```diff
@@ -267,6 +267,47 @@ function invoke(env) {
         .catch(exit);
     });
 
+  commander
+    .command('list')
+    .alias('status')
+    .description('        List all migrations files with status.')
+    .action(() => {
+      pending = initKnex(env, commander.opts())
```
kibertoad: can this use async/await instead of promise chain?

bin/cli.js
```diff
@@ -267,6 +267,47 @@ function invoke(env) {
...
+        .migrate.list()
+        .then(([completed, newMigrations]) => {
```
kibertoad: Would be great if you could extract whole logic for this operation into a separate file and only call it from cli.js

lib/migrate/Migrator.js
```diff
@@ -219,6 +219,27 @@ class Migrator {
       });
   }
 
+  // list all migrations
+  list(config) {
+    this._disableProcessing();
+    this.config = getMergedConfig(config, this.config);
+
+    return migrationListResolver
```
kibertoad: could you please use async/await instead?

test/jake/jakelib/migrate-test.js
```diff
@@ -543,6 +543,98 @@ test('migrate:down undos only the last run migration', (temp) => {
...
+    `node ${KNEX} list \
+    --client=sqlite3 \
+    --connection=${temp}/db \
+    --migrations-directory=${temp}/migrations`
+  )
+    .then(({ stdout }) => {
+      assert.include(stdout, `No Completed Migration files Found.`);
+      assert.include(stdout, `No Pending Migration files Found.`);
+
+      fs.writeFileSync(
```
kibertoad: using FileHelper with auto-cleanup is preferable for preparing files. see https://github.com/tgriesser/knex/blob/master/test/cli/migrate-make.spec.js

test/jake/jakelib/migrate-test.js
```diff
@@ -543,6 +543,98 @@ test('migrate:down undos only the last run migration', (temp) => {
...
+
+      return assertExec(
+        `node ${KNEX} migrate:up \
+        --client=sqlite3 \
+        --connection=${temp}/db \
+        --migrations-directory=${temp}/migrations`,
+        'create_books_table'
+      );
+    })
+    .then(({ stdout }) => {
```
kibertoad: async/await is preferable

kibertoad:
Much appreciated! Can you also create a documentation PR in https://github.com/knex/documentation?

vithalreddy:
@kibertoad Thanks for the review, Will do the required changes.

bin/utils/cli-config-utils.js
```diff
@@ -95,6 +95,38 @@ function success(text) {
   process.exit(0);
 }
 
+function print_migrations_list(completed, newMigrations) {
```
kibertoad: I wouldn't say that config-utils is a good place for this particular set of logic, I would create a new file "migrationsLister". Also please use camelCase for function name ;)

test/jake/jakelib/migrate-test.js
```diff
@@ -543,6 +543,100 @@ test('migrate:down undos only the last run migration', (temp) => {
     });
 });
 
+test('list prints migrations both completed and pending', async (temp) => {
+  const migrationFile1 = '001_create_animals_table.js';
+  const migrationFile2 = '002_add_age_column_to_animals_table.js';
+
+  let { stdout } = await assertExec(
```
kibertoad: does this need to be `let`? I assume we don't overwrite it anywhere.

vithalreddy: It is getting used below as well.

bin/utils/migrationsLister.js
```diff
@@ -0,0 +1,37 @@
+const color = require('colorette');
+
+const { success } = require('./cli-config-utils');
+
+function migrationsLister(completed, newMigrations) {
```
kibertoad: Sorry for the constant stream of requests. Could you please rename function to be a verb - "listMigrations"? File name is fine.

vithalreddy: No Problem at all.
Will rename and push the changes.

test/jake/jakelib/migrate-test.js
kibertoad: Can we just have to separate instances of stdout then? Reassigning variables in this fashion is confusing and is better avoided.

kibertoad:
@vithalreddy Thanks for doing the changes! This can be merged after documentation is created.

vithalreddy:
Thanks. Will add the documentation and submit PR for the same.

bin/cli.js
```diff
@@ -267,6 +268,19 @@ function invoke(env) {
         .catch(exit);
     });
 
+  commander
+    .command('list')
```
kibertoad: can you change it to be `migrate:list` for consistency?

bin/cli.js
```diff
@@ -267,6 +268,19 @@ function invoke(env) {
...
+    .alias('status')
```
kibertoad: migrate:status as alias

kibertoad:
Thanks!

vithalreddy:
@kibertoad Thanks for accepting and reviewing my pr.

PR #3410, opened 2019-08-23
Add SSL/TLS support for Postgres connection string

### LINKED ISSUE #852: SSL Query Parameter for Postgres
It seems like #516 has surfaced again. When I use a config object it seems to work fine with the `ssl` option, however this doesn't work with a connection string:

Works:

``` node
connection: {
  host: "host",
  port: "port",
  user: "user",
  password: "password",
  database: "database",
  ssl: true
 }
```

Does not work:

``` node
connection: "postgres://user:password@host:port/database?ssl=true"
```

hjr3:
- Upgrade pg-connection-string to 2.1.0

Fixes #852

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "liftoff": "3.1.0",
     "lodash": "^4.17.15",
     "mkdirp": "^0.5.1",
-    "pg-connection-string": "2.0.0",
+    "pg-connection-string": "2.1.0",
     "tarn": "^2.0.0",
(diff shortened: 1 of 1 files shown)


hjr3:
I released [pg-connection-string v2.1.0](https://github.com/iceddev/pg-connection-string/releases/tag/v2.1.0) which fixes the issues with ssl connection string. I tested this on my local fork.

kibertoad:
Thanks!

Issue #2763, opened 2018-08-12
SQlite3 backsticks/double quotes problem


kirrg001:
# Environment

Knex version: 0.12/0.14
Database + version: SQlite3 4.0.1
OS: OSX

Select applicable tempalate from below.
If issue is about oracledb support tag @ atiertant. For MSSql tag @ smorey2 .
Rest of dialects doesn't need tags.

# Questions about how to use knex

I have a SQlite3 database with double quotes and it was created with knex 0.12.
Then i've switched to knex 0.14 and executed `renameColumn`.
It crashed, because  [this line](https://github.com/tgriesser/knex/blob/v0.15.1/src/dialects/sqlite3/schema/ddl.js#L146) is always false and it doesn't create the table with the new table, always with the old.
It is always false, because e.g. "column" !== `column`.

Is there a way to migrate my SQlite3 database file? I assume knex won't be backwards compatible? But i guess many people could run into this.

kirrg001:
For reference: https://github.com/tgriesser/knex/pull/2087

tgriesser:
Blah I'm sorry @ErisDS and @kirrg001 I'm just seeing this now, been a busy couple of weeks for me. I am going to look through and see if I can get this fixed asap. I am seeing that Ghost was bumped to 2.0 yesterday, did you end up doing something else to fix it?

kirrg001:
> I am going to look through and see if I can get this fixed asap. 

That sounds amazing. Thanks so much!

> I am seeing that Ghost was bumped to 2.0 yesterday, did you end up doing something else to fix it?

Yeah, i've applied a quick fix, [see here](https://github.com/kirrg001/knex/commit/bce2fd7f6dd29e2fb4d203922ba6a807e3f95b22). Not sure it's 100% correct, but it's working fine for our use case at the moment 😇

elhigu:
Looks like we need to go through all the code and look for places where there has been used hardcoded quotes like here: https://github.com/tgriesser/knex/blob/v0.15.1/src/dialects/sqlite3/schema/ddl.js#L56

elhigu:
I'm surprised that test hadn't caught these... sounds like renaming parts will need more integration testing.

kirrg001:
@elhigu Any news on this? :)

I am running into trouble having two versions of knex in one project.

e.g. 

- Ghost uses knex 0.14.6 in our model layer
[...]

PR #3411, opened 2019-08-24
Fix handling of multiline SQL in SQLite3 schema

6uliver:
**Bug:** 
Multiline SQL parsed incorrectly in SQLite3 schema when using renameColumn or dropColumn.

Knex version: 0.19.2
Sqlite3 version: 4.1.0
Node.js version: 10.15.3

**Use case:**
I have a legacy SQLite3 database in production which has been created manually with DB Browser for SQLite (https://sqlitebrowser.org/) so I decided to use Knex migration capabilities from now on.
The problem is that the create statement in sqlite_master table is a multiline SQL (which has been created by the DB Browser and is absolutely valid) but the SQL parser in DDL implementation does not expect for it.

**Repro:**
1. Create an SQLite3 database and a table having some columns with DB Browser so the create statement contains newlines and tabs:
Screenshot 2019-08-24 at 10 29 21

```
CREATE TABLE "TestTable" (
	"id"	INTEGER,
	"description"	TEXT,
	PRIMARY KEY("id")
);
```

[test.db.zip](https://github.com/tgriesser/knex/files/3536999/test.db.zip)

2. Write a migration script which tries to rename a column or drop it.

[20190824103207_add_and_rename_column.js.zip](https://github.com/tgriesser/knex/files/3537003/20190824103207_add_and_rename_column.js.zip)

**ER:**
Columns are renamed/dropped

**AR:**
Migration script failes with the following error:
```
TypeError: Cannot read property '1' of null
    at SQLite3_DDL._doReplace (***/node_modules/knex/lib/dialects/sqlite3/schema/ddl.js:134:30)
...
```

Should you have any question or suggestion don't hesitate to ask, I will make the necessary changes quickly, so that you can merge it as soon as possible.


kibertoad:
How do you use it? By calling knex.raw?

6uliver:
> How do you use it? By calling knex.raw?

No, I'm not calling knex.raw, just using a simple migration script as you can see in the attached file:

```
exports.up = function(knex) {
  return knex.schema.table('TestTable', function (table) {
    table.dropColumn('description');
    table.renameColumn('id', 'testId');
  })
};
```

kibertoad:
I'm really confused then. If your script is valid javascript, why any changes are needed? I don't see any multiline SQL logic in example you gave, and without full integration test I don't see the real use-case.

6uliver:
> I'm really confused then. If your script is valid javascript, why any changes are needed? I don't see any multiline SQL logic in example you gave, and without full integration test I don't see the real use-case.

The multiline SQL is coming from the database file since SQLite3 is storing schema as a create statement in sqlite_master table. The multiline SQL create statement added by the DB Browser tool because my database is a legacy one and have not been created with a migration script from the beginning but by hand. DDL class is finding table schema with the getTableSql method which is returning the multiline SQL.

I kindly ask you to read again my original comment carefully, try out the migration script and the DB attached, run and inspect the failing tests separately commited in 82a46a0cac981fd20e8fd4750ad963d744d5e1a8

kibertoad:
@6uliver I will. However, given that this is not a typical use-case and current unit test covers non-self-explanatory case, full integration test is definitely needed in this case that would show that solution is working end to end.

6uliver:
@kibertoad ok, thanks for your quick response. I will add some end to end integration tests soon.

test/integration/migrate/index.js
```diff
@@ -72,6 +72,22 @@ module.exports = function(knex) {
       });
     }
 
+    if (knex.client.driverName === 'sqlite3') {
+      it('should not fail rename-and-drop-column with multiline sql from legacy db', () => {
+        return knex.migrate
```
kibertoad: Could you please use async/await instead?

test/integration/index.js
```diff
@@ -4,18 +4,23 @@ const knex = require('../../knex');
...
   return require('./suite')(logger(knex(config[dialectName])));
 });
 
-after(function(done) {
+before(function(done) {
   if (config.sqlite3 && config.sqlite3.connection.filename !== ':memory:') {
-    fs.unlink(config.sqlite3.connection.filename, function() {
-      done();
-    });
+    fs.copyFile(
```
kibertoad: Is this necessary for all tests in this suite?

kibertoad: Synchronous version would be preferable here, we don't really care about performance in this case.

test/integration/migrate/index.js
```diff
@@ -72,6 +72,22 @@ module.exports = function(knex) {
       });
     }
 
+    if (knex.client.driverName === 'sqlite3') {
+      it('should not fail rename-and-drop-column with multiline sql from legacy db', () => {
```
kibertoad: Probably sample legacy db that you are using should be only referenced within this test

6uliver:
@kibertoad I've added an end to end test. 

At the start of the test run for SQLite3 we copy sample.sqlite3 to test.sqlite3 and it contains a test table which has been created with DB Browser for SQLite (this causes the bug because it contains the multiline expression). The new test case has two migration script, one for renaming a column and the other for dropping the column. 

I've patched the double quote backward compatibility regex from `/CREATE\sTABLE\s".*"\s\("/` to `/CREATE\sTABLE\s".*"\s\(\s*"/` because after the parentheses the identifier starts in a newline so I added \s*.

test/integration/migrate/index.js
6uliver: Sure.

test/integration/migrate/index.js
6uliver: I tried to change the connection and file in config during this case, but I can not figure out how to change change Knex connection config after instantiation. Can you suggest how can I do it?

test/integration/index.js
```diff
@@ -4,18 +4,23 @@ const knex = require('../../knex');
...
+      path.resolve(__dirname, '../sample.sqlite3'),
```
kibertoad: Considering that this is a very specific db sample, can name be made more clear? E. g. "multilineCreateMaster.sqlite3"

test/integration/index.js
6uliver: We can reuse this copied database for other suites too, they created the test.sqlite3 anyway if it does not exist.

kibertoad: But it's a very specific DB for very specific case, why would anyone want to reuse it for unrelated test? And ideally tests should be as isolated as possible, to avoid breaking lost of tests when making unrelated change to test resources.

test/integration/migrate/index.js
kibertoad: Worst case scenario, you can instantiate a new Knex within this test, it is sqlite-specific anyway so multi-db support doesn't matter here. It is going to be a problem if we ever support alternate sqlite driver, so just to be on a safe side here, I would add `if (knex.client.driverName !== 'sqlite3') throw new Error('Please implement alternate instantiation logic for this driver')'` next to Knex instantiation to protect against someone adding alternate supported driverName blindly.

6uliver: I changed it as you requested, we use this special DB file only in this test. I had to make a copy before the test run and delete it after because knex migration tables changes, so the original file becomes changed in Git for every test run.

test/integration/index.js
6uliver: Yes, I agree absolutely, tests should be isolated from each other. I changed the code, see my other comment.

test/integration/index.js
6uliver: Done.

6uliver:
I fixed the code related to your comments. 

I had to change the identifier replacing logic too, because there were already some logic for backward compatibility related to double quotes and backsticks. I implemented this as a way to replace identifier style to the one which was given as the "to" parameter. The "from" parameter is stripped to get the identifier name without quotes and backsticks and replace it to "to" parameter as it received.

Examples:
in SQL, from, to, result
"id", "id", "newId" -> "newId"
"id", `id`, "newId" -> "newId"
"id", "id", `newId` -> `newId`
"id", `id`, `newId` -> `newId`

6uliver:
Travis failed on Node.js 8 because two Oracle DB migration tests failed with a timeout error: TimeoutError: Knex: Timeout acquiring a connection. The pool is probably full.

I don't think it's related with this PR, can you restart the build?

test/integration/index.js
```diff
@@ -5,18 +5,22 @@ const logger = require('./logger');
...
-
 Object.keys(config).forEach((dialectName) => {
   return require('./suite')(logger(knex(config[dialectName])));
 });
 
-after(function(done) {
+before(function() {
+  if (config.sqlite3 && config.sqlite3.connection.filename !== ':memory:') {
+    fs.copyFileSync(
+      __dirname + '/../multilineCreateMasterSample.sqlite3',
```
kibertoad: what's the difference between the two files?

test/integration/index.js
```diff
@@ -5,18 +5,22 @@ const logger = require('./logger');
...
-const Bluebird = require('bluebird');
-
 Object.keys(config).forEach((dialectName) => {
   return require('./suite')(logger(knex(config[dialectName])));
 });
 
-after(function(done) {
+before(function() {
+  if (config.sqlite3 && config.sqlite3.connection.filename !== ':memory:') {
+    fs.copyFileSync(
```
kibertoad: do we need to do the copying for the whole suite and not just one test?

kibertoad:
@kirrg001 Could you please check if changes in SQLite identifier name escaping is not going to break things for you?

kibertoad:
Related: #2763

test/integration/index.js
6uliver: I was thinking about copying in the test body but if the test fails it will not remove the file. We could copy in the test body and delete in the teardown function but I think it's better if they are in the same place.

test/integration/index.js
6uliver: I had to make a copy before the test run and delete it after because knex migration tables changes, so the DB file becomes changed in Git for every test run.

6uliver:
@kibertoad do you have any update on this? @kirrg001 seems to be inactive for a while.

kibertoad:
@6uliver I'm seeing this error on master now:
```
  1) Integration Tests
       "after all" hook in "Integration Tests":
     Error: EBUSY: resource busy or locked, unlink 'C:\sources\knex\test/test.sqlite3'
```

Do you think it is related to your changes?

kibertoad:
Released in 0.19.5

PR #3413, opened 2019-08-25
add ability to manually define schema for enu with useNative (#3307)

ivanovych666:
# Fixes issue: #3307
## Improve https://github.com/tgriesser/knex/pull/3400

There is no way to define schema manually, the schema name from withSchema will be always prepended to enumName.

#### Code:
```javascript
knex
      .schemaBuilder()
      .withSchema('test')
      .table('users', function(table) {
        table
          .enu('foo', ['bar', 'baz'], {
            useNative: true,
            enumName: 'type_schema.foo_bar',
          })
      })
      .toSQL();
```

#### Outputs:

```sql
-- ...
create type "test"."type_schema"."foo_bar" as enum ('bar', 'baz');
alter table "test"."users" add column "foo" "test"."type_schema"."foo_bar";
-- ...
```

### API proposal:
If enumName includes a dot, then don't prepend the schema name from withSchema to enumName.

#### Code:
```javascript
knex
      .schemaBuilder()
      .withSchema('test')
      .table('users', function(table) {
        table
          .enu('foo', ['bar', 'baz'], {
            useNative: true,
            existingType: true,
            enumName: 'type_schema.foo_bar'
          })
      })
      .toSQL();
```

#### Outputs:
```sql
-- ...
alter table "test"."users" add column "foo" "type_schema"."foo_bar";
-- ...
```

#### Code:
```javascript
knex
      .schemaBuilder()
      .withSchema('test')
      .table('users', function(table) {
        table
          .enu('foo', ['bar', 'baz'], {
            useNative: true,
            enumName: 'type_schema.foo_bar'
          })
      })
      .toSQL();
```

#### Outputs:
```sql
-- ...
create type "type_schema"."foo_bar" as enum ('bar', 'baz');
alter table "test"."users" add column "foo" "type_schema"."foo_bar";
-- ...
```

### DIFF
--- a/lib/dialects/postgres/schema/columncompiler.js
+++ b/lib/dialects/postgres/schema/columncompiler.js
@@ -36,7 +36,7 @@ Object.assign(ColumnCompiler_PG.prototype, {
 
     if (options.useNative) {
       let enumName = '';
-      const schemaName = this.tableCompiler.schemaNameRaw;
+      const schemaName = options.schemaName || this.tableCompiler.schemaNameRaw;
 
       if (schemaName) {
         enumName += `"${schemaName}".`;
--- a/test/unit/schema/postgres.js
+++ b/test/unit/schema/postgres.js
@@ -964,6 +964,58 @@ describe('PostgreSQL SchemaBuilder', function() {
     );
   });
 
+  it('adding enum with useNative, from manually defined schema and withSchema', function() {
+    const tableSchema = 'table_schema';
+    const tableName = 'table_name';
+    const typeSchema = 'type_schema';
+    const typeName = 'type_name';
+    const columnName = 'column_name';
+
+    tableSql = client
+      .schemaBuilder()
+      .withSchema(tableSchema)
+      .table(tableName, function(table) {
+        table.enu(columnName, ['foo', 'bar', 'baz'], {
+          useNative: true,
+          schemaName: typeSchema,
+          enumName: typeName,
+        });
+      })
+      .toSQL();
+    equal(2, tableSql.length);
+    expect(tableSql[0].sql).to.equal(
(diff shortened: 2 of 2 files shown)


kibertoad:
@ivanovych666 Wait, now I see what you mean - sometimes your table is in one schema (that is coming from `withSchema`, and your enums might very well be coming from the other schema, and it's actually not redundant, since it prevents you from combining the two).
In such case I think it would be confusing to silently ignore withSchema in case of some inputs, so explicit `schema` param is, indeed, better.

Could you please create a PR that would revert https://github.com/tgriesser/knex/pull/3400/commits/b29b3bdade2834701ff445bd6e085c0bdd55e2c5 instead?

ivanovych666:
@kibertoad , so the behavior will be the next:
- If the explicit `schema` is defined in `enu` options, then use it.
- If the schema from `withSchema()` is defined then use it.
- Else no schema should be used (means "public" schema).

Please confirm if this works.

Also, just as a suggestion, maybe instead of `schema` it is better to use `withSchema` parameter in `enu` options, to be more consistency? For example:

```javascript
knex
      .schemaBuilder()
      .withSchema('test')
      .table('users', function(table) {
        table
          .enu('foo', ['bar', 'baz'], {
            useNative: true,
            enumName: 'foo_bar',
            withSchema: 'type_schema'
          })
      })
      .toSQL();
```

kibertoad:
@ivanovych666 Yeah, logic sounds correct.
Regarding naming - actually, to be completely consistent, `schemaName` would be the best option; we only use `withSchema` as an operationm which is consistent with other usages (`with`, `withRecursive`), but we use `schemaName` as option name elsewhere.

kibertoad:
See comment about naming 😁. Change itself looks good, thanks!

ivanovych666:
Sorry, I saw it but made a mistake, I have updated the PR.

PR #3415, opened 2019-08-26
TypeScript definition: include schemaName in EnumOptions

ivanovych666:
This is addition to https://github.com/tgriesser/knex/pull/3413


kibertoad:
@ivanovych666 Can you mark optional ones as optional, while we are at it, btw?

kibertoad:
Awesome, thank you!

PR #3416, opened 2019-08-26
Migrations up/down commands: filename parameter

ivanovych666:
Add the ability to explicitly define the migration's filename for `up` and `down` commands (#1550).

### Current CLI
Run only the next migration:
```bash
migrate:up
```

Undo only the latest migration:
```bash
migrate:down
```

### Proposed CLI:
Add the optional `name` parameter for `up` and `down` commands.

Run only the specified or the next migration:
```bash
knex migrate:up <name>
```

Undo only the specified or the latest migration:
```bash
knex migrate:down <name>
```

#### Example 1:
Let's say we have two new migrations:
```
001_one.js
002_two.js
```
If we want to run only the `002_two.js` then we can use this:
```bash
knex migrate:up 002_two.js
```

#### Example 2:
Let's say we have two completed migrations:
```
001_one.js
002_two.js
```
If we want to revert only the `001_one.js` then we can use this:
```bash
knex migrate:down 001_one.js
```

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -190,11 +190,13 @@ function invoke(env) {
     });
 
   commander
-    .command('migrate:up')
-    .description('        Run the next migration that has not yet been run.')
-    .action(() => {
+    .command('migrate:up [<name>]')
+    .description(
+      '        Run the next or the defined migration that has not yet been run.'
+    )
+    .action((name) => {
       pending = initKnex(env, commander.opts())
-        .migrate.up()
+        .migrate.up(name)
         .then(([batchNo, log]) => {
           if (log.length === 0) {
             success(color.cyan('Already up to date'));
@@ -235,11 +237,11 @@ function invoke(env) {
     });
 
   commander
-    .command('migrate:down')
-    .description('        Undo the last migration performed.')
-    .action(() => {
+    .command('migrate:down [<name>]')
+    .description('        Undo the last or the defined migration performed.')
+    .action((name) => {
       pending = initKnex(env, commander.opts())
-        .migrate.down()
+        .migrate.down(name)
         .then(([batchNo, log]) => {
           if (log.length === 0) {
             success(color.cyan('Already at the base migration'));
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -196,7 +196,7 @@ function invoke(env) {
     )
     .action((name) => {
       pending = initKnex(env, commander.opts())
-        .migrate.up(name)
+        .migrate.up({ name })
         .then(([batchNo, log]) => {
           if (log.length === 0) {
             success(color.cyan('Already up to date'));
@@ -243,7 +243,7 @@ function invoke(env) {
     )
     .action((name) => {
       pending = initKnex(env, commander.opts())
-        .migrate.down(name)
+        .migrate.down({ name })
         .then(([batchNo, log]) => {
           if (log.length === 0) {
             success(color.cyan('Already at the base migration'));
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -192,7 +192,7 @@ function invoke(env) {
   commander
     .command('migrate:up [<name>]')
     .description(
-      '        Run the next or the defined migration that has not yet been run.'
+      '        Run the next or the specified migration that has not yet been run.'
     )
     .action((name) => {
       pending = initKnex(env, commander.opts())
@@ -238,7 +238,9 @@ function invoke(env) {
 
   commander
     .command('migrate:down [<name>]')
-    .description('        Undo the last or the defined migration performed.')
+    .description(
+      '        Undo the last or the specified migration that was already run.'
+    )
     .action((name) => {
       pending = initKnex(env, commander.opts())
         .migrate.down(name)
--- a/lib/migrate/Migrator.js
+++ b/lib/migrate/Migrator.js
@@ -97,7 +97,7 @@ class Migrator {
   }
 
   // Runs the next migration that has not yet been run
-  up(name, config) {
+  up(config) {
     this._disableProcessing();
     this.config = getMergedConfig(config, this.config);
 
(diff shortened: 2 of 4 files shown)


bin/cli.js
```diff
@@ -235,11 +237,11 @@ function invoke(env) {
     });
 
   commander
-    .command('migrate:down')
-    .description('        Undo the last migration performed.')
-    .action(() => {
+    .command('migrate:down [<name>]')
+    .description('        Undo the last or the defined migration performed.')
```
kibertoad: `specified` is likely to be clearer than `defined`

bin/cli.js
```diff
@@ -190,11 +190,13 @@ function invoke(env) {
     });
 
   commander
-    .command('migrate:up')
-    .description('        Run the next migration that has not yet been run.')
-    .action(() => {
+    .command('migrate:up [<name>]')
+    .description(
+      '        Run the next or the defined migration that has not yet been run.'
```
kibertoad: `specified` is likely to be clearer than `defined`

bin/cli.js
kibertoad: `that was already run` might be more consistent with woring of `up` command

lib/migrate/Migrator.js
```diff
@@ -108,16 +108,24 @@ class Migrator {
...
       })
       .then(([all, completed]) => {
-        const migrationToRun = getNewMigrations(
+        const newMigrations = getNewMigrations(
           this.config.migrationSource,
           all,
           completed
-        ).slice(0, 1);
+        );
+        const migrationToRun = name
```
kibertoad: Can you also throw an error if name was specified but migration was not found?

kibertoad:
Thanks! Can you also create a documentation PR at https://github.com/knex/documentation?

bin/cli.js
ivanovych666: done

bin/cli.js
ivanovych666: done

ivanovych666: done

lib/migrate/Migrator.js
ivanovych666: done, throws an error in `up` and in `down` methods

test/jake/jakelib/migrate-test.js
```diff
@@ -543,6 +589,60 @@ test('migrate:down undos only the last run migration', (temp) => {
...
+    });
+  });
+});
+
+test('migrate:down <name> throw an error', (temp) => {
+  const migrationsPath = `${temp}/migrations`;
+  const migrationFile1 = '001_one.js';
+
+  return assertExec(
+    `node ${KNEX} migrate:down ${migrationFile1} \
```
kibertoad: assertExecError is a better way to test negative scenarios

ivanovych666: Thank you, you are right. Done.

kibertoad:
@ivanovych666 Tests are failing now 😹

ivanovych666:
@kibertoad , (I hope) I have fixed integration tests by adding `null` as the first argument in `up` and `down` calls. But if we need to make this backward compatible, then in these methods we can use something like this:
```javascript
if (!config && name && typeof name === 'object') {
    config = name;
    name = null;
}
```
What do you think about this?

ivanovych666:
@kibertoad Is there any chance to have this pull request approved? If there is something more that I can do then please let me know.

lib/migrate/Migrator.js
```diff
@@ -169,7 +190,7 @@ class Migrator {
     });
   }
 
-  down(config) {
+  down(name, config) {
```
kibertoad: can't the name be part of config to keep this change backwards-compatible?

test/jake/jakelib/migrate-test.js
```diff
@@ -430,6 +433,49 @@ test('migrate:up runs only the next unrun migration', (temp) => {
...
+
+  fs.writeFileSync(`${migrationsPath}/${migrationFile2}`, migrationData);
+
+  return assertExec(
+    `node ${KNEX} migrate:up ${migrationFile2} \
+    --client=sqlite3 \
+    --connection=${temp}/db \
+    --migrations-directory=${migrationsPath}`,
+    'run_migration_002'
+  ).then(({ stdout }) => {
```
kibertoad: it is preferable to use async/await syntax

lib/migrate/Migrator.js
ivanovych666: done

test/jake/jakelib/migrate-test.js
ivanovych666: done

ivanovych666:
I have updated the pull request as per comments above.

ivanovych666:
Is there anything else that I can do here?

test/jake/jakelib/migrate-test.js
kibertoad: was this change pushed?

ivanovych666: You are right, thank you for mention this, I forgot to do this, now this is done. Please, let me know if there anything else.

ivanovych666:
@kibertoad , I'm not sure, but it looks like there is one test failed which should not be related to this pull request. I have made a rebase of my PR on your master, but this didn't help. Do you have any ideas on how to solve this?

kibertoad:
I think this is related to just released PG 12, since we are using the latest version. I'll fix it separately.

kibertoad:
@ivanovych666 New CLI test is failing in master: `object tested must be an array, a map, an object, a set, a string, or a weakset, but undefined given`

Could you please fix it in a separate PR?

ivanovych666:
Oops, sorry for that, new PR is ready here https://github.com/tgriesser/knex/pull/3466

kibertoad:
Released in 0.19.5

PR #3418, opened 2019-08-28
Override --connection if specified through cmd line arguments (#3305)

### LINKED ISSUE #3305: CLI --connection option is difficult to use
# Environment

Knex version: 0.17.3
Database + version: All (but using postgres)
OS: All (but using Ubuntu 18.x)

# Bug

> 1. Explain what kind of behaviour you are getting and how you think it should do

Conditions:

- There exists a knexfile
- And while running `npx knex  --client pg --connection <dsn> migrate:latest`

Result:

- DSN is ignored; even if no connection is specified in knexfile

Expected:

- An option to ignore a knexfile.js if it happens to be in the same path.
- And Either:
  - An error/warning that `connection` option is being ignored
  - Or the --connection argument should override the knexfile

[Here](https://github.com/tgriesser/knex/blob/858c7b6f1e1591075d7f51b647e0083731c2471e/bin/cli.js#L69) is the codepath that shows how hard it is to get --connection argument to work:

<a href="https://github.com/tgriesser/knex/blob/858c7b6f1e1591075d7f51b647e0083731c2471e/bin/cli.js#L69">Illustration of codepath</a>

> 2. Error message

- In the case that the knexfile leaves out the connection: `Error: Unable to acquire a connection`
- In the case that the knexfile has a connection string/object, and/or tries to connect to the db instance specified and fails with a connection error; or worse, it silently fails and connects to the wrong db instance.

> 3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real

N/A

vithalreddy:
override connection, client, and migrations-table-name if specified through cmd line arguments.

[Fixes #3305](https://github.com/tgriesser/knex/issues/3305)

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -65,6 +65,16 @@ function initKnex(env, opts) {
   }
 
   const resolvedConfig = resolveEnvironmentConfig(opts, env.configuration);
+
+  // override connection, client and migrationsTableName
+  // if specified through cmd line arguments
+
+  if (opts.connection) resolvedConfig.connection = opts.connection;
+  if (opts.client) resolvedConfig.client = opts.client;
+  if (opts.migrationsTableName) {
+    resolvedConfig.migrations.tableName = opts.migrationsTableName;
+  }
+
   const knex = require(env.modulePath);
   return knex(resolvedConfig);
 }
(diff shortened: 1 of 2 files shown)


BossCalifornia:
Is this going to happen??

kibertoad:
@BossCalifornia Could you rebase the PR?

BossCalifornia:
@kibertoad I don't have access to @vithalreddy's repo :(( I can try make my own with the same changes?

BossCalifornia:
Done https://github.com/BossCalifornia/knex/tree/pr/3418, should I create a new PR?

kibertoad:
@BossCalifornia yes, please!

BossCalifornia:
@kibertoad done! https://github.com/knex/knex/pull/4001

PR #3435, opened 2019-09-11
Make sqlite3 hasColumn case insensitive

### LINKED ISSUE #3434: Sqlite3 hasColumn should be case insensitive
# Environment

Knex version: 0.16.5
Database + version: sqlite3
OS: mac

# Bug

When runing the function `hasColumn` on sqlite the answer is case sensitive. Meaning that `test` and `Test` are considered as two different columns. 
In reality sqlite3 names are case insensitive. If you use `hasColumn` to check if a column exist and then create it this will cause a `duplicate column error` when you just change the case of one char. 

Happy to make a PR if this makes sense to you guys.

alexandrebodin:
This PR makes the sqlite3 `hasColumn` function case insensitive to respect sqlite3 convention.

Fix #3434.

It might some new tests ? let me know and I'll add the necessary ones.

### DIFF
--- a/test/integration/schema/index.js
+++ b/test/integration/schema/index.js
@@ -1058,12 +1058,13 @@ module.exports = function(knex) {
             return Promise.resolve();
           }
 
-          it('checks whether a column exists without being case sensitive, resolving with a boolean', function() {
-            return knex.schema
-              .hasColumn('accounts', 'FIRST_NAME')
-              .then(function(exists) {
-                expect(exists).to.equal(true);
-              });
+          it('checks whether a column exists without being case sensitive, resolving with a boolean', async () => {
+            const exists = await knex.schema.hasColumn(
+              'accounts',
+              'FIRST_NAME'
+            );
+
+            expect(exists).to.equal(true);
           });
         });
       });
--- a/test/integration/schema/index.js
+++ b/test/integration/schema/index.js
@@ -1048,6 +1048,24 @@ module.exports = function(knex) {
               expect(exists).to.equal(true);
             });
         });
+
+        describe('sqlite only', function() {
+          if (
+            !knex ||
+            !knex.client ||
+            !/sqlite3/i.test(knex.client.driverName)
+          ) {
+            return Promise.resolve();
+          }
+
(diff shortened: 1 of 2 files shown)


ricardograca:
Yes, this needs tests.

alexandrebodin:
Hi @ricardograca, 

I have looked into the `test/integration/schema` to add some tests about the hasColumn. The first thing I noticed is that test set a custom wrapIdentifier that upercases the values hence make the code case-insensitive. But in reality the default implementation doesn't do that. I don't understand why use this custtom wrapIdentifier and If I should remove it.

I might need a few explanations before adding the tests :) 

Thanks !

kibertoad:
@alexandrebodin I think that the reason why these wrappers are used is to test whether wrapping/postprocessing logic works in the first place. If default implementation is updated to do the wrapping by default, existing wrappers would need to be replaced with something else (e. g. applying some kind of prefixes).

elhigu:
There should be custom wrappers only in few tests, which are exactly making sure that overriding those wrappers works as expected. Default wrappers should be restored after those tests if those wrapper tests were not separate knex instances in the first place.

kibertoad:
@alexandrebodin Do you need any additional help?

alexandrebodin:
@kibertoad I will be able to work on it next week ;)

kibertoad:
@alexandrebodin Any updates :)?

test/integration/schema/index.js
```diff
@@ -1048,6 +1048,24 @@ module.exports = function(knex) {
...
+        describe('sqlite only', function() {
+          if (
+            !knex ||
+            !knex.client ||
+            !/sqlite3/i.test(knex.client.driverName)
+          ) {
+            return Promise.resolve();
+          }
+
+          it('checks whether a column exists without being case sensitive, resolving with a boolean', function() {
```
kibertoad: Could you use arrow function instead of `function()` and async/await instead of `then` chain?

alexandrebodin: Sure. Not a problem even if not consistent with the rest of the file ?

kibertoad: yeah, we want to eventually convert everything

PR #3438, opened 2019-09-12
Fixes types for Seeder#run

ksmithut:
The `seed.run` command returns a array of a single item, and that item contains the array of strings (filepaths of seeds that were run).

This comes from here: 
- https://github.com/tgriesser/knex/blob/master/lib/seed/Seeder.js#L37
- https://github.com/tgriesser/knex/blob/master/lib/seed/Seeder.js#L90
- https://github.com/tgriesser/knex/blob/master/lib/seed/Seeder.js#L162

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1826,7 +1826,7 @@ declare namespace Knex {
   class Seeder {
     constructor(knex: Knex);
     setConfig(config: SeederConfig): SeederConfig;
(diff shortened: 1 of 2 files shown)


ksmithut:
The test failures seem intermittent, but I could be wrong. Is there a way to trigger another run or is that test failure legitimate?

kibertoad:
yeah, that one is not important. Thanks!

Issue #2213, opened 2017-09-06
Regression: Concurrent subtransactions result in incorrect savepoint release order


joepie91:
## Affected versions

I have not determined the earliest version where this occurs, but it occurs in the latest release (`0.13.0`), and the issue still exists in `master`. I've tested with PostgreSQL, but there is no PostgreSQL-specific transaction logic, so presumably this affects other SQL databases as well (as the implicit-release behaviour involved is a standard feature).

## Reduced testcase

```js
'use strict';

const Promise = require("bluebird");
const knex = require("knex")(require("./knexfile"));

Promise.try(() => {
	return knex.transaction((trx) => {
		return Promise.all([
			trx.transaction((trx2) => {
				return Promise.delay(500);
			}),
			trx.transaction((trx3) => {
				return Promise.delay(600);
			}),
		]);
	});
}).then((result) => {
	console.log("All succeeded");
}).catch((err) => {
	console.error(err);
});
```

## Expected result

600 milliseconds delay, then an "All succeeded" message.

## Actual result

```
error: RELEASE SAVEPOINT trx6; - no such savepoint
    at Connection.parseE (/home/sven/projects/knex-subtransaction-bugcase/node_modules/pg/lib/connection.js:546:11)
    at Connection.parseMessage (/home/sven/projects/knex-subtransaction-bugcase/node_modules/pg/lib/connection.js:371:19)
    at Socket.<anonymous> (/home/sven/projects/knex-subtransaction-bugcase/node_modules/pg/lib/connection.js:114:22)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:547:20)
```

## Relevant debug logs:

```
  knex:tx trx1: Starting top level transaction +0ms
  knex:pool INFO pool postgresql:pg:client0 - dispense() clients=1 available=0 +4ms
  knex:client acquired connection from pool: __knexUid2 +30ms
  knex:query BEGIN; +1ms
  knex:bindings undefined +0ms
  knex:tx trx5: Starting nested transaction +5ms
  knex:tx trx6: Starting nested transaction +0ms
  knex:pool INFO pool postgresql:pg:client0 - dispense() clients=0 available=1 +1ms
  knex:pool INFO pool postgresql:pg:client0 - dispense() clients=0 available=2 +0ms
[...]

PR #3440, opened 2019-09-13
Fix for concurrent child transactions failing (#2213)

codeclown:
Fix for https://github.com/tgriesser/knex/issues/2213 where @joepie91 already did the bulk of the work and investigation. I needed this fixed for a personal project so adapted their test and reimplemented the check for `_previousSibling`.

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -210,6 +210,7 @@ class Transaction extends EventEmitter {
   // the original promise is marked completed.
   acquireConnection(client, config, txid) {
     const configConnection = config && config.connection;
+    const trx = this;
     return new Bluebird((resolve, reject) => {
       try {
         resolve(configConnection || client.acquireConnection());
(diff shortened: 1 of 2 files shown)


codeclown:
The new test failed on mssql with the message:

```
SAVE TRANSACTION trx422 - Can't acquire connection for the request. There is another request in progress.
```

Probably related to this: https://github.com/tediousjs/node-mssql/issues/302

kibertoad:
@codeclown I assume situation is not any worse than it was before for mssql in this scenario? If that's the case, probably it's ok to just skip mssql for the test.

codeclown:
@kibertoad I'm personally fine with that, it could be noted in the docs where there's a note about Redshift lack of savepoint support as well. I'll update the PR.

I'm not 100% content with the purely timing-based test here, though. I'll see if I could make it more concrete by adding an assurance that the promises resolve in the expected order always.

codeclown:
@kibertoad Updated PR, please see 2 commits.

kibertoad:
Much appreciated!

Issue #1747, opened 2016-10-18
Corrupt migration directory


annnfrann:
Every time I try to run a migration to production, I get the following error: `Error: The migration directory is corrupt, the following files are missing: 20160712115200_create_books.js`. This is an old migration file, not in the database I'm trying to migrate, and has quite probably been deleted. I've checked other issues and the docs, but I haven't found anything to help me with this problem. I'm not sure what to do.

elhigu:
If you are sure your database is in consistent state with your set of migrations you can look your `migrations` table from the database and remove the row, which mentions the missing migration file.

Other way to get rid of the error is to create migration file, which actually doesn't do anything with name: `20160712115200_create_books.js`. There is no checksum of the original migration file stored, so system will think that your newly created file was the one it sees in the `migrations`table.

Please close this issue afterwards if one of the tricks did work for you.

DariaDesiderata:
I am having the same issue. The solution proposed by @elhigu has absolutely nothing to do with the problem. Knex is somehow looking at old migrations folder instead of the current one. In my case the old folder is from March 2017 and the current one is from December 2017.

elhigu:
@DariaDesiderata knex tries to look migrations from the folder that is mentioned in knexfile. I might be able to help if you can show me the list of your migration files in those directories and contents of your `migrations` table.

If you have once ran migrations from your `March 2017` directory to your database, then you cannot just change the migration directory to point to `December 2017`, because knex requires that old migration files are not removed.

If you change the folder, then knex will think that there are some migrations missing. tdlr; knex does not support running migrations from multiple separate folders to the same database.

elhigu:
Though there is no reason why it couldn't support that. 

So if you make feature request for some option to disable migration directory consistency check I don't see why it couldn't be implemented if someone is motivated enough to do it.

abemm7787:
@annnfrann 
I would just copy and paste the 20160712115200_create_books.js  back into the explorer.

badbabykosh:
No one seems to be listening to the women here. I quote @dariawhitecastle..."I am having the same issue. The solution proposed by @elhigu has absolutely nothing to do with the problem. Knex is somehow looking at old migrations folder instead of the current one. In my case the old folder is from March 2017 and the current one is from December 2017." 
[...]

PR #3448, opened 2019-09-20
New config option to disable validateMigrationsList()

arrilot:
@elhigu here https://github.com/tgriesser/knex/issues/1747#issuecomment-349365656
> So if you make feature request for some option to disable migration directory consistency check I don't see why it couldn't be implemented if someone is motivated enough to do it.

So here we go

Motivation:
We use gitflow with features in different branches and deploy them separately to stage environment.
All branches on stage environment have one common database.
As a result we constantly face the same issue:
1. Developer A makes a new migration A in a new branch A. Branch A is deployed on stage
2. Developer B makes a new migration B in a new branch B (has nothing to do with branch A). Branch B cannot be deployed on stage because "The migration directory is corrupt".

Rolling back migrations is not a solution because not every migration can be rolled back
Reseting database is not a solution because it erases data in database aswell
Merging branch A into B is not a solution because it breaks independency of features.

We know that switching off migration order brings some inconsistency issues but they are far less important for our local/stage environments.

Some other migrators like node-pg-migrate have such feature. (https://github.com/salsita/node-pg-migrate/blob/master/docs/cli.md check-order)


elhigu:
> Rolling back migrations is not a solution because not every migration can be rolled back
Reseting database is not a solution because it erases data in database aswell
Merging branch A into B is not a solution because it breaks independency of features.

This is exactly the reason why I have been slightly against of having this feature included to knex. I haven't found many cases where having the switch actually would make sense and would not cause user's system to be inconsistent and broken.

I know this is a bit harsh, but to be honest your process of using staging environment is horribly wrong. It is inviting strange errors, which will appear only in staging environment and it doesn't make sure that multiple merged features actually works correctly when combined.

Effectively staging environment should mirror final production environment to verify that everything is fine before putting it to production.

If you need to test out every branch first separately in staging environment with shared database you should also clean up all changes done to DB when you revert the certain features code form the environment. 

> Reseting database is not a solution because it erases data in database aswell

Reseting database is the solution, you just need to take SQL dump of the database from the stage where it was before starting to test multiple feature branches. Then before or after running tests for each branch just restore database state from that SQL dump.

That way you will always have consistent database state with code. If you just run bunch of migrations from various features can make completely working feature to seem like broken one (just to name one: migrations also can remove data that other branch might need).

Anyhow IMO that kind of testing  should be done in testing environment instead of staging. In staging you should apply each feature on top of each other the same way it will be done finally to the production system.

arrilot:
> Effectively staging environment should mirror final production environment to verify that everything is fine before putting it to production.
> Anyhow IMO that kind of testing should be done in testing environment instead of staging. In staging you should apply each feature on top of each other the same way it will be done finally to the production system.

Looks like in my description i've used a term "stage environment" in a misleading way. 
I've meant test environment, not preproduction environment. Sorry about that.

A little bit more info about our setup:
Environments: `local (developers' PCs) -> test -> preproduction -> production`. Each of them has their own database
In preproduction envorinment everything is fine since only migrations from `release` branch are applyied there.
All of the discussion is about `test` environment where `develop` branch and many `feature/x` branches co-exist linked to a shared DB
In this environment QAs test features before merging them to develop branch. They also test develop branch before releasing it to preproduction/production.

I agree that sharing a database is kinda asking for troubles and not ideal, but we can't spin up a new database for every feature branch due to infrastracture limitations. We can probably make a separate database for `develop` in the future, though.

> Reseting database is the solution, you just need to take SQL dump of the database from the stage where it was before starting to test multiple feature branches. Then before or after running tests for each branch just restore database state from that SQL dump.

Reseting brings new problems to the table:
1. DevOps need to set up this resetting workflow in Jenkins (we have dozens of services btw)
2. It will be annoyning to constantly reset database every time a QA wants to test a feature. What if two QAs want to test different features at the same time? They are going to block and interfere with each other.

> If you just run bunch of migrations from various features can make completely working feature to seem like broken one (just to name one: migrations also can remove data that other branch might need).

We try to make all the features (migrations) backward compatible as much as possible. It also helps with AB tests and rolling back releases from production in case of accident.

As a result we prefer approach from this PR.

elhigu:
> A little bit more info about our setup:
> Environments: `local (developers' PCs) -> test -> preproduction -> production`. Each of them has their own database
> In preproduction envorinment everything is fine since only migrations from `release` branch are applyied there.
> All of the discussion is about `test` environment where `develop` branch and many `feature/x` branches co-exist linked to a shared DB
> In this environment QAs test features before merging them to develop branch. They also test develop branch before releasing it to preproduction/production.
> 
> I agree that sharing a database is kinda asking for troubles and not ideal, but we can't spin up a new database for every feature branch due to infrastracture limitations. We can probably make a separate database for `develop` in the future, though.
> 
> > Reseting database is the solution, you just need to take SQL dump of the database from the stage where it was before starting to test multiple feature branches. Then before or after running tests for each branch just restore database state from that SQL dump.
> 
> Reseting brings new problems to the table:
> 
> 1. DevOps need to set up this resetting workflow in Jenkins (we have dozens of services btw)

It could be integrated as part of test command and all services could be stored / restored by the same script.

> 2. It will be annoyning to constantly reset database every time a QA wants to test a feature. What if two QAs want to test different features at the same time? They are going to block and interfere with each other.

Ok, so this is environment where you are hand testing stuff... so there is no tests script where one could do the save/restore etc. Well sounds like you are out of luck then... 

> > If you just run bunch of migrations from various features can make completely working feature to seem like broken one (just to name one: migrations also can remove data that other branch might need).
> 
> We try to make all the features (migrations) backward compatible as much as possible. It also helps with AB tests and rolling back releases from production in case of accident.
> 
> As a result we prefer approach from this PR.

Good luck!

arrilot:
> PR is missing tests and documentation
@elhigu done

Looks like build is failing due to totally unrelated thing

      -    "balance": 12.240001
      +    "balance": 12.24

kibertoad:
@arrilot Pull from master, that test should be fixed now (there is a failing CLI test now, feel free to ignore it)

arrilot:
> @arrilot Pull from master, that test should be fixed now (there is a failing CLI test now, feel free to ignore it)

Done, thanks

PR #3449, opened 2019-09-21
This is a patch to fix issues #1901, #2990 and #2991

oitTim:
We use Oracle stored procedures and get the error 'expecting an array or an iterable object but got [object Null]' . This bugfix is kind of important as we convert old MuleSoft flows to Node endpoints.

Let me know if you need any more information or documentation.


package.json
```diff
@@ -36,6 +36,7 @@
     "liftoff": "3.1.0",
     "lodash": "^4.17.15",
     "mkdirp": "^0.5.1",
+    "oracledb": "^4.0.1",
```
kibertoad: It would be preferable to avoid adding this, but instead move tests to integration test suite and skipping it for non-oracle dbs.

oitTim: Where in the integrations folder should I place the tests?

kibertoad: I'd say that `https://github.com/tgriesser/knex/tree/master/test/integration/builder`, probably`additional`, since I assume this is not a select per se?

elhigu:
Did we have some wiki or cookbook for knex? That kind of entry would be great for documentation. Also great work, this has been very wanted feature 👍

oitTim:
I will create a cookbook on how to call an oracle stored procedure and get the results :D

elhigu:
https://github.com/tgriesser/knex/wiki/Recipes

oitTim:
Simple example added to the recipe.

kibertoad:
Released in 0.19.5

PR #3459, opened 2019-10-02
feat: Add types to the Migrator interface

lffg:
Add typings to the `list` and `forceFreeMigrationsLock` methods.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1828,7 +1828,6 @@ declare namespace Knex {
     currentVersion(config?: MigratorConfig): Promise<string>;
     up(config?: MigratorConfig): Promise<any>;
     down(config?: MigratorConfig): Promise<any>;
-    list(config?: MigratorConfig): Promise<any>;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -1828,6 +1828,8 @@ declare namespace Knex {
     currentVersion(config?: MigratorConfig): Promise<string>;
     up(config?: MigratorConfig): Promise<any>;
     down(config?: MigratorConfig): Promise<any>;
+    list(config?: MigratorConfig): Promise<any>;
```
kibertoad: I think list is already included and can be removed from this PR?

lffg: Where?

kibertoad: https://github.com/knex/knex/blob/master/types/index.d.ts#L1830

lffg: Ah, ok, thanks. :)

kibertoad:
Thanks!

Issue #3257, opened 2019-06-05
[Bluebird] Replace bluebird.promisify with util.promisify


kibertoad:
Pay attention to native promises not accepting context for method being executed as a second parameter (as opposed to bluebird)

PR #3470, opened 2019-10-08
Use util.promisify instead of Bluebird.promisify

maximelkin:
fixes https://github.com/tgriesser/knex/issues/3257

### DIFF
--- a/lib/migrate/MigrationGenerator.js
+++ b/lib/migrate/MigrationGenerator.js
@@ -1,7 +1,7 @@
 const fs = require('fs');
 const path = require('path');
+const { promisify } = require('util');
 const mkdirp = require('mkdirp');
-const {promisify} = require('util');
 const { writeJsFileUsingTemplate } = require('../util/template');
 const { getMergedConfig } = require('./configuration-merger');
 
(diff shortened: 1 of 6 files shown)


maximelkin:
Not sure what to do with Bluebird.promisifyAll in https://github.com/tgriesser/knex/blob/master/lib/dialects/oracle/index.js#L85

lib/migrate/MigrationGenerator.js
```diff
@@ -30,9 +30,7 @@ class MigrationGenerator {
     const dirs = this._absoluteConfigDirs();
 
     const promises = dirs.map((dir) => {
-      return bluebird
-        .promisify(fs.stat, { context: fs })(dir)
-        .catch(() => bluebird.promisify(mkdirp)(dir));
+      return promisify(fs.stat)(dir).catch(() => promisify(mkdirp)(dir));
```
kibertoad: `{ context: fs }` part is irrelevant?

maximelkin: yes, but i cant find it in official docs
https://nodejs.org/docs/latest-v8.x/api/util.html#util_util_promisify_original

kibertoad:
Bunch of conflicts came up.

kibertoad:
Thanks!

PR #3471, opened 2019-10-08
remove bluebird.each

maximelkin:
closes https://github.com/tgriesser/knex/issues/3256

### DIFF
--- a/lib/dialects/oracledb/index.js
+++ b/lib/dialects/oracledb/index.js
@@ -7,6 +7,7 @@ const ColumnCompiler = require('./schema/columncompiler');
 const { BlobHelper, ReturningHelper, isConnectionError } = require('./utils');
 const Bluebird = require('bluebird');
 const stream = require('stream');
+const { promisify } = require('util');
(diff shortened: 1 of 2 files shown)


maximelkin:
@kibertoad maybe its better to merge all prs in one?

lib/dialects/oracledb/index.js
```diff
@@ -128,7 +129,7 @@ Client_Oracledb.prototype.acquireRawConnection = function() {
           });
         });
       };
-      const fetchAsync = function(sql, bindParams, options, cb) {
+      const fetchAsync = promisify(function(sql, bindParams, options, cb) {
```
kibertoad: Does this function use any closure variables, or it can be moved outside to only be promisified once?

kibertoad:
@maximelkin Keeping PRs separate makes them easier to review :)

lib/dialects/oracledb/index.js
maximelkin: This function take connection from outside

PR #3474, opened 2019-10-09
Bluebird remove map mapSeries

maximelkin:


maximelkin:
Trying to find where unhandledRejection occurs, currently set WIP

maximelkin:
So it was from manual transactions, rewrite using callback-style

kibertoad:
Fantastic work, thank you!

PR #3477, opened 2019-10-12
refactor(bluebird): remove Bluebird.bind

maximelkin:


lib/migrate/Migrator.js
```diff
@@ -475,7 +474,7 @@ class Migrator {
   _waterfallBatch(batchNo, migrations, direction, trx) {
     const trxOrKnex = trx || this.knex;
     const { tableName, schemaName, disableTransactions } = this.config;
-    let current = Bluebird.bind({ failed: false, failedOn: 0 });
+    let current = Promise.resolve();
```
kibertoad: are those fields not used anywhere?

maximelkin: I didn't found any usages in project

maximelkin: 7a6937fdb329d0d0b29cd51fc28aea06e6903e0e first revision where failedOn exists, i think it just partially implemented feature

kibertoad: Fair enough, guess we can drop it.

kibertoad:
Sorry, there is a merge conflict now after refactoring of a seeder. Could you resolve them?

maximelkin:
Ok, i will resolve it later

lib/seed/Seeder.js
```diff
@@ -28,7 +20,7 @@ class Seeder {
   // Runs seed files for the given environment.
   async run(config) {
     this.config = this.setConfig(config);
-    const [all] = await this._seedData();
+    const all = await this._listAll();
```
kibertoad: Will types changed in https://github.com/knex/knex/pull/3438/files still be correct after this change?

lib/seed/Seeder.js
```diff
@@ -103,22 +85,14 @@ class Seeder {
     );
   }
 
-  // Generates the stub template for the current seed file, returning a compiled template.
```
kibertoad: @lorefnon I assume we stopped using it after your refactoring?

kibertoad:
@maximelkin Merged your other PR, that seems to have introduced a minor conflict as well.

lib/seed/Seeder.js
lorefnon: Yeah, i moved the lodash template usage to a shared utility which we can potentially use from other places as well.

lib/seed/Seeder.js
maximelkin: _seedData is just _listAll, but wrapping result inside promise into []

maximelkin:
@kibertoad done

PR #3479, opened 2019-10-12
Fix some issues around seed and migration generation

### LINKED ISSUE #3468: ReferenceError: name is not defined
# Environment

Knex version: 0.19.3
OS: macOS Mojave

# Bug

I am trying to use the following stub file for creating knex migration files and receive the error shown below.

```
// Update table name below
const name = 'TEMPTABLENAME';

exports.up = knex => {
    return knex.schema.createTable(name, table => {
        table.uuid(`${name}_id`).primary();
        table.timestamps(true, true);
    });
};

exports.down = knex => {
    return knex.schema.dropTableIfExists(name);
};
```

I believe this is cause by the string interpolation on line 6. Changing just line 6 to
```table.uuid('id').primary();``` makes the error go away but obviously doesn't really achieve what I am trying to do.

Error message
```
ReferenceError: name is not defined
    at eval (lodash.templateSources[0]:7:10)
    at MigrationGenerator._writeNewMigration (XXXX/server/node_modules/knex/lib/migrate/MigrationGenerator.js:66:9)
    at _ensureFolder.then.then (XXXX/server/node_modules/knex/lib/migrate/MigrationGenerator.js:24:27)
```

### PR DESCRIPTION
**Pull request by lorefnon**
Closes #3468

### DIFF
--- a/lib/migrate/MigrationGenerator.js
+++ b/lib/migrate/MigrationGenerator.js
@@ -2,7 +2,7 @@ const fs = require('fs');
 const path = require('path');
 const mkdirp = require('mkdirp');
 const bluebird = require('bluebird');
-const { template } = require('lodash');
+const { writeJsFileUsingTemplate } = require('../util/template');
 const { getMergedConfig } = require('./configuration-merger');
 
 class MigrationGenerator {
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: lib/migrate/MigrationGenerator.js [hunk 1]
```diff
@@ -30,42 +29,41 @@ class MigrationGenerator {
     const dirs = this._absoluteConfigDirs();
 
     const promises = dirs.map((dir) => {
-      return bluebird
-        .promisify(fs.stat, { context: fs })(dir)
-        .catch(() => bluebird.promisify(mkdirp)(dir));
+      return promisify(fs.stat, { context: fs })(dir)
```
kibertoad: Doesn't native promisify not accept the second param?

test/cli/seed-make.spec.js
```diff
@@ -0,0 +1,349 @@
...
+// To enable debug change to:
+// const NODE = 'node --inspect-brk';
+
+const {
+  seedStubOptionSetup,
+  expectContentMatchesStub,
+  setupFileHelper,
+} = require('./cli-test-utils');
+
+describe.only('seed:make', () => {
```
kibertoad: probably `.only` should be removed

lorefnon: Yeah, left it out by mistake.

PR #3496, opened 2019-10-25
feat: wrap subQuery with parenthesis when it appears as table name

### LINKED ISSUE #3485: please hint about missing "as()" when from(subquery) is missing it
# Environment

Knex version: 0.19.5
Database + version: postgresql 11
OS: linux/debian

I find myself often spending hours trying to figure how to make
```
knex('table').select(rawstr).from(knex('table').where(obj))
```
work.
It's very simple:
```
knex('table').select(rawstr).from(knex('table').where(obj).as('sub'))
```

However knex outputs a wrong sql statement (missing parenthesis) and when debugging, it looks like a bug in knex.

I think it would be nice to users if a missing as() could be detected and a warning printed.

edvardchen:
Fixes #3485

### DIFF
--- a/test/integration/builder/selects.js
+++ b/test/integration/builder/selects.js
@@ -1380,5 +1380,38 @@ module.exports = function(knex) {
         }
       }
     });
+
+    it('select from subquery', async function() {
+      const subquery = knex.from('accounts').whereBetween('id', [3, 5]);
+      return knex
+        .pluck('id')
+        .orderBy('id')
+        .from(subquery)
+        .then(
+          (rows) => {
+            expect(rows).to.deep.equal([3, 4, 5]);
+            expect(knex.client.driverName).to.oneOf(['sqlite3', 'oracledb']);
+          },
+          (e) => {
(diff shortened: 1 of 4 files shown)


kibertoad:
Could you please also add integration test that would actually execute the generated sql?

test/integration/builder/selects.js
```diff
@@ -1,6 +1,7 @@
 /*global expect, d*/
 'use strict';
 
+const mysqlErrors = require('mysql/lib/protocol/constants/errors');
```
kibertoad: Why depend on non-public part of external dependency and not just hardcode the constant internally?

edvardchen: Emm, OK. I will change it

kibertoad: My concern is that if there is a major change in the driver and both the constant is no longer there and field is no longer set, comparison is still going to be "undefined === undefined" and will pass. Thanks for changing it!

kibertoad:
Thanks!

PR #3497, opened 2019-10-25
Connection configuration provider - without changes to drivers

oranoran:
This is a simplified PR to implement the same functionality as suggested in #3403, but with the following differences:

1. All changes were made in client.js only, so they apply to all built-in and external drivers without any changes required in them, improving extensibility and maintainability.
2. Provided connection configurations can include an `expired()` function, allowing configurations to expire after a while.

The expiration feature is useful, for example, for expiring Amazon RDS IAM access tokens after their lifespan of 15 minute is over, like so:
```javascript

const connectionConfigProvider = () => {
    const { token, tokenExpiration } = await someCallToAwsToGetTheToken();
    return {
        host : 'your_host',
        user : 'your_database_user',
        password : token,
        database : 'myapp_test',
        expired: () => {
            return tokenExpiration <= Date.now();
        }
    };
}
```

I wrote the tests relying on sqlite3, hope this makes sense.
Please let me know if this direction seems right, then I'll add a documentation PR.


lib/client.js
```diff
@@ -254,16 +259,44 @@ Object.assign(Client.prototype, {
...
+        return; // static configuration, nothing to update
+      }
+      if (
+        this.connectionSettings.expired === undefined ||
+        !this.connectionSettings.expired()
+      ) {
+        return; // reuse existing connection
+      }
+      const providerResult = this.connectionConfigProvider();
+      if (!providerResult.then) {
```
kibertoad: why can't you just always `await` it?

oranoran: This allows the provider to be either synchronous or async.

lib/client.js
```diff
@@ -55,7 +55,12 @@ function Client(config = {}) {
     this.version = config.version;
   }
 
-  this.connectionSettings = cloneDeep(config.connection || {});
+  if (config.connection && config.connection instanceof Function) {
+    this.connectionConfigProvider = config.connection;
+    this.connectionSettings = { expired: () => true }; // causes the provider to be called on first use
```
kibertoad: `expirationChecker` might be a more self-explanatory name, to emphasize that it's not a flag.

lib/client.js
```diff
@@ -254,16 +259,44 @@ Object.assign(Client.prototype, {
     // choose the smallest, positive timeout setting and set on poolConfig
     poolConfig.acquireTimeoutMillis = Math.min(...timeouts);
 
+    const updatePoolConnectionSettingsFromProvider = () => {
+      if (!this.connectionConfigProvider) {
+        return; // static configuration, nothing to update
+      }
+      if (
+        this.connectionSettings.expired === undefined ||
```
kibertoad: Why not use falsiness check instead? If someone sets it to null explicitly, should be handled same way, and I assume that we don't want to support boolean values here.

lib/client.js
kibertoad: it's perfectly fine to await synchronous functions.

kibertoad:
TypeScript definitions probably need to be updated.

test/tape/connection-config-provider.js
```diff
@@ -0,0 +1,77 @@
+'use strict';
+
```
kibertoad: Can you also add integration test that would be run on all databases and verify connection settings to be reused across multiple connections?

oranoran:
> TypeScript definitions probably need to be updated.

I checked this, looks like the whole "connection" object is "any". Not changing anything else in the API, so didn't find anything that needs to be changed.

test/tape/connection-config-provider.js
oranoran: The current test does this for psqlite3 - creating 2 concurrent connections and counting the number of provider invocations. Just 1 invocation means the provider has been reused.

Is there a generic way to write a test to run on top of all databases?

lib/client.js
oranoran: will change

lib/client.js
oranoran: will change

lib/client.js
oranoran: According to the [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await):
If the value is not a Promise, it converts the value to a resolved Promise, and waits for it.

It sounds like this might split execution flow, causing a slight performance impact, especially on older Node versions (I think this has been optimized away in Node 12). For this reason I was trying not to add any async stuff along the core execution path unless it's required for supporting the new functionality.

Just bringing this consideration to attention - if you still feel it's better to use await, no problem I will change it.

test/tape/connection-config-provider.js
kibertoad: Yeah, check tests within `https://github.com/knex/knex/tree/master/test/integration` folder. We basically iterate over each test and pass `knex` instance to it with initialized connection. I assume you can reuse the code from how this iteration is done, but probably you will need to instantiate it separately, as you need different configuration.

kibertoad: @oranoran You can see how iteration is done here: https://github.com/knex/knex/blob/master/test/integration/index.js

lib/client.js
kibertoad: @oranoran Please correct me if I'm wrong, but this does not seem to be a hot path and is likely to be executed fairly rarely, hence microoptimization unlikely to be useful, and it does simplify code quite a bit in this case.

lib/client.js
```diff
@@ -279,34 +269,26 @@ Object.assign(Client.prototype, {
...
       }
-      const providerResult = this.connectionConfigProvider();
-      if (!providerResult.then) {
-        handleConnectionConfigProviderResult(providerResult);
-        return; // provider is synchronous, no need to return a promise
+      const providerResult = await this.connectionConfigProvider();
+      if (providerResult.expirationChecker) {
+        this.connectionConfigExpirationChecker =
+          providerResult.expirationChecker;
+        delete providerResult.expirationChecker;
```
kibertoad: why not just set to undefined? AFAIR it's still considerably faster.

oranoran: I replaced the delete of `this.connectionConfigExpirationChecker` a couple of lines later into null, because it seems a bit strange for the the client's own attributes to be just coming and going.

But as for this delete, there is a catch. The MySQL2 driver prints this nasty message if the expirationChecker property is present in the config, even if its value is `undefined`:
`Ignoring invalid configuration option passed to Connection: expirationChecker. This is currently a warning, but in future versions of MySQL2, an error will be thrown if you pass an invalid configuration option to a Connection`

kibertoad: Can you add comment with this explanation, so that noone removes it in the future?

oranoran:
@kibertoad Ok so the only open issue AFAIK is that I get a test failing on Oracle when Travis runs automatically. I can't reproduce this because when I try to run Oracle tests locally according to the instructions, I get an error.
```
 > DB='oracledb' mocha  --grep "Connection configuration" --exit -t 10000 test/index.js
Warning: Could not find any test files matching pattern: configuration"
sqlite does not support inserting default values. Set the `useNullAsDefault` flag to hide this warning. (see docs http://knexjs.org/#Builder-insert).
sqlite does not support inserting default values. Set the `useNullAsDefault` flag to hide this warning. (see docs http://knexjs.org/#Builder-insert).
sqlite does not support inserting default values. Set the `useNullAsDefault` flag to hide this warning. (see docs http://knexjs.org/#Builder-insert).
Knex: run
$ npm install oracledb --save
Cannot find module 'oracledb'
Error: Cannot find module 'oracledb'
    at Function.Module._resolveFilename (module.js:548:15)
    at Function.Module._load (module.js:475:25)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Client_Oracledb._driver (/Users/oran/proj/knex/lib/dialects/oracledb/index.js:30:20)```

**Comment by kibertoad** [comment]
It is a flaky test, let me rerun it.

**Comment by oranoran** [comment]
> > TypeScript definitions probably need to be updated.
> 
> I checked this, looks like the whole "connection" object is "any". Not changing anything else in the API, so didn't find anything that needs to be changed.

Actually, looking again at index.d.ts, I see there are ConnectionConfig interfaces defined for the various databases. Should I add an expiryChecker to each one?

**Comment by kibertoad** [comment]
Would be helpful, yes.

**Comment by oranoran** [comment]
The test is still failing:
```
  1) Integration Tests
       oracle | oracledb
         knex.migrate
           knex.migrate.up
             should only run the next migration that has not yet run if other migrations have already run:
      AssertionError: expected [ Array(1) ] to have a length of 2 but got 1
```

I can't reproduce it since I'm not able to run Oracle tests, any suggestions?

P.S the TS definitions I added aren't complete, I need to add a modified connection definition too. Will add later today.

kibertoad:
@oranoran It did pass after re-run, so as expected, flaky test, feel free to ignore.

kibertoad:
@oranoran Could you also create a documentation PR?

kibertoad:
@oranoran There is a conflict with types in master now.

kibertoad:
Thanks a lot for all the effort put into this!

oranoran:
> Thanks a lot for all the effort put into this!

Thanks for maintaining this awesome lib!

Ali-Dalal:
@oranoran Thanks for finalizing this 👍 

@kibertoad when can you release it the changes?

kibertoad:
@Ali-Dalal Released in 0.20.1

PR #3501, opened 2019-10-28
Fix undefined value error

acofer:
Since version 0.19.5, Oracle is throwing `type is undefined` when we call `.columnInfo()`. This fixes it.


acofer:
Disregard, just saw that it's fixed in 0.20.0.

PR #3502, opened 2019-10-29
Gracefully handle global promise pollution

lorefnon:
- Update QueryInterface to proxy to only standard promise methods
- Update QueryInterface type to be explicit about the methods being
  proxied

This primarily handles pollution of global Promise (either the type or
the global object at runtime) by other libraries.

Fixes: https://github.com/knex/knex/issues/3422#issuecomment-546930375

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1401,7 +1401,7 @@ declare namespace Knex {
   // Chainable interface
   //
 
-  interface ChainableInterface<T = any> extends Promise<T> {
+  interface ChainableInterface<T = any> extends Pick<Promise<T>, "then" | "catch" | "finally"> {
     toQuery(): string;
     options(options: { [key: string]: any }): this;
     connection(connection: any): this;


onehorsetown:
https://github.com/knex/knex/pull/3502/commits/e9fd66e2d2f5525bbaa6c5fe5b55a1a4bf98def1

This change broke TSLint for my TypeScript project.  Everything lints fine in `0.20.0`, but with `0.20.1`, I get tons of:

```
ERROR: (await-promise) ... Invalid 'await' of a non-Promise value.
ERROR: (no-invalid-await)  ...Refactor this redundant 'await' on a non-promise.
```

Trying to figure out why the tslint `no-invalid-await` and `await-promise` don't think this is a promise/thenable/etc.

lorefnon:
Do you have `"await-promise": [true, "Thenable"]` in your tslint config ?

types/index.d.ts
```diff
@@ -1401,7 +1401,7 @@ declare namespace Knex {
   // Chainable interface
   //
 
-  interface ChainableInterface<T = any> extends Promise<T> {
+  interface ChainableInterface<T = any> extends Pick<Promise<T>, "then" | "catch" | "finally"> {
```
elhigu: Does promise have "finally" method? I thought it was just a bluebird thing.

onehorsetown:
@lorefnon Yes, I've tried every variant of 

`"await-promise": [true, "Thenable", "PromiseLike", "Bluebird", etc...]`

I've stopped in the TSLint code and the name of the type at the point where it is checked is `__type`.   So, this actually makes the problem go away, but seems very wrong:

`"await-promise": [true, "__type"]`

The other linting error (`no-invalid-await`) is coming from the `tslint-sonarts` package which is checking if the type has a `then` method right here:

https://github.com/SonarSource/SonarTS/blob/4a4080b78001a78d758d1d0fa0190fb9496b6f57/sonarts-core/src/rules/noInvalidAwaitRule.ts#L57

At that point in the linting code, the `ChainableInterface` does not have a `then` method, but I have no idea why.

types/index.d.ts
onehorsetown: The Typescript`lib.es2018.promise.d.ts` definition added `finally`

https://github.com/microsoft/TypeScript/blob/234680851bbbe83dd6bcd8aac165a57adbf586e2/src/lib/es2018.promise.d.ts#L11

I don't follow the whole feature ratification process.  I just know `finally` works in Node 12.

72636c:
You can make an exception with `"await-promise": [true, "ChainableInterface"]`

onehorsetown:
> You can make an exception with `"await-promise": [true, "ChainableInterface"]`

@72636c - That fixes it.  Thanks.

The other issue I then have is that the `tslint-sonarts` linting plugin does not think the `Pick`ed type does not have a `then` method... which it, of course, does.

PR #3505, opened 2019-10-30
Fix typing file

MiOnim:
Two changes in the typings file:

- Fixed the type of `log` method in PoolConfig interface
- Removed `create`, `destroy` and `validate` methods from PoolConfig interface, because Knex is overriding these methods, and any user provided implementations of these methods are being ignored. Having these methods in the typings file will only confuse users.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1802,9 +1802,7 @@ declare namespace Knex {
 
   interface PoolConfig {
     name?: string;
-    create?: Function;
     afterCreate?: Function;
-    destroy?: Function;
     min?: number;
     max?: number;
     refreshIdle?: boolean;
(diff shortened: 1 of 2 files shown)


MiOnim:
Can someone please review this? CC: @elhigu

elhigu:
I think that at least validate should be able to be overridden by the user. But I believe that the best way to implement pool config typings would be to somehow inherit/extend it from tarn's config type and add javascript level errors, when trying to override create/destroy. I'm not completely sure if even overriding those must be denied if the one providing new implementations knows what they are doing 🤔

MiOnim:
@elhigu Knex seems to ignore any user provided implementation of validate/create/destroy, and pass its own implementation of these methods to tarn. I was just trying to update the typings file to express this intent accurately. Do you want me to change Knex's behavior and allow users to override those methods?

elhigu:
> Do you want me to change Knex's behavior and allow users to override those methods?

No need to change the current implementation. If some one starts requesting it it will be easy to improve afterwards.

MiOnim:
@elhigu In that case, this PR should be ready to be approved and merged?

elhigu:
Retriggered tests... for some reason there was no button visible in travis to retry those.

kibertoad:
Released in 0.20.2

Issue #3422, opened 2019-09-02
Property 'timeout' does not exist on type 'QueryBuilder'


3mini:
# Environment

Knex version: 0.19.3
Database + version: MySQL 5.7.22
OS: 10.14.6

If issue is about TypeScript definitions, tag @lorefnon.

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do
    
    I'm using **Typescript Version 3.5.1**.
    When I upgrade knex version **from 0.17.3 to 0.19.3**, I got a compile error regarding [timeout](http://knexjs.org/#Builder-timeout).
    It worked well before. 🙃

2. Error message
    ```
    Property 'timeout' does not exist on type 'QueryBuilder<unknown, 
    DeferredKeySelection<unknown, never, false, {}, false, {}, never>[]>'
    ```

kibertoad:
It was a bluebird method. We no longer support bluebird.
That said, timeout functionality is, indeed, useful. Wonder if we handle it otherwise right now.

Heikkohoo:
Same issue with the map - property

Property 'map' does not exist on type 'QueryBuilder<unknown, (DeferredKeySelection<unknown, string, true, {}, boolean, {}, unknown> | DeferredKeySelection<unknown, "id", true, {}, false, {}, never>)[]>'

elhigu:
> It was a bluebird method. We no longer support bluebird.
That said, timeout functionality is, indeed, useful. Wonder if we handle it otherwise right now.

IIRC http://knexjs.org/#Builder-timeout is not from bluebird, but actually implemented in knex. It actually cancels the query if timeout is triggered.

kibertoad:
@Heikkohoo `map` is definitely gone for good, just await results and call map on array that you get.
Please check if `timeout` actually works on latest version if you add ts-ignore, if it does, we need to add it to our knex types.

Heikkohoo:
@kibertoad thanks. Got the .map thinghy working.

kibertoad:
@3mini Types were fixed in 0.19.4 that was just published.

3mini:
@kibertoad Thanks 🙏

3mini:
@lorefnon There's still a compile error as below.

```
timeout(ms: number, options: {cancel?: boolean}): QueryBuilder<TRecord, TResult>;
                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'options' was not provided.
```

[...]

PR #3507, opened 2019-10-31
fix: Support nodejs v8 promise typing

alexkvak:


drwatsno:
Wait.. I think it will break someone who already using this typings :thinking:

drwatsno:
Anyway, could someone please explain this ```Pick<Promise<T>, "then" | "catch" | "etc">``` typing? It looks confusing

alexkvak:
@kibertoad I guess the solution of #3422 was wrong. 

Could you explain why the previous version 
`interface ChainableInterface<T = any> extends Promise<T>`
does the wrong things?

lorefnon:
@alexkvak can you please provide some context on what are you trying to fix ? 

> Wait.. I think it will break someone who already using this typings 🤔

Yes, finally will not be available in QueryInterface where as it is exposed.

alexkvak:
I try to return node v8 support.
If you create new project from scratch with @types/node@8 and latest knex version, you become Typescript compile error.

Sorry, I can’t provide some code just node.

lorefnon:
> Could you explain why the previous version
interface ChainableInterface<T = any> extends Promise<T>
does the wrong things?

So, typescript interfaces are mutable. And it is possible for other packages (like  `@types/bluebird-global` as was the case in https://github.com/knex/knex/issues/3422) to extend global interfaces and add keys etc. You can read more about this feature [here](https://www.typescriptlang.org/docs/handbook/declaration-merging.html).

ChainableInterface (what knex.select etc.) returns is not a Promise, but it exposes *some* promise methods which get [delegated](https://github.com/knex/knex/blob/master/lib/interface.js#L104) to the resolved promise when invoked.

When this happens then `ChainableInterface extends Promise` is wrong because it claims that it is compatible with Promise but in reality it is not. It is only exposing a subset of the methods.

This is usually not a problem in practice, but one case in particular (timeout) is problematic because Bluebird's timeout doesn't match the signature of an identically named function implemented in Knex. 

There shouldn't really be a conflict because Bluebird's timeout is not actually available through QueryInterface even if it was the global Promise. But the conflict arises because of the lax type definition of QueryInterface. 

> Anyway, could someone please explain this Pick<Promise<T>, "then" | "catch" | "etc"> typing? It looks confusing

It is conceptually similar to Lodash.pick. You get only the keys that you pick mapped to the corresponding values:

```
type A = { a: string, b: string, c: string }
type B = Pick<A, "a" | "b">;
// Same as: 

type B = { a: string, b: string }
```

lorefnon:
@alexkvak Can you please share the error that you get ?

alexkvak:
@lorefnon thanks!

kibertoad:
Released in 0.20.2

Issue #3472, opened 2019-10-09
Request for distinctOn Method


zjr:
# Environment

Knex version: 0.19.5
Database + version: PostgreSQL v11.5
OS: MacOS

# Feature discussion / request

### Background

This is about adding a `distinctOn` method to Knex, in addition to the existing `distinct`. This has been previously discussed in these issues: https://github.com/knex/knex/issues/51, https://github.com/knex/knex/issues/681.

### Use Case

As mentioned in the above referenced issues, it's currently possible to use `DISTINCT ON` through `raw`, but there are some problems with this. 

1. By default, Knex will add a `,` after the clause which is invalid in PostgreSQL and causes the query to fail. For example it will turn out looking like `select distinct on ("column1"), "column2" from "tbl_name"`.
2. If one were to write the entire select statement with `raw`, the query would no longer be reusable if additional selects needed to be added because Knex will append the select with a comma.
3. The best way I've come to to work around this is to do something like `knex('table').select(knex.raw('distinct on (??) ??', ['id', 'id']))`. The problem with this is the possibility for duplicate columns in select.

### What would support this?

A `distinctOn` function that would keep the `distinct on` portion separate from other `select`'s.

### API Proposal

```js
knex('users').distinctOn('id').select('name');
```
Resulting in
```SQL
select distinct on ("id") "name" from "users";
```

elhigu:
Sounds fine to me... If the method is called with any other DB (not supporting it) error should be thrown,  Anyone... feel free to implement and send PR for this 👍

kibertoad:
Released in 0.20.2

jsilveira2:
Hi, i updated my knex version to 0.20.2, but i still cant use distinctOn.
Knex return:  **knex(...).distinctOn is not a function**;

Do i have to call the method like this? knex('table1').distinctOn('columnA').

lorefnon:
@jsilveira2 Can you please provide an example of how you are using it ?

jsilveira2:
@lorefnon Yes of course, thanks for answer.

My complete query:

```
        const queryWhen = knex.raw(`
		        pe.id,
		        pp.fl_permitir,
		        rec.ds as "recurso",
		        op.ds as "operacao",
		        ROW_NUMBER() OVER(PARTITION BY pp.id_permissao ORDER BY CASE pp.id_perfil
		    ${perfisCase} END) AS rk
        `);

        const query = knex.with('permissoes', qb => {
[...]

PR #3513, opened 2019-11-03
Add support for distinct on for postgres #3472

lorefnon:


lib/query/compiler.js
```diff
@@ -574,6 +578,12 @@ assign(QueryCompiler.prototype, {
     );
   },
 
+  distinctOn(value) {
+    throw new Error(
+      '.distinctOn() is currently only supported on PostgreSQL'
```
kibertoad: How difficult would it be to support equivalent MySQL syntax? https://stackoverflow.com/questions/17673457/converting-select-distinct-on-queries-from-postgresql-to-mysql

Also does that also work on redshift?

lorefnon: @kibertoad I don't think it is a very good idea to emulate this feature in Knex because:

1. It introduces more complexity (both in implementation, and affects runtime performance) 
2. I don't see a straightforward emulation that is generically applicable 

For instance, consider the last example in that answer, if we add one more column (col5) to the order by then the solution will need more modification because we would need to find the minimum of col5 among all rows that have same value of col4 potentially introducing another level of nesting. 

Redshift doesn't support distinct on either.

kibertoad:
@lorefnon Can you also add documentation for it?

PR #3531, opened 2019-11-09
Expand SeedsConfig types

bensaufley:
This is not a change to behavior; it's just an update to the TypeScript types to reflect the actuality of what's in the codebase. Inspecting `lib/seed/Seeder.js` there are five unique references to `this.config.`, but only two were previously represented in the types.

I'm a big fan of generics and how they can help us write more explicit code but if you want to eliminate that for something more broad, like just `variables?: object` or `variables?: any` that's fine; I'm not using it.

I'm submitting this PR because I encountered the following issue:

`knexfile.ts` has the following:

```js
seeds: {
  directory: './db/seeds',
  stub: './db/seed.stub.ts',
}
```

Using the `.ts` stub file by way of `npx knex seed:make thing` generates a migration with a `.js` extension. Obviously that's not right. I can add `-x ts` in the command line, but it seems silly to need to do that every time when it's the only way I want to use it in the project.

So I thought, can I add `extension` as I can with `migrations`? TypeScript says no. But I tried it, with `} as any` for the seeds value, and it worked. So then I dug into the code, and yep there it is.

tl;dr `SeedsConfig` respects `extension` but the Type doesn't allow it currently. Everything else is secondary.

Since this isn't a change to behavior, and all added keys are optional, it shouldn't be a breaking change and I haven't written any tests. Please let me know if there's more I can do to ensure this meets the standards of the repo.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1622,7 +1622,7 @@ declare namespace Knex {
     nullable: boolean;
   }
 
-  interface Config {
+  interface Config<SV extends object = object> {
     debug?: boolean;
     client?: string | typeof Client;
     dialect?: string;
@@ -1636,7 +1636,7 @@ declare namespace Knex {
       origImpl: (value: string) => string,
       queryContext: any
     ) => string;
-    seeds?: SeedsConfig;
+    seeds?: SeedsConfig<SV>;
     acquireConnectionTimeout?: number;
     useNullAsDefault?: boolean;
     searchPath?: string | string[];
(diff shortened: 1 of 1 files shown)


lorefnon:
@bensaufley Thanks for this PR. 

I think in addition to this change, we can infer extension from type of knexfile. If knexfile is typescript then in most cases user would want migrations/seeds etc. generated in ts as well. I'll create a new issue for it. That can be taken up separately.

types/index.d.ts
```diff
@@ -1622,7 +1622,7 @@ declare namespace Knex {
     nullable: boolean;
   }
 
-  interface Config {
+  interface Config<SV extends object = object> {
```
lorefnon: object is not a very useful default type because you can't really do anything with it. `SV extends {} = any` would be something more convenient.

bensaufley: Makes sense to me. I don't actually tend to use `object` in my own code so I'm not sure what possessed me to do it here 😄

kibertoad:
Released in 0.20.2

PR #3537, opened 2019-11-12
Dispose oracle connection on connection error

abskmj:
Similar to what is already implemented for queries with `resultSet` option as true. This option is set for select queries.

https://github.com/knex/knex/blob/f87b28c3e3a6e2624d83ab9c4c68ecffc913aee5/lib/dialects/oracledb/index.js#L139-L150

Previously, the connection was disposed on connection error only for select queries. After this change, it will be disposed on connection error for any type of query. This will any non-usable connections from the connection pool.

### DIFF
--- a/lib/dialects/oracledb/index.js
+++ b/lib/dialects/oracledb/index.js
@@ -176,7 +176,21 @@ Client_Oracledb.prototype.acquireRawConnection = function() {
             fetchRowsFromRS(connection, result.resultSet, numRows);
           });
         } else {
-          connection.execute(sql, bindParams || [], options, cb);
+          connection.execute(sql, bindParams || [], options, function(
+            err,
+            result
+          ) {
+            if (err) {
+              // dispose the connection on connection error
+              if (isConnectionError(err)) {
+                connection.close().catch(function(err) {});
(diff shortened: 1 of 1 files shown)


kibertoad:
Could you please add test for this?

kibertoad:
@elhigu Does this look correct to you?

abskmj:
@kibertoad Could you please point me to the test cases written for this file? I couldn't find one

kibertoad:
@abskmj Probably https://github.com/knex/knex/blob/master/test/unit/dialects/oracledb.js makes most sense.

abskmj:
@kibertoad it's quite complex to set up a test case for this with Sinon. I would need some help with it.

kibertoad:
@elhigu Any recommendations?

elhigu:
Is there any ways to write an integration test instead of mocking? for example running query in one connection and terminating that connection through other connection?

kibertoad:
@jakecoffman Any chance you could review this PR? It is similar to the one you raised, and I wonder if we could release both fixes together.

kibertoad:
Included in #3611

kibertoad:
Released in knex@0.20.7

PR #3545, opened 2019-11-19
Initialize data to buffer instead of string for non-strings

MaliaGuerrero:
A variable to hold the Stream data is initialized to a string.  In the case that the type of the stream is not "string", data is concat'ed to a Buffer, which fails.  This changes initializes data to a Buffer in case that it is not a string.

### DIFF
--- a/lib/dialects/oracledb/index.js
+++ b/lib/dialects/oracledb/index.js
@@ -341,10 +341,7 @@ Client_Oracledb.prototype._query = function(connection, obj) {
  */
 function readStream(stream, type) {
   return new Promise((resolve, reject) => {
-    let data = '';
-    if (type !== 'string') {
-      data = Buffer(0);
-    }
+    let data = type === 'string' ? '' : Buffer(0);
 
     stream.on('error', function(err) {
       reject(err);
--- a/lib/dialects/oracledb/index.js
+++ b/lib/dialects/oracledb/index.js
@@ -342,6 +342,9 @@ Client_Oracledb.prototype._query = function(connection, obj) {
 function readStream(stream, type) {
   return new Promise((resolve, reject) => {
     let data = '';
+    if (type !== 'string') {
+      data = Buffer(0);
+    }
 
     stream.on('error', function(err) {
       reject(err);
--- a/test/unit/dialects/oracledb.js
+++ b/test/unit/dialects/oracledb.js
@@ -111,17 +111,14 @@ describe('OracleDb parameters', function() {
         });
     });
 
-    it('on blob', function() {
-      return knexClient
-        .raw(
-          'select TO_BLOB(\'67c1a1acaaca11a1b36fa6636166709b\') as "field" from dual'
-        )
-        .then(function(result) {
-          expect(result[0]).to.be.ok;
-          expect(result[0].field.toString('hex')).to.be.equal(
-            '67c1a1acaaca11a1b36fa6636166709b'
-          );
-        });
+    it('on blob', async () => {
+      const result = await knexClient.raw(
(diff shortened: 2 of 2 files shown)


MaliaGuerrero:
I'd love some help on running tests for this.  I can run `npm test` but there's no DB running so things fail.  I looked around for a test DB but couldn't find it.

elhigu:
https://github.com/knex/knex/blob/master/CONTRIBUTING.md#the-easy-way

There is docker-compose file for starting up test databases.

lib/dialects/oracledb/index.js
```diff
@@ -342,6 +342,9 @@ Client_Oracledb.prototype._query = function(connection, obj) {
 function readStream(stream, type) {
   return new Promise((resolve, reject) => {
     let data = '';
+    if (type !== 'string') {
```
kibertoad: why not only initialize data once, using ternary operator? (`let data = type === 'string'? '' : Buffer(0) `)

test/unit/dialects/oracledb.js
```diff
@@ -111,6 +111,19 @@ describe('OracleDb parameters', function() {
         });
     });
 
+    it('on blob', function() {
+      return knexClient
```
kibertoad: can you use async/await instead?

test/unit/dialects/oracledb.js
```diff
@@ -111,6 +111,19 @@ describe('OracleDb parameters', function() {
         });
     });
 
+    it('on blob', function() {
```
kibertoad: can you make it an arrow function?

test/unit/dialects/oracledb.js
MaliaGuerrero: This is the style of this entire file.  Is there a push to get all the files to use async away?  And also, are tests meant to be used on older versions?

test/unit/dialects/oracledb.js
MaliaGuerrero: Same comment as above.

test/unit/dialects/oracledb.js
kibertoad: we only support Node 8+, and will convert code eventually. new one is preferred in newer syntax.

lib/dialects/oracledb/index.js
MaliaGuerrero: I'm trying to follow the way it was before: https://github.com/knex/knex/commit/c9e305760d622071e1aaa6d8069cab25366d3b9b#diff-a86ace7e0baa34caa6f981181dc544e2

I'll see if ternary operators are common and implement if so.

kibertoad: Ternary operators are pretty idiomatic in JS. And knex old code def not a good example of how to write code 😂

MaliaGuerrero: Makes sense.  I'm a stickler for code style, but I get what you are saying.  :)

MaliaGuerrero: Fixed!  Let me know if there's anything else!

kibertoad:
Thanks!

kibertoad:
Released in 0.20.3

noisyui:
A related bug is that calling .stream() on a query to Oracle db return each chunk as Array, and the buffer column is returned as a Lob object as an entry in the array. And the query context  is not repected when calling the .stream() method.

kibertoad:
@noisyui Would you consider sending a PR?

noisyui:
@kibertoad I hope I can, but I don't think I'm the right person to do this at this moment, who grab this library for several days. Thank you!

PR #3547, opened 2019-11-20
Add connection string qs to connection params

AndreiIgna:
Like mentioned in here https://github.com/knex/knex/issues/2354#issuecomment-555968462 this will allow query string params to be passed to driver used. It already works like this in other libraries, and seems like it be helpful for multiple cases.

Example in MySql https://github.com/mysqljs/mysql/blob/master/lib/ConnectionConfig.js#L186

### DIFF
--- a/lib/util/parse-connection.js
+++ b/lib/util/parse-connection.js
@@ -56,7 +56,7 @@ function connectionObject(parsed) {
     }
   }
   if (parsed.query) {
-    for (var key in parsed.query) {
+    for (const key in parsed.query) {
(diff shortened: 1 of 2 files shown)


kibertoad:
Thanks! Could you add test for this?

AndreiIgna:
Added, tests seem to be failing for an unrelated part than this change

kibertoad:
Released in 0.20.3

AndreiIgna:
That's great 👍 just tested it and works fine in my project

PR #3551, opened 2019-11-24
Support 'only' option in type definitions

ooflorent:
The `only` option was missing from the type definitions.
So here is it.


kibertoad:
@lorefnon

kibertoad:
Released in 0.20.3

PR #3552, opened 2019-11-27
Remove bluebird.using

maximelkin:


maximelkin:
I dont understand acquireConnection implementation:
1) in base it declared with arguments (client, config, txid)
2) also it called in same file with same arguments
3) this method also overriden in mssql, oracle, oracledb, but they have only one argument - config
4) also where are calls to Client.prototype.acquireConnection, but this not so interesting

kibertoad:
@maximelkin Probably it works almost by accident.
See in mssql, for an example:

```
const configConnection = config && config.connection; // this will fail to resolve because client is passed instead of a config
<...>
      try {
        resolve(
          (t.outerTx ? t.outerTx.conn : null) ||
            configConnection ||
            t.client.acquireConnection()
        );
      }
// we will resolve this with a client mistakenly marked as configConnection 
<...>
conn.tx_ = conn.transaction(); // luckily for us, `client.js` has `transaction()` method defined, so probably that works and the whole monster works
```

Frankly, I think this is wrong and should be refactored ruthlessly.

lib/runner.js
```diff
@@ -79,9 +79,7 @@ Object.assign(Runner.prototype, {
     const stream = new PassThrough({ objectMode: true });
```
kibertoad: One `using` was missed at line 57.

kibertoad: Oh, wait, sorry, you only changed `runner.js`, `transaction.js` remains as-is.

kibertoad:
@maximelkin Opened https://github.com/knex/knex/pull/3554 to make sure that code is at least not lying - it could actually be that `client.connection` is what was intended to be consumed in the first place, so I didn't touch actual implementation yet, since it seems to be working. If CI passes, I'll merge it so that it's less confusing.

kibertoad:
@maximelkin #3554 was merged, you can rebase to de-confuse code :D

lib/transaction.js
```diff
@@ -208,32 +205,32 @@ class Transaction extends EventEmitter {
   // Acquire a connection and create a disposer - either using the one passed
   // via config or getting one off the client. The disposer will be called once
   // the original promise is marked completed.
-  acquireConnection(client, config, txid) {
+  acquireConnection(config, cb) {
```
kibertoad: why get rid of `txid`? it was always passed from this.txid anyway?

maximelkin: Yes

maximelkin: https://github.com/knex/knex/pull/3552/files#diff-e652b5ccb2708f628a9de325cf105fb7R19

lib/runner.js
```diff
@@ -22,7 +22,7 @@ Object.assign(Runner.prototype, {
   run() {
     const runner = this;
     return (
-      Bluebird.using(this.ensureConnection(), function(connection) {
+      this.ensureConnection(function(connection) {
```
maximelkin: return native promise, conversion to bluebird promise done in lib/interface.js

lib/dialects/mssql/transaction.js
```diff
@@ -60,43 +60,47 @@ module.exports = class Transaction_MSSQL extends Transaction {
...
         );
       } catch (e) {
         reject(e);
       }
     })
-      .then(function(conn) {
-        conn.__knexTxId = t.txid;
-        if (!t.outerTx) {
-          t.conn = conn;
+      .then((conn) => {
```
kibertoad: would be awesome to convert those `then` chains into async/await, but separate PR would be better for that.

kibertoad:
Looks fantastic, thank you!

PR #3554, opened 2019-11-27
Clarify acquireConnection() signature

kibertoad:


maximelkin:
@kibertoad I didn't see any place of passing config.connection, except sqlite3 ddl
so this works because config parameter for .transaction method not documented and in project this used only for sqlite3, which does not override this method

It should be safe to replace signature to correct

kibertoad:
@maximelkin Did the change, let's see if CI passes :)

maximelkin:
@kibertoad fails because config is optional parameter, so it can be undefined

kibertoad:
@maximelkin You are right! Thanks, pushed the fix.

PR #3556, opened 2019-12-02
Add missing withKeyName in ColumnBuilder

iamchathu:
Add missing method withKeyName to ColumnBuilder interface

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1588,14 +1588,17 @@ declare namespace Knex {
     comment(value: string): ColumnBuilder;
     alter(): ColumnBuilder;
     queryContext(context: any): ColumnBuilder;
+  }
+  
+  interface ForeignKeyBuilder {
     withKeyName(keyName: string): ColumnBuilder;
   }
 
-  interface ForeignConstraintBuilder {
+  interface ForeignConstraintBuilder extends ForeignKeyBuilder {
     references(columnName: string): ReferencingColumnBuilder;
   }
 
-  interface MultikeyForeignConstraintBuilder {
+  interface MultikeyForeignConstraintBuilder extends ForeignKeyBuilder {
     references(columnNames: string[]): ReferencingColumnBuilder;
   }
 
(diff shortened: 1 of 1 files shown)


lorefnon:
Sorry, I somehow missed this. 

I don't think ColumnBuilder is the right place for this method.

This method, along with the two others [here](https://github.com/knex/knex/blob/79ccb7d8bd4443823c649a524734b6f78ef075ad/lib/schema/tablebuilder.js#L232-L243) should be added to ForeignConstraintBuilder & MultikeyForeignConstraintBuilder [here](https://github.com/knex/knex/blob/master/types/index.d.ts#L1596-L1602). We can create a common interface which both of them inherit from, and add these methods there.

iamchathu:
@lorefnon Updated the PR

lexa0303:
Any updates on this? Pretty bad to not have a proper typings :)

kibertoad:
@iamchathu Could you please rebase the PR?

types/index.d.ts
```diff
@@ -1589,12 +1589,16 @@ declare namespace Knex {
...
+    withKeyName(keyName: string): ColumnBuilder;
+  }
 
-  interface ForeignConstraintBuilder {
+  interface ForeignConstraintBuilder extends ForeignKeyBuilder {
     references(columnName: string): ReferencingColumnBuilder;
   }
 
-  interface MultikeyForeignConstraintBuilder {
+  interface MultikeyForeignConstraintBuilder extends ForeignKeyBuilder {
```
zeotuan, line 1601: With this implementation, only when foreign key is made with ```foreign``` method will they be able to use withKeyName. 
Since both interfaces contain `references(columnName:string):ReferencingColumnBuilder`, shoudn't the ```withkeyName```  be put into ```interface ReferencingColumnBuilder``` or at least have this interface extends another interface that have all those method so that when foreign key is made with only ```references``` method, they can also those method?

kibertoad:
@OlivierCavadenti Can you review this?

kibertoad:
oh wait, this was finished in separate PR

PR #3562, opened 2019-12-04
Fix queryContext function defintion

andrewburgess:
According to the docs, calling `queryContext` with no arguments returns the context object. (ref: https://knexjs.org/#Builder-queryContext)

This PR adds the relevant overload to the function definition for that use case.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1330,6 +1330,7 @@ declare namespace Knex {
     wrap<TResult2 = TResult>(before: string, after: string): Raw<TResult>;
     toSQL(): Sql;
     queryContext(context: any): Raw<TResult>;
+    queryContext(): any;
   }
 
   interface RawBuilder<TRecord extends {} = any, TResult = any> {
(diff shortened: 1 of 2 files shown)


kibertoad:
Thanks! Could you also update https://github.com/knex/knex/blob/master/types/test.ts?

types/index.d.ts
```diff
@@ -1330,6 +1330,7 @@ declare namespace Knex {
     wrap<TResult2 = TResult>(before: string, after: string): Raw<TResult>;
     toSQL(): Sql;
     queryContext(context: any): Raw<TResult>;
+    queryContext(): any;
```
kibertoad: would it be better defined as Record<string, any>?

andrewburgess:
> Thanks! Could you also update https://github.com/knex/knex/blob/master/types/test.ts?

Can do

types/index.d.ts
andrewburgess: I'm not sure, would that be for the return type?

From what I can tell it returns exactly what was put in when called with an argument, so I figured it would be best to just have those types match

kibertoad: ok, fair :). does it work correctly if user passes e. g. string?

andrewburgess: Yep, this is the underlying implementation: https://github.com/knex/knex/blob/988fb243898d746a759d422762685a79eddf99ca/lib/helpers.js#L79-L85

And the docs say the queryContext is shallowly cloned when a query builder is cloned:
https://github.com/knex/knex/blob/f87b28c3e3a6e2624d83ab9c4c68ecffc913aee5/lib/query/builder.js#L93
which uses [`lodash.clone`](https://lodash.com/docs/4.17.15#clone) and should work on any passed in

kibertoad:
thanks!

kibertoad:
Released in 0.20.4

PR #3566, opened 2019-12-07
Fix debug logger messing up queries with %

### LINKED ISSUE #3564: Dynamic LIKE is Impossible
# Environment

Knex version: 0.20.2
Database + version: MS SQL Server

# Bug

It seems there is no way to do a dynamic like.

```javascript
const q = 'foo';
const term = `%${q}%`;
.whereRaw('target_name like ?', [term]) // rest of the query omited for clarity
```
Logs this:

> knex:bindings [ 'undefinedoolundefined] undefined

I tried many other ways:

```.whereRaw('target_name like ?', [`\%${q}\%`])```
> knex:bindings [ 'NaNoo%']

And:

```.where('target_name', 'like', `%${q}%`)```
>  knex:bindings [ 'NaNoo%' ]

And:

```.whereRaw(`target_name like '%${q}%'`)```
> where target_name like 'NaNoo%'

And:

````.whereRaw(`target_name like '\%${q}\%'`)````
> where target_name like 'NaNoo%'

If it's not a bug, I guess the documentation on how to do this is missing or hard to find.

### PR DESCRIPTION
**Pull request by lorefnon**
Resolves #3564

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -27,8 +27,11 @@ const { uniqueId, cloneDeep, defaults } = require('lodash');
 const Logger = require('./logger');
 
 const debug = require('debug')('knex:client');
-const debugQuery = require('debug')('knex:query');
+const _debugQuery = require('debug')('knex:query');
 const debugBindings = require('debug')('knex:bindings');
+
+const debugQuery = (sql, txId) => _debugQuery(sql.replace(/%/g, '%%'), txId)
+
 const { POOL_CONFIG_OPTIONS } = require('./constants');
 
 // The base client provides the general structure

### CODE REVIEW

File: lib/client.js [hunk 1]
```diff
@@ -27,8 +27,11 @@ const { uniqueId, cloneDeep, defaults } = require('lodash');
 const Logger = require('./logger');
 
 const debug = require('debug')('knex:client');
-const debugQuery = require('debug')('knex:query');
+const _debugQuery = require('debug')('knex:query');
 const debugBindings = require('debug')('knex:bindings');
+
+const debugQuery = (sql, txId) => _debugQuery(sql.replace(/%/, '%%'), txId)
```
kibertoad: wouldn't that mess up queries that had %% in them already? or that is not a realistic case? maybe we can limit regexp to just single instances of %?

lorefnon: I intended this to be `/%/g`. 

Although I cant imagine a situation where we would have multiple consecutive % in an sql query, if we do, I think we would always want them to be reported as many as there are. 

```
> debug('x')('%foo%')
  x NaNoo%

> debug('x')('%%foo%%')
  x %foo% +0ms

> debug('x')('%foo%'.replace(/%/g, '%%'))
  x %foo% +0ms

> debug('x')('%%foo%%'.replace(/%/g, '%%'))
  x %%foo%% +0ms
```

kibertoad: fair :)

PR #3579, opened 2019-12-12
Add create/drop schema methods to SchemaBuilder

arkie:


arkie:
FYI @kibertoad

kibertoad:
Thank you! Improved our error handling a bit so that users who would try to use these types on other dialects wouldn't be confused.

kibertoad:
Released in knex@0.20.7

PR #3580, opened 2019-12-14
Pass date objects through to the MySQL driver instead of modifying them

### LINKED ISSUE #3577: Time zone configuration is not used for insert causing dates to be inserted improperly
# Environment

Knex version: 0.16.3
Database + version: mysql 5.7
OS: MacOS 15.1

# Bug

When inserting into the database, knex ignores the timezone configuration and dates/times are always assumed to be local.

2. Error message

N/A

```
var knex = require('knex')({
        client: "mysql",
        connection: {
            ...,
            timezone: "UTC"
        },
);

const date = new Date(2019, 11, 10, 10)

await knex('table').insert([
    { date, name: "Pete" },
    { date, name: "Jeff" },
]);
```

If this code runs on a computer that is not in the UTC timezone, the inserted time is incorrect. The date string that should be in the insert query should be `'2019-12-10 18:00:00'` for a computer that is in the -08:00 offset, however, knex instead inserts the date as `'2019-12-10 10:00:00'`. 

I took a quick look through the source and it appears that in most/all? of the toQuery/toSQL calls (which do take a timezone as a parameter), the timezone from the configuration is not passed.

jordansoltman:
This resolves #3577 by making knex pass dates through to the driver instead of modifying them. However, this causes a single test failure that raises a question.

Since the dates are not being encoded by knex anymore, the `toString()` (aka `toQuery()`) methods get the raw `Date` object which causes it to print out the default Date `.toString()` value in the query.

I can't think of an easy way to work around this since the date actually gets modified by the driver... but since `toString()` according to the docs is "Useful for debugging, but should not be used to create queries for running them against DB." maybe this is okay? If so I can modify the test.

```
AssertionError: expected 'select * from `users` where updtime = \'2016-01-05 10:19:30.599\'' to equal 'select * from `users` where updtime = \'Tue Jan 05 2016 10:19:30 GMT+0000 (Coordinated Universal Time)\''
      + expected - actual
      -select * from `users` where updtime = '2016-01-05 10:19:30.599'
      +select * from `users` where updtime = 'Tue Jan 05 2016 10:19:30 GMT+0000 (Coordinated Universal Time)'
```

### DIFF
--- a/lib/dialects/mysql/index.js
+++ b/lib/dialects/mysql/index.js
@@ -52,7 +52,12 @@ Object.assign(Client_MySQL.prototype, {
     return new Transaction(this, ...arguments);
   },
 
-  _escapeBinding: makeEscape(),
+  _escapeBinding: makeEscape({
+      // Let mysql handle dates in the driver
+      escapeDate(val) {
+          return val;
+      }
+  }),
 
   wrapIdentifierImpl(value) {
     return value !== '*' ? `\`${value.replace(/`/g, '``')}\`` : '*';


jordansoltman:
I’m not really sure how the other drivers handle dates, but I’ll look into them, as well as some tests.

PR #3581, opened 2019-12-15
Separate each alter column into a separate statement in MSSQL

philipcullen:
MSSQL does not allow more than one column to be altered in a single statement. This patch overrides the default behaviour that combines them and will instead generate separate statements for each.

For example, this changes the behaviour of:

```
knex.schema.alterTable('test', function(t) {
     t.timestamp('foo').alter();
     t.timestamp('bar').alter();
})
```

from

`ALTER TABLE [test] ALTER COLUMN [foo] datetime2, [bar] datetime2`

to

```
ALTER TABLE [test] ALTER COLUMN [foo] datetime2
ALTER TABLE [test] ALTER COLUMN [bar] datetime2
```

### DIFF
--- a/lib/dialects/mssql/schema/tablecompiler.js
+++ b/lib/dialects/mssql/schema/tablecompiler.js
@@ -85,6 +85,20 @@ Object.assign(TableCompiler_MSSQL.prototype, {
 
   changeType() {},
 
+  alterColumns(columns, colBuilders) {
+    for (const col of columns.sql) {
+      this.pushQuery({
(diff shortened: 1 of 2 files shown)


jeremy-w:
I think this can be closed. It looks like the issue targeted by this PR has been fixed by another, judging from code addressing the problem here: https://github.com/knex/knex/blob/master/lib/dialects/mssql/schema/mssql-tablecompiler.js#L82-L95

Issue #3186, opened 2019-05-09
Crossdb support for INSERT ... ON CONFLICT / MERGE upsert


elhigu:
# Environment

# Feature discussion / request

1. Explain what is your use case

To support MERGE / ON CONFLICT features of databses.

2. Explain what kind of feature would support this

Full discussion about the feature is in https://github.com/tgriesser/knex/pull/2197#issuecomment-325934623

Conclusion by @jordansexton  was:

Okay. I'll finish the implementation for `DO NOTHING/UPDATE` on Postgres and `INSERT IGNORE/ON DUPLICATE KEY UPDATE` on MySQL in this PR, probably using two methods:
```javascript
insert({ ...inserts }).ignore([columns]?)
insert({ ...inserts }).merge({ ...updates }, [columns]?)
```

My reasoning is:

1. `replace` implies full replacement, whereas `merge` implies partial replacement (essentially, the semantics for a generic "object merge" are equivalent to what actually happens in the operation)
2. Only SQLite uses `REPLACE`. Since this PR won't support SQLite, and "replace" is a semantically different operation, it's probably better to have a different `replace` method (if supported at all).
3. The SQL standard uses `MERGE`. In the event that other RDBMS's converge upon the standard, `merge` is likely more future-proof than an ad-hoc alternative.
4. A followup PR to support MS SQL and Oracle will be able to use the same method and just work. I don't use either, but I'll take a stab at supporting both when this is settled. The second `columns` argument could be used to provide aliases for those instead.
3. Give some API proposal, how the feature should work

SherloxFR:
I think this feature is really important to implement, currently it seems I'm only able to do it this way:
```javascript
knex.raw(
  await knex
  .insert({
    a: 1,
    b: 2,
    c: 3,
    d: 4
  })
  .into('my_table')
  .toQuery() +
  await knex
  .raw(`
    ON DUPLICATE KEY UPDATE
      a = :a,
      b = :b,
      d = :d
  `, {
    a: 1,
    b: 2,
    d: 4,
  })
  .toQuery()
)
```

Surely we can all agree this is really an awful way to write SQL.

elhigu:
Removed a lengthy comment that basically said +1 and didnt add any value to the thread.

codeclown:
After taking a stab at implementing this, here's some more things to consider I unfortunately found out:

|          | can support ignore() | merge({ ... }, ['foo']) works | merge({ ... }, []) works |
| --- | --- | --- | --- |
[...]

PR #3593, opened 2019-12-24
Add raw support to QueryBuilder

### LINKED ISSUE #3032: RFC: Add `.raw` support to Builder instances analogous to `.whereRaw` et al.
# Environment

Knex version: any
Database + version: Azure SQL
OS: Linux

Hello Knex,

I would like to request that `.raw` be added to Builder instances for the purposes of appending dialect-specific statements to the end of built queries.

Hold onto your hats Ladies and Gentlemen.

My Knex instance is responsible for connecting to, and managing an SQL Azure database full of temporal tables (I'm the guy who wrote #2857).  Temporal tables in MSSQL allow you to roll-back the database and view it at any point in time and you access historical data by appending a `FOR SYSTEM_TIME AS OF [date]` statement to the end of your queries.

If the user has indicated they want to time-travel the database, then the program should apply this temporal statement to all of the outgoing queries. 

Consider the following Express API pseudo:

```js
function temporalMiddleware(req, res, next) {
    req.db.on('start', query => {
      if (req.timeTravelEnabled && req.timeTravelDate) {
         query.raw('FOR SYSTEM_TIME AS OF ?', req.timeTravelDate);
      }
    });
}
```
 Now, I know the documentation discourages modifying a builder in event emitters, but _the use of the start hook really is the best approach for this scenario_.  `knex.raw` would be well suited to this scenario if it did not return a new builder instance.  You cannot modify the `query` reference passed to the event listener.

My database API involves a lot of dialect-specific builder functions and a lot of them pass around a Knex instance specifically for the purpose of using Knex.raw: 

```js
function applySpecifics(knex) {
  return query => knex.raw('? SOMETHING_SPECIFIC', [query]);
}
```

Avoiding this deletes a fair bit of code smell in userland APIs:

```js
function applySpecifics() {
  return query => query.raw('? SOMETHING_SPECIFIC');
}
```

OR:

A context object be provided to event listeners to allow the `builder` property to be replaced with a different builder reference so that `knex.raw` may be used.  This is the easiest option, but is a fairly horrid breaking change which I can't imaging being popular.

Builder instances would support `.raw` methods which append raw statements to the built query when compiled.

The function synonymous with `knex.raw` be placed on `Builder.prototype` which allows raw statements to be appended to built statements at compile time.  The API should support bindings in the traditional fashion.

I'm happy to implement the feature, however I'm a bit of a fledgling when it comes to the query compiler and builder architecture, so let me know what you think.

willsoto:
- Introduces a new grouping named `raws` in order to support the new command
- Append only

Looking for feedback on the interface. If the implementation is acceptable, I can add more tests.

Closes #3032

### DIFF
--- a/lib/query/compiler.js
+++ b/lib/query/compiler.js
@@ -50,6 +50,7 @@ const components = [
   'offset',
   'lock',
   'waitMode',
+  'raws',
 ];
 
 assign(QueryCompiler.prototype, {
@@ -396,6 +397,20 @@ assign(QueryCompiler.prototype, {
     return sql.length > 1 ? sql.join(' ') : '';
   },
 
+  // Compiles all `raw` statements on the query.
+  raws() {
+    const raws = this.grouped.raws;
+    if (!raws) return;
+    const sql = [];
+    let i = -1;
+    while (++i < raws.length) {
+      const stmt = raws[i];
+      const val = this[stmt.type](stmt);
+      sql.push(val);
+    }
+    return sql.join(' ');
+  },
+
   group() {
     return this._groupsOrders('group');
   },
@@ -579,9 +594,7 @@ assign(QueryCompiler.prototype, {
   },
 
   distinctOn(value) {
-    throw new Error(
-      '.distinctOn() is currently only supported on PostgreSQL'
-    )
+    throw new Error('.distinctOn() is currently only supported on PostgreSQL');
   },
 
   // On Clause
@@ -704,6 +717,10 @@ assign(QueryCompiler.prototype, {
   // Compiles a "whereRaw" query.
   whereRaw(statement) {
     return this._not(statement, '') + this.formatter.unwrapRaw(statement.value);
+  },
+
+  raw(statement) {
+    return this.formatter.unwrapRaw(statement.value);
   },
 
   wrap(str) {
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -201,7 +201,7 @@ assign(Builder.prototype, {
     this._statements.push({
       grouping: 'columns',
       value,
-      distinctOn: true
+      distinctOn: true,
     });
     return this;
   },
@@ -866,6 +866,18 @@ assign(Builder.prototype, {
 
   orHavingRaw(sql, bindings) {
     return this._bool('or').havingRaw(sql, bindings);
+  },
+
+  raw(sql, bindings) {
+    const raw = sql instanceof Raw ? sql : this.client.raw(sql, bindings);
+    this._statements.push({
+      grouping: 'raws',
+      type: 'raw',
+      value: raw,
+      bool: this._bool(),
+      not: this._not(),
+    });
+    return this;
   },
 
   // Only allow a single "offset" to be set for the current query.
--- a/lib/query/methods.js
+++ b/lib/query/methods.js
@@ -88,4 +88,5 @@ module.exports = [
   'truncate',
   'transacting',
   'connection',
+  'raw',
 ];
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -3839,6 +3839,22 @@ describe('QueryBuilder', () => {
     );
   });
 
+  it('raw', () => {
+    testsql(
+      qb()
+        .select('*')
+        .from('users')
+        .whereRaw('1 = ?', [1])
+        .raw('FOR SYSTEM_TIME AS OF ?', ['12/24/2019']),
+      {
+        pg: {
+          sql: 'select * from "users" where 1 = ? FOR SYSTEM_TIME AS OF ?',
+          bindings: [1, '12/24/2019'],
+        },
+      }
+    );
+  });
+
   it('limits', () => {
     testsql(
       qb()


lib/query/compiler.js
```diff
@@ -396,6 +397,20 @@ assign(QueryCompiler.prototype, {
     return sql.length > 1 ? sql.join(' ') : '';
   },
 
+  // Compiles all `raw` statements on the query.
+  raws() {
```
elhigu: At some point there was meant to be a way to pass `?` to raw SQL string without need to escape it in case if no bindings were passed to the `.raw` call. So just concatenating SQL and bindings would break that feature.

Though I'm not sure if that was ever implemented.

willsoto:
> I would like to see different .raw calls to be bound to the query building stage

I agree. Any thoughts on how to achieve this?

> if builder does not have any where statements yet, then .raw() would append before .where part of the query etc.

Can you elaborate on this more? I wanna make sure I understand the intent...

> Another thing what worries me a bit is how to combine multiple raw statements.

Could you also discuss this point more?

willsoto:
@elhigu Any thoughts on my questions/comments? I am eager to take this to completion, I just want to make sure I am going about it the right way. Thanks!

elhigu:
>> if builder does not have any where statements yet, then .raw() would append before .where part of the query etc.
>Can you elaborate on this more? I wanna make sure I understand the intent...

>> I would like to see different .raw calls to be bound to the query building stage
> I agree. Any thoughts on how to achieve this?

I suppose that the most versatile way would be to have various .raw snippets stored in query builder and each of these snippets would have some dedicated place where in query they will be rendered. 

For example rawBeforeFrom, rawBeforeWhere etc. then when .raw() is given there could be another parameter to tell in which position it should be added... 

That would be a bit hacky implementation, but would make knex a lot more versatile and would allow users to pass much more DB specific declarations to queries that are mostly build with normal query builder APIs.

Better way to handle that would indeed be to completely rewrite query builder internal to have some universal graph representation of the query.

>> if builder does not have any where statements yet, then .raw() would append before .where part of the query etc.

> Can you elaborate on this more? I wanna make sure I understand the intent...

For example knex.raw('some query containing ?') would actaully send that string as SQL to DB driver instead of trying to convert it for example with pg driver to this `come query containing $1`.

But problem with that is when you do something like this:

knex.raw('? ?', [knex.raw('query 1 with ?'), knex.raw('query 2 with ?', 1)]) it probably just interpolates that raw to `sql: 'query 1 with ? query 2 with ?', bindings: [1]` and fails because of missing parameters.... I don't thing you don't need to resolve that for this issue (some automatic escaping feature for ? in some certain cases...)

I hope these clarified my earlier comments a bit.

willsoto:
Thanks for the elaboration.

Seeing as 

> Better way to handle that would indeed be to completely rewrite query builder internal to have some universal graph representation of the query.

Is out of scope for me 😅 is your suggestion is that I add methods along the lines of:

- `rawBeforeFrom`
- `rawBeforeWhere`
- `rawAfterFrom`
- `rawAfterWhere`
- `rawBeforeJoin`
etc

or have a single method `raw` that takes some options:

```js
knex("users").update().raw("on conflict (email)", {
  position: "beforeWhere"
}).where("condition");
```

Or some combination of both?

elhigu:
I would prefer having just single .raw method and have options as a third parameter. The same options could be also used for other stuff, like telling explicitly to knex that sql string questionmarks should not be replaced with parameter bindings.

wubzz:
> Better way to handle that would indeed be to completely rewrite query builder internal to have some universal graph representation of the query.

Completely agree with this.

```javascript
knex("users").update().raw("on conflict (email)", {
  position: "beforeWhere"
}).where("condition");
```

Completely disagree with this syntax.
<hr/>

It is way too ambiguous in my opinion, and feels like a hack to just append/prepend stuff at *certain 
**predefined** supported positions* of an sql in the current API.

Would make a lot more sense to implement support in the API for functions that fulfill some form of purpose. This specific example could've been

```javascript
knex('users')
    .where({...})
    .update({...})
    .onConflict([...], ...)
```

SQL-features that cannot/will not be covered due to being 'too uncommon' or whatever reason should instead be executed using the standard `Raw`. Feels like this adds more clutter to the `QueryBuilder` than it already has.

`Raw` is the complete opposite of `QueryBuilder` and its purpose is to be used for what the `QueryBuilder` cannot perform on its own. So trying to mix these two in the same cookie dough is very odd to me. It'll come out the oven as a very distasty cookie.

I cannot wrap my head around how this:

```javascript
knex('users')
    .select()
    .whereRaw('1 = ?', [1])
    .raw('FOR SYSTEM_TIME AS OF ?', ['12/24/2019'])
```

Is better, clearer, or in anyway an improvement over this:
```javascript
knex('users')
    .select()
    .whereRaw('1 = ? FOR SYSTEM_TIME AS OF ?', [1, '12/24/2019'])
```

Genuinely curious if there are any legitimate use-cases for this. Are there any examples besides the test?

elhigu:
There has been many cases where people would have needed to add raw snippets in certain places in query, which cannot have been added there without having to write query completely in single `.raw`call.

One place where it would used to be useful to have would have been to add `NOWAIT` or `SKIP LOCKED` declaration to the query (gladly there are builder methods now for that). I cannot remember any other places where I would have needed that from top of my head... Maybe in some upsert query.

I'm sure mssql / oracle etc. has more some strange additional declarations which can be given in only in certain points in query. Those places are neither so arbitary, since there are certain syntactical breakpoints in SQL where they make sense and may be really useful: 

`SELECT <place1> ...select clauses... <place2> FROM <place3> ...from clauses... <place4> WHERE ...`

All the places where those extra raw snippets should be able to be inserted should be carefully thought. 

This would only make and incremental improvement allowing to add that tiny bit of SQL that you still need to make your query to work without having to write everything as a raw query (which has also different result format than, queries made through query builder APIs).

Another place where people would have had that nice is to add some SQL before or after the whole query. For example comments having instruction to SQL load balancer. 

> Would make a lot more sense to implement support in the API for functions that fulfill some form of purpose. This specific example could've been

Yes, and those should be kept being implemented. It is complementary for having a generic way to be able to do the thing even when it does not have specific API there yet (like upsert functionality).

Those additional raw snippets are adding to the current `knex.raw()` which can be used to add custom SQL pretty much anywhere. except between some fixed parts in query building.

It is better to have even some way to do the query, instead of waiting for APIs for that specific functionality to appear in latest knex release.

> It is way too ambiguous in my opinion, and feels like a hack to just append/prepend stuff at certain
predefined supported positions of an sql in the current API.

There predefined positions are coming from the same places how query builder actually stores SQL that is being built.  Places can be designed so that they reduces maximally current limitations of knex.

Maybe hacky but should be carefully decided by looking where exactly in query building there are gaps, that does not allow passing any additional raw SQL in. I don't believe there are very many places like that.

willsoto:
@elhigu @wubzz - I do appreciate the conversation here and @wubzz's effort to make sure we aren't falling victim to the [xy problem](https://meta.stackexchange.com/questions/66377/what-is-the-xy-problem).

My goal in implementing this is ultimately to have the ability to have an upsert that leverages most of the query builder while not having to resort to making the whole thing `raw`. Admittedly, I could have probably just done #3186 but didn't feel like dealing with all the cross-db stuff 😅 

---

With that said, I do agree with @elhigu on this one. I think implementing a more flexible `raw` could help alleviate some short term issues for people. If enough people end up using similar enough `raw` statements, those can be rolled up into more bespoke methods as needed. 

---

Am I okay to move forward on the original proposed api?

elhigu:
@kibertoad @lorefnon @rhys-vdw @tgriesser any thoughts if extending knex's current raw functionality is a bad idea?

lorefnon:
> Better way to handle that would indeed be to completely rewrite query builder internal to have some universal graph representation of the query.

While I do think we should have a serializable representation of QueryBuilder (which doesn't break across versions), I don't believe it is an ideal solution for the use case here.

Because knex is responsible for converting this graph representation to SQL string, it will have to understand any dialect specific fields added by users in this graph representation. Otherwise it can lead to very brittle / error prone code if people are encouraged to append/prepend strings to the property values of this structure.

I think instead of extending raw with options like `{ position: "beforeWhere"}` a more flexible option is to make the [QueryCompiler](https://github.com/knex/knex/blob/master/lib/query/compiler.js) public, 
and make it accept (well documented) hooks for intercepting & transforming the strings it constructs at various places. 

The above is a bit vague, but if the approach sounds tenable at a high level I can check on implementing a poc this weekend.

elhigu:
That indeed sounds a bit more flexible, but remember that compiler hooks should be able to do transformation to both, SQL and bindings. Anyways that shouldn't be a problem, there is no reason why hooks couldn't transform both.

About makin QueryCompiler public, I wouldn't like to do that. It would expose unnecessary big surface of knex to public API and would make knex it more unstable and add number of breaking changes. Also knex should allow to set those hooks differently for each query, so having single extended QueryCompiler doesn't sound like a way to go. I don't see any benefit on publishing QueryCompiler completely.

So IMO API for setting those compiler hooks should be done through normal query builder `knex('table').compilerHook(...)` API or globally for all the queries like `knex.compilerHook(...)`. Pretty much the same way that currently events are allowed to be set.

lorefnon:
>  Also knex should allow to set those hooks differently for each query, so having single extended QueryCompiler doesn't sound like a way to go. 

Yeah I didn't mean a single extended QueryCompiler. I just meant that people could configure/extend  QueryCopiler and then pass it to QueryBuilder. 

But I see the point about stability and unnecessarily large public API. The compilerHook api sounds good to me.

willsoto:
Seems like there is general consensus to go a different direction on this. So I’ll close this pr since it sounds like it will be far more involved. Thanks for the discussion!

PR #3599, opened 2019-12-29
Update deprecated `dialect` to `client` in code example

JeremyBernier:
The code sample in the README.md uses a deprecated config option. When using `dialect`, one gets the warning: "Using 'this.dialect' to identify the client is deprecated and support for it will be removed in the future. Please use configuration option 'client' instead."


kibertoad:
thanks :)

PR #3604, opened 2020-01-02
fix(cli): prevent warning showing up with root command

### LINKED ISSUE #3603: [CLI]: warning message showing up on passing in the root command
Typing in `knex` is expected to show up the help information and exit. But, as of now a warning message shows up which isn't intended.

jamesgeorge007:
closes #3603 

Tweaked the code-base to make use of `commander.js` API that handles the scenario when no arguments were passed in.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -341,12 +341,11 @@ function invoke(env) {
         .catch(exit);
     });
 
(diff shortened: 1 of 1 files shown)


kibertoad:
Thanks!

PR #3606, opened 2020-01-03
fix #3605 not releasing connection from pool on disconnect

### LINKED ISSUE #3605: oracledb not releasing connection from pool after disconnect
# Environment

Knex version: 0.20.6
OS: Linux & OSX

@atiertant 

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do:

When oracledb falls of the network or the application can't connect and there's an open transaction, the pool doesn't empty. After the pool is full we see `Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?` and the application cannot be used until we restart it. 

2. Error message

`Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?`

See https://github.com/jakecoffman/knex-oracle-bug

The crux is: 

```
  console.log("Started transaction")

  await sleep(5000)
  console.log("Committing")

  try {
    await trx.commit()
  } catch (e) {
    console.log("Error committing:", e)
    await knex.destroy()
    return
  }

  console.log("Ok")
```

If you disconnect during the sleep then the `commit()` still finishes immediately (I expected it would block and throw on reject but it does not) and after a minute you see an Unhandled rejection and the connection isn't freed from the pool.

jakecoffman:
I took the liberty to use async/await as the code is much cleaner. Let me know if you want a less drastic change. I confirmed this works correctly against a live oracle database.

### DIFF
--- a/lib/dialects/oracledb/transaction.js
+++ b/lib/dialects/oracledb/transaction.js
@@ -70,18 +70,17 @@ module.exports = class Oracle_Transaction extends Transaction {
       } finally {
         debugTx('%s: releasing connection', this.txid);
         connection.isTransaction = false;
-        connection
-          .commitAsync()
-          .catch(function(err) {
-            t._rejecter(err);
-          })
-          .finally(function() {
-            if (!configConnection) {
-              t.client.releaseConnection(connection);
-            } else {
-              debugTx('%s: not releasing external connection', t.txid);
-            }
-          });
+        try {
+          await connection.commitAsync();
+        } catch (err) {
+          t._rejecter(err);
+        } finally {
+          if (!configConnection) {
+            await t.client.releaseConnection(connection);
+          } else {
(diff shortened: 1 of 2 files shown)


kibertoad:
Using async/await is definitely preferred, thank you! Could you please add test for this? (doesn't have to be integration test, just checking release call is being made would suffice)

jakecoffman:
It seems the change was too drastic and broke many tests so I'm trying a precision strike.

jakecoffman:
Interesting the build failed but the build on my repo passed https://travis-ci.org/jakecoffman/knex

kibertoad:
oh, that one is well-known to be flaky test, just ignore it. still would be nice to get a new test that was failing prior to this change, to make sure we don't break it again.

kibertoad:
thanks a lot!

PR #3611, opened 2020-01-05
dispose Oracle connection on connection error

jakecoffman:
added a test for PR #3537


kibertoad:
Wow! Thank you so much!

jakecoffman:
I'd like to fix #3612 too but I probably can't get to it until tomorrow.

kibertoad:
@jakecoffman Should I wait with releasing a new version of knex until then?

jakecoffman:
@kibertoad  It looks like there is more that I don't understand yet. You shouldn't wait for me to figure this out, might be a while! Thanks

kibertoad:
Released in knex@0.20.7

PR #3613, opened 2020-01-05
Improve Support for Liftoff's Preloaders

briandamaged:
Revises the Knex CLI's initialization logic so that it takes better advantage of Liftoff's module pre-loading.  More specifically:

- Removed the unsupported `"knexfile"` and `"knexpath"` arguments from the call to `Liftoff#launch(..)`.
- Removed Knex's custom logic for config file discovery.  (In certain cases, this logic would fail to discover configuration files unless they were either Javascript or Typescript.  Furthermore, this logic would never invoke Liftoff's module pre-loaders, which could then lead to syntax errors when opening the files)

Importantly: this PR changes the CLI's behavior when both the `--cwd` and `--knexfile` parameters are specified.  Previously, the location of the `--knexfile` was always relative to the value specified via `--cwd`.  Once this PR is merged, the two parameters will be independent of one another.  This better aligns w/ the semantics used by Liftoff.  For example:

```shell
knex --knexfile /path/to/shared/knexfile.js --cwd /var/www/project1 migrate:latest
knex --knexfile /path/to/shared/knexfile.js --cwd /var/www/project2 migrate:latest
```


bin/cli.js
```diff
@@ -39,37 +54,6 @@ function initKnex(env, opts) {
...
-      );
-    }
-
-    if (!env.configuration.ext) {
-      env.configuration.ext = path
-        .extname(resolvedKnexfilePath)
-        .replace('.', '');
-    }
-  }
-
```
briandamaged: This logic was essentially duplicating the "configuration file discovery" logic that was already provided by Liftoff.  However, modules discovered via this control path would never utilize any of Liftoff's module pre-loaders.  This is the reason that some users were observing syntax errors when attempting to load non-JS config files via the `--knexfile` parameter.  Ex: #2998

bin/cli.js
```diff
@@ -367,8 +351,7 @@ cli.on('requireFail', function(name) {
 cli.launch(
   {
     cwd: argv.cwd,
-    knexfile: argv.knexfile,
-    knexpath: argv.knexpath,
+    configPath: argv.knexfile,
```
briandamaged: The `Liftoff#launch(..)` method does not understand the `knexfile` and `knexpath` parameters.  Furthermore, it does not pass them though to the `env` object when calling the `invoke(..)` callback.

bin/utils/cli-config-utils.js
```diff
@@ -8,9 +8,8 @@ const argv = require('getopts')(process.argv.slice(2));
 
 function mkConfigObj(opts) {
   if (!opts.client) {
-    const path = resolveDefaultKnexfilePath();
     throw new Error(
-      `No default configuration file '${path}' found and no commandline connection parameters passed`
+      `No configuration file found and no commandline connection parameters passed`
```
briandamaged: This error message was rendering the config file name as `knexfile.undefined`.  This is because the `resolveDefaultKnexfilePath()` was expecting an argument for the desired extension.

However, Knex actually supports many different config file extensions, including `.ts` and `.coffee`.  So, it didn't really make sense to just show the `.js` extension.  Therefore, I removed the name from the error message.

bin/utils/cli-config-utils.js
```diff
@@ -31,34 +30,6 @@ function mkConfigObj(opts) {
...
-    `Failed to find configuration at default location of ${resolveDefaultKnexfilePath(
-      'js'
-    )}`
-  );
-}
-
-function resolveDefaultKnexfilePath(extension) {
-  return process.cwd() + `/knexfile.${extension}`;
-}
-
```
briandamaged: Dead code.  (This detail is handled by Liftoff now)

briandamaged:
FYI: This code currently breaks 1 unit test.  Specifically:

```
  6) CLI tests
       resolves knexfile correctly with cwd specified:
     node /home/brian/src/arclia/knex/bin/cli.js migrate:latest --cwd=test/jake-util/knexfile --knexfile=knexfile.js -> FAIL. 
Stdout:  
Error: /home/brian/src/arclia/knex/bin/utils/cli-config-utils.js:11
    throw new Error(
    ^
```

This error occurs because Liftoff treats `--knexfile` and `--cwd` as orthogonal parameters.  More specifically, `--knexfile` is _not_ evaluated relative to `--cwd`.

kibertoad:
@briandamaged I am somewhat concerned about breaking scripts for existing users, but I guess just one case with cwd is not too bad, as long as we publish it as a major version. Can you add one more cli-testlab-based test that was failing prior to this change, so that we don't break your fix again in the future? I guess adding ts file would work.

briandamaged:
@kibertoad : Sure!  I just wanted to get your feedback on the breaking change before moving forward w/ the other test cases.  I should have some time to write them in the next day or so.

briandamaged:
@kibertoad : I've added some new tests, but I'm noticing a discrepancy.

When both the `--cwd` and `--knexfile` flags are specified, it is changing the process's CWD to `path.dirname(knexfile)` (instead of honoring the `--cwd` value that was passed in explicitly).  It looks like this is potentially a bug in the Liftoff library itself:

https://github.com/js-cli/js-liftoff/issues/55

Anyway, I'll do a bit more digging into this tomorrow.

briandamaged:
@kibertoad : A closely-related question: what is the intention of the `cwd` value?  Is it supposed to represent:

- The directory that contains the Knexfile?
- The directory that contains the developer's application?  (ie: project root)
- An arbitrary directory based upon the specific project's needs?
- Something else?

kibertoad:
@briandamaged Prior discussions on this subject: https://github.com/knex/knex/pull/2959 and https://github.com/knex/knex/issues/2952
I would say that CWD is arbitrary point that is the root from which other paths are calculated. Important behaviour to preserve as this is how knex used to work historically - paths to seeds and migrations are resolved relatively to knexfile. Path to knexfile, however, should be resolved relatively to CWD, whatever user decides to pass as such.

kibertoad:
@briandamaged Would you like any help with this?

briandamaged:
@kibertoad : Howdy, and thx for following up!  I got pulled into something else during the last few weeks, so I haven't been able to investigate this any further.  However, I should have some time in the next few days.

Once I get back to this, I'll create a table that compares the "Old Behavior" and "New Behavior" for various scenarios.  From there, we can then discuss what the desired behavior in each scenario _should_ be, and implement things accordingly.  (Saying it another way: I think the "New Behavior" fixes certain problems, but it might introduce new problems as well.  So, we need to make sure we agree upon the requirements / expectations).

briandamaged:
@kibertoad : Alright, here's the table demonstrating the behaviors for various command line arguments:

https://docs.google.com/spreadsheets/d/19TMouhxdIosYGzubO-2lbxm8Ykx4jR-Iu74HXohFYMg/edit?usp=sharing

As you can see:

1. The `master` branch is effectively ignoring the `--knexfile` input.  This is the reason that `ts-node` was not being registered properly when loading Typescript migrations.
2. In the `fix/knexfile-cli-arg` branch, if `--cwd` is omitted, then the value of `env.cwd` will be inferred from `--knexfile`.  So, this aligns nicely w/ the "All paths within the Knexfile are relative to the Knexfile's directory" requirement that was mentioned... somewhere?  (I can't seem to find the documentation on it now)

So, based upon my understanding of the requirements, it looks like the behavior in `fix/knexfile-cli-arg` is correct.

briandamaged:
As for the travis-ci failure: it looks like the `ts-node` dev dependency probably was not installed for that particular scenario.  Is there an easy way to confirm that?

briandamaged:
@kibertoad : Okay, new update around the travis-ci error:  it appears that `nyc` uses the following default:

```
  --extension, -e             a list of extensions that nyc should handle in
                              addition to .js
                   [string] [default: [".js",".cjs",".mjs",".ts",".tsx",".jsx"]]
```

This causes the value of `require.extensions` to be modified once `nyc` hands control off to `mocha`.

The problem: notice that `".ts"` is already on that list.  This causes `Liftoff` to believe that Typescript support has already been registered properly.  Consequently, it decides that it does _not_ need to register `ts-node`, which then causes 🔥  when it attempts to open the Knexfile.

The "quick fix" would be to restrict `nyc` to a specific set of extensions; however, I suspect that will introduce more problems in the future.  (Especially if portions of Knex ever get rewritten in Typescript).  I'm also starting to think that `Liftoff` might be introducing more problems than it actually solves.

package.json
```diff
@@ -189,7 +190,10 @@
     "lines": 84,
     "statements": 82,
     "functions": 83,
-    "branches": 69
+    "branches": 69,
+    "extension": [
+      ".js"
+    ]
```
briandamaged: FYI: This is the `nyc` fix that I mentioned here:

https://github.com/knex/knex/pull/3613#issuecomment-579990257

Short story: it appears that `nyc`'s extension handling interferes with `Liftoff`'s extension handling.  Thus, when you run the unit tests under `nyc`, it fails to preload `ts-node` properly.

kibertoad:
Thanks! I'll take a look tomorrow.

briandamaged:
@kibertoad : FYI:  I just added a couple more rows to the spreadsheet.  These ones are for:

* An absolute `--cwd` path
* A relative `--knexfile` path

As you can see, the two branches behave a bit differently for this specific scenario.  So, which of these behaviors do we want to keep?

briandamaged:
@kibertoad : Forgot to mention: I think I figured out an adjustment to the new code that will allow it to recreate the old behavior.  To recap, let's say that:

* Shell's CWD:   `/private/tmp/foo`
* `--cwd` : `/private/tmp/bar`
* `--knexfile` : `./knexfile.js`

In this case, the CLI will ultimately resolve the Knexfile at:

* Old Behavior:  `/private/tmp/bar/knexfile.js`
* New Behavior: `/private/tmp/foo/knexfile.js`

Personally, I think that the New Behavior makes more sense.  That is: paths are always relative to the Shell's CWD.

kibertoad:
@briandamaged I agree that this makes more sense; however, this would be a breaking change and would require bumping second number. Could you also update https://github.com/knex/knex/blob/master/UPGRADING.md explaining under which circumstances users are likely to experience change in behaviour, and what they would need to change to get same result as before?

kibertoad:
If we can keep the way to recreate the old behaviour, though, that would be very helpful, as in general `knex` tries to be as conservative as possible with breaking changes.

briandamaged:
@kibertoad : Alright -- I went ahead and implemented a work-around that recreates the "classic" behavior.  So:

* If `--cwd` is specified, then `--knexfile` will be resolved relative to `--cwd`.  Otherwise, `--knexfile` will be resolved relative to the shell's CWD.
* Regardless of how `--knexfile` gets resolved, it will always invoke the appropriate Liftoff preloaders.  (ex: it will recognize that the file is Typescript and register the `ts-node` module)

bin/cli.js
```diff
@@ -368,11 +351,21 @@ cli.on('requireFail', function(name) {
...
+//      but we decided to retain the behavior for backwards-compatibility.  In
+//      particular: if `--knexfile` is a relative path, then it will be resolved
+//      relative to `--cwd` instead of the shell's CWD.
+//
+//      So, the easiest way to replicate this behavior is to have the CLI change
+//      its CWD to `--cwd` immediately before initializing everything else.  This
+//      ensures that Liftoff will then resolve the path to `--knexfile` correctly.
+if (argv.cwd) {
+  process.chdir(argv.cwd);
+}
```
briandamaged: @kibertoad : Here's the work-around that allows the CLI to retain the "classic" behavior.

briandamaged:
@kibertoad : Howdy again!  Just following up to see if there's anything you need on my side.  Thx!

kibertoad:
@briandamaged Sorry, just started at new job and got distracted. This looks great and I would be happy to merge this. Any reasons why you kept WIP in title?

briandamaged:
@kibertoad : Ah -- nope.  I just forgot to remove the `WIP:` prefix.  Lemme take care of that now...

Good luck on your new job, btw!  Doing something interesting, I hope?

kibertoad:
@briandamaged Thanks! Oh yeah, brand new team, greenfield project, I'm extremely hyped about it. The only unfortunate thing is that I switched stacks and returned to Java world, so there'll be less eating of own's dog food, unfortunately.

briandamaged:
@kibertoad: Ah -- oh well.  I've heard that the Java language / ecosystem has improved significantly over the last several years.  So, it might not be _that_ painful.  Plus, there are several languages that compile into Java bytecode, such as Kotlin and Scala.  So, you might be able to integrate those into whatever you're building.

kibertoad:
Well yeah, convincing folks over there to use Kotlin is one of my current dreams :D

PR #3615, opened 2020-01-07
#3751: Esm interop flag

D10221:
Enable esm interop behind a flag
Enable experimental, optional use
Minimum changes
Tests: migration/seeds/configuration


D10221:
I'll fix the build error, Node 8, eslint complains of syntax error, "import"

PR #3617, opened 2020-01-08
Print help only when there are no arguments

yeonhoyoon:
knex cli should print help only when there are no arguments.
current implementation always prints help even when given correct arguments, such as `knex migrate:status`.
the correct way to check argument length is via process.argv.
reference: https://github.com/tj/commander.js/#outputhelpcb

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -341,7 +341,7 @@ function invoke(env) {
         .catch(exit);
     });
 
-  if (!commander._args.length) {
+  if (!process.argv.slice(2).length) {
     commander.outputHelp();
   }
 
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "JSONStream": "^1.3.5",
     "chai": "^4.2.0",
     "chai-subset-in-order": "^2.1.3",
-    "cli-testlab": "^1.8.0",
+    "cli-testlab": "^1.10.0",
     "coveralls": "^3.0.9",
     "cross-env": "^6.0.3",
     "dtslint": "2.0.2",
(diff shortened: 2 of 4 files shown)


timorthi:
+1 also running into this, looks like a regression from #3604

kibertoad:
I'll take a look tomorrow, thanks. One quick note: could you please add cli-testlab-based test for this?

yeonhoyoon:
OK, but I'm not familiar with `cli-testlab`, so I will need some time to look into it.

kibertoad:
@yeonhoyoon Here is a super simple example: https://github.com/knex/knex/blob/master/test/cli/version.spec.js

and you can check other files in same directory if you'd need more than that

yeonhoyoon:
@kibertoad 
Thanks for your feedback, I added some specs to test help command.
There wasn't an assertion helper in the `cli-testlab` for expecting the output to _not include_ a given string, so I used the raw stdout variable instead to test the desired behavior.

kibertoad:
Much appreciated!

There are linting errors in the test:
```
/home/travis/build/knex/knex/test/cli/help.spec.js
   6:7  error  'cliPkg' is assigned a value but never used  no-unused-vars
  30:7  error  'expect' is not defined                      no-undef
```

After those are fixed, this can be merged :)

kibertoad:
@yeonhoyoon BTW, thank you for mentioning missing functionality in `cli-testlab`. I've added `notExpectedOutput` parameter and published 1.9.0, but it's a small thing, feel free to update it or leave as it is.

yeonhoyoon:
@kibertoad 
Apologies for leaving unused variables in the code, and thanks for quickly adding features to `cli-testlab`. I've changed the tests to use the newly added functionality. 
Please let me know if you have any other feedback.

kibertoad:
Thanks! <3

kibertoad:
Released in 0.20.8

PR #3629, opened 2020-01-17
Update index.d.ts, interface MsSqlConnectionConfig enableArithAbort: null | boolean is required now.

saostad:
Config object for mssql: see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mssql/index.d.ts is outdated. in tedious https://github.com/tediousjs/tedious/blob/master/src/connection.ts line 154, enableArithAbort: null | boolean is required now.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1701,7 +1701,7 @@ declare namespace Knex {
       appName?: string;
       abortTransactionOnError?: boolean;
       trustedConnection?: boolean;
-      enableArithAbort?: boolean,
+      enableArithAbort?: boolean;
     };
     pool?: {
       min?: number;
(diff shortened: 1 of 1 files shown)


kibertoad:
Users might be using different versions of mssql, though, so I'd still keep it as optional.

saostad:
> 
> 
> Users might be using different versions of mssql, though, so I'd still keep it as optional.

we can add it as optional filed, now it doesn't exist.

kibertoad:
@saostad Could you please adjust PR accordingly?

types/index.d.ts
```diff
@@ -1701,6 +1701,7 @@ declare namespace Knex {
       appName?: string;
       abortTransactionOnError?: boolean;
       trustedConnection?: boolean;
+      enableArithAbort?: boolean,
```
kibertoad: Could you change `,` to be `;` for consistency?

saostad: I did.

kibertoad:
Thanks!

PR #3634, opened 2020-01-21
replace Bluebird.timeout

maximelkin:
Breaking changes: TImeoutError now not extends Bluebird.TimeoutError
also this error exported explicitly

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "liftoff": "3.1.0",
     "lodash": "^4.17.15",
     "mkdirp": "^0.5.1",
-    "p-timeout": "^3.2.0",
     "pg-connection-string": "2.1.0",
     "tarn": "^2.0.0",
     "tildify": "2.0.0",
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "liftoff": "3.1.0",
     "lodash": "^4.17.15",
     "mkdirp": "^0.5.1",
+    "p-timeout": "^3.2.0",
     "pg-connection-string": "2.1.0",
     "tarn": "^2.0.0",
     "tildify": "2.0.0",
--- a/lib/client.js
+++ b/lib/client.js
@@ -25,6 +25,7 @@ const { makeEscape } = require('./query/string');
 const { uniqueId, cloneDeep, defaults } = require('lodash');
 
 const Logger = require('./logger');
+const { KnexTimeoutError } = require('./util/timeout');
 
 const debug = require('debug')('knex:client');
 const _debugQuery = require('debug')('knex:query');
@@ -344,11 +345,15 @@ Object.assign(Client.prototype, {
           debug('acquired connection from pool: %s', connection.__knexUid);
           return connection;
         })
-        .catch(TimeoutError, () => {
-          throw new Bluebird.TimeoutError(
-            'Knex: Timeout acquiring a connection. The pool is probably full. ' +
-              'Are you missing a .transacting(trx) call?'
-          );
+        .catch((error) => {
+          let convertedError = error;
+          if (error instanceof TimeoutError) {
+            convertedError = new KnexTimeoutError(
+              'Knex: Timeout acquiring a connection. The pool is probably full. ' +
+                'Are you missing a .transacting(trx) call?'
+            );
+          }
+          throw convertedError;
         });
     } catch (e) {
       return Bluebird.reject(e);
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -39,6 +39,7 @@
     "liftoff": "3.1.0",
     "lodash": "^4.17.15",
     "mkdirp": "^0.5.1",
+    "p-timeout": "^3.2.0",
```
kibertoad: Considering how small p-yimeout is, maybe it would make sense to inline implementation and use native finally instead of p-finally?

kibertoad: no strong opinion on that, though, sindresorhus libs are legit.

maximelkin:
Does we still support node 8?

package.json
maximelkin: Done

kibertoad:
@maximelkin Yeah, ideally we want to support Node 8 for a bit longer. If that makes implementation complex, we can stick with p-timeout, it's not a big deal.

maximelkin:
This can make harder to move from bluebird completely, because Promise.finally only from 10 node

kibertoad:
@maximelkin doesn't p-timeout rely on p-finally which supports node 8?

maximelkin:
Yes, it is
But p-timeout dependency removed

kibertoad:
@maximelkin I'd much rather use p-finally directly or p-timeout rather than keep bluebird or break node 8  compatibility.

maximelkin:
p-timeout required finally because of support  .cancel, we dont need it, so no troubles here

kibertoad:
ah, great :)

kibertoad:
@maximelkin Since https://github.com/knex/knex/pull/3639 is dropping support for Node 8 anyway, I guess this PR can do the same and we can merge both during next breaking release.

maximelkin:
Not directly related question:
bluebird will be replaced completely to es6 promises, or user may choose api/internal promises?

elhigu:
> bluebird will be replaced completely to es6 promises, or user may choose api/internal promises?

Knex will use only native promises and will not allow to choose the implementation.

maximelkin:
@kibertoad This pr actually not breaking node 8 compatibility, maybe we can merge it?

kibertoad:
@maximelkin There's a new unhandled Rejection on Node 8 happening ever since this was merged:

```
{ reason: Error: Transaction rejected with non-error: undefined
    at q.trxClient.query.catch.then.res (/home/travis/build/knex/knex/lib/transaction.js:15:2936)
From previous event:
    at Transaction._promise.acquireConnection.connection (/home/travis/build/knex/knex/lib/transaction.js:8:1205)
    at Bluebird.then.then (/home/travis/build/knex/knex/lib/transaction.js:18:1002)
From previous event:
    at Transaction.acquireConnection (/home/travis/build/knex/knex/lib/transaction.js:18:886)
    at new Transaction (/home/travis/build/knex/knex/lib/transaction.js:8:834)
    at Client_SQLite3.transaction (/home/travis/build/knex/knex/lib/client.js:7:2337)
    at Function.transaction (/home/travis/build/knex/knex/lib/util/make-knex.js:6:103)
    at Context.it (/home/travis/build/knex/knex/test/unit/knex.js:403:28)
    at callFn (/home/travis/build/knex/knex/node_modules/mocha/lib/runnable.js:395:21)
    at Test.Runnable.run (/home/travis/build/knex/knex/node_modules/mocha/lib/runnable.js:382:7)
    at Runner.runTest (/home/travis/build/knex/knex/node_modules/mocha/lib/runner.js:541:10)
    at /home/travis/build/knex/knex/node_modules/mocha/lib/runner.js:667:12
    at next (/home/travis/build/knex/knex/node_modules/mocha/lib/runner.js:450:14)
    at /home/travis/build/knex/knex/node_modules/mocha/lib/runner.js:460:7
    at next (/home/travis/build/knex/knex/node_modules/mocha/lib/runner.js:362:14)
    at Immediate._onImmediate (/home/travis/build/knex/knex/node_modules/mocha/lib/runner.js:428:5)
    at runCallback (timers.js:810:20)
    at tryOnImmediate (timers.js:768:5)
    at processImmediate [as _immediateCallback] (timers.js:745:5) }
```

Wonder if there's a gap in our timeout implementation and if sindresorhus one would be more reliable.

maximelkin:
@kibertoad 
I think it somehow connected with refactoring from bluebird pattern-match by class .catch to es6 .catch

PR #3638, opened 2020-01-27
Ft/don't interfere with "type": "module"

ExProbitasFiducia:
Adding `"type": "module"` in a package that depends on knex means files using the `.js` extension will be interpreted as es modules. This change allows the knexfile to use the `.cjs` extension so knex can `require` the knexfile, because `require` cannot import es modules.

This PR also adds the `.cjs` extension as a default extension, so it doesn't have to be explicitly included with `--knexfile="knexfile.cjs"`


kibertoad:
Could you please add an integration test for this? Check https://github.com/knex/knex/tree/master/test/cli for examples.

kibertoad:
Thank! Will take a look tomorrow.

kibertoad:
@ExProbitasFiducia This still needs a test, also there's a conflict now. Btw, is this change still needed now that Liftoff support was fixed?

kibertoad:
@ExProbitasFiducia Ping? Any chance you could finish this? If you need any help with writing a CLI test, I would be happy to help.

kibertoad:
@ExProbitasFiducia Do you need any help with writing a test?

PR #3639, opened 2020-01-27
Ft/support native esm

ExProbitasFiducia:
This PR Enables knex to support native esm for knexfiles, migrations and seeds.

It enables the dynamic import of knexfiles, migrations, and seeds using the import() syntax.

It inspects the consuming app's package.json to determine if the application is configured to interpret .js files as esm or cjs, and uses the according stub.


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "knex.js",
   "types": "types/index.d.ts",
   "engines": {
-    "node": ">=8"
+    "node": ">=10"
```
kibertoad: travis config file needs to be updated not to run tests on Node 8 as well

lib/seed/stub/mjs.stub
```diff
@@ -0,0 +1,13 @@
+
+exports.seed = function(knex) {
+  // Deletes ALL existing entries
+  return knex('table_name').del()
```
kibertoad: why not use async/await?

lib/seed/Seeder.js
```diff
@@ -98,7 +112,7 @@ class Seeder {
     const seedPath = this._getNewStubFilePath(name);
     await writeJsFileUsingTemplate(
       seedPath,
-      this._getStubPath(),
+      await this._getStubPath(),
```
kibertoad: I would prefer not to have inline await, better extract to a separate const

lib/seed/Seeder.js
```diff
@@ -76,11 +77,24 @@ class Seeder {
...
+  async _getStubPath() {
+    const packageDirectory = await pkgDir(process.cwd());
+    const consumingPackageJson = require(path.resolve(
+      packageDirectory,
+      'package.json'
+    ));
+    const consumingPackageIsModule = consumingPackageJson.type === 'module';
+    const stubExtension = this.config.extension;
+    const stubFormat =
+      stubExtension === 'js'
```
kibertoad: can this be broken somehow? ternary inside a ternary is not very easy to read

lib/seed/Seeder.js
```diff
@@ -76,11 +77,24 @@ class Seeder {
     return name;
   }
 
-  _getStubPath() {
-    return (
-      this.config.stub ||
-      path.join(__dirname, 'stub', this.config.extension + '.stub')
-    );
+  async _getStubPath() {
+    const packageDirectory = await pkgDir(process.cwd());
```
kibertoad: does this need to happen on every method invocation? can't be moved outside of the function?

lib/migrate/sources/fs-migrations.js
```diff
@@ -80,7 +82,7 @@ class FsMigrations {
 
   getMigration(migration) {
     const absoluteDir = path.resolve(process.cwd(), migration.directory);
-    return require(path.join(absoluteDir, migration.file));
+    return import(path.join(absoluteDir, migration.file));
```
kibertoad: what's the advantage of this change?

lib/migrate/MigrationGenerator.js
```diff
@@ -35,11 +36,24 @@ class MigrationGenerator {
     return Promise.all(promises);
   }
 
-  _getStubPath() {
-    return (
-      this.config.stub ||
-      path.join(__dirname, 'stub', this.config.extension + '.stub')
-    );
+  async _getStubPath() {
```
kibertoad: is this completely identical to what Seeder is doing? can this be extracted to some util?

bin/utils/cli-config-utils.js
```diff
@@ -48,6 +48,22 @@ function resolveKnexFilePath() {
     };
   }
 
+  const mjsPath = resolveDefaultKnexfilePath('mjs');
+  if (fs.existsSync(mjsPath)) {
```
kibertoad: I would avoid sync method calls in runtime. Since you are dropping Node 8 support anyway, you can use fs.promises instead

kibertoad:
This needs integration tests (see https://github.com/knex/knex/tree/master/test/cli) and I'd prefer to release a couple more non-breaking releases before we drop Node 8, but eventually we are going to need this, so thank you for this submission!

lib/migrate/sources/fs-migrations.js
ExProbitasFiducia: require cannot import es modules, whereas import can.

lib/migrate/MigrationGenerator.js
ExProbitasFiducia: It is, I think the code was duplicated in the current master as well. This PR was already getting large, so It seemed like a good idea not to make *too* many changes; that way the relationship with the current version could be a bit clearer for any reviewers/maintainers

bin/utils/cli-config-utils.js
ExProbitasFiducia: Agreed/

lib/seed/Seeder.js
ExProbitasFiducia: if the extension is js, check to see if the consuming package has type: module in its package.json. if it does, use the module format, if it doesnt, use esm format. Otherwise, just use the stub file with the extension as is.

I think that covers all cases. The problem with breaking it out is that we assign it to a const.
Maybe it's sensible to put it in an iife? either way.

lib/seed/Seeder.js
ExProbitasFiducia: I agree. It confuses the order in which things happen a bit.

lib/seed/stub/mjs.stub
ExProbitasFiducia: This file is in cjs format, while it should be in esm... good catch.

package.json
ExProbitasFiducia: ok

lib/seed/Seeder.js
ExProbitasFiducia: maybe. Is it possible that a particular consuming package can install a copy of knex, and then a dependency of that package can use the same copy of knex?

bin/utils/cli-config-utils.js
ExProbitasFiducia: theres a bunch of unrelated code that also uses promisify on the node utils. maybe it makes sense to make this change, along with updating that other code, in a separate pr.?

bin/cli.js
```diff
@@ -3,6 +3,7 @@
 const Liftoff = require('liftoff');
 const interpret = require('interpret');
 const path = require('path');
+const pkgDir = require('pkg-dir');
```
ExProbitasFiducia: pkg dir is being used to find the package.json of the project in which the clie is executed.

bin/utils/cli-config-utils.js
```diff
@@ -7,6 +7,7 @@ const color = require('colorette');
 const argv = require('getopts')(process.argv.slice(2));
 
 function mkConfigObj(opts) {
+  console.log('\n\n\n>>>mkConfigObj<<<\n\n');
```
kibertoad: is this needed?

lib/migrate/MigrationGenerator.js
kibertoad: fair

bin/utils/cli-config-utils.js
kibertoad: sure

package.json
```diff
@@ -28,6 +28,7 @@
     "stress:destroy": "docker-compose -f scripts/stress-test/docker-compose.yml stop"
   },
   "dependencies": {
+    "babel-plugin-transform-dynamic-import": "^2.1.0",
```
kibertoad: @briandamaged How do you feel about introducing babel plugin for this?

briandamaged: Hmm... I'll need to read through this PR more thoroughly to make sure I understand _how_ this plugin is being used.

(In particular: if the plugin is being used as a compilation step for `knex` itself, then that's probably fine.   However, if `knex` is becoming a top-level driver for `babel`, then it seems like that could become problematic.)

bin/cli.js
```diff
@@ -124,30 +124,41 @@ function invoke(env) {
...
-            type +
-            '.stub'
-        )
-        .then((code) => {
-          return fsPromised.writeFile(stubPath, code);
-        })
-        .then(() => {
-          success(color.green(`Created ${stubPath}`));
-        })
-        .catch(exit);
```
briandamaged: This change seems unrelated to native ESM support.  (It's just updating the code to use `async` / `await` syntax, correct?)

bin/cli.js
```diff
@@ -161,139 +172,147 @@ function invoke(env) {
...
+        const stub = getStubPath('migrations', env, opts);
+        if (stub) {
+          configOverrides.stub = stub;
+        }
+
+        const createdName = await knex.migrate.make(name, configOverrides);
+        success(color.green(`Created Migration: ${createdName}`));
+      } catch (error) {
+        console.log(error);
+        exit();
```
briandamaged: This change seems unrelated to native ESM support.  (It's just updating the code to use `async` / `await` syntax, correct?)

bin/cli.js
```diff
@@ -124,30 +124,41 @@ function invoke(env) {
...
-      if (env.configuration) {
-        exit(`Error: ${env.knexfile} already exists`);
+    .action(async () => {
+      try {
+        const packageDirectory = await pkgDir(process.cwd());
+        const consumingPackageJson = require(path.resolve(
+          packageDirectory,
+          'package.json'
+        ));
+        const consumingPackageIsModule = consumingPackageJson.type === 'module';
```
briandamaged: This logic appears to be duplicated in at least 3 places.  Perhaps consolidate it into a reusable function?

(On a related note: I suspect that `Liftoff` probably makes this information available already.  If so, then it would probably be best to use the values that `Liftoff` provides.  Otherwise, we'll probably encounter some weird discrepancies when `knex` / `Liftoff` disagree about which `package.json` should be used)

bin/cli.js
```diff
@@ -29,10 +30,11 @@ const fsPromised = {
...
 };
 
-function initKnex(env, opts) {
+const initKnex = async (env, opts) => {
   if (opts.esm) {
-    // enable esm interop via 'esm' module
-    require = require('esm')(module);
+    console.warn(
+      `The 'esm' opt is deprecated. Knex supports esm module imports using 'import()'`
+    );
```
briandamaged: What happens if an end-user was already using the `--esm` flag?  Are we sure that the new functionality will be 100% compatible with their existing Knexfiles?  (In other words: will anything break for these users?)

briandamaged:
@ExProbitasFiducia : Howdy, and thx for putting this PR together!  I'm just catching up w/ things, so it's probably going to take me a couple of days to read through everything.

However, after a quick glance, I'm not entirely sure what the scope of this PR is.  From what I can tell, it looks like it's doing several different things:

1. Introducing native support for ESM
2. Updating the syntax in several places to use `async` / `await`
3. Deprecating the `--esm` flag.  (it's unclear to me if the new code will be 100% compatible to the previous `--esm` flag behavior)
4. Revising the Migrator because...?  (more `async` / `await` revisions?)
5. Etc.

Would it be possible to split this PR into multiple smaller PRs that each have a narrowly-defined scope?  For example: for the "native support for ESM" task, it seems like the PR should only include the following changes:

1. Adding the ability to infer if the consuming package is a "module".  (Note: if I recall, there is already a separate library that can handle this step.  It might be the `esm` module that has been mentioned elsewhere)
2. If the consuming package is a "module", then add support for the "mjs" and "cjs" extensions.

Thanks in advance!

PR #3640, opened 2020-01-28
Refactor more tests to use cli-testlab

kibertoad:


lib/util/parse-connection.js
```diff
@@ -1,11 +1,13 @@
 const url = require('url');
 const { parse } = require('pg-connection-string');
 const parsePG = parse;
+const isWindows = process && process.platform && process.platform === 'win32';
 
 module.exports = function parseConnectionString(str) {
   const parsed = url.parse(str, true);
   let { protocol } = parsed;
-  if (protocol === null) {
+  const isDriveLetter = isWindows && protocol.length === 2;
```
kibertoad: Original implementation didn't support sqlite paths to begin with drive letter, e. g. "c:/knex/migrations"

PR #3647, opened 2020-02-01
Update QueryCompiler implementation to use classes

lorefnon:


elhigu:
This looks awesome 👍

lib/dialects/sqlite3/query/compiler.js
```diff
@@ -12,25 +12,30 @@ const {
...
+      );
+    }
+    const emptyStr = constant('');
+    this.forShare = emptyStr;
+    this.forUpdate = emptyStr;
   }
-}
 
-inherits(QueryCompiler_SQLite3, QueryCompiler);
+  // inherits(QueryCompiler_SQLite3, QueryCompiler);
```
kibertoad:
```suggestion
  // inherits(QueryCompiler_SQLite3, QueryCompiler);
```

kibertoad: Why the commented out code?

lorefnon: Oh, I thought I got rid of all of them. 

Just an old habit that makes it easier for me to review changes before committing.

lib/dialects/sqlite3/query/compiler.js
```diff
@@ -12,25 +12,24 @@ const {
...
 
-assign(QueryCompiler_SQLite3.prototype, {
-  // The locks are not applicable in SQLite3
-  forShare: emptyStr,
+    if (returning) {
+      this.client.logger.warn(
+        '.returning() is not supported by sqlite3 and will not have any effect.'
+      );
+    }
+    const emptyStr = constant('');
```
kibertoad: Do we need to redefine this on every call? Maybe this could be bumped to the top of the file to only be invoked once?

lorefnon: Ok, updated.

kibertoad:
Thanks!

PR #3649, opened 2020-02-02
WIP: Improving support for introducing adhoc (unsafe) transformations before SQL generation

lorefnon:
This primarily follows from the thread here: https://github.com/knex/knex/pull/3593#issuecomment-580346520

Proposed API adds a `transformRawTokens` to the `QueryBuilder` which enables users to transform the token list through a fluent visitor before the final SQL is generated. 

This is of course unsafe in the sense it is easy to construct invalid sql but can be an easy escape hatch when advanced users (who know what they are doing) want to insert dialect specific fragments 
 (for features that are not supported by knex) at appropriate places.

Example:

```js
knex('books')
  .insert([...])
  .transformRawTokens(t => {
    // t is a token visitor that allows you to easily traverse through the token list and 
    // insert custom tokens at adhoc places or substitute a range of tokens with another.
    t.findToken('insert').insertAfter('top 10')
  })
```

Generates: 

```sql
insert top 10 into [books] ...
```
Similarly,
```js
knex('dbo.DimCustomer as a')
  .innerJoin('dbo.FactInternetSales as b', 'a.CustomerKey', 'b.CustomerKey')
  .select('count(*)')
  .transformRawTokens(t => {
    t.last().insertAfter('OPTION (HASH JOIN)');
  })

```

Generates: 

```sql
SELECT COUNT (*) FROM [dbo].[DimCustomer] AS [a]  
INNER JOIN [dbo].[FactInternetSales] AS [b]   
ON ([a].[CustomerKey] = [b].[CustomerKey])  
OPTION (HASH JOIN); 
```

---

# TODO:

- [ ] Add support for placeholder bindings
- [ ] Update type definitions
- [ ] Add tests
- [ ] Update docs

### DIFF
--- a/lib/query/compiler.js
+++ b/lib/query/compiler.js
@@ -6,6 +6,8 @@ const QueryBuilder = require('./builder');
 const JoinClause = require('./joinclause');
 const debug = require('debug');
 
+const { TokenListVisitor } = require('./token-visitors');
+
 const {
   assign,
   bind,
@@ -18,6 +20,7 @@ const {
   omitBy,
   reduce,
   has,
+  isFunction
 } = require('lodash');
 const uuid = require('uuid');
 
@@ -52,6 +55,7 @@ class QueryCompiler {
     this.formatter = client.formatter(builder);
     // Used when the insert call is empty.
     this._emptyInsertValue = 'default values';
+    this._tokenTransformers = builder._tokenTransformers;
     this.first = this.select;
   }
 
@@ -88,6 +92,14 @@ class QueryCompiler {
       query.sql = val;
     } else {
       assign(query, val);
+    }
+
+    if (isString(query.sql) && Array.isArray(this._tokenTransformers)) {
+      const visitor = new TokenListVisitor(tokenList);
+      for (let i = 0; i < this._tokenTransformers.length; i++) {
+        this._tokenTransformers[i](visitor);
+      }
+      query.sql = visitor.toString();
     }
 
     if (method === 'select' || method === 'first') {
--- a/lib/query/compiler.js
+++ b/lib/query/compiler.js
@@ -6,7 +6,7 @@ const QueryBuilder = require('./builder');
 const JoinClause = require('./joinclause');
 const debug = require('debug');
 
-const { TokenListVisitor } = require('./token-visitors');
+const { SQLTokenListVisitor } = require('./token-visitors');
 
 const {
   assign,
@@ -19,8 +19,7 @@ const {
   map,
   omitBy,
   reduce,
-  has,
-  isFunction
+  has
 } = require('lodash');
 const uuid = require('uuid');
 
@@ -94,8 +93,8 @@ class QueryCompiler {
       assign(query, val);
     }
 
-    if (isString(query.sql) && Array.isArray(this._tokenTransformers)) {
-      const visitor = new TokenListVisitor(tokenList);
+    if (Array.isArray(this._tokenTransformers) && isString(query.sql)) {
+      const visitor = new SQLTokenListVisitor(query.sql);
       for (let i = 0; i < this._tokenTransformers.length; i++) {
         this._tokenTransformers[i](visitor);
       }
--- a/lib/query/token-visitors.js
+++ b/lib/query/token-visitors.js
@@ -1,0 +1,168 @@
+const { isEmpty, compact, isNil, isFunction, isString, cloneDeep, castArray } = require("lodash");
+const Tokenizer = require("tokenize-this");
+
+class TokenListVisitor {
+    constructor(tokens) {
+        if (isString(tokens)) tokens = this.tokenize(tokens);
+        this.tokens = tokens;
+    }
+    tokenize(rawStr) {
+        const tokens = [];
+        const tokenizer = new Tokenizer();
+        tokenizer.tokenize(rawStr, (token) => {
+            tokens.push(token);
+        });
+        return tokens;
+    }
+    toString() {
+        return compact(this.tokens).join(' ');
+    }
+    findToken(value, isCaseSensitive = false, startIdx = 0) {
+        const lowerValue = value.toLowerCase();
+        for (let idx = startIdx; idx < this.tokens.length; idx++) {
+            const didMatch = isCaseSensitive
+                ? this.tokens[idx] === value
+                : this.tokens[idx].toLowerCase() === lowerValue;
+            if (didMatch) return new TokenItemVisitor(idx, this);
+        }
+        return null;
+    }
+    first() {
+        return new TokenItemVisitor(0, this);
+    }
+    last() {
+        return new TokenItemVisitor(this.tokens.length-1, this);
+    }
+    _replaceRange(startIdx, endIdx, tokenList) {
+        const normalized = this._normalizeTokenList(tokenList);
+        this.tokens.splice(startIdx, endIdx - startIdx, ...normalized);
+        return new TokenRangeVisitor(startIdx, normalized.length, this);
+    }
+    _replaceAtIndex(i, tokenList) {
+        const normalized = this._normalizeTokenList(tokenList);
+        this.tokens.splice(i, 1, ...normalized);
+        return new TokenRangeVisitor(i, normalized.length, this);
+    }
+    _insertAtIndex(i, tokenList) {
+        const normalized = this._normalizeTokenList(tokenList);
+        if (normalized.length === 0) return null;
+        this.tokens.splice(i, 0, ...normalized);
+        return new TokenRangeVisitor(i, normalized.length, this);
+    }
+    _normalizeTokenList(tokenList) {
+        return compact(castArray(tokenList));
+    }
+    _selectRange(start, step, predicate) {
+        const { tokens } = this.tokenListVisitor;
+        const range = { start: null, end: null };
+        for (let i = start; i < tokens.length && i >= 0; i += step) {
+            const token = tokens[i];
+            const shouldSelect = predicate(token, i);
+            if (shouldSelect) {
+                if (i <= start) {
+                    range.start = i;
+                }
+                if (i >= start) {
+                    range.end = i;
+                }
+            } else break;
+        }
+        if (isNil(range.start) || isNil(range.next)) {
+            return null;
+        }
+        return new TokenRangeVisitor(range.start, range.end, this);
+    }
+}
+
+class TokenItemVisitor {
+    constructor(index, tokenListVisitor) {
+        this.index = index;
+        this.tokenListVisitor = tokenListVisitor;
+    }
+    token() {
+        return this.tokenListVisitor.tokens[this.index];
+    }
+    peek(count = 1) {
+        return this.tokenListVisitor.tokens.slice(this.index, this.index + count).join(' ');
+    }
+    next(token, isCaseSensitive = false) {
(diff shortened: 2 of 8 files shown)


test/unit/query/token-visitors.js
```diff
@@ -0,0 +1,83 @@
+const Knex = require('../../../lib/index');
+const { expect } = require('chai');
+const { TokenListVisitor } = require('../../../lib/query/token-visitors');
+
+describe.only("Token visitor", () => {
```
kibertoad: Probably ".only" wasn't intended to be committed :)

lib/query/token-visitors.js
```diff
@@ -0,0 +1,188 @@
...
+    extendRight(until) {
+        const rangeVisitor = this.tokenListVisitor._selectRange(this.endIndex, 1, until);
+        if (!isNil(rangeVisitor)) {
+            rangeVisitor.startIndex = this.startIndex;
+        }
+        return rangeVisitor
+    }
+    replace(replacement) {
+        if (isFunction(replacement)) {
+            let tokens = cloneDeep(this.tokenListVisitor.tokens.slice(this.startIndex, this.endIndex))
```
kibertoad: why use `let`? You intend to modify this variable at some point?

lib/query/compiler.js
```diff
@@ -90,6 +93,14 @@ class QueryCompiler {
       assign(query, val);
     }
 
+    if (isString(query.sql) && Array.isArray(this._tokenTransformers)) {
```
kibertoad: Probably case when query doesn't have `sql` is less frequent than when tokenTransformers are not set? If that's the case, probably it would make sense to swap conditions.

lib/query/token-visitors.js
```diff
@@ -0,0 +1,188 @@
...
+                    tokens[tokens.length - 1] + token,
+                    tokens[tokens.length - 1] + ' ' + token
+                ]) {
+                    if (operators[combined]) {
+                        tokens[tokens.length - 1] = combined;
+                        return;
+                    }
+                }
+            }
+            tokens.push(`${token}`);
```
kibertoad: why wrap in template?

lib/query/token-visitors.js
```diff
@@ -0,0 +1,188 @@
+const { isEmpty, compact, isNil, isFunction, isString, cloneDeep, castArray } = require("lodash");
+const Tokenizer = require("tokenize-this");
+const { operators } = require("../operators");
+
+class TokenListVisitor {
```
kibertoad: Name sounds somewhat confusing, it sounds like a very generic visitor, but it actually has a specific purpose; can it be reflected in the name?

lib/query/token-visitors.js
```diff
@@ -0,0 +1,188 @@
...
+            }
+            tokens.push(`${token}`);
+        });
+        return tokens;
+    }
+    toString() {
+        return compact(this.tokens).join(' ');
+    }
+    findToken(value, isCaseSensitive = false, startIdx = 0) {
+        const lowerValue = value.toLowerCase();
```
kibertoad: what if value is nil?

lib/query/token-visitors.js
```diff
@@ -0,0 +1,188 @@
...
+        for (let idx = startIdx; idx < this.tokens.length; idx++) {
```
kibertoad: why not juse use .find() method of JS arrays?

lib/query/token-visitors.js
```diff
@@ -0,0 +1,188 @@
...
+    _insertAtIndex(i, tokenList) {
+        const normalized = this._normalizeTokenList(tokenList);
+        if (normalized.length === 0) return null;
+        this.tokens.splice(i, 0, ...normalized);
+        return new TokenRangeVisitor(i, normalized.length, this);
+    }
+    _normalizeTokenList(tokenList) {
+        return compact(castArray(tokenList));
+    }
+    _selectRange(start, step, predicate) {
```
kibertoad: can you please add comment explaining this logic?

elhigu:
So is there going to be option to completely transform token content to be something else? It wasn't clear from initial API description.

lorefnon:
>  is there going to be option to completely transform token content to be something else?

Yes, the API above does support token replacements. Although I did miss to mention about it. 

Adding some more examples along with my mental model of the proposal: 

```js
knex('books')
  .insert([...])
  .transformRawTokens(t => {
    t.findToken('insert')     // What findToken returns is a TokenItemVisitor which is akin to a 
                              // cursor in the sequence of tokens.
                              // So at this point, my cursor is at the "insert" token
      .insertAfter('top 10')  // Insert something after the cursor (input is automatically tokenized 
                              // so that latter passes can also do things like findToken('top'))
  })
```

```js
knex('books')
  .select('*')
  .transformRawTokens(t => {
    t.findToken('select')       // Cursor is at select token
      .selecNextt()             // Select next token (this is unrelated to the this query being a select query)
                                // All the select* methods return a TokenRangeVisitor 
                                // which is essentially a selection of contiguous range of tokens
      .replaceWith('id')        // Replace the selected range of tokens with something else
                                // For the same logic as above, the input here is also automatically tokenized
                                // internally.
  })
```

All the select* methods accept a predicate function, so I can do things like: 

```js
t.findToken('select').selectNext((token) => token !== 'from')
```

which will select all tokens between `select` and `from`.

We can select (or expand a selection if we already have one) in either directions:

```js
t.findToken('from')
    .selectPrev((token) => token != 'select') // Keep selecting back until we hit a select
    .replaceWith('*')                         // replace the whole range with a '*'
```

elhigu:
For the replacement API I thought there would be way to get old value of the compilation that could be used to transform/replace it. Actually I originally thought that it would be the only API to transform those tokens since it effectively allows to do all the prepend, append, remove, replace and modify operations with single and quite simple API.

Like:
```
t.findToken('from')
    .selectPrev((token) => token != 'select') // Keep selecting back until we hit a select
    .transform(old => {
       return { sql: old.replace('foobar', 'FOOBAR'), bindings: old.bindings };
    })
```

lorefnon:
@elhigu  Regarding https://github.com/knex/knex/pull/3649#issuecomment-581857543, I think we can add a `transformRawSQL` in addition of `transformRawTokens`, which allows users to modify the complete generated SQL. 

I am a bit more inclined towards retaining the token oriented API because I feel raw string manipulation (using regex replacements etc.) can be more error prone and break in unexpected ways. 

Something like `findToken('insert')` will not match a token inside a string for example. Also I find the cursor oriented API slightly easier to reason through when nested queries are involved. So for instance if I seek the `insert` token and then seek the next `from`, the chance of accidentally transforming some unrelated `from` in a nested query is minimized.

Let me know your thoughts. I'll make any changes needed.

elhigu:
> I am a bit more inclined towards retaining the token oriented API because I feel raw string manipulation (using regex replacements etc.) can be more error prone and break in unexpected ways.

Right... I had only read the API description earlier and now that I just read the code I understood that this thing is actually parsing the SQL strings to tokens and doing replacements on those. I had first thought that those tokens were just referring to query builder internals structures where different where clauses etc. were stored.

How do you thing that SQL string should be tokenized? for example when there are operators or value literals embedded in the string? To get reasonable and robust tokens out of SQL string sounds quite hard problem to solve or am I missing something here? There is a reason why knex hasn't gone to the route where it would start parsing the SQL that is given to it so far. 

Anyways since we are matching to actual tokens inside strings, we can do modification to the token easily, since we already know when finding the token what is inside of it.

So I don't see any reason anymore why that plain string manipulation API I suggested earlier should be added (except that it sounds a lot simpler and from knex's perspective it should work easily for all the dialects and it's as expressive as the token replacement based API).

test/unit/query/token-visitors.js
lorefnon: Yes, removed.

lib/query/compiler.js
lorefnon: Yes, I agree. Updated.

lib/query/token-visitors.js
lorefnon: Added a string check now.

lib/query/token-visitors.js
lorefnon: Yeah, updated to SQLTokenListVisitor.

lib/query/token-visitors.js
lorefnon: It was done to stringify. Its no longer necessary after some other changes. removed.

lib/query/compiler.js
lorefnon: Yeah I agree. Update the order of checks.

lib/query/token-visitors.js
lorefnon: Updated to const.

test/unit/query/token-visitors.js
lorefnon: Yeah :) Removed.

lib/query/token-visitors.js
lorefnon: Because the starting index may not be 0.

lib/query/token-visitors.js
lorefnon: Yes, added now. Please lemme know if it is still unclear.

lorefnon:
> How do you thing that SQL string should be tokenized? for example when there are operators or value literals embedded in the string? To get reasonable and robust tokens out of SQL string sounds quite hard problem to solve or am I missing something here? There is a reason why knex hasn't gone to the route where it would start parsing the SQL that is given to it so far.

I use a generic tokenizer library that takes care of quotes being handled correctly. And I have additional handling over that to ensure that operators are not split out to multiple tokens. 

Other than that I don't think tokenization itself is a hard problem, esp. if all we do with them is concat them back to a string. If we were constructing an actual syntax tree by parsing the token list that would be hard (and dialect specific). 

But again I don't have full context on what knex has historically done so if we have encountered some issues with similar approaches before I'd like to find out more.

elhigu:
> I use a generic tokenizer library that takes care of quotes being handled correctly. 

Are you sure it handles all SQL quotes / escape stuff? For example mssql [] quotes etc.? There are multiple escaping schemes in many dialects and knex also supports to even override quoting schemes in knex to use something that knex is not using by default for certain dialect. 

> Other than that I don't think tokenization itself is a hard problem, esp. if all we do with them is concat them back to a string. If we were constructing an actual syntax tree by parsing the token list that would be hard (and dialect specific).

Tokenizing so that 95% works correctly isn't hard, but I really wouldn't like knex to have ton of bug reports in future, about tokenizer not working as expected in all the corner cases (probably those will arrive slowly though, since this is pretty advanced feature, that not many people need). That is why I would prefer having less smart way that works  100% of time correctly. 

Also when using token approach it will be necessary to document with very high detail how tokenizing is supposed to work.

lorefnon:
> Tokenizing so that 95% works correctly isn't hard, but I really wouldn't like knex to have ton of bug reports in future, about tokenizer not working as expected in all the corner cases (probably those will arrive slowly though, since this is pretty advanced feature, that not many people need). That is why I would prefer having less smart way that works 100% of time correctly.

@elhigu I see your point. I will close this pr since all the changes here are build around the assumption that we will operate on the list of tokens. I'll create a separate pr in near future that allows users to just intercept the raw string and modify that.

elhigu:
Sorry about the wasted effort you did with this PR  :(

PR #3650, opened 2020-02-05
remove bluebird

maximelkin:
Closing #1588

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -25,6 +25,7 @@ const { makeEscape } = require('./query/string');
 const { uniqueId, cloneDeep, defaults } = require('lodash');
 
 const Logger = require('./logger');
+const { availablePromiseMethods } = require('./util/promises');
 const { KnexTimeoutError } = require('./util/timeout');
 
 const debug = require('debug')('knex:client');
@@ -40,6 +41,14 @@ const { POOL_CONFIG_OPTIONS } = require('./constants');
 function Client(config = {}) {
   this.config = config;
   this.logger = new Logger(config);
+  const library = config.promiseLibrary || Promise;
+  this.promises = {
+    library,
+    availableMethods: availablePromiseMethods(
+      library,
+      config.promiseLibraryMethods
+    ),
+  };
 
   //Client is a required field, so throw error if it's not supplied.
   //If 'this.dialect' is set, then this is a 'super()' call, in which case
@@ -335,29 +344,26 @@ Object.assign(Client.prototype, {
   },
 
   // Acquire a connection from the pool.
-  acquireConnection() {
+  async acquireConnection() {
     if (!this.pool) {
-      return Bluebird.reject(new Error('Unable to acquire a connection'));
+      return Promise.reject(new Error('Unable to acquire a connection'));
     }
-    try {
-      return Bluebird.try(() => this.pool.acquire().promise)
-        .then((connection) => {
-          debug('acquired connection from pool: %s', connection.__knexUid);
-          return connection;
-        })
-        .catch((error) => {
-          let convertedError = error;
-          if (error instanceof TimeoutError) {
-            convertedError = new KnexTimeoutError(
-              'Knex: Timeout acquiring a connection. The pool is probably full. ' +
-                'Are you missing a .transacting(trx) call?'
-            );
-          }
-          throw convertedError;
-        });
-    } catch (e) {
-      return Bluebird.reject(e);
-    }
+    return this.pool
+      .acquire()
+      .promise.then((connection) => {
+        debug('acquired connection from pool: %s', connection.__knexUid);
+        return connection;
+      })
+      .catch((error) => {
+        let convertedError = error;
+        if (error instanceof TimeoutError) {
+          convertedError = new KnexTimeoutError(
+            'Knex: Timeout acquiring a connection. The pool is probably full. ' +
+              'Are you missing a .transacting(trx) call?'
+          );
+        }
+        throw convertedError;
+      });
   },
 
   // Releases a connection back to the connection pool,
@@ -370,14 +376,14 @@ Object.assign(Client.prototype, {
       debug('pool refused connection: %s', connection.__knexUid);
     }
 
-    return Bluebird.resolve();
+    return Promise.resolve();
   },
 
   // Destroy the current connection pool for the client.
   destroy(callback) {
     const maybeDestroy = this.pool && this.pool.destroy();
 
-    return Bluebird.resolve(maybeDestroy)
+    return Promise.resolve(maybeDestroy)
       .then(() => {
         this.pool = void 0;
 
@@ -390,7 +396,7 @@ Object.assign(Client.prototype, {
           callback(err);
         }
 
-        return Bluebird.reject(err);
+        return Promise.reject(err);
       });
   },
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -1,5 +1,3 @@
-const Bluebird = require('bluebird');
-
 const Raw = require('./raw');
 const Ref = require('./ref');
 const Runner = require('./runner');
@@ -25,7 +23,6 @@ const { makeEscape } = require('./query/string');
 const { uniqueId, cloneDeep, defaults } = require('lodash');
 
 const Logger = require('./logger');
-const { availablePromiseMethods } = require('./util/promises');
 const { KnexTimeoutError } = require('./util/timeout');
 
 const debug = require('debug')('knex:client');
@@ -41,14 +38,6 @@ const { POOL_CONFIG_OPTIONS } = require('./constants');
 function Client(config = {}) {
   this.config = config;
   this.logger = new Logger(config);
-  const library = config.promiseLibrary || Promise;
-  this.promises = {
-    library,
-    availableMethods: availablePromiseMethods(
-      library,
-      config.promiseLibraryMethods
-    ),
-  };
 
   //Client is a required field, so throw error if it's not supplied.
   //If 'this.dialect' is set, then this is a 'super()' call, in which case
--- a/lib/client.js
+++ b/lib/client.js
@@ -335,24 +335,22 @@ Object.assign(Client.prototype, {
   // Acquire a connection from the pool.
   async acquireConnection() {
     if (!this.pool) {
-      return Promise.reject(new Error('Unable to acquire a connection'));
+      throw new Error('Unable to acquire a connection');
     }
-    return this.pool
-      .acquire()
-      .promise.then((connection) => {
-        debug('acquired connection from pool: %s', connection.__knexUid);
-        return connection;
-      })
-      .catch((error) => {
-        let convertedError = error;
-        if (error instanceof TimeoutError) {
-          convertedError = new KnexTimeoutError(
-            'Knex: Timeout acquiring a connection. The pool is probably full. ' +
-              'Are you missing a .transacting(trx) call?'
-          );
-        }
-        throw convertedError;
-      });
+    try {
+      const connection = await this.pool.acquire().promise;
+      debug('acquired connection from pool: %s', connection.__knexUid);
+      return connection;
+    } catch (error) {
+      let convertedError = error;
+      if (error instanceof TimeoutError) {
+        convertedError = new KnexTimeoutError(
+          'Knex: Timeout acquiring a connection. The pool is probably full. ' +
+          'Are you missing a .transacting(trx) call?'
+        );
+      }
+      throw convertedError;
+    }
   },
 
   // Releases a connection back to the connection pool,
--- a/lib/dialects/oracle/index.js
+++ b/lib/dialects/oracle/index.js
@@ -5,7 +5,6 @@ const { promisify } = require('util');
 
 const inherits = require('inherits');
 const Client = require('../../client');
-const Bluebird = require('bluebird');
 const { bufferToString } = require('../../query/string');
 const Formatter = require('./formatter');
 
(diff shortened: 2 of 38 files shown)


maximelkin:
Wait for https://github.com/knex/knex/pull/3634

maximelkin:
todo one of:
1) moving tests from bluebird
2) global or knex instance promise lib setup
2.1) adding methods in runtime vs adding another level of prototype at knex instantiation
2.2) searching for all external async api methods and wrapping with user defined lib

lib/client.js
```diff
@@ -335,29 +333,26 @@ Object.assign(Client.prototype, {
...
-            convertedError = new KnexTimeoutError(
-              'Knex: Timeout acquiring a connection. The pool is probably full. ' +
-                'Are you missing a .transacting(trx) call?'
-            );
-          }
-          throw convertedError;
-        });
-    } catch (e) {
-      return Bluebird.reject(e);
+      return Promise.reject(new Error('Unable to acquire a connection'));
```
qubyte: No need for an explicit `Promise.reject` here. Since this is an `async` function, all you have to do is throw the error and the caller will see the promise this function returns become rejected.

maximelkin: Done

briandamaged:
Howdy!  Just curious: what's the intended scope of this PR?

1. Remove the dependency on the Bluebird library?
2. Rewrite existing logic to take advantage of `async` / `await`?
3. Something else?

The reason I'm asking: this seems like it could potentially touch a _lot_ of code, which will make it challenging to code review.  So, perhaps the overall effort can be split into several smaller PRs?

kibertoad:
I agree with the idea of smaller PRs being better and this is something we should in general strive for; however, at this point it's likely that PR split is going to take non-trivial effort; if that is the case, I'd rather review a larger PR than lose part of refactoring due to simple revert.

kibertoad:
That said, since PR is pretty large already, it might make sense to wrap up what was already changed into final state, review, merge and continue in a separate PR, rather than keep expanding this one.

maximelkin:
In some places conversion from promise chain to async/await done because of node 8 support and .finally usage, if we can remove node 8 support, some changes can be reverted to make pr smaller

lib/interface.js
```diff
@@ -81,31 +81,21 @@ module.exports = function(Target) {
     );
   };
 
+  Target.prototype.asCallback = function(cb) {
+    const promise = this.then();
+    promise.then(
```
maximelkin: todo: move to utils

briandamaged:
In terms of the "smaller PR" discussion: we don't necessarily need to revert any changes.  Instead, we could use this branch as a basis for generating smaller incremental patches:

https://stackoverflow.com/questions/28192623/create-patch-or-diff-file-from-git-repository-and-apply-it-to-another-different

For example:

1. Generate a patch that includes only "trivial changes" within the `/lib` folder.  (Ex: files where you only had to rename `Bluebird.resolve` to `Promise.resolve`, etc.   Also: notice that the `/test` folder is remaining untouched.  This will give us confidence that the new code is still working correctly)
2. Generate a patch that includes some of the "nontrivial changes" within the `/lib` folder.  (Ex: places where code had to be rewritten because it relied upon a Bluebird feature.  Once again: the `/test` folder remains untouched).  You might need to repeat this step depending upon the complexity of the changes.
3. Generate a patch for the `/lib` folder.

There might be more steps as well, of course.  The main concept: you don't need to manually recreate the changes.  Instead, you can use this branch as the basis for generating multiple patch sets.  This will then make it significantly easier to review and merge the code.

I can assist you with this if you'd like to give it a try.

briandamaged:
Side note: I also see why this is such a challenging PR.  I just replaced a Bluebird Promise in 1 line of code, and suddenly 15 seemingly-unrelated sections of the code exploded.  🙀 

It turns out that those sections of the code all relied upon the `.finally(..)` method at some point within their control flow.  So, I temporarily re-wrapped the Promise w/ `Bluebird.resolve(..)` just to work around the issue.

test/utils.js
```diff
@@ -0,0 +1,8 @@
+const assert = require('assert');
```
kibertoad: could be chai for consistency

test/tape/invalid-db-setup.js
```diff
@@ -54,12 +55,13 @@ module.exports = (knexfile) => {
...
-              t.fail(`Stream query should have failed with non timeout error`);
-            })
             .catch((e) => {
+              t.notOk(
+                e instanceof KnexTimeoutError,
+                'Stream query should have failed with non timeout error'
+              );
               t.ok(
-                e.message.indexOf('i-refuse-to-exist') > 0,
+                e.message.includes('i-refuse-to-exist'),
```
kibertoad: nice :)

kibertoad:
Preferably we keep Node 8 support until Node 14 lands.

test/utils.js
```diff
@@ -0,0 +1,8 @@
...
+
+const expectError = (promise) =>
+  promise.then(() => assert.fail("Shouldn't have gotten here."), (e) => e);
```
briandamaged: You might also consider using the `chai-as-promised` plugin:

https://www.chaijs.com/plugins/chai-as-promised/

This encapsulates the pattern of "failing when an exception didn't occur".

lib/interface.js
```diff
@@ -81,31 +82,18 @@ module.exports = function(Target) {
...
-      'delay',
-    ],
-    function(method) {
-      Target.prototype[method] = function() {
-        const promise = this.then();
-        return promise[method].apply(promise, arguments);
-      };
-    }
-  );
+  availablePromiseMethods().forEach(function(method) {
```
briandamaged: The list produced by `availablePromiseMethods()` is significantly shorter than the list that it's replacing.  Was this intentional?  (Granted: I'm assuming that many of the methods in the original list were Bluebird-specific)

lib/transaction.js
```diff
@@ -251,6 +253,15 @@ class Transaction extends EventEmitter {
         }
       });
   }
+
+  then(onResolve, onReject) {
+    return this._promise.then(onResolve, onReject);
+  }
```
briandamaged: FYI: this is probably related to the "`availablePromiseMethods()` list is significantly shorter than the list that it's replacing" concern that I mentioned elsewhere.  The `availablePromiseMethods()` method is currently omitting `"then"`.

In other words: if the list produced by `availablePromiseMethods()` included `"then"`, then you probably could just rely upon meta-programming to generate this method automatically.

(On the flipside: given that the list of Promise methods only contains 2-3 elements now, it might no longer be worth leveraging meta-programming for this.)

lib/util/promises.js
```diff
@@ -0,0 +1,5 @@
+function availablePromiseMethods() {
+  return Promise.prototype.finally ? ['catch', 'finally'] : ['catch'];
```
briandamaged: Hmm... based upon the Knex documentation, it looks like the `.finally(..)` method is intentionally being made available to older versions of Node.  (Specifically: release notes for version `0.20.2`).  So, perhaps we should eliminate the conditional logic here and just _always_ provide the `.finally(..)` method?  (Of course, the meta-programming might interfere with this.  So... perhaps we should avoid meta-programming for generating these methods?)

maximelkin: 0.20.2 is about typings only, as I can see

lib/interface.js
maximelkin: Yes, this methods are from bluebird

test/utils.js
maximelkin: code removed

test/utils.js
maximelkin: done

lib/util/promises.js
briandamaged: These release notes for `0.20.2` on  http://knexjs.org/  mention this:

> Fix regression in older version of node when Promise#finally was not available #3507

(Perhaps this line was omitted from another copy of the release notes?)

maximelkin:
I can't change lib/ code without changing test/, because tests depends on .finally, which is incompatible with node 8

lib/dialects/oracle/index.js
```diff
@@ -80,10 +79,10 @@ Object.assign(Client_Oracle.prototype, {
   // Get a raw connection, called by the `pool` whenever a new
   // connection needs to be added to the pool.
   acquireRawConnection() {
-    return new Bluebird((resolver, rejecter) => {
+    return new Promise((resolver, rejecter) => {
       this.driver.connect(this.connectionSettings, (err, connection) => {
         if (err) return rejecter(err);
-        Bluebird.promisifyAll(connection);
+        connection.executeAsync = promisify(connection.execute);
```
maximelkin: here attention needed: check all file and notice, what from connection only execute (with suffix async) is used

lib/dialects/oracledb/index.js
```diff
@@ -254,96 +253,100 @@ Client_Oracledb.prototype._query = function(connection, obj) {
...
-
-    if (obj.method === 'update') {
-      const modifiedRowsCount = obj.rowsAffected.length || obj.rowsAffected;
-      const updatedObjOutBinding = [];
-      const updatedOutBinds = [];
-      const updateOutBinds = (i) =>
-        function(value, index) {
-          const OutBindsOffset = index * modifiedRowsCount;
-          updatedOutBinds.push(outBinds[i + OutBindsOffset]);
+  return connection
```
maximelkin: github is bad at diffing with spaces: here only Bluebird.resolve removed

lib/transaction.js
maximelkin: This will not work, because interface has non-trivial then (but this can be replaced to some internal method too, but this too complex

lib/dialects/oracle/index.js
briandamaged: FYI: I discovered earlier today that `lib/dialects/oracle` is actually partially-dead code.  It is not used anywhere except as a parent class for `lib/dialects/oracledb`.

So, you might want to double-check that this method is even used by `lib/dialects/oracledb`.  If not, then it might be easiest to just remove the entire method.

lib/transaction.js
maximelkin: helper which generates array of promise methods helps with optional .finally

test/integration/builder/inserts.js
```diff
@@ -1148,39 +1147,37 @@ module.exports = function(knex) {
         if (/redshift/i.test(knex.client.driverName)) {
           return;
         }
+        this.timeout(10000);
```
maximelkin: previously this was timeout on bluebird chain

lib/transaction.js
briandamaged: Right -- we're saying the same thing.  Basically: you can't use meta-programming for this specific case.  However, given that there are only a handful of cases now, it might be worth discarding the meta-programming in favor of something more explicit.

lib/util/promises.js
maximelkin: I checked #3507 pr

lib/dialects/oracle/index.js
maximelkin: It's easier to drop support for https://www.npmjs.com/package/oracle entirely, but this is not logical part of this merge request

briandamaged:
@maximelkin + @kibertoad : I've been reviewing the code, but the huge number of changes makes it difficult to be very confident about the code review (fun fact: it's actually 63 pages worth of changes).  Would it be possible to split this into at least 2 separate PRs?  In particular:

1. The first PR contains only the revisions to the `/test` code.  This way, we can at least be confident that the old code works with the new test cases.  (In other words: this helps us reduce the possibility of having 2 bugs that cancel one another out.  Ex:  a bug in the `/src` code that gets ignored due to a bug in the `/test` code)
2. The second PR contains the revisions to the actual code.  (And ideally: no more changes to the test cases.  This way, we can be confident that the code is doing the same thing before and after the changes)

(Ideally, I'd really like to see each of these split into 2-3 separate PRs as well.  But, splitting things this way will at least allow us to be a lot more confident in the code reviews).

In terms of strategy, you can do the following:

1. Create a copy of your branch, and rebase it against `master`:

```
git checkout bluebird_removal
git checkout -b bluebird_removal-rebase
git rebase -p master
```

(This will replay your commits against `master`, which ensures that we'll get a clean diff in the next step)

2. Create a patch of the changes against `master`:

```
git diff master > bluebird_removal.patch
```

3. Switch back to `master`, and apply only the test-related changes from the patch:

```
git checkout master
git apply --include="test/*" bluebird_removal.patch
```

4. Reintroduce the `chai-as-promised` dependency to `package.json`.  (Note: we can't use the branch's `package.json` since it also removed `bluebird`)

```
npm install --save-dev chai-as-promised
```

5. Create a new branch for these changes:

```
git checkout -b bluebird_removal-tests_only
```

At this point, it should be possible to run the tests against the older code.  Once we're confident that the tests are good, we can merge them.  Afterwards, we can apply the patch to create a new PR that contains the other half of the changes:

```
# Note: We're assuming that the test changes have already been merged into master
git checkout master
git apply --exclude="test/*" --exclude="package.json" bluebird_removal.patch
```

Again: I apologize for asking you to do this, but it would really allow everybody to be a lot more confident about the changes that were implemented.

lib/transaction.js
maximelkin: done: 9ef63698542dc7e9606d89b3d0b5055c9f47ba61
Sorry for delay, didn't quickly realize your proposed solution

maximelkin:
https://github.com/knex/knex/pull/3683
https://github.com/knex/knex/pull/3682

PR #3658, opened 2020-02-10
Add check to only create native enum once

### LINKED ISSUE #3657: Altering a colum to an enum with useNative: true fails, tries to create enum twice
# Environment

Knex version: 0.20.9
Database + version: Postgres 12
OS: Linux (Ubuntu)

# Bug

If you attempt to alter a column to change it's type to an enum, and you use the `useNative: true` enum option, knex will try to create the enum twice, which causes a "type already exists" error.

Here is an example migration which demonstrates the problem:

```typescript
export async function up(knex: Knex): Promise<any> {
  return knex.schema
    .debug(true)
    .createTable("orders", (table) => {
      table.text("drink");
    })
    .alterTable("orders", (table) => {
      // This one works
      table.enum("size", ["small", "medium", "large"], {
        enumName: "drink_sizes",
        useNative: true,
      });

      // This one tries to create the enum twice
      table
        .enum("drink", ["coffee", "tea", "water"], {
          enumName: "drink_types",
          useNative: true,
        })
        .alter();
    });
}
```

If you look at the debug output from that migration, you will see that the first enum in the alterTable part generates the sql fine, but the second one (using `.alter()`) will create two identical `create type` sql statements, which causes Postgres to fail with 'type "drink_types" already exists'

EDIT: 

tl;dr Altering table to enum seems to try to create column type twice:

https://runkit.com/embed/xemoadwjsojw

mogzol:
This will fix #3657


kibertoad:
Can you also add test for this? I think we had integration tests for enums already, should be easy to expand.

lib/dialects/postgres/schema/columncompiler.js
```diff
@@ -8,6 +8,7 @@ const { isObject } = require('lodash');
 function ColumnCompiler_PG() {
   ColumnCompiler.apply(this, arguments);
   this.modifiers = ['nullable', 'defaultTo', 'comment'];
+  this._nativeEnumCreated = false;
```
elhigu: Are you sure this works, when there are multiple native enums declared / altered in the same table? This is a column compiler, so it probably does, but anyways set of tests for this would be needed. Also other tests than just having the most trivial happy path tested (cases where multiple columns are created and some altered)

kibertoad:
@mogzol Do you need any help with this?

mogzol:
Oh sorry I forgot about this actually. I don't really have much free time right now if someone else want's to take this over, or wait until closer to the end of the year and I should have a chance to fix it.

mogzol:
Updated with some tests and a slightly different approach, now the result of the type call in `getColumnType` in the column compiler is cached so it's only run once (per column), which should fix this issue for any types that have side effects, instead of just fixing the specific postgres enu type.

kibertoad:
@mogzol Could you also please add an integration test to `test/integration2`?

lib/dialects/postgres/schema/columncompiler.js
mogzol: Added some tests

mogzol:
@kibertoad Yes, added some integration tests. I also found that altering native enum columns still failed because postgres doesn't allow casting directly between enum types, [they must be converted to text first](https://www.munderwood.ca/index.php/2015/05/28/altering-postgresql-columns-from-one-enum-to-another/), so I added a check in the postgres column compiler for that.

kibertoad (approved):

kibertoad:
Thank you!

PR #3659, opened 2020-02-12
include SSL in postgres connection definition

samjeffress:
Reflect the type at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pg/index.d.ts


kibertoad:
Released in 0.20.10.

PR #3661, opened 2020-02-12
Fix: Knex CLI calls process.chdir() before opening Knexfile

briandamaged:
Addresses the issue outlined here:

https://github.com/knex/knex/issues/3660

Short story: the Knex CLI was opening the Knexfile before calling `process.chdir(...)`.  Unfortunately, the Knexfile in question was resolving some paths relative to `process.cwd()`.  As a result, these paths ended up being relative to the wrong directory.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -29,6 +29,15 @@ const fsPromised = {
 };
 
 function initKnex(env, opts) {
+  checkLocalModule(env);
+  if (process.cwd() !== env.cwd) {
+    process.chdir(env.cwd);
+    console.log(
+      'Working directory changed to',
+      color.magenta(tildify(env.cwd))
+    );
+  }
+
   if (opts.esm) {
     // enable esm interop via 'esm' module
     require = require('esm')(module);
@@ -46,15 +55,6 @@ function initKnex(env, opts) {
 
     // TODO: Should this property be documented somewhere?
     env.configuration.ext = path.extname(p).replace('.', '');
-  }
-
-  checkLocalModule(env);
-  if (process.cwd() !== env.cwd) {
-    process.chdir(env.cwd);
-    console.log(
-      'Working directory changed to',
-      color.magenta(tildify(env.cwd))
-    );
   }
 
   const resolvedConfig = resolveEnvironmentConfig(opts, env.configuration);
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -28,6 +28,17 @@ const fsPromised = {
   writeFile: promisify(fs.writeFile),
 };
 
+function openKnexfile(configPath) {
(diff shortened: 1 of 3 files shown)


kibertoad:
Wasn't your preferred approach not to change process altogether?

briandamaged:
@kibertoad : It is; however, it looks like that is going to take a bit of work.  It looks like several modules within Knex are implicitly coupled with `process.cwd()`.  So, it will probably take a little while to revise those areas.

In the meantime, the current PR at least fixes this specific regression.  (But yeah -- arguably, the Knexfile itself should not be calculating paths relative to `process.cwd()` since this value will be inconsistent.  For example: `process.cwd()`  will be different if you load the Knexfile from your own application instead of the Knex CLI)

briandamaged:
@kibertoad : Hmm... the Travis CI failure seems to be an unrelated transaction failure.  Is there an easy way to kick off the job again?

kibertoad:
Thanks! I'll publish version with hotfix in the evening.

briandamaged:
@kibertoad : Thx!  And yeah, I'll look into the "decouple the code from `process.cwd()`" thing in the next few days.  After skimming through the code a bit, it looks like there's implicit coupling in at least half a dozen different places.  So, it's gonna take a little effort to figure out how to implement the change wo/ breaking backward compatibility in the process.

kibertoad:
Sorry, there will be a delay in releasing new version, there is a possible regression in timeout code that need investigation :-/

briandamaged:
np!

briandamaged:
@kibertoad : In terms of the timeouts, are you referring to the latest Travis-CI failure?

https://travis-ci.org/knex/knex/jobs/650139383?utm_medium=notification&utm_source=github_status

If so, I noticed that error recently as well.  It seems to be one of those fun "random" errors...

briandamaged:
@kibertoad : I think I've tracked down at least part of the problem.  Take a look at the following:

https://github.com/knex/knex/blob/77df705189c712105cfab293ae26fff41de8e6ae/lib/runner.js#L263

Suppose that `await cb(connection)` raises an unexpected error.  Further suppose that `await this.client.releaseConnection(this.connection);` also raises an error.  In this case, this 2nd error will interrupt the 2nd error before it has been properly handled, resulting in an "unhandled rejection".

kibertoad:
@maximelkin thoughts on explanation above?

maximelkin:
image
```code
async function f() {
 try {
    await Promise.reject(new Error('a1'));
  } finally {
  await Promise.reject(new Error('b2'));
 }
}
```
undefined
> f().catch(console.error)

maximelkin:
https://github.com/knex/knex/pull/3665

PR #3662, opened 2020-02-13
`fn.now` takes optionally a precision argument.

pmcnr-hx:
`fn.now` can optionally take a precision argument, so reflect that in the TypeScript definition.


kibertoad:
Released in 0.20.10.

PR #3665, opened 2020-02-14
fix unhandledrejection for rollback of non managed transaction

maximelkin:


maximelkin:
Somehow connected with switches between bluebird <-> native promises

briandamaged:
Agreed, and I suspect that `Promise.race(..)` might be involved as well.  The `mysql` / `mysql2` dialects are repeatedly failing for me for this specific test case:

https://github.com/knex/knex/blob/77df705189c712105cfab293ae26fff41de8e6ae/test/integration/builder/additional.js#L863

Both of these are implemented in terms of `Promise.race(..)`.  Additionally, it appears that there is a race between:

1. A standard node.js `setTimeout(..)` call
2. A resource acquisition within the `Tarn` library.  (Specifically, when attempting to obtain a connection from a pool.  The acquisition is timing out, but the promise rejection never ends up being handled)

lib/transaction.js
```diff
@@ -195,7 +195,7 @@ class Transaction extends EventEmitter {
 
             value = new Error(`Transaction rejected with non-error: ${value}`);
           }
-          this._rejecter(value);
+          setImmediate(() => this._rejecter(value));
```
briandamaged: I just tried out your fix, but I'm still observing the problem within this test case:

https://github.com/knex/knex/blob/77df705189c712105cfab293ae26fff41de8e6ae/test/integration/builder/additional.js#L863

Here's a sample of my output:

```
 % DB="mysql mysql2" KNEX_TEST_TIMEOUT=60000 ./node_modules/.bin/mocha --exit -t 10000 test/index.js
sqlite does not support inserting default values. Set the `useNullAsDefault` flag to hide this warning. (see docs http://knexjs.org/#Builder-insert).
sqlite does not support inserting default values. Set the `useNullAsDefault` flag to hide this warning. (see docs http://knexjs.org/#Builder-insert).
sqlite does not support inserting default values. Set the `useNullAsDefault` flag to hide this warning. (see docs http://knexjs.org/#Builder-insert).


  Integration Tests
    mysql | mysql
      Additional
Unhandled rejection: KnexTimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?
    at /Users/brianlauber/Documents/src/knex/lib/client.js:351:30
From previous event:
    at Client_MySQL.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/client.js:348:15)
    at Client_MySQL.cancelQuery (/Users/brianlauber/Documents/src/knex/lib/dialects/mysql/index.js:175:32)
    at /Users/brianlauber/Documents/src/knex/lib/runner.js:177:37
From previous event:
    at Runner.query (/Users/brianlauber/Documents/src/knex/lib/runner.js:169:13)
    at /Users/brianlauber/Documents/src/knex/lib/runner.js:41:23
    at /Users/brianlauber/Documents/src/knex/lib/runner.js:261:24
From previous event:
    at Runner.ensureConnection (/Users/brianlauber/Documents/src/knex/lib/runner.js:259:8)
    at Runner.run (/Users/brianlauber/Documents/src/knex/lib/runner.js:27:12)
    at Raw.Target.then (/Users/brianlauber/Documents/src/knex/lib/interface.js:15:43)
    at Context.<anonymous> (/Users/brianlauber/Documents/src/knex/test/integration/builder/additional.js:908:10)
    at callFn (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:395:21)
    at Test.Runnable.run (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:382:7)
    at Runner.runTest (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:541:10)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:667:12
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:450:14)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:460:7
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:362:14)
    at Immediate.<anonymous> (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:428:5)
    at processImmediate (internal/timers.js:439:21) {
  name: 'KnexTimeoutError'
}
        ✓ .timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection (5135ms)
    mysql | mysql2
      Additional
Unhandled rejection: KnexTimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?
    at /Users/brianlauber/Documents/src/knex/lib/client.js:351:30
From previous event:
    at Client_MySQL2.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/client.js:348:15)
    at Client_MySQL2.cancelQuery (/Users/brianlauber/Documents/src/knex/lib/dialects/mysql/index.js:175:32)
    at /Users/brianlauber/Documents/src/knex/lib/runner.js:177:37
From previous event:
    at Runner.query (/Users/brianlauber/Documents/src/knex/lib/runner.js:169:13)
    at /Users/brianlauber/Documents/src/knex/lib/runner.js:41:23
    at /Users/brianlauber/Documents/src/knex/lib/runner.js:261:24
From previous event:
    at Runner.ensureConnection (/Users/brianlauber/Documents/src/knex/lib/runner.js:259:8)
    at Runner.run (/Users/brianlauber/Documents/src/knex/lib/runner.js:27:12)
    at Raw.Target.then (/Users/brianlauber/Documents/src/knex/lib/interface.js:15:43)
    at Context.<anonymous> (/Users/brianlauber/Documents/src/knex/test/integration/builder/additional.js:908:10)
    at callFn (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:395:21)
    at Test.Runnable.run (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:382:7)
    at Runner.runTest (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:541:10)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:667:12
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:450:14)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:460:7
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:362:14)
    at Immediate.<anonymous> (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:428:5)
    at processImmediate (internal/timers.js:439:21) {
  name: 'KnexTimeoutError'
}
        ✓ .timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection (5125ms)


  2 passing (10s)

Unhandled rejections: 2
{
  reason: KnexTimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?
      at /Users/brianlauber/Documents/src/knex/lib/client.js:351:30
  From previous event:
      at Client_MySQL.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/client.js:348:15)
      at Client_MySQL.cancelQuery (/Users/brianlauber/Documents/src/knex/lib/dialects/mysql/index.js:175:32)
      at /Users/brianlauber/Documents/src/knex/lib/runner.js:177:37
  From previous event:
      at Runner.query (/Users/brianlauber/Documents/src/knex/lib/runner.js:169:13)
      at /Users/brianlauber/Documents/src/knex/lib/runner.js:41:23
      at /Users/brianlauber/Documents/src/knex/lib/runner.js:261:24
  From previous event:
      at Runner.ensureConnection (/Users/brianlauber/Documents/src/knex/lib/runner.js:259:8)
      at Runner.run (/Users/brianlauber/Documents/src/knex/lib/runner.js:27:12)
      at Raw.Target.then (/Users/brianlauber/Documents/src/knex/lib/interface.js:15:43)
      at Context.<anonymous> (/Users/brianlauber/Documents/src/knex/test/integration/builder/additional.js:908:10)
      at callFn (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:395:21)
      at Test.Runnable.run (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:382:7)
      at Runner.runTest (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:541:10)
      at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:667:12
      at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:450:14)
      at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:460:7
      at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:362:14)
      at Immediate.<anonymous> (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:428:5)
      at processImmediate (internal/timers.js:439:21) {
    name: 'KnexTimeoutError'
  }
}
{
  reason: KnexTimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?
      at /Users/brianlauber/Documents/src/knex/lib/client.js:351:30
  From previous event:
      at Client_MySQL2.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/client.js:348:15)
      at Client_MySQL2.cancelQuery (/Users/brianlauber/Documents/src/knex/lib/dialects/mysql/index.js:175:32)
      at /Users/brianlauber/Documents/src/knex/lib/runner.js:177:37
  From previous event:
      at Runner.query (/Users/brianlauber/Documents/src/knex/lib/runner.js:169:13)
      at /Users/brianlauber/Documents/src/knex/lib/runner.js:41:23
      at /Users/brianlauber/Documents/src/knex/lib/runner.js:261:24
  From previous event:
      at Runner.ensureConnection (/Users/brianlauber/Documents/src/knex/lib/runner.js:259:8)
      at Runner.run (/Users/brianlauber/Documents/src/knex/lib/runner.js:27:12)
      at Raw.Target.then (/Users/brianlauber/Documents/src/knex/lib/interface.js:15:43)
      at Context.<anonymous> (/Users/brianlauber/Documents/src/knex/test/integration/builder/additional.js:908:10)
      at callFn (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:395:21)
      at Test.Runnable.run (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:382:7)
      at Runner.runTest (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:541:10)
      at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:667:12
      at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:450:14)
      at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:460:7
      at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:362:14)
      at Immediate.<anonymous> (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:428:5)
      at processImmediate (internal/timers.js:439:21) {
    name: 'KnexTimeoutError'
  }
}
```

maximelkin:
$ node
async function main() {
   const a = Promise.reject(new Error('A!'))
    Promise.resolve(a.then(console.log)).catch(e => console.log('handled1'))
     a.then(() => console.log('never'))
}

> handled1
(node:7931) UnhandledPromiseRejectionWarning: Error: A!
    at main (repl:2:26)
    at repl:1:1
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at REPLServer.defaultEval (repl.js:240:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:468:10)
    at emitOne (events.js:121:20)
    at REPLServer.emit (events.js:211:7)
    at REPLServer.Interface._onLine (readline.js:280:10)
(node:7931) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:7931) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.


this is case of 2 unhandled rejections

lib/dialects/mysql/index.js
```diff
@@ -190,7 +190,10 @@ Object.assign(Client_MySQL.prototype, {
       .finally(() => {
         // NOT returning this promise because we want to release the connection
         // in a non-blocking fashion
-        acquiringConn.then((conn) => this.releaseConnection(conn));
+        acquiringConn
+          .then((conn) => this.releaseConnection(conn))
+          // error handled in another chain of promises
+          .catch(() => {});
```
briandamaged: I think this might just be obscuring/hiding the underlying problem.

I agree that there is a race between 2 separate promise chains, and one of them is releasing the connection first.  However, it seems like the code should be restructured in a manner where the "release the connection" responsibility occurs _after_ the race has been resolved.  That is, instead of doing this:

```
RACE
  - pool.acquire().catch(()=>  releaseConnection())
  - delay().catch(()=> releaseConnection())
```

The code should be structured like this:

```
try {
  RACE
    - pool.acquire()
    - delay()
} catch() {
  releaseConnection()
}
```

(Sry -- I'm being sloppy about the Promise notation.  The main concept is that you shouldn't have 2 promise chains that each take responsibility for releasing the connection.  Instead, you should conclude the race, and then release the promise chain in 1 location)

maximelkin: we cant releaseConnection  in catch/finally, if pool.acquire() in try block, because we doesn't have reference to it

maximelkin: Maybe pool.acquire should have timeout

maximelkin: Ok, we can pass connectTimeout while acquiring connection in mysql: https://github.com/mysqljs/mysql#connection-options

maximelkin: Also pg does not give this error because acquiring connection not limited by time

briandamaged: So, Knex is also providing a pooling layer on top of MySQL via the Tarn library.  One of the "timeouts" that we are observing is actually coming from Tarn.  In particular, take a look at line 351 below:

https://github.com/knex/knex/blob/77df705189c712105cfab293ae26fff41de8e6ae/lib/client.js#L348-L357

Typically, the Tarn connection pool would already be populated with MySQL connections.  Hence, the "timeout" doesn't necessarily mean that a connection w/ MySQL timed out.  It could also mean:

1. The Tarn pool was depleted, and no connections were returned to the pool in time
2. The Tarn pool was destroyed by mistake.  (I already checked for this, but this did not appear to be the case)
3. Tarn keeps trying to allocate a new resource for the pool, but it's repeatedly failing.  Eventually, the "acquisition" timeout expires.  (This could be the case if MySQL connections are repeatedly being destroyed before Tarn has an opportunity to hand them off)

maximelkin: we can request tarn.js add timeout/cancellation of resource request

briandamaged: k -- new observation.  The testcase fails differently if you change the pool size to 2:

https://github.com/knex/knex/blob/77df705189c712105cfab293ae26fff41de8e6ae/test/integration/builder/additional.js#L877

```
  1) Integration Tests
       mysql | mysql
         Additional
           .timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection:

      AssertionError: expected { Object (timeout, name, ...) } to deeply equal { Object (timeout, name, ...) }
      + expected - actual

       {
      -  "message": "Defined query timeout of 1ms exceeded when running query."
      +  "message": "After query timeout of 1ms exceeded, cancelling of query failed."
         "name": "KnexTimeoutError"
         "timeout": 1
       }
      
      at /Users/brianlauber/Documents/src/knex/test/integration/builder/additional.js:912:71
  From previous event:
      at Context.<anonymous> (test/integration/builder/additional.js:911:15)
      at processImmediate (internal/timers.js:439:21)
```

It seems that the code is attempting to establish 2 separate connections for some reason.  So, there are 2 separate timeouts occurring:

1. The "1 ms" timeout as instructed by the testcase itself
2. The "100 ms" timeout for the pool resource acquisition

Furthermore, take a look at line 181 below:

https://github.com/knex/knex/blob/77df705189c712105cfab293ae26fff41de8e6ae/lib/dialects/mysql/index.js#L174-L196

For whatever reason, the cancelation operation is also introducing a 100ms delay.  So, I suspect the following is happening:

1. The initial connection times out after roughly 1ms
2. For whatever reason, the cancelation operation is mistakenly establishing a 2nd connection.
3. MEANWHILE, it is waiting 100ms before actually canceling the first connection
4. By then, the Tarn resource acquisition has already timed out for the 2nd connection

briandamaged: ... which makes sense.  It looks like the test case is intended to handle exactly this scenario.  The issue is that it's then failing to handle the Tarn acquisition error itself.

maximelkin: .timeout(1) is for initial query
timeout(100) is for query which cancelling initial query

maximelkin: cancellation operation establishing second connection because first connection is hanged, we cannot queue 2 parallel operations in one connection

maximelkin: pool size should be = 1, because test reproducing case of timeout acquiring connection for cancel query

briandamaged: I see exactly how/why the issue is occurring.  I'm just not 100% sure what to do about it at this point.  🤔 

Basically: the `cancelQuery(..)` method is written so that it tries to "give up" if it's unable to establish a secondary connection after 100ms.  So, the sequence of events is:

1. It starts trying to establish a secondary connection.
2. After 100ms, it finally gives up.
3. HOWEVER, it still invokes the `.finally(..)` block.
4. The `.finally(..)` block calls `acquiringConn.then(..)` again, which establishes a _second_ Promise chain.
5. The Tarn resource acquisition timeout is never handled in this second Promise chain.

Interestingly, the Tarn resource acquisition error _is_ being handled by the call to `Promise.race(..)`.  Unfortunately, after the race has already been concluded, the second promise chain is established, and the error is dropped on the floor.

So, I guess my main question: why is the `cancelQuery(..)` method enforcing a 100ms timeout in the first place?  Shouldn't it just rely upon the pool's settings instead?

cc: @maximelkin  + @kibertoad

maximelkin: @bywo

briandamaged: @bywo + @maximelkin : So yeah: my main question is "Do we actually need the 100ms timeout that is baked into the `cancelQuery(..)` method?".  Ex: why not just do this?

```
    const conn = await this.acquireConnection();
    try {
      await this.query(conn, {
        method: 'raw',
        sql: 'KILL QUERY ?',
        bindings: [connectionToKill.threadId],
        options: {},
      });
    } finally {
      await this.releaseConnection(conn);
    }
  },
```

This eliminates the race condition and ensures that the Tarn resource allocation error can be raised/caught.

briandamaged: FYI: Just created an alternative merge request w/ this proposed solution:

https://github.com/knex/knex/pull/3666

All of the tests seem to be passing on my local machine, at least.  We'll see what Travis-CI has to say about it.

kibertoad:
@maximelkin Is this still useful with https://github.com/knex/knex/pull/3666 merged?

maximelkin:
@kibertoad partially - need to revert lib/dialects/mysql/index.js

maximelkin:
Third unhandled rejection is disappear, I will re-read code again later, because don't understand why

lib/transaction.js
briandamaged: Q: what problem is this line intended to fix?

maximelkin: Unhandled rejection Error: Transaction rejected with non-error: undefined
from https://travis-ci.org/knex/knex/jobs/650139383?utm_medium=notification&utm_source=github_status

maximelkin: All troubles are going from
```js

trx.select()
// ...blablabla...

await trx.executionPromise
```

briandamaged: Gotcha.  Have you been able to recreate this error after #3666  was merged?  (The link you provided looks like it's from a Travis-CI run that is 5 days old)

maximelkin: Yes, locally on node 8

briandamaged: Interesting.  Are you seeing the issue locally when you run the tests?  Or, are you only seeing it when you run the snippet that you provided above?  ☝️ 

So far, I have been unable to recreate the issue when running the tests locally.  I'll give it another try once I get back to my apartment.  (in the next 30 mins or so)

maximelkin: knex.transaction() - returns this.initPromise
executionPromise - rejected because of 
1) lib/transactions.js#78 - without handler executionPromise returned at line 109 and will not produce unhandled rejection
2) lib/transactions.js#198

maximelkin: "rejects execution promise if there was a manual rollback and transaction is set to reject" - this test

sqlite3, node 8, several retries

maximelkin: I dont understand how this supposed to work, lol

maximelkin: I see first implementation and examples - no executionPromise .catch required
And then I see https://github.com/knex/knex/pull/3328

maximelkin: For easy reproduction add delay after await trx.rollback()

briandamaged: k -- just got back, so I'll try recreating the issue in the next few minutes.

One other thing I noticed: it appears that some test cases will "fail" if you omit a specific database from the `DB="..."` environment variable.  For instance: you will observe "failures" if you run the tests via:

```
DB="mysql mysql2" KNEX_TEST_TIMEOUT=60000 ./node_modules/.bin/mocha --exit -t 10000 test/index.js
```

However, these are not actually failures in the Knex library itself.  It's actually a problem w/ the way the test cases are written.  More specifically: the test cases are referencing data that is only available when `"sqlite3"` is specified within `DB="..."` .  (BTW: I'm going to create a separate issue for this)

So, with this in mind: what command are you running to kick off the tests?  It might provide some insight into why the tests are failing.

maximelkin: @kibertoad can you help with explanation, how .executionPromise should work?
executionPromise must be awaited immediately, or after transaction commit/rollback?

maximelkin: @briandamaged  
```
mocha ./test/unit/knex.js --grep "^knex rejects execution promise if there was a manual rollback and transaction is set to reject$"
```

maximelkin:
TODO:
lib/transaction.js line 275 handle trx.transaction rejection

lib/transaction.js
```diff
@@ -32,6 +32,11 @@ class Transaction extends EventEmitter {
         this.initRejectFn = reject;
         container = (transactor) => {
           resolve(transactor);
+          // because user must await executionPromise in same tick
+          // todo^ is this true?
+          transactor.executionPromise = transactor.executionPromise.catch((err) => {
+            throw err;
+          });
```
briandamaged: Hmm... I suspect that this isn't actually addressing the underlying race condition.  Instead, it is just adding a slight delay in the event processing queue, which makes the race condition less likely to occur.

maximelkin: Yes, this change is wrong too

lib/transaction.js
kibertoad: @maximelkin executionPromise is a way to expose state of transaction to the user, so that he can e. g. handle the rollback or proceed with the rest of the flow after commit has happened. Having that in mind, I'd say that it should be awaited after transaction commit/rollback.
Maybe we need to rename the field to make it more explicit, e. g. transactionPromise.

briandamaged: @maximelkin : I've been able to recreate the failure locally using the command-line you provided.  I've identified a few suspicious sections of code, but so far none of them seem to directly impact this specific issue.

For example: take a look at the following:

https://github.com/knex/knex/blob/72886086b65583b395d2c34851ef60a5a6d5fd75/lib/transaction.js#L228-L233

These lines seem strange for at least 2 reasons:

1. There is a `.catch(()=> {})` statement, which might be hiding other bugs.  (More specifically: it might be causing Node to overlook some "unresolved promises".  This will require some more investigation to confirm this)
2. The conditional logic seems to be largely unnecessary.  In particular, `this._previousSibling` should _always_ be defined thanks to the following lines in the constructor:

https://github.com/knex/knex/blob/72886086b65583b395d2c34851ef60a5a6d5fd75/lib/transaction.js#L124-L128

But yeah: so far, these oddities don't seem to have any direct impact on the issue you're observing.

cc: @kibertoad

briandamaged:
@maximelkin + @kibertoad : Here's a new piece of info that might help w/ the investigation.

First, here is the stacktrace that is attached to the "unresolved promise":

```
Error: Transaction rejected with non-error: undefined
    at q.trxClient.query.catch.then (/Users/brianlauber/Documents/src/knex/lib/transaction.js:196:21)
From previous event:
    at Transaction._promise.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/transaction.js:67:32)
    at Bluebird.then.then (/Users/brianlauber/Documents/src/knex/lib/transaction.js:236:24)
From previous event:
    at Transaction.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/transaction.js:234:8)
    at new Transaction (/Users/brianlauber/Documents/src/knex/lib/transaction.js:58:26)
    at Client_SQLite3.transaction (/Users/brianlauber/Documents/src/knex/lib/client.js:125:12)
    at Function.transaction (/Users/brianlauber/Documents/src/knex/lib/util/make-knex.js:40:31)
    at Context.it.only (/Users/brianlauber/Documents/src/knex/test/unit/knex.js:403:28)
    at callFn (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:395:21)
    at Test.Runnable.run (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:382:7)
    at Runner.runTest (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:541:10)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:667:12
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:450:14)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:460:7
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:362:14)
    at Immediate._onImmediate (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:428:5)
    at runCallback (timers.js:810:20)
    at tryOnImmediate (timers.js:768:5)
    at processImmediate [as _immediateCallback] (timers.js:745:5)
```

I modified the test case so that it prints out the stacktrace for the exception that it catches.  Interestingly, *the value that gets printed changes depending upon whether or not the "unresolved promise" occurred.

When the "unresolved promise" occurs, the test case prints the following stacktrace:

```
Error: Transaction rejected with non-error: undefined
    at q.trxClient.query.catch.then (/Users/brianlauber/Documents/src/knex/lib/transaction.js:196:21)
From previous event:
    at Transaction._promise.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/transaction.js:67:32)
    at Bluebird.then.then (/Users/brianlauber/Documents/src/knex/lib/transaction.js:236:24)
From previous event:
    at Transaction.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/transaction.js:234:8)
    at new Transaction (/Users/brianlauber/Documents/src/knex/lib/transaction.js:58:26)
    at Client_SQLite3.transaction (/Users/brianlauber/Documents/src/knex/lib/client.js:125:12)
    at Function.transaction (/Users/brianlauber/Documents/src/knex/lib/util/make-knex.js:40:31)
    at Context.it.only (/Users/brianlauber/Documents/src/knex/test/unit/knex.js:403:28)
    at callFn (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:395:21)
    at Test.Runnable.run (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:382:7)
    at Runner.runTest (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:541:10)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:667:12
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:450:14)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:460:7
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:362:14)
    at Immediate._onImmediate (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:428:5)
    at runCallback (timers.js:810:20)
    at tryOnImmediate (timers.js:768:5)
    at processImmediate [as _immediateCallback] (timers.js:745:5)
Error: Transaction rejected with non-error: undefined
    at q.trxClient.query.catch.then (/Users/brianlauber/Documents/src/knex/lib/transaction.js:196:21)
From previous event:
    at Transaction._promise.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/transaction.js:67:32)
    at Bluebird.then.then (/Users/brianlauber/Documents/src/knex/lib/transaction.js:236:24)
From previous event:
    at Transaction.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/transaction.js:234:8)
    at new Transaction (/Users/brianlauber/Documents/src/knex/lib/transaction.js:58:26)
    at Client_SQLite3.transaction (/Users/brianlauber/Documents/src/knex/lib/client.js:125:12)
    at Function.transaction (/Users/brianlauber/Documents/src/knex/lib/util/make-knex.js:40:31)
    at Context.it.only (/Users/brianlauber/Documents/src/knex/test/unit/knex.js:403:28)
    at callFn (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:395:21)
    at Test.Runnable.run (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:382:7)
    at Runner.runTest (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:541:10)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:667:12
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:450:14)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:460:7
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:362:14)
    at Immediate._onImmediate (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:428:5)
    at runCallback (timers.js:810:20)
    at tryOnImmediate (timers.js:768:5)
    at processImmediate [as _immediateCallback] (timers.js:745:5)
```

Notice that this appears to contain the "unresolved promise" stacktrace as a subcomponent.

In contrast, here is what gets printed when the "unresolved promise" does NOT occur:

```
Error: Transaction rejected with non-error: undefined
    at q.trxClient.query.catch.then (/Users/brianlauber/Documents/src/knex/lib/transaction.js:196:21)
From previous event:
    at Transaction._promise.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/transaction.js:67:32)
    at Bluebird.then.then (/Users/brianlauber/Documents/src/knex/lib/transaction.js:236:24)
From previous event:
    at Transaction.acquireConnection (/Users/brianlauber/Documents/src/knex/lib/transaction.js:234:8)
    at new Transaction (/Users/brianlauber/Documents/src/knex/lib/transaction.js:58:26)
    at Client_SQLite3.transaction (/Users/brianlauber/Documents/src/knex/lib/client.js:125:12)
    at Function.transaction (/Users/brianlauber/Documents/src/knex/lib/util/make-knex.js:40:31)
    at Context.it.only (/Users/brianlauber/Documents/src/knex/test/unit/knex.js:403:28)
    at callFn (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:395:21)
    at Test.Runnable.run (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runnable.js:382:7)
    at Runner.runTest (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:541:10)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:667:12
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:450:14)
    at /Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:460:7
    at next (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:362:14)
    at Immediate._onImmediate (/Users/brianlauber/Documents/src/knex/node_modules/mocha/lib/runner.js:428:5)
    at runCallback (timers.js:810:20)
    at tryOnImmediate (timers.js:768:5)
    at processImmediate [as _immediateCallback] (timers.js:745:5)
```

As you can see, it matches the stacktrace that would have been included in the "unresolved promise".

So: I don't have any specific conclusions related to this yet.  But, I suspect that this will help pinpoint the source of the race condition.

briandamaged:
@maximelkin + @kibertoad : Update on my "different stacktraces" message  (https://github.com/knex/knex/pull/3665#issuecomment-587951963)

Looks like that might have been a false alarm.  It was a quirk in the console output that caused it to _look_ like a really long stacktrace.  In reality, the error was actually getting printed from 2 different places.  (Which is why there the weird disconnect in the middle of the stacktrace)

briandamaged:
@maximelkin + @kibertoad : FYI  -- I created another PR for you to look at over here:

#3671 

It was initially intended as a minor code cleanup; however, it might have inadvertently fixed the race condition as well.

PR #3666, opened 2020-02-15
Fixed unresolved promise in cancelQuery(..) ...

briandamaged:
... more specifically: cancelQuery(..) was attempting to
"cancel the cancellation" after 100ms.  However, it was not
actually achieving this objective.  In reality, the cancellation
was still running in the background even though the caller had
already moved on.

Later on, the cancellation would ACTUALLY fail due to a resource
allocation issue (ie: no more connections in the Tarn pool).
This would then result in an unhandled Promise rejection.

### DIFF
--- a/lib/dialects/mysql/index.js
+++ b/lib/dialects/mysql/index.js
@@ -171,27 +171,18 @@ Object.assign(Client_MySQL.prototype, {
 
   canCancelQuery: true,
 
-  cancelQuery(connectionToKill) {
-    const acquiringConn = this.acquireConnection();
-
-    // Error out if we can't acquire connection in time.
-    // Purposely not putting timeout on `KILL QUERY` execution because erroring
-    // early there would release the `connectionToKill` back to the pool with
-    // a `KILL QUERY` command yet to finish.
-    return timeout(acquiringConn, 100)
-      .then((conn) =>
-        this.query(conn, {
-          method: 'raw',
-          sql: 'KILL QUERY ?',
-          bindings: [connectionToKill.threadId],
-          options: {},
-        })
-      )
-      .finally(() => {
-        // NOT returning this promise because we want to release the connection
-        // in a non-blocking fashion
-        acquiringConn.then((conn) => this.releaseConnection(conn));
+  async cancelQuery(connectionToKill) {
+    const conn = await this.acquireConnection();
+    try {
+      return await this.query(conn, {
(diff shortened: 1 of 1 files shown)


lib/dialects/mysql/index.js
```diff
@@ -171,27 +171,18 @@ Object.assign(Client_MySQL.prototype, {
...
   canCancelQuery: true,
 
-  cancelQuery(connectionToKill) {
-    const acquiringConn = this.acquireConnection();
-
-    // Error out if we can't acquire connection in time.
-    // Purposely not putting timeout on `KILL QUERY` execution because erroring
-    // early there would release the `connectionToKill` back to the pool with
-    // a `KILL QUERY` command yet to finish.
-    return timeout(acquiringConn, 100)
```
briandamaged: Short story: I remove the `100ms` timeout here.

I think the logic was intended to "cancel the cancellation" if it took too long.  In reality, it was just returning control to the caller after 100ms; yet the actual cancellation operation was still running.

briandamaged:
@maximelkin  + @kibertoad  + @bywo :  Proposing this as the fix for the "timeout" issue that was being observed.  Although the PR at #3665 makes the unhandled promise issue go away, it is ultimately just obscuring the underlying problem.  (the underlying problem being: there is an exception that cannot be caught/handled by the caller)

kibertoad:
This is very old code from 2016, wonder if it still makes sense with how Tarn.js works.
@elhigu @koskimas Any thoughts? For me new behaviour makes more sense, but this is obviously a pretty sensitive part of the code.

elhigu:
Sounds like the reason why timeout was there was to prevent case where connection could not have been acquired for cancelation. 

This implementation looks fine to me. AFAIK Tarn's acquire connection timeout should automatically handle this case as it is used in this fix.

However it should be made sure that other dialect cancelation methods are fixed as well.

kibertoad:
@elhigu Only other dialect that seem to have `cancelQuery` implementation is Postgre, and it doesn't have this kind of logic already, which would explain why we were getting this test fail only on mysql.

kibertoad:
Thank you so much!

briandamaged:
np!  I'm just glad we were able to find the cause for the intermittent failures!

Thx to @maximelkin  as well for helping to brainstorm about possible race conditions / etc.

Issue #3225, opened 2019-05-26
Flaky Oracle test


kibertoad:
# Environment

Knex version: 0.17.0
Database + version: Oracle
Node 6

From CI:

```
 1) oracle | oracledb
       knex.migrate
         knex.migrate.up
           should only run the first migration if no migrations have run:
      AssertionError: expected [] to have a length of 1 but got 0
      + expected - actual
      -0
      +1
      
      at /home/travis/build/tgriesser/knex/test/integration/migrate/index.js:534:38
  From previous event:
      at Builder.Target.then (lib/interface.js:38:24)
      at knex.migrate.up.then (test/integration/migrate/index.js:533:16)
  From previous event:
      at Context.<anonymous> (test/integration/migrate/index.js:530:12)
```

kibertoad:
Observed on Node 10:
```
  1) oracle | oracledb
       knex.migrate
         knex.migrate.status
           should return 0 if code matches DB:
      AssertionError: expected -1 to equal 0
      + expected - actual
      --1
      +0
      
      at /home/travis/build/tgriesser/knex/test/integration/migrate/index.js:160:39
  From previous event:
      at Context.<anonymous> (test/integration/migrate/index.js:159:12)
  From previous event:
```

elhigu:
Thanks. I'll fix those. Probably adding forUpdate() to some query when checking value from DB is enough. There are already other disabled tests which will catch this error.

kibertoad:
Released in 0.20.10.

PR #3668, opened 2020-02-16
oracledb: commit was a no-op causing race conditions 

### LINKED ISSUE #3612: oracledb: commit is a no-op
# Environment

Knex version: 0.20.6
OS: Linux & OSX

# Bug

The oracledb commit is a no-op. I think it should perform the actual commit immediately.

2. Error message

There's no error message.

I noticed this in our production code that the commit seemed to not work immediately. The integration tests also seem to exhibit this behavior which is why it may fail sometimes (race conditions.)

Here's the code path it takes:

1. Transaction starts and it sets [isTransaction=true](https://github.com/knex/knex/blob/aea29d0a7c00a4e6e0f6eafdf57240ef211813cb/lib/dialects/oracledb/transaction.js#L60)
2. When commit is called it calls `commitAsync` which [immediately resolves doing nothing](https://github.com/knex/knex/blob/aea29d0a7c00a4e6e0f6eafdf57240ef211813cb/lib/dialects/oracledb/index.js#L112)
3. Later it sets [isTransaction=false](https://github.com/knex/knex/blob/aea29d0a7c00a4e6e0f6eafdf57240ef211813cb/lib/dialects/oracledb/transaction.js#L72) so it really does the commit this time.

The problem is in the tests and in my production code I assume the commit actually happened when commit was called, so we occasionally see errors due to the race condition it creates.

jakecoffman:
fix #3612


kibertoad:
@akaraivanov @jakecoffman I assume this still needs new unit tests with Sinon mocks?

kibertoad:
refs #3225

jakecoffman:
I'm trying to see if the integration tests I enabled will fail in travis so I am running them repeatedly. Would that be enough to say that this fix is tested? It would probably be more meaningful than a mocked out db call.

kibertoad:
Yeah, fair :)

kibertoad:
Seem to be green, so at least not worse than before. Let's land this and observe oracle behaviour in master over longer period of time.

PR #3671, opened 2020-02-19
Fix race condition in non-container transactions

briandamaged:
Minor cleanup to some of the Transaction logic.  However, it looks like this cleanup might have also inadvertently fixed the race condition that was being investigated here:

https://github.com/knex/knex/pull/3665

cc: @maximelkin + @kibertoad

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -225,12 +225,11 @@ class Transaction extends EventEmitter {
       .then((connection) => {
         connection.__knexTxId = this.txid;
 
-        return (this._previousSibling
-          ? this._previousSibling.catch(() => {})
-          : Promise.resolve()
-        ).then(function() {
-          return connection;
-        });
+        return this._previousSibling
+          .catch(() => {}) // TODO: Investigate this line
+          .then(function() {
+            return connection;
+          });
       })
       .then(async (connection) => {
         try {
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -55,7 +55,11 @@ class Transaction extends EventEmitter {
       outerTx ? 'nested' : 'top level'
     );
 
-    this._promise = this.acquireConnection(config, (connection) => {
+    // FYI: As you will see in a moment, this Promise will be used to construct
+    //      2 separate Promise Chains.  This ensures that each Promise Chain
+    //      can establish its error-handling semantics without interfering
+    //      with the other Promise Chain.
+    const basePromise = this.acquireConnection(config, (connection) => {
       const trxClient = (this.trxClient = makeTxClient(
         this,
         client,
@@ -109,6 +113,11 @@ class Transaction extends EventEmitter {
       }
     });
 
+
+    // FYI: This is the Promise Chain for EXTERNAL use.  It ensures that the
+    //      caller must handle any exceptions that result from `basePromise`.
+    this._promise = basePromise.then((x)=> x);
+
     this._completed = false;
 
     // If there's a wrapping transaction, we need to wait for any older sibling
@@ -118,7 +127,11 @@ class Transaction extends EventEmitter {
     this._previousSibling = Bluebird.resolve(true);
     if (outerTx) {
       if (outerTx._lastChild) this._previousSibling = outerTx._lastChild;
-      outerTx._lastChild = this._promise;
+
+      // FYI: This is the Promise Chain for INTERNAL use.  It serves as a signal
+      //      for when the next sibling should begin its execution.  Therefore,
+      //      exceptions are caught and ignored.
+      outerTx._lastChild = basePromise.catch(()=> {});
     }
   }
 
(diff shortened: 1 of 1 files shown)


lib/transaction.js
```diff
@@ -74,13 +74,7 @@ class Transaction extends EventEmitter {
...
         })
         .then((transactor) => {
-          if (this.initPromise) {
-            transactor.executionPromise = executionPromise.catch((err) => {
-              throw err;
-            });
-          } else {
-            transactor.executionPromise = executionPromise;
-          }
+          transactor.executionPromise = executionPromise;
```
briandamaged: @maximelkin :  I'm fairly certain that this change is the one that fixed the race condition.

Your PR (#3665) included this change as well; however, it relocated the `.catch((err)=> {throw err})` logic to a different part of the code.  In contrast, this PR just removes that logic entirely.

lib/transaction.js
```diff
@@ -225,12 +219,11 @@ class Transaction extends EventEmitter {
...
-          ? this._previousSibling.catch(() => {})
-          : Promise.resolve()
-        ).then(function() {
-          return connection;
-        });
+        return this._previousSibling
+          .catch(() => {}) // TODO: Investigate this line
+          .then(function() {
+            return connection;
+          });
```
briandamaged: This was just a minor code cleanup.  It turns out that `this._previousSibling` is _always_ defined due to the way the constructor is written.  So, there was no need for conditional logic here.

Additionally, take note of the `TODO`.  I'm 95% sure that the code should _not_ be ignoring that exception.  However, I left it in place for now so that we can minimize the impact of this PR.

briandamaged:
At a high-level, it seems like Knex is blurring the lines between:

1. The `Transaction`
2. The `Transactor`.  (The thing that looks like a Knex Client, but it's actually scoped to a single transaction)
3. The operational logic around the `Transactor`.  ("If the callback to `knex.transaction(..)` returns a `Promise`, then automatically commit or rollback the `Transactor` depending upon the outcome.  Otherwise, assume that the caller will be responsible for committing/rolling back the `Transactor`")

In the current version of the Knex code, the `Transaction` class is responsible for instantiating the `Transactor` and handling the `Transactor`'s operational logic as well.  I suspect that we should extract these responsibilities into a higher layer of the code, such as here:

https://github.com/knex/knex/blob/72886086b65583b395d2c34851ef60a5a6d5fd75/lib/util/make-knex.js#L39-L50

lib/transaction.js
maximelkin: This catch added in #3328, so I am unsure about this changes

lib/transaction.js
briandamaged: Actually... I suspect that the `.catch(()=> {})` line was the reason for the "silent failures" mentioned here:

https://github.com/knex/knex/pull/3328#issuecomment-508403321

I'm going to test out a different implementation that will (hopefully!) allow for proper sibling->sibling linking without swallowing exceptions.

briandamaged: Alright -- I just pushed the code for establishing 2 separate Promise Chains.  So, instead of doing this:

```
this._promise = createPromise()
outer._lastChild = this._promise.catch(()=> {})
// Oops!  Now `this.promise` will always think that its exceptions have been handled!
```

It is doing this:

```
const basePromise = createPromise();
this._promise = basePromise.then((x)=> x)
outer._lastChild = this._promise.catch(()=> {})
// Yay!  Now `this._promise` understands that its exceptions still need to be handled!
```

lib/transaction.js
```diff
@@ -225,12 +232,11 @@ class Transaction extends EventEmitter {
       .then((connection) => {
         connection.__knexTxId = this.txid;
 
-        return (this._previousSibling
-          ? this._previousSibling.catch(() => {})
-          : Promise.resolve()
-        ).then(function() {
-          return connection;
-        });
+        return this._previousSibling
```
maximelkin: acquireConnection also has overrides in several dialects, check them please

briandamaged: The `Transaction`'s `acquireConnection` method delegates to the underlying `Dialect`'s `acquireConnection` method:

https://github.com/knex/knex/blob/72886086b65583b395d2c34851ef60a5a6d5fd75/lib/transaction.js#L216-L224

So, there isn't actually an inheritance relationship there.

briandamaged: Actually, nvm.  I see what you mean. I'll check it in a little bit.

(Related note: seems like there is a lot of duplication within those implementations)

briandamaged: So, the other implementations are actually ignoring the details around `this._previousSibling`.  That is potentially a problem, but not a problem that was introduced by this specific change.

briandamaged:
@maximelkin + @kibertoad : This current PR seems to fix the "unresolved Promise" issue that we were observing.  However, I've actually done some additional cleanup/fixing as well:

https://github.com/briandamaged/knex/pull/1/files

This builds upon the existing PR and untangles the `Transaction` / operational logic.  So, the `Transaction` class no longer needs to change its behavior depending upon how it was called.  But, I'm keeping this as a separate PR for now just in case we want to merge things more incrementally.

kibertoad:
This looks great, thanks a lot!

kibertoad:
@briandamaged Can you also create the second PR?

briandamaged:
@kibertoad : Sure!

PR #3673, opened 2020-02-19
Cleanup and extract Transaction Workflow logic

briandamaged:
This extracts the "operational" logic related to container / non-container Transactions.  Saying it another way: it means that the `Transaction` class no longer needs to be aware of "how" it was instantiated.

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -14,23 +14,11 @@ const { uniqueId, isUndefined } = require('lodash');
 // Acts as a facade for a Promise, keeping the internal state
 // and managing any child transactions.
 class Transaction extends EventEmitter {
-  constructor(client, container, config, outerTx) {
+  constructor(client, container, config = {}, outerTx = null) {
     super();
     this.doNotRejectOnRollback = config.doNotRejectOnRollback;
 
     const txid = (this.txid = uniqueId('trx'));
-
-    // If there is no container provided, assume user wants to get instance of transaction and use it directly
-    if (!container) {
-
-
-      this.initPromise = new Promise((resolve, reject) => {
-        this.initRejectFn = reject;
-        container = (transactor) => {
-          resolve(transactor);
-        };
-      });
-    }
 
     this.client = client;
     this.logger = client.logger;
@@ -94,12 +82,6 @@ class Transaction extends EventEmitter {
         });
 
       return executionPromise;
-    }).catch((err) => {
-      if (this.initRejectFn) {
-        this.initRejectFn(err);
-      } else {
-        throw err;
-      }
     });
 
 
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -16,17 +16,13 @@ const { uniqueId, isUndefined } = require('lodash');
 class Transaction extends EventEmitter {
   constructor(client, container, config, outerTx) {
     super();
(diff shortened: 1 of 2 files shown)


lib/transaction.js
```diff
@@ -14,35 +14,12 @@ const { uniqueId, isUndefined } = require('lodash');
 // Acts as a facade for a Promise, keeping the internal state
 // and managing any child transactions.
 class Transaction extends EventEmitter {
-  constructor(client, container, config, outerTx) {
+  constructor(client, container, config = {}, outerTx = null) {
     super();
+    this.doNotRejectOnRollback = config.doNotRejectOnRollback;
```
kibertoad: Doesn't this change default value to `false` from current `true`?

lib/transaction.js
```diff
@@ -225,12 +203,11 @@ class Transaction extends EventEmitter {
...
 
-        return (this._previousSibling
-          ? this._previousSibling.catch(() => {})
-          : Promise.resolve()
-        ).then(function() {
-          return connection;
-        });
+        return this._previousSibling
+          // .catch(() => {}) // TODO: Investigate this line
+          .then(function() {
```
kibertoad: probably can be replaced with async/await?

lib/transaction.js
briandamaged: Yes and No.  It appears that `Transaction` construction is always kicked off from `lib/util/make-knex.js` .  The logic for handling defaults has been extracted to that location.

But... to be on the safe side, we might also want to provide an ULTRA-default value within the `Transaction` constructor as well.

kibertoad: Right, I see, you handle default values elsewhere now.

lib/transaction.js
briandamaged: It's worth a shot, at least.  My main concern was that `Transaction` itself is a Promise-like object.  So, we need to make sure it isn't partially-resolved by accident.

lib/util/make-knex.js
```diff
@@ -36,17 +36,32 @@ function initContext(knexFn) {
...
+      // Backwards-compatibility: the default value of `config.doNotRejectOnRollback`
+      // changes depending upon whether or not a `container` was provided.
+      if(isUndefined(config.doNotRejectOnRollback)) {
+        config.doNotRejectOnRollback = !container;
+      }
 
-      if (container) {
+      if(container) {
+        const trx = this.client.transaction(container, config);
+        trx.userParams = this.userParams;
```
briandamaged: This seems a bit strange to me.  Is there a reason that `userParams` cannot be passed in via the `config` object?

lib/transaction.js
briandamaged: Actually: Just realized that this PR is showing items that were merged from the other PR (#3671 ) .  I might need to rebase the branch and create a new PR instead.

lib/util/make-knex.js
kibertoad: They probably could, `userParams` is just a slightly special beast that we use for handling e. g. enabling/disabling preprocessing of queries in addition to its primary purpose of parametrizing specific copy of knex builder :)

briandamaged:
Closing this PR out because there were duplicate commits in its history.

The rebased version of this PR can be found here:  #3674

PR #3674, opened 2020-02-19
Cleanup and extract Transaction Workflow logic (rebase)

briandamaged:
This extracts the "operational" logic related to container / non-container Transactions. Saying it another way: it means that the `Transaction` class no longer needs to be aware of "how" it was instantiated.

(Rebase of #3673 )

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -14,23 +14,11 @@ const { uniqueId, isUndefined } = require('lodash');
 // Acts as a facade for a Promise, keeping the internal state
 // and managing any child transactions.
 class Transaction extends EventEmitter {
-  constructor(client, container, config, outerTx) {
+  constructor(client, container, config = {}, outerTx = null) {
     super();
     this.doNotRejectOnRollback = config.doNotRejectOnRollback;
 
     const txid = (this.txid = uniqueId('trx'));
-
-    // If there is no container provided, assume user wants to get instance of transaction and use it directly
-    if (!container) {
-
-
-      this.initPromise = new Promise((resolve, reject) => {
-        this.initRejectFn = reject;
-        container = (transactor) => {
-          resolve(transactor);
-        };
-      });
-    }
 
     this.client = client;
     this.logger = client.logger;
@@ -94,12 +82,6 @@ class Transaction extends EventEmitter {
         });
 
       return executionPromise;
-    }).catch((err) => {
-      if (this.initRejectFn) {
-        this.initRejectFn(err);
-      } else {
-        throw err;
-      }
     });
 
 
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -16,17 +16,13 @@ const { uniqueId, isUndefined } = require('lodash');
 class Transaction extends EventEmitter {
   constructor(client, container, config, outerTx) {
     super();
+    this.doNotRejectOnRollback = config.doNotRejectOnRollback;
 
     const txid = (this.txid = uniqueId('trx'));
 
     // If there is no container provided, assume user wants to get instance of transaction and use it directly
     if (!container) {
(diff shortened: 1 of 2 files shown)


lib/transaction.js
```diff
@@ -14,35 +14,12 @@ const { uniqueId, isUndefined } = require('lodash');
 // Acts as a facade for a Promise, keeping the internal state
 // and managing any child transactions.
 class Transaction extends EventEmitter {
-  constructor(client, container, config, outerTx) {
+  constructor(client, container, config = {}, outerTx = null) {
     super();
+    this.doNotRejectOnRollback = config.doNotRejectOnRollback;
```
briandamaged: @kibertoad : Should we provide an ULTRA-default value here?  If so: what should it be?

lib/util/make-knex.js
```diff
@@ -36,17 +36,32 @@ function initContext(knexFn) {
...
+        const trx = this.client.transaction(container, config);
+        trx.userParams = this.userParams;
         return trx;
+      } else {
+        const initPromise = new Promise((resolve, reject) => {
+          _resolve = resolve;
+          _reject = reject;
+        });
+
+        const trx = this.client.transaction(_resolve, config);
```
maximelkin: maybe move this lines inside Promise constructor?

lib/transaction.js
kibertoad: I would still default to true, TBH. This whole parameter exists to preserve compatibility with old-style (callback) transactions that used to work the "FALSE" way, and personally I find this behaviour less helpful (see https://github.com/knex/knex/issues/3246)

lib/util/make-knex.js
briandamaged: Oh -- you're right!  I didn't even notice that the refactoring allowed for the code to be cleaned up in that way.  Neat!

(I'll push the change once the tests finish running)

maximelkin: also _resolve and _reject wasn't declared, but this will be fixed by that move

kibertoad:
Yeah, so build failed with
```
/home/travis/build/knex/knex/lib/util/make-knex.js
  55:11  error  '_resolve' is not defined  no-undef
  56:11  error  '_reject' is not defined   no-undef
  59:45  error  '_resolve' is not defined  no-undef
  61:19  error  '_reject' is not defined   no-undef
```

lib/util/make-knex.js
briandamaged: Oops -- looks like I deleted the `let` statement for those when I was rearranging the code earlier.  Anyhoo: the latest commit moves everything inside of the Promise constructor.

lib/transaction.js
briandamaged: k -- just pushed an update for that.

briandamaged:
@kibertoad : Yeah -- I deleted the `let` statements by accident when rearranging a few lines of code.  But, good news: @maximelkin had a great suggestion for a way to further cleanup that part of the code.  So, the issues should be resolved once Travis-CI finishes

lib/transaction.js
briandamaged: On a related note: perhaps the behavior can eventually be extracted out of the Transactor class via a Decorator or Proxy of some kind?

kibertoad: IIRC, Proxy are rather expensive performance-wise, so I would prefer to avoid them on a hot path. Simple decorator might make sense, though.

briandamaged:
Alright -- I recommend that we go ahead and merge this PR once we're comfortable with it.  There are other sections of the `Transaction` logic that can potentially be cleaned up now; however, refactoring those areas within the same PR will make it more challenging to code review everything.

Thoughts?

kibertoad:
I agree.

PR #3679, opened 2020-02-22
WIP: Cleanup/extract transaction ordering logic

briandamaged:
Experimental refactoring of the `Transaction` code.

Note: there's another PR that touches a lot of the same code; therefore, there is no intention of ever merging this branch.  But, the lessons learned in this branch will then be used in a separate PR later on.

### DIFF
--- a/lib/dialects/oracle/transaction.js
+++ b/lib/dialects/oracle/transaction.js
@@ -40,38 +40,4 @@ module.exports = class Oracle_Transaction extends Transaction {
       });
   }
 
-  acquireConnection(config, cb) {
-    const configConnection = config && config.connection;
-    return new Bluebird((resolve, reject) => {
-      try {
-        resolve(configConnection || this.client.acquireConnection());
-      } catch (e) {
-        reject(e);
-      }
-    })
-      .then((connection) => {
-        connection.__knexTxId = this.txid;
-
-        return connection;
-      })
-      .then((connection) => {
-        if (!this.outerTx) {
-          connection.setAutoCommit(false);
-        }
-        return connection;
-      })
-      .then(async (connection) => {
-        try {
-          return await cb(connection);
-        } finally {
-          debugTx('%s: releasing connection', this.txid);
-          connection.setAutoCommit(true);
-          if (!configConnection) {
-            this.client.releaseConnection(connection);
-          } else {
-            debugTx('%s: not releasing external connection', this.txid);
-          }
-        }
-      });
-  }
 };
--- a/lib/dialects/oracledb/transaction.js
+++ b/lib/dialects/oracledb/transaction.js
@@ -54,21 +54,12 @@ module.exports = class Oracle_Transaction extends Transaction {
   acquireConnection(config, cb) {
     const configConnection = config && config.connection;
     const t = this;
-    return new Bluebird((resolve, reject) => {
+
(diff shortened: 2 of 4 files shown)


lib/dialects/oracle/transaction.js
```diff
@@ -40,38 +40,4 @@ module.exports = class Oracle_Transaction extends Transaction {
...
-          debugTx('%s: releasing connection', this.txid);
-          connection.setAutoCommit(true);
-          if (!configConnection) {
-            this.client.releaseConnection(connection);
-          } else {
-            debugTx('%s: not releasing external connection', this.txid);
-          }
-        }
-      });
-  }
```
briandamaged: Dead Code.

briandamaged: As in: this was already dead code before this refactoring started.

lib/dialects/oracledb/transaction.js
```diff
@@ -54,21 +54,12 @@ module.exports = class Oracle_Transaction extends Transaction {
...
-          .catch(reject);
-      } catch (e) {
-        reject(e);
-      }
-    }).then(async (connection) => {
+
+    return this._acquireConnection(config).then(async (connection) => {
       try {
+        connection.__knexTxId = this.txid;
+        connection.isTransaction = true;
```
briandamaged: Various sections of the `Transaction` code are manipulating values on the `connection` object.  This is problematic for a number of reasons.

In this specific case, I suspect that Oracle connections are not being released properly when a Transaction is more than 2 levels deep.

lib/transaction.js
```diff
@@ -48,60 +50,12 @@ class Transaction extends EventEmitter {
...
-
-      return executionPromise;
-    });
-
+    const basePromise = this._kickThingsOff(config, container);
 
     // FYI: This is the Promise Chain for EXTERNAL use.  It ensures that the
     //      caller must handle any exceptions that result from `basePromise`.
-    this._promise = basePromise.then((x)=> x);
+    this._promise = Bluebird.resolve(basePromise).then((x)=> x);
```
briandamaged: Added `Bluebird.resolve(..)` as a temporary work-around since various portions of the code still rely upon the `.finally(..)` method.

lib/dialects/oracle/transaction.js
maximelkin: Why this is dead code?

briandamaged: There are a few reasons:

1. `"oracle"` is not listed within the `SUPPORTED_CLIENTS` constant:

https://github.com/knex/knex/blob/c9be84890cf7d3211c6103d7c914a60b4b84647b/lib/constants.js#L10-L20

2. The `lib/dialects/oracle` module is only referenced by the `lib/dialects/oracledb` module.
3. The `lib/dialects/oracledb` module sub-classes this code and then overrides the `acquireConnection(..)` method.

Additionally, tests continue to succeed even after this code has been removed.

kibertoad:
aaaaand conflicts are all over the place. as expected

briandamaged:
@kibertoad : Ah -- yeah, that's why this was an experimental/throw-away branch.  I just wanted to provide everybody w/ a preview of the refactoring.

briandamaged:
Closing this out in favor of #3698

PR #3681, opened 2020-02-23
Add support to transform raw SQL through user specified transformers

lorefnon:
Follows from the comments in https://github.com/knex/knex/pull/3649

### DIFF
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -36,6 +36,7 @@ function Builder(client) {
   this.client = client;
   this.and = this;
   this._single = {};
+  this._rawTransformers = [];
   this._statements = [];
   this._method = 'select';
   if (client.config) {
@@ -77,6 +78,14 @@ assign(Builder.prototype, {
     return this.client.queryCompiler(this).toSQL(method || this._method, tz);
   },
 
+  transformRaw(transformer) {
+    if (!isFunction(transformer)) {
+      throw new TypeError("Expected provided transformer to be a function");
+    }
+    this._rawTransformers.push(transformer);
+    return this;
+  },
+
   // Create a shallow clone of the current query builder.
   clone() {
     const cloned = new this.constructor(this.client);
@@ -84,6 +93,7 @@ assign(Builder.prototype, {
     cloned._single = clone(this._single);
     cloned._statements = clone(this._statements);
     cloned._debug = this._debug;
+    cloned._rawTransformers = clone(this._rawTransformers);
 
     // `_option` is assigned by the `Interface` mixin.
     if (!isUndefined(this._options)) {
--- a/lib/query/compiler.js
+++ b/lib/query/compiler.js
@@ -11,6 +11,7 @@ const {
   bind,
   compact,
   groupBy,
+  isNil,
   isEmpty,
   isString,
   isUndefined,
@@ -52,6 +53,7 @@ class QueryCompiler {
     this.formatter = client.formatter(builder);
     // Used when the insert call is empty.
     this._emptyInsertValue = 'default values';
+    this._rawTransformers = builder._rawTransformers;
     this.first = this.select;
   }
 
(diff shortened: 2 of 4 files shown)


elhigu:
Hmm.. looks like this thing is actually missing the part, where you can tell which part of SQL compilation this is applied like it was discussed here https://github.com/knex/knex/pull/3593#issuecomment-577842601 

So this would not give any extra information about compiler's statement part which is being modified. Though I kind of like simplicity of this one too. I'm not sure if having the hooks like that actually solves anything that cannot be resolved as easy by this one. 

Maybe if there are multiple subquery builders and only like main builder should have the transformation applied it would be useful 🤔

elhigu:
One thing that I would like to have working is:

```
const subquery = knex('table2').select('id');

knex('table')
  .whereIn('id', subquery)
  .transformRaw((query) => {
    query.sql = query.sql.replace('select', 'SELECT');
  });
```

Which would modify in this case only the actual query builder SQL and not the subquery. Which I suppose means that the transformations should be ran before interpolating all SQL to single long query.

kibertoad:
@lorefnon Would you like this included in the next knex release? Shall we wait for this PR?

elhigu:
I don't think there is any reason for holding release for this. Better to make feature carefully and make another release if there is hurry for anyone to get this into use.

lorefnon:
Sorry for stalling on this. Got occupied with some other things. 

Regarding https://github.com/knex/knex/pull/3681#issuecomment-590738101 : 

IMHO this can't be done in a clean way, without effectively rewriting how the QueryCompiler works across dialects. Currently the sql fragments are pieced together as strings sequentially, so there isn't a point where we have some outer sql template with placeholders into which the inner sql gets injected. 

Eg. when we are generating the sql for a select, but the time the "where" component gets processed we don't have the complete outer sql yet but we simply concat the complete inner sql and continue processing the subsequent components of the outer select. 

So if we want to implement something that distinguishes between outer query and sub-query etc. the most logical way seems that instead of intermediate steps of compiler generating partial strings, we generate a tree of sql fragments from each step and then the final toSQL is the only one responsible for stringifying the tree as the last step of query generation. 

> Maybe if there are multiple subquery builders and only like main builder should have the transformation applied it would be useful 🤔

I think what you were suggesting here is kind'a similar but not too sure. Basically in the above, the string generation will happen in the top level QueryCompiler and inner query compilers will only be generating the fragment trees which become a part of the outer tree. We can keep a reference to the QueryBuilder which each tree node is coming from.

I am explicitly using the term fragment tree and not a syntax tree because we wouldn't parse the raw strings at any point. So things passed as strings to knex will remain strings, just the hierarchy knex generates will have a structured representation that will be exposed to the transformer.

I *think* it is possible to implement this in a way that doesn't break third party dialects as long as we are fine with the transformers not working for them either.

elhigu:
I believe you might be right about it. I'll give this a look and comment some more after that.

elhigu:
Yea, looks like the way how the whole query compiler works would need to be changed 🙄Since this cannot be done well easily, I suppose I'm now fine with anything you'll come up with 😅Thanks for putting so much effort on this.

lorefnon:
Thanks :) I'll take a stab at implementing this in non-breaking way and see how it goes.

OlivierCavadenti:
@lorefnon can I take it ? I plan to finish this  if possible.

OlivierCavadenti:
> @OlivierCavadenti Sure, please do.
> 
> Sorry for having to discontinue this due to a change in personal priorities at my end.

I will read all comments, but did you explore any specific solutions ?

OlivierCavadenti:
I will work on this next month and I am quite interested about :

`The protoc code generator does something similar by adding comment markers with [insertion points](https://developers.google.com/protocol-buffers/docs/reference/java-generated#plugins) into the generated code which plugins can target. They are kind'a clunky to deal with but may solve use cases for some people.`

Could be interesting to implements a insert point feature in query compilers but I need to figure how it's can impact the basecode.
The main idea is to insert transformers in steps of query compiler of all dialects, but we need to be sure that is not a re-implementation of raw feature and we don't put intruse code in all the code... 

`For my own use cases, I mostly construct complex queries through liquid templates now with a bunch of custom helpers that reduce the scope of errors. I have found that to be easier from maintenability perspective.`

Can you explain me more how you do this (with an example) ?

PR #3682, opened 2020-02-23
remove dependency on bluebird methods from tests

maximelkin:
check https://github.com/knex/knex/pull/3650 for older discussions

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -4,17 +4,10 @@
 require('source-map-support').install();
 
 global.sinon = require('sinon');
+require('./chai-setup');
+global.chai.use(require('sinon-chai'));
 
-const chai = (global.chai = require('chai'));
-
-chai.use(require('sinon-chai'));
-chai.should();
-
-const bluebird = require('bluebird');
-global.expect = chai.expect;
 global.d = new Date();
-
-bluebird.longStackTraces();
 
 // '.timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection' produced unhandled rejection and it's unclear how to avoid that
 const EXPECTED_REJECTION_COUNT = 2;
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -4,7 +4,7 @@
 
 const Knex = require('../../../knex');
 const _ = require('lodash');
-const bluebird = require('bluebird');
+const delay = require('../../../lib/util/delay');
 
 module.exports = function(knex) {
   describe('Additional', function() {
@@ -190,44 +190,6 @@ module.exports = function(knex) {
             ]);
           });
       });
-    });
-
-    it('should forward the .mapSeries() function from bluebird', function() {
-      const asyncTask = function() {
-        return new Promise(function(resolve, reject) {
-          const output = asyncTask.num++;
-          setTimeout(function() {
-            resolve(output);
-          }, Math.random() * 200);
-        });
-      };
-      asyncTask.num = 1;
-
-      const returnedValues = [];
-      return knex('accounts')
-        .select()
-        .limit(3)
-        .mapSeries(function(account) {
-          return asyncTask().then(function(number) {
-            returnedValues.push(number);
-          });
-        })
-        .then(function() {
-          expect(returnedValues[0] == 1);
-          expect(returnedValues[1] == 2);
-          expect(returnedValues[2] == 3);
-        });
-    });
-
-    it('should forward the .delay() function from bluebird', function() {
-      const startTime = new Date().valueOf();
-      return knex('accounts')
-        .select()
-        .limit(1)
-        .delay(300)
-        .then(function(accounts) {
-          expect(new Date().valueOf() - startTime > 300);
-        });
     });
 
     it('should truncate a table with truncate', function() {
@@ -835,10 +797,7 @@ module.exports = function(knex) {
           // This query will hang if a connection gets released back to the pool
           // too early.
           // 50ms delay since killing query doesn't seem to have immediate effect to the process listing
-          return bluebird
-            .resolve()
-            .then()
-            .delay(50)
+          return delay(50)
             .then(function() {
               return getProcessesQuery;
             })
@@ -860,7 +819,7 @@ module.exports = function(knex) {
         });
     });
 
-    it('.timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection', function() {
+    it('.timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection', async () => {
       // Only mysql/postgres query cancelling supported for now
       const driverName = knex.client.driverName;
       if (
@@ -903,23 +862,21 @@ module.exports = function(knex) {
 
       const query = testQueries[driverName]();
 
-      return query
-        .timeout(1, { cancel: true })
-        .then(function() {
-          throw new Error("Shouldn't have gotten here.");
-        })
-        .catch(function(error) {
-          expect(_.pick(error, 'timeout', 'name', 'message')).to.deep.equal({
-            timeout: 1,
-            name: 'KnexTimeoutError',
-            message:
-              'After query timeout of 1ms exceeded, cancelling of query failed.',
-          });
-        })
-        .finally(() => knexDb.destroy());
+      try {
+        await expect(
+          query.timeout(1, { cancel: true })
+        ).to.eventually.be.rejected.and.to.deep.include({
+          timeout: 1,
+          name: 'KnexTimeoutError',
+          message:
+            'After query timeout of 1ms exceeded, cancelling of query failed.',
+        });
+      } finally {
+        await knexDb.destroy();
+      }
     });
 
-    it('.timeout(ms, {cancel: true}) should release connections after failing if connection cancellation throws an error', function() {
+    it('.timeout(ms, {cancel: true}) should release connections after failing if connection cancellation throws an error', async function() {
       // Only mysql/postgres query cancelling supported for now
       const driverName = knex.client.driverName;
       if (!_.startsWith(driverName, 'pg')) {
@@ -951,6 +908,7 @@ module.exports = function(knex) {
       const knexPrototype = Object.getPrototypeOf(knexDb.client);
       const originalWrappedCancelQueryCall =
         knexPrototype._wrappedCancelQueryCall;
+
       knexPrototype._wrappedCancelQueryCall = (conn) => {
         return knexPrototype.query(conn, {
           method: 'raw',
@@ -961,32 +919,27 @@ module.exports = function(knex) {
       const queryTimeout = 10;
       const secondQueryTimeout = 11;
 
-      return getTestQuery()
-        .timeout(queryTimeout, { cancel: true })
-        .then(function() {
-          throw new Error("Shouldn't have gotten here.");
-        })
-        .catch(function(error) {
-          expect(_.pick(error, 'timeout', 'name', 'message')).to.deep.equal({
-            timeout: queryTimeout,
-            name: 'error',
-            message: `After query timeout of ${queryTimeout}ms exceeded, cancelling of query failed.`,
-          });
-        })
-        .then(() => {
-          knexPrototype._wrappedCancelQueryCall = originalWrappedCancelQueryCall;
-          return getTestQuery().timeout(secondQueryTimeout, { cancel: true });
-        })
-        .catch(function(error) {
-          expect(_.pick(error, 'timeout', 'name', 'message')).to.deep.equal({
-            timeout: secondQueryTimeout,
-            name: 'KnexTimeoutError',
-            message: `Defined query timeout of ${secondQueryTimeout}ms exceeded when running query.`,
-          });
-        })
-        .finally(() => {
-          return knexDb.destroy();
+      try {
+        await expect(
+          getTestQuery().timeout(queryTimeout, { cancel: true })
+        ).to.be.eventually.rejected.and.deep.include({
+          timeout: queryTimeout,
+          name: 'error',
+          message: `After query timeout of ${queryTimeout}ms exceeded, cancelling of query failed.`,
         });
+
+        knexPrototype._wrappedCancelQueryCall = originalWrappedCancelQueryCall;
+
+        await expect(
+          getTestQuery().timeout(secondQueryTimeout, { cancel: true })
+        ).to.be.eventually.rejected.and.deep.include({
+          timeout: secondQueryTimeout,
+          name: 'KnexTimeoutError',
+          message: `Defined query timeout of ${secondQueryTimeout}ms exceeded when running query.`,
+        });
+      } finally {
+        await knexDb.destroy();
+      }
     });
 
     it('Event: query-response', function() {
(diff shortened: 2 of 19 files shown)


briandamaged:
@maximelkin : Thx for splitting this into a separate PR!  I'll have some time to look over it later this afternoon.

test/index.js
```diff
@@ -4,18 +4,11 @@
 require('source-map-support').install();
 
 global.sinon = require('sinon');
+require('./chai-setup');
+global.chai.use(require('sinon-chai'));
```
briandamaged: Perhaps this should be moved into `test/chai-setup.js` as well?

test/index.js
```diff
@@ -4,18 +4,11 @@
...
 
-const chai = (global.chai = require('chai'));
-
-chai.use(require('sinon-chai'));
-chai.should();
-
-const bluebird = require('bluebird');
-global.expect = chai.expect;
 global.d = new Date();
```
briandamaged: Hmm... having a global variable named `d` seems a bit dangerous since `d` is commonly used as a temporary variable name.  This could lead to some surprises when writing test cases.

(Fixing this might be out of scope for the current PR, though.  But, if we can fix this in 1-2 lines, then it might be worth it).

test/index.js
```diff
@@ -4,18 +4,11 @@
...
 
-bluebird.longStackTraces();
-
```
briandamaged: One concern: `bluebird.longStackTraces()` has actually been very helpful in tracking down race conditions in the code!  We might want to figure out a way to optionally enable this feature just for testing purposes.

(The challenge, of course, is that it then also becomes difficult to verify that the code is working properly _without_ relying upon `Bluebird`.  It's like it needs to be a test dependency, but also _not_ be a test dependency at the same time)

cc: @kibertoad

test/integration/builder/additional.js
```diff
@@ -951,6 +908,7 @@ module.exports = function(knex) {
       const knexPrototype = Object.getPrototypeOf(knexDb.client);
       const originalWrappedCancelQueryCall =
         knexPrototype._wrappedCancelQueryCall;
+
       knexPrototype._wrappedCancelQueryCall = (conn) => {
```
briandamaged: Hmm... we should probably be using a `sinon` sandbox here rather than patching the prototype ourselves.  This will help guarantee that the prototype is restored properly once the test case concludes.  (Although this is out of scope for this PR, it might be worth implementing this fix if it's easy to do)

cc: @kibertoad

test/integration/builder/additional.js
```diff
@@ -961,32 +919,27 @@ module.exports = function(knex) {
...
+      try {
+        await expect(
+          getTestQuery().timeout(queryTimeout, { cancel: true })
+        ).to.be.eventually.rejected.and.deep.include({
+          timeout: queryTimeout,
+          name: 'error',
+          message: `After query timeout of ${queryTimeout}ms exceeded, cancelling of query failed.`,
         });
+
+        knexPrototype._wrappedCancelQueryCall = originalWrappedCancelQueryCall;
```
briandamaged: This relates to my "use `sinon` sandboxing" comment from a moment ago.

Notice: if the previous `expect(..)` statement fails, this this line will never be executed.  This means that the prototype will never be restored, which will lead to some really confusing errors in later test cases.

test/integration/builder/additional.js
```diff
@@ -961,32 +919,27 @@ module.exports = function(knex) {
...
-          expect(_.pick(error, 'timeout', 'name', 'message')).to.deep.equal({
-            timeout: queryTimeout,
-            name: 'error',
-            message: `After query timeout of ${queryTimeout}ms exceeded, cancelling of query failed.`,
-          });
-        })
-        .then(() => {
-          knexPrototype._wrappedCancelQueryCall = originalWrappedCancelQueryCall;
-          return getTestQuery().timeout(secondQueryTimeout, { cancel: true });
-        })
```
briandamaged: Interestingly, it looks like the original test case was missing one of these:

```
        .then(function() {
          throw new Error("Shouldn't have gotten here.");
        })
```

Which means: the original test case was potentially overlooking some failures.

test/integration/builder/additional.js
```diff
@@ -961,32 +919,27 @@ module.exports = function(knex) {
...
+
+        await expect(
+          getTestQuery().timeout(secondQueryTimeout, { cancel: true })
+        ).to.be.eventually.rejected.and.deep.include({
+          timeout: secondQueryTimeout,
+          name: 'KnexTimeoutError',
+          message: `Defined query timeout of ${secondQueryTimeout}ms exceeded when running query.`,
+        });
+      } finally {
+        await knexDb.destroy();
```
briandamaged: It looks like there are several lines between:

1. Instantiating the `knexDb` instance, and
2. The beginning of the `try{} finally{}` code

Which means: if a failure occurs anywhere between these steps, then `knexDb` will never be destroyed.  So, we should probably begin the `try{}` statement a bit earlier.  Ex:

```
const knexDb = new Knex(knexConfig);

// Notice: the `try{}` begins immediately after instantiating `knexDb`
try {
  // Notice: this line used to be outside of the `try{}`
  const rawTestQueries = {
      pg: (sleepSeconds) => `SELECT pg_sleep(${sleepSeconds})`,
  };

  // ... more lines ...
} finally {
  await knexDb.destroy();
}
```

test/integration/builder/inserts.js
```diff
@@ -1148,39 +1147,37 @@ module.exports = function(knex) {
...
+          })
+          .then(function() {
+            expect.fail('Should not reach this point');
+          })
+          .catch(function(error) {
+            //Should reach this point before timeout of 10s
+            expect(error.message.toLowerCase()).to.include(
+              'batchinsertduplicatekey'
+            );
+          });
```
briandamaged: Most of this diff was due to indentation.  Short story: code is establishing the timeout via `chai` instead of `Bluebird` now.

test/integration/builder/transaction.js
```diff
@@ -290,13 +290,14 @@ module.exports = function(knex) {
             expect(count).to.equal(5);
             return knex('test_schema_migrations').count('*');
           })
+          .then(() => expect.fail('should never reach this line'))
```
briandamaged: Nice fix!

kibertoad:
@briandamaged I agree with your observations regarding tech debt accrued in our tests, we should create a JIRA issue summing up the scope of needed changes.

test/integration/builder/transaction.js
```diff
@@ -489,14 +493,12 @@ module.exports = function(knex) {
...
-        .then(function() {
-          throw new Error('should not get here');
-        })
-        .catch(KnexTimeoutError, function(error) {});
+      try {
+        await db.transaction(() => db.transaction(() => ({})));
+        expect.fail('should not get here');
+      } catch (error) {
+        expect(error).to.be.an.instanceof(KnexTimeoutError);
+      }
```
briandamaged: FYI: You can probably also do:

```
await expect(
  db.transaction(() => db.transaction(() => ({})))
).to.be.rejectedWith(KnexTimeoutError)
```

test/integration/builder/transaction.js
```diff
@@ -397,8 +401,8 @@ module.exports = function(knex) {
       if (/mssql/i.test(knex.client.driverName)) {
         return Promise.resolve();
       }
-      const first = Bluebird.delay(50);
-      const second = first.then(() => Bluebird.delay(50));
+      const first = delay(50);
+      const second = first.then(() => delay(50));
```
briandamaged: Your fix here looks good, but this is another test case we should revisit later on.

Looking at the code, I'm not convinced that this test case is sufficient / reliable.  (Especially since we recently discovered that some `Transaction` subclasses were not handling this case properly).  So, we'll need to take another look at this once we finish w/ the Bluebird removal.

test/index.js
maximelkin: Nope, because this related to sinon too

test/index.js
maximelkin: note: usages can be found via search /* global.?d */

test/index.js
maximelkin: Not affects tests directly, so we can just return it back and remove in second pr

test/integration/builder/additional.js
maximelkin: As i can see, lines between try block and knex instantiation dont throws, except if
I move new Knex after if

test/integration/builder/transaction.js
maximelkin: done

test/integration/builder/additional.js
briandamaged: Given that this is a test case, it's not really a huge concern.  But, generally speaking, it's best to establish "object disposal guarantees" immediately after instantiating an object.  For example, consider the following scenario:

```
const connection = await createConnection();

// But... we haven't established the connection disposal guarantee yet!
doSomethingwith(connection);

// At last, we're setting up the connection disposal guarantee
try {
  // Do stuff...
} finally {
  await connection.dispose();
}
```

Even if `doSomethingwith(connection)` is safe today, it might start throwing exceptions in a future version of the code.  This will then introduce a connection leak into the code, which might prove to be very difficult to find.

test/tape/crossdb-compatibility.js
```diff
@@ -12,97 +13,99 @@ module.exports = function(knex) {
...
+        ]);
+
+        const res = await knex('test_table');
+        t.assert(
+          res.length === 5,
+          'multiple rows with nulls could be added despite of unique constraints'
+        );
+      } finally {
+        t.end();
+      }
```
briandamaged: The logic appears to be equivalent.

(Side note: this really seems like it should be at least 3 separate test cases.  Given the current design of the test case, the test runner will only document assertions that succeeded.  If an assertion fails, the test runner will just report something like:  "Expected 3 assertions  (Only detected 2)" )

test/tape/crossdb-compatibility.js
```diff
@@ -12,97 +13,99 @@ module.exports = function(knex) {
...
+            t.unique('third');
+            t.unique('fourth');
+          });
+        t.assert(
+          true,
+          'Creating / dropping / creating unique constraint was a success'
+        );
+      } finally {
+        t.end();
+      }
```
briandamaged: Looks good!  (Another  `.finally(..)` -> `finally{}` replacement)

test/unit/knex.js
```diff
@@ -285,11 +284,10 @@ describe('knex', () => {
 
     const knexWithParams = knex.withUserParams({ userParam: '451' });
 
-    return knexWithParams.transaction((trx) => {
+    return knexWithParams.transaction(async (trx) => {
       expect(trx.userParams).to.deep.equal({
         userParam: '451',
       });
-      return bluebird.resolve();
```
briandamaged: I see.  So, this was to work around the `Transaction` class's special handling for callbacks that return Promises.  (Otherwise, the `Transaction` class would assume that the caller is responsible for committing / rolling back the transaction)

test/tape/pool.js
```diff
@@ -30,51 +30,37 @@ test(`pool evicts dead resources when factory.validate rejects`, (t) => {
...
-    });
+      )
+    );
+
+  const connectionsGroup1 = await allocateConnections();
+  connectionsGroup1.forEach((con) => pool.release(con));
+  connectionsGroup1.forEach((con) => {
+    // fake kill connections
+    con.error = 'connection lost';
+  });
```
briandamaged: It's not important, but it looks like these `.forEach(..)` statements could be combined.

test/tape/pool.js
```diff
@@ -30,51 +30,37 @@ test(`pool evicts dead resources when factory.validate rejects`, (t) => {
...
+  connectionsGroup2.forEach((con) => pool.release(con));
+
+  const connectionsGroup3 = await allocateConnections();
+  connectionsGroup3.forEach((con) =>
+    t.ok(con.id > 4 && con.id < 11, 'Released working connection was used')
+  );
+  connectionsGroup3.forEach((con) => pool.release(con));
+
+  await pool.destroy();
+  t.end();
```
briandamaged: The logic here appears to be equivalent to the logic it is replacing.

test/tape/pool.js
```diff
@@ -101,39 +87,36 @@ test('#823, should not skip pool construction pool config is not defined', funct
...
+        )
+      );
+    } catch (e) {
+      t.fail(
+        `Should have created new connection and execute the query, got : ${e}`
+      );
+    } finally {
+      t.end();
+      await knex.destroy();
+    }
```
briandamaged: The new logic looks like it's equivalent to the old logic.

test/tape/transactions.js
```diff
@@ -318,27 +308,24 @@ module.exports = function(knex) {
...
-          'Second sibling transaction completed'
-        );
       });
+    } finally {
+      t.equal(
+        secondTransactionCompleted,
+        true,
+        'Second sibling transaction completed'
+      );
+    }
```
briandamaged: `.finally(..)` -> `finally{}`

test/tape/transactions.js
```diff
@@ -352,18 +339,17 @@ module.exports = function(knex) {
...
-      })
-      .finally(function() {
-        return knex('test_table').then(function(results) {
-          t.equal(results.length, 2, 'Parent transaction inserted two records');
-        });
       });
+    } finally {
+      const results = await knex('test_table');
+      t.equal(results.length, 2, 'Parent transaction inserted two records');
+    }
```
briandamaged: `.finally(..)` -> `finally{}`

test/tape/transactions.js
```diff
@@ -380,19 +366,19 @@ module.exports = function(knex) {
...
-          'Second sibling transaction completed'
-        );
       });
+    } finally {
+      t.equal(
+        secondTransactionCompleted,
+        true,
+        'Second sibling transaction completed'
+      );
+    }
```
briandamaged: `.finally(..)` -> `finally{}`

test/tape/transactions.js
```diff
@@ -465,20 +449,19 @@ module.exports = function(knex) {
...
+        true,
+        'Second sibling transaction completed'
+      );
+      const results = await knex('test_table');
+      t.equal(results.length, 1, 'Parent transaction inserted one record');
+    }
   });
 
-  test('#625 - streams/transactions', 'postgresql', function(t) {
+  test('#625 - streams/transactions', 'postgresql', (t) => {
```
briandamaged: Just curious: was there a reason that this needed to be changed into an arrow function?

test/index.js
briandamaged: It's only "related to sinon" in the sense that it delegates to the sinon library under the hood.  However, it does not actually change the behavior of sinon itself in any way.  (It's a plugin that adds a handful of new functions to chai)

But yeah -- it's not really important for this PR anyway, so we can skip this change for now.

test/tape/transactions.js
kibertoad: In general we are replacing anonymous functions with arrow functions, so there didn't have to be need.

kibertoad:
@briandamaged Yeah, it looks good to me, and you did very through analysis on it already, so I think we are good to go. I created separate issue to track test refactorings.

kibertoad:
@briandamaged Will you do the honours?

briandamaged:
@kibertoad : Done!

PR #3683, opened 2020-02-23
remove dependency on bluebird methods from sources

maximelkin:
see https://github.com/knex/knex/pull/3650
wait for https://github.com/knex/knex/pull/3682

### DIFF
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -1,4 +1,3 @@
-const Bluebird = require('bluebird');
 const Transaction = require('../../transaction');
 const { isUndefined } = require('lodash');
 const debug = require('debug')('knex:tx');
@@ -9,11 +8,9 @@ module.exports = class Transaction_MSSQL extends Transaction {
     return conn.tx_.begin().then(this._resolver, this._rejecter);
   }
 
-  savepoint(conn) {
+  async savepoint(conn) {
     debug('%s: savepoint at', this.txid);
-    return Bluebird.resolve().then(() =>
-      this.query(conn, `SAVE TRANSACTION ${this.txid}`)
-    );
+    return this.query(conn, `SAVE TRANSACTION ${this.txid}`);
   }
 
   commit(conn, value) {
@@ -48,13 +45,11 @@ module.exports = class Transaction_MSSQL extends Transaction {
     );
   }
 
-  rollbackTo(conn, error) {
+  async rollbackTo(conn, error) {
     debug('%s: rolling backTo', this.txid);
-    return Bluebird.resolve()
-      .then(() =>
-        this.query(conn, `ROLLBACK TRANSACTION ${this.txid}`, 2, error)
-      )
-      .then(() => this._rejecter(error));
+    await this.query(conn, `ROLLBACK TRANSACTION ${this.txid}`, 2, error);
+
+    this._rejecter(error);
   }
 
   // Acquire a connection and create a disposer - either using the one passed
@@ -62,7 +57,7 @@ module.exports = class Transaction_MSSQL extends Transaction {
   // the original promise is marked completed.
   acquireConnection(config, cb) {
     const configConnection = config && config.connection;
-    return new Bluebird((resolve, reject) => {
+    return new Promise((resolve, reject) => {
       try {
         resolve(
           (this.outerTx ? this.outerTx.conn : null) ||
--- a/lib/dialects/oracle/index.js
+++ b/lib/dialects/oracle/index.js
@@ -5,7 +5,6 @@ const { promisify } = require('util');
 
 const inherits = require('inherits');
 const Client = require('../../client');
-const Bluebird = require('bluebird');
 const { bufferToString } = require('../../query/string');
 const Formatter = require('./formatter');
 
@@ -80,10 +79,10 @@ Object.assign(Client_Oracle.prototype, {
   // Get a raw connection, called by the `pool` whenever a new
   // connection needs to be added to the pool.
   acquireRawConnection() {
-    return new Bluebird((resolver, rejecter) => {
+    return new Promise((resolver, rejecter) => {
       this.driver.connect(this.connectionSettings, (err, connection) => {
         if (err) return rejecter(err);
-        Bluebird.promisifyAll(connection);
+        connection.executeAsync = promisify(connection.execute);
         if (this.connectionSettings.prefetchRowCount) {
           connection.setPrefetchRowCount(
             this.connectionSettings.prefetchRowCount
@@ -116,7 +115,7 @@ Object.assign(Client_Oracle.prototype, {
   },
 
   _stream(connection, obj, stream, options) {
-    return new Bluebird(function(resolver, rejecter) {
+    return new Promise(function(resolver, rejecter) {
       stream.on('error', (err) => {
         if (isConnectionError(err)) {
           connection.__knex__disposed = err;
--- a/lib/dialects/oracledb/index.js
+++ b/lib/dialects/oracledb/index.js
@@ -5,7 +5,6 @@ const inherits = require('inherits');
 const QueryCompiler = require('./query/compiler');
 const ColumnCompiler = require('./schema/columncompiler');
 const { BlobHelper, ReturningHelper, isConnectionError } = require('./utils');
-const Bluebird = require('bluebird');
 const stream = require('stream');
 const { promisify } = require('util');
 const Transaction = require('./transaction');
@@ -82,7 +81,7 @@ Client_Oracledb.prototype.prepBindings = function(bindings) {
 // connection needs to be added to the pool.
 Client_Oracledb.prototype.acquireRawConnection = function() {
   const client = this;
-  const asyncConnection = new Bluebird(function(resolver, rejecter) {
+  const asyncConnection = new Promise(function(resolver, rejecter) {
     // If external authentication don't have to worry about username/password and
     // if not need to set the username and password
     const oracleDbConfig = client.connectionSettings.externalAuth
@@ -112,7 +111,7 @@ Client_Oracledb.prototype.acquireRawConnection = function() {
         return rejecter(err);
       }
       connection.commitAsync = function() {
-        return new Bluebird((commitResolve, commitReject) => {
+        return new Promise((commitResolve, commitReject) => {
           this.commit(function(err) {
             if (err) {
               return commitReject(err);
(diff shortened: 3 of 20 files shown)


kibertoad:
@maximelkin There's a minor conflict now.

lib/dialects/mssql/transaction.js
```diff
@@ -48,21 +45,19 @@ module.exports = class Transaction_MSSQL extends Transaction {
...
+  async rollbackTo(conn, error) {
     debug('%s: rolling backTo', this.txid);
-    return Bluebird.resolve()
-      .then(() =>
-        this.query(conn, `ROLLBACK TRANSACTION ${this.txid}`, 2, error)
-      )
-      .then(() => this._rejecter(error));
+    await this.query(conn, `ROLLBACK TRANSACTION ${this.txid}`, 2, error);
+
+    this._rejecter(error);
```
briandamaged: This appears to replicate the old logic...

... BUT: It's a bit surprising that `_this.rejector(error)` only gets invoked when `this.query(..)` succeeds.  We should investigate this further once we've merged this PR.

briandamaged: Yeah... this behavior seems to be inconsistent across `Transaction` implementations.  So, we can follow up on this once we resume the `Transaction` refactoring.

lib/dialects/postgres/index.js
```diff
@@ -273,22 +272,20 @@ Object.assign(Client_PG.prototype, {
...
-    });
+    const conn = await this.acquireConnection();
+
+    try {
+      return await this._wrappedCancelQueryCall(conn, connectionToKill);
+    } finally {
+      // NOT returning this promise because we want to release the connection
+      // in a non-blocking fashion
+      this.releaseConnection(conn);
+    }
```
briandamaged: Logic appears to be equivalent to the logic that it replaced.

(We should probably consider adding an `await` to the `this.releaseConnection(..)` call in a future PR, though.  Well, more specifically: figure out a good way to handle unhandled promise rejections from that call)

maximelkin:
@kibertoad done

lib/transaction.js
```diff
@@ -234,6 +235,23 @@ class Transaction extends EventEmitter {
...
+  }
+
+  finally(onFinally) {
+    return this._promise.finally(onFinally);
+  }
+
+  asCallback(cb) {
+    callbackify(() => this._promise)(cb);
+    return this._promise;
+  }
```
briandamaged: Hmm... it looks like `.asCallback(..)` is mostly a Bluebird method.  Unfortunately, we mentioned it in the Knex documentation, and therefore we (probably) need to continue supporting it.

However, it looks like Bluebird supports an additional `options` parameter:

https://github.com/petkaantonov/bluebird/blob/master/src/nodeify.js#L45

So, even if we continue providing an `.asCallback(..)` method, there's still a good chance that we're gonna break backwards compatibility by introducing this change.

@kibertoad + @maximelkin : Thoughts?

lib/transaction.js
```diff
@@ -234,6 +235,23 @@ class Transaction extends EventEmitter {
...
+    return this._promise.then(onResolve, onReject);
+  }
+
+  catch(onReject) {
+    return this._promise.catch(onReject);
+  }
+
+  finally(onFinally) {
+    return this._promise.finally(onFinally);
+  }
```
briandamaged: Might want to do this instead:

```
async finally(onFinally) {
  try {
    return await this._promise
  } finally {
    await onFinally();
  }
}
```

This way, we can provide a `.finally(..)` method without having to rely upon the underlying Promise implementation.

(Note: we should probably also verify that `onFinally` is actually a function before invoking it.)

lib/transaction.js
```diff
@@ -215,11 +214,13 @@ class Transaction extends EventEmitter {
...
-          .then(function() {
-            return connection;
-          });
+        return (
+          this._previousSibling
+            // .catch(() => {}) // TODO: Investigate this line
+            .then(function() {
+              return connection;
+            })
+        );
```
briandamaged: Added parentheses to make multi-line statement more apparent.

lib/dialects/mssql/transaction.js
maximelkin: we can create issue where track full review of code

lib/transaction.js
maximelkin: We not documented options neither on site nor typings
as solution we can assert second parameter is undefined

kibertoad: Then we can freely consider options entity not a part of public api and remove it without semver major.

lib/transaction.js
maximelkin: Providing working finally when native promises not supports it can confuse user
Also this can introduce unexpected errors:
selectFromDb()
+    .then(x => map(x))
    .finally(() => {})

start throwing

lib/transaction.js
maximelkin: I vote for bump semver major (second number of course), just to warn users

kibertoad: It's always a hassle, because then downstream (such as bookshelf) need to bump their versions as well, and in this case it feels hardly warranted.
I would reserve (semi-)major version bump for when we drop Node 8 support.

lib/transaction.js
briandamaged: Very true.  In that case, we might want to inject this function into `Transaction.prototype` conditionally.  Otherwise, end-users might get confused when their `Transaction` instance _claims_ to have a `.finally(..)` method, but the method always throws an exception.

(Unfortunately, this might also complicate things on the Typescript side...)

lib/transaction.js
briandamaged: My guess is that this release is prolly gonna surprise end-users one way or another.  There's a good chance that their code probably relied upon `bluebird` methods even though `knex` didn't explicitly advertise support for them.

lib/interface.js
```diff
@@ -81,31 +81,17 @@ module.exports = function(Target) {
...
+    return promise;
+  };
+
+  Target.prototype.catch = function(onReject) {
+    return this.then().catch(onReject);
+  };
+
+  Target.prototype.finally = function(onFinally) {
+    return this.then().finally(onFinally);
+  };
```
briandamaged: @maximelkin : Based upon our earlier discussion, I think this one should probably be conditionally injected as well.

lib/transaction.js
maximelkin: Typescript typings not affected, because querybuilder extends part of globally defined Promise interface, so finally will be available only if finally enabled

maximelkin: >Otherwise, end-users might get confused when their Transaction instance claims to have a
I think user won't do ('finally' in trx), but ok, I will fix that case

briandamaged: Yeah -- just saw the Typescript part of the code.  I think you're correct, but I'm gonna triple-check it via a Typescript compiler just to be sure.

But, my other concern still stands: anybody using reflection to see the list of methods will see `.finally(..)` listed.  Example from my `node 8` console:

```
> var Transaction = require('./lib/transaction')
undefined
> "finally" in Transaction.prototype
true
```

briandamaged: Sounz good.  On the plus side: we can remove the conditional logic once `node 8` support is dropped 😅

maximelkin: done

lib/util/finally-mixin.js
```diff
@@ -0,0 +1,10 @@
+const noop = require('./noop');
+
+const finallyMixin = (prototype) =>
+  Object.assign(prototype, {
+    finally(onFinally) {
+      return this.then().finally(onFinally);
+    },
+  });
+
+module.exports = Promise.prototype.finally ? noop : finallyMixin;
```
briandamaged: We should probably leave a comment here to explain why this logic is conditional.

```suggestion
// FYI: Support for `Promise.prototype.finally` was not introduced until Node 9.
//        Therefore, Knex will need to conditionally inject support for `.finally(..)`
//        until support for Node 8 is officially dropped.
module.exports = Promise.prototype.finally ? noop : finallyMixin;
```

lib/interface.js
```diff
@@ -81,31 +82,15 @@ module.exports = function(Target) {
...
+    const promise = this.then();
+    callbackify(() => promise)(cb);
+    return promise;
+  };
+
+  Target.prototype.catch = function(onReject) {
+    return this.then().catch(onReject);
+  };
+
+  finallyMixin(Target.prototype);
```
briandamaged: Ah -- good idea w/ implementing `.finally(..)` support as a Mixin!

lib/dialects/oracle/index.js
```diff
@@ -80,10 +79,10 @@ Object.assign(Client_Oracle.prototype, {
   // Get a raw connection, called by the `pool` whenever a new
   // connection needs to be added to the pool.
   acquireRawConnection() {
-    return new Bluebird((resolver, rejecter) => {
+    return new Promise((resolver, rejecter) => {
       this.driver.connect(this.connectionSettings, (err, connection) => {
         if (err) return rejecter(err);
-        Bluebird.promisifyAll(connection);
+        connection.executeAsync = promisify(connection.execute);
```
briandamaged: This seems like it is probably reasonable.  I'll have a better idea once I finish reviewing `dialects/oracledb`

lib/dialects/oracledb/index.js
```diff
@@ -259,100 +258,100 @@ Client_Oracledb.prototype._query = function(connection, obj) {
...
-    if (obj.returningSql) {
-      const response = await connection.executeAsync(
-        obj.returningSql(),
-        rowIds,
-        { resultSet: true }
-      );
-      obj.response = response.rows;
-    }
-    return obj;
-  });
```
briandamaged: This is literally all just a gigantic re-indentation.  (I confirmed this by copying the content from both branches, normalizing the indentation, and comparing the diffs.  End-result: the diffs were identical)

lib/transaction.js
briandamaged: Looks good -- thx!

lib/dialects/oracle/index.js
briandamaged: Confirmed -- this appears to be fine.  But, we might want to refactor this in a future commit.  (In particular: it's injecting data into a 3rd-party component.  This could become a problem if the component's interface changes in the future)

maximelkin:
https://github.com/Vincit/objection.js/issues/950 related

PR #3684, opened 2020-02-23
a little refactor query/compiler.js

maximelkin:
todo: benchmark this changes


maximelkin:
todo2: review todo in code

maximelkin:
Too many tests depends on alphabetical sort of columns
also user code possibly do same thing

elhigu:
> Too many tests depends on alphabetical sort of columns also user code possibly do same thing

I thought we had already removed alphabetical sorting from insert couple of years ago (or at least from updates) because in some cases actual order of columns in insert / update does matter (IIRC there was some way to write some dependencies between columns which were evaluated in order). Maybe that change was postgresql dialect specific 🤔

PR #3688, opened 2020-02-25
ci improvements

maximelkin:


scripts/docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
 
   mysql:
     image: mysql
-    command: --default-authentication-plugin=mysql_native_password
+    command: --default-authentication-plugin=mysql_native_password --sync_binlog=0 --skip-innodb_doublewrite --innodb-flush-log-at-trx-commit=0 --innodb-flush-method=nosync
```
kibertoad: can you add next to this line comment, explaining rationale behind these parameters?

maximelkin: Ok, later

kibertoad:
What is still to be done for this PR?

maximelkin:
Find options for oracle, mssql

maximelkin:
Not sure about using ramfs - we use too many instances

kibertoad:
@maximelkin Another useful improvement would be to migrate from Travis to CircleCI, btw, since Travis has rather frustrating limits on how often jobs are fired, and in general its future doesn't look particularly bright: https://www.reddit.com/r/devops/comments/at3oyq/it_looks_like_ibera_is_gutting_travis_ci_just_a/

Not sure how difficult such migration would be, but I don't think we are using anything particularly fancy out of Travis features.

kibertoad:
@ksatirli Can you take a quick expert glance on our existing Travis config (including maximelkin's improvements in this PR) and assess how difficult would it be to do same things with CircleCI?

maximelkin:
here only database tuning options, so It shouldn't affect travisci -> circleci migration

kibertoad:
Yeha, I would hope CircleCI also supports passing DB options

maximelkin:
So, disabling durability on databases not affected test performance (222 seconds as usual)

kibertoad:
@maximelkin Could be that Travis does that by default for CI DBs, which makes sense. I would still keep these changes somewhere around, because there is no guarantee CircleCI does the same thing, we would need to measure again; and TBH I wouldn't mind having these commands in config explicitly.

maximelkin:
@kibertoad no, because we running databases via docker-compose

kibertoad:
Ah, I see

elhigu:
> Another useful improvement would be to migrate from Travis to CircleCI, btw, since Travis has rather frustrating limits on how often jobs are fired

Has there been problems with travis so far? I don't see why couldn't we have both running at the same time and see fi Circle CI suits us any better. So far I have been pretty happy with tarvis though (after changing to docker setup). Being able to run tests also in windows would be great though.

kibertoad:
@elhigu That's a very good point, we should do that!

maximelkin:
Coverage decreased because of switch from node 8 to node 12, should we temporary reduce it?

kibertoad:
@maximelkin Wonder why version switch would affect coverage. Any particular reason why switch was needed prior to (incoming) dropping of Node 8 support?

kibertoad:
@maximelkin Was it because of https://github.com/DefinitelyTyped/DefinitelyTyped/issues/42275?

briandamaged:
@kibertoad + @maximelkin : It looks like the oracle tests are only running for node 8.  So, that probably explains why the coverage dropped.

kibertoad:
@briandamaged That was intentional, though, to speed up execution, I think.

briandamaged:
Right.  I'm just saying that it would also cause the coverage to drop since the Oracle code is not being tested.

maximelkin:
>Coverage decreased because of switch from node 8 to node 12, should we temporary reduce it?

Wait
this sound really strange, because coverage checked for all node versions, so changes shouldn't affect it

briandamaged:
@maximelkin : By default, Travis-CI currently uses the `npm run test` script.  This runs the unit tests, but it does not calculate the coverage for them.

maximelkin:
Ok, I got this

maximelkin:
TODO: exclude tests from coverage

kibertoad:
@maximelkin What's the end result of this improvement? Did it speed up tests?

package.json
```diff
@@ -96,6 +96,7 @@
     "mysql": "^2.18.1",
     "mysql2": "^2.1.0",
     "nyc": "^15.0.0",
+    "oracledb": "^4.2.0",
```
kibertoad: Considering that testing Oracle is the slowest step, do we want to run it on all Node versions?

elhigu: We could add an additional fast test that doesn't pull oracle, but I think its better to have more complete tests than fast ones. In either way it will take long to all test setups to finnish

maximelkin: Yes, because in other case we can miss some breaking changes from node for oracledb

PR #3689, opened 2020-02-25
Prepare to release 0.20.11

kibertoad:


lib/raw.js
```diff
@@ -14,7 +14,7 @@ const {
   isNumber,
 } = require('lodash');
 const saveAsyncStack = require('./util/save-async-stack');
-const uuid = require('uuid');
+const uuid = require('uuid').v4;
```
maximelkin: uuid/v4 instead
```suggestion
const uuid = require('uuid/v4');
```

test/integration/builder/inserts.js
```diff
@@ -2,7 +2,7 @@
 
 'use strict';
 
-const uuid = require('uuid');
+const uuid = require('uuid').v4;
```
maximelkin:
```suggestion
const uuid = require('uuid/v4');
```

lib/query/compiler.js
```diff
@@ -19,7 +19,7 @@ const {
   reduce,
   has,
 } = require('lodash');
-const uuid = require('uuid');
+const uuid = require('uuid').v4;
```
maximelkin:
```suggestion
const uuid = require('uuid/v4');
```

maximelkin:
Can you please pause release?
Just found a bug https://github.com/knex/knex/pull/3690

lib/query/compiler.js
kibertoad: Actually, this is not going to work and was actively discouraged by `uuid`, as it couples import to specific directory structure. What I should have used instead is alias for { v4 } require, but I was too lazy to look up syntax for that; will do it now, since that would definitely look cleaner.

maximelkin: Ok, I see

briandamaged:
Howdy!  Is there anything I can do to assist on this front?

briandamaged:
@kibertoad : It looks like updating `@types/node` to version `13.7.7` might resolve the CI failure we're observing.  But, I'm not sure if this is the approach we want to take or not.

kibertoad:
@briandamaged I don't mind updating types at least temporarily, our tests would probably break if we start using some features that are too new, but at least locally I'm still getting errors with tarn types (should be resolved when next tarn version is released, though).

briandamaged:
@kibertoad : I'm not 100% sure on this, but it looks like the `@types/node` version doesn't necessarily mirror the runtime version of Node.  It looks like there are subdirectories for the various runtimes:

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/node

Which... would kinda make sense?  Otherwise, how would the `@types/node` project release bug fixes for the type definitions?

(Again: I'm not 100% sure on this.  I only have basic knowledge of Typescript)

maximelkin:
I think they align versions, check recently published tags in npm

PR #3690, opened 2020-02-26
bugfix mysql query errors ignored

maximelkin:


maximelkin:
http://bluebirdjs.com/docs/api/catch.html#filtered-catch
for check

maximelkin:
So i got this rough test, but its hungs on sqlite and mysql (with both solutions)
```
        knex.transaction(async (trx2) => {
          return trx2.transaction(async (trx) => {
            await trx.savepoint();
            await trx.schema.createTable('test1111124', (t) => t.string('a'));
            await trx.savepoint();
          });
        })
```

briandamaged:
@maximelkin : I think there was just a typo in your original test case.  Try this:

```
    it.only('does not shallow exception when error during transaction occurs', async () => {
      await expect(
        knex.transaction(async (trx) => {
          await trx.raw('incorrect sql');
        })
      ).to.eventually.be.rejected;
    });
```

(The `.to.eventually.be.rejected`  was mistakenly chained to the Knex transaction in the original draft of the test case)

briandamaged:
@maximelkin : However... that test case seems to pass when run against `master` as well.  It looks like something else is intercepting / handling the error before it ever reaches the code that contained the fix.

So... we'll probably need to create some additional test cases as well

maximelkin:
Error from failed rollback/etc transaction methods ignored

kibertoad:
Should this be marked as WIP?

briandamaged:
@kibertoad : Yeah -- it should probably be marked `WIP:` until we figure out a good way to test this.

(On the flipside, though: the changes do appear to recreate the old logic.  We'd just like to introduce some tests to avoid future regressions)

maximelkin:
@briandamaged can you help me with testcase? I dont know how to get raw connection from transaction

test/integration/builder/transaction.js
```diff
@@ -659,5 +659,17 @@ module.exports = function(knex) {
         return builder;
       });
     });
+
+    it('does not shallow exception when error during transaction occurs', async () => {
```
briandamaged: Typo:  "shallow" -> "swallow"

briandamaged:
@maximelkin : Sure!  I've actually been dabbling w/ some test cases for the last day or two, but I haven't had any success yet.  I'll try to create some new ones based upon your latest commits.

briandamaged:
And by "success", I mean:

* The case fails in `master`, but
* succeeds in this branch

So far, all of my test cases have succeeded in both branches, which suggests that they aren't hitting the correct control path.

maximelkin:
@briandamaged check please new test case in branch, it works on my machine

briandamaged:
@maximelkin : The test case succeeds in this branch.

However, when I run it against `master`, it only fails for `mysql`.  Any idea why it would succeed for `mysql2` ?

briandamaged:
@maximelkin : nvm -- stupid question.  It's because of the conditional at the beginning of the test case.

However, I think we should probably try to design the test case so that it can be run against all of the different Client implementations.  (I'm removing the conditional to see how it behaves now...)

test/integration/builder/transaction.js
```diff
@@ -659,5 +659,18 @@ module.exports = function(knex) {
         return builder;
       });
     });
+
+    it('#3690 does not swallow exception when error during transaction occurs', async () => {
+      if (knex.client.driverName !== 'mysql') return;
```
briandamaged: @maximelkin : I think we need to remove this conditional from the beginning of the test.  With the conditional removed:

1. The test cases still passes in this branch, and
2. The test case fails in `master`.  (Specifically: against the `mysql` and `mysql2` Clients, which is the behavior we were expecting to see)

maximelkin: This possibly not works for non mysql, because each client needs special kill connection command

maximelkin: Currently just as proof of concept

briandamaged: Ah -- that makes sense, then.  So yeah: we should figure out a good way to run this test case against all of the various Clients.

test/integration/builder/transaction.js
```diff
@@ -659,5 +659,29 @@ module.exports = function(knex) {
...
+      };
+
+      const killConnection = killConnectionMap[knex.client.driverName];
+      if (!killConnection) return;
+
+      await expect(
+        knex.transaction(async (trx2) => {
+          await killConnection(await trx2.client.acquireConnection(), trx2);
+          await trx2.transaction(async () => 2);
+        })
```
briandamaged: @maximelkin : It looks like some of the exceptions are being thrown for the wrong reason.  Try removing the `await expect(..)` wrapper and watch what happens.  Here's what I'm seeing on my side when I do that:

```

  1) Integration Tests
       oracle | oracledb
         Transactions
           #3690 does not swallow exception when error during transaction occurs:
     TypeError: connection.drop is not a function
      at oracledb (test/integration/builder/transaction.js:674:52)
      at /Users/brianlauber/Documents/src/knex/test/integration/builder/transaction.js:682:15

  2) Integration Tests
       mssql | mssql
         Transactions
           #3690 does not swallow exception when error during transaction occurs:
     RequestError: KILL @p0 - Incorrect syntax near '@p0'.


  3) Integration Tests
       postgresql | pg
         Transactions
           #3690 does not swallow exception when error during transaction occurs:
     Error: ROLLBACK - Client has encountered a connection error and is not queryable

```

(The errors for `oracledb` and `mssql` look incorrect to me)

maximelkin: Sorry, not tested locally oracle and mssql

briandamaged: Ah -- np.  It took me a while to figure out a good way to test those ones locally.  I'll see if I can figure out a good solution for the Oracle / MSSQL ones

maximelkin: mssql works differently with placeholders?

maximelkin: ok, its not allowed to kill with placeholder

maximelkin:
```
Unhandled rejection: TypeError: Cannot assign to read only property 'originalError' of object 'TransactionError: Requests can only be made in the LoggedIn state, not the Final state'
    at conn.tx_.rollback.then (/knex/lib/dialects/mssql/transaction.js:42:38)
```
:man_shrugging:

test/integration/suite.js
```diff
@@ -22,11 +22,13 @@ module.exports = function(knex) {
...
-    require('./builder/updates')(knex);
-    require('./builder/transaction')(knex);
+    require('./builder/updates')(knex);*/
+    require('./builder/transaction')(
+      knex
+    ); /*
     require('./builder/deletes')(knex);
     require('./builder/additional')(knex);
-    require('./datatype/bigint')(knex);
+    require('./datatype/bigint')(knex);*/
```
briandamaged: FYI: You can also run specific test cases by appending `.only` to the declaration.  Ex:

```
it.only('tests whatever', function() {

});
```

maximelkin: yes, thanks, but i was too lazy to check is where another describe nesting

maximelkin: (this not intended to be committed lol)

test/integration/builder/transaction.js
```diff
@@ -659,5 +659,34 @@ module.exports = function(knex) {
         return builder;
       });
     });
+
+    it('#3690 does not swallow exception when error during transaction occurs', async () => {
```
briandamaged: We actually should be using standard functions instead of arrow functions when creating `mocha` test cases.  Otherwise, we'll actually lose some `mocha` functionality:

https://remarkablemark.org/blog/2016/08/20/arrow-functions-in-mocha/

maximelkin: For this.skip()?

test/integration/builder/transaction.js
```diff
@@ -659,5 +659,34 @@ module.exports = function(knex) {
...
+        oracledb: async (connection, trx) => {
+          const id = (await trx.raw(
+            `select sys_context('userenv','sessionid') id from dual;`
+          ))[0].id;
+          return knex.raw(`Alter System Kill Session '${id}'`);
+        },
+      };
+
+      const killConnection = killConnectionMap[knex.client.driverName];
+      if (!killConnection) return;
```
briandamaged: If we use a standard function when declaring the `mocha` test, then we can do the following:

```suggestion
      if (!killConnection) {
        return this.skip();
      }
```

This will mark the test as "pending" rather than "succeeded" for the cases that we currently do not cover.

test/integration/builder/transaction.js
briandamaged: Exactly.

test/integration/suite.js
briandamaged: Yeah -- I figured as much.  lol

test/integration/builder/transaction.js
```diff
@@ -659,5 +659,34 @@ module.exports = function(knex) {
...
+          const id = (await trx.raw('select @@SPID as id'))[0].id;
+          return knex.raw(`KILL ${id}`);
+        },
+        oracledb: async (connection, trx) => {
+          const id = (await trx.raw(
+            `select sys_context('userenv','sessionid') id from dual;`
+          ))[0].id;
+          return knex.raw(`Alter System Kill Session '${id}'`);
+        },
+      };
```
briandamaged: So: it looks like this logic is already encapsulated in a `.cancelQuery(..)` method for the various clients that support it.  Ex:

https://github.com/knex/knex/blob/b025aea31881e9a08c67ae0ccb6a4c1ceeae2ef2/lib/dialects/mysql/index.js#L172-L184

So, we can probably just call that method instead.

Additionally, Clients that support this functionality also set `canCancelQuery = true`.  So, this means we can probably also do:

```
if(!knex.client.canCancelQuery) {
  return this.skip();
}
```

maximelkin: This is not cancelling query, this is killing connection/session

briandamaged: Ah -- good call.  I just saw the `KILL` keyword and jumped to conclusions.

Anyhoo: for now, maybe we should limit the scope to the handful of clients that we've been able to test successfully  (mysql, mysql2, pg, and maybe redshift).  We can then mark the others as "pending" and follow up on this after the `0.20.11` release.

@maximelkin + @kibertoad : Thoughts?

maximelkin:
added oracle
todo: oracle and mssql broken on that test

test/integration/builder/transaction.js
maximelkin: +1
fix mssql and oracle unhandled rejections may require more hours to fix

test/integration/builder/transaction.js
maximelkin: done

test/integration/builder/transaction.js
kibertoad: Sounds good :)

maximelkin:
@kibertoad @briandamaged tests passed, oracle and mssql cases commented with todo

test/integration/builder/transaction.js
```diff
@@ -659,5 +659,46 @@ module.exports = function(knex) {
         return builder;
       });
     });
+
+    it('#3690 does not swallow exception when error during transaction occurs', async function() {
+      const killConnectionMap = {
+        mysql: async (connection) => knex.raw('KILL ?', [connection.threadId]),
```
kibertoad: so mysql and mysql2 are same? can we set arrow function to const and reuse it?

test/integration/builder/transaction.js
```diff
@@ -659,5 +659,46 @@ module.exports = function(knex) {
...
+        mysql2: async (connection) => knex.raw('KILL ?', [connection.threadId]),
+        pg: async (connection) =>
+          knex.raw('SELECT pg_terminate_backend(?)', [connection.processID]),
+        redshift: async (connection) =>
```
kibertoad: are we running tests on redshift in ci?

maximelkin: No, can be removed

test/integration/builder/transaction.js
maximelkin: done

test/integration/builder/transaction.js
maximelkin: done

PR #3692, opened 2020-02-28
fix(types): Client.validateConnection must be sync

mastermatt:
The result of this method is passed directly to tarn `validate`, which does not resolve the value only truthy checks it.


elhigu:
Tarn will support async validate in next release.

mastermatt:
https://github.com/Vincit/tarn.js/pull/45

PR #3695, opened 2020-03-01
Mark missing test cases as "pending" rather than "passed"

briandamaged:
Currently, the Knex codebase marks several tests as "passing" even though no test was actually run.  This is then giving us a false sense of how well the codebase has really been tested.  So, this is a quick pass over the code that marks several of these missing test cases as `"pending"` instead of `"passing"`.  From there, we can then decide upon an appropriate way to refactor each of these test cases.

### DIFF
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -452,7 +452,7 @@ module.exports = function(knex) {
 
     it('#2184 - should properly escape table name for SQLite columnInfo', function() {
       if (knex.client.driverName !== 'sqlite3') {
-        return;
+        return this.skip();
       }
 
       return knex.schema
@@ -664,10 +664,10 @@ module.exports = function(knex) {
     it('.timeout() should throw TimeoutError', function() {
       const driverName = knex.client.driverName;
       if (driverName === 'sqlite3') {
-        return;
+        return this.skip();
       } //TODO -- No built-in support for sleeps
       if (/redshift/.test(driverName)) {
-        return;
+        return this.skip();
       }
       const testQueries = {
         pg: function() {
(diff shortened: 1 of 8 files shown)


briandamaged:
I suspect that most of these test cases fall into one of two categories:

1. The test case is applicable to this combination of inputs, but it was never written.
2. The test case is not applicable to this combination of inputs.  (Ex: the feature is not supported by this specific Dialect)

For this 2nd category, we will ultimately want to restructure the code so that the combination of inputs is detected / discarded _before_ a test case has started.  Thus, the scenario will be omitted from the output entirely (Instead of being marked as `"pending"` or `"passing"`).  Ideally, we should introduce some primary indicators/metadata into the codebase to handle this.  For example, instead of doing this:

```
it('throws an error if client does not support dropSchema', async function() {
  if(/pg/i.test(knex.client.driverName) {
    // Why?
    return this.skip();
  }

  // ...
}
```

It should be doing something like this:

```
// Got it.  This test case is only applicable when the Client does not support schema dropping
if(!example.Client.canDropSchema) {
  it('throws an error if client does not support dropSchema', async function() {
    // ...
  }
}
```

test/integration/builder/additional.js
```diff
@@ -758,7 +758,7 @@ module.exports = function(knex) {
         expect(addTimeout).to.throw(
           'Query cancelling not supported for this dialect'
         );
-        return;
+        return; // TODO: Use `this.skip()` here?
```
briandamaged: In this case, it looks like there should potentially be 2 separate test cases.  The 2nd test case is only applicable to a subset of Dialects.

test/tape/pool.js
```diff
@@ -101,7 +101,16 @@ test('#2321 dead connections are not evicted from pool', async (t) => {
...
+
+      // TODO: It looks like this test case can trigger the race condition
+      //       outlined in this issue comment:
+      //
+      //         https://github.com/knex/knex/issues/3636#issuecomment-592005391
+      //
+      //       For now, we're working around this problem by introducing a
+      //       1 second delay.  But, we should revisit this once the connection
+      //       pool logic has been refactored.
+      await delay(1000); // wait driver to notice connection errors (2ms was enough locally)
```
briandamaged: Technically, this change really shouldn't be part of this PR.  But, the race condition caused the CI to fail, and this seemed like the easiest way to kick things off again.  (Related note: is there a re-run button in Travis-CI?  I haven't been able to find it)

briandamaged: nvm -- just found the button in Travis-CI.  I just had to login before it appeared.

kibertoad:
Yup, makes sense.

PR #3697, opened 2020-03-02
Cleanup/remove dead oracle code

briandamaged:
It appears that `/lib/dialects/oracle/` is mostly dead code.  More specifically:

1. It is not listed within the `SUPPORTED_CLIENTS` constant.
2. Portions of the code are subclassed by `/lib/dialects/oracledb/`, but...
3. ... `/lib/dialects/oracledb/` overrides many of the methods from `/lib/dialects/oracle/`

So, this PR just removes some of the dead code so that we can have less code to maintain going forward.

### DIFF
--- a/lib/dialects/oracle/index.js
+++ b/lib/dialects/oracle/index.js
@@ -49,20 +49,6 @@ Object.assign(Client_Oracle.prototype, {
     return new TableCompiler(this, ...arguments);
   },
 
-  prepBindings(bindings) {
-    return map(bindings, (value) => {
-      // returning helper uses always ROWID as string
-      if (value instanceof ReturningHelper && this.driver) {
-        return new this.driver.OutParam(this.driver.OCCISTRING);
-      } else if (typeof value === 'boolean') {
-        return value ? 1 : 0;
-      } else if (Buffer.isBuffer(value)) {
-        return bufferToString(value);
-      }
-      return value;
-    });
-  },
-
   // Return the database for the Oracle client.
   database() {
     return this.connectionSettings.database;
--- a/lib/dialects/oracle/index.js
+++ b/lib/dialects/oracle/index.js
@@ -1,19 +1,13 @@
 // Oracle Client
 // -------
-const { map, flatten, values } = require('lodash');
-const { promisify } = require('util');
-
 const inherits = require('inherits');
 const Client = require('../../client');
-const { bufferToString } = require('../../query/string');
-const Formatter = require('./formatter');
 
-const QueryCompiler = require('./query/compiler');
 const SchemaCompiler = require('./schema/compiler');
 const ColumnBuilder = require('./schema/columnbuilder');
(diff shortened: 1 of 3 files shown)


lib/dialects/oracle/transaction.js
```diff
@@ -2,75 +2,4 @@ const Transaction = require('../../transaction');
...
-          if (!configConnection) {
-            this.client.releaseConnection(connection);
-          } else {
-            debugTx('%s: not releasing external connection', this.txid);
-          }
-        }
-      });
-  }
-};
+module.exports = class Oracle_Transaction extends Transaction {};
```
briandamaged: Every method of the `Transaction` class is overridden.  But, I'm leaving this class in place for now just to minimize changes to other files.

briandamaged:
I forgot to mention: it also looks like the `oracle` library hasn't been supported in the last 3 years:

https://www.npmjs.com/package/oracle

lib/dialects/oracle/transaction.js
briandamaged: Just realized that `/lib/dialects/oracledb` wasn't even subclassing this code.  So, now I've nuked it.

lib/dialects/oracle/index.js
```diff
@@ -30,22 +23,6 @@ Object.assign(Client_Oracle.prototype, {
...
-  },
-
-  formatter() {
-    return new Formatter(this, ...arguments);
-  },
-
-  queryCompiler() {
-    return new QueryCompiler(this, ...arguments);
-  },
-
```
briandamaged: `Client_Oracledb` overrides all of these methods

lib/dialects/oracle/index.js
```diff
@@ -62,44 +39,6 @@ Object.assign(Client_Oracle.prototype, {
...
-    });
-  },
-
-  // Used to explicitly close a connection, called internally by the pool
-  // when a connection times out or the pool is shutdown.
-  async destroyRawConnection(connection) {
-    const close = promisify((cb) => connection.close(cb));
-    return close();
-  },
-
```
briandamaged: `Client_Oracledb` overrides all of these methods

lib/dialects/oracle/index.js
```diff
@@ -135,61 +74,6 @@ Object.assign(Client_Oracle.prototype, {
...
-            return response;
-          }
-          // return an array with values if only one returning value was specified
-          return flatten(map(response, values));
-        }
-        return obj.rowsAffected;
-      default:
-        return response;
-    }
-  },
```
briandamaged: `Client_Oracledb` overrides all of these methods

briandamaged:
Alright -- I think this is a good stopping point.  Currently, the PR just focuses on the dead code that could be removed easily.

In a future PR, we can begin migrating the non-dead code into the `oracledb` dialect.

PR #3698, opened 2020-03-02
Cleanup/extract transaction ordering logic (implementation 2.0)

briandamaged:
While refactoring some of the other `Transaction` logic, we noticed that some of the `Transaction` subclasses were not enforcing the execution ordering guarantees.  So, this PR restructures the code to ensure that this guarantee is always enforced.

Note: this is a re-implementation of #3679 .  It was easier to re-implement the logic rather than trying to resolve all of the merge conflicts.


lib/transaction.js
```diff
@@ -211,30 +223,19 @@ class Transaction extends EventEmitter {
...
+      } finally {
+        if (!configConnection) {
+          debug('%s: releasing connection', this.txid);
+          this.client.releaseConnection(connection);
+        } else {
+          debug('%s: not releasing external connection', this.txid);
         }
-      });
+      }
+    });
```
briandamaged: Dang... it looks like the `git` commit hooks automatically fixed the indentation.  (I was hoping to leave the indentation as-is, and fix it in a separate PR.  Oh well.)

lib/transaction.js
```diff
@@ -211,30 +223,19 @@ class Transaction extends EventEmitter {
...
         reject(e);
       }
-    })
-      .then((connection) => {
+    }).then(async (connection) => {
+      try {
         connection.__knexTxId = this.txid;
-
-        return (
-          this._previousSibling
```
briandamaged: Short story: `Transaction#acquireConnection(..)` is no longer responsible for waiting on `this._previousSibling`.  This responsibility has been moved to `Transaction#_kickThingsOff(..)` so that its enforcement can be guaranteed across `Transaction` subclasses.

lib/transaction.js
```diff
@@ -49,53 +49,7 @@ class Transaction extends EventEmitter {
...
-              });
-          }
-          return null;
-        })
-        .catch((e) => {
-          return this._rejecter(e);
-        });
-
-      return executionPromise;
-    });
```
briandamaged: This logic was extracted into `Transaction#_kickThingsOff(..)`

test/integration/builder/transaction.js
```diff
@@ -415,6 +409,31 @@ module.exports = function(knex) {
...
+          trx.transaction(async function(trx2) {
+            const rows = await trx2(TABLE_NAME);
+
+            // Even though `trx1` was delayed, `trx2` waited patiently for `trx1`
+            // to finish.  Therefore, `trx2` discovers that there is already 1 row.
+            expect(rows.length).to.equal(1);
+          }),
+        ]);
+      });
+    });
```
briandamaged: The other test case for #2213 would still succeed for `oracledb` even though the transaction ordering had not been properly enforced.

So, this raises an important question: should this transaction ordering behavior apply to `oracledb` as well?  If so, then I think the current PR is good as-is.  Otherwise, we might need to do some additional refactoring to extract this behavior even further.

briandamaged:
So, I think this is a good stopping point for the current PR.  #3679 contained a few additional revisions, but I'd rather handle these in a separate PR this time.

lib/transaction.js
```diff
@@ -200,6 +154,64 @@ class Transaction extends EventEmitter {
     return this;
   }
 
+  async _kickThingsOff(config, container) {
+    // FYI: This is temporarily stalling things so that the constructor
```
maximelkin: maybe move _previousSibling higher than this._kickThingsOff?

briandamaged: It's a bit of a chicken-and-egg problem because `_previousSibling` is derived from `basePromise` (Well, more precisely: the previous Transaction's `basePromise`).  My concern is that this will introduce a subtle race condition that can cause `outerTx._lastChild` to point to the wrong `Promise`.

But... we might be able to address that by changing how/when `outerTx._lastChild` gets assigned.  I'll have to think about that a bit more.

lib/transaction.js
```diff
@@ -49,53 +49,7 @@ class Transaction extends EventEmitter {
...
-          }
-          return null;
-        })
-        .catch((e) => {
-          return this._rejecter(e);
-        });
-
-      return executionPromise;
-    });
+    const basePromise = this._kickThingsOff(config, container);
```
kibertoad: `_kickThingsOff` this sounds slightly too informal and vague, can we make it more specific?

lib/transaction.js
```diff
@@ -200,6 +154,64 @@ class Transaction extends EventEmitter {
     return this;
   }
 
+  async _kickThingsOff(config, container) {
+    // FYI: This is temporarily stalling things so that the constructor
+    //      can finish initializing the Transaction.  Otherwise,
+    //      `this.previousSibling` will still be `undefined`.
+    await Promise.resolve();
```
kibertoad: This looks hacky. We can't reshuffle things around so that this method is only invoked after constructor is done?

lib/transaction.js
briandamaged: Saying this another way: I think this relates to a larger problem that I'm seeing throughout the `Transaction` class.

There are currently several places throughout the `Transaction` class where the `Transaction` needs to be aware of how/when it is being constructed/used.  In this case, the `Transaction` needs to change its behavior depending upon whether or not there was an `outerTx`.

So, one of my main goals in refactoring the `Transaction` class is to move these responsibilities into a higher layer of the code.  But, there will probably be 2-3 separate PRs before this goal is completed.

lib/transaction.js
briandamaged: FYI: There is another conversation thread discussing this a few lines above.

I agree that this is "hacky"; however, this PR didn't actually introduce the hack.  It was implicitly occurring in the previous implementation as well.  Specifically, it:

1. Obtained a `connection`.  (which stalled things)
2. Waited for `_this.previousSibling`.  (Which was now defined thanks to the stall)
3. Continued processing `connection`.

So, this new implementation is simply making the hack more apparent.

kibertoad: Makes sense :)

lib/transaction.js
briandamaged: Yeah -- I haven't been able to think of a good name for it so far.  My guess is that this function will most likely disappear by the time the overall refactoring has been completed.

It's a bit tricky to pick a name because this function is currently doing too many things.  Ex:

1. Ensuring that the previous Transaction has completed
2. Acquiring a connection
3. (Possibly) setting up disposal for the connection
4. Creating a Transactor
5. (Possibly) setting up commit/rollback guarantees for the Transactor
6. Yielding control back to the caller

So yeah... it's tricky to come up w/ a name that encompasses all of those things.  lol

briandamaged: How about: `_evaluateContainer(..)` ?

test/integration/builder/transaction.js
```diff
@@ -395,12 +395,6 @@ module.exports = function(knex) {
     });
 
     it('#2213 - should wait for sibling transactions to finish', function() {
-      if (/redshift/i.test(knex.client.driverName)) {
-        return Promise.resolve();
-      }
```
briandamaged: TBH: I'm not really sure why this test case was skipped for `redshift`.  Does anybody have a `redshift` test environment that we can try?

maximelkin: maybe redshift not supports savepoints?

maximelkin: https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-functions.html
Yes, this is not supported

lib/transaction.js
briandamaged: Here's what I'm planning for some upcoming PRs:

As mentioned previously, the `Transaction` class often needs to change its behavior depending upon how/when it is being constructed/used.  However, it does not have any direct way to know this information.  Instead, it infers the details by:

1. Checking for the presence of `outerTx`.  (Let's the `Transaction` know whether or not it is a "top-level" transaction)
2. Checking for the presence of `config.connection`.  (This seems to be another way for the `Transaction` to determine whether it is top-level or not)
3. Checking to see if `container` returned a `Promise`.  (Helps the `Transaction` decide if it's responsible for the commit/rollback operations)

It looks like most of these details can be extracted into the Factory components that are instantiating the `Transactions`.  More specifically: the `Factory` components can pass in the appropriate strategy functions / polymorphic objects when instantiating the `Transaction`s.

For example: right now, there are several places in the code where `Transaction` changes its behavior depending upon whether or not `outerTx` was defined.  Instead, the code will be revised so that `outerTx` is _always_ defined as one of two implementations:

1. One implementation that is used by "Top-Level" transactions
2. A second implementation that is used by "Nested" transactions

So, how do we decide which implementation to use?  From what I can tell, it's actually fairly straightforward:

1. Factories instantiated through `makeKnex` always use the "Top-Level" implementation
2. Factories instantiated through `makeTransactor` always use the "Nested" implementation

(Note: right now, the main difference between these two Factories seems to be whether or not `outerTx` is passed into the `Transaction` constructor.  So, this seems to align nicely w/ this proposal)

Once this change is in place, it becomes possible for the `Transaction` class to delegate various decisions to this object.  For example:

1. Does the `connection` need to be disposed once the `Transaction` has concluded?
2. Does the `Transaction` need to register itself as a sibling?
3. Etc.

That's the high-level idea, at least.  But, as you probably guessed, this also means that this is going to change the API for the `Transaction` constructor.  This is a somewhat-breaking change, but fortunately it appears that the `Transaction` constructor is never actually called directly.

Anyhoo: this is my main reason for leaving the "Temporarily stall things" hack in place for now.  We can go ahead and fix it in this PR, but it also means that the scope of this PR will increase quite a bit.  I'd rather merge this minor enhancement/fix first, and then tackle the larger code restructuring afterwards.

cc: @maximelkin + @kibertoad

test/integration/builder/transaction.js
briandamaged: Aiight -- I'll have this fixed in a few mins.  I'm just re-running the unit tests first to make sure I didn't break anything when changing the test cases.

lib/transaction.js
maximelkin: @briandamaged nice!

briandamaged: So, an interesting discovery while investigating some of these ideas: it appears that `config.connection` is only used by `sqlite3`.  From what I can tell, it's being used for 2 reasons:

1. To facilitate a handful of `sqlite3` unit tests. (ie: allow the unit tests to bypass certain logic)
2. To ensure that there is only ever 1 connection created.  (it looks like `sqlite3` might not work properly when there is more than one connection in the pool)

As an experiment, I removed the `config.connection` logic.  Most things seem to work fine, but a handful of `sqlite3` operations break because they are unable to obtain a connection (because the pool max is 1).  So, I'm doing a bit more investigation to figure out why `sqlite3` is attempting to create new connections in those specific cases.

PR #3702, opened 2020-03-04
whatwg url connection string parsing

maximelkin:
tldr: stop using legacy url.parse api and use new URL() instead

Changes:
https://nodejs.org/api/url.html#url_percent_encoding_in_urls

### DIFF
--- a/lib/util/parse-connection.js
+++ b/lib/util/parse-connection.js
@@ -62,10 +62,10 @@ function connectionObject(parsed) {
     connection.port = parsed.port;
   }
   if (parsed.username || parsed.password) {
-    connection.user = parsed.username;
+    connection.user = decodeURIComponent(parsed.username);
   }
   if (parsed.password) {
-    connection.password = parsed.password;
+    connection.password = decodeURIComponent(parsed.password);
   }
   if (parsed.searchParams) {
     for (const [key, value] of parsed.searchParams.entries()) {
--- a/lib/util/parse-connection.js
+++ b/lib/util/parse-connection.js
@@ -1,13 +1,23 @@
-const url = require('url');
 const { parse } = require('pg-connection-string');
 const parsePG = parse;
 const isWindows = process && process.platform && process.platform === 'win32';
 
+/**
+ * @param str
+ * @returns {URL}
+ */
+function tryParse(str) {
+  try {
+    return new URL(str);
+  } catch (e) {
+    return null;
+  }
+}
+
 module.exports = function parseConnectionString(str) {
-  const parsed = url.parse(str, true);
-  let { protocol } = parsed;
-  const isDriveLetter = isWindows && protocol && protocol.length === 2;
-  if (protocol === null || isDriveLetter) {
+  const parsed = tryParse(str);
+  const isDriveLetter = isWindows && parsed && parsed.protocol.length === 2;
+  if (!parsed || isDriveLetter) {
     return {
       client: 'sqlite3',
       connection: {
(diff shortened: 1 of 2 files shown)


kibertoad:
Would it be possible to add test for this?

also are you still sure you don't want to be a maintainer :-p? You make a great dynamic duo with @briandamaged

maximelkin:
>Would it be possible to add test for this?
Ok, next evening

>also are you still sure you don't want to be a maintainer :-p?
¯\_(ツ)_/¯

maximelkin:
Some tests for this logic are in test/tape/parse-connection.js

kibertoad:
@maximelkin If there are no complaints after people upgrade to 0.20.11 and we don't need to release a hotfix, I think we could bump version to 0.21.0 and drop Node 8 already. April is close.

kibertoad:
Since there is a separate 0.20 branch for hotfixes, and master is for 1.0.0, we can merge this breaking change after CI is green and documentation is updated.

lib/util/parse-connection.js
```diff
@@ -1,20 +1,32 @@
-const url = require('url');
+const { URL } = require('url');
```
TrySound, line 1: Since node 10 URL is available globally like in browsers. This require can be removed.

lib/util/parse-connection.js
```diff
@@ -1,20 +1,31 @@
...
+    return new URL(str);
+  } catch (e) {
+    return null;
+  }
+}
+
 module.exports = function parseConnectionString(str) {
-  const parsed = url.parse(str, true);
-  let { protocol } = parsed;
-  const isDriveLetter = isWindows && protocol && protocol.length === 2;
```
maximelkin, line 9: protocol is always string, so && protocol && removed

kibertoad: Here is the reason why it was made defensive: #3743
It was failing under certain circumstances.

lib/util/parse-connection.js
maximelkin: done

kibertoad:
@maximelkin Could you add notes about incompatible changes to `UPGRADING.md`?

test/tape/parse-connection.js
```diff
@@ -49,6 +51,92 @@ test('mysql connection protocol with query string params', function (t) {
...
+        user: 'username',
+        host: 'path.to.some-url',
+        port: '3306',
+        database: 'testdb',
+      },
+    }
+  );
+});
+
+test('decodes username and password', function(t) {
```
kibertoad, line 113: we also need a test for sqlite Windows path, e. g. 'C:/node/database.sqlite'

kibertoad (changes requested):
See comments

lib/util/parse-connection.js
maximelkin: This check now not actual, because url.parse can return protocol null, and in URL case this is impossible

kibertoad:
@maximelkin Tests are failing now

maximelkin:
yea, i know, just searched for fix

kibertoad (approved):

test/tape/parse-connection.js
maximelkin: done

kibertoad:
Awesome, thanks!

kibertoad:
@maximelkin Can you also update https://github.com/knex/knex/blob/master/UPGRADING.md with notes about changes in how connection string is interpreted?

maximelkin:
https://nodejs.org/api/url.html#url_percent_encoding_in_urls
If you use something unusual for hostname/database name/auth parameters in url (not A-z, digits, dot, dash) - check how it would work.

kibertoad:
That would work, if there is also an explicit mention that previously rules for legacy API applied, and now the WHATWG API do, with the URL reference that you gave.

Issue #3699, opened 2020-03-03
Knex is not clearing timeouts after they have become unnecessary


prk3:
# Environment

Knex version: 0.20.10
Database + version: postgres 12.1
OS: Linux
Node: 12.15.0
Jest: 25.1.0

# Bug

Jest prints a warning about open handles when a test creates and rollbacks a transaction. When I comment out transacting code, the warning disappears. The connection to postgres database is closed in afterAll call.

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your
 tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

`--detectOpenHandles` flag does not detect anything.

Here is the reproduction code:
```javascript
const Knex = require('knex');

const db = Knex({
  client: process.env.DB_TYPE || 'pg',
  connection: {
    host: process.env.DB_HOST || 'localhost',
    port: Number(process.env.DB_PORT) || 5432,
    user: process.env.DB_USER || 'postgres',
    password: process.env.DB_USER || 'postgres',
    database: process.env.DB_DATABASE || 'postgres',
  },
});

describe('test', () => {
  afterAll(async () => {
    await db.destroy();
  });

  it('transacts', async () => {
    const t = await db.transaction();

    await t.rollback();
  });
});
```

elhigu:
Is it only with transactions or does making any query provoke this? Does it happen also if you commit transaction correctly?

Also that transaction code is  invalid. You should use knex.transactionProvider for that kind of scenario.

```
  it('transacts', async () => {   
    const trxProvider = knex.transactionProvider();
    const trx = await trxProvider();
    trx.rollback(new Error('Fail'));
  
    try {
      await trx.executionPromise;
    } catch (err) {
      // just to prevent unhandled rejection caused by rollback
    }
  }); 
```

or just the safer handler way with implicit commit / rollback , which works for most use cases just fine:

 ```
  it('transacts', async () => {
     try {
        // just implicitly rolls back because of thrown exception
        await db.transaction(async trx => {
           threw now Error("will implicitly rollback because handler did reject");
        }); 
     } catch (err) {
        // to prevent unhandled rejection
     }
  }); 
```

elhigu:
Sorry I just read the issue again... I'm pretty sure this just happens because of invalid use of knex. Might be that await db.destroy() doesn't complete because of invalid use of transactions there might be connection leased from pool that is never returned there. 

[...]

PR #3703, opened 2020-03-04
timeout(..) function clears unnecessary timeouts

briandamaged:
As noted in #3699 , the `timeout(..)` function wasn't clearing the calls to `setTimeout(..)` once the wrapped Promise had completed.  As a result, this would cause programs to hang for a few seconds if they attempted to exit shortly after resolving a `timeout(..)`.

### DIFF
--- a/lib/util/timeout.js
+++ b/lib/util/timeout.js
@@ -7,11 +7,21 @@ class KnexTimeoutError extends Error {
   }
 }
 
+function timeout(promise, ms) {
+  return new Promise(async function(resolve, reject) {
+    const id = setTimeout(function() {
+      reject(new KnexTimeoutError('operation timed out'));
+    }, ms);
+
+    try {
+      resolve(await promise);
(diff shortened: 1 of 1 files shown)


lib/util/timeout.js
```diff
@@ -7,11 +7,21 @@ class KnexTimeoutError extends Error {
   }
 }
 
+function timeout(promise, ms) {
+  return new Promise(async function(resolve, reject) {
```
kibertoad: I guess I need to rewatch https://www.youtube.com/watch?v=e3Nh350b6S4&feature=youtu.be, but wasn't using async inside promise callback kinda dangerous?

briandamaged: Yeah -- it can be dangerous if there is a bug in your error handling.  So, I think that's the reason they advise against it.

In either case: I already rewrote the logic to do the same thing without `async`.

PR #3705, opened 2020-03-06
Fixed a few unhandled Promise rejections in a test case

briandamaged:
Fortunately, these unhandled Promise rejections turned out to be in the test case code!  (Rather than being a deeper problem that had to be tracked down)

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -10,7 +10,7 @@ global.chai.use(require('sinon-chai'));
 global.d = new Date();
 
 // '.timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection' produced unhandled rejection and it's unclear how to avoid that
-const EXPECTED_REJECTION_COUNT = 2;
+const EXPECTED_REJECTION_COUNT = 0;
 const rejectionLog = [];
 process.on('unhandledRejection', (reason) => {
   console.error('Unhandled rejection:', reason);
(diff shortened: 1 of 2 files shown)


maximelkin:
Maybe also reduce unhandledRejections counter to 0 in tests?

briandamaged:
@maximelkin : I was thinking the same thing.

@kibertoad + @elhigu : Thoughts on setting `EXPECTED_REJECTION_COUNT = 0` ?  We've fixed most (all?) of the other unhandled promise rejections over the last few weeks.  (I'll run the tests on loop in a few minutes just to see if any other cases arise)

kibertoad:
@briandamaged Yeah, setting it to 0 and seeing if tests are still green would be a great idea.

briandamaged:
Alright -- I went ahead and set `EXPECTED_REJECTION_COUNT` to `0`.  I've been running the tests on loop for about an hour now, and so far there have been no failures.

test/index.js
```diff
@@ -10,7 +10,7 @@ global.chai.use(require('sinon-chai'));
 global.d = new Date();
 
 // '.timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection' produced unhandled rejection and it's unclear how to avoid that
```
kibertoad: probably comment is obsolete now

PR #3706, opened 2020-03-06
Fix: Added missing call to _reject in Transactor#transaction

briandamaged:


lib/transaction.js
```diff
@@ -280,13 +280,7 @@ function makeTransactor(trx, connection, trxClient) {
...
     } else {
       return new Promise((resolve, _reject) => {
-        trxClient.transaction(
-          (nestedTrx) => {
-            resolve(nestedTrx);
-          },
-          options,
-          trx
-        );
+        trxClient.transaction(resolve, options, trx).catch(_reject);
```
kibertoad: in which case was this triggered? is it difficult to simulate in a test?

briandamaged: Hmm... I can probably come up w/ a test case for it.  I'll give it a try in a few minutes.

briandamaged: Thanks for reviewing all of these, btw!  Any hesitations before we merge this change?

kibertoad:
I regret nothing.

PR #3707, opened 2020-03-06
Rewrote Transaction#acquireConnection() methods to use async

briandamaged:
The code should be functionally equivalent before and after this change.  It's mostly a minor cleanup so that the next phase of refactoring will be easier.

### DIFF
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -55,48 +55,38 @@ module.exports = class Transaction_MSSQL extends Transaction {
   // Acquire a connection and create a disposer - either using the one passed
   // via config or getting one off the client. The disposer will be called once
   // the original promise is marked completed.
-  acquireConnection(config, cb) {
+  async acquireConnection(config, cb) {
     const configConnection = config && config.connection;
-    return new Promise((resolve, reject) => {
-      try {
-        resolve(
-          (this.outerTx ? this.outerTx.conn : null) ||
-            configConnection ||
-            this.client.acquireConnection()
-        );
-      } catch (e) {
-        reject(e);
+    const conn =
+      (this.outerTx && this.outerTx.conn) ||
+      configConnection ||
+      (await this.client.acquireConnection());
+
+    try {
+      conn.__knexTxId = this.txid;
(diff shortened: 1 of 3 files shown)


lib/dialects/mssql/transaction.js
```diff
@@ -55,48 +55,38 @@ module.exports = class Transaction_MSSQL extends Transaction {
...
-      } catch (e) {
-        reject(e);
+    const conn =
+      (this.outerTx && this.outerTx.conn) ||
+      configConnection ||
+      (await this.client.acquireConnection());
+
+    try {
+      conn.__knexTxId = this.txid;
+      if (!this.outerTx) {
```
maximelkin: this part was outside try catch

briandamaged: Yeah -- there were a handful of lines that were outside of the `try/finally` in the previous code.  But, because the connection had already been established, they really should have been inside the `try` block.  Otherwise, errors would cause connections to be unreleased.

(Granted: the possibility of an error in that code was extremely low.  This is mostly just a precaution)

maximelkin: Ok

lib/dialects/oracledb/transaction.js
```diff
@@ -50,40 +50,28 @@ module.exports = class Oracle_Transaction extends Transaction {
     return this.query(conn, `SAVEPOINT ${this.txid}`);
   }
 
-  acquireConnection(config, cb) {
+  async acquireConnection(config, cb) {
     const configConnection = config && config.connection;
```
maximelkin: hm
why here configConnection retrieved, but not used in transaction?

maximelkin: This is not introduced in pull request, so shouldn't be fixed right now

briandamaged: Oh -- it's actually referenced towards the very bottom.  It helps the `Transaction` decide whether or not it should release the connection.

(But, you're actually correct: even though the code is there, it doesn't actually do anything.  It's either leftover from an older feature, or code that was copied unnecessarily)

PR #3708, opened 2020-03-06
WIP: Experiment/inject transaction connection

briandamaged:
Experimental branch that extracts the connection acquisition logic out of `Transaction`.  Instead, `Transactions` are provided with a `withConnection(..)` function that handles the details.

The end-result: `Transactions` no longer need to figure out if the connection should be released.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -33,6 +33,11 @@ const debugQuery = (sql, txId) => _debugQuery(sql.replace(/%/g, '%%'), txId);
 
 const { POOL_CONFIG_OPTIONS } = require('./constants');
 
+const ReuseConnection = (connection) =>
+  async function withConnection(next) {
+    return next(connection);
+  };
+
 // The base client provides the general structure
 // for a dialect specific client object.
 function Client(config = {}) {
--- a/lib/dialects/sqlite3/schema/ddl.js
+++ b/lib/dialects/sqlite3/schema/ddl.js
@@ -19,6 +19,11 @@ const {
   chunk,
 } = require('lodash');
 
+const ReuseConnection = (connection) =>
+  async function(next) {
+    return next(connection);
+  };
+
 // So altering the schema in SQLite3 is a major pain.
 // We have our own object to deal with the renaming and altering the types
 // for sqlite3 things.
@@ -282,7 +287,7 @@ assign(SQLite3_DDL.prototype, {
           return omit(row, mappedFrom);
         });
       },
-      { connection: this.connection }
+      { withConnection: ReuseConnection(this.connection) }
     );
   },
 
@@ -312,7 +317,7 @@ assign(SQLite3_DDL.prototype, {
             );
           });
       },
-      { connection: this.connection }
+      { withConnection: ReuseConnection(this.connection) }
     );
   },
 
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -18,6 +18,12 @@ const ReuseConnection = (connection) =>
     return next(connection);
   };
 
+const InjectTransactionID = (txid) => (next) =>
+  async function(connection) {
+    connection.__knexTxId = txid;
+    return next(connection);
+  };
+
 // FYI: This is defined as a function instead of a constant so that
 //      each Transactor can have its own copy of the default config.
 //      This will minimize the impact of bugs that might be introduced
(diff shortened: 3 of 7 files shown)


briandamaged:
FYI: I don't have any plans to merge this branch.  I just created the PR to get initial feedback on the high-level approach.  Afterwards, we can go back and implement this more cleanly.

lib/client.js
```diff
@@ -33,6 +33,11 @@ const debugQuery = (sql, txId) => _debugQuery(sql.replace(/%/g, '%%'), txId);
 
 const { POOL_CONFIG_OPTIONS } = require('./constants');
 
+const ReuseConnection = (connection) =>
+  async function withConnection(next) {
+    return next(connection);
+  };
+
```
briandamaged: Oops -- I thought I deleted this part already.  I'll fix this in a little while.  (I currently have the tests running on loop so that we can detect unhandled promise rejections)

lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -19,6 +19,11 @@ const {
   chunk,
 } = require('lodash');
 
+const ReuseConnection = (connection) =>
+  async function(next) {
+    return next(connection);
+  };
+
```
briandamaged: FYI: This function is currently duplicated in a few places.  We can consolidate it as part of the re-implementation.

lib/transaction.js
```diff
@@ -163,79 +175,69 @@ class Transaction extends EventEmitter {
...
-          return null;
-        })
-        .catch((e) => {
-          return this._rejecter(e);
+    const { withConnection } = config;
+
+    const withEverything = pipe([
+      withConnection,
+      InjectTransactionID(this.txid),
+    ]);
```
briandamaged: I'm not sure if I'll keep the `pipe([...])` function in the re-implementation.  It's a functional programming concept that many developers might find mystifying.

briandamaged:
This PR builds upon #3707 , which makes the diffs a bit challenging to read.  So, here's a high-level overview of the concepts.

Originally, all of the `Transaction#acquireConnection(..)` implementations included some form of the following pattern:

```javascript
async acquireConnection(config, cb) {
  // Figure out how the connection should be obtained.  Specifically: did the caller
  // provide a connection, or does the Transaction need to procure one itself?
  const configConnection = config && config.connection;
  const connection = configConnection || (await this.client.acquireConnection());

  try {
    // Do other post-processing on the connection, some of which might be
    // implementation-specific
    // ....

    return await cb(connection);
  } finally {
    // Do other post-processing on the connection, some of which might be
    // implementation-specific
    // ....

    if(configConnection) {
      // Do nothing.  The caller passed in a connection explicitly, so it is their
      // responsibility to release it.
    } else {
      // This Transaction instance acquired the connection itself; therefore, it
      // is the Transaction's responsibility to release it.
      await this.client.releaseConnection(connection);
    }
  }
}
```

As you can see, the Transaction is doing quite a bit of work to figure out the caller's intentions!

With the new code, this logic has been reduced to:

```javascript
async acquireConnection(connection, cb) {
  try {
    // Do other post-processing on the connection, some of which might be
    // implementation-specific
    // ....

    return await cb(connection);
  } finally {
    // Do other post-processing on the connection, some of which might be
    // implementation-specific
    // ....
  }
}
```

As you can see, the Transaction no longer needs to figure out if it's responsible for acquiring/releasing the connection.  Instead, it can focus on the Dialect-specific details.

So, where does the connection come from, then?  Short answer: the Transaction delegates control to a `withConnection(..)` strategy function, which encapsulates the details.  For example:

```
// The `withConnection` function will be provided by the caller
const { withConnection } = config;

return withConnection(async (connection) {
  return this.acquireConnection(connection, async ()=> {
    // ...
  });
});
```

So, let's suppose that we're creating a top-level `Transaction` from a `knex` instance.  In this case, it's the `knex` instance's responsibility to acquire and release the connection.  So, its `withConnection(..)` implementation will look like this:

```
async function withConnection(cb) {
  const connection = await this.client.acquireConnection();
  try {
    return await cb(connection);
  } finally {
    await this.client.releaseConnection(connection);
  }
}
```

In contrast, let's suppose we're creating a nested `Transaction`.  In this case, we always want to reuse the existing connection.  So, the `withConnection(..)` implementation will look like this:

```
async function withConnection(cb) {
  return cb(outerTx.connection);
}
```

Notice that it doesn't release the connection.  That's because it is the caller's responsibility to release the connection.

Anyway, that's the gist of this refactoring.  There's still more work to be done, but it at least extracts this responsibility out of the `Transaction` class.

PR #3709, opened 2020-03-06
Updated .gitignore files to ignore testing artifacts

briandamaged:
Various testing artifacts get left on disk if you kill the tests while they're still running.  I updated the `.gitignore` files so that it would be less likely for these to get committed by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,10 @@ tmp
 bin/migrations-folder
 bin/db.sqlite3
 
+# Other artifacts created during testing
+/knexfile.js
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -16,6 +16,10 @@ tmp
 bin/migrations-folder
 bin/db.sqlite3
 
+# Other artifacts created during testing
+/knexfile.js
```
briandamaged: I used an absolute path here so that it only ignores this specific artifact.

Issue #3553, opened 2019-11-27
Postgres client not parsing node-native BigInt insert values


Bouncey:
# Environment

Knex version: 0.20.2
Database + version: Postgres 10.10
OS: Ubuntu

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do
Postgres client not parsing node-native `BigInt` insert values correctly, other clients can handle them correctly.

2. Error message
None

3. Reduced test code
```js
const knex = require('knex');

const pgBuilder = knex({ client: 'pg' });
const mySqlBuilder = knex({ client: 'mysql' });

const rows = [{ count: BigInt(Number.MAX_SAFE_INTEGER * 2) }];

function testBigIntParsing(builder) {
  const sql = builder('my_table')
    .insert(rows)
    .toString();

  console.log('SQL: ', sql);
  console.log('Rows: ', rows);
}


testBigIntParsing(pgBuilder);
testBigIntParsing(mySqlBuilder);
```
Output: 
```sh
SQL:  insert into "my_table" ("count") values ('')
Rows:  [ { count: 18014398509481982n } ]
SQL:  insert into `my_table` (`count`) values ('18014398509481982')
Rows:  [ { count: 18014398509481982n } ]
```

kibertoad:
Does the same work when you are using `pg` driver directly?

Bouncey:
If I use `pg` directly, the `BigInt(Number.MAX_SAFE_INTEGER * 2)` value gets added to the `count` column correctly.

elhigu:
Try `.toSQL()` or running real query with BigInt.  `.toString()` should be only used for debugging. However I suppose that could be fixed. IIRC pg driver indeed calls toString() of parameter binding automatically... or maybe it was some special function for conversion from object -> sql.

kibertoad:
@Bouncey Would you consider contributing PR with a failing integration test that passes with other dbs? Should be an easy fix from there.

Bouncey:
`.toSQL` does indeed parse the `BigInt` correctly.

I'll see what I can muster up with a PR

mercmobily:
Hi! Thanks for the clear repro — this comes down to using `toString()` vs `toSQL()` and how bindings are serialised.

[...]

PR #3710, opened 2020-03-07
add support for bigint

maximelkin:
todo: fix tests


.eslintrc.js
```diff
@@ -20,6 +20,9 @@ module.exports = {
     'no-fallthrough': warning,
     'require-atomic-updates': 0,
   },
+  globals: {
+    BigInt: true,
+  },
```
briandamaged: Do we need to officially drop support for node 8 first?

test/integration/builder/inserts.js
```diff
@@ -1302,5 +1302,17 @@ module.exports = function(knex) {
           });
       });
     });
+
+    describe('#3553 insert with bigint', function() {
+      if (typeof BigInt === 'undefined') return;
```
briandamaged: Hmm... does `"#3553 insert with bigint"` still get printed in the test case output?  If so, then we might want to move the conditional outside of the `describe(..)` block:

```
if(typeof BigInt !== 'undefined') {
  describe('#3553 insert with bigint', function() {
    //
  });
}
```

test/integration/builder/inserts.js
```diff
@@ -1302,5 +1302,17 @@ module.exports = function(knex) {
...
+
+      it('should allow insert with BigInt', async function() {
+        await knex('accounts').insert([{ id: BigInt(12) }]);
```
briandamaged: Might want to:

1. Test w/ a number that is really large.  (Ex:  `BigInt("1234567812345678123456781234567812345678")`)
2. Read the number back out of the database as well.

Also: it looks like the `MSSQL` changes the datatype depending upon the size of the integer.  So, there are at least 5 different scenarios that should be tested there.  (And, probably more like 7-9 cases if you count the boundary conditions as well).

test/integration/builder/inserts.js
```diff
@@ -1302,5 +1302,17 @@ module.exports = function(knex) {
...
+      });
+      it('should allow insert array with BigInt', async function() {
+        if (knex.client.driverName !== 'pg') return true;
+        await knex('accounts').insert([{ id: [BigInt(12)] }]);
```
briandamaged: It would probably be a good idea to use a larger number here as well.

test/integration/builder/inserts.js
```diff
@@ -1302,5 +1302,17 @@ module.exports = function(knex) {
...
     });
+
+    describe('#3553 insert with bigint', function() {
+      if (typeof BigInt === 'undefined') return;
+
+      it('should allow insert with BigInt', async function() {
+        await knex('accounts').insert([{ id: BigInt(12) }]);
+      });
+      it('should allow insert array with BigInt', async function() {
+        if (knex.client.driverName !== 'pg') return true;
```
briandamaged: Might want to do `return this.skip()` for now.

(Eventually, we should figure out a way to restructure the test cases so that these skipped scenarios get omitted.  But, marking them as `pending` will at least remind us to revisit this later)

test/integration/builder/updates.js
```diff
@@ -339,5 +339,36 @@ module.exports = function(knex) {
...
     });
+
+    describe('#3553 update with bigint', function() {
+      if (typeof BigInt === 'undefined') return;
+
+      it('should allow insert with BigInt', async function() {
+        const firstName = '#3553';
+        await knex('accounts').insert([{ balance: 5, first_name: firstName }]);
+        await knex('accounts').update([
+          { balance: BigInt(Number.MAX_SAFE_INTEGER) },
```
briandamaged: `Number.MAX_SAFE_INTEGER` is still a regular integer, correct?  If so, then you might want to use a larger number here.

briandamaged: Oh, and of course: read the number back out of the database as well.  (Just to make sure it didn't get truncated or anything like that)

test/integration/datatype/bigint.js
```diff
@@ -122,4 +122,33 @@ module.exports = function(knex) {
...
+      })
+      .then(function() {
+        return knex(tableName).insert({ value: value });
+      })
+      .then(function() {
+        return knex(tableName).first('value');
+      })
+      .then(function(response) {
+        expect(response.value).to.be.eql(value.toString());
+      });
```
briandamaged: You could use `async` / `await` syntax here if you wanted to.

test/unit/query/builder.js
```diff
@@ -132,7 +132,7 @@ function testquery(chain, valuesToCheck, selectedClients) {
     newChain.client = selectedClients[key];
     const sqlString = newChain.toQuery();
     const checkValue = valuesToCheck[key];
-    expect(checkValue).to.equal(sqlString);
+    expect(checkValue).to.equal(sqlString, `for driver ${key}`);
```
briandamaged: Unrelated enhancement?

test/unit/query/builder.js
```diff
@@ -9863,4 +9863,31 @@ describe('QueryBuilder', () => {
       }
     );
   });
+
+  it('#3553 should correctly compile insert with BigInt', function() {
+    testquery(
+      qb()
+        .insert([{ a: BigInt(12) }])
```
briandamaged: `BigInt("1234567182348761238746512348765")` just to make sure it's not coercing the values back into regular `Number` instances.

maximelkin:
@briandamaged Thanks for review! I will add test cases

test/integration/builder/inserts.js
maximelkin: I think this really not matters, because with node 8 drop this line will be removed

kibertoad:
Unless we absolutely want this in 0.20, we can mark it with `breaking change` label and drop Node 8 already.

maximelkin:
@kibertoad this merge is not breaking node 8 support
but i think its ok to merge this in 1.x version
(sorry for long inactivity, with this merge)

kibertoad:
@maximelkin Didn't it include specific checks for pre-10 Node, though?

maximelkin:
typeof something === 'bigint'
this is compatible with node 8

.eslintrc.js
maximelkin: fixed, now globals for test/ directory only

maximelkin:
todo: workaround for https://github.com/mapbox/node-sqlite3/issues/1058

maximelkin:
When selecting decimal from postgres, type of resulting value can be string or number, depending on size of data
This is because of old version pg-types for https://github.com/brianc/node-postgres

maximelkin:
mssql when selecting decimals without any settings give simple numbers, so this explodes

maximelkin:
God bless mysql drivers - the most simple configuration for getting strings for decimals and bigints

maximelkin:
On oracledb side probably we should always set fetchAsString and then parse to normal integer by passed metadata (numeric, int64, float64 -> string, less precision numbers to js native)

maximelkin:
mysql give errors because accounts.id declared as bigint autoincrement

maximelkin:
Maybe we should split merge request into fixing support of int64 and decimals for databases and support of js bigint?

kibertoad:
@maximelkin Yeah, split makes sense.

kibertoad:
@maximelkin Should we extract the working part and merge that for now?

maximelkin:
I think this pr will be adding support for querybuilding with .toString() only, but not sure is this useful because some libs dont support bigint yet

.eslintrc.js
kibertoad: Support for Node 8 was dropped already, so this is good to go.

kibertoad:
@maximelkin I would recommend pulling from master for this PR for vastly faster CI :)

Issue #3686, opened 2020-02-25
Refactor tests


kibertoad:
1. We should consider retiring chai.should() in favor of expect(..). (Why? Because chai.should() monkey-patches various Javascript objects, which can occasionally lead to some surprising results);

2. Se should consider eliminating global variables (e. g. `global.expect`) that are referenced throughout various test cases.

3. We should be using a sinon sandbox rather than patching the prototype ourselves. This will help guarantee that the prototype is restored properly once the test case concludes. (see test/integration/builder/additional.js -> `const knexPrototype = Object.getPrototypeOf(knexDb.client)` etc)

Originally raised in https://github.com/knex/knex/pull/3682

mercmobily:
Hi! Thanks for opening this housekeeping item. Here is a quick checklist of the test clean‑ups and where we are right now.

- [x] Retire `chai.should()` in favour of `expect(...)` across tests
- [x] Eliminate implicit globals used across suites (e.g. `global.expect`, `global.d`) — `global.d` removed in #3747; `expect` is imported locally
- [ ] Use a sinon sandbox consistently rather than patching prototypes directly
  - [ ] Replace manual prototype patching in `test/integration2/query/misc/additional.spec.js` (`const knexPrototype = Object.getPrototypeOf(knexDb.client) ...`) with `sinon.createSandbox()` and proper restore
  - [ ] Audit remaining tests for direct prototype patching and convert to sandbox + restore

Additional improvements (from a quick audit):

- [ ] Standardise setup/teardown helpers and per‑dialect test utilities
- [ ] Group and label integration tests by dialect to reduce cross‑dialect assumptions
- [ ] Reduce flakiness by isolating long‑running or timeout‑sensitive cases

Progress so far: #3712 (merged) removed several implicit globals; #3747 (merged) removed `global.d` and localised variables where needed.

PR #3712, opened 2020-03-08
Removed some globals from tests

briandamaged:
Removed a few of the globals from the test cases.  This partially addresses one of the items mentioned in #3686

### DIFF
--- a/test/chai-setup.js
+++ b/test/chai-setup.js
@@ -3,5 +3,3 @@ const chai = require('chai');
 chai.should();
 chai.use(require('chai-as-promised'));
 chai.use(require('sinon-chai'));
-
-global.expect = chai.expect;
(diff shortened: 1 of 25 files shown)


briandamaged:
As far as I can tell,  `global.d` is the last global.  Unfortunately, this one looks like it's going to be a bit more challenging to remove.  (There appears to be some tight coupling between the test cases and some testing-related functions.)

kibertoad:
Thanks for the PR, really cheered me up while I'm wrestling with Perl here 😂

briandamaged:
Ouch -- Perl!  In that case... maybe I should clean a few other things up as well.  lol

briandamaged:
k -- looks like the build succeeded.  Any hesitations before we merge it?

kibertoad:
Nope, looks beautiful.

PR #3716, opened 2020-03-09
Cleanup/remove transaction stalling

briandamaged:
This removes the `await Process.resolve()` work-around from the Transaction initialization logic.

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -45,6 +45,10 @@ class Transaction extends EventEmitter {
       outerTx ? 'nested' : 'top level'
     );
 
+    // `this` can potentially serve as an `outerTx` for another
+    // Transaction.  So, go ahead and establish `_lastChild` now.
+    this._lastChild = Promise.resolve();
+
     // FYI: As you will see in a moment, this Promise will be used to construct
     //      2 separate Promise Chains.  This ensures that each Promise Chain
     //      can establish its error-handling semantics without interfering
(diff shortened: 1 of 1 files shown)


briandamaged:
This brings us another step closer to cleaning up some of the `outerTx` logic.  I'm still trying to figure out a good way to restructure the logic so that an `outerTx` is _always_ passed into a `Transaction`.  (Which would eliminate the need for some of the conditional logic in the code)

lib/transaction.js
```diff
@@ -45,6 +46,12 @@ class Transaction extends EventEmitter {
       outerTx ? 'nested' : 'top level'
     );
 
+    // `this` can potentially serve as an `outerTx` for another
+    // Transaction.  So, go ahead and establish `_lastChild` now.
+    this._lastChild = Promise.resolve();
```
kibertoad: When do we overwrite this?

briandamaged: It gets overwritten towards the bottom of each child Transaction's constructor:

```
outerTx._lastChild = basePromise.catch(() => {});
```

BTW: I think I'm gonna mark this PR as `WIP:` for now.  I'm still thinking about extracting the `outerTx` thing a bit further.

briandamaged:
I went ahead and removed the `WIP:` prefix.  It looks like cleaning up some of the other logic around `outerTx` will end up increasing the scope quite a bit; so, I'd prefer to handle that in a separate PR.

lib/transaction.js
```diff
@@ -37,6 +37,7 @@ class Transaction extends EventEmitter {
     this.logger = client.logger;
     this.outerTx = outerTx;
     this.trxClient = undefined;
+    this._completed = false;
```
briandamaged: Moved this line to earlier in the constructor just to be on the safe side.  (Since the Promise logic is no longer being stalled in the same manner)

lib/transaction.js
```diff
@@ -155,14 +155,6 @@ class Transaction extends EventEmitter {
   }
 
   async _evaluateContainer(config, container) {
-    // FYI: This is temporarily stalling things so that the constructor
-    //      can finish initializing the Transaction.  Otherwise,
-    //      `this.previousSibling` will still be `undefined`.
-    await Promise.resolve();
-
-    // Wait for the earlier Transactions to complete before proceeding.
-    await this._previousSibling;
```
briandamaged: This step has been moved into the constructor.

PR #3717, opened 2020-03-10
Fix/method binding on knex proxy

briandamaged:
The `make-knex.js` module was taking a few shortcuts when creating the `knex` proxy.  As a result, most (all?) of the methods within the `context` object were bound to `knex` instead of `context`.  This would then cause reference errors when one `context` method attempted to call another `context` method.

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -255,7 +255,7 @@ finallyMixin(Transaction.prototype);
 function makeTransactor(trx, connection, trxClient) {
   const transactor = makeKnex(trxClient);
 
-  transactor.withUserParams = () => {
+  transactor.context.withUserParams = () => {
     throw new Error(
       'Cannot set user params on a transaction - it can only inherit params from main knex instance'
     );
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -271,14 +271,9 @@ function makeTransactor(trx, connection, trxClient) {
       options.doNotRejectOnRollback = true;
     }
 
-    if (container) {
-      return trxClient.transaction(container, options, trx);
-    } else {
-      return new Promise((resolve, _reject) => {
-        trxClient.transaction(resolve, options, trx).catch(_reject);
-      });
-    }
+    return this._transaction(container, options, trx);
   };
+
   transactor.savepoint = function(container, options) {
     return transactor.transaction(container, options);
   };
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -264,7 +264,7 @@ function makeTransactor(trx, connection, trxClient) {
   transactor.isTransaction = true;
   transactor.userParams = trx.userParams || {};
 
-  transactor.transaction = function(container, options) {
+  transactor.context.transaction = function(container, options) {
     if (!options) {
       options = { doNotRejectOnRollback: true };
     } else if (isUndefined(options.doNotRejectOnRollback)) {
--- a/lib/util/make-knex.js
+++ b/lib/util/make-knex.js
@@ -38,15 +38,14 @@ function initContext(knexFn) {
     // when transaction is ready to be used.
     transaction(container, _config) {
       const config = Object.assign({}, _config);
-      config.userParams = this.userParams || {}
-      if(isUndefined(config.doNotRejectOnRollback)) {
+      config.userParams = this.userParams || {};
+      if (isUndefined(config.doNotRejectOnRollback)) {
         // Backwards-compatibility: default value changes depending upon
         // whether or not a `container` was provided.
         config.doNotRejectOnRollback = !container;
       }
 
-
-      if(container) {
+      if (container) {
         const trx = this.client.transaction(container, config);
         return trx;
       } else {
@@ -145,6 +144,26 @@ function redefineProperties(knex, client) {
     };
   });
 
+  const METHOD_PROXIES = [
+    'raw',
+    'batchInsert',
+    'transaction',
+    'transactionProvider',
+    'initialize',
+    'destroy',
+    'ref',
+    'withUserParams',
+    'queryBuilder',
+    'disableProcessing',
+    'enableProcessing',
+  ];
+
+  for (const m of METHOD_PROXIES) {
+    knex[m] = function(...args) {
+      return knex.context[m](...args);
+    };
+  }
+
   Object.defineProperties(knex, {
     context: {
       get() {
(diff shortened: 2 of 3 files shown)


lib/transaction.js
```diff
@@ -264,7 +264,7 @@ function makeTransactor(trx, connection, trxClient) {
   transactor.isTransaction = true;
   transactor.userParams = trx.userParams || {};
 
-  transactor.transaction = function(container, options) {
+  transactor.context.transaction = function(container, options) {
```
briandamaged: It looks like this line was only working because of the way the proxy shortcuts had been setup.  In other words: one bug was canceling another bug.

So, now that the proxies are being setup correctly, we need to make sure we override the method in the correct place.  (on the `context` object rather than the `knex` function)

lib/util/make-knex.js
```diff
@@ -7,6 +7,80 @@ const QueryInterface = require('../query/methods');
...
+    configurable: true,
+  },
+
+  fn: {
+    get() {
+      return new FunctionHelper(this.client);
+    },
+    configurable: true,
+  },
+};
```
briandamaged: I replaced all of the references to `knex` with references to `this` instead.  This means that the property definitions can now be identical for every `knex` instance.  Therefore, I moved the property definitions into the top-level module so that they could be reused.  (Rather than creating a new instance of the definitions for each `knex` instance)

lib/util/make-knex.js
```diff
@@ -140,83 +213,18 @@ function redefineProperties(knex, client) {
...
-      get() {
-        return new Seeder(knex);
-      },
-      configurable: true,
-    },
+  for (const m of CONTEXT_METHODS) {
+    knex[m] = function(...args) {
+      return this.context[m](...args);
+    };
+  }
```
briandamaged: This was really the key change.  Previously, these methods were not actually being proxied.  Instead, they would be injected directly into the `knex` function.  For example:

```
knex.transaction = context.transaction;
knex.raw = context.raw
```

The problem: this meant that the methods would be invoked through `knex` now instead of `context`.  Thus, `this` would point to `knex` instead of `context`.  As a result, it was impossible to define methods in `context` such as:

```
const context = {
  transaction(container, _options) {
    // Do some implementation-specific stuff
    const outerTx = createOuterTxForTopLevelTransactions();

    // This line would fail because `this` is pointing to `knex` instead of `context`!
    return this._transaction(container, options, outerTx);
  },

  _transaction(container, options, outerTx) {
    // Do stuff that is common across all implementations...
  }
}
```

test/integration/builder/additional.js
```diff
@@ -1089,7 +1089,7 @@ module.exports = function(knex) {
       });
       it('should capture stack trace on raw query', () => {
         return knex.raw('select * from some_nonexisten_table').catch((err) => {
-          expect(err.stack.split('\n')[2]).to.match(/at Function\.raw \(/); // the index 2 might need adjustment if the code is refactored
+          expect(err.stack.split('\n')[2]).to.match(/at Object\.raw \(/); // the index 2 might need adjustment if the code is refactored
```
briandamaged: Previously, the stacktrace said `Function` here because the underlying method was not actually being proxied.  Instead, it was being invoked through `knex`, which was a Function.

lib/util/make-knex.js
```diff
@@ -38,20 +124,25 @@ function initContext(knexFn) {
...
+      if (container) {
+        const trx = this.client.transaction(container, config, outerTx);
         return trx;
       } else {
         return new Promise((resolve, reject) => {
-          const trx = this.client.transaction(resolve, config);
+          const trx = this.client.transaction(resolve, config, outerTx);
           trx.catch(reject);
         });
       }
```
briandamaged: I refactored the `transaction(..)` method just to demonstrate that the `this` pointer was referencing the correct object now.  So, this allows `this.transaction(..)` to call `this._transaction(..)` without encountering any reference errors.

lib/util/make-knex.js
```diff
@@ -138,89 +229,46 @@ function _copyEventListeners(eventName, sourceKnex, targetKnex) {
...
+  //       allow external components to directly manipulate `knex`
+  //       data structures?  Or, should we come up w/ a different
+  //       approach that avoids side-effects / mutation?
+  //
+  //      (FYI: I noticed this issue because I attempted to integrate
+  //       this logic directly into the `KNEX_PROPERTY_DEFINITIONS`
+  //       construction.  However, `KNEX_PROPERTY_DEFINITIONS` is
+  //       constructed before any `knex` instances are created.
+  //       As a result, the method extensions were missing from all
+  //       `knex` instances.)
```
briandamaged: I decided to omit this part from the PR until we have a chance to discuss it.

On one hand, we could probably work around this by using an actual `Proxy(..)`.  On the other hand, it seems strange that we're allowing `knex` data / structures to be directly manipulated in this way.  Perhaps we can come up w/ some way to preserve the original `QueryBuilder`, yet also allow end-users to create/use custom `QueryBuilder`s?

lib/transaction.js
```diff
@@ -264,21 +264,16 @@ function makeTransactor(trx, connection, trxClient) {
...
     }
 
-    if (container) {
-      return trxClient.transaction(container, options, trx);
-    } else {
-      return new Promise((resolve, _reject) => {
-        trxClient.transaction(resolve, options, trx).catch(_reject);
-      });
-    }
+    return this._transaction(container, options, trx);
```
briandamaged: `this` correctly refers to the `transaction.context` now, which means that this method call works now.  Hooray!

briandamaged:
In the near-term, this PR just fixes the `this` pointer within the `context` methods.

Longer-term, this also allows us to treat `context` like a regular Object or Class.  Most importantly, this means that we can subclass it rather than patching its methods on the fly.  For example:

```
// Default implementation
class Context {
  constructor() {
    // ...
  }

  transaction(container, config) {
    // ...
  }
}

// Implementation used specifically by Transactions
class TransactionContext extends Context {
  constructor({outerTx}) {
    super();
    this.outerTx = outerTx;
  }

  transaction(container, options) {
    return this._transaction(container, options, this.outerTx);
  }
}
```

Of course, there are pros and cons w/ this approach.  But, the possibility will at least be open to us now.

kibertoad:
@briandamaged The reason why I'm concerned about using actual Proxy here is performance. See this: https://thecodebarbarian.com/thoughts-on-es6-proxies-performance

briandamaged:
@kibertoad : Agreed about the `Proxy` performance.  That's the main reason I decided to omit that part of the refactoring.

briandamaged:
@maximelkin : do you also want a chance to look at this PR before it gets merged?  Or, should we go ahead and merge it?

maximelkin:
Go ahead, if this covered by tests

briandamaged:
@maximelkin : Thx for reviewing it!

briandamaged:
Oh -- and thx @kibertoad as well!

PR #3722, opened 2020-03-11
WIP: Feature/connection leasing

briandamaged:
Redesigns the `Transaction` logic so that the caller is ultimately responsible for acquiring/releasing the connection.

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -12,6 +12,8 @@ const debug = Debug('knex:tx');
 
 const { uniqueId, isUndefined } = require('lodash');
 
+const { ReuseConnection } = require('./util/connection-leasing');
+
 // FYI: This is defined as a function instead of a constant so that
 //      each Transactor can have its own copy of the default config.
 //      This will minimize the impact of bugs that might be introduced
@@ -30,6 +32,7 @@ class Transaction extends EventEmitter {
     super();
     this.userParams = config.userParams;
     this.doNotRejectOnRollback = config.doNotRejectOnRollback;
+    const withConnection = config.withConnection;
 
     const txid = (this.txid = uniqueId('trx'));
 
@@ -46,30 +49,9 @@ class Transaction extends EventEmitter {
       outerTx ? 'nested' : 'top level'
     );
 
-    // `this` can potentially serve as an `outerTx` for another
-    // Transaction.  So, go ahead and establish `_lastChild` now.
-    this._lastChild = Promise.resolve();
-
-    const _previousSibling = outerTx ? outerTx._lastChild : Promise.resolve();
-
-    // FYI: As you will see in a moment, this Promise will be used to construct
-    //      2 separate Promise Chains.  This ensures that each Promise Chain
-    //      can establish its error-handling semantics without interfering
-    //      with the other Promise Chain.
-    const basePromise = _previousSibling.then(() =>
-      this._evaluateContainer(config, container)
+    this._promise = withConnection((conn) =>
+      this._evaluateContainer(conn, container)
     );
-
-    // FYI: This is the Promise Chain for EXTERNAL use.  It ensures that the
-    //      caller must handle any exceptions that result from `basePromise`.
-    this._promise = basePromise.then((x) => x);
-
-    if (outerTx) {
-      // FYI: This is the Promise Chain for INTERNAL use.  It serves as a signal
-      //      for when the next sibling should begin its execution.  Therefore,
-      //      exceptions are caught and ignored.
-      outerTx._lastChild = basePromise.catch(() => {});
-    }
   }
 
   isCompleted() {
@@ -154,8 +136,11 @@ class Transaction extends EventEmitter {
     return this;
   }
 
-  async _evaluateContainer(config, container) {
-    return this.acquireConnection(config, (connection) => {
+  async _evaluateContainer(conn, container) {
+    // This will be used when creating nested Transactions
+    this.reuseConnection = ReuseConnection(conn);
+
+    return this.acquireConnection(conn, (connection) => {
       const trxClient = (this.trxClient = makeTxClient(
         this,
         this.client,
@@ -207,22 +192,9 @@ class Transaction extends EventEmitter {
   // Acquire a connection and create a disposer - either using the one passed
   // via config or getting one off the client. The disposer will be called once
   // the original promise is marked completed.
-  async acquireConnection(config, cb) {
-    const configConnection = config && config.connection;
-    const connection =
-      configConnection || (await this.client.acquireConnection());
-
-    try {
-      connection.__knexTxId = this.txid;
-      return await cb(connection);
-    } finally {
-      if (!configConnection) {
-        debug('%s: releasing connection', this.txid);
-        this.client.releaseConnection(connection);
-      } else {
-        debug('%s: not releasing external connection', this.txid);
-      }
-    }
+  async acquireConnection(connection, cb) {
+    connection.__knexTxId = this.txid;
+    return await cb(connection);
   }
 
   then(onResolve, onReject) {
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -256,11 +256,15 @@ function makeTransactor(trx, connection, trxClient) {
   transactor.isTransaction = true;
   transactor.userParams = trx.userParams || {};
 
-  transactor.context.transaction = function(container, options) {
-    if (!options) {
-      options = { doNotRejectOnRollback: true };
-    } else if (isUndefined(options.doNotRejectOnRollback)) {
+  transactor.context.transaction = function(container, _config) {
+    const options = Object.assign({}, _config);
+
+    if (isUndefined(options.doNotRejectOnRollback)) {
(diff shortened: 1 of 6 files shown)


kibertoad:
@briandamaged @maximelkin I suggest we create 0.20 branch for potential hotfixes a d start merging all the breakung changes such as dropping node 8 into master in preparation for 1.0.0

briandamaged:
@kibertoad : I think that's a good idea.

kibertoad:
Done!

lib/util/connection-leasing.js
```diff
@@ -0,0 +1,26 @@
...
+  function reuseConnection(next) {
+    const basePromise = previous.then(() => next(connection));
+    previous = basePromise.catch((err) => {});
+    return basePromise.then((x) => x);
+  }
+
+  return reuseConnection;
+}
+
+Object.assign(exports, {
```
kibertoad: Why Object.assign and not module.exports = {}?

lib/util/connection-leasing.js
```diff
@@ -0,0 +1,26 @@
...
+  async function withConnection(next) {
+    const connection = await client.acquireConnection();
+    try {
+      return await next(connection);
+    } finally {
+      await client.releaseConnection(connection);
+    }
+  };
+
+function ReuseConnection(connection) {
```
kibertoad: PascalCase looks confusing here; this doesn't look like a class or class-like function. Would it be possible to clarify naming a bit, to make it more self-explanatory what functions in this file do?

briandamaged: Ah -- I tend to use PascalCase for factory functions as well.  In this case: it's a function that returns a connection leasing function.

lib/transaction.js
```diff
@@ -256,13 +228,17 @@ function makeTransactor(trx, connection, trxClient) {
   transactor.isTransaction = true;
   transactor.userParams = trx.userParams || {};
 
-  transactor.context.transaction = function(container, options) {
-    if (!options) {
-      options = { doNotRejectOnRollback: true };
-    } else if (isUndefined(options.doNotRejectOnRollback)) {
+  transactor.context.transaction = function(container, _config) {
+    const options = Object.assign({}, _config);
```
kibertoad: Since we are using fairly new Node version, `const options = {..._config}` seems to be a simpler way.

lib/util/connection-leasing.js
briandamaged: `Object.assign(..)` tends to behave slightly better when you have circular dependencies.  But yeah: you really shouldn't have circular dependencies in the first place.  lol

lib/util/connection-leasing.js
kibertoad: If it's a factory function, then probably it should be a noun and not a verb.

kibertoad: However, since this is a factory function for making functions... Maybe naming should explain that fact somehow?

briandamaged: Makes sense.  I'll take care of that in a little bit.

I'm also gonna post a "thinking out loud" comment in a little bit to clarify some of the other things that seem to relate to all of this.  (Ex: connection pooling, etc)

briandamaged:
Okay... thinking out loud ™️  :

Earlier today, I was following up to a question here:

https://github.com/knex/knex/issues/756#issuecomment-597617787

This got me thinking: a connection that is used in this manner won't be subject to the same `_previousSibling` guarantees.  So, if somebody attempted to run several queries at once using this connection, then things would probably catch on fire.  🔥 

Saying it another way: there is logic in place right now that guarantees that nested `Transactions` will run in the correct order.  But, this logic breaks down if you do something like this:

```javascript
await knex.transaction(async function(trx) {
  const connection = await trx.client.acquireConnection();
  try {
    await Promise.all([
      trx.transaction(..),
      doSomethingWith(connection),  // Uh oh...
      trx.transaction(..),
      doSomethingElseWith(connection),  // Uh oh...
    ]);
  } finally {
    await trx.client.releaseConnection(connection);
  }
});
```

The lines marked with `Uh oh...` will attempt to run immediately, which will interfere with the nested Transactions.  Oops!

Alright, here's a quick inductive leap: what if Transactions didn't have special "connection ordering guarantees" (ie: the `_previousSibling` logic)?  What if instead Transactions just had a connection pool with exactly 1 connection in it?  In this case, it seems like we'd have a better way to prevent multiple pieces of code from using the same connection concurrently.  Ex:

```javascript
await knex.transaction(async function(trx) {
  await Promise.all([
    trx.transaction(..),
    trx.leaseConnection((connection)=> doSomethingWith(connection)),
    trx.transaction(..),
    trx.leaseConnection((connection)=> doSomethingElseWith(connection)),
  ]);
});
```

And, just as importantly: this would also work for cases where a connection was obtained/released via `trx.acquireConnection()` / `trx.releaseConnection()`

So, you might notice that the `Transaction` class no longer contains the "connection ordering guarantee" stuffs (ie: `_previousSibling` logic).  Instead, this logic has been encapsulated into `ReuseConnection(..)`, which is temporarily simulating a pool with exactly 1 connection.

Anyway, I'll probably investigate this a bit more tomorrow.  Specifically: would it be possible to reuse `tarn` to create new pools with exactly 1 connection?  This would potentially save us from having to reinvent certain wheels.  But, it's also not clear how we'd want to configure the various timeouts for temporary connection pool.

cc: @kibertoad , @maximelkin

maximelkin:
Looks nice
1) need to be checked any logic in dbms about rollback transaction after error
2) Just edge case for tests
```js
const promise = knex.transaction(async (trx) => {
trx.leaseConnection(conn => delay(1000).then(() => conn.query('select 1')));
});
// promise should be resolved after conn.query('select 1')
```
3) so basically .leaseConnection is nested .transaction, but without guaranties of rollback on error?

maximelkin:
2) not works, because you can't use connection after finish of transaction callback block

briandamaged:
@maximelkin 

In terms of 3: Not exactly.  The `.leaseConnection(..)` method would be equivalent to:

```
// Assume that this method is defined on a Client instance for now
async leaseConnection(next) {
  const connection = await this.acquireConnection();
  try {
    return await next(connection);
  } finally {
    await this.releaseConnection(connection);
  }
}
```

So, it procures a `connection` for the duration of `next(connection)`.

This is _sorta_ what the code was already doing.  However, take a look at this:

https://github.com/knex/knex/blob/37d9c3034743705782dce42355abd3df4f0ebf84/lib/transaction.js#L341-L346

As you can see, the `trxClient` is defining fake `acquireConnection(..)` / `releaseConnection(..)` methods.  The main problem: the `connection` is never actually removed from a "pool" of any kind!  So, it's possible for several consumers to all acquire the same connection at the same time:

```javascript
await knex.transaction(async function(trx) {
  async function doStuff() {
    const conn = trx.client.acquireConnection();
    try {
      await blah(conn);
    } finally {
      await trx.client.releaseConnection(conn);
    }
  }

  // Both calls to doStuff() will acquire the same connection at the same time!
  return Promise.all([
    doStuff(),
    doStuff(),
  ]);
});
```

So, the idea is to create a new Pool for the Transaction.  But, the pool has the following characteristics

1. The pool contains exactly 1 connection.
2. The pool itself _does not_ create the connection.
3. Instead, the connection is "borrowed" (or "leased") from an external source.  (Ex: the parent Transaction)

Here's an analogy: let's say you work for a Software Company.  You can think of a Software Company as a pool of "Software Engineers".   Now, let's say there's a large project that will require 3 dedicated Software Engineers.  The Software Company won't hire 3 new Engineers just for this project.  Instead, it will allocate 3 of its Engineers to create a new Team (ie: a pool within a pool!) to focus on this project.

And, I think we can apply the same concept to Knex's Transaction logic.  Whenever a new Transaction is started, the caller will allocate 1 connection to focus on the Transaction.  Likewise, if there is a nested Transaction, then the Parent Transaction will allocate its one-and-only connection to each of its Child Transactions.  Consequently, this will then guarantee the same transaction ordering behaviors that we used to achieve via the `_previousSibling` logic.

lib/transaction.js
```diff
@@ -207,22 +192,9 @@ class Transaction extends EventEmitter {
...
-      return await cb(connection);
-    } finally {
-      if (!configConnection) {
-        debug('%s: releasing connection', this.txid);
-        this.client.releaseConnection(connection);
-      } else {
-        debug('%s: not releasing external connection', this.txid);
-      }
-    }
+  async acquireConnection(connection, cb) {
```
maximelkin: This method probably should be renamed, because its not about acquiring connection anymore

briandamaged: Yeah -- I agree that the method should be renamed.  The question is whether or not we want to maintain backwards-compatibility.  In particular: it appears that several developers have worked around limitations in `knex` by calling `acquireConnection(..)` directly.  So, renaming the method would break all of those work-arounds.

At the moment, I'm continuing w/ some refactoring using the current names.  At the end, I might create some method aliases / etc. so that we can maintain as much backwards-compatibility as possible.  (And, probably have the aliases log a deprecation warning of some kind)

PR #3724, opened 2020-03-12
Typescript: Makes the ChainableInterface conform to Promise<T>

flovilmart:
This PR addresses a problem that is often raised in Typescript as the ChainableInterface does not conform to `Promise<T>`.

It add the missing `Symbol.toStringTag` set to `Promise(Target.name)` (so the Builder would be) `Promise(Builder)`.

I wasn't sure of the signature, for the extend, as `finally` may not be there and it doesn't seem to be polyfilled. 

Was the `Pick<K>` being defensive as `interface.js` does not clone the `Promise` prototype?

### DIFF
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -93,7 +93,7 @@ module.exports = function(Target) {
   };
 
   Object.defineProperty(Target.prototype, Symbol.toStringTag, {
-    get: () => `Promise(${Target.name})`
+    get: () => 'object',
   });
 
   finallyMixin(Target.prototype);
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -92,5 +92,9 @@ module.exports = function(Target) {
     return this.then().catch(onReject);
   };
 
+  Object.defineProperty(Target.prototype, Symbol.toStringTag, {
+    get: () => `Promise(${Target.name})`
+  });
+
   finallyMixin(Target.prototype);
 };
--- a/test/unit/interface.js
+++ b/test/unit/interface.js
@@ -31,7 +31,7 @@ describe('interface', function() {
 
     _interface(SomeClass);
     const fakeInstance = new SomeClass();
-    chai.expect(fakeInstance[Symbol.toStringTag]).to.eq('Promise(SomeClass)');
+    chai.expect(fakeInstance[Symbol.toStringTag]).to.eq('object');
     fakeInstance._asyncStack = ['line1', 'line2', 'line3'];
     fakeInstance.then();
   });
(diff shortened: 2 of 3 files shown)


lorefnon:
> `ChainableInterface<T = any> extends Promise<T>`

This is not desirable because if the global Promise type has been mutated then we would end up advertising that ChainableInterface has methods that it actually doesn't. 

We had discussed this before in https://github.com/knex/knex/issues/3422. 

Does adding `typeof Symbol.toStringTag` to `ExposedPromiseKeys` not solve your problem here ? If so, can you please provide an example of a problem that this PR is intended to solve ?

flovilmart:
@lorefnon this makes total sense.

Will try with your suggestion

flovilmart:
After looking, adding `typeof Symbol.toStringTag` to `ExposedPromiseKeys` does not work as `typeof Symbol.toStringTag` resolves as a `symbol` which is not what we want. I updated the PR by splitting the ChainableInterface declaration in 2 parts.

1.  the Promise<T> conformance
2. the additional methods available on ChainableInterface<T>.

This way it fixes properly the issue for me.

lorefnon:
I don't quite understand why the splitting is necessary. After automatic merging, won't it be equivalent to: 

```ts
  interface ChainableInterface<T = any> extends Pick<Promise<T>, keyof Promise<T> & ExposedPromiseKeys> {
    readonly [Symbol.toStringTag]: string;
    
    toQuery(): string;
    options(options: Readonly<{ [key: string]: any }>): this;
    connection(connection: any): this;
   ...
}
```

flovilmart:
It’s for readability. The 1st part of the declaration is everything Promise, the second part is all the specifics. Happy to put it all together if you prefer.

Would you want the extra property at the top or the bottom of the interface declaration?

lorefnon:
Oh ok. Sorry, I misunderstood the intent. 

It is understandable to keep it in its own interface but I don't think taking advantage of interface merging is the right approach because that was primarily designed to support cases like cross-library monkey-patching. 

I suggest doing something like: 

```ts
interface StringTagSupport {
    readonly [Symbol.toStringTag]: string
}

interface ChainableInterface<T = any> extends StringTagSupport, ... 
```

And then we can use that interface in other places where we add `Symbol.toStringTag` member to customize how `toString` behaves.

flovilmart:
Thanks! I updated the PR with the suggestions.

lorefnon:
Thank you. Changes look good to me.

flovilmart:
@lorefnon feel free to merge :) I can't

kibertoad:
Released in 0.20.13

flovilmart:
Thanks @kibertoad

PR #3730, opened 2020-03-15
typings: support Raw types for insert, where, update

maximelkin:
Closes https://github.com/knex/knex/issues/3652

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -417,7 +417,7 @@ declare namespace Knex {
 
   type DbRecord<TRecord> = Readonly<SafePartial<MaybeRawRecord<TRecord>>>;
 
-  type DbRecordSet<TRecord> = Readonly<MaybeArray<DbRecord<TRecord>>>;
+  type DbRecordArr<TRecord> = Readonly<MaybeArray<DbRecord<TRecord>>>;
 
   //
   // QueryInterface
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -678,8 +688,8 @@ declare namespace Knex {
       >[]
     >(
       columnName: K1,
-      value: TRecord[K1],
-      returning: K1
+      value: DbColumn<TRecord[K1]>,
+      returning: K2
```
maximelkin: side fix

types/index.d.ts
```diff
@@ -691,12 +701,12 @@ declare namespace Knex {
       >[]
     >(
       columnName: K1,
-      value: TRecord[K1],
-      returning: readonly K1[]
+      value: DbColumn<TRecord[K1]>,
+      returning: readonly K2[]
```
maximelkin: side fix

maximelkin:
@kibertoad this can be merged, since this is not breaking change

lorefnon:
@maximelkin Thanks for this PR. The changes look good :) 

The only minor comment I have is on the naming: `DbRecordSet` -> `DbRecordArr` to convey that it is not actually a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set).

maximelkin:
Set in name means not data structure, but some amount of items
And this is not always array

maximelkin:
also, what do you think about renaming DbColumn to DbAttribute?

kibertoad:
@lorefnon Is this good enough to be merged already?

lorefnon:
Yeah, I think the current set of changes look good. 

I don't have a preference around DbColumn vs DbAttribute.

Issue #3721, opened 2020-03-11
oracledb's Oracle_Transaction.acquireConnection ignores external connection passed in config


kenbellows:
Hey there! I'm writing an API using knex that queries an OracleDB instance in a way that requires me to manually interact with a `node-oracledb` connection object before kicking off a transaction. I was happy to discover that I could manually pass in a connection object that I had modified using `const trx = knex.transaction(undefined, { connection })`, which solves my problem perfectly.

However, I found that the connection object I was passing in did not seem to be the one making the db calls. For example, `node-oracledb` exposes a magic setter on `connection` objects called `connection.clientId` that sets `SYS_CONTEXT('userenv', 'client_identifier')` in the db. So I tried this test code:

```js
const knex = require('knex')({
  client: 'oracledb',
  connection: { /* ... connection info ... */ }
})

(async () => {
  const connection = await knex.client.acquireConnection()
  connection.clientId = 'testuser'
  console.log(await knex.transaction(async trx => {
    return trx.raw(`SELECT SYS_CONTEXT('userenv', 'client_identifier') FROM DUAL`)
  }, { connection })
})()
```
My expected output, what I get when I use the `node-oracledb` library directly, is:
```js
[ { "SYS_CONTEXT('userenv', 'client_identifier')": "testuser" } ]
```
My actual output was:
```js
[ { "SYS_CONTEXT('userenv', 'client_identifier')": null } ]
```
which indicates that the connection object used to make the query has no `clientId` set.

I did some digging through Knex and found the problem. There's a small bug in `lib/dialects/oracledb/transaction.js` in [the `acquireConnection()` function](https://github.com/knex/knex/blob/master/lib/dialects/oracledb/transaction.js#L53-L76):
```js
 async acquireConnection(config, cb) {
    const configConnection = config && config.connection;

    const connection = await this.client.acquireConnection();
    try {
      connection.__knexTxId = this.txid;
      connection.isTransaction = true;
      return await cb(connection);
    } finally {
      debugTx('%s: releasing connection', this.txid);
      connection.isTransaction = false;
      try {
        await connection.commitAsync();
      } catch (err) {
        this._rejecter(err);
      } finally {
        if (!configConnection) {
          await this.client.releaseConnection(connection);
        } else {
          debugTx('%s: not releasing external connection', this.txid);
        }
      }
    }
  }
```

Note the first few lines:
```js
const configConnection = config && config.connection;
const connection = await this.client.acquireConnection();
```
[...]

PR #3731, opened 2020-03-15
Fix: Oracle_Transaction#acquireConnection(..) can use config.connection

briandamaged:
As noted in #3721 , OracleDB's `Transaction#acquireConnection(..)` implementation was not using `config.connection` when it was passed in.  So, this PR addresses that bug.

### DIFF
--- a/test/integration/builder/transaction.js
+++ b/test/integration/builder/transaction.js
@@ -749,4 +749,41 @@ module.exports = function(knex) {
       process.removeListener('unhandledRejection', fn);
     }
   });
+
+  context('when a `connection` is passed in explicitly', function() {
+    beforeEach(function() {
+      this.sandbox = sinon.createSandbox();
+    });
+
+    afterEach(function() {
+      this.sandbox.restore();
+    });
+
+    it('assumes the caller will release the connection', async function() {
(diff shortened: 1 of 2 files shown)


briandamaged:
@kibertoad : Which branch should we be using for `0.20.x` fixes?  (This PR is currently merging into `master`)

test/integration/builder/transaction.js
```diff
@@ -749,4 +749,41 @@ module.exports = function(knex) {
...
+  context('when a `connection` is passed in explicitly', function() {
+    beforeEach(function() {
+      this.sandbox = sinon.createSandbox();
+    });
+
+    afterEach(function() {
+      this.sandbox.restore();
+    });
+
+    it('assumes the caller will release the connection', async function() {
```
briandamaged: FYI: This test case fails for OracleDB when run against the code in `master`.

kibertoad:
Apparently I've created 0.20.x in the wrong repo 😂. Pushed it correctly from current master now; I don't think any of breaking changes made it in, but could you please double-check?

briandamaged:
@kibertoad : Thx for merging!

Based upon the git logs, it looks like there were no breaking changes between the `0.20.11` release and the current `master` branch.

PR #3741, opened 2020-03-20
Upgrade package mkdirp and improve it's usage

### LINKED ISSUE #3740: Upgrade dependency mkdirp - it is using an outdated version adding an unnecessary package for the users
Knex uses an outdated version of `mkdirp` package in it's `dependencies`.
https://github.com/knex/knex/blob/998c92ca97c323c7e7c1cb8788dd4c5e4ac3e6bd/package.json#L40

Can we update this to use the latest version offered by the package [1.0.3](https://www.npmjs.com/package/mkdirp/v/1.0.3)?

Please confirm and I can do a PR for this. It won't be a big [change](https://github.com/knex/knex/search?q=mkdirp&unscoped_q=mkdirp).

**Why upgrade?**
1. The current version `0.5.1` is deprecated and not on their supported list of versions. Emits this message:
   ```
   warning mkdirp@0.5.3: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
   ```
2. It requires a dependency `minimist` which has a vulnerability. The new version `1.0.3` has [zero dependencies](https://www.npmjs.com/package/mkdirp/v/1.0.3), so it's an unnecessary package for us and it will be removed. See this:
   image

kabirbaidhya:
**Changes**
* Upgrade [`mkdirp`](https://www.npmjs.com/package/mkdirp) to `^1.0.3`. Check issue #3740 for details.
* The new version is already Promise-friendly so update the usage a bit to address that.
* De-duplicate `_ensureFolder` logic by extracting the common logic into util function under `lib/util/fs.js`.
* Added tests for the new util function.
* Added CLI tests on the following files to ensure `seed:make` and `migration:make` commands do create directories (recursively) when they do not exist.
  - https://github.com/knex/knex/blob/master/test/cli/migrate-make.spec.js
  - https://github.com/knex/knex/blob/master/test/cli/seed-make.spec.js

Resolves #3740

### DIFF
--- a/lib/util/fs.js
+++ b/lib/util/fs.js
@@ -1,0 +1,19 @@
+const fs = require('fs');
+const { promisify } = require('util');
+const mkdirp = require('mkdirp');
+
+/**
+ * Ensures the given path exists.
+ *  - If the path already exist, it's fine - it does nothing.
+ *  - If the path doesn't exist, it will create it.
+ *
+ * @param {string} path
+ * @returns {boolean}
+ */
+function ensureDirectoryExists(dir) {
+  return promisify(fs.stat)(dir).catch(() => mkdirp(dir));
+}
+
+module.exports = {
+  ensureDirectoryExists,
+};
--- a/lib/util/fs.js
+++ b/lib/util/fs.js
@@ -3,6 +3,14 @@ const { promisify } = require('util');
 const mkdirp = require('mkdirp');
 
 /**
+ * Promisified verion fs.stat() function.
+ *
+ * @param {string} path
+ * @returns {Promise}
+ */
+const stat = promisify(fs.stat);
+
+/**
  * Ensures the given path exists.
  *  - If the path already exist, it's fine - it does nothing.
  *  - If the path doesn't exist, it will create it.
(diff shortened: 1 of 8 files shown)


lib/util/fs.js
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Ensures the given path exists.
+ *  - If the path already exist, it's fine - it does nothing.
+ *  - If the path doesn't exist, it will create it.
+ *
+ * @param {string} path
+ * @returns {Promise}
+ */
+function ensureDirectoryExists(dir) {
+  return promisify(fs.stat)(dir).catch(() => mkdirp(dir));
```
kibertoad: Can you extract `promisify(fs.stat)` outside of the function to make sure it is only invoked once?

kabirbaidhya: Done.

kibertoad:
@kabirbaidhya Well, there are some existing CLI tests already, see
https://github.com/knex/knex/blob/master/test/cli/migrate-make.spec.js
https://github.com/knex/knex/blob/master/test/cli/seed-make.spec.js

If you could add tests for creating directories recursively to both of them, that would be fantastic.

kabirbaidhya:
@kibertoad Done. Added CLI tests.

kibertoad:
Thanks a lot!

PR #3742, opened 2020-03-21
Add unit tests for escape utility functions

kabirbaidhya:
Follow up PR from https://github.com/knex/knex/pull/3568.
- Add tests for the main escape creator function `makeEscape` covering various cases both with default and custom config.
- Add tests for `dateToString` utility function to cover various it's logical cases
- Fix a potential issue by adding a default argument - that would show up if just the date is passed:
   ```
   TypeError: Cannot read property 'timeZone' of undefined
   ```
**Test Cases**
image

### DIFF
--- a/lib/query/string.js
+++ b/lib/query/string.js
@@ -101,7 +101,7 @@ function escapeString(val, finalEscape, ctx) {
   return "'" + escapedVal + "'";
 }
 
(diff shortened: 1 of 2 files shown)


kibertoad:
Thanks a lot!

kibertoad:
Released in 0.20.13

PR #3748, opened 2020-03-23
CLI: adds support for asynchronous knexfile loading

flovilmart:
We plan to use knex CLI but our credentials are safely stored in a vault, which requires to be accessed asynchronously.

This PR adds the support for letting users have the knexfile options returned asynchronously.


bin/cli.js
```diff
@@ -28,9 +28,14 @@ const fsPromised = {
...
 
-function openKnexfile(configPath) {
-  const config = require(configPath);
-
+async function openKnexfile(configPath) {
+  let config = require(configPath);
+  if (typeof config  === 'function') {
+    config = config();
+  }
+  if (config instanceof Promise) {
```
kibertoad: This is redundant, you can simply await config() and if it's not a promise, it will just return the value.

test/cli/knexfile-test.spec.js
```diff
@@ -65,6 +65,33 @@ module.exports = {
         );
       });
 
+      it('Run migrations with knexfile returning function passed', () => {
```
kibertoad: Can you also add same tests for seeder execution? Ideally the two should work symmetrically.

kibertoad:
One thing that I'm concerned about is that this implements support for async knexfile usage in CLI context, but not for general knex usage, meaning knexfile is no longer universally applicable.

You may want to check out https://github.com/knex/knex/pull/3497 - it may potentially address your use-case; if it's not, we can proceed with your implementation.

flovilmart:
@kibertoad you're right, #3497  addresses my needs for now.
Would you like to have full async support for the knexfile implemented extending my current implementation, or are you happy with the current implementation?
If so, I can update my PR to extend the support. otherwise, I'll close as I can work around it.

kibertoad:
Your extension is pretty non-intrusive; if the rest can be done in similarly non-intrusive way, and you could provide documentation for that, I think it could be merged.

flovilmart:
>  but not for general knex usage, meaning knexfile is no longer universally applicable.

@kibertoad I am not sure I follow. I can't see in the codebase where it loads `knexfile`. Can you point it out?

kibertoad:
@flovilmart lib/knex.js -> function Knex(config) {

Now I'm thinking whether or not user would expect promise to be supported here; it actually probably is going to be way simpler if they would resolve knexfile from the promise up-front and then passed that to the knex. So you are right, no additional changes are needed.

flovilmart:
awesome. Then again, I am not particularly attached to this PR, but it make is easier to deal with async knexfiles.

kibertoad:
@flovilmart Could you also submit PR to https://github.com/knex/documentation to mention that knexfile can return a promise?

flovilmart:
@kibertoad opened this one: https://github.com/knex/documentation/pull/259

kibertoad:
There's a minor conflict, could you please fix it?

kibertoad:
Thanks!

flovilmart:
Thanks for the quick response @kibertoad. I praised your commitment in my team this morning 🙌

kibertoad:
@flovilmart Ahaha, I'm touched! How urgently do you need this feature? I can cherry-pick this change to 0.20 prior to next major release, since it looks non-breaking.

flovilmart:
it's not urgent at all, we're in the process of integrating migrations to our codebases. If you wanna go ahead and cherry pick, that's appreciated nonetheless :)

kibertoad:
Released in  0.20.14.

flovilmart:
Thanks!

PR #3749, opened 2020-03-23
Minor improvements on the usage of fs utilities - unify all the fs functions into the same util/fs to simplify things

kabirbaidhya:
- Move all the promisified declarations of fs functions such as `readFile`, `writeFile` and `readdir` to `util/fs.js` module. This also deduplicates the usage a bit.
- Update usages and import them from the same module. 
- Add few tests - just for a quick validation of the module's exports


kabirbaidhya:
CI is passing now.

PR #3752, opened 2020-03-24
coverage calculation fix

maximelkin:


maximelkin:
@kibertoad can you revert it?
Actually I saw no differences, and dig it deeper:
https://github.com/istanbuljs/schema/blob/master/default-exclude.js
defaults already excludes tests

kibertoad:
Sure!

Issue #3738, opened 2020-03-18
Invalid TypeScript signatures for orderBy


danielrearden:
# Environment

Knex version: 0.20.11
Database + version: any
OS: MacOS 10.14.16

@lorefnon

# Bug

This is the current definition:

```
interface OrderBy<TRecord = any, TResult = unknown[]> {
  (columnName: keyof TRecord, order?: 'asc' | 'desc'): QueryBuilder<
    TRecord,
    TResult
  >;
  (columnName: string, order?: string): QueryBuilder<TRecord, TResult>;
  (
    columnDefs: Array<
      keyof TRecord | Readonly<{ column: keyof TRecord; order?: 'asc' | 'desc' }>
    >
  ): QueryBuilder<TRecord, TResult>;
  (
    columnDefs: Array<string | Readonly<{ column: string; order?: string }>>
  ): QueryBuilder<TRecord, TResult>;
  <TRecordInner, TResultInner>(
    subQueryBuilder: QueryBuilder<TRecordInner, TResultInner>
  ): QueryBuilder<TRecord, TResult>;
}
```

The last signature was added [with this PR](https://github.com/knex/knex/pull/3491/files). However, this signature doesn't account for specifying the order. Minimally, the order parameter should be added, but playing around with the API suggests a query builder can also now be passed in when using an object like `{ column: qb, order: 'desc' }`. Suggested changes:

```
interface OrderBy<TRecord = any, TResult = unknown[]> {
  (columnName: keyof TRecord, order?: 'asc' | 'desc'): QueryBuilder<
    TRecord,
    TResult
  >;
  (columnName: string, order?: string): QueryBuilder<TRecord, TResult>;
  (
    columnDefs: Array<
      keyof TRecord | Readonly<{ column: keyof TRecord; order?: 'asc' | 'desc' }>
    >
  ): QueryBuilder<TRecord, TResult>;
  (
    columnDefs: Array<string | Readonly<{ column: string; order?: string }>>
  ): QueryBuilder<TRecord, TResult>;
  <TRecordInner, TResultInner>(
    subQueryBuilder: QueryBuilder<TRecordInner, TResultInner>,

    // Add this parameter:

    order?: string
  ): QueryBuilder<TRecord, TResult>;

    // And this signature:

  <TRecordInner, TResultInner>(
    columnDefs: Array<string | Readonly<{ column: string | QueryBuilder<TRecordInner, TResultInner>; order?: string }>>
[...]

PR #3757, opened 2020-03-25
#3738: Update signature of orderBy to support QueryBuilder inside array

lorefnon:


kibertoad:
wonder what happened to Travis

PR #3760, opened 2020-03-26
named mssql rowcount as AFFECTEDROWS

helloyou2012:
Named mssql `rowcount` as `AFFECTEDROWS` like [`sequelize`](https://github.com/sequelize/sequelize/blob/5822bc793abe1cffd6220ba7888a5f925c51cb02/lib/dialects/mssql/query-generator.js#L539)


kibertoad:
First of all, this is a breaking change. Second, why make it consistent with Sequelize?

elhigu:
making it consistent across knex dialects could make sense though (if it is not already).

PR #3763, opened 2020-03-27
Implement "UPSERT" (Postgres/MySQL/Sqlite)

nicoburns:
CC: #3186

^ I'll write it up there.

Documentation PR: https://github.com/knex/documentation/pull/260

### DIFF
--- a/lib/dialects/mysql/query/compiler.js
+++ b/lib/dialects/mysql/query/compiler.js
@@ -21,7 +21,7 @@ class QueryCompiler_MySQL extends QueryCompiler {
   // Compiles an `insert` query, allowing for multiple
   // inserts using a single query statement.
   insert() {
-    let sql = QueryCompiler.prototype.insert.call(this);
+    let sql = super.insert();
     if (sql === '') return sql;
 
     const { ignore, merge, insert } = this.single;
@@ -35,13 +35,7 @@ class QueryCompiler_MySQL extends QueryCompiler {
     let sql = ' on duplicate key update ';
     if (updates) {
       const updateData = this._prepUpdate(updates);
-      if (typeof updateData === 'string') {
-        sql += updateData;
-      } else {
-        sql += updateData.join(',');
-      }
-
-      return sql;
+      return sql + updateData.join(',');
     } else {
       const insertData = this._prepInsert(insert);
       if (typeof insertData === 'string') {
@@ -51,7 +45,7 @@ class QueryCompiler_MySQL extends QueryCompiler {
       }
 
       sql += insertData.columns
-        .map((column) => this.formatter.wrapString(column))
+        .map((column) => this.formatter.wrapIdentifier(column))
         .map((column) => `${column} = values(${column})`)
         .join(', ');
 
--- a/lib/dialects/mysql/query/compiler.js
+++ b/lib/dialects/mysql/query/compiler.js
@@ -9,12 +9,12 @@ class QueryCompiler_MySQL extends QueryCompiler {
     super(client, builder);
 
     const { returning } = this.single;
-
     if (returning) {
       this.client.logger.warn(
         '.returning() is not supported by mysql and will not have any effect.'
       );
     }
+
     this._emptyInsertValue = '() values ()';
   }
 
--- a/lib/dialects/mysql/query/compiler.js
+++ b/lib/dialects/mysql/query/compiler.js
@@ -45,7 +45,7 @@ class QueryCompiler_MySQL extends QueryCompiler {
       }
 
       sql += insertData.columns
-        .map((column) => this.formatter.wrapIdentifier(column))
+        .map((column) => this.formatter.wrapAsIdentifier(column))
         .map((column) => `${column} = values(${column})`)
         .join(', ');
 
@@ -103,7 +103,7 @@ class QueryCompiler_MySQL extends QueryCompiler {
         'select * from information_schema.columns where table_name = ? and table_schema = ?',
       bindings: [table, this.client.database()],
       output(resp) {
-        const out = resp.reduce(function(columns, val) {
+        const out = resp.reduce(function (columns, val) {
           columns[val.COLUMN_NAME] = {
             defaultValue: val.COLUMN_DEFAULT,
             type: val.DATA_TYPE,
--- a/lib/dialects/mysql/query/compiler.js
+++ b/lib/dialects/mysql/query/compiler.js
@@ -18,6 +18,47 @@ class QueryCompiler_MySQL extends QueryCompiler {
     this._emptyInsertValue = '() values ()';
   }
 
+  // Compiles an `insert` query, allowing for multiple
+  // inserts using a single query statement.
+  insert() {
+    let sql = QueryCompiler.prototype.insert.call(this);
+    if (sql === '') return sql;
+
+    const { ignore, merge, insert } = this.single;
+    if (ignore) sql = sql.replace('insert into', 'insert ignore into');
+    if (merge) sql += this._merge(merge.updates, insert);
+
+    return sql;
+  }
+
+  _merge(updates, insert) {
+    let sql = ' on duplicate key update ';
+    if (updates) {
+      const updateData = this._prepUpdate(updates);
+      if (typeof updateData === 'string') {
+        sql += updateData;
+      } else {
+        sql += updateData.join(',');
+      }
+
+      return sql;
+    } else {
+      const insertData = this._prepInsert(insert);
+      if (typeof insertData === 'string') {
+        throw new Error(
+          'If using merge with a raw insert query, then updates must be provided'
+        );
+      }
+
+      sql += insertData.columns
+        .map((column) => this.formatter.wrapString(column))
+        .map((column) => `${column} = values(${column})`)
+        .join(', ');
+
+      return sql;
+    }
+  }
+
   // Update method, including joins, wheres, order & limits.
   update() {
     const join = this.join();
@@ -68,7 +109,7 @@ class QueryCompiler_MySQL extends QueryCompiler {
         'select * from information_schema.columns where table_name = ? and table_schema = ?',
       bindings: [table, this.client.database()],
       output(resp) {
-        const out = resp.reduce(function (columns, val) {
+        const out = resp.reduce(function(columns, val) {
           columns[val.COLUMN_NAME] = {
             defaultValue: val.COLUMN_DEFAULT,
             type: val.DATA_TYPE,
--- a/lib/dialects/postgres/query/compiler.js
+++ b/lib/dialects/postgres/query/compiler.js
@@ -25,9 +25,9 @@ class QueryCompiler_PG extends QueryCompiler {
     if (sql === '') return sql;
 
     const { returning, ignore, merge, insert } = this.single;
-    if (returning) sql += this._returning(returning);
     if (ignore) sql += this._ignore(ignore);
     if (merge) sql += this._merge(merge.updates, merge.columns, insert);
+    if (returning) sql += this._returning(returning);
 
     return {
       sql: sql,
@@ -67,7 +67,7 @@ class QueryCompiler_PG extends QueryCompiler {
   }
 
   _returning(value) {
-    return ` returning ${this.formatter.columnize(value)}`;
+    return value ? ` returning ${this.formatter.columnize(value)}` : '';
   }
 
   _ignore(columns) {
--- a/lib/dialects/postgres/query/compiler.js
+++ b/lib/dialects/postgres/query/compiler.js
@@ -21,11 +21,16 @@ class QueryCompiler_PG extends QueryCompiler {
   // Compiles an `insert` query, allowing for multiple
   // inserts using a single query statement.
   insert() {
-    const sql = QueryCompiler.prototype.insert.call(this);
+    let sql = QueryCompiler.prototype.insert.call(this);
     if (sql === '') return sql;
-    const { returning } = this.single;
+
+    const { returning, ignore, merge, insert } = this.single;
+    if (returning) sql += this._returning(returning);
+    if (ignore) sql += this._ignore(ignore);
+    if (merge) sql += this._merge(merge.updates, merge.columns, insert);
+
     return {
-      sql: sql + this._returning(returning),
+      sql: sql,
       returning,
     };
   }
@@ -62,7 +67,41 @@ class QueryCompiler_PG extends QueryCompiler {
   }
 
   _returning(value) {
-    return value ? ` returning ${this.formatter.columnize(value)}` : '';
+    return ` returning ${this.formatter.columnize(value)}`;
+  }
+
+  _ignore(columns) {
+    return ` on conflict (${this.formatter.columnize(columns)}) do nothing`;
+  }
+
+  _merge(updates, columns, insert) {
+    let sql = ` on conflict (${this.formatter.columnize(
+      columns
+    )}) do update set `;
+    if (updates) {
+      const updateData = this._prepUpdate(updates);
+      if (typeof updateData === 'string') {
+        sql += updateData;
+      } else {
+        sql += updateData.join(',');
+      }
+
+      return sql;
+    } else {
+      const insertData = this._prepInsert(insert);
+      if (typeof insertData === 'string') {
+        throw new Error(
+          'If using merge with a raw insert query, then updates must be provided'
+        );
+      }
+
+      sql += insertData.columns
+        .map((column) => this.formatter.wrapString(column.split('.').pop()))
+        .map((column) => `${column} = excluded.${column}`)
+        .join(', ');
+
+      return sql;
+    }
   }
 
   // Join array of table names and apply default schema.
--- a/lib/dialects/postgres/query/compiler.js
+++ b/lib/dialects/postgres/query/compiler.js
@@ -24,9 +24,10 @@ class QueryCompiler_PG extends QueryCompiler {
     let sql = QueryCompiler.prototype.insert.call(this);
     if (sql === '') return sql;
 
-    const { returning, ignore, merge, insert } = this.single;
-    if (ignore) sql += this._ignore(ignore);
-    if (merge) sql += this._merge(merge.updates, merge.columns, insert);
+    const { returning, onConflict, ignore, merge, insert } = this.single;
+    if (onConflict && ignore) sql += this._ignore(onConflict);
+    if (onConflict && merge)
+      sql += this._merge(merge.updates, onConflict, insert);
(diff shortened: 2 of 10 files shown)


SherloxTV:
Hey, thanks for the PR !

Overall LGTM, but it's missing a test case for passing an array of fields into `.onConflict`.

I could add those if you'd like :smile:

test/integration/builder/inserts.js
```diff
@@ -1278,6 +1278,112 @@ module.exports = function(knex) {
         });
     });
 
+    it('will silently do nothing when multiple inserts are made into a unique column and ignore is specificied', async function() {
```
kibertoad: specificied -> specified

kibertoad:
This also needs documentation - could you create a PR in https://github.com/knex/documentation?

SherloxTV:
@nicoburns tell me if you want help on the missing test case / the documentation, I'd be glad to help

nicoburns:
I've put in a documentation PR here: https://github.com/knex/documentation/pull/260.

Feedback on the documentation would definitely be helpful. I found it a little tricky to balance being clear and describing all possible combinations of options, which keeping the explanation concise. Ideally there should probably be a separate section on upserting with a guide-level explanation that explains things like setting of the schema to make it work (similar to the section on transactions). But thats a much bigger change to the structure of the docs, and could always come later.

Help implementing extra tests is also working. In addition to @SherloxTV's suggestion of a test that covers passing an array of fields into .onConflict, an integration test that covers use of merge with an array of inserted objects if probably a good idea.

test/integration/builder/inserts.js
nicoburns: fixed

lib/dialects/mysql/query/compiler.js
```diff
@@ -18,6 +18,47 @@ class QueryCompiler_MySQL extends QueryCompiler {
     this._emptyInsertValue = '() values ()';
   }
 
+  // Compiles an `insert` query, allowing for multiple
+  // inserts using a single query statement.
+  insert() {
+    let sql = QueryCompiler.prototype.insert.call(this);
```
maximelkin: here we can use super.insert() for simplicity

lib/dialects/mysql/query/compiler.js
```diff
@@ -18,6 +18,47 @@ class QueryCompiler_MySQL extends QueryCompiler {
...
+    if (merge) sql += this._merge(merge.updates, insert);
+
+    return sql;
+  }
+
+  _merge(updates, insert) {
+    let sql = ' on duplicate key update ';
+    if (updates) {
+      const updateData = this._prepUpdate(updates);
+      if (typeof updateData === 'string') {
```
maximelkin: _prepUpdate return only arrays

lib/dialects/mysql/query/compiler.js
```diff
@@ -18,6 +18,47 @@ class QueryCompiler_MySQL extends QueryCompiler {
...
+    } else {
+      const insertData = this._prepInsert(insert);
+      if (typeof insertData === 'string') {
+        throw new Error(
+          'If using merge with a raw insert query, then updates must be provided'
+        );
+      }
+
+      sql += insertData.columns
+        .map((column) => this.formatter.wrapString(column))
```
maximelkin: wrapIdentifier?

maximelkin: This method works just like wrapString, but without checking for including ' as '

lib/dialects/postgres/query/compiler.js
```diff
@@ -65,6 +70,40 @@ class QueryCompiler_PG extends QueryCompiler {
...
+    } else {
+      const insertData = this._prepInsert(insert);
+      if (typeof insertData === 'string') {
+        throw new Error(
+          'If using merge with a raw insert query, then updates must be provided'
+        );
+      }
+
+      sql += insertData.columns
+        .map((column) => this.formatter.wrapString(column.split('.').pop()))
```
maximelkin: Why here needed `.split('.').pop()`?

types/index.d.ts
```diff
@@ -739,6 +739,32 @@ declare namespace Knex {
...
+      >[]
+    >(
+      columns: readonly TKey[]
+    ): QueryBuilder<TRecord, TResult2>;
+    returning<TResult2 = SafePartial<TRecord>[]>(
+      column: string | readonly string[]
+    ): QueryBuilder<TRecord, TResult2>;
+
+    ignore(): QueryBuilder<TRecord, TResult>;
+    merge(data?: DbRecordArr<TRecord>): QueryBuilder<TRecord, TResult>;
```
maximelkin: I think here should be `DbRecord` instead of `DbRecordArr`

lib/dialects/mysql/query/compiler.js
nicoburns: Ah yeah, I just copy-pasted that from somewhere else. Have changed it to use super.

lib/dialects/mysql/query/compiler.js
nicoburns: Good point.

lib/dialects/mysql/query/compiler.js
nicoburns: Ah yeah, I guess inserts would never contain `as`.

lib/dialects/postgres/query/compiler.js
nicoburns: I was thinking that the column specified for the insert may be of the form `tableName.columnName`, in which case I only wanted the column name. I guess specifying the table name might not be supported for inserts anyway?

types/index.d.ts
nicoburns: Ah yeah, that sounds right. This function doesn't accept an array.

lib/dialects/mysql/query/compiler.js
maximelkin: Can you also fix other places?

tsondergaard:
I'm using knex for a project that can use either sqlite, mssql or oracle. For the time being this project would let me use onConflict() with only sqlite. My question is, what would a program look like that would like to use the onConflict() syntax when supported by the current dialect. Specifically what goes in the if-clause in place of hasOnConflict() here?

```javascript
if (hasOnConflict()) {
    // Use onConfict to upsert
} else {
    // use delete followed by insert
}
```

lib/dialects/postgres/query/compiler.js
```diff
@@ -16,16 +16,21 @@ class QueryCompiler_PG extends QueryCompiler {
   }
 
   // is used if the an array with multiple empty values supplied
-  
 
   // Compiles an `insert` query, allowing for multiple
   // inserts using a single query statement.
   insert() {
-    const sql = QueryCompiler.prototype.insert.call(this);
+    let sql = QueryCompiler.prototype.insert.call(this);
```
maximelkin: can you also apply review fixes from mysql driver?

lib/dialects/sqlite3/query/compiler.js
```diff
@@ -104,7 +111,50 @@ class QueryCompiler_SQLite3 extends QueryCompiler {
       }
       blocks[i] = block.join(', ');
     }
-    return sql + ' select ' + blocks.join(' union all select ');
+    sql += ' select ' + blocks.join(' union all select ');
+
+    const { onConflict, ignore, merge } = this.single;
+    if (onConflict && ignore) sql += ' where true' + this._ignore(onConflict);
```
elhigu, line 117: Could we setup prettier to force having braces also in one-liner `if` statements? 

Those are pretty error prone when code is modified later on and in merging errors (here is one pretty public example of it https://blog.codecentric.de/en/2014/02/curly-braces/ ). 

Also in this case where the first `if` doesn't have braces and the second `else if` does hurt my eyes 😅

kibertoad:
@nicoburns Do you need any help finishing this?

lib/dialects/mysql/query/compiler.js
```diff
@@ -18,6 +18,41 @@ class QueryCompiler_MySQL extends QueryCompiler {
...
+      const insertData = this._prepInsert(insert);
+      if (typeof insertData === 'string') {
+        throw new Error(
+          'If using merge with a raw insert query, then updates must be provided'
+        );
+      }
+
+      sql += insertData.columns
+        .map((column) => this.formatter.wrapIdentifier(column))
+        .map((column) => `${column} = values(${column})`)
```
SpenserJ, line 49: As of [MySQL 8.0.20](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-20.html), `values(column)` is deprecated and they recommend using row aliases (added in 8.0.19):
`INSERT INTO t1 (id,val) VALUES (1,15),(2,10) AS new ON DUPLICATE KEY UPDATE val = new.val;`

alubbe:
as someone who woud love to see this merged, what are the meaning of the two labels here? One says missing documentation, but there is a PR for it here https://github.com/knex/documentation/pull/260. The other one says missing tests, but there are a few hundred lines of testing code here. What cases specifically would you like to see tested? I do see some merge conflicts against `master` that need resolving though.

elhigu:
> as someone who woud love to see this merged, what are the meaning of the two labels here? One says missing documentation, but there is a PR for it here [knex/documentation#260](https://github.com/knex/documentation/pull/260). The other one says missing tests, but there are a few hundred lines of testing code here. What cases specifically would you like to see tested? I do see some merge conflicts against `master` that need resolving though.

Integration tests, which verifies that when those queries are ran against real database they behave as expected. In other words all this: https://github.com/knex/knex/issues/3186#issuecomment-605314991

Mostly the same queries, whose generation is currently covered by query builder code generation tests.

Alexsey:
Shouldn't this also support `batchInsert`?

elhigu:
@Alexsey No. `.batchInsert` is very opinionated and one use only helper (which should not even be a part of knex, but it should be a separate extension plugin at most).

vazra:
any help required to get this merged? @nicoburns @elhigu @kibertoad

kibertoad:
@vazra Yes, please! Would be great to add missing tests (see comments) and resolve conflicts.

ARKKYN:
any update on this pr

A-Tokyo:
Any updates? 🚀

nicoburns:
Sorry to have left this for so long. Life has been busy! Some thoughts:

#### Regarding Tests
There are already some integration tests as part of this PR. Is there anything in particular that you feel hasn't been covered?

One thing which has occurred to me might be good to cover is the interaction of this feature with `knex.raw()`. This should probably have both unit and integration tests.

#### Regarding the extra parameter for MySQL

I'm somewhat reluctant to implement this. Because:

- The only thing it does is assert that the user of the library has read the documentation, which they should have done anyway. It seems likely that a user who has not read the documentation will simply copy and paste an example along with the parameter.
- There is no way for us to verify that the columns listed actually correspond to the PRIMARY KEY (we'd need to add some kind of generated code from the database schema, which would go well beyond knex's current scope).
- Postgres and SQLlite have a similar invariant: that the columns listed in the onConflict part of the query have a UNIQUE index on them. So making sure that the schema is setup correctly for using this feature is necessary for all databases, not just MySQL.

elhigu:
> There are already some integration tests as part of this PR

```
Integration tests, which verifies that when those queries are ran against real database they behave as expected. In other words all this: #3186 (comment)

Mostly the same queries, whose generation is currently covered by query builder code generation tests.
```

Those tests that are currently included are not really integration tests (even that it says integration in folder name). Real integration tests actually executed the queries against real database. Otherwise there is no guarantees that any of these queries are returning expected data on every supported DB.

> Regarding the extra parameter for MySQL
> I'm somewhat reluctant to implement this. Because:
> The only thing it does is assert that the user of the library has read the documentation, which they should have done anyway. It seems likely that a user who has not read the documentation will simply copy and paste an example along with the parameter.

Not throwing an error in this kind of cases, where knex silently generate wrong code without warning for those DBs who doesn't support certain feature has been a problem and people are then adding bug reports of it. Easier to just have code to notice invalid usage.

> There is no way for us to verify that the columns listed actually correspond to the PRIMARY KEY (we'd need to add some kind of generated code from the database schema, which would go well beyond knex's current scope).

No such verification is needed, it should be safe enough and verbose to read from code to have that extra parameter as a seal of trust from programmer that this conflict statement actually is valid and crossdb compatible,

> Postgres and SQLlite have a similar invariant: that the columns listed in the onConflict part of the query have a UNIQUE index on them. So making sure that the schema is setup correctly for using this feature is necessary for all databases, not just MySQL.

I had no idea that also other DBs had these limitations. This is actually a good reason why it is pretty much useless to add that error then. We can just add good documentation about these pitfalls and that should be enough.

nicoburns:
I've rebased this PR against the latest master.

> Those tests that are currently included are not really integration tests (even that it says integration in folder name). Real integration tests actually executed the queries against real database. Otherwise there is no guarantees that any of these queries are returning expected data on every supported DB.

Oh, I was under the impression that the integration tests were executing against a real database. Where do the tests which execute against a real database live?

kibertoad:
@nicoburns  test/integration/ is generally the right place, but most of tests in /builder/ are asserted by checking just the generated sql (see the ` .testSql` method usage) as opposed to asserting actual changes inside the DB. You can add missing test derived from your existing one, just replace `testSql` with actually executing SQL and then retrieving the data and asserting on it.

tsheaff:
@nicoburns thanks for all your work on this so far. It sounds like the last piece is fleshing out the integration tests? Are you looking for help with that piece, or are you planning to get this finished yourself? Let us know, Can't wait for this feature to be merged 🎉

nicoburns:
@tsheaff I'm planning to get around to this at some point. But I'm not quite sure when I'll have time. If anyone else wants to finish it up then I'm quite happy for them to do so.

I looked into the tests again after @kibertoad's comment above, and I believe my integration tests *are* running against an actual database. They call `.testSQL` (which doesn't run the commands), but they *also* `await` the query, select the data back out of the database and run assertions against the retrieved data.

However, I believe there are still a couple of outstanding items:
1. It seems that my tests don't actually pass anymore (not sure why - I'm sure they used to!)
1. I think it would be good to add tests around the usage of this feature with `knex.raw()` (in value position).

tsheaff:
Amazing @nicoburns thanks! Looks like it's super close. The ramining failing tests I'm seeing are:

```
  1) Integration Tests
       oracle | oracledb
         Inserts
           will silently do nothing when multiple inserts are made into a unique column and ignore is specified:
     Error: insert into "upsert_tests" ("email", "name") values (:1, :2) returning "email" into :3 - ORA-00001: unique constraint (SYSTEM.upsert_tests_email_unique) violated
  
  2) Integration Tests
       oracle | oracledb
         Inserts
           will silently do nothing when multiple inserts are made into a composite unique column and ignore is specified:
     Error: insert into "upsert_composite_key_tests" ("email", "name", "org") values (:1, :2, :3) returning "email" into :4 - ORA-00001: unique constraint (SYSTEM.pWLdp/gASIHRY5hXRVs73MBLb1E) violated
  
  3) Integration Tests
       mssql | mssql
         Inserts
           will silently do nothing when multiple inserts are made into a unique column and ignore is specified:
     RequestError: insert into [upsert_tests] ([email], [name]) output inserted.[email] values (@p0, @p1) - Cannot insert duplicate key row in object 'dbo.upsert_tests' with unique index 'upsert_tests_email_unique'. The duplicate key value is (ignoretest@example.com).

  4) Integration Tests
       mssql | mssql
         Inserts
           will silently do nothing when multiple inserts are made into a composite unique column and ignore is specified:
     RequestError: insert into [upsert_composite_key_tests] ([email], [name], [org]) output inserted.[email] values (@p0, @p1, @p2) - Cannot insert duplicate key row in object 'dbo.upsert_composite_key_tests' with unique index 'upsert_composite_key_tests_org_email_unique'. The duplicate key value is (acme-inc, ignoretest@example.com).
```

So it's specific to mssql and oracle which don't implement "upsert". If it were an issue remaining with Postgres I'd hop on it since I'm an expert in Postgres, but I've worked with mssql and oracle zero in my career so I don't feel comfortable with such an important diff on a DB engine I'm not familiar with.

Anyone from the community who's expressed interest able to hop on and fix these last failures? cc @vazra @ARKKYN @A-Tokyo or others?

nicoburns:
I've pushed a commit ignoring those tests for oracle and mssql, as those platforms are not currently supported by this feature. Perhaps we ought to have some kind of runtime check that throws an error if someone tries to use this feature with those databases?

For those following this PR, there is also still on going work on the corresponding documentation taking place here: https://github.com/knex/documentation/pull/260

kibertoad:
>Perhaps we ought to have some kind of runtime check that throws an error if someone tries to use this feature with those databases?

Sounds good!

nicoburns:
> Perhaps we ought to have some kind of runtime check that throws an error if someone tries to use this feature with those databases?

This is now implemented.

Additionally I have created a new "class" (ES5 syntax) `OnConflictQueryBuilder` which is returned from `.onConflict()` and wraps the regular query builder. It has two methods: `.ignore()` and `.merge()` which modify and return your query. The end result of this is that the interface for onConflict queries is the same for valid queries, but invalid combinations of `.onConflict()`, `.ignore()`, and `.merge()` are impossible. I also added a `.then()` method to `OnConflictQueryBuilder` which throws, to catch the case where someone tries to `await` an query that ends in a `.onConflict()` (which is invalid without a matching `.ignore()` or `.merge()`.

edd83:
🎉 Amazing work @nicoburns, definitely wants this feature to Knex instead of doing custom queries, @elhigu anything else to add to this PR? 😄

kibertoad (approved):

tsheaff:
Thanks for all the hard work on this @nicoburns 👏 

@kibertoad how long would you expect for it to take to get this master-merged commit into the next official release version?

kibertoad:
Will be released as soon as documentation is finalized by @nicoburns

kibertoad:
Released in 0.21.10

vvo:
Thank you so much to the knex contributors for implementing this wonderful feature. Saves a lot of boilerplate code on my side. Congrats on the API too.

lib/dialects/sqlite3/query/compiler.js
glensc: totally support the curly braces, regardless of the 1st of April post date :)

KrisLau:
@kibertoad Someone implemented it for mssql here: https://github.com/knex/knex/pull/6050. I tested it and it seems to work well at least on my local environment

melroyvandenberg:
It's not documented? https://knexjs.org/guide/query-builder.html#upsert

PR #3783, opened 2020-04-06
Support more modes to set the host as a Postgres socket

danielrozenberg:
v2.2.0 of pg-connection-string adds this feature: https://github.com/iceddev/pg-connection-string/pull/34


kibertoad:
Released in  0.20.14.

PR #3784, opened 2020-04-07
Add types for `.distinctOn`

victorandree:
Interface as provided by #3513.

There's a separate PR for docs at https://github.com/knex/documentation/pull/257

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -432,6 +432,7 @@ declare namespace Knex {
     into: Table<TRecord, TResult>;
     table: Table<TRecord, TResult>;
     distinct: Distinct<TRecord, TResult>;
+    distinctOn: DistinctOn<TRecord, TResult>;
 
     // Joins
     join: Join<TRecord, TResult>;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -907,6 +908,16 @@ declare namespace Knex {
...
 
+  interface DistinctOn<TRecord extends {}, TResult = {}[]> {
+    <ColNameUT extends keyof TRecord>(
+      ...columnNames: readonly ColNameUT[]
+    ): QueryBuilder<TRecord, TResult>;
+
+    <ColNameUT extends keyof TRecord>(
+      columnNames: readonly ColNameUT[]
+    ): QueryBuilder<TRecord, TResult>;
+  }
```
lorefnon, line 919: It is also good to add another signature with `columnNames: readonly string[]` to accomodate things like `distinctOn('users.foo')` where the argument is not the same as column key.

lorefnon:
@victorandree Thanks for your PR. Besides the minor change requested, this looks good.

victorandree:
Thanks @lorefnon! I force pushed to add 2 string interfaces:

```typescript
(...columnNames: readonly string[]): QueryBuilder<TRecord, TResult>;
(columnNames: readonly string[]): QueryBuilder<TRecord, TResult>;
```

kibertoad:
Released in 0.20.15

PR #3787, opened 2020-04-10
Fix Knex interface types

tkrotoff:
I believe `interface Knex` and `function Knex` should have the same types. (well I don't understand why they should have different types).

This causes some problems with Bookshelf.

Without this PR:

Screen Shot 2020-04-10 at 15 12 46

<br>

Screen Shot 2020-04-10 at 15 13 40

With this PR:

Screen Shot 2020-04-10 at 15 01 18

I've run `dtslint types` and everything is OK

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -322,7 +322,7 @@ interface PgTableOptions {
   only?: boolean;
 }
 
-interface Knex<TRecord extends {} = any, TResult = any[]>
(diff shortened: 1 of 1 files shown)


lorefnon:
Yes, it makes sense for both of them to have the same defaults. Thanks for the PR.

kibertoad:
Released in  0.20.14.

PR #3791, opened 2020-04-13
add delayedThen/delayedCatch methods

sloops77:
Im building a lightweight framework around knex. 

delayedThens and delayedCatches (im not sure if there is a more correct name) allow me to declaratively handle postprocessing of responses for finds, inserts, updates, etc that are unique to those queries, while still allowing the developer to be in control of when execution happens AND can continue to modify the querybuilder outside of my framework seamlessly.

Question 1: is this something of interest?
Question 2: if so then where should i add tests? Under unit, integration, both?

### DIFF
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -11,8 +11,22 @@ module.exports = function(Target) {
     }).join(';\n');
   };
 
+  Target.prototype.delayThen = function() {
+    this.interceptors = this.interceptors || [];
+    this.interceptors.push(['then', arguments]);
+    return this;
+  };
+
+  Target.prototype.delayCatch = function() {
+    this.interceptors = this.interceptors || [];
+    this.interceptors.push(['catch', arguments]);
+    return this;
+  };
+
   // Create a new instance of the `Runner`, passing in the current object.
   Target.prototype.then = function(/* onFulfilled, onRejected */) {
+    this.interceptors = this.interceptors || [];
+    this.interceptors.push(['then', arguments]);
     let result = this.client.runner(this).run();
 
     if (this.client.config.asyncStackTraces) {
@@ -26,7 +40,11 @@ module.exports = function(Target) {
       });
     }
 
-    return result.then.apply(result, arguments);
+    for(const interceptor of this.interceptors) {
+      result = result[interceptor[0]].apply(result, interceptor[1]);
+    }
+
+    return result;
   };
 
   // Add additional "options" to the builder. Typically used for client specific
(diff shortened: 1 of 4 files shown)


elhigu:
Question 1: not really wanted like this... you might like to checkout objection.js it is light weight and already has this kind of feature.

If we are going to add more this kind of hooks they should be designed carefully to meet enough wide usecases. Right now you should be able to implement something like this with your own custom https://knexjs.org/#Installation-post-process-response function (not sure if it includes catching).

sloops77:
Sorry - github notifications means that i didnt catch this.

@elhigu 

I dont want to use an ORM. The reason i want delayThen & delayedCatches is to make a framework that is much lighter weight than something like objection, it is seemless to the regular knex user, but allows them to be much DRYer. It permits the programmer to just get data out of the database. There are no models or classes, and relationships are handled using knex queries & transactions directly.

For me, the best would be for there to be a pipeline that can be hooked into before execution is triggered by the client code. `delayThen` & `delayCatch` allows me to setup these elements of the execution pipeline bespoke for each function i expose. 
Whatever solution is created is not intended to be used by regular use cases, but is aimed at advanced use cases or framework builders, who would be able to leverage and extend your work in a far simpler way than `postProcessResponse` allows. 

I get you would prefer to think about this architecture more thoroughly before implementing it so maybe it should go into the pot for your next major version. 

let me give you a flavour for how i am using it.

The framework exposes a `findById` function and a `findOne` function. The function `findById` just takes an id parameter, and `delayThen` is used to declaratively specify that the framework errors if no result is returned. The `findOne` function accepts arbitrary filters and does NOT error if no result is returned.  `postProcessResponse` does not permit such customizations easily, at least not without having an awful interface that as it must be shared amongst all.

elhigu:
> I get you would prefer to think about this architecture more thoroughly before implementing it so maybe it should go into the pot for your next major version

I'm pretty much not writing any code for knex anymore, so it is up to individual contributors to decide, which features/bug fixes they like to implement.

> The findOne function accepts arbitrary filters and does NOT error if no result is returned. postProcessResponse does not permit such customizations easily, at least not without having an awful interface that as it must be shared amongst all.

I'm pretty sure that with extending knex with special delayThen method which sets handler to query builder context and which is then read / executed by postProcessResponse hook that could be implemented as a plugin. Nothing specially horrible shared awfulness there (it would be global shared afwfulness, but in the knex core code otherwise).

I would really want knex to evolve to even more simple tool just for making query building / modifying to be awesome with it and I want that knex would have powerful APIs for extending it to allow other people to provide those additional functions to it. 

That being said, if there is a good powerful unopiniated design for hooks, which doesn't complicate knex too much I don't see it impossible of having those. Though current status of knex is quite complicated mess... so stuffing more and more things to top of that mess does not sound like a good idea.

PR #3800, opened 2020-04-15
Fix: Support for `.finally(..)` on knex's Promise-alikes

briandamaged:


kibertoad:
@briandamaged is ` knex.select(1).finally(() => console.log('huh'))` still breaking after this change?

briandamaged:
@kibertoad : Looks like it passes.  Want me to add a unit test for that as well, or is there already one available in a separate branch?

kibertoad:
No such test yet, so would be good to add.

briandamaged:
@kibertoad : k -- it looks like the unit tests and such are passing now.  Should we merge it?

kibertoad:
Yup! Thanks a lot. I'll cherry-pick and cut new 0.20.x release later today.

PR #3804, opened 2020-04-16
Reduce size of lodash in bundle

tkalliom:
When rolling up the current master plus pg, the resulting bundle is 961KB, of which more than half consists of just lodash. This PR removes unnecessary usage of lodash and changes the remaining requires so that only the necessary parts are loaded. The size of the same bundle is down 660KB, so roughly half of lodash is left out.

### DIFF
--- a/lib/dialects/mssql/index.js
+++ b/lib/dialects/mssql/index.js
@@ -1,6 +1,8 @@
 // MSSQL Client
 // -------
-const { map, flatten, values } = require('lodash');
+const flatten = require('lodash/flatten');
+const map = require('lodash/map');
+const values = require('lodash/values');
 const inherits = require('inherits');
 
 const Client = require('../../client');
--- a/lib/dialects/mysql/schema/columncompiler.js
+++ b/lib/dialects/mysql/schema/columncompiler.js
@@ -3,7 +3,7 @@
 const inherits = require('inherits');
 const ColumnCompiler = require('../../../schema/columncompiler');
 
-const { isObject } = require('lodash');
+const isObject = require('lodash/isObject');
 
 function ColumnCompiler_MySQL() {
   ColumnCompiler.apply(this, arguments);
--- a/lib/dialects/mysql/schema/compiler.js
+++ b/lib/dialects/mysql/schema/compiler.js
@@ -3,7 +3,7 @@
 const inherits = require('inherits');
 const SchemaCompiler = require('../../../schema/compiler');
 
-const { some } = require('lodash');
+const some = require('lodash/some');
 
 function SchemaCompiler_MySQL(client, builder) {
   SchemaCompiler.call(this, client, builder);
--- a/lib/dialects/sqlite3/query/compiler.js
+++ b/lib/dialects/sqlite3/query/compiler.js
@@ -3,14 +3,12 @@
 const QueryCompiler = require('../../../query/compiler');
 
 const noop = require('../../../util/noop');
-const {
-  constant,
-  each,
-  isEmpty,
-  isString,
-  reduce,
-  identity,
-} = require('lodash');
+const constant = require('lodash/constant');
+const each = require('lodash/each');
+const identity = require('lodash/identity');
+const isEmpty = require('lodash/isEmpty');
+const isString = require('lodash/isString');
+const reduce = require('lodash/reduce');
 
 const emptyStr = constant('');
 
--- a/lib/dialects/sqlite3/query/compiler.js
+++ b/lib/dialects/sqlite3/query/compiler.js
@@ -2,12 +2,12 @@
 
 const QueryCompiler = require('../../../query/compiler');
 
+const noop = require('../../../util/noop');
 const {
   constant,
   each,
   isEmpty,
   isString,
-  noop,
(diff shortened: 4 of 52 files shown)


lib/dialects/mssql/index.js
```diff
@@ -1,6 +1,8 @@
 // MSSQL Client
 // -------
-const { map, flatten, values } = require('lodash');
+const flatten = require('lodash/flatten');
```
kibertoad, line 3: Doesn't JS support `flat` operation natively in recent enough Node.js?

lib/dialects/mssql/index.js
```diff
@@ -1,6 +1,8 @@
...
+const map = require('lodash/map');
+const values = require('lodash/values');
```
kibertoad, line 5: Why not use Object.values instead?

lib/dialects/mysql/schema/columncompiler.js
```diff
@@ -74,7 +72,7 @@ Object.assign(ColumnCompiler_MySQL.prototype, {
   },
 
   datetime(precision) {
-    if (isObject(precision)) {
+    if (precision instanceof Object) {
```
kibertoad, line 75: can you create an internal isObject function and refer to it from everywhere, so that it would be easier to replace underlying implementation in the future? there are certain edge cases for this check.

lib/dialects/mysql/schema/compiler.js
```diff
@@ -3,7 +3,7 @@
 const inherits = require('inherits');
 const SchemaCompiler = require('../../../schema/compiler');
 
-const { some } = require('lodash');
+const some = require('lodash/some');
```
kibertoad, line 6: Should be possible to replace with native `some`

lib/dialects/sqlite3/query/compiler.js
```diff
@@ -2,20 +2,17 @@
...
-  noop,
-  reduce,
-  identity,
-} = require('lodash');
+const constant = require('lodash/constant');
+const each = require('lodash/each');
+const identity = require('lodash/identity');
+const isEmpty = require('lodash/isEmpty');
+const isString = require('lodash/isString');
+const noop = require('lodash/noop');
```
kibertoad, line 10: there is already an internal implementation of noop that can be used: lib\util\noop.js

lib/query/builder.js
```diff
@@ -7,25 +7,20 @@ const { EventEmitter } = require('events');
...
-  toArray,
-  reject,
-  includes,
-  last,
-  isPlainObject,
-} = require('lodash');
+const assign = require('lodash/assign');
+const clone = require('lodash/clone');
+const each = require('lodash/each');
+const includes = require('lodash/includes');
```
kibertoad, line 13: It seems that `includes` can be replaced with native implementation

kibertoad:
@tkalliom This is amazing, thank you!

lib/dialects/mysql/schema/columncompiler.js
maximelkin: i.e. Object.create(null)

lib/dialects/mysql/schema/compiler.js
maximelkin: this place and next - lodash for many functions accept also object

lib/migrate/Migrator.js
```diff
@@ -141,15 +134,10 @@ class Migrator {
...
-          isEmpty(
-            filter(migrationsToRun, (migration) => {
-              const migrationContents = this.config.migrationSource.getMigration(
-                migration
-              );
-
-              return !this._useTransaction(migrationContents);
-            })
-          );
+          (!migrationToRun ||
```
maximelkin, line 137: lost some here or this is intended change?

lib/dialects/mysql/schema/compiler.js
tkalliom: I was not certain whether `resp` is always an array, or if it could be `undefined` in some situations. `_.some(undefined, ...)` returns `false` while `undefined.some` throws.

lib/query/builder.js
tkalliom: Knex supports running in the browser. I didn’t find a documentation of what versions are supported, so I sticked to ES5

lib/dialects/mssql/index.js
kibertoad: If you check package.json and readme, it's already Node 8+

lib/dialects/mysql/schema/columncompiler.js
kibertoad: That also works

lib/dialects/mssql/index.js
kibertoad: Sounds good!

tkalliom:
^ it appears as if there is some flakiness in the CI as Node 8, 12 and 13 runs passed but 10 failed with a couple of ”ORACLE initialization or shutdown in progress” – can someone with the access retrigger?

kibertoad:
Released in 0.21.0

PR #3809, opened 2020-04-18
Bump dependencies

kibertoad:


kibertoad:
Latest tarn.js types throw an error for pre-3.7 versions: "An accessor cannot be declared in an ambient context", so I had to bump it. @lorefnon do you think that makes sense?

lorefnon:
I don't have any issue with bumping to TS 3.7+. 

If someone comes up with a request to support lower versions of TS we can send a PR to tarn to use [downlevel-dts](https://github.com/sandersn/downlevel-dts).

kibertoad:
@maximelkin Looks like Node 13 is consistently failing with Oracle when newest Tarn.js is used. Any ideas?

kibertoad:
Wait, this is the first time. Let me rerun...

PR #3820, opened 2020-04-22
rewrite delay to not node-only version

maximelkin:


kibertoad:
Can you explain the reason behind this change?

maximelkin:
Users write comments about broken delay() function when they run tests in browser environment (yeah they didn't know about it, but this is a bit tediously).
Also in readme indicated browser support.

(personally i dont think what somebody really use knex in browser, lol)

kibertoad:
Makes sense!

kibertoad:
Released in knex@0.21.1

PR #3822, opened 2020-04-22
Add migrate:unlock command, truncate on forceFreeMigrationsLock

zachelrath:
# Changes
- Adds new `migrate:unlock` command to CLI which invokes `knex.migrations.forceFreeMigrationsLock()`, to provide for a CLI-based way of forcibly resetting the `knex_migrations_lock` table in the event that it gets into a bad state.
- Adjusts the behavior of `forceFreeMigrationsLock` to truncate the `knex_migrations_lock` table to ensure that there is only 1 row in the event that multiple rows have been inserted into that table, in which case [migrations will fail](https://github.com/knex/knex/blob/master/lib/migrate/Migrator.js#L299-L310)

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -58,7 +58,6 @@ function invoke(env) {
   env.modulePath = env.modulePath || env.knexpath || process.env.KNEX_PATH;
 
   const filetypes = ['js', 'coffee', 'ts', 'eg', 'ls'];
-  let pending = null;
 
   const cliVersion = [
     color.blue('Knex CLI version:'),
@@ -110,7 +109,7 @@ function invoke(env) {
       }
       checkLocalModule(env);
       const stubPath = `./knexfile.${type}`;
-      pending = readFile(
+      readFile(
         path.dirname(env.modulePath) +
           '/lib/migrate/stub/knexfile-' +
           type +
@@ -148,7 +147,7 @@ function invoke(env) {
         configOverrides.stub = stub;
       }
 
-      pending = instance.migrate
+      instance.migrate
         .make(name, configOverrides)
         .then((name) => {
           success(color.green(`Created Migration: ${name}`));
@@ -161,7 +160,7 @@ function invoke(env) {
     .description('        Run all migrations that have not yet been run.')
     .option('--verbose', 'verbose')
     .action(() => {
-      pending = initKnex(env, commander.opts())
+      initKnex(env, commander.opts())
         .then((instance) => instance.migrate.latest())
         .then(([batchNo, log]) => {
           if (log.length === 0) {
@@ -181,7 +180,7 @@ function invoke(env) {
       '        Run the next or the specified migration that has not yet been run.'
     )
     .action((name) => {
-      pending = initKnex(env, commander.opts())
+      initKnex(env, commander.opts())
         .then((instance) => instance.migrate.up({ name }))
         .then(([batchNo, log]) => {
           if (log.length === 0) {
@@ -207,7 +206,7 @@ function invoke(env) {
     .action((cmd) => {
       const { all } = cmd;
 
-      pending = initKnex(env, commander.opts())
+      initKnex(env, commander.opts())
         .then((instance) => instance.migrate.rollback(null, all))
         .then(([batchNo, log]) => {
           if (log.length === 0) {
@@ -228,7 +227,7 @@ function invoke(env) {
       '        Undo the last or the specified migration that was already run.'
     )
     .action((name) => {
-      pending = initKnex(env, commander.opts())
+      initKnex(env, commander.opts())
         .then((instance) => instance.migrate.down({ name }))
         .then(([batchNo, log]) => {
           if (log.length === 0) {
@@ -249,7 +248,7 @@ function invoke(env) {
     .command('migrate:currentVersion')
     .description('        View the current version for the migration.')
     .action(() => {
-      pending = initKnex(env, commander.opts())
+      initKnex(env, commander.opts())
         .then((instance) => instance.migrate.currentVersion())
         .then((version) => {
           success(color.green('Current Version: ') + color.blue(version));
@@ -262,7 +261,7 @@ function invoke(env) {
     .alias('migrate:status')
     .description('        List all migrations files with status.')
     .action(() => {
-      pending = initKnex(env, commander.opts())
+      initKnex(env, commander.opts())
         .then((instance) => {
           return instance.migrate.list();
         })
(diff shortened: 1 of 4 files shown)


lib/migrate/Migrator.js
```diff
@@ -278,10 +278,18 @@ class Migrator {
...
 
-    const lockTable = getLockTableName(this.config.tableName);
+    const lockTableName = getLockTableName(this.config.tableName);
+
     return getSchemaBuilder(this.knex, this.config.schemaName)
       .hasTable(lockTable)
-      .then((exist) => exist && this._freeLock());
+      .then((exist) => {
+        if (exist) {
+          await getTable(this.knex, lockTableName, this.config.schemaName).del();
```
kibertoad, line 287: `this.config.schemaName` and `lockTableName` usage is not consistent.

kibertoad, line 287: can you rewrite whole chain to be async/await?

test/jake/jakelib/migrate-test.js
```diff
@@ -717,6 +717,40 @@ test('migrate runs "esm" modules', (temp) => {
   });
 });
 
+test('migrate:unlock restores lock table to one row with is_locked=0', (temp) => {
+  const db = new sqlite3.Database(`${temp}/db`);
+  // Seed multiple rows into the lock table
+  return new Promise((resolve, reject) =>
```
kibertoad, line 723: Could you please use async/await instead of a promise chain?

lib/migrate/Migrator.js
zachelrath: @kibertoad Can you explain what you mean by `not consistent`? `lockTableName` is used in multiple locations in the outer function, which is why i assigned it to a `const`.

zachelrath: Sure, I could do that.

test/jake/jakelib/migrate-test.js
```diff
@@ -717,6 +717,40 @@ test('migrate runs "esm" modules', (temp) => {
...
+    db.run('INSERT INTO knex_migrations_lock(is_locked) VALUES (?),(?),(?)', [1, 0, 1], (err) => {
+      err ? reject(err) : resolve();
+    })
+  )
+    .then(() => {
+      return assertExec(
```
kibertoad, line 729: Would be awesome if you could use new style of executing script and asserting results - see https://github.com/knex/knex/blob/master/test/cli/migrate.spec.js for an example

test/jake/jakelib/migrate-test.js
zachelrath: yep, can do that, was just trying to maintain consistency with other tests in this file

lib/migrate/Migrator.js
kibertoad: yeah, this is fair. still for symmetry sake would be nice to do `const lockSchemaName = ` as well.

test/jake/jakelib/migrate-test.js
kibertoad: this is legacy set of tests, I really need to add comment pointing into non-obsolete tests as a reference.

test/jake/jakelib/migrate-test.js
zachelrath: Ah okay, is that the desired directory for net-new CLI tests? If so I will move my test over into that file.

kibertoad: yes, we use testlab-cli instead of plain jake now.

zachelrath: Sounds good! The new format is much cleaner and more familiar, so I'm all for switching over.

test/cli/migrate-unlock.spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+describe('migrate:unlock', () => {
+    before(() => {
+        process.env.KNEX_PATH = '../knex.js';
+    });
+
+    const rootDir = getRootDir();
+    const dbPath = path.resolve(rootDir, 'db');
+
+    it('should restore lock table to one row with is_locked=0', async () => {
```
kibertoad, line 18: Can you also add test for the case when there was no lock in the first place?

test/cli/migrate-unlock.spec.js
```diff
@@ -0,0 +1,46 @@
...
+            db.all('SELECT * FROM knex_migrations_lock', {}, (err, rows) => {
+                if (err) {
+                    reject(err);
+                }
+                resolve(rows);
+            });
+        });
+        expect(rows.length).to.equal(1);
+        expect(rows[0].is_locked).to.equal(0);
+        db.close();
```
kibertoad, line 42: db instantiation and closing probably could be moved to `before` and `after` blocks

test/cli/migrate-unlock.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    before(() => {
+        process.env.KNEX_PATH = '../knex.js';
+    });
+
+    const rootDir = getRootDir();
+    const dbPath = path.resolve(rootDir, 'db');
+
+    it('should restore lock table to one row with is_locked=0', async () => {
+        
+        const db = await new sqlite3.Database(dbPath);
```
kibertoad, line 20: probably it can also be in-memory db for simplicity as it probably makes no difference in this scenario

zachelrath: @kibertoad Can you explain what you mean by "in-memory db" or point me to an example? Is that an internal-only Knex dialect that's only available in test contexts? I haven't an example of that in any other knex tests i've looked at.

kibertoad: from one of the knex tests:
```
    beforeEach(() => {
      connection = new sqlite3.Database(':memory:');
    });
```

bin/cli.js
```diff
@@ -58,7 +58,6 @@ function invoke(env) {
   env.modulePath = env.modulePath || env.knexpath || process.env.KNEX_PATH;
 
   const filetypes = ['js', 'coffee', 'ts', 'eg', 'ls'];
-  let pending = null;
```
kibertoad, line 61: it wasn't used for anything?

kibertoad:
@zachelrath Thanks a lot! Could you also update documentation at https://github.com/knex/documentation?

bin/cli.js
zachelrath: Yeah `pending` was assigned but was never used

test/cli/migrate-unlock.spec.js
zachelrath: I tried going down the in-memory path but it ended up complicating the test a lot more, as I had to create and tear-down a temporary Knexfile and  in order to have the CLI command write to the correct in-memory DB --- I couldn't get this to work from the CLI:

```execCommand(`node ${KNEX} migrate:unlock --client=sqlite3 --connection=:memory:`);```

Just using the test sqlite DB as the PR currently has in place right now works just fine and seems to be the least-complex option. 

Would you mind if we just retained the current approach?

zachelrath:
@kibertoad Docs PR: https://github.com/knex/documentation/pull/266

kibertoad:
Thanks a lot!

kibertoad:
Released in knex@0.21.1

PR #3829, opened 2020-04-26
Add support for .cjs files by default

dirty-cake:
This PR should solve the problem which I get after specified `"type": "module"` on the `package.json`. Since `--esm` flag doesn't work on node.js 14.0, the only way to use migration CLI with module packages is to specify all files as `.cjs`. But thay are doesn't supported by default, and there is no documented way to add it to the available list, but I was find one.


kibertoad:
Thank you!

kibertoad:
Released in knex@0.21.1

PR #3832, opened 2020-04-28
Allowed to set default schema from knexfile file

paul4156:


kibertoad:
@paul4156 Could you add test for this, and also documentation at https://github.com/knex/documentation ?

kibertoad:
Also I assume TS types need to be updated in https://github.com/knex/knex/blob/master/types/index.d.ts

paul4156:
OK sure.

kibertoad:
@paul4156 Any chance you could finish the PR?

lib/knex.js
```diff
@@ -54,6 +54,11 @@ function Knex(config) {
   if (config.userParams) {
     newKnex.userParams = config.userParams;
   }
+
+  if (config.defaultSchema) {
+    newKnex.withSchema(config.defaultSchema);
```
CodisRedding, line 59: As far as I can tell using `withSchema` on its own like this doesn't prepend the schema. At least not when printing out statements with debug on .

PR #3835, opened 2020-05-01
Implemented listTables method

gpetrov:
Implements a new method listTables which queries and returns a list of table names in the current database. See #360.

Improved upon the original PR https://github.com/knex/knex/pull/2560

Update: superseded by https://github.com/knex/knex-schema-inspector

### DIFF
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -269,6 +269,14 @@ module.exports = function (knex) {
       expect(knex.fn.now(6).toQuery()).to.equal('CURRENT_TIMESTAMP(6)');
     });
 
+    it('should list tables', () => {
+      return knex()
+        .listTables()
+        .then((res) => {
+          expect(res).to.be.an('array').that.includes('10_test_table');
+        });
+    });
+
     it('gets the columnInfo', function () {
       return knex('datatype_test')
         .columnInfo()
--- a/lib/dialects/mssql/query/compiler.js
+++ b/lib/dialects/mssql/query/compiler.js
@@ -214,7 +214,7 @@ class QueryCompiler_MSSQL extends QueryCompiler {
       sql: sql,
       bindings: bindings,
       output(resp) {
-        return resp.rows.map((table) => {
+        return resp.map((table) => {
           return table.TABLE_NAME;
         });
       },
--- a/lib/dialects/mssql/query/compiler.js
+++ b/lib/dialects/mssql/query/compiler.js
@@ -189,6 +189,38 @@ class QueryCompiler_MSSQL extends QueryCompiler {
     return 'with (HOLDLOCK)';
   }
 
+  // Compiles a `listTables` query
+  listTables() {
+    let schema = this.single.schema;
+
+    if (schema) {
+      schema = this.client.customWrapIdentifier(schema, identity);
+    }
+
+    let sql = `select table_name from information_schema.tables where `;
+    const bindings = [];
+
+    if (schema) {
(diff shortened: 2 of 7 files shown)


gpetrov:
There you go @kibertoad

kibertoad:
@gpetrov This looks great, thank you! Could you also please add integration test with schema for postgre and mssql? Also this needs to be documented, could you submit minimal documentation for https://github.com/knex/documentation?

test/integration/builder/additional.js
```diff
@@ -269,6 +269,14 @@ module.exports = function (knex) {
       expect(knex.fn.now(6).toQuery()).to.equal('CURRENT_TIMESTAMP(6)');
     });
 
+    it('should list tables', () => {
+      return knex()
+        .listTables()
+        .then((res) => {
```
kibertoad, line 275: you can use async/await for simplicity.

gpetrov:
After further development it turned out that more general schema retrieval function will be much more convenient. Those should return more the knex generic options and not the database specific (unlike the data type in columInfo)

Initially I just extended columnInfo - but as this method is already included in made test and its output is fixed - it might be wiser to add a complete new method

So a new proposal for changes:
| function | description|
|--|--|
|knex.schema.tables()| to get all tables, with properties
|knex.schema.views()| to get all views
|knex.schema.columns(table)| to get all columns for a table/view

**tables** should return:
|property| description|
|--|--|
|engine| the mySQL engine only
|schema| the database schema
|collation| the collation if any mySQL only
|comment| comment for the table if available

**columns** should return:
|property| description|
|--|--|
|type|knex type - so integer, string, text
|maxLength| max length of field 
|nullable|nullable
|defaultValue|the default value
|primary|is it a primary key
|increments|is it an autoincrement (created with increments)
|references|the knex references value of the field in other table
|inTable|the referenced table
|onDelete|delete rule if specified
|onUpdate|update rule if specified
|comment|the column comment

What do you think - @kibertoad ?

gpetrov:
Any feedback @kibertoad ? I don't want to do all the major rewrite without any confirmation or feedback as it is quite a lot of work.

funnylookinhat:
Bump @gpetrov @kibertoad - I'd like to see this merged :)

MattieBelt:
Any update on this, would love to have have an function to retrieve all the tables.

kibertoad:
@MattieBelt See https://github.com/knex/knex-schema-inspector
@rijkvanzanten How production ready do you think it already is?

rijkvanzanten:
We don't have a full test suite yet, so it's always a little tricky to promise things at that stage, but generally speaking all methods are written for all DB vendors and are working in my own (in production) projects 👍

kibertoad:
@rijkvanzanten Btw, maybe you could submit a PR with a mention of `knex-schema-inspector` for knex readme? Probably more people would realize they need it in their life that way.

gpetrov:
@kibertoad what do you want to do this with PR?

I do have integrated much better versions of `listTables` and `columnInfo` for all databases on my fork that return rich information.
But now I feel we shouldn't change the initial `listTables` and `columnInfo` as many tests and dependencies depend on their simple output. And are also light to execute

Maybe we can create additional versions like `listTablesEx` and `columnInfoEx` ? That return the richer object info but are also heavier to execute.

In the same part I would like also to continue with listViews and views creation and maybe `listPrimaryKeys` and `listForeignKeys`

julianwagle:
Hey I see this PR was not merged, was there anything with similar functionality that was merged instead?

Issue #3834, opened 2020-04-30
Custom migration source is always overwritten with FsMigrations


KrishnaPG:
# Environment

Knex version: 0.21.1
Database + version: SQLite 3
OS: Windows 10

# Bug

When using a customer migration source, the call to `make` method is not using the supplied custom source. Rather it always goes back to the built-in `FsMigrations`.

After debugging, the culprit seems to be at this line: https://github.com/knex/knex/blob/fe8340a6274202b5c59cb623f3e387aa7801f569/lib/migrate/Migrator.js#L298 

The correct code should be:
````
  make(name, config) {
    return this.generator.make(name, config);  //<- should send config, not this.config
  }
````
### Background
To give a background, I am creating a memory-backed migration source that does *not* use file-system. But unable to make the Knex use it, since the `FsMigrations` is always kicking-in.

That is because down the line in the series of calls, a call to `getMergedConfig(this.config, this.config)` ends up clearing the user-supplied custom migration source and resetting it to the built-in `FsMigrations`. Here is that part of the code, just for reference: 
https://github.com/knex/knex/blob/fe8340a6274202b5c59cb623f3e387aa7801f569/lib/migrate/configuration-merger.js#L27-L42

elhigu:
Please add reproduction code. IIRC there is a test for non-FsMigrations based migration source.

KrishnaPG:
Please try the below to reproduce the bug:
````

class MigrationSource {
	getMigrations() {
		console.log("getMigrations got called: ");
		return Promise.resolve(['migration1'])
	}
	getMigrationName(migration) {
		console.log("getMigrationName: ", migration);
		return migration;
	}
	getMigration(migration) {
		console.log("getMigration got called with: ", migration);
	}
};

const migrationSource = new MigrationSource();

const knex = require('knex')({
	client: 'sqlite3',
	connection: { filename: './.db/db.sqlite' },
	migrations: { migrationSource }
});

knex.migrate.make("test1", {  });
````
If it is working, then you should see console log output.

KrishnaPG:
Also, after struggling for the past few days working on this custom migration source implementation, it turns out that this whole API section needs complete refactoring. 

For example, the `getMigrations()` method supports promises, but the `getMigrationName()` and `getMigration()` does not support promises, nor even callbacks (and rather they expect the result to be returned from these API calls in synchronous fashion) !!

kibertoad:
@KrishnaPG I finally have some time, will take a look today.

[...]

PR #3838, opened 2020-05-05
Refactor migration logic to use async/await. Add test case for using classes with custom migration sources

kibertoad:
refs #3834

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -49,7 +49,11 @@ async function initKnex(env, opts) {
     ? await openKnexfile(env.configPath)
     : mkConfigObj(opts);
 
-  const resolvedConfig = resolveEnvironmentConfig(opts, env.configuration, env.configPath);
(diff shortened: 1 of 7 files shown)


kibertoad:
@KrishnaPG To address the specific problem that you were having - you were using it in a different way than Knex expects you to - you were returning list of migrations as an object and not as a list migration names. However unintuitive that might be, changing this signature now would be a breaking change. I'll revise our documentation to see how we can make it more clear.

Speaking of supporting asynchronous resolution of migration names and specific migrations - it is less straightforward than one might think. I did some early refactoring to simplify doing that later, but it still has ways to go. If you have some time to contribute, you can take a look at lib\migrate\Migrator.js -> as you can see, there are quite a bit of cases when we are calling these methods in some kind of callback cycle, and these are beasts for use-cases with promises.

PR #3839, opened 2020-05-05
Warn user if custom migration source is being reset

kibertoad:
refs #3834

### DIFF
--- a/lib/migrate/Migrator.js
+++ b/lib/migrate/Migrator.js
@@ -43,7 +43,10 @@ class Migrator {
       this.knex.userParams = this.knex.userParams || {};
     }
 
-    this.config = getMergedConfig(this.knex.client.config.migrations);
+    this.config = getMergedConfig(
+      this.knex.client.config.migrations,
+      this.knex.client.logger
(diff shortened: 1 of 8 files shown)


kibertoad:
@KrishnaPG As you can see in this PR, this functionality should be working. The only case when your custom migration source would be overriden is if you kept any of fs-related options intact. New version would display an error in such case.

lib/migrate/Migrator.js
```diff
@@ -43,7 +43,10 @@ class Migrator {
       this.knex.userParams = this.knex.userParams || {};
     }
 
-    this.config = getMergedConfig(this.knex.client.config.migrations);
+    this.config = getMergedConfig(
+      this.knex.client.config.migrations,
+      this.knex.client.logger
```
anandnimkar, line 48: The second argument here needs to be `undefined`, not `this.knex.client.logger`

kibertoad: Sorry for not replying before, but yeah, that was a keen observation, thank you for it.

PR #3840, opened 2020-05-05
Add development Docker configuration

kibertoad:
It's not very convenient to delete all the unneeded db configs from the main Docker config (I'm looking at you, Oracle), so I created a dev one.


scripts/docker-compose-dev.yml
```diff
@@ -0,0 +1,23 @@
+version: '3'
+
+services:
```
elhigu, line 3: I don't think there is a need for additional file... with docker compose one can select which services should be started.

kibertoad: oh, that is nifty. will adjust npm script and remove the file then

package.json
```diff
@@ -25,6 +25,8 @@
     "test:cli": "cross-env KNEX_PATH=../knex.js KNEX=bin/cli.js jake -f test/jake/Jakefile",
     "db:start": "docker-compose -f scripts/docker-compose.yml up --build -d mysql oracledbxe postgres mssql; docker-compose -f scripts/docker-compose.yml up initmssqlknexdb waitmysql waitpostgres waitoracledbxe",
     "db:stop": "docker-compose -f scripts/docker-compose.yml down",
+    "db:start:dev": "docker-compose -f scripts/docker-compose-dev.yml up -d",
```
elhigu, line 28: I suppose this does the same thing without an additional docker-compose file
```
docker-compose -f scripts/docker-compose.yml up --build -d postgres && docker-compose -f scripts/docker-compose.yml up waitpostgres
```

Also naming could be db:start:postgres

kibertoad: I get this error when I try to run your command:
```
Creating scripts_waitpostgres_1 ... done
Attaching to scripts_waitpostgres_1
waitpostgres_1     | psql: error: could not connect to server: could not connect to server: Connection refused
waitpostgres_1     |    Is the server running on host "postgres" (172.19.0.2) and accepting
waitpostgres_1     |    TCP/IP connections on port 5432?
waitpostgres_1     |  ?column?
waitpostgres_1     | ----------
waitpostgres_1     |         1
waitpostgres_1     | (1 row)
waitpostgres_1     |
scripts_waitpostgres_1 exited with code 0
```

Any ideas?

kibertoad: Also `docker-compose -f scripts/docker-compose.yml down postgres && docker-compose -f scripts/docker-compose.yml down waitpostgres` doesn't work, what would be the right way to shut it down?

elhigu: actually I don't think I have ever tried to shutdown only a single service.... maybe docker-compose stop postgres is enough?

elhigu:
```
Creating scripts_waitpostgres_1 ... done
Attaching to scripts_waitpostgres_1
waitpostgres_1     | psql: error: could not connect to server: could not connect to server: Connection refused
waitpostgres_1     |    Is the server running on host "postgres" (172.19.0.2) and accepting
waitpostgres_1     |    TCP/IP connections on port 5432?
waitpostgres_1     |  ?column?
waitpostgres_1     | ----------
waitpostgres_1     |         1
waitpostgres_1     | (1 row)
waitpostgres_1     |
scripts_waitpostgres_1 exited with code 0
```

Looks like correct output. Script tries to connect once and it fails and the second time query is sent to server with success so that waiting docker container exists.

package.json
elhigu: or even `docker-compose -f scripts/docker-compose.yml rm -fs postgres` which stops and removes the container.

kibertoad: apparently docker-compose is smart enough, stopping whole script manages to identify what is running and stop just that.

PR #3845, opened 2020-05-07
Update README to reflect correct handling of native map function

skreis:
There is an example in the README which presents `.map` as exposed with the knex library.

This had previously been valid when bluebird functions were leaked in the library and documentation.

Further described in https://github.com/knex/knex/issues/3704#issuecomment-595782683


kibertoad:
Thank you!

skreis:
@kibertoad you're very welcome, have a great day!

PR #3846, opened 2020-05-08
fix typings for raw querybuilders

maximelkin:
bindings for whereRaw/groupByRaw etc should be one object or array, not vararg


lorefnon:
Thanks for fixing this. The previous bindings were incorrect.

PR #2493, opened 2018-02-19
Added the removal of a columns default constraint, before dropping the column

tetsupanda:
PR targeting issue #2492, which is specific to MSSQL.

PR #3855, opened 2020-05-15
Added the removal of a columns default constraint, before dropping the column (MSSQL), with integration tests

### LINKED ISSUE #2492: Dropping MSSQL column with default constraint does not work
# Environment

Knex version: 0.14.4
Database + version: MSSQL

# Bug

When attempting to drop a column (via a migration) with a MSSQL db, an error is returned.  After digging into this, the error is that the column cannot be dropped because it has a default constraint.  The default constraint has to be dropped first, which does not seem to be currently supported.

2. Error message (in Node.js): TransactionError: Transaction has been aborted.

When running the generated SQL (ALTER TABLE [Accounts] DROP COLUMN [IsAccountDisabled]) manually in SQL server, this error is shown:

Msg 5074, Level 16, State 1, Line 12
The object 'DF__Accounts__IsAcco__56E8E7AB' is dependent on column 'IsAccountDisabled'.
Msg 4922, Level 16, State 9, Line 12
ALTER TABLE DROP COLUMN IsAccountDisabled failed because one or more objects access this column

Here is the migration raising the error (fails when running the rollback; default constraint created in up migration):

const TABLE_NAME = 'Accounts';
const COLUMN_NAME = 'IsAccountDisabled';

    return knex.schema.alterTable(TABLE_NAME, t => {
        t.boolean(COLUMN_NAME).notNullable().default(false);
      });
};

    return knex.schema.hasColumn(TABLE_NAME, COLUMN_NAME).then(exists => {
        console.log(`Exists: ${exists}`);
    
        if (exists) {
            return knex.schema.table(TABLE_NAME, t => {
              t.dropColumn(COLUMN_NAME);
            });
        } else {
          return;
        }
      })
};

C0Nd3Mnd:
Basically the same as #2493, just with integration tests added (and some code formatting).

Fixes #2492.

### DIFF
--- a/test/integration/migrate/migration-integration-tests.js
+++ b/test/integration/migrate/migration-integration-tests.js
@@ -649,6 +649,19 @@ module.exports = function (knex) {
               });
           });
       });
+
+      it('should drop a column with a default constraint (mssql)', function () {
+        return knex.migrate
+          .latest({
+            directory: 'test/integration/migrate/drop-with-default-constraint',
+          })
+          .then(() => {
+            return knex.migrate.rollback({
+              directory:
+                'test/integration/migrate/drop-with-default-constraint',
+            });
+          });
+      });
     });
 
     describe('knex.migrate.down', () => {
--- a/test/integration/migrate/migration-integration-tests.js
+++ b/test/integration/migrate/migration-integration-tests.js
@@ -650,17 +650,14 @@ module.exports = function (knex) {
           });
       });
 
-      it('should drop a column with a default constraint (mssql)', function () {
-        return knex.migrate
-          .latest({
-            directory: 'test/integration/migrate/drop-with-default-constraint',
-          })
-          .then(() => {
-            return knex.migrate.rollback({
-              directory:
-                'test/integration/migrate/drop-with-default-constraint',
-            });
-          });
+      it('should drop a column with a default constraint (mssql)', async () => {
+        await knex.migrate.latest({
+          directory: 'test/integration/migrate/drop-with-default-constraint',
+        });
+
+        await knex.migrate.rollback({
+          directory: 'test/integration/migrate/drop-with-default-constraint',
+        });
       });
     });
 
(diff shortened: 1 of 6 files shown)


test/integration/migrate/migration-integration-tests.js
```diff
@@ -649,6 +649,19 @@ module.exports = function (knex) {
               });
           });
       });
+
+      it('should drop a column with a default constraint (mssql)', function () {
+        return knex.migrate
```
kibertoad, line 654: nit: you can use async-await here

C0Nd3Mnd: Oh, true. I just looked at how another test was done. Fixed.

C0Nd3Mnd:
I had to update a test that was affected by this change. It should be working fine now.

kibertoad:
Thanks a lot!

kibertoad:
@C0Nd3Mnd How urgently do you need version with this fix released?

C0Nd3Mnd:
@kibertoad We have workarounds for this problem in our migration scripts right now, so it works as-is.

It would be nice to be able to remove these workarounds soon though, so a version with this fix in it in the following days/weeks would be greatly appreciated!

NGPixel:
@kibertoad Any update of when this fix will be released?

kibertoad:
@NGPixel Thanks you for a ping! Will try to get it out today, worst case scenario - tomorrow.

kibertoad:
@NGPixel @C0Nd3Mnd Sorry for the delay. Published 0.21.2. Please let me know if it works!

KrisLau:
@kibertoad Doesn't seem to be working in the latest knex version? Constraint on the column:
image

Migration file: 
```js
export const up = async knex => {
  await knex.schema.alterTable('user', table => {
    table.integer('email_type').defaultTo(0);
  });
  await knex.schema.alterTable('team', table => {
    table.boolean('allow_fans').defaultTo(false);
  });
};

export const down = async knex => {
  await knex.schema.table('team', table => {
    table.dropColumn('allow_fans');
  });
  await knex.schema.table('user', table => {
    table.dropColumn('email_type');
  });
};
```

PR #3857, opened 2020-05-18
Add support for AWS Aurora MySQL Data API

txase:
AWS Aurora supports an HTTP Data API that proxies SQL queries as HTTP requests (including transactions!). This PR shows how to create a new dialect to support the Aurora Data Api.

I'm in favor of merging this, but I understand the interest in limiting new dialects built into knex. I'd be fine with maintaining this separately, as suggested in the contributing guidelines, but it's unclear how a complex dialect like this can be instantiated outside of knex itself. I can't think of a way to tell knex to load the `mysql` dialect but with a different implementation and configuration format. Please advise on whether this should be merged into knex or how it could be loaded separately.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -7,6 +7,7 @@ const CLIENT_ALIASES = Object.freeze({
 
 const SUPPORTED_CLIENTS = Object.freeze(
   [
+    'aurora-data-mysql',
     'mssql',
     'mysql',
     'mysql2',
--- a/lib/dialects/aurora-data-mysql/index.js
+++ b/lib/dialects/aurora-data-mysql/index.js
@@ -1,0 +1,256 @@
+// AWS Aurora MySQL Data API Client
+// -------
+const Client_MySQL = require('../mysql');
+const Transaction = require('./transaction');
+
+function getAuroraDataValue(value) {
+  if ('blobValue' in value) {
+    return Buffer.from(value.blobValue, 'base64');
+  } else if ('booleanValue' in value) {
+    return value.booleanValue;
+  } else if ('doubleValue' in value) {
+    return value.doubleValue;
+  } else if ('isNull' in value) {
+    return null;
+  } else if ('longValue' in value) {
+    return value.longValue;
+  } else if ('stringValue' in value) {
+    return value.stringValue;
+  } else {
+    const type = Object.keys(value)[0];
+    throw new Error(`Unknown value type '${type}' from row`);
+  }
+}
+
+function hydrateRecord(record, fields) {
+  return record.reduce((row, value, index) => {
+    const field = fields[index];
+
+    value = getAuroraDataValue(value);
+
+    switch (field.typeName) {
+      case 'DECIMAL':
+        value = Number(value);
+        break;
+
+      case 'DATE':
+      case 'DATETIME':
+      case 'TIMESTAMP':
+      case 'YEAR':
+        value = new Date(value + 'Z');
+        break;
+
+      case 'CHAR':
+        if (field.precision === 5) {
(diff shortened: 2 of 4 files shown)


elhigu:
Probably would be a good idea to create a template project for writing those external dialects, showing how to do the basic inheriting the classes and setting up peer dependencies etc.

txase:
I don't see a way to do it without it being really hacky. Like, creating a knex object with the 'mysql' dialect and then overwriting a bunch of the object's prototype chain. Am I missing something?

kibertoad:
@txase Is there a drop-in mock replacement for AWS rdsdataservice that could be used for automated test purposes? If we are to consider making an exception here, new dialect needs to be runnable in CI and pass the existing testing suite (potentially omitting some tests as not relevant).

txase:
@kibertoad It's possible to mock the AWS SDK, but it would be simply testing that the right AWS SDK calls are made with the right values. There's no way to run a local version of an Aurora Data API that saves values to a sqlite table, for example.

There are a lot of tests, and I couldn't figure out which ones make the most sense to try to write for this kind of dialect, where the dialect syntax is already covered and there's no way to run a fake version of the Data API service to test against. If you could point to existing tests that make sense to replicate for this dialect I'd be happy to take a look at doing so.

Thanks!

kibertoad:
@txase In general we prefer to have a shared set of tests ran across multiple dialects, to make sure that dialects are compliant with how knex is doing things and overall work correctly. If that is not possible for this particular dialect, value of having it upstream is diminished, as we can get regressions at any time without someone noticing it until encountered in production.

How difficult would it be to implement a pass-through version of `Aurora Data API` that just operates on a single instance of MySQL and proxies calls to MySQL driver? If that is not feasible, we could help you to setup a separate npm package for a drop-in dialect. If some kind of extensions points within knex codebase would make that easier, we'd be happy to discuss and implement something.

elhigu:
> Like, creating a knex object with the 'mysql' dialect and then overwriting a bunch of the object's prototype chain.

Just inheriting dialect from another, the same way that knex internally inherits different variants of dialects.

txase:
@elhigu Can you provide an example of how this would work? It's not possible (as far as I can tell) to create a different dialect and load it because there is a hardcoded set of dialects that knex supports: https://github.com/knex/knex/blob/master/lib/constants.js#L8-L18.

You also can't access a dialect's class with knex without reaching into internals. The documented API only allows for requesting a dialect by name as the `client` property of the configuration. Without direct access to the class, a separate module can't inherit it through a public API.

elhigu:
> You also can't access a dialect's class with knex without reaching into internals. The documented API only allows for requesting a dialect by name as the client property of the configuration. Without direct access to the class, a separate module can't inherit it through a public API.

Yep external dialects must require base class implementations etc. from knex npm package and use knex internals and must be kept uptodate  possibly on every patch update of knex (so they should be released in worst case for each exact knex version). Though those interfaces are not updated very often so it is not as bad as it sounds.

Though keeping up external dialect can be easier than knex core dialect if there is test suite that is better suited to test only necessary parts of the new dialect instead of adding that dialect to be included in all tests of knex... Though in addition of having template project for external dialects knex could also have some external integration test suite for those to make sure that basic stuff like transactions (with different locking) etc. are working as expected for every external dialect.

For passing external dialect to knex config you should do this (as mentioned in contributing.md):

```
  client: require('knex-aurora-dialect'),
  connection : {...}
});
```

txase:
I managed to create an independent driver: [knex-aurora-data-api-mysql](https://www.npmjs.com/package/knex-aurora-data-api-mysql). Closing this issue now :).

PR #3865, opened 2020-05-22
Prefer `void` as return type on migration generator ts stub

vcanales:
The liberal use of `any` is frowned upon by some default `tslint` configurations. That combined with the fact the these functions, as far as I know, don't _need_ to return anything, makes it reasonable to use `void` over `any` for the return type.

If this change is accepted, I could also update docs to reflect that returning on `up` and `down` functions is not necessary, unless I'm missing something!

### DIFF
--- a/lib/migrate/stub/ts.stub
+++ b/lib/migrate/stub/ts.stub
@@ -8,7 +8,7 @@ export async function up(knex: Knex): Promise<Knex.SchemaBuilder> {
     });
 }
 <% } else { %>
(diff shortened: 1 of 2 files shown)


kibertoad:
Thanks!

kibertoad:
@vcanales `returning on up and down functions is not necessary, unless I'm missing something!` Only if they are async. In general you should return a promise, whether it is implicit or explicit one.

PR #3873, opened 2020-06-06
feat: new config parameter / CLI flag to prefixing seed filename with a timestamp

### LINKED ISSUE #3623: Prefix seeds with timestamp like migrations

When using the knex cli, you can run `knex migrate:make $name` and it'll generate a `$timestamp_$name.js` in the migrations directory. Is it possible to do the same for the seeds? Because when I run `knex seed:make $name` it generates a `$name.js` _without_ the timestamp and I have to add it in manually :/

christianodejesus:
I've created the new timestampFilenamePrefix seeder config parameter and --timestamp-filename-prefix CLI flag to prefixing seed filenames with a timestamp, like migration files.

Documentation PR: https://github.com/knex/documentation/pull/271

Close #3623.


test/cli/seed-make.spec.js
```diff
@@ -378,4 +379,75 @@ development: {
...
+    beforeEach(() => {
+      fileHelper = setupFileHelper();
+    });
+
+    afterEach(() => {
+      fileHelper.cleanup();
+    });
+
+    before(() => {
+      process.env.KNEX_PATH = '../knex.js';
```
kibertoad, line 397: can you reset it to pre-test value in `after` statement?

bin/cli.js
```diff
@@ -94,7 +94,11 @@ function invoke(env) {
       'environment, default: process.env.NODE_ENV || development'
     )
     .option('--esm', 'Enable ESM interop.')
-    .option('--specific [path]', 'Specify one seed file to execute.');
+    .option('--specific [path]', 'Specify one seed file to execute.')
+    .option(
+      '--timestamp-filename-prefix',
+      'Enable a timestamp prefix on name of generated seed files.'
```
kibertoad, line 100: `on name` -> `in the name`, I assume?

bin/cli.js
```diff
@@ -94,7 +94,11 @@ function invoke(env) {
       'environment, default: process.env.NODE_ENV || development'
     )
     .option('--esm', 'Enable ESM interop.')
-    .option('--specific [path]', 'Specify one seed file to execute.');
+    .option('--specific [path]', 'Specify one seed file to execute.')
+    .option(
```
kibertoad, line 98: Should this be a root level parameter?

lib/seed/Seeder.js
```diff
@@ -83,6 +84,11 @@ class Seeder {
 
   _getNewStubFileName(name) {
     if (name[0] === '-') name = name.slice(1);
+
+    if (this.config.timestampFilenamePrefix === true) {
+      name = yyyymmddhhmmss() + '_' + name;
```
kibertoad, line 89: It would be preferable to use string template

lib/util/timestamp.js
```diff
@@ -0,0 +1,21 @@
+// Ensure that we have 2 places for each of the date segments.
+function padDate(segment) {
+  segment = segment.toString();
+  return segment[1] ? segment : `0${segment}`;
```
kibertoad, line 4: Probably this can be used: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

test/cli/seed-make.spec.js
```diff
@@ -378,4 +379,75 @@ development: {
...
+      );
+
+      await execCommand(
+        `${NODE} ${KNEX} seed:make ${testFileName} --knexpath=../knex.js`,
+        {
+          expectedOutput: 'Created seed file',
+        }
+      );
+
+      const fileCount = fileHelper.fileGlobExists(seedGlobPath);
```
kibertoad, line 448: Isn't this going to be flaky due to slight mismatch of time between setting testFileName and generating actual stub? Wonder if we could mock timer to make this test deterministic or use less strict glob that ignores seconds part.

lib/seed/Seeder.js
christianodejesus: Yes, I was going to use string template, but I decided to just keep it the way it was.

lib/util/timestamp.js
christianodejesus: I only moved this code from lib/migrate/MigrationGenerator.js to here, to not change the logic and possibly causes any problem.

test/cli/seed-make.spec.js
christianodejesus: Ok

bin/cli.js
christianodejesus: I think not. But I put the new parameter where the "--specific" was, and this one only affect seeds too.

bin/cli.js
christianodejesus: Yes

test/cli/seed-make.spec.js
christianodejesus: I put a timestamp in the end of filename to eliminate chances of create files with same name between tests. I not testing that this timestamp value will be the same that the new option generates. But I will change these tests to more clearly logic.

bin/cli.js
christianodejesus: I tested it now, if I remove the parameter from root level, it will not be shown on the help command.

kibertoad:
@christianodejesus Sorry for the delay in merging this. If you could resolve conflicts, I'll merge.

christianodejesus:
> @christianodejesus Sorry for the delay in merging this. If you could resolve conflicts, I'll merge.
@kibertoad I did a merge from knex/master and conflicts are resolved. Ok!?

kibertoad (approved):

kibertoad:
Thanks a lot!

PR #3885, opened 2020-06-15
Add .timeout(ms) to .raw()'s typescript typings

tomi:
`.timeout(ms)` is [supported for `knex.raw()`](https://github.com/knex/knex/blob/master/lib/raw.js#L46), but it's missing from the TS typings.


kibertoad:
Released in 0.21.3

PR #3890, opened 2020-06-19
Provided fix for long-running timeout error

bashirk:
Provided a fix for the production deployment bug:
`KnexTimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?`

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -238,7 +238,7 @@ Object.assign(Client.prototype, {
   },
 
   poolDefaults() {
-    return { min: 2, max: 10, propagateCreateError: false };
+    return { min: 2, max: 10, propagateCreateError: true };
   },
 
   getPoolSettings(poolConfig) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -238,7 +238,7 @@ Object.assign(Client.prototype, {
   },
 
   poolDefaults() {
-    return { min: 2, max: 10, propagateCreateError: true };
+    return { min: 2, max: 10, propagateCreateError: false };
   },
 
   getPoolSettings(poolConfig) {


kibertoad:
@elhigu We really need a comment with explanation in that part of the code. Probably it's the fourth or fifth time this comes up :)

kibertoad:
@bashirk Are you sure this fixes anything and does not just obscure the problem?

bashirk:
@kibertoad yeah, it does obscure the "trx error" problem literally

elhigu:
> @elhigu We really need a comment with explanation in that part of the code. Probably it's the fourth or fifth time this comes up :)

I thought that `pool timeout error` it would have been explained with greater details in various places various times :) And that propagate error option, which must _*never*_ be used with knex also... IIRC I had added comment about it in the code too, but maybe not.

If someone likes to update documentation about both of those things I could write some draft about both things. 

About the propagate error:

```
propagateCreateError : false
```

pool option must never be used with knex or it will cause a loop where acquiring new connection fails and pool would just keep on retrying until some other (for example acquireConnectionTimout) timeout kicks in. 

If error is not propagated, for example problems, where connection to database is lost or credentials are wrong will cause knex to poll / hang until long timeout. Actually in that case one would like to get error immediately to hand and for example start serving error pages to clients instead of starting to accumulate waiting page requests which then all will fail after time out. 

Knex is _*not designed to work*_ with tarn.js with that option disabled.

bashirk:
I clearly understand what you've stated, and how disabling the option might cause something to break in tarn.js. Cheers!

Issue #3870, opened 2020-06-03
postProcessResponse not called for streamed queries ?


mlarcher:
# Environment

Knex version: 0.21.1
Database + version: postgres 11
OS: node:14.2.0 docker container

# Feature discussion / request

It seems that `postProcessResponse` is not called on the resulting rows when using `.stream()` on a query. Can someone tell me if this is a bug or the intended behavior, and if there is a known workaround by any chance ?

elhigu:
It is not intentional and I don't believe that there is any workaround for it.

mlarcher:
Thank you for the info @elhigu 
I guess it counts as a bug then, as it was quite unexpected.

sirfilip:
I can pick this one up

kibertoad:
Would be awesome!

sirfilip:
Quick question is it ok if i add test coverage and can go with jest or mocha it is all the same to me, please tell me what is the project preference ?

kibertoad:
@sirfilip Test coverage is actually mandatory for PRs with rare exceptions. Take a look at existing integration tests, they use mocha.

sirfilip:
Thanks @kibertoad  looks like i need to get something from my sight :). Now all makes sence. I am on it!

sirfilip:
^^^ it is messy not a real pr just want to make sure that i am going in the right direction. It looks like i will have to wrap each row of the stream inside the client.postProcessResponse with the row and connection as context i think that i am maybe missing someting

cubikrichard:
@sirfilip Did your fork need more work before creating a PR? Is there anything I can help with to keep it moving?

sirfilip:
Sorry for the late response what i submitted was a direction that i took to resolve this issue and it feels a bit messy so not really sure if i should continue with the implementation for the rest of the drivers. But if it looks ok from your side i can continue and implement this for the other drivers quickly, probably in the following days. Please letme know. I dont have any problem with discarding the code cause i really hope there is a cleaner way to implement this.

cubikrichard:
I used your fork as inspiration, but, it was possible to hook into the stream at a higher level, so to avoid having to implement it for each driver. 

`lib/runner.js` https://github.com/knex/knex/pull/3931/files

sirfilip:
That is awesome! I like your approach by far better then the one i took. I vote for your proposal :). Great job !!!

andreaspalsson:
[...]

PR #3893, opened 2020-06-23
#3870 Fix proposal

sirfilip:


kibertoad:
@sirfilip Should this work only for SQLite? Other DBs are failing now.

sirfilip:
That is correct it is implemented for sqlite only will add it to the other dialects hopefully this week

kibertoad:
Much appreciated!

kibertoad:
@sirfilip Is there anything I could help with?

sirfilip:
Sorry for the delay got new laptop working on it tonight

sirfilip:
Guess will need some help with pg cause it is using the `pg-query-stream` package so will need to either dig in into the src or maybe can give some heads up :) ?

sirfilip:
Hmm this approach is not going to work cause each dialect has own streaming support, will take some time to solve each one of them which is telling me maybe the whole approach is wrong and that i should look for a way to call `postProcessResponse` later in the stack and dodge the need to deal with all of the dialects, so will spend some time lnvestigating this.

kibertoad:
@sirfilip That makes sense. Please ping me if you'll need any help or have any questions.

sirfilip:
Will have to pause the development from my side cause getting overwhelmed with something else people, can someone please take this over ?

PR #3899, opened 2020-06-28
Escape single quoted values passed to defaultTo function

### LINKED ISSUE #3898: single quoted value passed to `defaultTo` method is not escaped properly
# Environment

Knex version: da54cf1ecf0acef4b3d3d51cd2656e4faf10d3e9
Database + version: sqlite3@4.2.0
OS: Gnu/Linux

# Bug

1. Explain what kind of behavior you are getting and how you think it should do
  - value with single quotes passed to `defaultTo` function is not escaped properly for Sqlite3 database.

2. Error message
  - Syntax error from Sqlite

3. Reduced test code
  ```javascript
  it('correctly escape singleQuotes passed to defaultTo()', function () {
    tableSql = client
      .schemaBuilder()
      .table('users', function (table) {
        table.string('foo').defaultTo("single 'quoted' value");
      })
      .toSQL();

    equal(
      tableSql[0].sql,
      "alter table `users` add column `foo` varchar(255) default 'single ''quoted'' value'"
    );
  });
  ```

  - I was facing this issue for Sqlite3 dialect . I think, this bug is only applicable to Sqlite

harish2704:
Fixes #3898

Signed-off-by: Harish Karumuthil <harish2704@gmail.com>


kibertoad:
Thanks!

Issue #2324, opened 2017-11-13
stream.end() does not return connection to pool - race condition


rkaw92:
I'm currently using the stream API on Node v8.1.4 with knex 0.14.0.

It seems that the connection is correctly returned to the pool when I read the entirety of the stream's contents. However, if I don't reach the end of the stream, but instead use `stream.end()` immediately (or do nothing, it really doesn't seem to matter), the connection is never returned to the pool.

This is contrary to [the wiki](https://github.com/tgriesser/knex/wiki/Manually-Closing-Streams), which states that one should call `.end()`  to return the connection to the pool, which it does not seem to be doing in my case.

Consider this trivial example (assuming `CREATE TABLE items (id SERIAL);`, and it **needs to be empty**, or otherwise see *Note 1*):
```js
const db = require('knex')({
	client: 'pg',
	connection: 'postgres://postgres:postgres@127.0.0.1/postgres'
});

const queryStream1 = db('items').select('*').stream();
queryStream1.on('data', function() {});
//.end() commented out for now - everything works as expected
//queryStream1.end();
```

Now, uncomment the last line (`.end()`) - this does not return the connection to the pool. Moreover, it actually *prevents the connection from being returned*!

Upon inspecting `Client_PG#_stream`, it seems that this is a race condition. Here is what I think is happening:
* User calls `.stream()` - this is implemented as `Runner#stream`
  * A stream object is created using `new PassThrough({objectMode: true});`
  * This line runs: `const promise = Promise.using(this.ensureConnection(), function(connection) {`
    * When `ensureConnection()` fulfills, the function will be entered entered and it will call `runner.client.stream(runner.connection, sql, stream, options);` - however, this **does not happen just yet**, as we need to wait for the ensureConnection promise
  * The stream is returned
* The user calls `.end()` on the stream, which causes it to emit event **end** immediately, so it will not be emitted again in the future
* Now, the `runner.client.stream(...)` invocation is actually made, and the underlying dialect's client registers an `.on('end')` handler on the stream (in case of PostgreSQL, it's done [here](https://github.com/tgriesser/knex/blob/0.14.0/src/dialects/postgres/index.js#L187)). However, **it is too late**, as the **end** event has already been emitted and we can no longer detect the end of the stream here.
* Thus, the Promise returned from underlying driver's `_stream` invocation is never resolved nor rejected, and the connection never returns to the pool.

[...]

PR #3900, opened 2020-06-28
Ensure DB stream is destroyed when the PassThrough is destroyed (#2324).

issa-tseng:
there are a great many details [in the comments i left on this ticket](https://github.com/knex/knex/issues/2324#issuecomment-643053327) on why this change is necessary.

in the end, because y'all support back to node 10, which does not have native `streams.pipeline`, i went with this approach. the other advantage here is that it solves the problem at its source, for all adapters using this code.

i went to try to write a test for this, but i spent a few hours and had a lot of trouble getting them to run at all. some but not all of the unit tests run totally happily, but none of the integration or cli tests really work. seems there are four or five separate issues with them that i am having trouble getting to the bottom of. and it seems like streams aren't really all that well tested at all to begin with, so i figured i'd at least just submit this pr.

i am still real happy to write a test here, but i'm going to need some help getting the tests to work if i'm going to do that.


issa-tseng:
it looks like the tests are failing to run on CI due to a typescript dependency issue.

issa-tseng:
rebasing over master to see if that resolves the test issues.

lib/runner.js
```diff
@@ -77,7 +77,16 @@ Object.assign(Runner.prototype, {
...
 
     const runner = this;
-    const stream = new PassThrough({ objectMode: true });
+    let queryStream;
+    const stream = new PassThrough({
+      objectMode: true,
+      destroy() {
+        if (!queryStream) return;
+        queryStream.destroy();
+        queryStream.emit('error', new Error('stream destroyed'));
```
petslane, line 86:
```suggestion
        queryStream.destroy(new Error('stream destroyed'));
```

Based on https://nodejs.org/api/stream.html#stream_readable_destroy_error you can just pass `Error` to `destroy()`

petslane: This should work, but not for `mysql2`. `queryStream` from `mysql2` is not initial readable stream and destroying  this stream does not destroy source stream, so for `mysql2` this still does not work. But this is `mysql2` [issue #1168](https://github.com/sidorares/node-mysql2/issues/1168).

But this PR change is correct. If `PassThrough` stream here is destroyed, then by default this does not destroy parent/source stream, so it must be destroyed separately, otherwise source stream will stay paused (because of unpipe). And it is up  to underlying dialect's client to also make sure that source stream is actually destroyed and query to database cancelled.

ducaale:
Regarding typescript errors, I think changing `lint:types` to use the locally installed version of typescript should fix it. Not sure if that is acceptable though.
```
dtslint --localTs node_modules/typescript/lib types
```

kibertoad:
@issa-tseng Sorry for the delay in getting back to this. I've synced your branch with master which, I believe, should no longer fail the CI, and added the extremely basic test for handling errors in streams. If CI passes, I'll merge this and release a new version already.

kibertoad:
@issa-tseng It is failing on Node 14/mssql for some reason, due to `queryStream.destroy` not being available. I added workaround for the time being, and opened https://github.com/tediousjs/tedious/issues/1139 just in case it's a problem on Tedious end, but maybe you have any ideas?

kibertoad:
@issa-tseng Merging it as-is to unblock current usage, but hopefully this can be followed-up in a separate PR.

PR #3910, opened 2020-07-02
Mssql decimal fix

hansnull:
Hi knex guys,

I have written a fix for large decimals in mssql, related to https://github.com/tediousjs/tedious/issues/1058.
My fix adds dynamic scaling for decimal values and prevents an UInt64 overflow in tedious.

Best regards,
hansnull


lib/dialects/mssql/index.js
```diff
@@ -16,6 +16,7 @@ const ColumnCompiler = require('./schema/columncompiler');
 
 const SQL_INT4 = { MIN: -2147483648, MAX: 2147483647 };
 const SQL_BIGINT_SAFE = { MIN: -9007199254740991, MAX: 9007199254740991 };
+const MAX_UINT64 = 18446744073709551615;
```
elhigu, line 19: How did you get that numerical value? You know you cannot represent that big number exactly in javascript without using some special encoding like string or loosing precision.
```
> 18446744073709551615
18446744073709552000
> 
```

hansnull: Thats true... do you want me to include https://github.com/MikeMcl/bignumber.js for the check ?

elhigu:
Please try to explain super clearly what is the problem you are now trying to solve here.  What do you mean by dynamic scaling and why it is needed / wanted.

hansnull:
I tried to store the following decimal value with knex and MSSQL: 10000000000005.74
No error was thrown, but MSSQL stores actually: -20037647.7924
The root cause for that is the following line:
https://github.com/tediousjs/tedious/blob/4120817754e91a32d9419093770771df590365ca/src/data-types/decimal.ts#L75
My fix calculates the scale and precision dynamically which fixes the problem for the number 10000000000005.74
To make sure that no other value overflows the UINT64 I added the check.

elhigu:
Is this trying to recognize that if there is bigint coming from DB and it is too big to store in javascript number that error should not be thrown? 🤔

hansnull:
Because javascript numbers are actually float values, the check is not necessary... I will remove it.

elhigu:
If the root cause is in tedious:

> https://github.com/tediousjs/tedious/blob/4120817754e91a32d9419093770771df590365ca/src/data-types/decimal.ts#L75

Cannot it be fixed in tedious if the root cause is there?

hansnull:
It is also related to knex, because the default for the scale is 10 and for precision 38, which is not needed for a number like 10.01

hansnull:
> 
> 
> > https://github.com/tediousjs/tedious/blob/4120817754e91a32d9419093770771df590365ca/src/data-types/decimal.ts#L75
> 
> Cannot it be fixed in tedious if the root cause is there?

I don't think so... see https://github.com/tediousjs/tedious/issues/1058

hansnull:
You can use my added unit test to compare the master version with my fix.

lib/dialects/mssql/index.js
```diff
@@ -316,7 +316,17 @@ Object.assign(Client_MSSQL.prototype, {
...
+        let scale = 10;
+        const bindingParts = binding.toString().split('.');
+        if (bindingParts.length === 2) {
+          scale = Math.min(bindingParts[1].length, scale);
+          precision = Math.min(
+            bindingParts[0].length + bindingParts[1].length,
+            precision
+          );
+        }
+        req.input(`p${i}`, this.driver.Decimal(precision, scale), binding);
```
elhigu, line 329: How about not trying to write javascript number (double) as a decimal, but as `this.driver.Float(53)` to try to write exact value represented by `number` to the database?

If writing decimal to float type of column in this way works, I don't see why other way around wouldn't work even better (without losing any precicion). 

If one really wants to write decimals to DB I suppose string or some other exact format should be used.

elhigu: Makes me wonder if currently writing floats is also broken here....

lib/dialects/mssql/index.js
elhigu: > > > https://github.com/tediousjs/tedious/blob/4120817754e91a32d9419093770771df590365ca/src/data-types/decimal.ts#L75
> > 
> > 
> > Cannot it be fixed in tedious if the root cause is there?
> 
> I don't think so... see [tediousjs/tedious#1058](https://github.com/tediousjs/tedious/issues/1058)

Actually now I investigated that tedious issue and few other related ones and it should be possible to fix in tedious (actually probably pretty easily by just writing value for example in 32bit pieces instead of trying to create one big value that might get too big...), but right now they just don't care to fix it since they want to write bigger reform how they are handling types. Anyways in this case if writing number always as `double` works it should be also more performant.

hansnull: Writing the numbers as double may work, but it would need a table column migration or am I wrong?

lib/dialects/mssql/index.js
hansnull: Do you want me to add more tests?

lib/dialects/mssql/index.js
elhigu: I don't think it should need any migrations at least currently writing to float table with decimals seems to work, so I don't believe it should work just fine also other way around.

lib/dialects/mssql/index.js
elhigu: Yes. I added comment about it to test case file.

elhigu: Yes. And to change this implementation to write 

```
req.input(`p${i}`, this.driver.Float(53), binding);
```

instead of `Decimal`.

hansnull: I will give it a try and add more tests

lib/dialects/mssql/index.js
hansnull: Writing floats works (see tests).

elhigu:
Everything else seems to work just fine, except for typescript tests (`npm run lint:types && npm run lint`)seems to be failing for some reason.

hansnull:
The failing tests seem not to be caused by my changes.

kibertoad:
@hansnull Could you please pull latest changes from master?

hansnull:
> 
> 
> @hansnull Could you please pull latest changes from master?

Done

kibertoad:
Released in 0.21.7

PR #3920, opened 2020-07-11
Fix getMergedConfig calls to include client logger

anandnimkar:
This fixes issues with logging that were introduced in this PR (https://github.com/knex/knex/pull/3839).

This fixes the following issue that I logged: https://github.com/knex/knex/issues/3919


kibertoad:
Thanks!

PR #3923, opened 2020-07-14
Fix for issue 3914

drew-royster:
Fix for https://github.com/knex/knex/issues/3914

Added unit tests for generateCombinedName

This broke these two tests
```
  1) Query Building Tests
       Oracle SchemaBuilder
         allows dropping a unique compound index with too long generated name:
  2) Query Building Tests
       OracleDb SchemaBuilder
         allows dropping a unique compound index with too long generated name:
```

Let me know how to proceed to remediate those issues in the way that matches the pattern you have laid out in other places.

Thanks!

### DIFF
--- a/lib/dialects/oracle/utils.js
+++ b/lib/dialects/oracle/utils.js
@@ -1,6 +1,6 @@
 function generateCombinedName(logger, postfix, name, subNames) {
   const crypto = require('crypto');
-  const limit = 30;
+  const limit = 128;
   if (!Array.isArray(subNames)) subNames = subNames ? [subNames] : [];
   const table = name.replace(/\.|-/g, '_');
   const subNamesPart = subNames.join('_');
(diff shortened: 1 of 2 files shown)


kibertoad:
Better solution would be to keep default as-is to avoid introducing breaking changes, but make it configureable for user who need it.

kibertoad:
If you want to be consistent with precedents, take a look at following files:
lib/dialects/postgres/index.js:
```
        if (!client.version) {
          return client.checkVersion(connection).then(function (version) {
            client.version = version;
            resolver(connection);
          });
        }
```

and then inside the lib/dialects/postgres/schema/columncompiler.js:
```
  if (!client.version || parseFloat(client.version) >= 9.2)
    return jsonb ? 'jsonb' : 'json';
```

We could use the same for Oracle.

kibertoad:
@drew-royster Do you need any help with this?

PR #3931, opened 2020-07-17
support postProcessResponse for streams (#3870)

cubikrichard:
#3870 

When using `query.stream()` the postProcessResponse method is not called.

Example:

```js
const knex = require('./knex');
const { PassThrough } = require('stream');

async function main() {
  const db = knex({
    client: 'pg',
    connection: 'postgres://localhost:5432/db',
    postProcessResponse: (result, queryContext) => {
      console.log('postProcessResponse', result, queryContext); // only called once, for the "normal" query
      return result;
    },
  });

  // normal promise - works
  await db('a_table').select('*').limit(3);

  // stream - doesn't work
  const stream = db('a_table').select('*').limit(3).stream();
  stream.pipe(new PassThrough({ objectMode: true }));
  await new Promise(resolve => stream.on("finish", () => resolve()));
}

main()
  .then(() => {
    console.log('✅');
    process.exit(0);
  })
  .catch((err) => {
    console.error(err);
    console.log('❌');
    process.exit(1);
  });

```

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -1,7 +1,7 @@
 const { KnexTimeoutError } = require('./util/timeout');
 const { timeout } = require('./util/timeout');
 
-let PassThrough;
+let Transform;
 
 // The "Runner" constructor takes a "builder" (query, schema, or raw)
 // and runs through each of the query statements, calling any additional
@@ -73,12 +73,19 @@ Object.assign(Runner.prototype, {
     // Determines whether we emit an error or throw here.
     const hasHandler = typeof handler === 'function';
 
-    // Lazy-load the "PassThrough" dependency.
-    PassThrough = PassThrough || require('stream').PassThrough;
+    // Lazy-load the "Transform" dependency.
+    Transform = Transform || require('stream').Transform;
 
     const runner = this;
-    const stream = new PassThrough({ objectMode: true });
+    const queryContext = this.builder.queryContext();
 
+    const stream = new Transform({
+      objectMode: true,
+      transform: (chunk, _, callback) => {
+        callback(null, this.client.postProcessResponse(chunk, queryContext));
+      },
+    });
+    
     let hasConnection = false;
     const promise = this.ensureConnection(function (connection) {
       hasConnection = true;
--- a/test/integration/builder/additional.js
+++ b/test/integration/builder/additional.js
@@ -84,6 +84,35 @@ module.exports = function (knex) {
             expect(res.queryContext).to.equal('the context');
           });
       });
+
+      it('should process response done through a stream', (done) => {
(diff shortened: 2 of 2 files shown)


elhigu:
Looks good, but I'm not enough used to stream APIs to be able to if this is fine solution or if it is missing anything.

KristjanTammekivi:
I don't think the tests would work properly on failure, the expects are executed in a different context

cubikrichard:
Good spot.

I'm not sure if Chai has something similar to jest `expect.assertions` (https://jestjs.io/docs/en/expect#expectassertionsnumber) but I can work around that with a count. 

Will update accordingly.

KristjanTammekivi:
Don't know an elegant way, you can assign the queryContext, callCount, count to a variable and then in the done callback do a try { expect().... done() } catch (err) { done(err); }, not very pretty but should do the trick

cubikrichard:
```js
it('should process response done through a stream', (done) => {
  let response;
  const stream = knex('accounts').limit(1).stream();
  
  stream.on('data', (res) => {
    response = res;
  });
  stream.on('finish', () => {
    expect(response.callCount).to.equal(1);
    done();
  });
});

it('should pass query context for responses through a stream', (done) => {
  let response;
  const stream = knex('accounts')
    .queryContext('the context')
    .limit(1)
    .stream();

  stream.on('data', (res) => {
    response = res;
  });
  stream.on('finish', () => {
    expect(res.queryContext).to.equal('the context');
    done();
  });
});

it('should process response for each row done through a stream', (done) => {
  const stream = knex('accounts').limit(5).stream();
  let count = 0;
  stream.on('data', () => count++);
  stream.on('finish', () => {
    expect(count).to.equal(5);
    done();
  });
});
```

Maybe a bit easier to read and fixes the problem - just rerunning the tests for fail conditions now to check they fail.

If the stream never finishes `done()` will never be called, and if the `data` callback isn't called _as expected_ the `finish` callback will error.

KristjanTammekivi:
The problem with this is that in case of an error it's hidden from the user, you'll just get a "Done callback wasn't executed within 2000ms" or whatever. But it is a bit more readable, I agree. ¯\_(ツ)_/¯

cubikrichard:
It is a fair point; although in the context for this test I feel we can assume that the implementation of `query.stream()` is _correct_ and therefore can have confidence that `stream.on('finish', () => {})` will be called. 

In that case, the error will actually read "response.queryContext did not equal 'the context'" for example

cubikrichard:
Is there anything I can do to help get this merged?

kibertoad:
@cubikrichard Let me look at it a bit more, hopefully, I'll merge it today and release a new version over weekend.

Issue #3921, opened 2020-07-11
Overly verbose logging when migrations directory specified


anandnimkar:
# Environment

Knex version: 0.21.2
Database + version: Irrelevant
OS: Irrelevant

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do

Due to changes introduced in https://github.com/knex/knex/pull/3839, on every call to `migrate.latest()` I now get two messages in my logs stating `"FS-related option specified for migration configuration. This resets migrationSource to default FsMigrations"`

I think this is overly verbose and possibly unnecessary. I run migrations programmatically at the start of every invocation of a Lambda function and now this is showing up in my logs.

At the very least, can we just have this message once per migrate.latest() call? Because `getMergedConfig` (in the file `lib/migrate/configuration-merger.js`) is called multiple times, it is showing up twice.

I also introduced a separate issue to ensure that `getMergedConfig` is called with the appropriate client logger when supplied. See https://github.com/knex/knex/issues/3919 and https://github.com/knex/knex/pull/3920.

2. Error message

See 1 above.

3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real
   database connection to MySQL or PostgreSQL, then single file example which initializes
   needed data and demonstrates the problem.

Set migrations directory in knexfile, then run `migrate.latest()` or `migrate.rollback()`

ksmithut:
I might also add that I'm fine if there's a more preferred method of setting custom migration directories (perhaps setting migrationSource to an instance of `FsMigrations`), but it would be nice if it were documented (and supported in the types).

elhigu:
You can set migration directory in configuration (https://knexjs.org/#Migrations-API)

```
{
  migrations: {
    directory: ['myCustomMigrationDir1','myCustomMigrationDir2']
  }
}
```

ksmithut:
@elhigu Right, the problem is that when using the `knex` js api for running migrations, you get verbose logging `"FS-related option specified for migration configuration. This resets migrationSource to default FsMigrations"` when using that option using the latest version

zjr:
Same if running `knex migrate` from the command line. I get the message twice.

jonahmolinski:
@jalfd exactly!

There should be a way to suppress this message or clear documentation on how to avoid it.

elhigu:
> @jalfd exactly!
> 
> There should be a way to suppress this message or clear documentation on how to avoid it.

[...]

PR #3936, opened 2020-07-19
Make the migrationSource / FsMigrations warning quieter

joshkel:
The warning that was added in #3839 is noisier than necessary; see #3921. It should suffice to warn only if the migrationSource is actually modified.


kibertoad:
@joshkel Please let me know if you need help with writing a test!

kibertoad:
Released in 0.21.6

Issue #3937, opened 2020-07-20
 ESM module interop fails if  calling module/package is of type 'module'   


D10221:
# Environment

Knex version: 0.21.2

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do

    ESM module interop fails if  calling module/package is of type 'module'   

2. Error message
    Error [ERR_REQUIRE_ESM]: Must use import to load ES Module
3. Reduced test code, 
```json
{
  "name": "knex14",
  "version": "1.0.0",
  "main": "index.js", 
  "type": "module",
  "dependencies": {
    "knex": "^0.21.2",
    "sqlite3": "^5.0.0"
  }
}
```sh
./node_modules/.bin/knex --esm migrate:up
```

This is more of an [esm](https://github.com/standard-things/esm) bug, see [#868](https://github.com/standard-things/esm/issues/868).  
That in turns is because a change in Node v13.8+ default behavior.

Solution.
As explained on [#868](https://github.com/standard-things/esm/issues/868). 
Provide a require extension to avoid Node's resolution 'error' .
```
//  knex/bin/cli.js line:46
require.extensions['.js'] = (m, fileName)=> 
      m._compile(require("fs").readFileSync(fileName, "utf8"), fileName);
```

D10221:
Here is a pr [#3938](https://github.com/knex/knex/pull/3938) with the proposed solution

kibertoad:
Released in 0.21.3

rajsinghUSA:
Thank you greatly for your work on this! I have been waiting for this feature for a long time, and I really appreciate your PR.

Unfortunately, I am getting this error:
```js
raj@r-sb2:~/repos/testserver$ npx knex@latest migrate:latest
npx: installed 159 in 7.501s
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /home/raj/repos/testserver/knexfile.js
require() of ES modules is not supported.
require() of /home/raj/repos/testserver/knexfile.js from /home/raj/.npm/_npx/9669/lib/node_modules/knex/lib/util/import-file.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename knexfile.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/raj/repos/testserver/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1273:13)
[...]

PR #3938, opened 2020-07-20
 ESM module interop fails if calling module/package is of type 'module' #3937 

D10221:
Provide a require extension to avoid Node's resolution 'error' .

### DIFF
--- a/test/jake-util/knexfile-esm-module/seeds/one.js
+++ b/test/jake-util/knexfile-esm-module/seeds/one.js
@@ -1,0 +1,1 @@
+export function seed(...args) {}
--- a/test/jake-util/knexfile-esm-module/seeds/one.js
+++ b/test/jake-util/knexfile-esm-module/seeds/one.js
@@ -1,1 +1,6 @@
-export function seed(...args) {}
+/**
+ * @param {import("../../../../")} knex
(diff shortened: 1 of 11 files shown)


kibertoad:
Thanks! Can you add CLI test for this?

D10221:
I thought it was covered b the existing one.
I can make an specific one :)

test/jake-util/knexfile-esm-module/seeds/one.js
```diff
@@ -0,0 +1 @@
+export function seed(...args) {}
```
kibertoad, line 1: Is this intended syntax? Why use `...args`?

D10221: You are correct, there is no need for the `...args`  here.
While the syntax is valid, there is no need to allocate unused variables.
I think I put them there for self documenting  reasons,.. and I just forgot to remove them when I realized there it was no need to assert it received the correct parameters, as this only needs to be successfully imported. and it shouldn't test the seeding functionality it self.

I'll remove the dead code, it might take bit, I pruned mi docker images, still downloading :)

kibertoad:
Thanks!

PR #3946, opened 2020-07-24
Add `null` as valid binding type

Naddiseo:
This change allows `null` into the bindings parameter of `.raw()` to typecheck properly

@lorefnon could you verify if this is the correct way to fix this


lorefnon:
@Naddiseo Thanks for your PR. Yes, not having null in binding types was an oversight.

kibertoad:
Released in 0.21.3

PR #3956, opened 2020-07-29
Fix migration source name in rollback all

caseywebdev:
This looks like it's an oversight, probably not caught because of low custom MigrationSource usage. Until this is merged, the temporary workaround is to make sure your custom `migration` objects have a `file` property that matches their `name`.

### DIFF
--- a/lib/migrate/Migrator.js
+++ b/lib/migrate/Migrator.js
@@ -183,7 +183,9 @@ class Migrator {
           return all
             ? allMigrations
                 .filter((migration) => {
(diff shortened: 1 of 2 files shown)


kibertoad:
@caseywebdev Could you add a CLI test for this to make sure we don't break it again?

caseywebdev:
Added

kibertoad:
Thank you!

kibertoad:
Released in 0.21.3

Issue #3786, opened 2020-04-09
Seed data in subfolders of config folder


lioaslan:
Hi,
I wonder that I can seed data from many sub-folders of the folder I config to seed?
```
seed/
     user/
          userAccount.js
          userProfile.js
     pagekage/
          pagekage1.js
          pagekage2.js
....
```
I have to put file in sub-folder because my project I'm working is so complicate and sub-folder is required for clarity.

Thanks for any helps.
________
So sorry for my poor English

kibertoad:
Assuming you have configuration like this:
```
development: {
  client: ...,
  connection: { ... },
  seeds: {
      directory: './seed'
  }
}
```

You can use multiple directories:

development: {
  client: ...,
  connection: { ... },
  seeds: {
      directory: ['./seed/user', 'seed/pagekage']
  }
}
```

kibertoad:
@syphuongcuong3 Note that very old knex versions don't support that, so if that fails to work, consider updating knex.

kibertoad:
Actually, I double-checked and it seems that only migrations support multiple folders for now. I'll give it a shot today or tomorrow evening, there's an incomplete PR for this: #3357

syphuongcuong3:
Assume that, I have 10 or 20 sub-folders of user/ and 20 sub-folder of package/
So I have to update 30-40 strings in directory
I hope that seed can run in file inside sub-folder of the directory.
Thanks

lioaslan:
Hope you fix it soon.
Thanks so much

mdartic:
Any news on this @kibertoad ?
I need this feature for a project I currently have, I'm dividing seeds in several folders.
Could I help on the PR #3357 ?

kibertoad:
thank you for the ping, drowning in multiple workstreams. author of that PR claimed that approach ended up not working. if you could investigate, and finish his work with CLI test, I'll make sure to review urgently.

mdartic:
Ok, I'll give it a try.
I don't know the knex codebase, so I'm gonna look on the "migration" side (as you said it was already working).
Any info about why the PR approach wasn't the good one ?
Do you think we have to implement the subfolder drilling too ?
Is this feature already implemented in the migration one ?

kibertoad:
@mdartic I would imagine that quite a bit of code that is currently duplicated between migrations and seeds will have to be synced up. Don't see a reason why it wouldn't work after that.

kibertoad:
[...]

PR #3967, opened 2020-08-06
3786 seed data subfolders

mdartic:
Work in progress for the #3786 issue.
@kibertoad could you tell me if this WIP look good for you ?
I need some advice to add tests on the codebase. 
Could you give me some directions ?

### DIFF
--- a/lib/seed/Seeder.js
+++ b/lib/seed/Seeder.js
@@ -11,6 +11,11 @@ const {
 } = require('../util/fs');
 const { writeJsFileUsingTemplate } = require('../util/template');
 
+const filterByLoadExtensions = (extensions) => (value) => {
+  const extension = path.extname(value);
+  return includes(extensions, extension);
+};
+
 // The new seeds we're performing, typically called from the `knex.seed`
 // interface on the main `knex` object. Passes the `knex` instance performing
 // the seeds.
@@ -54,17 +59,10 @@ class Seeder {
     // (getFilepathsInFolderRecursively does this)
     // if false, we need to sort all the seeds
     if (this.config.sortDirsSeparately) {
-      return seeds.filter(this._filterByLoadExtensions(loadExtensions));
+      return seeds.filter(filterByLoadExtensions(loadExtensions));
     } else {
-      return seeds.filter(this._filterByLoadExtensions(loadExtensions)).sort();
+      return seeds.filter(filterByLoadExtensions(loadExtensions)).sort();
     }
-  }
-
-  _filterByLoadExtensions(extensions) {
-    return function (value) {
-      const extension = path.extname(value);
-      return includes(extensions, extension);
-    };
   }
 
   // Ensures a folder for the seeds exist, dependent on the
--- a/lib/seed/Seeder.js
+++ b/lib/seed/Seeder.js
@@ -150,7 +150,7 @@ class Seeder {
       if (!directory) {
         // eslint-disable-next-line no-console
         console.warn(
-          'Failed to resolve config file, knex cannot determine where to generate migrations'
+          'Failed to resolve config file, knex cannot determine where to run or make seeds'
         );
       }
       return path.resolve(process.cwd(), directory);
--- a/lib/seed/Seeder.js
+++ b/lib/seed/Seeder.js
@@ -73,7 +73,6 @@ class Seeder {
 
   // Validates seed files by requiring and checking for a `seed` function.
   _validateSeedStructure(filepath) {
-    // const seed = require(path.join(this._absoluteConfigDir(), name));
     const seed = require(filepath);
     if (typeof seed.seed !== 'function') {
       throw new Error(
@@ -100,7 +99,6 @@ class Seeder {
     const dirs = this._absoluteConfigDirs();
     const dir = dirs.slice(-1)[0]; // Get last specified directory
     return path.join(dir, fileName);
-    // return path.join(this._absoluteConfigDir(), this._getNewStubFileName(name));
   }
 
   // Write a new seed to disk, using the config and generated filename,
@@ -119,10 +117,8 @@ class Seeder {
   // Runs a batch of seed files.
   async _waterfallBatch(seeds) {
     const { knex } = this;
-    // const seedDirectory = this._absoluteConfigDir();
     const log = [];
     for (const seedPath of seeds) {
-      // const seedPath = path.join(seedDirectory, seedName);
       const seed = require(seedPath);
       try {
         await seed.seed(knex);
--- a/lib/seed/Seeder.js
+++ b/lib/seed/Seeder.js
@@ -43,18 +43,18 @@ class Seeder {
   // Lists all available seed files as a sorted array.
   async _listAll(config) {
     this.config = this.setConfig(config);
-    const loadExtensions = this.config.loadExtensions;
     const seeds = (
       await Promise.all(
         this._absoluteConfigDirs().map(getFilepathsInFolderRecursively)
       )
     ).flat();
-    return seeds
-      .filter((value) => {
-        const extension = path.extname(value);
-        return includes(loadExtensions, extension);
-      })
-      .sort();
+    return seeds.filter(this._filterByLoadExtensions).sort();
+  }
+
+  _filterByLoadExtensions(value) {
+    const extension = path.extname(value);
+    const loadExtensions = this.config.loadExtensions;
+    return includes(loadExtensions, extension);
   }
 
   // Ensures a folder for the seeds exist, dependent on the
--- a/lib/seed/Seeder.js
+++ b/lib/seed/Seeder.js
@@ -59,7 +59,7 @@ class Seeder {
 
   // Ensures a folder for the seeds exist, dependent on the
   // seed config settings.
-  async _ensureFolder() {
+  _ensureFolder() {
     const dirs = this._absoluteConfigDirs();
     const promises = dirs.map(ensureDirectoryExists);
     return Promise.all(promises);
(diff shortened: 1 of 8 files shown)


lib/seed/Seeder.js
```diff
@@ -41,22 +44,25 @@ class Seeder {
...
     const loadExtensions = this.config.loadExtensions;
-    return readdir(this._absoluteConfigDir()).then((seeds) =>
-      seeds
-        .filter((value) => {
-          const extension = path.extname(value);
-          return includes(loadExtensions, extension);
-        })
-        .sort()
-    );
+    const seeds = (
```
kibertoad, line 47: Could you please replicate sortDirsSeparately parameter from migrations for consistency? (https://github.com/knex/knex/blob/fa906278b823728f02349be7315d536de6feb7b1/lib/migrate/sources/fs-migrations.js#L47)

To provide context: this would allow users to either run seeds directory by directory (in the order specified in configuration), or have global execution order (all filenames sorted irrespective of the directory they are located in)

lib/seed/Seeder.js
```diff
@@ -41,22 +44,25 @@ class Seeder {
...
+      await Promise.all(
+        this._absoluteConfigDirs().map(getFilepathsInFolderRecursively)
+      )
+    ).flat();
+    return seeds
+      .filter((value) => {
```
kibertoad, line 53: I would assume this filter function is used in other places as well, probably we can extract it.

lib/seed/Seeder.js
```diff
@@ -41,22 +44,25 @@ class Seeder {
...
 
   // Ensures a folder for the seeds exist, dependent on the
   // seed config settings.
   async _ensureFolder() {
-    const dir = this._absoluteConfigDir();
-
-    await ensureDirectoryExists(dir);
+    const dirs = this._absoluteConfigDirs();
+    const promises = dirs.map(ensureDirectoryExists);
+    return Promise.all(promises);
```
kibertoad, line 65: If you want to be consistent with original implementation, should be `await Promise.all(promises);`.
Alternatively function doesn't need to be async anymore.

lib/seed/Seeder.js
```diff
@@ -87,7 +96,11 @@ class Seeder {
   }
 
   _getNewStubFilePath(name) {
-    return path.join(this._absoluteConfigDir(), this._getNewStubFileName(name));
+    const fileName = this._getNewStubFileName(name);
+    const dirs = this._absoluteConfigDirs();
+    const dir = dirs.slice(-1)[0]; // Get last specified directory
+    return path.join(dir, fileName);
+    // return path.join(this._absoluteConfigDir(), this._getNewStubFileName(name));
```
kibertoad, line 103: should this be commented out?

lib/seed/Seeder.js
```diff
@@ -106,10 +119,10 @@ class Seeder {
   // Runs a batch of seed files.
   async _waterfallBatch(seeds) {
     const { knex } = this;
-    const seedDirectory = this._absoluteConfigDir();
+    // const seedDirectory = this._absoluteConfigDir();
```
kibertoad, line 122: should this be commented out?

lib/seed/Seeder.js
```diff
@@ -129,8 +142,23 @@ class Seeder {
...
+   */
+  _absoluteConfigDirs() {
+    const directories = Array.isArray(this.config.directory)
+      ? this.config.directory
+      : [this.config.directory];
+    return directories.map((directory) => {
+      if (!directory) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          'Failed to resolve config file, knex cannot determine where to generate migrations'
```
kibertoad, line 157: is the warning message correct here?

lib/util/fs.js
```diff
@@ -31,11 +31,42 @@ function ensureDirectoryExists(dir) {
...
 
+/**
+ * Read recursively a directory,
+ * sorting folders and files by alphabetically order
+ *
+ * @param {string} dir
+ * The directory to recursively analyse
+ */
+async function getFilepathsInFolderRecursively(dir) {
+  let results = [];
```
kibertoad, line 42: looks like it can be `const`

lib/util/fs.js
```diff
@@ -31,11 +31,42 @@ function ensureDirectoryExists(dir) {
...
+          // Add directory to array [comment if you need to remove the directories from the array]
+          const subFiles = await getFilepathsInFolderRecursively(currentFile);
+          results = results.concat(subFiles);
+        } else {
+          results.push(currentFile);
+        }
+      })
+    );
+  } catch (error) {
+    throw new Error(error);
```
kibertoad, line 59: why catch it in the first place?

kibertoad:
@mdartic Check this out for how to write CLI tests for seeding: https://github.com/knex/knex/blob/master/test/cli/seed.spec.js

lib/util/fs.js
mdartic: we reaffect results with the `concat` later 

https://github.com/knex/knex/pull/3967/files/f5c3c9a0f79271b265ec3f4519b3acda3c68f01f#diff-38843cf9c53f32dffeb2e781e43b4d2fR52

in this case I need a `let` variable ?

lib/util/fs.js
mdartic: the use of `readdir` and `stat` make me think it could break at the beginning,
do you have an other idea ?

lib/seed/Seeder.js
mdartic: I choose the second option, to match the `MigrationGenerator` implementation.
I don't know where to put this "common" method.
Do we have to make a `class` and make `MigrationGenerator` and `Seeder` ones extends them ?

lib/seed/Seeder.js
mdartic: This is fixed in the latest commit. Do I let you resolve the conversation ?

mdartic:
I'm going to write some tests.
I've pushed some commits fixing some of your comments.
I have to add the `sortDirsSeparately` option too.

lib/seed/Seeder.js
mdartic: I've implemented the `sortDirsSeparately`. Tell me if it sounds ok for you.

mdartic:
@kibertoad I've pushed some tests, it "works" on sqlite on my PC.
I hope CI will pass.
Tell me if it's ok for you.
I've seen we have to update the documentation too, and seems the PR https://github.com/knex/documentation/pull/220 looks fine for what I've added ? Maybe if it's ok for this PR, we could reopen the documentation PR ?

lib/seed/Seeder.js
```diff
@@ -41,22 +45,34 @@ class Seeder {
...
+    // if false, we need to sort all the seeds
+    if (this.config.sortDirsSeparately) {
+      return seeds.filter(this._filterByLoadExtensions(loadExtensions));
+    } else {
+      return seeds.filter(this._filterByLoadExtensions(loadExtensions)).sort();
+    }
+  }
+
+  _filterByLoadExtensions(extensions) {
+    return function (value) {
```
kibertoad, line 64: can be an arrow function for brevity, `(value) => {<...>}`

should this be a class method, though, considering that it is completely static?

lib/seed/Seeder.js
kibertoad: Either way works, you can have a superclass, or just import from some kind of commons util file with static functions.

lib/util/fs.js
kibertoad: but even if it breaks in the beginning, what do we gain from immediately rethrowing an error wrapped into generic Error?

lib/seed/Seeder.js
mdartic: Where do you put this method in this case, in which file ?

lib/util/fs.js
kibertoad: Oh, I see. It can be quite confusing for other developers when you reassign the results list, though.

Do we need to do it in a mutable way? Wouldn't destructuring work here?
`arr1.push(...arr2) // Adds arr2 items to end of array`

test/cli/seed.spec.js
```diff
@@ -20,6 +20,16 @@ describe('seed:run', () => {
     );
   });
 
+  it('find recursively files', () => {
+    return execCommand(
+      `node ${KNEX} seed:run --knexfile=test/jake-util/seeds-knexfile-directories.js`,
+      {
+        expectedOutput: 'Ran 3 seed files',
```
kibertoad, line 27: would be helpful to also assert expected seed names

test/jake-util/seeds-recursive/first/third/first.js
```diff
@@ -0,0 +1 @@
+exports.seed = (knex, Promise) => {};
```
kibertoad, line 1: I think `Promise` is not longer used nor passed. Am I wrong?

lib/seed/Seeder.js
kibertoad: can be in this same file for now

test/cli/seed.spec.js
mdartic: Added

lib/seed/Seeder.js
mdartic: Done

test/jake-util/seeds-recursive/first/third/first.js
mdartic: You're right ;-)

mdartic:
I've added some changes to the PR.
Tell me what do you think.
Concerning all your conversation, do you want me I click "Resolve" when it's ok for me ? 
I would prefer you do this, and I know you agree with the code.

kibertoad:
@mdartic I think it looks great now, thanks a lot for your contribution! Will release new version soon.

Issue #3978, opened 2020-08-13
Problem with ESM modules


umaar:
### Environment

Knex version: 0.21.4
OS: macOS
Node: v14.4.0

```sh
 ➜  ./node_modules/.bin/knex --esm migrate:latest --knexfile knexfile.mjs
Requiring external module /Users/umarhansa/development/project/node_modules/interpret/mjs-stub
TypeError: Cannot read property 'get' of undefined
```

That error is referring to the start of my knex.mjs file which looks like this:

```js
import config from 'config';
const environment = config.get('environment');
```

When I log the value of `config`, it's `undefined` meaning something is broken with ESM via knex. This was working in `0.21.1`

Also please see: https://github.com/knex/knex/issues/3882 - can you include more ESM related tests please? It will help catch these issues before users do, thanks. So far I've experienced quite a few issues already with ESM + knex. Totally appreciate it's still a relatively 'new' thing.

kibertoad:
@D10221 Could you please take a look at this?

D10221:
@umaar 
Hi,  
It seems related to [esm #851](https://github.com/standard-things/esm/issues/851)
'.mjs' limitations are not well documented.
I repro your case , 'mjs' extension might not be well supported by 'esm' , or not work at all 
We might have to fallback to native `import` for that case (`mjs`) 


Would it be possible for you to try your setup with `js' extension instead ?

umaar:
Hi, I will be able to try later and let you know - but just a reminder, this was working in knex version `0.21.1` <- that version also used the `esm` module I think. So possibly something has changed?

Also I haven't tried recently, but a few months ago when I was trying to make ESM work with Knex, I had to call my knexfile `.mjs`. I would've preferred to use `.js` but it didn't work at the time.

hydroper:
I'm having the same issue with latest version. I tried a foo.js at the same directory as knexfile.js.

![image](https://user-images.githubusercontent.com/52384977/90156921-75700d80-dd63-11ea-9de4-b34e512b32c6.png)

![image](https://user-images.githubusercontent.com/52384977/90156934-79039480-dd63-11ea-9ebf-112cbbcfa5d0.png)

D10221:
@umaar   
Yes it did changed.
'esm' wasn't working when the host package is `type:"module"`, latest change is trying to solve that.
I did try latest 'knex' and locally hack the changes back (pseudo revert to previous behavior)  
But is the same result you have reported here when using 'mjs' extension

[...]

PR #3980, opened 2020-08-13
#3978 mjs modules

D10221:
Fallback to native 'import' on mjs files
Tested on Node 10, 12, 14 ... Ubuntu, Windows and WSL

NOTE Node v10.x is supported behind experimental modules flag 

Let see what the CI has to say :)


bin/cli.js
```diff
@@ -23,7 +23,10 @@ const { readFile, writeFile } = require('./../lib/util/fs');
 const { listMigrations } = require('./utils/migrationsLister');
 
 async function openKnexfile(configPath) {
-  let config = require(configPath);
+  const importFile = require('../lib/util/import-file');
+  let config = await importFile(configPath);
+  // needs more testing
+  // config = config.default || config; //Allow default export || named exports
```
kibertoad, line 29: should the commented out code be here?

lib/util/import-file.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * imports 'mjs', else requires.
+ * NOTE: require me late!
+ * @param {string} filepath
+ * @todo WARN on version 10 and '--experimental-modules' and '--esm'
+ */
+module.exports = function importFile(filepath) {
+  return filepath.endsWith('.mjs')
+    ? import(require('url').pathToFileURL(filepath))
```
kibertoad, line 9: do we need to lazy-load `url` as well here?

test/cli/seed.spec.js
```diff
@@ -182,3 +182,36 @@ describe('seed:run', () => {
...
+      else resolve();
+    })
+  );
+  await new Promise((resolve) => db.close(() => resolve()));
+  const version = Number((/v(\d+)/i.exec(process.version) || [])[1]);
+  return execCommand(
+    [
+      `node`,
+      // TODO: document this !
+      version === 10 && '--experimental-modules',
```
kibertoad, line 202: what does providing version give us here?

kibertoad: nvm, read the code wrong

lib/util/import-file.js
D10221: Not really , it can be a top level import. I'm maybe a bit paranoid there
I guess I test run can decide on it.

bin/cli.js
D10221: Well, I think is a missing feature.
Been thinking from the beginning on the 'default' export, 
but there is always more important TODO's, 
Didn't want to mix 'concern' and introduce unneeded changes
Can't make my mind up on that :)
SO I guess it should be removed

D10221:
BTW : Thank you very much for your input !

I feel is 100% finished,
I would like to think a bit more on edge cases for the mjs/esm combination.
Mostly mjs knexfile and js migrations/seeds.
And final clean up. 

I sent the PR last night 3:00 AM (my time), and I was a bit  burnout :)

But I wanted to see what the CI has to say 

Because  I had 1 test failure on my Docker setup , due to timeouts 
And I thought I just probably was getting some ENV wrong

Thanks again

lib/util/import-file.js
D10221: Run a few tests with the top level `require` 
And i get some test failures, ... timeout related 
It could be just my environment 
But, I think it could  be making the initialization a bit more expensive. 
Perhaps its better to leave it 'lazy' , and let only the `--esm` users pay the minimal performance penalty 
As Its cached by require anyway

These are the failures
```

  1) Integration Tests
       sqlite3 | sqlite3
         Updates
           should immediately return updated value for other connections when updating row to DB returns:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (.../knex/test/all-tests-suite.js)
      at listOnTimeout (internal/timers.js:551:17)
      at processTimers (internal/timers.js:494:7)

  2) Integration Tests
       sqlite3 | sqlite3
         Updates
           should increment a float value:
     AssertionError: expected 676.53 to be close to 0 +/- 0.001
      at .../knex/test/integration/builder/updates.js:185:57

  3) Integration Tests
       sqlite3 | sqlite3
         Updates
           should allow returning for updates in postgresql:
     KnexTimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?
      at Client_SQLite3.acquireConnection (lib/client.js:349:26)
```

PR #3984, opened 2020-08-15
Update dtslint

lorefnon:


kibertoad:
Thanks! Wonder if it was a bug in older dtslint or some local issue.

PR #4000, opened 2020-08-24
Added schema creation (postgres) if needed on first migration

spanyukov:
fix a problem with first migration when you don't have schema yet.
Before you had a bug like this:
`error: create table "config"."knex_migrations" ("id" serial primary key, "name" varchar(255), "batch" integer, "migration_time" timestamptz) - schema "config" does not exist`

### DIFF
--- a/lib/migrate/table-creator.js
+++ b/lib/migrate/table-creator.js
@@ -33,7 +33,7 @@ function ensureTable(tableName, schemaName, trxOrKnex) {
 }
 
 function _createMigrationTable(tableName, schemaName, trxOrKnex) {
-  return getSchemaBuilder(trxOrKnex, schemaName).createTable(
+  return getSchemaBuilder(trxOrKnex, schemaName).createSchemaIfNotExists(schemaName).createTable(
     getTableName(tableName),
     function (t) {
       t.increments();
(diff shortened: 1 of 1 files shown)


ike18t:
I'd love to have this functionality as well. Unfortunately the commit broke the build but a conditional call of createSchemaIfNotExists based on if a schemaName exists should resolve the issue.

spanyukov:
> I'd love to have this functionality as well. Unfortunately the commit broke the build but a conditional call of createSchemaIfNotExists based on if a schemaName exists should resolve the issue.

Its working only with Postgres yet.. I will fix it and make PR again.

ike18t:
@spanyukov Do you think you'll have time to update this PR? I don't mind opening a new PR if not.

spanyukov:
> @spanyukov Do you think you'll have time to update this PR? I don't mind opening a new PR if not.

I added condition.

ike18t:
Your changes unfortunately don't pass CI. You are probably better off appending the createSchemaIfNotExists to https://github.com/knex/knex/blob/9b82d0653a64bb56397275d524d2a51f06e1a41c/lib/migrate/table-creator.js#L70
Then the change only needs to happen in one place and that place is already "schema aware".

Also there is docker-compose file if you want to run tests locally to verify your changes.

spanyukov:
> Your changes unfortunately don't pass CI. You are probably better off appending the createSchemaIfNotExists to
> 
> https://github.com/knex/knex/blob/9b82d0653a64bb56397275d524d2a51f06e1a41c/lib/migrate/table-creator.js#L70
> 
> 
> Then the change only needs to happen in one place and that place is already "schema aware".
> Also there is docker-compose file if you want to run tests locally to verify your changes.

thanks! I put this code to schema builder getter as you described.

PR #4003, opened 2020-08-27
Unify SeedsConfig and SeederConfig

xuanduc987:
These two types seems to describe the same thing while being out of sync.


lorefnon:
@xuanduc987 Yeah you are right. Thanks for updating this.

lorefnon (approved):

xuanduc987:
Rebased onto master

kibertoad (approved):

kibertoad:
Thanks!

PR #4011, opened 2020-08-31
Throw error when specific seed file cannot be found

### LINKED ISSUE #4005: Knex seed should notify when a specific seed is not found
# Environment

Knex version: 0.21.5
Database + version: PostgreSQL: 12.4
OS: macOS 10.15.6

# Bug

I am calling `knex.seed.run({ specific: 'some-seeder.ts' });` to invoke a **specific** seed file that my integration tests **rely** on.
I had renamed that seeder, and the command could not find it, but:

- did not throw an error
- did not show a warning or notice to let me know it was ignored (failed silently)

I figured out what was the problem by debugging and realizing that the seeder was not invoked at all.

2. Error message

There is none, it fails silently, this is my issue.

```
// a test in a jest environment
beforeAll(async () => {
  // not finding a **specific** seeder should throw, or at least notice that the seeder cannot be executed
  await knex.seed.run({ specific: 'some-file-that-does-not-exists' });
});
```

I have integration tests relying on a test database which I need to seed with specific data for my tests to work (reflecting the business workflow of the client I work for)

An update to the `seed.run` source code to handle the case when the `specific` option is provided but the seeder cannot be found.

I can think of a few options:
- throw an error in such a case (could also be an opt-in flag/option for `seed.run()` to throw an error in such a case)
- log a message stating that one specific seeder cannot be found

I am happy to contribute with a PR after I get some approval on the way forward :)

flo-sch:
fix(seed): throw error when specific seed file cannot be found

fixes #4005

### DIFF
--- a/lib/seed/Seeder.js
+++ b/lib/seed/Seeder.js
@@ -25,11 +25,15 @@ class Seeder {
   // Runs seed files for the given environment.
   async run(config) {
     this.config = this.setConfig(config);
-    const all = await this._listAll();
-    const files =
-      config && config.specific
-        ? all.filter((file) => path.basename(file) === config.specific)
-        : all;
+    let files = await this._listAll();
+    if (config && config.specific) {
+      files = files.filter((file) => path.basename(file) === config.specific);
+      if (files.length === 0) {
+        throw new Error(
+          `Invalid argument provided: the specific seed "${config.specific}" does not exist.`
+        );
+      }
+    }
     return this._runSeeds(files);
   }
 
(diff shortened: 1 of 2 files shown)


flo-sch:
It looks like some unrelated tests are failing on node v10:

> Should have created new connection and execute the query, got : Error: select 1 - This socket has been ended by the other party

Is that a known issue?

elhigu:
I haven't seen that before, sounds like CI was having some issues... for some reason I don't have controls in travis to re-run CI anymore so I had to close/open to do that.

flo-sch:
Looks like it worked this time... Not sure what happened in Travis then 🤷‍♂

elhigu:
Thanks!

PR #4022, opened 2020-09-07
Add two new where clause

basit1327:
1) **whereGreaterThan (column, value)**
_Usage_: knex('users').whereGreaterThan('votes', 100)
_Output_: select * from `users` where `votes` > 100

2) **whereLessThan (column, value)**
_Usage_: knex('users').whereLessThan('votes', 100)
_Output_: select * from `users` where `votes` < 100

### DIFF
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -546,6 +546,40 @@ assign(Builder.prototype, {
     return this._bool('or').whereNotBetween(column, values);
   },
 
+  // Adds a `where greater than` clause to the query.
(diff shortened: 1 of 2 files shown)


elhigu:
You can create this as your own knex extension. 

Also implementation is too opinionated. There are usecases where one doesn't like to have a number value on the right side. Also If we go to this path then next we would have `.whereGreaterOrEqualThan()`... etc. 

And all features must have good tests (also for error handling) and documentation included.

PR #4033, opened 2020-09-16
Allow string[] type for directory in SeedsConfig

aMoniker:
I'm pretty sure you can do this, since I had it working for multiple seeds without the typing.


kibertoad (approved):

kibertoad:
Thanks!

PR #4038, opened 2020-09-18
Warn if whereNot is used with 'in' or 'between'

simon-lorenz:
Today I was confused by the fact that `whereNot('col', 'in', [ values ])` is not valid and does fail silently. I struggled with this for quite some time before finding this part of the documentation:

> CAVEAT: WhereNot is not suitable for "in" and "between" type subqueries. You should use "not in" and "not between" instead.

If I understand correctly, it's never allowed to call `whereNot()` with "in" or "between". This PR logs the warning from the documentation, helping developers to avoid this error.

### DIFF
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -371,6 +371,13 @@ assign(Builder.prototype, {
 
   // Adds an `not where` clause to the query.
   whereNot() {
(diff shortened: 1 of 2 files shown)


kibertoad (approved):

kibertoad:
Thank you!

PR #4040, opened 2020-09-19
Fix double merging of config for migrator

### LINKED ISSUE #3834: Custom migration source is always overwritten with FsMigrations

kibertoad:
fixes #3834
fixes #3936

### DIFF
--- a/lib/migrate/Migrator.js
+++ b/lib/migrate/Migrator.js
@@ -307,8 +307,7 @@ class Migrator {
 
   // Creates a new migration, with a given name.
   make(name, config) {
-    this.config = getMergedConfig(config, this.config, this.knex.client.logger);
-    return this.generator.make(name, this.config, this.knex.client.logger);
+    return this.generator.make(name, config, this.knex.client.logger);
   }
 
   _disableProcessing() {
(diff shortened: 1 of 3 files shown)


kibertoad:
@hallais @KrishnaPG Please take a look!

PR #4047, opened 2020-09-24
Override knexfile options with CLI options

KoryNunn:
This PR allows overriding options in a passed knexfile with other options passed to the CLI.


kibertoad:
There's a test failing, could you please take a look?

kibertoad:
@KoryNunn Could you please also update documentation in https://github.com/knex/documentation?

bin/utils/cli-config-utils.js
```diff
@@ -6,6 +6,28 @@ const tildify = require('tildify');
...
+
+  if (opts.client) {
+    config.client = opts.client;
+  }
+
+  if (opts.connection) {
+    config.connection = opts.connection;
+  }
+
+  if (opts.migrationsDirectory) {
```
kibertoad, line 20: I wonder if there is a way to make all overrides case-insensitive. Probably we can do check on lowercased entry pairs.

kibertoad:
@KoryNunn Do you need any help with this?

bin/utils/cli-config-utils.js
OlivierCavadenti: As I can understand, the goal is to lower case the properties in options object before merge with config in knexfile ?

KoryNunn:
Hi just a quick update. I no longer use knex, and as such have no capability for valuable input. I'll leave this PR open in case it is of use to anyone else, but also not against it being closed.

bin/utils/cli-config-utils.js
kibertoad: But some options are camelCase, no?

OlivierCavadenti:
> Hi just a quick update. I no longer use knex, and as such have no capability for valuable input. I'll leave this PR open in case it is of use to anyone else, but also not against it being closed.

It's ok no problem at all, I will finish it !

bin/utils/cli-config-utils.js
OlivierCavadenti: Mmh, I will check that this week.

OlivierCavadenti:
@kibertoad 
Doc here https://github.com/knex/documentation/pull/395
By the way, I assume I can also close https://github.com/knex/knex/pull/5006 which I have opened but it's a duplicate of this one ?
For your comment with camelCase, I made some tests, but since cli options with commander have fixed case and transformed in camelCase properties, I don't see any problems, and unit test is ok. Maybe I miss something ?

kibertoad (approved):

kibertoad:
Yeah, looks good, and we can close 5006 too.

kibertoad:
Released in 1.0.5

PR #4051, opened 2020-09-27
feat(QueryBuilder): add method clear(statement)

intech:
Add method `clear(statement)` in `QueryBuilder` for https://github.com/knex/knex/issues/1476

#### TODO:
- [x] approve with the community
- [x] docs
- [x] tests

### DIFF
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -1040,7 +1040,7 @@ assign(Builder.prototype, {
   
   // Remove everything from statement clause
   clear(statement) {
-    if(!['select', 'columns', 'where', 'group', 'order', 'having', 'limit', 'offset'].includes(statement))
+    if(!['with', 'select', 'columns', 'where', 'union', 'join', 'group', 'order', 'having', 'limit', 'offset'].includes(statement))
       throw new Error(`Knex Error: unknown statement '${statement}'`);
     if(statement === 'select') statement = 'columns';
     this._clearGrouping(statement);
@@ -1250,7 +1250,11 @@ assign(Builder.prototype, {
 
   // Helper function for clearing or reseting a grouping type from the builder
   _clearGrouping(grouping) {
-    this._statements = reject(this._statements, { grouping });
+    if(grouping in this._single) {
+      delete this._single[grouping];
+    } else {
+      this._statements = reject(this._statements, { grouping });
+    }
   },
 
   // Helper function that checks if the builder will emit a select query
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -1037,12 +1037,12 @@ assign(Builder.prototype, {
     this._clearGrouping('having');
     return this;
   },
-  
+
   // Remove everything from statement clause
   clear(statement) {
     if(!['with', 'select', 'columns', 'where', 'union', 'join', 'group', 'order', 'having', 'limit', 'offset'].includes(statement))
       throw new Error(`Knex Error: unknown statement '${statement}'`);
-    if(statement === 'select') statement = 'columns';
+    if(statement === 'select') { statement = 'columns'; }
     this._clearGrouping(statement);
     return this;
   },
@@ -1251,7 +1251,7 @@ assign(Builder.prototype, {
   // Helper function for clearing or reseting a grouping type from the builder
   _clearGrouping(grouping) {
     if(grouping in this._single) {
-      delete this._single[grouping];
+      this._single[grouping] = undefined;
     } else {
       this._statements = reject(this._statements, { grouping });
     }
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -968,7 +968,6 @@ assign(Builder.prototype, {
   // Clears increments/decrements
   clearCounters() {
     this._single.counter = {};
-
     return this;
   },
 
@@ -1040,8 +1039,9 @@ assign(Builder.prototype, {
 
   // Remove everything from statement clause
   clear(statement) {
-    if(!['with', 'select', 'columns', 'where', 'union', 'join', 'group', 'order', 'having', 'limit', 'offset'].includes(statement))
+    if(!['with', 'select', 'columns', 'where', 'union', 'join', 'group', 'order', 'having', 'limit', 'offset', 'counter', 'counters'].includes(statement))
       throw new Error(`Knex Error: unknown statement '${statement}'`);
+    if(statement.startsWith('counter')) return this.clearCounters();
     if(statement === 'select') { statement = 'columns'; }
     this._clearGrouping(statement);
     return this;
--- a/lib/query/builder.js
+++ b/lib/query/builder.js
@@ -1008,33 +1008,42 @@ assign(Builder.prototype, {
     return this;
   },
 
-  // Remove everything from select clause
+  // Deprecated. Remove everything from select clause
   clearSelect() {
     this._clearGrouping('columns');
     return this;
   },
 
-  // Remove everything from where clause
+  // Deprecated. Remove everything from where clause
   clearWhere() {
     this._clearGrouping('where');
     return this;
   },
 
-  // Remove everything from group clause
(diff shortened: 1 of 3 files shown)


kibertoad:
Thank you for the contribution! Could you please add tests for this?

intech:
> Thank you for the contribution! Could you please add tests for this?

Oh sure!
I am worried that I just got acquainted with the project three days ago and really need your approval. Did I take everything into account in logic?

lib/query/builder.js
```diff
@@ -1008,35 +1008,44 @@ assign(Builder.prototype, {
...
 
-  // Remove everything from order clause
+  // Deprecated. Remove everything from order clause
   clearOrder() {
     this._clearGrouping('order');
     return this;
   },
 
-  // Remove everything from having clause
+  // Deprecated. Remove everything from having clause
```
kibertoad, line 1035: Why mark existing methods as deprecated?

lib/query/builder.js
```diff
@@ -1008,35 +1008,44 @@ assign(Builder.prototype, {
...
   clearHaving() {
     this._clearGrouping('having');
     return this;
   },
+  
+  // Remove everything from statement clause
+  clear(statement) {
+    if(!['with', 'select', 'columns', 'where', 'union', 'join', 'group', 'order', 'having', 'limit', 'offset'].includes(statement))
+      throw new Error(`Knex Error: unknown statement '${statement}'`);
+    if(statement === 'select') statement = 'columns';
```
kibertoad, line 1045: Could you please add `{}`?

lib/query/builder.js
```diff
@@ -1241,7 +1250,11 @@ assign(Builder.prototype, {
 
   // Helper function for clearing or reseting a grouping type from the builder
   _clearGrouping(grouping) {
-    this._statements = reject(this._statements, { grouping });
+    if(grouping in this._single) {
+      delete this._single[grouping];
```
kibertoad, line 1254: Is there a particular reason to use `delete` over "set to undefined" here? Delete is slow.

lib/query/builder.js
intech: For further refactoring, remove duplicate features.
Leave backward compatibility to major version.
What do you think?

lib/query/builder.js
intech: Specify what it is about?

kibertoad: `if (xyz) {  statement }`

intech:
@kibertoad Fixed the code. Added a test. Updated the documentation.
Please review. Thanks!

intech:
@kibertoad I propose to merge so that people can already use.
And this PR for docs https://github.com/knex/documentation/pull/289

intech:
@kibertoad we are still waiting for your approval :hourglass_flowing_sand:

lib/query/builder.js
kibertoad: Isn't there a difference between those methods, though? From my understanding, existing ones are more granular, while new ones clears everything. Do I read code wrong?

intech: `clear("order")` method completely replaces `clearOrder()`, the difference is that clear gives you more options for limit, offset and other

test/unit/query/builder.js
```diff
@@ -743,6 +743,45 @@ describe('QueryBuilder', () => {
     );
   });
 
+  it('clear by statements', () => {
```
elhigu, line 746: More tests needed to make sure that each clear statement works separately and doesn't clear anything unexpected + test for non-happy-path while using non existing statement.  

Also some test showing that it does not clear anything from subquery would be good to have.

intech: I agree, I'll do it.

lib/query/builder.js
```diff
@@ -1008,36 +1008,45 @@ assign(Builder.prototype, {
     return this;
   },
 
-  // Remove everything from select clause
+  // Deprecated. Remove everything from select clause
```
kibertoad, line 1011: Can you please add non-deprecated alternative in comments to every deprecated method?

kibertoad: @elhigu Do you think we should output deprecation warning here as well?

elhigu: I think those separate clear methods can be deprecated as long as list of valid string literals are listed in typings.

kibertoad (changes requested):
Could you also update types in https://github.com/knex/knex/blob/master/types/index.d.ts?

kibertoad:
@intech Could you please update Typescript types? then it should be good to be merged.

intech:
@kibertoad I'll do it along with additional tests. A little more patience. Thanks.

types/index.d.ts
```diff
@@ -533,6 +533,7 @@ declare namespace Knex {
     clearOrder(): QueryBuilder<TRecord, TResult>;
     clearHaving(): QueryBuilder<TRecord, TResult>;
     clearCounters(): QueryBuilder<TRecord, TResult>;
+    clear(statement: string): QueryBuilder<TRecord, TResult>;
```
kibertoad, line 536: Could you narrow down the type of `statement` to the list of supported statements?

intech: @kibertoad I don't know typescript, I hope I did it right. I would be grateful for your review.

types/index.d.ts
```diff
@@ -424,6 +424,7 @@ declare namespace Knex {
   //
   // QueryInterface
   //
+  type Statements = "with" | "select" | "columns" | "where" | "union" | "join" | "group" | "order" | "having" | "limit" | "offset";
```
elhigu, line 427: Naming this ClearStatements would be less probably to conflict with some other random type declarations.

elhigu: Other that that this looks what I was expecting 👍

elhigu (approved):
Looks good to me. Was the documentation up to date?

intech:
> Looks good to me. Was the documentation up to date?

@elhigu Check https://github.com/knex/documentation/pull/289/files#diff-ca136b3642dd50189c1489da6c71328577609b9df883bbe284e739c2de7a7660R1110

elhigu:
> > Looks good to me. Was the documentation up to date?
> 
> @elhigu Check https://github.com/knex/documentation/pull/289/files#diff-ca136b3642dd50189c1489da6c71328577609b9df883bbe284e739c2de7a7660R1110

Old methods could be marked as deprecated 🤔

intech:
@elhigu there is not a single deprecated method in the documentation, I do not know how best to do it. Do you have ideas?

intech:
Perhaps this option will work?

elhigu:
Looks perfect 👍

intech:
@elhigu in the commit code the menu item is added. In the screenshot I made a visualization in devtools to see :)

intech:
It will remain for me to cover all statements with tests, I doubt that `clearCounter()` will work correctly there
https://github.com/knex/knex/blob/993c2104c54a49dddf3cee8e1dc59a2d54282972/lib/query/builder.js#L969-L970

is cleared and
https://github.com/knex/knex/blob/993c2104c54a49dddf3cee8e1dc59a2d54282972/lib/query/builder.js#L1253-L1255

occurs in the `_clearGroups ()` method.
I think it will throw an error. I recommend waiting for the test coverage. I'll do it by the end of the week.

intech:
I think everything.
There is a separate situation with the `.clearCounters()` method I removed the description of `deprecated` on it.
In the new method `.clear()` added 2 more values ​​for the parameter `counter` and `counters`, both are aliases for the method `.clearCounters()`.

It looks right in my opinion, I didn't complicate one method, but added sugar to it.

kibertoad (approved):

kibertoad:
Thanks a lot!

kibertoad:
Released in 0.21.9

Issue #4045, opened 2020-09-22
Duplicated semicolons on oracle toString() / toQuery()


guicovred:
# Environment

Knex version: 0.21.5
Database + version: Oracle
OS: any
@atiertant

# Bug

`toQuery()` & `toSql()` are returning duplicate semicolons for queries that pass through the method `wrapSqlWithCatch()`

Example: 

```
var knex = require('knex')({
    client: 'oracledb',
    connection: {
        host: '127.0.0.1',
        user: 'your_database_user',
        password: 'your_database_password',
        database: 'myapp_test'
    }
});

var builder = knex.schema.dropTableIfExists('book');

console.log(builder.toQuery())
```
outputs: 
```
begin execute immediate 'drop table "book"'; exception when others then if sqlcode != -942 then raise; end if; end;;
begin execute immediate 'drop sequence "book_seq"'; exception when others then if sqlcode != -2289 then raise; end if; end;
```

It happens because both `wrapSqlWithCatch()` and `toQuery()` are appending a semicolon to end of query.
https://github.com/knex/knex/blob/master/lib/dialects/oracle/utils.js#L30
https://github.com/knex/knex/blob/master/lib/interface.js#L14

kibertoad:
Released in 0.21.7

PR #4052, opened 2020-09-29
Ensure that semicolon is not appended to statements that already end with a semicolon

guicovred:
This solves #4045:

`toQuery()` & `toSql()` are returning duplicate semicolons for queries that pass through the method `wrapSqlWithCatch()`

Example: 

```
var knex = require('knex')({
    client: 'oracledb',
    connection: {
        host: '127.0.0.1',
        user: 'your_database_user',
        password: 'your_database_password',
        database: 'myapp_test'
    }
});

var builder = knex.schema.dropTableIfExists('book');

console.log(builder.toQuery())
```
outputs: 
```
begin execute immediate 'drop table "book"'; exception when others then if sqlcode != -942 then raise; end if; end;;
begin execute immediate 'drop sequence "book_seq"'; exception when others then if sqlcode != -2289 then raise; end if; end;
```

### DIFF
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -11,7 +11,7 @@ module.exports = function (Target) {
       .map((statement) => {
         return this.client._formatQuery(statement.sql, statement.bindings, tz);
       })
-      .join(';\n');
+      .reduce((a, c) => a.concat(a.endsWith(';') ? '\n' : ';\n', c));
   };
 
   // Create a new instance of the `Runner`, passing in the current object.
--- a/test/unit/schema/oracle.js
+++ b/test/unit/schema/oracle.js
@@ -963,4 +963,12 @@ describe('Oracle SchemaBuilder', function () {
       expect(spy.thirdCall.args).to.deep.equal(['users', 'table context']);
     });
   });
+
+  it('test converting a sql wrapped with catch to string, #4045', function () {
+    tableSql = client.schemaBuilder().dropTableIfExists('book');
+
+    expect(tableSql.toQuery()).to.equal(
+      'begin execute immediate \'drop table "book"\'; exception when others then if sqlcode != -942 then raise; end if; end;\nbegin execute immediate \'drop sequence "book_seq"\'; exception when others then if sqlcode != -2289 then raise; end if; end;'
+    );
+  });
 });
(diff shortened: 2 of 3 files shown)


kibertoad:
Thanks! Could you please add test for this? Also CI seems to be failing.

guicovred:
Of course, in reality this does not seem to be the solution, I will take a closer look.

guicovred:
Well, I believe it will now work as expected. But I don't know where to put the tests.

This specific case would be testing whether converting a set of queries, some ending with semicolons, would work as expected, with all of them ending with just one semicolon, except the last, where nothing would be appended.

The query builder already has tests to verify that most queries are converted correctly, the schema builder does not and that would require refactoring a lot for such a small adjustment I think.

It is not something strongly related to a dialect, nor to the query builder, nor to the schema builder and nor a specific statement. 

What do you think?

kibertoad:
@guicovred Is it possible to write a test that would have been failing prior to this change, but no longer does after it?

guicovred:
Ok, this is possible. I was thinking about something more generic before. I added the test. Thanks!

guicovred:
Seems like this failed CI tests was introduced by nodejs v14.13 (20 hours ago).

[esm-interop.spec.js#L164](https://github.com/knex/knex/blob/master/test/cli/esm-interop.spec.js#L164) and [esm-interop.spec.js#L179](https://github.com/knex/knex/blob/master/test/cli/esm-interop.spec.js#L179)

Apparently he exception is from here: [lexer.js#L892](https://github.com/nodejs/node/blob/1e8cb08edcbbfe01e7ef186a09d4781b33b490cc/deps/cjs-module-lexer/lexer.js#L892)

kibertoad:
#4054 4054

kibertoad:
@guicovred Please pull from master, that should fix it.

guicovred:
> @guicovred Please pull from master, that should fix it.

Done. Thanks!

guicovred:
@kibertoad Is there still anything i can do to get it merged?

kibertoad (approved):

kibertoad:
Thanks a lot!

guicovred:
I thank you!

PR #4053, opened 2020-09-30
MSSQL '?' solution

KalebMatthews:
Now escapes properly if literal '?' is needed.

Change credit to @caihaibin1991 (unable to link).

```
Coded and tested. '?' will be replaced with @p# while '\?' will remain as '?'.
```


elhigu:
Missing regression tests

caihaibin1991:
@KalebMatthews thank you

test/unit/dialects/mssql.js
```diff
@@ -30,4 +30,26 @@ describe('MSSQL unit tests', () => {
       'select * from [projects] where [id" = 1 UNION SELECT 1, @@version -- --] = ?'
     );
   });
+
+  it('should include "?" in statement', async () => {
```
elhigu, line 34: test naming is pretty cryptic, maybe following would be more clear about tests intent:

`should not convert escaped "?" -marks in raw SQL to native parameter bindings` 

the same applies to second test:

`should convert "?" -marks in raw SQL to native parameter binding`

kibertoad:
@KalebMatthews Could you please pull latest changes from master?

kibertoad:
Thank you!

PR #4056, opened 2020-10-02
Avoid lodash typecheks

TrySound:
Lodash is quite big project. Even with direct imports it loads [tons](https://github.com/knex/knex/pull/3804) of
code and still bloats node_modules. This is a problem especially since lodash
mostly used as a polyfill for modern js features.

In this diff I attempted to reduce lodash usage by replacing type checks
with `typeof` operator which might be sufficient.

Also replaced lodash/isObject with custom simplified utility which does not
consider functions as objects and allows to simplify code in one place.

### DIFF
--- a/lib/dialects/oracle/query/compiler.js
+++ b/lib/dialects/oracle/query/compiler.js
@@ -6,7 +6,6 @@ const compact = require('lodash/compact');
 const identity = require('lodash/identity');
 const isEmpty = require('lodash/isEmpty');
 const isPlainObject = require('lodash/isPlainObject');
-const isString = require('lodash/isString');
 const reduce = require('lodash/reduce');
 const QueryCompiler = require('../../../query/compiler');
 const { ReturningHelper } = require('../utils');
@@ -78,7 +77,7 @@ class QueryCompiler_Oracle extends QueryCompiler {
 
     const sql = {};
 
-    if (isString(insertData)) {
+    if (typeof insertData === 'string') {
       return this._addReturningToSqlAndConvert(
         `insert into ${this.tableName} ${insertData}`,
         returning
--- a/lib/dialects/oracle/query/compiler.js
+++ b/lib/dialects/oracle/query/compiler.js
@@ -9,6 +9,7 @@ const isPlainObject = require('lodash/isPlainObject');
 const reduce = require('lodash/reduce');
 const QueryCompiler = require('../../../query/compiler');
 const { ReturningHelper } = require('../utils');
+const { isString } = require('../../../util/is');
 
 const components = [
   'columns',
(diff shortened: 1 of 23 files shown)


lib/dialects/oracle/query/compiler.js
```diff
@@ -78,7 +77,7 @@ class QueryCompiler_Oracle extends QueryCompiler {
 
     const sql = {};
 
-    if (isString(insertData)) {
```
kibertoad, line 81: can you extract this to an internal isString function as well, so that it would be easy for us to make it more robust if needed?

lib/migrate/Migrator.js
```diff
@@ -476,7 +474,7 @@ class Migrator {
   _useTransaction(migrationContent, allTransactionsDisabled) {
     const singleTransactionValue = get(migrationContent, 'config.transaction');
 
-    return isBoolean(singleTransactionValue)
+    return typeof singleTransactionValue === 'boolean'
```
kibertoad, line 477: same, could you please introduce an internal isBoolean method?

lib/migrate/Migrator.js
```diff
@@ -30,7 +28,7 @@ inherits(LockError, Error);
 class Migrator {
   constructor(knex) {
     // Clone knex instance and remove post-processing that is unnecessary for internal queries from a cloned config
-    if (isFunction(knex)) {
+    if (typeof knex === 'function') {
```
kibertoad, line 31: same for functions

TrySound:
Is this really necessary? All cases covers old browsers behaviours or oldschool usage of Number/String/Boolean as constructor. IMO such utilities only complicates developing because they always need to be found and then imported.

elhigu:
> Is this really necessary? All cases covers old browsers behaviours or oldschool usage of Number/String/Boolean as constructor. IMO such utilities only complicates developing because they always need to be found and then imported.

Yep it is necessary, otherwise there will be different ways to check the same thing every where.

TrySound:
@elhigu @kibertoad You can check master for "typeof" keyword and find a bunch of the same checks. So do you really want to introduce different ways to check the same thing instead of fixing it? Code was not consistent and you want to keep it this way?

elhigu:
@TrySound did you check out all the different ways how people can do various type checkings in javascript? It is better to have those helpers and fix non consistent places to use them. That way it will be clear how checks should be done and they will be easier to change if some corner case is found (the latter reason is just complementary... the first one is actually important enough that refactoring should be done).

TrySound:
Ok. Done.

kibertoad (approved):

kibertoad:
Thank you!

PR #4057, opened 2020-10-03
Add optional chaning.

futirstanjo:
Hi, my name is Junior, i have one suggest for you.

Knex ❤️ 

Its possible use optional chaining here for a better syntax ?!

ty for attention, have a nice day


kibertoad:
Thank you for your contribution!
Unfortunately, optional chaining is only available in Node 14+, and we still have to support Node 10. Therefore we can't merge this PR at this time. I'm sorry.

PR #4058, opened 2020-10-03
Make toSQL().toNative() work for Raw to match the API for QueryBuilder

### LINKED ISSUE #3628: toNative is not implemented for Raw
# Environment

Knex version: 0.20.8
Database + version: pg 9.6
OS: Linux

@lorefnon 

# Bug

This should work according to the TypeScript types, but it throws:

` knex.raw("id = ?", ["hello"]).toSQL().toNative()`

The error is:

`TypeError: knex.raw(...).toSQL(...).toNative is not a function`

The relevant TypeScript types are:

```typescript
interface Raw<TResult = any>
  extends events.EventEmitter,
    ChainableInterface<ResolveResult<TResult>> {
  toSQL(): Sql;
  // ...
}

interface Sql {
  method: string;
  options: any;
  bindings: Value[];
  sql: string;
  toNative(): SqlNative; // <--
}
```

Code to reproduce:

```typescript
import * as knex from "knex"
const kpg = knex({client: "pg"})
knex.raw("id = ?", ["hello"]).toSQL().toNative()
```

lorefnon:
Resolves #3628


elhigu:
Implementation looks good, but it is missing regression test.

elhigu:
And thanks for adding this, it has been a bit painful of not being able to do that with knex.raw 😅

test/tape/raw.js
```diff
@@ -91,3 +91,15 @@ test('raw bindings are optional, #853', function (t) {
...
   t.deepEqual(sql.bindings, [4]);
 });
+
+test('Allows retrieval of raw query through toNative', function (t) {
+  t.plan(2);
+  t.equal(raw('count(*) as user_count, status').toSQL().toNative(), {
+    sql: 'count(*) as user_count, status',
+    bindings: []
+  });
+  t.equal(raw('select * from users where id = ?', [1]).toSQL().toNative(), {
```
elhigu, line 101: Could that raw dialect to be changed something else that mysql, because with it the binding format is exactly the same between `.toSQL()` and `.toSQL().toNative()`.

kibertoad: @lorefnon

kibertoad:
@lorefnon Do you need any help with this?

kibertoad:
@elhigu I've added tests for other dialects, but they still seem to return exact same values?.. Is this to be expected?..

kibertoad:
@lorefnon @elhigu I've updated tests to use actual dialect-specific clients, now there is a difference. Could you please check if output is what it should be now?

PR #4060, opened 2020-10-05
Replace mkdirp with recursive flag

TrySound:
Ref https://nodejs.org/docs/latest/api/fs.html#fs_fs_mkdir_path_options_callback

In node 10 mkdir got new flag to support mkdirp like functionality.


kibertoad:
This is great, thank you!

kibertoad (approved):

PR #4064, opened 2020-10-13
use tarn configs instead of generic-pool

zfben:
use tarn configs instead of generic-pool


kibertoad:
Could you please pull from master?

zfben:
> Could you please pull from master?

Yep, please review again :)

kibertoad (approved):

kibertoad:
Released in 0.21.12

PR #4078, opened 2020-10-24
Fixes build issue related to Node.js 14

wickedest:
The error message change from node v14.13.0 to v14.14.0, breaking a number of unit tests.

If all is good, please kindly tag the PR as "hacktoberfest-accepted" for credit towards [hacktoberfest](https://hacktoberfest.digitalocean.com/faq).

### DIFF
--- a/test/cli/esm-interop.spec.js
+++ b/test/cli/esm-interop.spec.js
@@ -859,7 +859,7 @@ describe('esm interop and mjs support', () => {
   });
 
   for (const spec of fixture) {
-    it.only(spec.title, async function () {
+    it(spec.title, async function () {
       const {
         testCase,
         knexfile = 'knexfile.js',
--- a/test/cli/esm-interop.spec.js
+++ b/test/cli/esm-interop.spec.js
@@ -162,7 +162,9 @@ const fixture = [
     knexArgs: ['migrate:latest', isNode10 && `--esm`],
     expectedErrorMessage: isNode10
       ? 'Error: Cannot load module from .mjs'
-      : semver.major(process.version) === 14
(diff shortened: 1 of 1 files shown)


test/cli/esm-interop.spec.js
```diff
@@ -162,8 +162,8 @@ const fixture = [
     knexArgs: ['migrate:latest', isNode10 && `--esm`],
     expectedErrorMessage: isNode10
       ? 'Error: Cannot load module from .mjs'
-      : semver.gte(process.version, 'v14.13.0')
-      ? 'Unexpected export statement in CJS module'
+      : semver.major(process.version) === 14
```
kibertoad, line 165: Are you sure about this change? This behaviour is different across specific 14.x versions, see https://github.com/knex/knex/pull/4055/files, so I would expect that less granular condition is going to break older versions.

kibertoad: So I assume that correct fix would separate between pre-14.13.0, 14.13.0 and post-14.13.0

wickedest: That all depends - are these tests meant to be run on arbitrary minor versions, or are we talking the _latest_ minor version.  The CI is installing 14:
image

I reckoned that the intention is to always test on the latest minor/patch.  But y, if it is important for the test framework to run on 14.13.0, then an additional specific check would be required.

kibertoad: Considering that this is a test and adding additional check is not affecting runtime, I would prefer to be flexible with version support here.

wickedest: Tweaked.

kibertoad:
Thank you!

PR #4080, opened 2020-10-25
change .map to .forEach

ggorlen:
It doesn't make sense to use .map here; this allocates and produces an array of `undefined` (the return value of `console.log`) only to be tossed to the garbage collector. Better to use `forEach` if no modifications are applied to the array elements and the intent is to produce a side effect (logging).

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,9 +81,9 @@ knex.schema
       .select('users.user_name as user', 'accounts.account_name as account')
   )
 
-  // map over the results
+  // Display the results
   .then(rows =>
-    rows.map(row => {
+    rows.forEach(row => {
       console.log(row)
     })
   )


kibertoad:
@ggorlen I would say it's more of an imperfect documentation in the first place; usually developers map query results to transform the data, almost never they do just the logging part. Hence a better improvement would be to add some kind of mapping logic in the existing documentation. Could you do that?

ggorlen:
@kibertoad I see what you mean. However, if I add another promise block before the `forEach` block like:

```js
  // Optionally, map the results to select a property
  .then(rows => 
    rows.map(row => row.user)
  )
```
it begs the question, why not simply exclude the `'accounts.account_name as account'` portion of the query? This sends the wrong message about where to filter database results.

Another try:
```js
  // Map the results, if desired, to add a new property
  .then(rows => 
    rows.map(row => ({...row, active: true}))
  )
```
Now the query is correct, but it's getting a bit unnecessarily complex and potentially confusing for newcomers. A final try, going for practicality:
```js
  // Respond with the results
  .then(rows => 
    response.json(rows)
  )
```
Great, unless the reader isn't using knex in a web application context and has no idea what `response.json` is/does. Unlikely but possible.

Do you have a more naturally-motivated yet general transformation on the results in mind? If it's understood that the purpose is a contrived example, then maybe the original is sufficient, but the `forEach` still seems like the cleanest way to show consuming results for this case, with the implication that it is a stand-in for a `map`, `response.json` and so forth being pretty easy to grasp for the typical reader. Maybe making the comment explicit like `// Consume the results` or `// Consume the results as needed for the application` would better indicate that it's not _just_ about displaying them.

Maybe I'm overthinking it!

kibertoad:
That's a fair point. I think you can do some kind of data transformation there, e. g. derive some field based on combination of two columns, e. g. isActive = row.finishedAt= null and row.startedAt != null.

ggorlen:
`finishedAt`/`startedAt` don't appear to be columns in these table schemas, so I think the `rows.map(row => ({...row, active: true}))` idea would illustrate this sort of transformation without changing the schema, with the caveats mentioned above.

kibertoad:
@ggorlen Incorporated your suggestion into https://github.com/knex/knex/commit/5f34fc170ac02090ce359c54bf77bfcd624b14d6, although I'm not sure how successful my attempt to list you as a co-author was :-/

Issue #4072, opened 2020-10-21
knex-cli prints usage help text twice


emilwihlander:
# Environment

Knex CLI version: 0.21.6
Knex Local version: 0.21.2
Database + version: N/A
OS: Windows 10

# Bug

When writing `knex` in Powershell it prints the usage help text twice, it should only be printed once.

```
PS C:\***> knex
Usage:  [options] [command]

Options:
  -V, --version                   output the version number
  --debug                         Run with debugging.
  --knexfile [path]               Specify the knexfile path.
  --knexpath [path]               Specify the path to knex instance.
  --cwd [path]                    Specify the working directory.
  --client [name]                 Set DB client without a knexfile.
  --connection [address]          Set DB connection without a knexfile.
  --migrations-directory [path]   Set migrations directory without a knexfile.
  --migrations-table-name [path]  Set migrations table name without a knexfile.
  --env [name]                    environment, default: process.env.NODE_ENV || development
  --esm                           Enable ESM interop.
  --specific [path]               Specify one seed file to execute.
  --timestamp-filename-prefix     Enable a timestamp prefix on name of generated seed files.
  -h, --help                      display help for command

Commands:
  init [options]                          Create a fresh knexfile.
  migrate:make [options] <name>           Create a named migration file.
  migrate:latest [options]                Run all migrations that have not yet been run.
  migrate:up [<name>]                     Run the next or the specified migration that has not yet been run.
  migrate:rollback [options]              Rollback the last batch of migrations performed.
  migrate:down [<name>]                   Undo the last or the specified migration that was already run.
  migrate:currentVersion                  View the current version for the migration.
  migrate:list|migrate:status             List all migrations files with status.
  migrate:unlock                          Forcibly unlocks the migrations lock table.
  seed:make [options] <name>              Create a named seed file.
  seed:run [options]                      Run seed files.
  help [command]                  display help for command
Usage: cli [options] [command]

Options:
  -V, --version                   output the version number
  --debug                         Run with debugging.
  --knexfile [path]               Specify the knexfile path.
  --knexpath [path]               Specify the path to knex instance.
  --cwd [path]                    Specify the working directory.
  --client [name]                 Set DB client without a knexfile.
  --connection [address]          Set DB connection without a knexfile.
  --migrations-table-name [path]  Set migrations table name without a knexfile.
  --env [name]                    environment, default: process.env.NODE_ENV || development
  --esm                           Enable ESM interop.
  --specific [path]               Specify one seed file to execute.
[...]

PR #4082, opened 2020-10-26
fix help text being printed twice (#4072)

emilwihlander:
When the cli is called without arguments it prints the help text twice. This pull request makes it print it only once. See issue #4072

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -352,10 +352,6 @@ function invoke(env) {
         .catch(exit);
     });
 
-  if (!process.argv.slice(2).length) {
-    commander.outputHelp();
-  }
-
   commander.parse(process.argv);
 }
 
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "chai": "^4.2.0",
     "chai-as-promised": "^7.1.1",
     "chai-subset-in-order": "^2.1.4",
-    "cli-testlab": "^2.0.0",
+    "cli-testlab": "^2.1.1",
     "coveralls": "^3.1.0",
     "cross-env": "^7.0.2",
     "dtslint": "3.6.14",
--- a/test/cli/esm-interop.spec.js
+++ b/test/cli/esm-interop.spec.js
@@ -43,7 +43,6 @@ const fixture = [
     testCase: 'knexfile-esm-module',
     knexArgs: ['migrate:latest'],
     dropDb: true,
-    expectedOutput: 'Batch 1 run: 1 migrations',
     expectedErrorMessage: isNode10
       ? 'Unexpected token export'
       : 'Must use import to load ES Module',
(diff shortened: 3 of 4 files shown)


kibertoad:
Could you please add CLI test for this?

emilwihlander:
I looked at it but with the library used to test the cli there was no easy way to check if something appears multiple times as far as I could see. The tests do test whether or not it prints at least once at the moment and it may be overkill to add a complicated test for a bug that doesn't have any function-breaking effect.

kibertoad:
@emilwihlander Wouldn't `notExpectedOutput` with duplicated text do the trick?

emilwihlander:
By using something like `notExpectedOutput: ['usage', 'usage']`?

My understanding of the cli-testlab library is that if you specify multiple strings for `notExpectedOutput` it checks if any of the strings exist rather than if all exist. Therefore I don't find a reasonable way to test for multiple occurrences.

This is the function used by the library:
```
function assertNotOutput(
  stdout: string,
  notExpectedOutput: string[] | string | undefined,
  rejectFn: (error: AssertionError) => void
) {
  const notExpectedOutputs = toStringArray(notExpectedOutput)

  const errors: string[] = []
  if (notExpectedOutputs && notExpectedOutputs.length > 0) {
    notExpectedOutputs.forEach((notExpectedEntry: string) => {
      if (stdout.includes(notExpectedEntry)) {
        errors.push(`Expected output not to include "${notExpectedEntry}", but it was actually "${stdout}"`)
      }
    })
  }
  if (errors.length > 0) {
    rejectFn(new AssertionError(errors.join('\n')))
  }
}
```

kibertoad:
@emilwihlander Can't this be done by using one single long assertion?

kibertoad:
If not, I will extend cli-testlab to support occurrence counting.

emilwihlander:
I could add the entire help text printed plus ` + '\nusage'`, I guess. The test would break (always succeed) if the help text would be changed though since the specific help text would never match again. That isn't desirable.

emilwihlander:
You could either add a flag (maybe `expectedOutputOnce`?) or extend the current option to accept a object to specify the number of times a string should occur, e.g. `expectedOutput: {'usage': 1}` to make the checking easier :)

emilwihlander:
I just thought of a possible workaround, 
```
it('Prints help when no arguments are given', () => {
  return execCommand(`node ${KNEX}`, {
    expectedErrorMessage: 'Process exited with error',
    expectedOutput: 'Usage',
    notExpectedOutput: '\nUsage',
  });
});
```
should work right? I don't have the test suite running on my computer so I can't test locally, could you try @kibertoad?

kibertoad:
@emilwihlander Ended up extending cli-testlab to have more explicit assertions, let's see if CI passes now :)

kibertoad (approved):

kibertoad:
Thank you!

PR #4083, opened 2020-10-26
Make toQuery behavior consistent with pre-0.21.7

kibertoad:
Accept empty schema builder.

fixes https://github.com/mikro-orm/mikro-orm/issues/961

### DIFF
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -7,6 +7,9 @@ module.exports = function (Target) {
   Target.prototype.toQuery = function (tz) {
     let data = this.toSQL(this._method, tz);
     if (!Array.isArray(data)) data = [data];
+    if (!data.length) {
+      return '';
+    }
     return data
       .map((statement) => {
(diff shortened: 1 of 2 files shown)


kibertoad:
@B4nan Could you take a look? I would like to release fix tonight to make sure latest knex version is fully compatible with mikro-orm.

lib/interface.js
```diff
@@ -7,6 +7,9 @@ module.exports = function (Target) {
   Target.prototype.toQuery = function (tz) {
     let data = this.toSQL(this._method, tz);
     if (!Array.isArray(data)) data = [data];
+    if (!data.length) {
```
kibertoad, line 10: Result of breakage was introduction of `reduce` after .map - `reduce` throws an error on empty array when there is no default collector.

B4nan:
Yes I can confirm this fixed my tests too.

kibertoad:
@lookfirst It returns an array, an empty array, if there is nothing. Changing that logic would be a veeeeery breaking change :)

PR #4084, opened 2020-10-26
Fix comment escaping for MySQL and PostgreSQL

### LINKED ISSUE #4081: Knex does not escape single quote in comment function
# Environment

Knex version: 0.21.6
Database + version: MySQL 8.0.22
OS: Windows 10 (Docker)

# Bug
When using the comment function the single quote (') isn't escaped and therefore makes the query invalid.

When running the following command:
```
knex.schema.createTable('test', t => {
  t.text('column1').primary().comment("The table's first column")
  t.text('column2').comment("The table's second column")
})
  .then(r => console.log('success'))
  .catch(e => console.log(e))
```
with this knexfile:
```
const defaults = {
  client: 'mysql2',
  version: process.env.MYSQL_VERSION,
  connection: {
    host: process.env.MYSQL_HOST,
    user: 'root',
    password: process.env.MYSQL_ROOT_PASSWORD,
    database: process.env.MYSQL_DATABASE,
  },
  migrations: {
    tableName: 'knex_migrations'
  }
}
```
The following error is printed:
```
Error: 
create table `test` (
    `column1` text comment 'The table's first column',
    `column2` text comment 'The table's second column'
)
- You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 's first column', `column` text comment 'The table's second column')' at line 1
```

As seen, the single quote (') isn't escaped as \\' and therefore makes the query invalid. I haven't tried this with other databases but the insert function seems to escape the single quote when I tested.

Here's an example in runkit: https://runkit.com/emilwihlander/5f97014c4458ca001ad49a69

kibertoad:
fixes #4081

### DIFF
--- a/lib/dialects/mysql/schema/columncompiler.js
+++ b/lib/dialects/mysql/schema/columncompiler.js
@@ -4,6 +4,8 @@ const { inherits } = require('util');
 const ColumnCompiler = require('../../../schema/columncompiler');
 const { isObject } = require('../../../util/is');
 
+const commentEscapeRegex = /(?<!\\)'/g;
+
 function ColumnCompiler_MySQL() {
   ColumnCompiler.apply(this, arguments);
   this.modifiers = [
@@ -150,7 +152,7 @@ Object.assign(ColumnCompiler_MySQL.prototype, {
         'Your comment is longer than the max comment length for MySQL'
       );
     }
-    return comment && `comment '${comment.replace("'", "\\'")}'`;
+    return comment && `comment '${comment.replace(commentEscapeRegex, "\\'")}'`;
   },
 
   first() {
--- a/lib/dialects/mysql/schema/columncompiler.js
+++ b/lib/dialects/mysql/schema/columncompiler.js
@@ -150,7 +150,7 @@ Object.assign(ColumnCompiler_MySQL.prototype, {
         'Your comment is longer than the max comment length for MySQL'
       );
     }
-    return comment && `comment '${comment}'`;
+    return comment && `comment '${comment.replace("'", "\\'")}'`;
   },
 
   first() {
--- a/lib/dialects/postgres/schema/columncompiler.js
+++ b/lib/dialects/postgres/schema/columncompiler.js
@@ -4,6 +4,7 @@
 const { inherits } = require('util');
 const ColumnCompiler = require('../../../schema/columncompiler');
 const { isObject } = require('../../../util/is');
+const commentEscapeRegex = /(?<!')'(?!')/g;
 
 function ColumnCompiler_PG() {
   ColumnCompiler.apply(this, arguments);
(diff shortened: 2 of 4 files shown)


kibertoad:
@emilwihlander @wickedest Could you please review?

lib/dialects/mysql/schema/columncompiler.js
```diff
@@ -4,6 +4,8 @@ const { inherits } = require('util');
 const ColumnCompiler = require('../../../schema/columncompiler');
 const { isObject } = require('../../../util/is');
 
+const commentEscapeRegex = /(?<!\\)'/g;
```
wickedest, line 7: Maybe this is normal for knex (e.g. to play it safe), but I wouldn't think it should concern itself with users pre-escaping their inputs to a particular dialect.  The interface, `comment(string)` is simple enough and should escape whatever is necessary to be escaped for the particular dialect.

lib/dialects/postgres/schema/columncompiler.js
```diff
@@ -101,13 +102,16 @@ Object.assign(ColumnCompiler_PG.prototype, {
...
     const columnName = this.args[0] || this.defaults('columnName');
+    const escapedComment = comment
+      ? `'${comment.replace(commentEscapeRegex, "''")}'`
+      : 'NULL';
 
     this.pushAdditional(function () {
       this.pushQuery(
         `comment on column ${this.tableCompiler.tableName()}.` +
           this.formatter.wrap(columnName) +
           ' is ' +
```
wickedest, line 113: ' is ${escapedComment}'

emilwihlander (commented):
At first, I was unsure how new support for negative lookbehind regex was in node js but it seems to be supported since 8.10.0 and since knex is > 10 it shouldn't be a problem.

The test cases cover the problems I've had! LGTM.

lib/dialects/mysql/schema/columncompiler.js
emilwihlander: My quickfix for this bug was to escape the string myself. It's quite possible that other people have done the same and that would make their current code produce a different result. But I do agree with you, it shouldn't care if the quote is pre-escaped.

kibertoad: @wickedest I'm concerned that otherwise this would have been a breaking change, since previously this functionality would only work if users escaped themselves, and that is no longer going to work.

wickedest: better to play it safe then

wickedest (approved):

lib/dialects/mysql/schema/columncompiler.js
ackqlik: hey this is broken for me.

PR #4089, opened 2020-10-29
Replace uuid with copied nanoid logic

### LINKED ISSUE #3774: Switch to v1 uuids since 0.20.5 causes issues when mocking global date
The switch from v4 to v1 uuids for _knexQueryUids in v0.20.5 (https://github.com/knex/knex/pull/3596) actually caused us a few issues.

Long story short, it is trivial to artificially trigger uuidv1's 10M operations/sec limit when running a longish test suite with a mocked date object that does not advance. Our use of knex.raw() was somewhere in the thousands when our integration tests started throwing errors.

Not sure if it's something you want to address and I did work around it, but given that uuidv4's performance was claimed not to be a bottleneck in practice (https://github.com/knex/knex/issues/3595), it is arguable that knex suddenly becoming time-sensitive in a minor release is an order of magnitude worse than the original problem.

kibertoad:
Makes ids more easily mockable.
fixes #3774

### DIFF
--- a/lib/query/compiler.js
+++ b/lib/query/compiler.js
@@ -15,7 +15,7 @@ const isEmpty = require('lodash/isEmpty');
 const map = require('lodash/map');
 const omitBy = require('lodash/omitBy');
 const reduce = require('lodash/reduce');
-const uuid = require('uuid');
+const { nanoid } = require('../util/nanoid');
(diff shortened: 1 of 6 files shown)


kibertoad:
@benwiggins @cipacda Could you please take a look?

elhigu (approved):

PR #4091, opened 2020-10-30
Remove redundant async in tests

artursvonda:
This fixes the issue that was causing tests to fail although I'm not sure what's the actual cause. Might be a TypeScript error. The `async` in this case was redundant anyway. And it also works if we do
```
  await knex.transaction((trx) => {
    const result = await trx.insert({ name: 'Old Books' }, 'id').into('articles');
    return result
});
```
Anyway, this should fix test errors for now without losing any safety of tests.

### DIFF
--- a/types/test.ts
+++ b/types/test.ts
@@ -1262,12 +1262,12 @@ const main = async () => {
   // Transactions:
 
   // $ExpectType any[]
-  await knex.transaction(async (trx) => {
+  await knex.transaction((trx) => {
     return trx.insert({ name: 'Old Books' }, 'id').into('articles');
   });
 
   // $ExpectType Pick<Article, "id" | "subject">[]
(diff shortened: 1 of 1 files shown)


rijkvanzanten:
```js
await knex.transaction((trx) => {
    const result = await trx.insert({ name: 'Old Books' }, 'id').into('articles');
    return result
});
```

Shouldn't work right?

The function context of the callback in `transaction()` isn't an async function, and you can't `await` anything outside of an `async` function 🤔

artursvonda:
Ah, sorry. There's a typo. There should be async there for transaction callback as you would expect. Good catch! :)

kibertoad (approved):

rijkvanzanten:
@kibertoad are you sure this works as expected? See my comment above

PR #4092, opened 2020-10-30
Add support for `dropForeign` in SQLite

rijkvanzanten:
Fixes https://github.com/knex/knex/issues/2753, fixes https://github.com/knex/knex/issues/1598

I loosely based this of the existing `dropIndex` and `dropColumn` / `renameColumn` methods. I wasn't able to reuse the `_doReplace` method, as that's hardcoded to only rename/drop columns.

I was trying to add tests for this as well, but had some trouble getting started with those. I was searching for tests for `dropColumn` and `renameColumns` to base it of, but those don't seem to exist 🤔


kibertoad:
@rijkvanzanten 

```
/home/travis/build/knex/knex/lib/dialects/sqlite3/schema/ddl.js
  317:45  error  'from' is not defined             no-undef
  337:46  error  Unnecessary escape character: \`  no-useless-escape
  337:53  error  Unnecessary escape character: \`  no-useless-escape
```

CI failure seems to be due to trivial linting issues.

kibertoad:
On a subject of `dropColumn` tests - we have some for PostgreSQL (see postgres.js), I think it should be fairly easy to repurpose them for SQLite as well.

rijkvanzanten:
@kibertoad Updated the method and manually tested it to make sure it actually works 🙂  Should be ready to merge now. 

I tried going through the tests to see if I could copy over something from the postgres side, and kinda got lost in the sauce.. I did ran into this though:

https://github.com/knex/knex/blob/f953ff4f8f632b9d4e967e6072ac99a3da1e5fbb/test/integration/schema/index.js#L1887-L1926

Is this secretly already testing dropForeign on sqlite, or am I totally misreading this?

kibertoad:
@rijkvanzanten Note the `if ... else` clause. I think original code is not testing `dropForeign` per se, just doing the cleanup that is not possible in SQLite. It should be probably possible to create a new test by removing the if/then clause and running dropForeign on SQLite as well, although I would imagine that if it was originally failing with `'Named foreign key not correctly created.'`, then some other adjustments might be needed as well.

kibertoad:
Linting is failing: 
```
 336:46  error  Unnecessary escape character: \`  no-useless-escape
  336:53  error  Unnecessary escape character: \`  no-useless-escape
```

kibertoad:
@rijkvanzanten If you are having problems with tests, I can write some for you.

rijkvanzanten:
@kibertoad That would be much appreciated, thank you

kibertoad:
@rijkvanzanten Could you please take a look at the test that I added? It is failing for me for some reason. One thing that does not seem to be supported is dropping foreign key with custom key name, but even if I modify test not to depend on that, it still fails when trying to insert row which would violate removed foreign key.
You can easily reproduce this locally without any setup by changing `foreign-keys.js` line 
```
  describe('Schema', () => {
```
into
```
  describe.only('Schema', () => {
```

and then running `npm run test:sqlite`.

Note that due to enabling foreign key check for SQLite integration tests, some existing tests started to fail, but these should be more or less easy fixes.

rijkvanzanten:
Will do, thanks for setting up the tests 💪

kibertoad:
@rijkvanzanten Is there anything I can help with on this PR? Really would love to get this into next minor release before we switch to next breaking version.

rijkvanzanten:
Heya! I haven't had the time to look back into this unfortunately.. I could really use your help figuring out those tests though! 

> into next minor release before we switch to next breaking version.

What is the rough timeline for this? Is `knex` going `1.0.0`?!

kibertoad:
@rijkvanzanten Plan is to release 0.22.0 with #4122, #3702 and probably some TypeScript changes. Then the only remaining thing that I want in before releasing 1.0.0 is #2796 that is already mid-way.
So I wouldn't hold my breath for 1.0.0 in December, but early next year is definitely a possibility.

Any specific questions about the tests?

kibertoad:
And with regards to 0.22.0, I'll get on prepping it right after this PR is done :)

rijkvanzanten:
Nice! 

> One thing that does not seem to be supported is dropping foreign key with custom key name,

This is correct, the dropForeign method is based around parsing the table create SQL and does some string splitting etc based on the column name. I can take a peak at extending this to also support index names 👍 

I was mostly wondering about:

> even if I modify test not to depend on that, it still fails when trying to insert row which would violate removed foreign key.

Knex failing when trying to insert a row which violates a key constraint sounds correct to me? Maybe I'm misunderstanding what you're saying 🙂

kibertoad:
@rijkvanzanten What I mean is this:

```
        await knex.schema.alterTable('foreign_keys_table_one', (table) => {
          table.dropForeign(['fkey_two']);
          table.dropForeign([], 'fk_fkey_threeee');
        });

        await knex('foreign_keys_table_one').insert({
          fkey_two: 9999,
          fkey_three: 99,
        });
```

Since we dropped the foreign keys in the first block, second block should no longer fail. But it does. Which for me looks like we are not actually dropping the foreign keys.

rijkvanzanten:
> `table.dropForeign([], 'fk_fkey_threeee');`

I think it makes sense that this one fails currently. Like you previously correctly stated, deleting by key name isn't yet supported in this PR

kibertoad:
@rijkvanzanten Yeah, but that's what I mean by `even if I modify test not to depend on that`. Try removing ` .withKeyName('fk_fkey_threeee');` and make both dropForeign operations use the first param. the test still fails.

rijkvanzanten:
Ahhh I see. Weird.. I've been using this method from my fork for a little bit now with success. I'll pick it back up once my schedule is opening up a little bit! (Hopefully early next week) 😬 (We're going through a pretty big launch rn, so things got a little hectic)

kibertoad:
@rijkvanzanten Good luck with the launch! Hectic week here as well, I should actually be finishing the practice exercity for candidate tryout tomorrow :D. A fun night is ahead.

rijkvanzanten:
@kibertoad, I finished implementing the foreign key name support, only to realize that SQLite doesn't even support setting custom names for foreign key constraints ☹️

How do you want to proceed? We could throw a warning when the user tries setting a custom key constraint name I suppose..

rijkvanzanten:
@kibertoad Added support for custom key names!

There's still 3 more tests failing.. but from what I can tell, those are unrelated to `dropForeign`? It seems like the implementation for `renameColumn` doesn't take foreign key constraints into account 🤔 

```
  1) Integration Tests
       sqlite3 | sqlite3
         Schema
           renameColumn
             without mappers
               renames the column:
     Error: DROP TABLE "rename_column_test" - SQLITE_CONSTRAINT: FOREIGN KEY constraint failed
  

  2) Integration Tests
       sqlite3 | sqlite3
         Schema
           renameColumn
             without mappers
               successfully renames a column referenced in a foreign key:
     Error: DROP TABLE "rename_column_test" - SQLITE_CONSTRAINT: FOREIGN KEY constraint failed
  

  3) Integration Tests
       sqlite3 | sqlite3
         Schema
           renameColumn
             without mappers
               successfully renames a column referenced by another table:
```

rijkvanzanten:
On further inspection, I think it's this `after()` that doesn't clean up the foreign key constraints on drop table:

```
after(() =>
          knex.schema
            .dropTable('rename_column_foreign_test')
            .dropTable('rename_column_test')
            .dropTable('rename_col_test')
        );
```

kibertoad:
@rijkvanzanten Looking into it now, thanks for the heads-up!

rijkvanzanten:
Awesome! I'll be around [on Discord](https://directus.chat) or [your Gitter](https://gitter.im/tgriesser/knex) if you want to discuss anything synchronously 👍

kibertoad (approved):

kibertoad:
@rijkvanzanten We did it!

PR #4102, opened 2020-11-01
Types: Make .ignore() and .merge() return QueryBuilder rather than QueryInterface

nicoburns:
Fixes part (2) from https://github.com/knex/knex/issues/4101


kibertoad (approved):

kibertoad:
Released in 0.21.12

PR #4105, opened 2020-11-02
Fix types for insert to allow array

artursvonda:
Looks like I broke array insert type. This adds tests for it and fix as well.


kibertoad (approved):

kibertoad:
Released in 0.21.13

PR #4109, opened 2020-11-04
Move TS tests to tsd, part 1

kibertoad:
tsd is more explicit and IDE-friendly than dtslint comment magic.

Also TSLint is deprecated, so future of dtslint is uncertain.


lorefnon:
Yes, tsd looks much better. My biggest pain point with dtslint was that it would do a string comparision. So tests could fail because type signature was not formatted in the exact same way or the representation was not the same (eg. tests broke when TS made aliases like `Pick<Foo>` lazy). tsd doesn't seem to suffer from the same issues.

PR #4119, opened 2020-11-15
Add types for userParams and withUserParams

theGC:
Add types to index.d.ts for missing `knex.userParams` param and `knex.withUserParams({})` method, as suggested in issue: https://github.com/knex/knex/issues/3962

The types are pretty open as looking at the source there don't seem to be any restrictions on what the user can pass, so long as it's a literal object, the values can be anything.


kibertoad (approved):

kibertoad:
thanks!

PR #4122, opened 2020-11-17
Drop liftoff

TrySound:
Liftoff has very poor support and a lot of legacy dependencies.
The package takes half of knex size.
https://packagephobia.com/result?p=liftoff
https://packagephobia.com/result?p=knex

The package is not abandoned though there are no clear deadlines.
https://github.com/js-cli/js-liftoff/pull/114

I also found most liftoff features are not used by knex and in some
cases the usage is wrong.

For example I found `--knexpath` is not used at all. It's consumed
from liftoff's env.knexpath which is always undefined.

I reimplemented all resolutions with a few small utilities like escalade
and resolve-from.

v8flags and flagged-respawn looks like black magic so I suggest to
recommend using `NODE_OPTIONS="...flags..." npm run knex` instead.

For config transpilation still used interpret and rechoir which is used
in liftoff internally.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
-const Liftoff = require('liftoff');
+const rechoir = require('rechoir');
 const interpret = require('interpret');
+const resolveFrom = require('resolve-from');
+const escalade = require('escalade/sync');
 const path = require('path');
 const tildify = require('tildify');
 const commander = require('commander');
@@ -17,7 +19,7 @@ const {
   getSeedExtension,
   getStubPath,
 } = require('./utils/cli-config-utils');
-const { readFile, writeFile } = require('./../lib/util/fs');
+const { existsSync, readFile, writeFile } = require('./../lib/util/fs');
 
 const { listMigrations } = require('./utils/migrationsLister');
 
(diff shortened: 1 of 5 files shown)


bin/cli.js
```diff
@@ -57,19 +59,103 @@ async function initKnex(env, opts) {
   return knex(resolvedConfig);
 }
 
-function invoke(env) {
-  env.modulePath = env.modulePath || env.knexpath || process.env.KNEX_PATH;
+function findUpModulePath(cwd, packageName) {
```
kibertoad, line 62: Could you please move all the new util code into separate file(s)?

TrySound: bin/utils/cli-config-utils.js?

kibertoad: Not sure if all of new functions are config-related, but yeah, you can introduce new files there as well.

kibertoad:
@elhigu Do you think this should go in as a semver major?

kibertoad:
@TrySound Thanks! This is a great change and very appreciated.

TrySound:
@kibertoad Yes, since I dropped v8flags support. Sure, welcome.

kibertoad:
I'll use the opportunity to get "the infamous triplet" for types in as well then :)

bin/cli.js
TrySound: modulePath is probably config related

elhigu:
> @elhigu Do you think this should go in as a semver major?

Yep, that would be the safest.

kibertoad:
@TrySound Could you please update https://github.com/knex/knex/blob/master/UPGRADING.md with instructions about using `NODE_OPTIONS="...flags..." npm run knex` for 0.22.0?

TrySound:
Hi @kibertoad is this good to go in 0.22?

kibertoad:
@TrySound Yup! Still waiting for #4148 to land and @rijkvanzanten wanted to get createForeignKey functionality in for the last 0.21 release, but after that I'm starting the 0.22 and landing this PR.

TrySound:
Looks like travis is dying

kibertoad:
@TrySound Travis is indeed at its last legs, but in this case it seems to be a Docker configuration issue on Travis end: `You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit`.

Either way, it's good we mostly migrated to GitHub Actions already :).

Btw, I'm working on wrapping up the loose ends on 0.21 today, hopefully will merge this PR already, thank you for your patience!

kibertoad (approved):

kibertoad:
@TrySound BTW, does this implementation automatically supports TypeScript files, or we need to manually add some autoloaders?

TrySound:
I think all loaders are listed in https://github.com/gulpjs/interpret

kibertoad:
@TrySound Do they work out-of-the-box or require something to be done on knex or its users' side?

TrySound:
I believe the same as before
https://github.com/js-cli/js-liftoff/blob/master/lib/register_loader.js#L10

kibertoad:
Gotcha, thanks!

Issue #4129, opened 2020-11-23
Eliminate high coupling to `debug`


Download:
## Feature Description

Reduce coupling and footprint of minified file by using `anylogger` i.s.o `debug`

### The problem

Currently, this library has [a dependency](https://github.com/knex/knex/blob/v0.16.2/package.json#L15) on [`debug`](https://npmjs.com/package/debug). Though that is an excellent library, this dependency has 2 major drawbacks:

* This library is now forcing `debug` onto all developers that use this library (high coupling)
* `debug` is 3.1kB minified and gzipped, directly adding 3.1kB to the minimum footprint of this library

### The solution

Please have a look at [`anylogger`](https://npmjs.com/package/anylogger). It's a logging facade specifically designed for libraries. It achieves these goals:

* Decouple the library from the underlying logging framework
* Reduce the minimal bundle footprint. Anylogger is only 370 bytes.

The decoupling is achieved by only including the minimal facade to allow client code to do logging and using adapters to back that facade with an actual logging framework. The reduced minimal footprint follows naturally from this decoupling as the bulk of the code lives in the adapter or logging library.

There are already adapters for some popular logging frameworks and more adapters can easily be created:

* [`anylogger-console`](https://npmjs.com/package/anylogger-console) (to use the console i.s.o some logging framework)
* [`anylogger-debug`](https://npmjs.com/package/anylogger-debug)
* [`anylogger-loglevel`](https://npmjs.com/package/anylogger-loglevel)
* [`anylogger-log4js`](https://npmjs.com/package/anylogger-log4js)
* [`ulog`](https://npmjs.com/package/ulog) (logger with native anylogger support)

If this library were to switch to `anylogger`, you could still install `debug` as a dev-dependency and then `require('anylogger-debug')` in your tests to have your tests work exactly as they always did, with `debug` as the logging framework, while still decoupling it from `debug` for all clients.

### Workarounds

We can use a Webpack alias to alias `debug` to `anylogger`, then use an anylogger adapter to route it to the logging framework we choose, but it requires specific configuration in Webpack and it would be much better if it worked like this out of the box.

### Contributions

I Would be willing to contribute a Pull Request for this feature if the Knex developers agree it would be useful

### Disclaimer
[...]

PR #4135, opened 2020-11-25
Replaced debug and console with anylogger

Download:
* Uninstalled debug as runtime dependency
* Installed anylogger as runtime dependency to replace debug
* Installed debug and anylogger-debug adapter as development dependencies
* Required anylogger i.s.o debug everywhere debug was used
* Created an anylogger logger everywhere console was used
* Removed eslint exceptions for no-console
* Added a parameter `level` to `logger._log` so e.g. error messages will be actually logged with console.error. This allows stderr redirection.
* Required enylogger-debug in testInitializer.js so tests keep using debug
* Replaced console in testInitializer.js with an anylogger logger to complete decoupling from the console

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
-/* eslint no-console:0, no-var:0 */
+/* eslint no-var:0 */
+const log = require('anylogger')('knex:cli');
 const Liftoff = require('liftoff');
 const interpret = require('interpret');
 const path = require('path');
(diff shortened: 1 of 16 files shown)


Download:
Hi, I'm not expecting you guys to merge this, I just made this PR so we can have a more concrete discussion about #4129

mercmobily:
Anylogger doesn't seem to be actively developed anymore. Closing it.

Download:
What development do you need? It is a *facade*, meant to be small and unchanging....

PR #4137, opened 2020-11-25
Document gotcha for running mssql tests locally

kibertoad:


kibertoad:
@elhigu Have you ever encountered this issue yourself? Running the code on Windows is the only way how I can explain why it works in CI but fails locally, although I have no idea why it would make a difference for Docker environment.

elhigu:
Nope I have never seen that 🤔

kibertoad:
@elhigu Does it work locally without a hitch for you? Which OS are you running?

elhigu:
I've been using mostly MacOS for knex work, but also I haven't ran mssql (or any other) knex tests for a long time.

PR #4138, opened 2020-11-25
chore/dont do typecheck of tsd tests, because tsd already does that

maximelkin:


kibertoad (approved):

kibertoad:
@maximelkin So the problem is due to mismatch between tsc and tsd tsconfigs?

maximelkin:
Tsd somehow ignores expectError errors, but tsc doesn't.
Existing tests not using expectError, so we got this error only now.

PR #4139, opened 2020-11-26
Added type for withKeyName

### LINKED ISSUE #4063: Missing support for withKeyName in TypeScript type declarations
# Environment

Knex version: 0.21.6
Database + version: Not relevant
OS: Not relevant

Typescript issue: @lorefnon 

# Bug

I am trying to rename a foreign key. Minimal example:

```ts
import * as Knex from "knex";

export async function up(knex: Knex): Promise<void> {
  await knex.schema.createTable("xyz", (t) => {
    t.increments("id");
    
    t.foreign("id").references("another.table").withKeyName("cg_link");
  });
}
```

2. Error message

`Property 'withKeyName' does not exist on type 'ReferencingColumnBuilder'.ts(2339)`

See above

Alucard17:
Closes #4063

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1747,6 +1747,7 @@ declare namespace Knex {
     comment(value: string): ColumnBuilder;
     alter(): ColumnBuilder;
     queryContext(context: any): ColumnBuilder;
+    withKeyName(keyName: string): ColumnBuilder;
   }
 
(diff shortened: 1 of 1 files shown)


kibertoad (approved):

kibertoad:
Thanks!

kibertoad:
Released in 0.21.13

PR #4147, opened 2020-12-07
Support non-standard Oracle port

### LINKED ISSUE #4146: Can't connect to OracleDB using host+port combo
# Environment

Knex version: 0.21.12
Database + version: OracleDB (any)
OS: any

# Bug

When connecting to an OracleDB database using a combination of host+port, Knex isn't able to acquire the connection (fails with `ORA-12170: TNS:Connect timeout occurred`) This is most likely because of:

https://github.com/knex/knex/blob/073df48cc6f31ba0ea1f732341d15b631fe1d386/lib/dialects/oracledb/index.js#L99-L101

I believe changing `client.connectionSettings.host + '/' + client.connectionSettings.database` to `client.connectionSettings.host + ':' + client.connectionSettings.port + '/' + client.connectionSettings.database` might do the trick 

Ref https://github.com/directus/directus/issues/3266

kibertoad:
fixes #4146

Unfortunately, no integration test for this, as we'd need a separate instance of Oracle for this :-/

### DIFF
--- a/lib/dialects/oracledb/index.js
+++ b/lib/dialects/oracledb/index.js
@@ -81,6 +81,24 @@ Client_Oracledb.prototype.prepBindings = function (bindings) {
   });
 };
 
+function resolveConnectString(connectionSettings) {
+  if (connectionSettings.connectString) {
+    return connectionSettings.connectString;
+  }
+
+  if (!connectionSettings.port) {
+    return connectionSettings.host + '/' + connectionSettings.database;
+  }
+
+  return (
+    connectionSettings.host +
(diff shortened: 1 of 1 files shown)


rijkvanzanten (approved):
LGTM!

I tried this locally, and I'm able to connect properly now with the following setup:

```js
const database = knex({
  client: 'oracledb',
  connection: {
    host: 'localhost',
    port: 12345, // << Note the non-standard port
    user: 'xxx',
    password: 'xxx',
    database: 'xxx',
  }
});
```

PR #4148, opened 2020-12-07
Add support for WHERE clauses to "upsert" queries

### LINKED ISSUE #4145: Add support for WHERE clause to upsert
# Environment

Knex version: 0.12.12
Database + version: Postgres 10.7
OS: MacOS Catalina 10.15.7

I am using the new "upsert" feature added in #3763, which works great! But using PostgreSQL, I need to add a WHERE clause to an "upsert" statement (ON CONFLICT ...), so that I can conditionally "upsert" only certain rows. 

I looked through the Github issues and through the #3763 PR which added support for "upsert" and it doesn't seem like this feature is currently supported.

The PostgreSQL documentation shows the syntax for using a WHERE clause in INSERT ... ON CONFLICT queries: https://www.postgresql.org/docs/10/sql-insert.html.

An example of how this feature is used: https://stackoverflow.com/questions/44571870/postgresql-on-conflict-with-a-where-clause

Adding support for a WHERE clause to "upsert" statements (ON CONFLICT ...) in DB engines that support it would meet my needs.

The PostgreSQL compiler (and perhaps MySQL/SQLite compilers?) needs to be updated to support a `.where` method after an `onConflict().merge()` chain.

For PostgreSQL at least, I think the changes would go somewhere around here:
https://github.com/knex/knex/blob/master/lib/dialects/postgres/query/compiler.js#L29

markdboyd:
This is a follow-up to #3763, which adds support for WHERE clauses to "upsert" queries for the DB engines that support this functionality.

Based on my research on the DB engines that support "upsert" queries, it seems like PostgreSQL and SQLite support WHERE clauses for them, but MySQL does not:

https://www.postgresql.org/docs/10/sql-insert.html
https://sqlite.org/lang_upsert.html
https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html

To do:

- [x] Add support for WHERE clause to upsert queries in PostgreSQL and add tests
- [x] Add support for WHERE clause to upsert queries in SQLite and add tests
- [x] Update documentation: https://github.com/knex/documentation/pull/300

Closes #4145

### DIFF
--- a/lib/dialects/mysql/query/compiler.js
+++ b/lib/dialects/mysql/query/compiler.js
@@ -26,7 +26,13 @@ class QueryCompiler_MySQL extends QueryCompiler {
 
     const { ignore, merge, insert } = this.single;
     if (ignore) sql = sql.replace('insert into', 'insert ignore into');
-    if (merge) sql += this._merge(merge.updates, insert);
+    if (merge) {
+      sql += this._merge(merge.updates, insert);
+      const wheres = this.where();
+      if (wheres) {
+        throw new Error('.onConflict().merge().where() is not supported for mysql');
+      }
+    }
 
     return sql;
   }
--- a/lib/dialects/postgres/query/compiler.js
+++ b/lib/dialects/postgres/query/compiler.js
@@ -26,10 +26,11 @@ class QueryCompiler_PG extends QueryCompiler {
 
     const { returning, onConflict, ignore, merge, insert } = this.single;
     if (onConflict && ignore) sql += this._ignore(onConflict);
-    if (onConflict && merge)
+    if (onConflict && merge) {
       sql += this._merge(merge.updates, onConflict, insert);
       const wheres = this.where();
       sql += wheres ? ` ${wheres}` : '';
+    }
     if (returning) sql += this._returning(returning);
 
     return {
--- a/lib/dialects/postgres/query/compiler.js
+++ b/lib/dialects/postgres/query/compiler.js
@@ -28,6 +28,8 @@ class QueryCompiler_PG extends QueryCompiler {
     if (onConflict && ignore) sql += this._ignore(onConflict);
     if (onConflict && merge)
       sql += this._merge(merge.updates, onConflict, insert);
+      const wheres = this.where();
+      sql += wheres ? ` ${wheres}` : '';
     if (returning) sql += this._returning(returning);
 
     return {
--- a/lib/dialects/postgres/query/compiler.js
+++ b/lib/dialects/postgres/query/compiler.js
@@ -29,7 +29,7 @@ class QueryCompiler_PG extends QueryCompiler {
     if (onConflict && merge) {
       sql += this._merge(merge.updates, onConflict, insert);
       const wheres = this.where();
-      sql += wheres ? ` ${wheres}` : '';
+      if (wheres) sql += ` ${wheres}`;
     }
     if (returning) sql += this._returning(returning);
 
--- a/test/integration/builder/inserts.js
+++ b/test/integration/builder/inserts.js
@@ -1499,6 +1499,11 @@ module.exports = function (knex) {
               'insert into "upsert_tests" ("email", "name") values (?, ?) on conflict ("email") do update set "email" = excluded."email", "name" = excluded."name" where "upsert_tests"."role" = ? returning "email"',
               ['mergetest@example.com', 'AFTER', 'tester']
             );
+            tester(
+              'sqlite3',
+              'insert into `upsert_tests` (`email`, `name`) values (?, ?) on conflict (`email`) do update set `email` = excluded.`email`, `name` = excluded.`name` where `upsert_tests`.`role` = ? returning `email`',
+              ['mergetest@example.com', 'AFTER', 'tester']
+            );
           });
       } catch (err) {
         if (/oracle|mssql/i.test(knex.client.driverName)) {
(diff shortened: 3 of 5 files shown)


lib/dialects/postgres/query/compiler.js
```diff
@@ -26,8 +26,11 @@ class QueryCompiler_PG extends QueryCompiler {
 
     const { returning, onConflict, ignore, merge, insert } = this.single;
     if (onConflict && ignore) sql += this._ignore(onConflict);
-    if (onConflict && merge)
+    if (onConflict && merge) {
       sql += this._merge(merge.updates, onConflict, insert);
+      const wheres = this.where();
+      sql += wheres ? ` ${wheres}` : '';
```
kibertoad, line 32: I would simlify this to a `if (wheres)` check instead of empty concatenation, same as here:
`if (returning) sql += this._returning(returning);`

test/integration/builder/inserts.js
```diff
@@ -1465,6 +1465,110 @@ module.exports = function (knex) {
...
+          .where('upsert_tests.role', 'tester')
+          .testSql(function (tester) {
+            tester(
+              'pg',
+              'insert into "upsert_tests" ("email", "name") values (?, ?) on conflict ("email") do update set "email" = excluded."email", "name" = excluded."name" where "upsert_tests"."role" = ? returning "email"',
+              ['mergetest@example.com', 'AFTER', 'tester']
+            );
+          });
+      } catch (err) {
+        if (/oracle|mssql/i.test(knex.client.driverName)) {
```
kibertoad, line 1504: Skipping test altogether same as is done for `redshift` probably would be more transparent in test results log.

test/integration/builder/inserts.js
```diff
@@ -1465,6 +1465,110 @@ module.exports = function (knex) {
       expect(rows[0].name).to.equal('AFTER');
     });
 
+    it('conditionally updates rows when inserting a duplicate key to unique column and merge with where clause matching row(s) is specified', async function () {
+      if (/redshift/i.test(knex.client.driverName)) {
```
kibertoad, line 1469: shouldn't mysql be also skipped if you are saying it is not supported?

kibertoad:
Since you are still planning to add support for SQLite, that will resolve the currently failing test. Could you please add the sanity check for MySQL, which would throw an error if someone tries to use it with WHERE clause, and adjust assertion accordingly?

lib/dialects/postgres/query/compiler.js
markdboyd: Will do

test/integration/builder/inserts.js
markdboyd: I don't fully understand the test setup - but it seems like the difference between redshift and oracle/mssql is that for oracle/mssql we expect an error to be thrown, but not redshift? Should I preserve that distinction or just add mssql/oracle to the skip?

lib/dialects/postgres/query/compiler.js
markdboyd: 706ce6e42513c52547c045827a504a32991e941b

markdboyd:
@kibertoad I think I have addressed your feedback, minus one question I had about the integration tests

lib/dialects/mysql/query/compiler.js
```diff
@@ -26,7 +26,13 @@ class QueryCompiler_MySQL extends QueryCompiler {
 
     const { ignore, merge, insert } = this.single;
     if (ignore) sql = sql.replace('insert into', 'insert ignore into');
-    if (merge) sql += this._merge(merge.updates, insert);
+    if (merge) {
+      sql += this._merge(merge.updates, insert);
+      const wheres = this.where();
+      if (wheres) {
```
markdboyd, line 32: It feels like there might be a better way to check whether `where` is called after a `merge` statement for MySQL, but I could not figure out what it is

test/integration/builder/inserts.js
kibertoad: TBH, I think we don't even run redshift in CI, so feel free to just keep ignoring it.
BTW, shouldn't MySQL also be included in error assertion?

markdboyd: Ah yeah. Added MySQL to have an error assertion:

90cba493009f09ceb4457c4040cfb1da406129d2

test/integration/builder/inserts.js
```diff
@@ -1465,6 +1465,130 @@ module.exports = function (knex) {
       expect(rows[0].name).to.equal('AFTER');
     });
 
+    it('conditionally updates rows when inserting a duplicate key to unique column and merge with where clause matching row(s) is specified', async function () {
+      if (/redshift|mysql/i.test(knex.client.driverName)) {
```
kibertoad, line 1469: since you are asserting the error now, probably no longer you need to skip it

markdboyd: Sorry about that. Fixed: 80f366d67fd90d4792470a3418668849c54daa36

kibertoad:
This can be merged after documentation is added to https://github.com/knex/documentation

markdboyd:
@kibertoad OK, I have created a documentation PR: https://github.com/knex/documentation/pull/300

kibertoad (approved):

kibertoad:
Thanks a lot!

kibertoad:
Released in 0.21.13

PR #4153, opened 2020-12-10
Add support for application_name in PG connection types

aitskovi:
Adds support for application_name as a valid type.

This option is already supported in `pg`, see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pg/index.d.ts

Also confirmed that it passes through correctly. Let me know if I should also be adding a test for this typing change.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2020,6 +2020,7 @@ declare namespace Knex {
     connectionTimeoutMillis?: number;
     keepAliveInitialDelayMillis?: number;
     ssl?: boolean | ConnectionOptions;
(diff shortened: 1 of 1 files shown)


kibertoad (approved):

kibertoad:
Thank you! I think it's a simple enough change not to need test :)

kibertoad:
Released in 0.21.13

aitskovi:
Awesome, thanks for the quick turnaround.

kibertoad:
It was my pleasure!

PR #4154, opened 2020-12-10
Fix performance of asyncStackTraces with enable-source-maps node flag

istarkov:
Reading stack trace with '--enable-source-maps' flag on node can be very costly, 
in our case when using knex it takes 100-200ms for every query what in case of multiple queries gives huge delays.

Instead of reading stack trace at saveAsyncStack function better to read it when needed at catch error.

### DIFF
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -25,9 +25,14 @@ module.exports = function (Target) {
       result = result.catch((err) => {
         err.originalStack = err.stack;
         const firstLine = err.stack.split('\n')[0];
-        this._asyncStack.unshift(firstLine);
+
+        const { error, lines } = this._asyncStack;
+        const stackByLines = error.stack.split('\n');
+        const asyncStack = stackByLines.slice(lines);
+        asyncStack.unshift(firstLine);
+
         // put the fake more helpful "async" stack on the thrown error
-        err.stack = this._asyncStack.join('\n');
+        err.stack = asyncStack.join('\n');
         throw err;
       });
     }
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -26,6 +26,9 @@ module.exports = function (Target) {
         err.originalStack = err.stack;
         const firstLine = err.stack.split('\n')[0];
 
+        // a hack to get a callstack into the client code despite this
+        // node.js bug https://github.com/nodejs/node/issues/11865
+        // see lib/util/save-async-stack.js for more details
         const { error, lines } = this._asyncStack;
         const stackByLines = error.stack.split('\n');
         const asyncStack = stackByLines.slice(lines);
--- a/lib/util/save-async-stack.js
+++ b/lib/util/save-async-stack.js
@@ -2,8 +2,13 @@ module.exports = function saveAsyncStack(instance, lines) {
   if (instance.client.config.asyncStackTraces) {
     // a hack to get a callstack into the client code despite this
     // node.js bug https://github.com/nodejs/node/issues/11865
-    const stackByLines = new Error().stack.split('\n');
-    stackByLines.splice(0, lines);
-    instance._asyncStack = stackByLines;
+
+    // Save error here but not error trace
+    // reading trace with '--enable-source-maps' flag on node can be very costly
+
+    instance._asyncStack = {
+      error: new Error(),
+      lines,
+    };
   }
 };
--- a/test/unit/interface.js
+++ b/test/unit/interface.js
@@ -30,9 +30,13 @@ describe('interface', function () {
     }
 
     _interface(SomeClass);
+
     const fakeInstance = new SomeClass();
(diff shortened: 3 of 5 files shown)


test/unit/interface.js
```diff
@@ -30,9 +30,13 @@ describe('interface', function () {
     }
 
     _interface(SomeClass);
+
     const fakeInstance = new SomeClass();
     chai.expect(fakeInstance[Symbol.toStringTag]).to.eq('object');
-    fakeInstance._asyncStack = ['line1', 'line2', 'line3'];
+    fakeInstance._asyncStack = {
```
kibertoad, line 36: How difficult would it be to create a less synthetic unit test (e. g. for an in-memory SQLite db instance) that would have an actual asyncStack?

lib/util/save-async-stack.js
```diff
@@ -2,8 +2,13 @@ module.exports = function saveAsyncStack(instance, lines) {
...
     // a hack to get a callstack into the client code despite this
     // node.js bug https://github.com/nodejs/node/issues/11865
-    const stackByLines = new Error().stack.split('\n');
-    stackByLines.splice(0, lines);
-    instance._asyncStack = stackByLines;
+
+    // Save error here but not error trace
+    // reading trace with '--enable-source-maps' flag on node can be very costly
+
+    instance._asyncStack = {
```
kibertoad, line 9: This is a breaking change, right?

lib/util/save-async-stack.js
```diff
@@ -2,8 +2,13 @@ module.exports = function saveAsyncStack(instance, lines) {
   if (instance.client.config.asyncStackTraces) {
     // a hack to get a callstack into the client code despite this
     // node.js bug https://github.com/nodejs/node/issues/11865
```
kibertoad, line 4: So from my understanding this could be removed after Node 10 support is dropped, is that correct?

lib/interface.js
```diff
@@ -25,9 +25,14 @@ module.exports = function (Target) {
       result = result.catch((err) => {
         err.originalStack = err.stack;
         const firstLine = err.stack.split('\n')[0];
-        this._asyncStack.unshift(firstLine);
+
+        const { error, lines } = this._asyncStack;
```
kibertoad, line 29: Shouldn't the comment of
```
 // a hack to get a callstack into the client code despite this
 // a hack to get a callstack into the client code despite this
 // node.js bug https://github.com/nodejs/node/issues/11865
```
be moved here then?

test/unit/interface.js
istarkov: Less synthetic was already in tests here https://github.com/istarkov/knex/blob/master/test/integration/builder/additional.js#L1127
There was no need to change it. As output is the same.

lib/util/save-async-stack.js
istarkov: `_` means private var ;-) and not exposed in any docs, so it depends

lib/util/save-async-stack.js
istarkov: Can't be removed even after node 10, node stil has some issues with async traces

lib/interface.js
istarkov: Probably duplicated? Now it places at initialisation

test/unit/interface.js
istarkov: interface.js also is not used as I see in tests at all, probably need to be removed at all

lib/interface.js
kibertoad: So which part relates to working around that particular Node issue? Reference should be there, or both, if it applies to both :)

test/unit/interface.js
kibertoad: See test/db-less-test-suite.js:

```
  // require('./unit/interface'); ToDo Uncomment after fixed
```

Can you uncomment it?

test/unit/util/save-async-stack.js
```diff
@@ -13,7 +13,11 @@ describe('saveAsyncStack', function () {
     };
     saveAsyncStack(fakeInstance, 1);
```
kibertoad, line 15: Could you please add `test/unit/util/nanoid.js` and `test/unit/util/save-async-stack.js` into  `test/db-less-test-suite.js`?

istarkov: Done

test/unit/interface.js
istarkov: No, it didnt work before, it doesnt work now. I have no idea why and I don't have enough knowledge why it was commented and not working at all. Probably standalone PR must be created by someone smart to fix it.

lib/interface.js
istarkov: done

kibertoad (approved):

kibertoad:
Thank you!

kibertoad:
Released in 0.21.13

PR #4156, opened 2020-12-11
Fix(seed:run): irregular seed file execution order

progfay:
## Problem

- Run `knex seed:run` with `seeds: { sortDirsSeparately: true }` option
	- Run seed files with irregular order (unsorted)
	- Change order every time you run

See https://github.com/progfay/knex-irregular-file-order-poc

## Factor

- In `getFilepathsInFolder` function, push file path to array asynchronously

## Solution

- To avoid async pushing

### DIFF
--- a/lib/util/fs.js
+++ b/lib/util/fs.js
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const { flatten } = require('lodash');
 const os = require('os');
 const path = require('path');
 const { promisify } = require('util');
@@ -46,23 +47,22 @@ function ensureDirectoryExists(dir) {
  * All found files, concatenated to the current dir
  */
 async function getFilepathsInFolder(dir, recursive = false) {
-  const results = [];
   const pathsList = await readdir(dir);
-  await Promise.all(
-    pathsList.sort().map(async (currentPath) => {
-      const currentFile = path.resolve(dir, currentPath);
-      const statFile = await stat(currentFile);
-      if (statFile && statFile.isDirectory()) {
-        if (recursive) {
-          const subFiles = await getFilepathsInFolder(currentFile, true);
-          results.push(...subFiles);
+  return flatten(
+    await Promise.all(
+      pathsList.sort().map(async (currentPath) => {
+        const currentFile = path.resolve(dir, currentPath);
+        const statFile = await stat(currentFile);
+        if (statFile && statFile.isDirectory()) {
+          if (recursive) {
+            return await getFilepathsInFolder(currentFile, true);
+          }
+          return [];
         }
-      } else {
-        results.push(currentFile);
-      }
-    })
+        return [currentFile];
+      })
+    )
   );
-  return results;
 }
 
 module.exports = {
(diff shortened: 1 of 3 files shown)


test/unit/seed/seeder.js
```diff
@@ -301,6 +300,7 @@ describe('Seeder._listAll', function () {
             '/test/integration/seed/2-seeds/recursive-folder/ls-seed.ls',
           process.cwd() +
             '/test/integration/seed/2-seeds/recursive-folder/ts-seed.ts',
+          process.cwd() + '/test/integration/seed/2-seeds/ts-seed.ts',
```
kibertoad, line 303: Why change the order?

test/unit/util/fs.js
```diff
@@ -94,4 +95,47 @@ describe('FS functions', () => {
       expect(fs.existsSync(newPath)).to.equal(true);
     });
   });
+
+  describe('getFilepathsInFolder', () => {
+    it('should return sorted file list under the directory.', async () => {
+      const directory = await createTemp();
+      fs.writeFileSync(path.join(directory, 'testfile1.txt'), 'testfile1');
```
kibertoad, line 102: Can you either use `fileHelp` for automatic cleanup (see test/cli/migrate.spec.js for a usage example) or add manual cleanup steps?

kibertoad: Wait, I see you are using `createTemp()`. When is it deleted?

kibertoad: Checked the implementation, all good here, sorry.

lib/util/fs.js
```diff
@@ -46,23 +47,22 @@ function ensureDirectoryExists(dir) {
...
-  await Promise.all(
-    pathsList.sort().map(async (currentPath) => {
-      const currentFile = path.resolve(dir, currentPath);
-      const statFile = await stat(currentFile);
-      if (statFile && statFile.isDirectory()) {
-        if (recursive) {
-          const subFiles = await getFilepathsInFolder(currentFile, true);
-          results.push(...subFiles);
+  return flatten(
+    await Promise.all(
```
kibertoad, line 52: I wonder if we even need a promise here in the first place. Since seeding is usually not done constantly on the runtime, but executed on-demand, it's not performance critical, so it might make sense to simplify implementation and make it sync.

test/unit/seed/seeder.js
progfay: Comments of `Seeder._listAll`:
https://github.com/knex/knex/blob/bc04b7ab30f152a677de53e2cc6e7df995a12f98/lib/seed/Seeder.js#L51-L52

And I thought `sort` means:
1. Same order as `seeds.directory` property in `knexfile.js` (process in `Seeder._listAll`)
2. Sub-ordered by alphabet (include nested files) (process in `getFilepathsInFolder`)

Test code is violate this rule, so I change order.

lib/util/fs.js
progfay: I agree your opinion!
But, this changing make `getFilepathsInFolder`, `Seeder._listAll` and `Seeder.run` to sync function from async function.

Can I change it?

kibertoad: Great point. Double-checked how it is with Migrations, and they are async there, which is actually important, since there is also a MigrationSource abstraction there, which might be retrieving migrations e. g. from a database or other storage. Since we might eventually want to have same for Seeds, it's better to keep implementation consistent, so let's keep it the way it is.

kibertoad (approved):

kibertoad:
Thank you!

kibertoad:
Released in 0.21.13

PR #4157, opened 2020-12-11
Avoid connection getting stuck on socket hangup

### LINKED ISSUE #4128: UnhandledPromiseRejectionWarning
# Environment

Knex version: 0.21.5
Database + version: MSSQL 2017
OS: MAC OS

For MSSql tag @smorey2.

# Problem

I am using transactions running in a cron job every 15 seconds hitting the database and performing read, write and update calls bound in **async await try catch block**.

Sample Code:
```
async function updateSetup(tableId) {
    try {
        let newId;
        await dbData.transaction(async trx => {
            let id = await trx('SETUP').select('VALUEZ').where('KEY_NAME', tableId);
            newId = id[0].VALUEZ + 1;
            await trx('SETUP').update({ 'VALUEZ': newId }).where('KEY_NAME', tableId);
        })
        return newId;
    } catch (e) {
        console.log(e);
        throw e;
    }
}

```

**Test**
I was having socket hangup issues and was trying to test the code by switching my internet on and off, to check what all errors will be generated. In one particular case I get this warning  below  and it does not get caught in the catch block:

```
node:41181) UnhandledPromiseRejectionWarning: TypeError: Cannot assign to read only property 'originalError' of object 'TransactionError: Requests can only be made in the LoggedIn state, not the Final state'
    at /Users/aviral/projects/texma-sync copy/node_modules/knex/lib/dialects/mssql/transaction.js:41:38
<node_internals>/internal/process/warning.js:27
(node:41181) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 10)
<node_internals>/internal/process/warning.js:27
(node:41181) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
<node_internals>/internal/process/warning.js:27
``` 

**Issue**
After that  even if I restore the net connection my code resumes after several seconds as this below and pretty much stuck at this point, I have to restart my application manually from PM2 to get rid of this issue:

```
RequestError: select * from [MSGTABLE] where [STATUSZ] = @p0 - Connection lost - socket hang up
    at handleError (/Users/aviral/projects/texma-sync/node_modules/mssql/lib/tedious/request.js:366:15)
    at Connection.emit (<node_internals>/events.js:214:15)
    at Connection.socketError (/Users/aviral/projects/texma-sync/node_modules/tedious/lib/connection.js:1294:12)
    at Connection.socketEnd (/Users/aviral/projects/texma-sync/node_modules/tedious/lib/connection.js:1312:12)
    at Socket.<anonymous> (/Users/aviral/projects/texma-sync/node_modules/tedious/lib/connection.js:1131:14)
    at Socket.emit (<node_internals>/events.js:214:15)
    at endReadableNT (<node_internals>/_stream_readable.js:1178:12)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {code: 'EREQUEST', originalError: Error: Connection lost - socket hang up
  … (/Users/aviral/projects/texma-sync/node_…, name: 'RequestError', number: undefined, lineNumber: undefined, …}
```

I don't know how to resolve this issue, because net connectivity failure is a constant issue at production site.

kibertoad:
fix #4128

### DIFF
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -38,7 +38,6 @@ module.exports = class Transaction_MSSQL extends Transaction {
         this._rejecter(err);
       },
       (err) => {
-        if (error) err.originalError = error;
         return this._rejecter(err);
       }
     );
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -38,6 +38,13 @@ module.exports = class Transaction_MSSQL extends Transaction {
         this._rejecter(err);
       },
       (err) => {
+        if (error) {
+          try {
+            err.originalError = error;
+          } catch (_err) {
+            // This is to handle https://github.com/knex/knex/issues/4128
+          }
+        }
         return this._rejecter(err);
       }
     );


elhigu:
LGTM

kibertoad:
@elhigu So I failed to reproduce the issue locally, and there is actually a test that triggers it already, and it fails in CI now. Should we just change the assertion, or make the error handling more defensive in that place? (try-catch the field setting).

elhigu:
The test was relaying on that err.originalError exists? how could that have worked if setting it should throw an error 🤔

kibertoad:
@elhigu Because it's not that straightforward :). See original issue: 
```
    On LAN - Windows 10 - No issues, as soon as I switch off internet it gives connectivity issue and as soon as I switch internet back on it reconnects and starts sending data.
    On Cloud - Linux Ubuntu 18.04 - I've put a log below. I am connecting to the same sql server instance to two separate databases.
```
So probably it depends on particular configuration or deployment details.

elhigu:
ummm... so maybe just some certain error coming from driver or from some driver version is different that doesn't allow setting that attribute :) yeah... if we could pass that originalError information in some other more robust way it would be better, but I'm fine also with the try-catch best effort hack.

PR #4158, opened 2020-12-13
Bugfix: Uses esm import if package.json type is module

SnirShechter:
Fixes https://github.com/knex/knex/issues/4159

### DIFF
--- a/lib/util/import-file.js
+++ b/lib/util/import-file.js
@@ -5,7 +5,7 @@
  * @todo WARN on version 10 and '--experimental-modules' and '--esm'
  */
 module.exports = function importFile(filepath) {
-  return filepath.endsWith('.mjs')
+  return filepath.endsWith('.mjs') || process.env.npm_package_type === 'module'
     ? import(require('url').pathToFileURL(filepath))
     : require(filepath);
 };
--- a/lib/util/import-file.js
+++ b/lib/util/import-file.js
@@ -4,8 +4,9 @@
  * @param {string} filepath
  * @todo WARN on version 10 and '--experimental-modules' and '--esm'
  */
+const isTypeModule = process.env.npm_package_type === 'module'
 module.exports = function importFile(filepath) {
-  return filepath.endsWith('.mjs') || process.env.npm_package_type === 'module'
+  return filepath.endsWith('.mjs') || isTypeModule
     ? import(require('url').pathToFileURL(filepath))
     : require(filepath);
 };
--- a/lib/util/import-file.js
+++ b/lib/util/import-file.js
@@ -1,10 +1,12 @@
+// When run via npm, we can leverage the injected environment variables to infer the import type
+const isTypeModule = process.env.npm_package_type === 'module'
+
 /**
  * imports 'mjs', else requires.
  * NOTE: require me late!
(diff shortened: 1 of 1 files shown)


kibertoad:
@SnirShechter Could you please add reference to to documentation for `npm_package_type`? Can't find it.

kibertoad:
I've just tried running a test after having set `type` to `module` in package.json, and I don't see `npm_package_type` on my process.env

kibertoad:
OK, so it does work, but only when you execute something from npm, which makes sense.

SnirShechter:
> I've just tried running a test after having set `type` to `module` in package.json, and I don't see `npm_package_type` on my process.env. When is it set?

npm automatically adds the whole package.json as environment variables.

npm_package_type is basically package.json "type": "module"

I'm open to other suggestions if it looks hacky :)

lib/util/import-file.js
```diff
@@ -5,7 +5,7 @@
  * @todo WARN on version 10 and '--experimental-modules' and '--esm'
  */
 module.exports = function importFile(filepath) {
-  return filepath.endsWith('.mjs')
+  return filepath.endsWith('.mjs') || process.env.npm_package_type === 'module'
```
kibertoad, line 8: Could you please read `process.env.npm_package_type` value once and cache it? `process.env` calls are expensive (https://www.reddit.com/r/node/comments/7thtlv/performance_penalty_of_accessing_env_variables/)

kibertoad: Also I would swap the order of checks, if it's set per-process, no need to check filepath.

SnirShechter: Yes, of course. I wasn't aware it's expensive, good to know.

Maybe there's a better way to check for esm?

lib/util/import-file.js
```diff
@@ -4,8 +4,9 @@
  * @param {string} filepath
  * @todo WARN on version 10 and '--experimental-modules' and '--esm'
  */
+const isTypeModule = process.env.npm_package_type === 'module'
```
kibertoad, line 7: JSDoc above it belong to the function, so it would make sense to move the constant a bit higher.

lib/util/import-file.js
kibertoad: If there is, I'm not aware of it, unfortunately. I think this solution will do nicely for now. How urgently do you need this change published, btw?

SnirShechter: Not urgent, take your time.

It's for a personal project. I can always use [patch-package](https://github.com/ds300/patch-package) just in case.

kibertoad:
Thanks! If CI is green, this is good to go.

lib/util/import-file.js
kibertoad: Gotcha. We plan to do the final release in 0.21 line next week, so this one will be included.

SnirShechter: Sounds good. Thanks!

SnirShechter:
> Thanks! If CI is green, this is good to go.

Superb!

My first (tiny) Knex contribution :)

kibertoad:
And it's greatly appreciated :)

PR #4160, opened 2020-12-14
Make sure query-response and query-error events contain _knexTxId

xuanduc987:
The docs said that these events contain both __knexUid and __knexTxId but I found that it were lacking __knexTxId


kibertoad (approved):

kibertoad:
Thanks!

kibertoad:
Released in 0.21.14

PR #4161, opened 2020-12-14
Improved integration test framework

kibertoad:
This kind of a setup is much more IDE-friendly, because it does not rely on tests being executed from very specific suite, hence has significantly lower barrier to entry for new developers.


test/integration2/schema/foreign-keys.spec.js
```diff
@@ -0,0 +1,78 @@
+const { expect } = require('chai');
+const { getAllDbs, getKnexForDb } = require('../util/knex-instance-provider');
+
+describe.skip('Schema', () => {
```
kibertoad, line 4: This is an example test that is also something that @rijkvanzanten can use for their own PR, it's also a PoC for the new mechanism.

test/integration2/util/knex-instance-provider.js
```diff
@@ -0,0 +1,143 @@
+const { promisify } = require('util');
+const knex = require('../../../lib');
+
+const Db = {
+  PostgresSQL: 'postgres',
+  MySQL: 'mysql',
+  MySQL2: 'mysql2',
+  MSSQL: 'mssql',
+  SQLite: 'sqlite3',
+  Oracle: 'oracle',
```
elhigu, line 10: is oracle nowadays oracle and not oracledb ?

kibertoad: You are right, that was a mistake on my part. Fixed.

PR #4162, opened 2020-12-14
Add primary/foreign support to SQLite on alterTable

### LINKED ISSUE #3351: Allow creating foreign keys in sqlite
# Environment

Knex version: 0.18.3
Database + version: sqlite3 v4.0.7
OS: macos

# Bug / Feature request

Currently knex complaints about adding a foreign key when altering a table (`SQLite3 Foreign & Primary keys may only be added on create`). I understand why you have it there, but it is not 100% true. You can create foreign keys on existing table - you just have to add new column with that key, the thing you cannot do is to alter existing column (making it foreign key). 

Currently I am trying to use knex as a query builder/runner in my ORM (https://github.com/B4nan/mikro-orm), and the way I am talking about is how it works there currently (so it is not theoretical, it is something I already have implemented without knex - see here https://github.com/B4nan/mikro-orm/tree/master/lib/schema). 

Here is example SQL that is valid and I would like to be able to build that with knex schema builder:

```sql
ALTER TABLE "author" ADD "favourite_book_id" INTEGER DEFAULT NULL REFERENCES "book" ("id") ON DELETE SET NULL ON UPDATE CASCADE;
```

The reason why I don't want to create this FK when creating the table is that there are references in both ways - author table references book and vice versa. It is a model example, not a real world usage.

rijkvanzanten:
Fixes #3351, fixes https://github.com/directus/directus/issues/2904


lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -328,7 +328,7 @@ assign(SQLite3_DDL.prototype, {
         }
 
         const updatedDefs = defs
-          .split(',')
+          .split(/,(?![^(]*\))/g)
```
kibertoad, line 331: This regex is too complex for my brain :D. Could you extract it into a public constant and add a couple of unit tests that would document what it does?

lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -359,6 +359,106 @@ assign(SQLite3_DDL.prototype, {
...
+        const matched = oneLineSQL.match(/^CREATE TABLE\s+(\S+)\s*\((.*)\)/);
+
+        const columnDefinitions = matched[2];
+
+        if (!columnDefinitions) {
+          throw new Error('No column definitions in this statement!');
+        }
+
+        let newColumnDefinitions = columnDefinitions
+          .split(/,(?![^(]*\))/g) // comma, except when in parentheses
```
kibertoad, line 380: is this the same regex as before? can it be a reused constant?

lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -359,6 +359,106 @@ assign(SQLite3_DDL.prototype, {
...
+          .map((line) => line.trim())
+          .filter((line) => line.startsWith('primary') === false)
+          .map((line) => line.replace(/primary key/i, ''));
+
+        let primaryKeyDef = `primary key(${columns.join(',')})`;
+
+        if (constraintName) {
+          primaryKeyDef = `constraint ${constraintName} ${primaryKeyDef}`;
```
kibertoad, line 388: I would prefer a second constant here, with ternary operator assigning to it either this, or just `primaryKeyDef `

lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -359,6 +359,106 @@ assign(SQLite3_DDL.prototype, {
...
+        }
+
+        newColumnDefinitions.push(primaryKeyDef);
+        newColumnDefinitions = newColumnDefinitions.join(', ');
```
kibertoad, line 392: can we avoid mutation here and introduce a new variable instead?

lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -359,6 +359,106 @@ assign(SQLite3_DDL.prototype, {
...
+
+  foreign: async function (foreignInfo) {
+    return this.client.transaction(
+      async (trx) => {
+        this.trx = trx;
+
+        const tableInfo = (await this.getTableSql())[0];
+        const currentSQL = tableInfo.sql;
+
+        const oneLineSQL = currentSQL.replace(/\s+/g, ' ');
```
kibertoad, line 415: I see lots of overlap with previous method. maybe some code can be moved to a shared internal function?

test/integration2/schema/foreign-keys.spec.js
```diff
@@ -1,7 +1,7 @@
 const { expect } = require('chai');
 const { getAllDbs, getKnexForDb } = require('../util/knex-instance-provider');
 
-describe.skip('Schema', () => {
+describe('Schema', () => {
```
kibertoad, line 4: oh wow, it fit as-is :D

lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -359,6 +360,112 @@ assign(SQLite3_DDL.prototype, {
...
+        const matched = oneLineSQL.match(/^CREATE TABLE\s+(\S+)\s*\((.*)\)/);
+
+        const columnDefinitions = matched[2];
+
+        if (!columnDefinitions) {
+          throw new Error('No column definitions in this statement!');
+        }
+
+        let newColumnDefinitions = columnDefinitions
```
kibertoad, line 424: Could you also make this one immutable?

test/integration2/schema/primary-keys.spec.js
```diff
@@ -0,0 +1,70 @@
+const { expect } = require('chai');
+const { getAllDbs, getKnexForDb } = require('../util/knex-instance-provider');
+
+describe('Schema', () => {
+  describe.skip('Foreign keys', () => {
```
kibertoad, line 5: Name still needs to be changed, btw

rijkvanzanten:
@kibertoad Question regarding tests for primary key: seeing that Knex doesn't have a way to cross-platform retrieve the set primary key / indexes, how would you like me to test if the primary is set correctly?

I could do a similar trick to:

```js
if (knex.client.driverName === 'sqlite3') {
  expect(err.message).to.equal(
    `insert into \`foreign_keys_table_one\` (\`fkey_three\`, \`fkey_two\`) values (99, 9999) 
 SQLITE_CONSTRAINT: FOREIGN KEY constraint failed`
  );
}
```

where I use knex.raw to match it based on raw info coming from the db?

kibertoad:
@rijkvanzanten Can't you just check if unique constraint is set by trying to insert an existing id?

rijkvanzanten:
That would be a way to test the unique-nature of a PK yeah, but it doesn't really give us the full picture 🤔 Like how would we test if the custom constraint name is set correctly?

kibertoad:
@rijkvanzanten You can try deleting it, I guess, if that throws an error for non-existing constraints, that should work.

kibertoad (approved):

kibertoad:
Released in 0.21.15

PR #4164, opened 2020-12-17
Timestamp UTC Standardization

Tolterix:
Currently you are using whatever timezone the user who creates migrations has for their migration file names. This Pull will change the timestamps to all be UTC by default. I also added in a Rest API parameter to the function so you can pass it a standard offset(the const array at the top of timestamp.js) to change the timezone for the file. I would think adding a timezone option in the knexfile.js for users would allow teams to use a central timezone such as EST or JST. Right now teams that are not co-geo-located together have to finagle files to get migrate:latest to work properly since the files do not use UTC by default.

TLDR: formats all times for migrate:make and seed:make files to UTC.

### DIFF
--- a/lib/util/timestamp.js
+++ b/lib/util/timestamp.js
@@ -1,13 +1,12 @@
 function yyyymmddhhmmss(...time) {
 	const rYear = [31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
 	const timeZones = {
-		'60':   'ECT', '120':  'EET', '180':  'EAT',
-		'210':  'MET', '240':  'NET', '300':  'PLT', '330':  'IST',
-		'360':  'BST', '420':  'VST', '480':  'CTT', '540':  'JST',
-		'570':  'ACT', '600':  'AET', '660':  'SST', '720':  'NST',
-		'-660': 'MIT', '-600': 'HST', '-540': 'AST', '-480': 'PST',
-		'-420': 'MST', '-360': 'CST', '-300': 'EST',
-		'-240': 'PRT', '-210': 'CNT',
+		'60':   'ECT', '120':  'EET', '180':  'EAT', '210':  'MET',
+		'240':  'NET', '300':  'PLT', '330':  'IST', '360':  'BST',
+		'420':  'VST', '480':  'CTT', '540':  'JST', '570':  'ACT',
+		'600':  'AET', '660':  'SST', '720':  'NST', '-660': 'MIT',
+		'-600': 'HST', '-540': 'AST', '-480': 'PST', '-420': 'MST',
+		'-360': 'CST', '-300': 'EST', '-240': 'PRT', '-210': 'CNT',
 		'-180': 'BET', '-60':  'CAT'
 	}
 	
@@ -51,8 +50,8 @@ function yyyymmddhhmmss(...time) {
 		}
 	}
 	
-	dateTime[3]++;
 	dateTime[2]++;
+	dateTime[3]++;
 	
 	if ((dateTime[4]%4 !== 0) && (dateTime[2] === 29) && (dateTime[3] === 2)) {
 		if (forward) {
--- a/lib/util/timestamp.js
+++ b/lib/util/timestamp.js
@@ -1,16 +1,78 @@
-// Get a date object in the correct format, without requiring a full out library
-// like "moment.js".
-function yyyymmddhhmmss() {
-  const d = new Date();
-
-  return (
-    d.getFullYear().toString() +
-    (d.getMonth() + 1).toString().padStart(2, '0') +
-    d.getDate().toString().padStart(2, '0') +
-    d.getHours().toString().padStart(2, '0') +
-    d.getMinutes().toString().padStart(2, '0') +
-    d.getSeconds().toString().padStart(2, '0')
-  );
+function yyyymmddhhmmss(...time) {
+	const rYear = [31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
+	const timeZones = {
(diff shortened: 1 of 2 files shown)


Eyegee-Eleven:
This is a great feature for my team who is not co-geo-located! I hope the CI test gets fixed and this gets added to the main project!

lib/util/timestamp.js
```diff
@@ -1,16 +1,84 @@
...
+		'-180': 'BET', '-60':  'CAT'
+	}
+	
+	let d = new Date()
+	let offset = d.getTimezoneOffset()
+	let zoneName = 'UTC'
+	if (time.length > 0) {
+		//'2016-02-28T21:12:45.678'
+		let regex = new RegExp('^[0-9]{4}-[0-9]{2}-[0-9]{2}T([0-9]{2}:){2}[0-9]{2}\.[0-9]{3}$')
+		if ()
```
kibertoad, line 19: is this a leftover code?

lib/util/timestamp.js
```diff
@@ -1,16 +1,84 @@
...
-
-  return (
-    d.getFullYear().toString() +
-    (d.getMonth() + 1).toString().padStart(2, '0') +
-    d.getDate().toString().padStart(2, '0') +
-    d.getHours().toString().padStart(2, '0') +
-    d.getMinutes().toString().padStart(2, '0') +
-    d.getSeconds().toString().padStart(2, '0')
-  );
+function yyyymmddhhmmss(...time) {
```
kibertoad, line 1: What's the point of adding this parameter if it's never passed from anywhere? Also what's the benefit of having it over just using UTC always?

elhigu: >  Also what's the benefit of having it over just using UTC always?

I'm wondering the same.

lib/util/timestamp.js
Tolterix: This is something I was working on very late as I was working with one of my coworkers to validate times given in, as this method allows for no parameters or two parameters

lib/util/timestamp.js
Tolterix: the point of having the parameter is that if a team doesnt want UTC, lets say they are all co-located in a EST timezone, then this parameter would allow for, in the future, letting that timezone be given somewhere else. For instance, if a team wanted to all use EST, possibly if a parameter was added into the knexfile as timezone: EST, then the function could translate the offset for that if it is given in as a parameter. That sections definitely needs a bit of work depending on how that could/would get implemented in the future, I just thought teams should be given the choice of using the default of UTC or changing to their own specified timezone.

kibertoad: @Tolterix But it's never passed from knexfile into `yyyymmddhhmmss` right now.
Still, why any team might want more specific time, considering that there is no significance to the timestamp whatsoever, and it is used merely to ensure that seed/migration files are automatically sorted by the order of their creation, which is achieved with UTC perfectly?

Tolterix: The fact that it is never passed in now won't affect it at all, as I am using a Rest API with the ...time. The functionality of how you use it everywhere in the code right now wouldn't change other than changing the time to UTC and adding the string 'UTC' to the beginning of the timestamp, if you implement that part of this. The main part about the sorting and creation falls back to when things are migrated in teams. I have discussed this with a couple of my coworkers who have worked with this and they have concluded that having their own timezone would help for clarity for them dealing with migrations. @kibertoad

kibertoad: @Tolterix What would be the clarity about? Why would anyone care at what particular hour seed file was created?

Tolterix: Not the seed, but definitely when migrations happen.

Tolterix: I guess that tracking when migrations are made would help when doing multiple migrations that are very similar if you don't delete them. Might be an edge case I will admit, and UTC time as default for everyone would definitely resolve the base issue. Other timezones was just a nicety that I thought would be good to have in the future.

elhigu: I still don't see any reason for not using UTC... commit history tells when migration was created and necessarily migrations are not even applied in the same order if some feature whose migration was created later on is merged to the production first etc. Also nothing prevents your team to add any timezone or any other tag to the migration name even if it is not generated automatically. 

Why is it necessary to know in what exact time in some exact timezone certain migration file has been created? (it is not even exactly the same time when that migration contents has been written...).

Tolterix: how does migrate latest work if it isnt looking at that timestamp? or is the timestamp literally just a randomly generated id and it doesnt matter what it is? because if that is the case you could just remove the timestamp function and put 20 random characters on the front of the files name.

kibertoad: @Tolterix `latest` retrieves list of migrations that were already run, compares it against full list of migrations, and runs everything that wasn't run yet. As I've said, timestamp is used to ensure that migrations are sorted by creation date, and for that purpose always using UTC is perfectly appropriate.

Tolterix: Awesome, yeah with that knowledge it makes perfect sense to refactor it to not include all the other stuff and just use UTC. I'll refactor my repo appropriately and do another pull after.

kibertoad: Thank you!

Tolterix:
Submitted new pull request as discussed. See Timestamp UTC Standardization for Migrations #4245 @ https://github.com/knex/knex/pull/4245 for details.

kibertoad:
Released in 0.95.0-next2

kibertoad:
@Eyegee-Eleven @Tolterix Released in 0.95.0

PR #4169, opened 2020-12-21
Update index.d.ts fixes #3549

### LINKED ISSUE #3549: Property 'after' does not exist on type 'ColumnBuilder'
# Environment

Knex version: ^0.20.2
Database + version: MySQL
OS: macOS 10.14.6

_@lorefnon / typescript_

# Bug

Should be able to use after() on ColumnBuilder

2. Error message

`Property 'after' does not exist on type 'ColumnBuilder'. Did you mean 'alter'?ts(2551)`

```js
knex.schema.alterTable('users', (t: Knex.TableBuilder) => {
      t.integer('creator_id')
        .after('source') // <== No worky
        .unsigned()
      t.foreign('creator_id').references('users.id')
    })
```

nvitaterna:
add after to columnbuilder definition

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1808,6 +1808,7 @@ declare namespace Knex {
     alter(): ColumnBuilder;
     queryContext(context: any): ColumnBuilder;
     withKeyName(keyName: string): ColumnBuilder;
+    after(keyName: string): ColumnBuilder;
   }
 
   interface ForeignConstraintBuilder {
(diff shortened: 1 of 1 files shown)


kibertoad:
Thanks! Can you also add `first`? It also seems to be missing.

types/index.d.ts
```diff
@@ -1808,6 +1808,7 @@ declare namespace Knex {
     alter(): ColumnBuilder;
     queryContext(context: any): ColumnBuilder;
     withKeyName(keyName: string): ColumnBuilder;
+    after(keyName: string): ColumnBuilder;
```
kibertoad, line 1811: is `keyName` accurate here? I would imagine `field` or `column` would be more accurate.

nvitaterna: `columnName` would actually make sense, and is used slightly above in the `references` function.

kibertoad (approved):

kibertoad:
Thanks a lot!

PR #4172, opened 2020-12-25
Fix type definition of Transaction

HeavenFox:
The default of template variable `TResult` for `Transaction` is `any`, whereas for Knex it's `unknown[]`, so calling functions on transaction objects may give wrong type.

```
const a = await knex.select('*').from<Tuple>('table'); // type Tuple[]
knex.transaction(async trx => {
    const b = await trx.select('*').from<Tuple>('table'); // type Tuple | Tuple[]
})
```


kibertoad:
@lorefnon Does this look correct to you?

kibertoad:
@lorefnon Could you also take a look at this, if this is correct? I've rebased it.

types/index.d.ts
```diff
@@ -1680,7 +1680,7 @@ export declare namespace Knex {
     asCallback(callback: Function): Promise<T>;
   }
 
-  interface Transaction<TRecord extends {} = any, TResult = any>
+  interface Transaction<TRecord extends {} = any, TResult = any[]>
```
lorefnon, line 1683: Why are we changing this ? I don't think this is generally correct - transaction can wrap non-select operations too.

kibertoad: @HeavenFox could you please comment?

lorefnon: Ok, I think for `trx.select` we will need to update the default, rather than changing the result type of Transaction itself.

kibertoad: @lorefnon Do you mean
```
  interface ColumnNameQueryBuilder<TRecord = any, TResult = unknown[]> {
    // When all columns are known to be keys of original record,
    // we can extend our selection by these columns
    (columnName: '*'): QueryBuilder<
      TRecord,
      ArrayIfAlready<TResult, DeferredKeySelection<TRecord, string>>
    >;
```
or some other part?

lorefnon: Something incomprehensible is happening here. Let me check back and update in some time. I am not entirely sure how we end up with that weird union type.

lorefnon: Sorry for creating confusion here. This PR, as implemented, is good to merge. 

When using `update`, `insert` etc. we will unwrap the array type wherever relevant, so it is fine if the QueryBuilder's result type is an array type.

We end up with that erroneous union type because the `ArrayIfAlready` type is buggy. It should actually be: 

```
type ArrayIfAlready<T1, T2> = AnyToUnknown<T1> extends any[] ? T2[] : T2;
```
to prevent any type from matching both branches of conditional. 

I'll create separate PR for that.

kibertoad: No worries, thank you for verifying it!

kibertoad: @lorefnon Any chance you could still do that fix?

kibertoad: @lorefnon ping?..

kibertoad:
Released in 0.95.0

Issue #4174, opened 2020-12-26
Problem with onConflict + merge


ncsft:
# Environment

Knex version: 0.21.12
Database + version: postgesql 12
OS: all

# Bug
Error when 33k+ parameters

```
error: insert into "bug_table" ("id", "value") values ($1, $2), (...) on conflict ("id") do update set "id" = excluded."id", "value" = excluded."value" - bind message has 464 parameter formats but 0 parameters
```

```
    const rows = [];
    for (let i = 0; i < 33000; i++) {
      rows.push({ id: i, value: i });
    }
    await knex.table('bug_table')
      .insert(rows)
      .onConflict('id')
      .merge();
```
Table definition
```
CREATE TABLE "public"."bug_table" (
    "id" int4 NOT NULL,
    "value" int4 NOT NULL,
    PRIMARY KEY ("id")
);
```

kibertoad:
@ncsft Are you sure this is not a driver limitation? Can you try running same statement directly using `pg` library in Node.js?

ncsft:
> @ncsft Are you sure this is not a driver limitation? Can you try running same statement directly using `pg` library in Node.js?

No. Thats works fine
```
knex.raw(
      knex.table('bug_table').insert(rows).toQuery() + ' ON CONFLICT (id) DO UPDATE SET ' +
      Object.keys(rows[0]).map(field => `${field}=EXCLUDED.${field}`).join(', ')
    );
```

kibertoad:
@ncsft Thanks, I'll take a look. Appreciate the repro!

kibertoad:
It stops working at 32768 elements, so it really looks like a 16-bit integer overflow somewhere, but I am yet to find where.

kibertoad:
@ncsft 
Having investigated further, I believe this is not an issue on knex's behalf. There is a difference between the two queries that you have provided - native knex one in the end passes two params to `pg`:

*  `text` with the value `insert into "merge_table" ("id", "value") values ($1, $2), ($3, $4), ($5, $6), ($7, $8), ($9, $10), ($11, $12), ($13, $14), ($15, $16), ($17, $18), ($19, $20), ($21, $22), ($23, $24) <...>on conflict ("id") do update set "id" = excluded."id", "value" = excluded."value"` 
* `values` with the value `Array(65536) [0,0,1,1,2,2 <...>]`,

while `knex.raw` that you used inlines everything:
[...]

PR #4175, opened 2020-12-26
Extract query execution-related code from client

kibertoad:
refs #4174

### DIFF
--- a/lib/internal/query-executioner.js
+++ b/lib/internal/query-executioner.js
@@ -36,7 +36,9 @@ function enrichQueryObject(connection, queryParam, client) {
 function executeQuery(connection, queryObject, client) {
   return client._query(connection, queryObject).catch((err) => {
     err.message =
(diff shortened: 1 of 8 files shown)


lib/internal/query-executioner.js
```diff
@@ -0,0 +1,55 @@
...
+  if (typeof queryObject === 'string') {
+    queryObject = { sql: queryObject };
+  }
+
+  queryObject.bindings = client.prepBindings(queryObject.bindings);
+  queryObject.sql = client.positionBindings(queryObject.sql);
+
+  const { __knexUid, __knexTxId } = connection;
+
+  client.emit('query', Object.assign({ __knexUid, __knexTxId }, queryObject));
```
kibertoad, line 30: I'm not too happy about event emitting and debug logic sneakily being in a method called `enrichQueryObject`, but it is shared across both cases where we use this method.
If someone feels strongly that it's worth to extract those four lines into a separate reused method and has a good name in mind, I'm open to moving them.

elhigu (approved):

PR #4178, opened 2020-12-27
Try to introduce some organization into how files are grouped

kibertoad:
As the golden rule of complexity says, there shouldn't be more than seven sub-entities per entity (https://en.wikipedia.org/wiki/The_Magical_Number_Seven,_Plus_or_Minus_Two), so I did my best to reorganize stuff around. It is obviously very opinionated, and any ideas how to name things better or organize them into more cohesive groups would be more than welcome.

Considering how much noise there is in this PR, let me sum up the changes:

1) There are new folders in `lib` directory: `execution` (containing `transaction` and `runner`), `knex-builder` (containing `Knex`, `make-knex` and other files that are only used by them). 
2) Existing directory `schema` was moved into `schema/builder`, three more sub-folders were added - `util`, `migrate` and `seed` (and I know that technically seeding has nothing to do with schema, but logically it belongs together with migrations and schema-building, I just couldn't come up with a generic term that would combine them - it would be CLI, but both seeds and migrations can be invoked programatically as well. Please let me know if you can think of a word that applies to all three). Also `util` files that are only used by files in this folder were moved here.
3) `interface` was renamed into `builderInterfaceAugmenter`. Still not super-informative, but `interface` basically didn't mean anything at all to a person who didn't already know what it was doing. Better name suggestions would be awesome.
4) functionhelper converted into class.
5) Minor cleanup.


maximelkin:
Does query/ and schema/ have intersections beside Raw class?

maximelkin:
Not sure about second - i think merge migrate & seed ok, but builder is something separate from that.

kibertoad:
@maximelkin `lib/util/helpers.js` is also shared. It's a prime candidate for splitting, btw.
Also `lib/util/is.js` and `lib/util/save-async-stack.js`

kibertoad:
@maximelkin Fair point. Any good one word that describes both seeds and migrations?

kibertoad:
And yeah, from shared dependencies perspective `seed` and `migrate` share a lot among themselves, but nothing with `builder`, so cohesion-wise they shouldn't be combined. So I guess `schema-builder` and, ugh, "something-something DB incremental evolution".

maximelkin:
In documentation they named Migrations https://knexjs.org/#Migrations, so why not?

maximelkin:
Also knex-builder could contain things, which is actually execution/ part

kibertoad:
> In documentation they named Migrations https://knexjs.org/#Migrations, so why not?

Kinda confusing, but at least that'd be consistent, so yeah, renamed to that.

maximelkin:
Also, do you have visual schema for builder/compiler/etc usage (i.e. when doing query)?

kibertoad:
> Also knex-builder could contain things, which is actually execution/ part

Do you mean `batch-insert` and `delay`? It's currently only used inside `make-knex` that exposes it, hence the choice of a location. But yeah, I see your point, could be less confusing if we move it there.
Should we also move `FunctionHelper` to `schema` as well then? Even though it is exposed via `make-knex`, it is actually only used in schema building.

kibertoad:
> Also, do you have visual schema for builder/compiler/etc usage (i.e. when doing query)?

Not yet, but we definitely need some kind of a diagram there.

maximelkin:
FunctionHelper can be used for normal queries - `knex('a').update({updated_at: knex.now()})`

maximelkin:
Not in this pr, but we should decide about camelCase vs train-case filenames. This can confuse new contributors.

kibertoad:
@maximelkin I mostly tried to follow the train-case as it seemed to be the most common style applied in Knex codebase, but I'm open to be convinced otherwise :)

kibertoad:
@maximelkin Is there anything left before this can be approved?

kibertoad:
Merging it to unblock further refactorings, but if anyone has any additional thoughts or comments after the fact, please keep commenting, I'll address them in a separate PR.

PR #4180, opened 2020-12-28
drop fake-client and Knex.raw

maximelkin:


kibertoad (approved):

kibertoad:
Yay, less legacy code!

PR #4181, opened 2020-12-28
Introduce the "infamous triplet" export

kibertoad:
This allows users to import knex in a variety of ways, as a default or a named import, using ES6 imports or CommonJS. Makes forcing `esModuleInterop` unnecessary. Allows to clearly separate between `knex` as function that user calls, and `Knex` as namespace and interface for knex builder.


kibertoad:
@felixmosh This breaks dynamic typing for custom QueryBuilder extensions from #3334, could you by any chance take a look?

felixmosh:
I think that the augmentation of the module should be like this:

```ts
declare module '../types' {
  namespace Knex {
    interface QueryBuilder {
      customSelect<TRecord, TResult>(
        value: number
      ): Knex.QueryBuilder<TRecord, TResult>;
    }
  }
}
```

kibertoad:
@felixmosh Thanks a lot, that worked!

kibertoad:
@lorefnon @elhigu Thoughts on this change?

lorefnon:
Looks good to me. Seems like something that make life of some people easier. 

Although when using typescript this is seldom a real problem because TS would usually do the right thing for you.

PR #4183, opened 2020-12-28
Add ecosystem doc

kibertoad:


kibertoad:
@felixmosh I've noticed that you wrote some plugins for knex, would you mind if we feature them on our knex tooling list? And if you are okay with that, which ones would you like mentioned?

felixmosh:
I've 2 plugins that I'm using right now,

[knex-paginate](https://github.com/felixmosh/knex-paginate) - this one allows you to handle pagination tasks
[knex-on-duplicate-update](https://github.com/felixmosh/knex-on-duplicate-update) - which probably will be deprecated once I will migrate my code base to use `onConflict`

kibertoad:
@felixmosh Got it, thanks!

Issue #581, opened 2014-11-26
Set transaction isolation level


tgriesser:
It'd be nice to have a way to specify the isolation level on the transaction

Also mentioned by @llambda in #138, it'd be nice to have both a better api for injecting connection specific config default settings (currently it's possible with an `afterCreate` hook)

rprieto:
This would definitely be nice :+1: 
AFAIK, the current workaround looks something like (for `Postgres`):

```js
knex.transaction(function(trx) {
   return trx
   .raw('set transaction isolation level repeatable read;')
   .then(function() {
      return trx.select(/* ... */);
   });
})
```

Does this look correct to you?
And were you thinking of something like these?

```js
knex.transaction({isolation: 'repeatable read'}, function(trx) { /* ... */ });
knex.transaction.repeatableRead(function(trx) { /* ... */ });
```

bendrucker:
Your workaround looks correct. As for the new feature, I'd go with the first version but invert the arguments. Callback first, options second.

rprieto:
Thanks! I suggested the callback as the last argument for CoffeeScript / ES6 support, where the `(trx) =>` function syntax makes it much easier if there's nothing behind it. But you'd have to add argument type checking to keep it backwards-compatible. I guess otherwise we'll just extract that function and call `knex.transaction(fn, {isolation: '...'})`.

cspotcode:
@rprieto's workaround won't work for MySQL, since "SET TRANSACTION ISOLATION LEVEL" applies to the *next* transaction, not the current one.

From: https://dev.mysql.com/doc/refman/5.0/en/set-transaction.html
> Without any `SESSION` or `GLOBAL` keyword, the statement sets the isolation level for the next (not started) transaction performed within the current session. Subsequent transactions revert to using the `SESSION` isolation level.

hendrul:
:( This is a blocker for me, and is sad cuz I really like bookshelf/knex combination. If there where some work around to this for mysql I'll be very happy.

honestserpent:
Any news on this feature?

elhigu:
@rprieto's suggestion for the backwards compatible API looks good to me.

honestserpent:
Any news on this feature?

kibertoad:
@honestserpent I could try working on it. Which databases are you interested in? Would you consider contributing sample SQLs to illustrate what expected output is?

honestserpent:
I'm working on Postgres.
[...]

PR #4185, opened 2020-12-29
Add transaction isolation support

ayroblu:
Attempt to solve #581 
Probably needs some extra tests, not sure where the best place to put them is.
Also probably needs to implement for all drivers rather than just half of them

Syntax is `knex.transaction(...).setIsolationLevel('string')`

### DIFF
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -111,6 +111,7 @@ function nameToIsolationLevelEnum(level) {
       `Unknown Isolation level, was expecting one of: ${JSON.stringify(keys)}`
     );
   }
+  return knownEnum;
 }
 
 // Based on: https://github.com/tediousjs/node-mssql/blob/master/lib/isolationlevel.js
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -101,19 +101,23 @@ module.exports = class Transaction_MSSQL extends Transaction {
 /** MSSQL is annoying this way */
 function nameToIsolationLevelEnum(level) {
   if (!level) return;
-  level = level.toLowerCase();
-  switch (level) {
-    case 'read uncommitted':
-      return 'READ_UNCOMMITTED';
-    case 'read committed':
-      return 'READ_COMMITTED';
-    case 'repeatable read':
-      return 'REPEATABLE_READ';
-    case 'snapshot':
-      return 'SNAPSHOT';
-    case 'serializable':
-      return 'SERIALIZABLE';
-    default:
-      return level;
+  level = level.toUpperCase().replace(' ', '_');
+  const knownEnum = isolationEnum[level];
+  if (!knownEnum) {
+    const keys = Object.keys(isolationEnum).map((key) =>
+      key.toLowerCase().replace('_', ' ')
+    );
+    throw new Error(
+      `Unknown Isolation level, was expecting one of: ${JSON.stringify(keys)}`
+    );
   }
 }
+
+// Based on: https://github.com/tediousjs/node-mssql/blob/master/lib/isolationlevel.js
+const isolationEnum = {
+  READ_UNCOMMITTED: 0x01,
+  READ_COMMITTED: 0x02,
+  REPEATABLE_READ: 0x03,
+  SERIALIZABLE: 0x04,
+  SNAPSHOT: 0x05,
+};
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -4,7 +4,9 @@ const debug = require('debug')('knex:tx');
 module.exports = class Transaction_MSSQL extends Transaction {
   begin(conn) {
     debug('%s: begin', this.txid);
-    return conn.tx_.begin().then(this._resolver, this._rejecter);
+    return conn.tx_
+      .begin(this.isolationLevel)
+      .then(this._resolver, this._rejecter);
   }
 
   async savepoint(conn) {
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -100,6 +100,7 @@ module.exports = class Transaction_MSSQL extends Transaction {
 
 /** MSSQL is annoying this way */
 function nameToIsolationLevelEnum(level) {
+  if (!level) return;
   level = level.toLowerCase();
   switch (level) {
     case 'read uncommitted':
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -104,11 +104,10 @@ function nameToIsolationLevelEnum(level) {
   level = level.toUpperCase().replace(' ', '_');
   const knownEnum = isolationEnum[level];
   if (!knownEnum) {
-    const keys = Object.keys(isolationEnum).map((key) =>
-      key.toLowerCase().replace('_', ' ')
-    );
     throw new Error(
-      `Unknown Isolation level, was expecting one of: ${JSON.stringify(keys)}`
+      `Unknown Isolation level, was expecting one of: ${JSON.stringify(
+        humanReadableKeys
+      )}`
     );
   }
   return knownEnum;
@@ -122,3 +121,6 @@ const isolationEnum = {
   SERIALIZABLE: 0x04,
   SNAPSHOT: 0x05,
 };
+const humanReadableKeys = Object.keys(isolationEnum).map((key) =>
+  key.toLowerCase().replace('_', ' ')
+);
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -5,7 +5,7 @@ module.exports = class Transaction_MSSQL extends Transaction {
   begin(conn) {
     debug('%s: begin', this.txid);
     return conn.tx_
-      .begin(this.isolationLevel)
+      .begin(nameToIsolationLevelEnum(this.isolationLevel))
       .then(this._resolver, this._rejecter);
   }
 
@@ -97,3 +97,22 @@ module.exports = class Transaction_MSSQL extends Transaction {
     }
   }
 };
+
+/** MSSQL is annoying this way */
+function nameToIsolationLevelEnum(level) {
+  level = level.toLowerCase();
+  switch (level) {
+    case 'read uncommitted':
+      return 'READ_UNCOMMITTED';
+    case 'read committed':
+      return 'READ_COMMITTED';
+    case 'repeatable read':
+      return 'REPEATABLE_READ';
+    case 'snapshot':
+      return 'SNAPSHOT';
+    case 'serializable':
+      return 'SERIALIZABLE';
+    default:
+      return level;
+  }
+}
--- a/lib/dialects/mssql/transaction.js
+++ b/lib/dialects/mssql/transaction.js
@@ -98,7 +98,6 @@ module.exports = class Transaction_MSSQL extends Transaction {
   }
 };
 
-/** MSSQL is annoying this way */
 function nameToIsolationLevelEnum(level) {
   if (!level) return;
   level = level.toUpperCase().replace(' ', '_');
--- a/lib/dialects/oracledb/transaction.js
+++ b/lib/dialects/oracledb/transaction.js
@@ -4,7 +4,11 @@ const debugTx = require('debug')('knex:tx');
 
 module.exports = class Oracle_Transaction extends Transaction {
   // disable autocommit to allow correct behavior (default is true)
-  begin() {
+  begin(conn) {
+    if (this.isolationLevel) {
+      // Only supports read committed and serializable, there's also a "read only"
+      return this.query(conn, `SET TRANSACTION ${this.isolationLevel}`);
+    }
     return Promise.resolve();
   }
 
--- a/lib/dialects/oracledb/transaction.js
+++ b/lib/dialects/oracledb/transaction.js
@@ -14,7 +14,12 @@ module.exports = class Oracle_Transaction extends Transaction {
           'Oracle only supports read committed and serializable transactions, ignoring the isolation level param'
         );
       } else {
-        return this.query(conn, `SET TRANSACTION ${this.isolationLevel}`);
+        this.client.logger.warn(
+          'Not supporting Oracle transaction isolation right now'
+        );
+        // I tried this, but it didn't work
+        // Doc here: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SET-TRANSACTION.html
+        // return this.query(conn, `SET TRANSACTION ${this.isolationLevel}`);
       }
     }
     return Promise.resolve();
--- a/lib/dialects/oracledb/transaction.js
+++ b/lib/dialects/oracledb/transaction.js
@@ -2,12 +2,20 @@ const Transaction = require('../../execution/transaction');
 const { timeout, KnexTimeoutError } = require('../../util/timeout');
 const debugTx = require('debug')('knex:tx');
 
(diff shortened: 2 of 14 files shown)


types/index.d.ts
```diff
@@ -328,6 +328,10 @@ interface DMLOptions {
   includeTriggerModifications?: boolean;
 }
 
+interface TransactionMethods<T> {
+  setIsolationLevel: (value: string) => T;
```
kibertoad, line 332: Would be better to use an enum with all possible values, also runtime code should throw an error if unknown isolation level is passed.

test/unit/knex.js
```diff
@@ -447,6 +447,13 @@ describe('knex', () => {
       return knex.destroy();
     });
 
+    it('supports transaction isolation', async () => {
```
kibertoad, line 450: Would be better to run it in an integration test to cover all the dbs. See test/integration2 for a reference.
If you have Docker installed, you can use `npm run db:start:no-oracle` to have all the dbs locally.
Also test needs to ensure that isolation level is actually doing something.

lib/knex-builder/make-knex.js
```diff
@@ -141,10 +141,8 @@ function initContext(knexFn) {
         const trx = this.client.transaction(container, config, outerTx);
         return trx;
       } else {
-        return new Promise((resolve, reject) => {
-          const trx = this.client.transaction(resolve, config, outerTx);
-          trx.catch(reject);
```
kibertoad, line 146: I don't see this catch blocked added elsewhere, was its removal intentional?

ayroblu: I don't really get why this Promise is here, when in the other branch it just returns the trx element, so ideally I'd find a way to remove it, based on the tests failing, that might not be possible :/

lib/dialects/postgres/transaction.js
```diff
@@ -0,0 +1,13 @@
+const Transaction = require('../../execution/transaction');
+
+class Transaction_PG extends Transaction {
+  begin(conn) {
+    if (this.isolationLevel) {
+      return this.query(conn, `BEGIN ISOLATION LEVEL ${this.isolationLevel};`);
+    } else {
```
kibertoad, line 7: `else` is redundant here.

lib/dialects/sqlite3/transaction.js
```diff
@@ -0,0 +1,12 @@
+const Transaction = require('../../execution/transaction');
+
+class Transaction_Sqlite extends Transaction {
+  begin(conn) {
+    const setIsolationLevel = this.isolationLevel
```
kibertoad, line 5: Does it have to be different from how it's done in PG, which has no ternary operator and `Promise.resolve`?

lib/execution/transaction.js
```diff
@@ -78,7 +78,10 @@ class Transaction extends EventEmitter {
   }
 
   begin(conn) {
-    return this.query(conn, 'BEGIN;');
+    const setIsolationLevel = this.isolationLevel
+      ? this.query(conn, `SET ISOLATION LEVEL ${this.isolationLevel};`)
```
kibertoad, line 82: Which databases would that work for? I think this will currently affect Oracle and MySQL.

lib/execution/transaction.js
```diff
@@ -93,6 +96,11 @@ class Transaction extends EventEmitter {
     return this.query(conn, `RELEASE SAVEPOINT ${this.txid};`, 1, value);
   }
 
+  setIsolationLevel = (value) => {
+    this.isolationLevel = value;
```
kibertoad, line 100: Would be good to validate value here.

lib/execution/transaction.js
```diff
@@ -180,6 +188,9 @@ class Transaction extends EventEmitter {
           // Directly thrown errors are treated as automatic rollbacks.
           let result;
           try {
+            if (!container) {
```
kibertoad, line 191: Why this change?

lib/knex-builder/make-knex.js
kibertoad: Saying that our transaction logic is messy and complicated would be an... understatement. There are lots of edge cases there.

types/index.d.ts
ayroblu: Well normally there are 4 levels, mssql has 5 (kinda 6), sqlite has 1, kinda 2, from my testing mysql (5.7) says it has 4 but really only works with 3. So I'm not sure exactly the best interface for it

lib/dialects/mssql/transaction.js
```diff
@@ -95,3 +97,27 @@ module.exports = class Transaction_MSSQL extends Transaction {
...
   }
 };
+
+/** MSSQL is annoying this way */
+function nameToIsolationLevelEnum(level) {
+  if (!level) return;
+  level = level.toUpperCase().replace(' ', '_');
+  const knownEnum = isolationEnum[level];
+  if (!knownEnum) {
+    const keys = Object.keys(isolationEnum).map((key) =>
```
kibertoad, line 107: I would suggest calculating this once and just having this as a const.

lib/dialects/sqlite3/transaction.js
```diff
@@ -0,0 +1,17 @@
+const Transaction = require('../../execution/transaction');
+
+class Transaction_Sqlite extends Transaction {
+  begin(conn) {
+    // SQLite doesn't really support isolation levels, it is serializable by
```
kibertoad, line 5: Throwing an error is fine.

types/index.d.ts
kibertoad: Supporting all 5 (or) 6 on validation level is fine, DB will throw an error if it doesn't support something.

ayroblu:
Urg I don't get why it's still failing, it passes fine locally :/
image
And I get unrelated failures when I run the whole thing
image

ayroblu:
Okay, it's timing out on `ALTER DATABASE knex_test SET ALLOW_SNAPSHOT_ISOLATION ON` in CI hmmm

ayroblu:
Well that should cover the main parts, let me know if you'd like to change anything, the original issue suggests a different API and I'm not really familiar enough with all the chaining options in terms of what's best.
Probably I guess you might want some more tests I guess otherwise? I had to change the docker compose file cause I couldn't get the alter table to pass so not sure I can do much about that.

lib/execution/transaction.js
```diff
@@ -78,7 +94,13 @@ class Transaction extends EventEmitter {
   }
 
   begin(conn) {
-    return this.query(conn, 'BEGIN;');
+    const setIsolationLevel = this.isolationLevel
```
kibertoad, line 97: Might be  a good idea to rename this variable so that it doesn't have same name as the class method.

lib/execution/transaction.js
```diff
@@ -93,6 +115,18 @@ class Transaction extends EventEmitter {
     return this.query(conn, `RELEASE SAVEPOINT ${this.txid};`, 1, value);
   }
 
+  setIsolationLevel(isolationLevel) {
```
kibertoad, line 118: `transaction()` method already has optional `config` parameter, why not just add isolation level there?

scripts/docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
       - bash
       - -c
       # https://docs.microsoft.com/en-us/sql/relational-databases/logs/control-transaction-durability?view=sql-server-ver15#bkmk_DbControl
-      - 'until /opt/mssql-tools/bin/sqlcmd -S mssql -U sa -P S0meVeryHardPassword -d master -Q "CREATE DATABASE knex_test; ALTER DATABASE knex_test SET DELAYED_DURABILITY = FORCED"; do sleep 5; done'
+      - 'until /opt/mssql-tools/bin/sqlcmd -S mssql -U sa -P S0meVeryHardPassword -d master -Q "CREATE DATABASE knex_test; ALTER DATABASE knex_test SET ALLOW_SNAPSHOT_ISOLATION ON; ALTER DATABASE knex_test SET DELAYED_DURABILITY = FORCED"; do sleep 5; done'
```
kibertoad, line 23: This can't be applied in runtime using knex.raw?

test/integration2/transaction/set-isolation-level.spec.js
```diff
@@ -0,0 +1,63 @@
...
+            return;
+          }
+          const trx = await knex
+            .transaction()
+            .setIsolationLevel('read committed');
+          const result1 = await trx(tableName).select();
+          await knex(tableName).insert({ id: 1, value: 1 });
+          const result2 = await trx(tableName).select();
+          await trx.commit();
+          expect(result1).to.not.equal(result2);
```
kibertoad, line 42: Wouldn't this assertion pass even if you didn't use transaction isolation? One of the selects happens before the insert, so it is obvious that results won't match.

test/integration2/transaction/set-isolation-level.spec.js
```diff
@@ -0,0 +1,63 @@
...
+        it('Expect to avoid read skew when repeatable read (snapshot isolation)', async () => {
+          if (isSQLite(knex)) {
+            return;
+          }
+          // NOTE: for mssql, it requires an alter database call that happens in docker-compose
+          const isolationLevel = isMssql(knex) ? 'snapshot' : 'repeatable read';
+          const trx = await knex
+            .transaction()
+            .setIsolationLevel(isolationLevel);
+          const result1 = await trx(tableName).select().timeout(500);
```
kibertoad, line 54: why do we need timeout?

types/index.d.ts
```diff
@@ -349,11 +355,11 @@ interface Knex<TRecord extends {} = any, TResult = unknown[]>
   transaction(
     transactionScope?: null,
     config?: any
-  ): Promise<Knex.Transaction>;
+  ): Promise<Knex.Transaction> & TransactionMethods<Promise<Knex.Transaction>>;
```
kibertoad, line 358: Shouldn't this also be added to TransactionProvider?

kibertoad:
Thank you for the contribution! Could you also please update the documentation? https://github.com/knex/documentation

lib/dialects/sqlite3/transaction.js
```diff
@@ -0,0 +1,18 @@
...
+    // default and so we override it to ignore isolation level.
+    // There is a `PRAGMA read_uncommitted = true;`, but that's probably not
+    // what the user wants
+    if (this.isolationLevel) {
+      throw new Error(
```
kibertoad, line 10: Having slept on this, I think warning that the isolation level will be ignored would be better, since SQLite is often used as a stand-in for a real DB, and having "if (test_env)" conditions in code would be a huge hassle.
I think transaction has reference to the client, so it should be possible to do `this.client.logger.warn`.

scripts/docker-compose.yml
ayroblu: I tried this, but it threw an error, I guess there's side effects elsewhere
`ALTER DATABASE knex_test SET SNAPSHOT ON WITH ROLLBACK IMMEDIATE`
https://stackoverflow.com/a/1253452/6725059

test/integration2/transaction/set-isolation-level.spec.js
ayroblu: Yes that's the point, even though you're using a transaction, you aren't getting the isolation that you might expect when running queries, which is the same as not having a transaction at all.

test/integration2/transaction/set-isolation-level.spec.js
ayroblu: You don't, I can remove that, it was just helped my testing

types/index.d.ts
ayroblu: Uhhh I haven't touched the TransactionProvider, so I'm not sure if it's the same code path, but sure, I could add it there

lib/execution/transaction.js
ayroblu: Ummm, I could? It's basically the same, but I've never seen any documentation for the config param and I thought the chaining style was more idiomatic 🤷

kibertoad: Yeah, I'll add types and documentation for the config param. And currently there is no overload for .transaxtion with just a config param, I think.

ayroblu:
@kibertoad Can you help me with oracle? I'm not sure I understand the error, or how oracle really works, maybe worth skipping?
`Error: Transaction query already complete, run with DEBUG=knex:tx for more info`

Docs for isolation with oracle are here: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SET-TRANSACTION.html#GUID-F11E1E30-5871-48D1-8266-F80A1DF126A1

lib/knex-builder/make-knex.js
```diff
@@ -123,6 +123,12 @@ function initContext(knexFn) {
     // If container is not provided, returns a promise with a transaction that is resolved
     // when transaction is ready to be used.
     transaction(container, _config) {
+      // Overload support of `transaction(config)`
+      if (!_config && typeof container === 'object') {
```
kibertoad, line 127: It would be safer to use method `isObject` from `lib/util/is.js`, because null is also an object.

lib/dialects/oracledb/transaction.js
```diff
@@ -2,9 +2,26 @@ const Transaction = require('../../execution/transaction');
...
+  begin(conn) {
+    if (this.isolationLevel) {
+      // There's also a "read only", but that's not really an "isolationLevel"
+      if (!supportedIsolationLevels.includes(this.isolationLevel)) {
+        this.client.logger.warn(
+          'Oracle only supports read committed and serializable transactions, ignoring the isolation level param'
+        );
+      } else {
+        this.client.logger.warn(
+          'Not supporting Oracle transaction isolation right now'
```
kibertoad, line 18:
```suggestion
          'Transaction isolation is not currently supported for Oracle'
```

lib/dialects/oracledb/transaction.js
```diff
@@ -2,9 +2,26 @@ const Transaction = require('../../execution/transaction');
...
+        );
+        // I tried this, but it didn't work
```
kibertoad, line 20: Don't have Oracle instance locally, and debugging with Travis is a pain, so yeah, it's fine to skip Oracle for now.

kibertoad (approved):

clemens-smartparking:
It's not on npm yet, right? or do I miss something? what is the release process?

kibertoad:
@clemens-smartparking Basically the last two things left before 0.95.0 goes full on live: 
1) @tylerjwatson finishes error reporting improvement he needed to get feature parity with his branch;
2) @tylerjwatson confirms that preview version of a new mssql driver behaves correctly.

In case you don't use mssql or would like to contribute to testing the (considerably improved) mssql driver, you can use `0.95.0-next1` already, it is 99.9% what `0.95.0` final will be.

tylerjwatson:
Many apologies for the delay, we've had production problems that have sopped up all my free time.  I am working through getting this done, but it may be a little while yet.

kibertoad:
@tylerjwatson No worries! You are doing a big favour to the project, and we are grateful that you find it possible to contribute altogether. We'd still like to release late February; if you don't believe it would be feasible for you to finish before that, we might have to just bite the bullet and release what we have, and rely on knex community for testing it.

clemens-smartparking:
Thanks for the clarification and thanks for your work! I planned to test for sqlite and pg. However, there seem to be some TS type problems with objection.js when using 0.95.0-next1. Any easy fix for that?

kibertoad:
@clemens-smartparking Unfortunately, that is not that straightforward. First of all, Objection TS types need to be adjusted, as seen in https://github.com/Vincit/objection.js/pull/1992. Second, looks like Objection is importing some of knex internal files directly, so that will need to be adjusted as well. I'll spend some time this week to fix incompatibilities on Objection side, hopefully some kind of `next` release can be done for Objection as well.

PR #4186, opened 2020-12-30
fallback to json for sqlite3 when using jsonb

7opf:


kibertoad:
Thanks! Could you please add a test for this?
See `test/integration2/schema/primary-keys.spec.js` for a good example.

test/integration2/schema/jsonb.spec.js
```diff
@@ -0,0 +1,57 @@
+const { expect } = require('chai');
+const { getAllDbs, getKnexForDb } = require('../util/knex-instance-provider');
+
+describe('Schema', () => {
+  describe('json columns', () => {
+    // no support for json in mssql, omit test
+    getAllDbs()
+      .filter((db) => db !== 'mssql')
```
kibertoad, line 8: Does it work on Oracle or that also has to be skipped?

7opf:
> Thanks! Could you please add a test for this?
> See `test/integration2/schema/primary-keys.spec.js` for a good example.

@kibertoad - I've added a test which will check `PRAGMA table_info()` for sqlite and `information_schema.columns` for the other sql drivers (except mssql which does have JSON data type). btw - this might be a separate issue, but attempting to create a jsonb column with mssql backend currently results in an error from the backend, not sure if this is desired behaviour for knex?

kibertoad:
@7opf I would say that it's preferable to fail-fast there rather than trick user into thinking that JSON is supported, when it is not.

test/integration2/schema/jsonb.spec.js
7opf: Seems to be the same situation as mssql - it's not a supported data type. Attempting to create such a schema gives `Error: create table "table_with_jsonb_col1" ("jsonb_col1" undefined) - ORA-00902: invalid datatype`

7opf: the reason I hadn't put it in the test, was because it wasn't part of `getAllDbs`, I assumed there was a reason for it not being there so I didn't explicitly add it.

kibertoad: Let's skip it too then.

7opf: I'll add it to the filter

7opf:
> @7opf I would say that it's preferable to fail-fast there rather than trick user into thinking that JSON is supported, when it is not.

The databases do "support" JSON through e.g. as VARCHAR or NVARCHAR in mssql. This behaviour is similar to what I saw for old versions of postgres in knex below:   https://github.com/knex/knex/blob/a31f0400e61d795dc7dbce92ef1f71d0d45a0926/lib/dialects/postgres/schema/columncompiler.js#L119-L123

Knex could "fallback" to the JSON-equivalent data type in mssql and oracledb, rather than failing (silently or not).

kibertoad:
Oracle is not included in default list, but can still come from an ENV override (which is the case for Travis), so it's better to skip it explicitly.
I'm not sure `TEXT` can be considered an equivalent to JSON, you can do fairly limited things with it. Either way, changing behaviour for it is definitely out of scope for this PR.

kibertoad (approved):

kibertoad:
Thanks a lot!

kibertoad:
Released in 0.95.0

PR #4189, opened 2020-12-31
Fix SQLite foreign key constraints when altering a table

### LINKED ISSUE #4155: Foreign keys prevent table altering migrations on SQLite when foreign keys pragma is on
# Environment

Knex version: 0.21.12
Database + version: sqlite3 3.27.2
OS: Linux

# Bug

When trying to alter an SQLite table that is referenced by a foreign key, and when the foreign keys pragma is on, migrations fail. This seems to be due to how SQLite limitations are worked around – renaming or dropping a column is implemented as dropping the whole table and making a replacement.

We are setting the pragma as suggested in https://github.com/knex/knex/issues/453#issuecomment-54160324. I think the implementations of dropping and renaming a column should start by checking whether the pragma is on, and in such case, disable it and restore after the operation is complete. (Only the particular operation should be “wrapped”; if the whole migration was, then even `DROP TABLE` would run despite references)

A workaround might be to skip the pragma for migrations. However, I don’t think it is possible to use different settings for migrations in the knexfile?

2. Error message

```
Error: DROP TABLE "table_being_altered" - SQLITE_CONSTRAINT: FOREIGN KEY constraint failed
```

3. Reduced test code

See https://runkit.com/embed/05t8t5jgr1j1

```javascript
require('sqlite3');
const Knex = require('knex');

const knexSqlite = Knex({
  client: 'sqlite',
  connection: ':memory:',
  pool: {
    afterCreate: (conn, cb) => conn.run("PRAGMA foreign_keys = ON", cb)
  }
});

(async function run() {

  // first migration
  await knexSqlite.schema.createTable('cars', (t) => {
    t.string('id').primary();
    t.string('comment');
  });
  await knexSqlite.schema.createTable('wheels', (t) => {
    t.string('id').primary();
t.integer('car_id').unsigned().notNullable().references('id').inTable('cars');
  });

  // data population
  await knexSqlite('cars').insert({ id: "a", comment: 'Lorem ipsum' });
  await knexSqlite('wheels').insert([{ id: "a1", car_id: "a" }]);
  
  // second migration
  await knexSqlite.schema.table('cars', (t) => {
    t.dropColumn('comment');
  });

  console.log('test table data:', await knexSqlite('cars'));
})();
```

nickrum:
Some notes/questions:

- Is it safe to execute raw queries directly on the client?
- What was the reasoning behind doing `disableProcessing()` / `enableProcessing()` in `getTableSql()`? It seems like this does not work when using the client directly instead of through a transaction.
- Unless I am missing something, with this change `reinsertMapped()` should be treated like a critical section, so another invocation of the function can't interfere with the procedure of querying, disabling and reenabling foreign key constraint enforcement.
- When using indices, the manual foreign key check will likely fail as they are not recreated for the new table. (According to the [SQLite Docs for ALTER TABLE](https://sqlite.org/lang_altertable.html) this also applies to triggers and views, but I'm not sure how relevant that is for Knex)
- With this change would it be possible to enable foreign key constraint enforcement for SQLite by default? (Maybe with an option to disable it)

Fixes #4155

### DIFF
--- a/lib/dialects/sqlite3/schema/ddl.js
+++ b/lib/dialects/sqlite3/schema/ddl.js
@@ -12,6 +12,9 @@ const {
   dropOriginal,
   renameTable,
   getTableSql,
+  isForeignCheckEnabled,
+  setForeignCheck,
+  executeForeignCheck,
 } = require('./internal/sqlite-ddl-operations');
 const { parseCreateTable, parseCreateIndex } = require('./internal/parser');
 const {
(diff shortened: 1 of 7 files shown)


kibertoad:
> Is it safe to execute raw queries directly on the client?

It's preferable to avoid doing that (although we have cases of this, but we intend to refactor that out). The reason for that - knex aims to be a querybuilder, and certain projects use it that way, simply generating SQL statements that they execute on their own. Ideally that should be preserved.

> What was the reasoning behind doing disableProcessing?

See #3033

> When using indices, the manual foreign key check will likely fail as they are not recreated for the new table

Why not? That sounds confusing for the user.

> With this change would it be possible to enable foreign key constraint enforcement for SQLite by default?

Why should it be default? Isn't it something that user should opt-in on their own? I'm open to exposing this config as an option param, though.

kibertoad:
@nickrum A heads-up - I'm currently reworking foreign key creation to output proper SQL, you should be able to reuse same solution after it is done.

kibertoad:
But in order to unblock you for now, since changes might take some time, feel free to execute changes directly on client for now if that makes things easier for you.

kibertoad:
@nickrum After #4194 is merged, you can use the mechanism used there as a reference.

nickrum:
@kibertoad Thank you for the info and the heads-up. In that case I will wait until that PR is merged as I don't think it would be possible to output proper SQL for the foreign key check only without also refactoring the whole DDL generation.

I did find a way to essentially do what this PR currently does with generated SQL statements, but there is one catch: As, to my knowledge at least, it is not possible to enable/disable the foreign key check based on a previous query, the SQL statements have to be generated based on a query of "`PRAGMA foreign_keys` at "compile-time" or based on a config option.

> See #3033

I see, thanks. Is there a way to disable processing without a transaction? The function doesn't seem to exist on the client directly.

> Why not? That sounds confusing for the user.

Because I didn't find the time to implement this yet. It is a bit more complicated as it requires modifying the `CREATE INDEX` query used to create the index if a column has been renamed or dropped.

> Why should it be default? Isn't it something that user should opt-in on their own? I'm open to exposing this config as an option param, though.

Mainly because of https://github.com/knex/knex/issues/453#issuecomment-54160324. I am not sure if users expect the check to be turned on by default because of the way other database vendors do it or the opposite because it is the default in SQLite. In any case, a config option sounds like a good idea.

kibertoad:
@nickrum PR was just merged!

> the SQL statements have to be generated based on a query of "PRAGMA foreign_keys at "compile-time

That's not a problem. Check how that PR is implemented, there is an async function involved that generates query SQL based on results of previous queries, same approach should work for you.

> Is there a way to disable processing without a transaction?

Where do you need it? You'll have an instance of transaction inside ddl.

kibertoad:
@nickrum Awesome! Do you need help writing a test?

nickrum:
> Where do you need it? You'll have an instance of transaction inside ddl.

Inside `getTableSql()`. I had to remove the transaction because enabling/disabling the foreign keys check inside a transaction is not allowed.

> Do you need help writing a test?

I would appreciate any help regarding tests. I still have to port the foreign keys check to the new `generateReinsertCommands()` though. Currently it only works with the old method.

Also, one of the tests is failing because `renameColumn()` doesn't rename references inside other tables.

kibertoad:
@nickrum New style of integration tests are actually quite easy to run. Since you are only interested in SQLite, what you can do is mark your test suite with `describe.only` and then run `npm run test:sqlite`. Best place for new tests would be here: test/integration2/schema/foreign-keys.spec.js
You can adjust existing ones slightly to replicate the case that you need.

kibertoad:
> . I still have to port the foreign keys check to the new generateReinsertCommands() though. Currently it only works with the old method.

In order for the new method to work, make sure to update `lib/dialects/sqlite3/schema/sqlite-tablecompiler.js` to use `statementsProducer` instead of `output`, then array of string commands will be executed automatically. See https://github.com/knex/knex/blob/92d8f49724493d25cad8224fe9831f51fbb7f6c9/lib/dialects/sqlite3/schema/sqlite-tablecompiler.js#L160 as an example.

nickrum:
@kibertoad Thanks, I think I will be able to figure this out.

lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -58,50 +58,52 @@ class SQLite3_DDL {
   }
 
   getTableSql() {
-    this.trx.disableProcessing();
-    return this.trx.raw(getTableSql(this.tableName())).then((result) => {
-      this.trx.enableProcessing();
-      return result;
-    });
+    return this.client.transaction(
```
kibertoad, line 61: Wouldn't this result in starting one more transaction on top of already running one?

kibertoad: ah, I see you just pushed existing transaction creation logic slightly deeper.

lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -477,14 +442,47 @@ class SQLite3_DDL {
...
+            );
+
+            if (foreignKeysViolations.length > 0) {
+              throw new Error('FOREIGN KEY constraint failed');
+            }
+          }
+        },
+        { connection: this.connection }
+      );
+    } finally {
```
kibertoad, line 479: nice touch executing this in `finally` block.

kibertoad:
@nickrum If you'll get stuck with writing tests, please let me know, would be happy to help.

kibertoad:
Friendly ping :). Happy to pick up any pieces of remaining work if you need help with anything.

nickrum:
@kibertoad sorry for the delay :) The beginning of a year is always a little busy. I also had a few issues implementing the foreign keys check for generated SQL statements, but I think I got this now. If you could pick up writing some tests, that would help a lot.

lib/dialects/sqlite3/schema/ddl.js
nickrum: Exactly, I had to do this because enabling/disabling the foreign keys check inside a transaction does not work. But the transaction is still needed to rollback in case the check fails.

lib/dialects/sqlite3/schema/ddl.js
nickrum: Just to be on the safe side! Unfortunately, this does not work with generated SQL statements.

nickrum:
@kibertoad I finally found the time to implement my idea regarding foreign keys check with generated SQL statements.

Unfortunately it is not exactly intuitive. I had to create temporary tables and triggers to dynamically rollback the changes if a foreign key error occurs.

There are two independent checks because the manual foreign keys check can report violations in two different ways. Either the check returns a list of violations for every row that violates a foreign key constraint, or the check command simply fails if a whole column referenced by some other column can not be found.

The change to `queryArray()` is needed to execute all queries before throwing an error to make sure the temporary tables and triggers are removed and the foreign keys check is reenabled (this is essentially what the `finally` does in `reinsertMapped()`).

kibertoad:
@nickrum Does the new logic only work when executed on client directly? Is it not possible to generate SQL for all of the operations that users could execute themselves, so that they could use knex purely as a query builder?

nickrum:
It should work without the client as long as every generated SQL is executed regardless of any errors. This is definitely not ideal but, to my knowledge at least, there is no easy way around this as the foreign key constraint enforcement definitely has to be reenabled after the transaction.

I can only think of two alternatives. One would be to basically execute the whole procedure in `reinsertMapped()` at build time of the query and based on that throw an error at build time. The other would be to manually validate the constraints at build time of the query. Both approaches have the disadvantage that they depend on the state of the database at build time which would make the usage of the generated queries less flexible.

nickrum:
@kibertoad Let me know which approach you prefer or if you can think of another way to go about this.

kibertoad:
@nickrum I would say that with the query builder approach it is not possible to have a completely safe solution, so I would argue that if user goes for query generation and executes it manually, they should be responsible for wrapping it into transaction and rolling back properly should any part of it (such as executing the foreign check) fails. When we are executing it directly (using `runner.js` -> `queryArray()`), we should be the ones ensuring that there is an atomic transaction for the whole array. Not sure why we might need any additional safety mechanisms beyond that.

nickrum:
@kibertoad Good point. What about the problem that `PRAGMA foreign_keys` does not work inside a transaction? A user who goes for query generation has to somehow know that those must not be executed inside a transaction. The same thing also applies to us when we are executing the queries directly in `queryArray()`.

kibertoad:
@nickrum Yeah, that is definitely a problem. I don't think that a good solution for that exists; considering that we don't know when and how user will execute generated SQL, we can't really provide him with an easy way to rollback their changes, so I would say that we should simply throw an error and let the user handle situation in this case.

kibertoad:
@nickrum Do you think that there is anything left, other than fixing tests?

nickrum:
@kibertoad Sounds good to me. Regarding executing the queries directly inside `queryArray()`, I am not sure how to start a transaction inside that function and how to exclude the `PRAGMA` statements from the transaction in a clean way. Any ideas?

After finding a solution for that, fixing the tests and writing new tests to verify the added functionality, this PR should be good to go.

kibertoad:
@nickrum If you believe that it is important to explicitly delineate those transactions, I would suggest adjusting signature of `generateDdlCommands` and `statementsProducer` to return an array of arrays, where each array is a separate transaction. Then it should be possible to wrap them properly in execution.

kibertoad:
@nickrum Does the suggested design make sense? Would you need any help with implementing it?
After thinking some more about it, I realize that in-place execution is likely to be problematic since by default Migrator executes changes in a single transaction, and modifying logic to ensure that we commit and start a new transaction at the right moment is going to be rather complicated with current design, and I wouldn't do it within this PR. How feasible is it to ensure that correct SQL statements are generated for `generateDdlCommands` for manual execution, where user can handle transactions correctly, and make in-place execution just using Migrator "more or less work" without proper transactionality, that would be implemented later separately?

nickrum:
> If you believe that it is important to explicitly delineate those transactions

I do think so because, while we can't tell how a user executes the generated SQL statements, if one of the altering functions is executed directly on the client, without a transaction it is not guaranteed that the whole procedure is reverted and we might leave the database in an inconsistent state.

> Does the suggested design make sense? Would you need any help with implementing it?

It does make sense and the implementation of the SQLite part should be straightforward.

> I realize that in-place execution is likely to be problematic since by default Migrator executes changes in a single transaction

Not sure if I understand this correctly as my overall knowledge about Knex is rather limited, but currently `SQLite3_DDL` already starts a transaction for every executed DDL function which is executed in the `processResponse()` call of `Runner.query()`. So shouldn't it be possible to come up with a version of `Runner.query()` that executes the query inside a transaction and that does work with Migrator without having to ensure that we commit and start a new transaction at the right moment?

> How feasible is it to ensure that correct SQL statements are generated for `generateDdlCommands` for manual execution, where user can handle transactions correctly, and make in-place execution just using Migrator "more or less work" without proper transactionality, that would be implemented later separately?

Without a transaction there would be no way to revert the changes in case of a foreign key constraint violation. If this can be tolerated, then it is feasible.

kibertoad:
> Not sure if I understand this correctly as my overall knowledge about Knex is rather limited, but currently SQLite3_DDL already starts a transaction for every executed DDL function which is executed in the processResponse() call of Runner.query(). So shouldn't it be possible to come up with a version of Runner.query() that executes the query inside a transaction and that does work with Migrator without having to ensure that we commit and start a new transaction at the right moment?

Fair point, let's try for that.

kibertoad:
@nickrum Any chance you could create a preliminary PR changing signature of `generateDdlCommands` to return an array, so that we could implement the rest of the changes without introducing breaking changes within 0.95.0?

nickrum:
@kibertoad Sure! I feel like it is not clear which queries should be executed inside a transaction and which queries must not if `generateDdlCommands` returns an array of arrays, though. What do you think of returning an object like:

```javascript
{
  pre: [],
  sql: [],
  post: []
}
```

This way, if foreign key constraint enforcement is disabled, the user could simply ignore the `pre` and `post` properties and use `.generateDdlCommands().sql` without having to worry about at which position in the array of arrays the actual queries to alter the schema are.

kibertoad:
@nickrum Sounds like a great idea!

OlivierCavadenti:
What is the status of this PR ? @nickrum

nickrum:
@kibertoad @OlivierCavadenti Sorry this took so long! I made several attempts throughout the year at finishing this PR but there were a lot of show-stoppers preventing me from doing so.

Initially I thought I could use the SQLite's `PRAGMA defer_foreign_keys` to defer the foreign keys check until the alter table commands have been executed. This wasn't possible due to several issues. That's why I had to add an additional `check` property to the object returned by `generateDdlCommands` to execute the check explicitly.

lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -43,365 +46,351 @@ class SQLite3_DDL {
...
+      },
+      { connection: this.connection }
+    );
   }
 
-  renameTable() {
-    return this.trx.raw(renameTable(this.alteredName, this.tableName()));
+  async isForeignCheckEnabled() {
+    const result = await this.client
+      .raw(isForeignCheckEnabled())
```
kibertoad, line 69: can we rename one of the methods to make this code less confusing?

nickrum: Sure! I kept the names the same because the functions `createNewTable`, `copyData`, `dropOriginal`, `renameTable` and `getTableSql`, which existed before this PR, are also called exactly the same as their corresponding methods in the `SQLite3_DDL` class. Should I rename all of them?

kibertoad: all good then, out-of-scope for this pr

kibertoad (approved):
LGTM

nickrum:
In just pushed a small commit that wraps the `statementsProducer` transaction and the post query in a try-finally-block to align with the way the `alter()` function is implemented.

OlivierCavadenti:
>Sorry this took so long! I made several attempts throughout the year at finishing this PR but there were a lot of show-stoppers preventing me from doing so.

No problem, if I ask, it's to know if we can finish the work or not, I understand that everyone is busy !
Anyway, that's for this hard and amazing work, I can merge.

rijkvanzanten:
@kibertoad hate to be that guy, but do you think we could sneak in a 0.95.15 before the big 1.0 release? 😬 Between this and https://github.com/knex/knex/pull/4923, a couple mission critical bugfixes for SQLite use have made it in to the project, but aren't available for use yet. Totally understand if v1 is the exclusive focus rn tho! ✌🏻

kibertoad:
@rijkvanzanten Sure. can you give a laundry list of prs you want in it?

kibertoad:
Released in 1.0.1.

@rijkvanzanten Do you still need 0.95 backport, or 1.x is sufficient?

AlphaMoonbaseBerlin:
@kibertoad 
Just FIY: This issue brings in some issues with [Directus](https://github.com/directus/directus), which uses knex as its ORM. 
Because of the switch to sqlite/vscode, they did not update knex to version 1. Some Discussion [here](https://github.com/directus/directus/issues/10842)

PR #4194, opened 2021-01-01
Return complete list of DDL commands for creating foreign keys in SQLite

### LINKED ISSUE #3351: Allow creating foreign keys in sqlite

kibertoad:
fixes #3351

### DIFF
--- a/lib/dialects/sqlite3/schema/internal/sqlite-ddl-operations.js
+++ b/lib/dialects/sqlite3/schema/internal/sqlite-ddl-operations.js
@@ -1,12 +1,16 @@
 const identity = require('lodash/identity');
 const chunk = require('lodash/chunk');
 
-async function insertChunked(trx, chunkSize, target, iterator, result) {
+function insertChunked(trx, chunkSize, target, iterator, existingData) {
+  const result = [];
   iterator = iterator || identity;
-  const chunked = chunk(result, chunkSize);
+  const chunked = chunk(existingData, chunkSize);
   for (const batch of chunked) {
-    await trx.queryBuilder().table(target).insert(batch.map(iterator));
+    result.push(
+      trx.queryBuilder().table(target).insert(batch.map(iterator)).toQuery()
+    );
   }
+  return result;
 }
 
 function createTempTable(createTable, tablename, alteredName) {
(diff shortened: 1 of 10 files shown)


kibertoad:
@B4nan Could you please review? If this works for you, I could revise all other SQLite commands to follow the same pattern.

lib/dialects/sqlite3/schema/internal/sqlite-ddl-operations.js
```diff
@@ -1,21 +1,34 @@
...
+    );
   }
+  return result;
 }
 
 function createTempTable(createTable, tablename, alteredName) {
   return createTable.sql.replace(tablename, alteredName);
 }
 
 async function copyData(trx, tableName, alteredName) {
```
kibertoad, line 20: `copyData` and `reinsertData` will be removed in the end, we'll use `copyAllData` for everything, as it's significantly better for large amount of data.

test/integration2/schema/foreign-keys.spec.js
```diff
@@ -37,14 +40,71 @@ describe('Schema', () => {
...
+              'foreign_keys_table_one',
+              (table) => {
+                table
+                  .foreign('fkey_three')
+                  .references('foreign_keys_table_three.id')
+                  .withKeyName('fk_fkey_threeee');
+              }
+            );
+            const queries = await builder.generateDdlCommands();
+            expect(queries).to.eql([
```
kibertoad, line 65: @B4nan This is the format that you would be getting. Would that work?

nickrum:
Looks very promising. It is unfortunate though that it is not possible to return the generated commands directly from `toSQL()`. @kibertoad Have you considered changing that function to return a Promise? (Although that would be a rather heavy breaking change)

kibertoad:
Yes, exactly. Heavily breaking change and one not needed for the majority of cases.

kibertoad:
Hence the new metod generateDdlCommands.

B4nan:
Finally had some time to review this, it looks good but it there are some issues:

- sometimes the meta queries are there (like `pragma ...`), which I am not interested in
- other drivers need to be probably adjusted too, like in mysql the `renameColumn` now prints `show fields from `author` where field = ?;` instead of the actual rename query. also looking at the implementation I would very much like just a rename column query, while it tries to be too smart and fix FKs and indexes too (which I have already handled myself just fine, and wont to have control over those) - could it be at least configurable?
- the TS typings are wrong, it says there will be array of strings but its actually array of objects with `sql` and `bindings` keys
- what is the exact purpose of the `pre` and `post` arrays? so far i never seen anything filled in there. would be great if those could be used for the "too smart" queries i mentioned about the mysql's rename column method, so i can just ignore them this way

PR #4200, opened 2021-01-03
Use SQLite rename column instead of ddl helper

nickrum:
Since SQLite version 3.25.0 renaming columns is supported natively.

Knex depends on `sqlite3@5.0.0` which bundles SQLite version 3.32.3, so this should be safe unless someone manually builds `sqlite3` with an older version.


kibertoad (approved):

kibertoad:
This is a very welcome change, thank you!

PR #4202, opened 2021-01-04
Sqlite drop composite foreign keys

schmetzyannick:
Closes https://github.com/knex/knex/issues/4201

### DIFF
--- a/lib/dialects/sqlite3/schema/ddl.js
+++ b/lib/dialects/sqlite3/schema/ddl.js
@@ -308,10 +308,10 @@ class SQLite3_DDL {
               if (defLine.includes(indexName)) return false;
               return true;
             } else {
-              const matched = defLine.match(/\(`(\S+)`\)/);
-              const columnName = matched[1];
+              const matched = defLine.match(/\(`([^)]+)`\)/);
+              const columnNames = matched[1].split(', ');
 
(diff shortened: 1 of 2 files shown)


lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -308,10 +308,17 @@ class SQLite3_DDL {
...
             } else {
-              const matched = defLine.match(/\(`(\S+)`\)/);
-              const columnName = matched[1];
-
-              return columns.includes(columnName) === false;
+              const matched = defLine.match(/\(`([^)]+)`\)/);
+              const columnNames = matched[1].split(', ');
+
+              let unknownColumnIncluded = false;
+              for (let i = 0; i < columnNames.length; i++) {
```
kibertoad, line 315: Wouldn't implementation based on `some` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) be simpler?

schmetzyannick: You are right that would be simpler. But when I look at the code snippet I think the check should be performed  with Array.every, because all columns of the FK are relevant, so my fault.

kibertoad (approved):

kibertoad:
Thank you!

PR #4207, opened 2021-01-05
Add pg-mem to documentation

oguimbal:
Just adds [pg-mem](https://github.com/oguimbal/pg-mem) to documentation, which now supports Knex.

FYI, `pg-mem`is a small full-js library emulating an in-memory Postgres instance, which is bindable to virtually every library aiming postgres (see [supported bindings](https://github.com/oguimbal/pg-mem#-libraries-adapters) here)

### DIFF
--- a/ECOSYSTEM.md
+++ b/ECOSYSTEM.md
@@ -10,4 +10,4 @@ Plugins and tools maintained by the knex team are listed under [Core](#core) whi
 #### [Community](#community)
 
 - [`knex-paginate`](https://github.com/felixmosh/knex-paginate) Extension of the query builder with `paginate` method that helps with pagination tasks.
-- [`pg-mem`](https://github.com/oguimbal/pg-mem) emulates an in-memory postgres instance bound to Knex in order to write blazingly fast tests [(details here)](https://github.com/oguimbal/pg-mem/wiki/Libraries-adapters#-knex)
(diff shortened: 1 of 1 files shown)


kibertoad (approved):

kibertoad:
Thank you! Good luck with your project. May I suggest to add an npm badge to your repo?

oguimbal:
Thanks :)

Right, that's a sensible suggestion.

elhigu:
have you tried if knex's test suite's integration tests passes with this library?

kibertoad:
@elhigu Considering that amount of ground that knex covers, I would be extremely surprised, but that'd be an interesting experiment. I'll give it a shot tomorrow.

oguimbal:
@elhigu @kibertoad At first sight, that will not work.

The first thing that make me think so is that Knex tests seems to be spanning things such as isolation levels, and concurrency.

Isolation levels are not implemented with pg-mem... Given that it is intended for testing, its concurrency model is very simple. A pg-mem database is an immutable map under the hood. Period.. So you have transactions, but no isolation levels, and writes on 
 transactions cannot be concurrently committed (and by extension, there is no such thing as locking for writing mechanisms).

nb: If you play with it, be aware that pg-mem declares very few functions, so if your test uses a function (for instance, knex runs a `select version();` query when booting), you'll have to declare them via `db.public.declareFunction(...)`.

PR #4225, opened 2021-01-12
Fix SQLite foreign on delete when altering a table

bittersweetshimmer:
It's just a typo fix for #4162.

### DIFF
--- a/test/integration2/schema/foreign-keys.spec.js
+++ b/test/integration2/schema/foreign-keys.spec.js
@@ -87,6 +87,10 @@ describe('Schema', () => {
           });
 
           it('generates correct SQL for the new foreign key operation with an on delete clause', async () => {
+            if (!isSQLite(knex)) {
+              return;
+            }
+
             const builder = knex.schema.alterTable(
               'foreign_keys_table_one',
               (table) => {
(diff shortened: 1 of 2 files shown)


kibertoad:
@bittersweetshimmer Thanks! Could you also add test that was clearly missing for this case? Probably you could easily make one out of existing ones in test/integration2/schema/foreign-keys.spec.js

test/integration2/schema/foreign-keys.spec.js
```diff
@@ -86,6 +86,58 @@ describe('Schema', () => {
...
+                  .foreign('fkey_three')
+                  .references('foreign_keys_table_three.id')
+                  .withKeyName('fk_fkey_threeee')
+                  .onDelete('CASCADE');
+              }
+            );
+
+            const queries = await builder.generateDdlCommands();
+
+            if (isSQLite(knex)) {
```
kibertoad, line 103: If this test only makes sense for SQLite, can you add
`          if (!isSQLite(knex)) {
            return;
          }
`

to it?

bittersweetshimmer: On it.

kibertoad:
Thank you!

kibertoad (approved):

kibertoad:
Released in 0.95.0

PR #4235, opened 2021-01-16
fix mssql remove unused pool config option

dhensby:
The `evictionRunIntervalMillis` config option for the mssql pool hasn't been used since v5 of the library.

In the upcoming v7 release this option causes an error as the pool (tarnjs) validates the options provided.


kibertoad:
@dhensby Since that breaks compatibility with mssql 4 (which we are OK not to support in our next release which will be semver major), could you please adjust `lib/dialects/mssql/index.js` to throw an error on mssql < 5 instead of a current workaround?
Also `UPGRADING.md` will need a note that only mssql 5+ is now supported.

lib/dialects/mssql/index.js
```diff
@@ -31,7 +31,6 @@ function Client_MSSQL(config = {}) {
     min: 1,
     max: 1,
     idleTimeoutMillis: Number.MAX_SAFE_INTEGER,
-    evictionRunIntervalMillis: 0,
```
kibertoad, line 34: Why remove it instead of replacing with reapIntervalMillis?

dhensby: Considering this hasn't been doing anything for the current version of mssql, I assumed it wasn't needed.

I can add `reapIntervalMillis` if you like?

kibertoad, line 34: @elhigu @koskimas I checked tarn.js and it appear that replicating behaviour of the old pool is no longer possible here, setting `reapIntervalMillis` to 0 is not supported. Was that intentional?

kibertoad: @dhensby I'm not sure why it was disabled in the first place, could be related to mssql always still having its own pool under-the-hood, which I'm not sure is still the case. Wonder if we double-pool it still when used from knex. If not, this whole default config could be dropped.

kibertoad: Related: https://github.com/Vincit/tarn.js/issues/56

dhensby:
> @dhensby Since that breaks compatibility with mssql 4 (which we are OK not to support in our next release which will be semver major), could you please adjust `lib/dialects/mssql/index.js` to throw an error on mssql < 5 instead of a current workaround?

I think this will "break" support for v4 and v5 (v6 is when we started using tarn as our pool, so this setting hasn't had any effect since then).

If I add a check for the current version of mssql, we could provide the right config based on that instead of throwing an error (though v4 is EOL and v5 will be when I release v7).

Lastly, the `package.json` for this lib only claims support for `^6.3.1` of mssql lib, which is why I didn't think dropping this config option would need mentioning as it has no affect in v6.

kibertoad:
@dhensby I'm OK with either adding check or just outright dropping support for 4 and 5, if from your perspective 6 and 7 are strictly better options that are more stable than either 4 or 5.
Our `package.json` is based on the fact that we only execute tests on ^6.3.1, so it is difficult to have confidence in other versions still working. People still eventually use knex with older versions though.

lib/dialects/mssql/index.js
kibertoad: @dhensby Probably the best solution here would be to disable knex pooling altogether when using node-mssql, and just use your pool, but I'm not sure if you expose it in a way that would allow us to drop-in it where we have ours now.

dhensby: We're using tarn under the hood and it's exposed via `.pool` property on our pool, which would give your library direct access to the tarn pool if that's what you needed? Though it's tarn v1 in mssql v6 and tarn v3 in mssql v7.

However, that's not something I have the time to do right now and was hoping I could just provide a quick fix to prevent errors in mssql v7 while continuing to provide support for v6.

dhensby:
Yes, 6 and 7 are much better options than the others (which aren't really maintained now) and the older versions use node-pool instead of tarnjs

lib/dialects/mssql/index.js
kibertoad: @dhensby Fair point, and this sounds like the kind of an optimization we should so ourselves on our end. Within the scope of this PR let's just remove this parameter and drop support for v4.

dhensby:
I've made changes to the driver and the UPGRADING file, however I had targeted this at 0.21 branch as I'd hoped / expected this to just be a bug fix (essentially adding support for mssql v7 whilst having no effect on mssql v6)...

kibertoad:
@dhensby Could you retarget this to master and for 0.21 do the PR with just the `evictionRunIntervalMillis` removed?

lib/dialects/mssql/index.js
elhigu: That option is not needad because min: 1, max: 1 already prevents idle connection from being discarded anyways.

dhensby:
@kibertoad created the new PR for 0.21 and targeted this one at master

kibertoad:
@dhensby Thank you! I will use this branch as a basis for reworked integration with MSSQL that will try to make use of existing tarn pool, so please don't get alarmed that it doesn't get merged for couple more weeks.

kibertoad:
Superseded by #2857

Issue #4199, opened 2021-01-03
Support for MySQL optimizer hints


martinmacko47:
# Environment

Knex version: n/a
Database + version: MySQL
OS: n/a

# Feature discussion / request

1. Explain what is your use case

    Sometimes we need to run a query with an [optimizer hint]. Such as:

    ```sql
    select /*+ semijoin(firstmatch, loosescan) */ * from ...
    select /*+ max_execution_time(1000) */ * from ...
    ```
    
    Please note that optimizer hits are something else than [index hints] which have different syntax and use.

2. Explain what kind of feature would support this

    AFAIK, currently there is no way how to add such optimizer hints in Knex, besides falling back to raw SQL.
    It would be usefull if there was a way how to tell Knex to add such hints to the generated query.

    The hints may be used at the beginning of each query:

    ```sql
    select /*+ ... */ ...
    insert /*+ ... */ ...
    replace /*+ ... */ ...
    update /*+ ... */ ...
    delete /*+ ... */ ...
    ```
    
    And at the beginning of subqueries:
    
    ```sql
    (select /*+ ... */ ... )
    (select ... ) union (select /*+ ... */ ... )
    (select /*+ ... */ ... ) union (select /*+ ... */ ... )
    update ... where x in (select /*+ ... */ ...)
    insert ... select /*+ ... */ ...
    ```

3. Give some API proposal, how the feature should work

    We could add a new query builder method `.addHint('semijoin(firstmatch, loosescan)')` that would add the hint after the query verb.
    
    I don't know how and if there is support of some kind of optimizer hints in other SQL dialents. So I can't tell how would this work for them.

[optimizer hint]: https://dev.mysql.com/doc/refman/5.7/en/optimizer-hints.html
[index hints]: https://dev.mysql.com/doc/refman/5.7/en/index-hints.html

kibertoad:
@martinmacko47 Would https://github.com/knex/knex/pull/2815 allow to achieve same thing, or syntax for comments and hints is different?

martinmacko47:
@kibertoad If I understand #2815 correctly, it adds comments **before** the verb. Like in:

```
/* comment */ select ...
```

However MySQL optimizer hits must be **directly after** the verb otherwise they don't work:

```
select /*+ semijoin(firstmatch, loosescan) */ ...
```

For instance, for select method it would need to be rendered [here between `select` and `distinctClause`].

[...]

PR #4243, opened 2021-01-20
Allow optimizer hints

martinmacko47:
Implements #4199

Optimizer hints using comment `/*+` syntax support in dialects:
- MySQL: Supported and working.
- Postgres: Not supported in vanilla Postgres. Supported in some extensions like in EDB Postgres Advanced Server, where the [syntax is the same](https://www.enterprisedb.com/edb-docs/d/edb-postgres-advanced-server/user-guides/database-compatibility-for-oracle-developers-guide/11/Database_Compatibility_for_Oracle_Developers_Guide.1.038.html
) as in MySQL. Not tested as I don't have access to EDB Postgres Advanced Server.
- Oracle: Supported. The [syntax is the same](https://docs.oracle.com/cd/B13789_01/server.101/b10752/hintsref.htm) as in MySQL. Not tested as I don't have access to Oracle DB.
- MSSQL: Not supported.
- Sqlite3: Not supported.

It is still good to keep the hints implementation even for dialects that don't support them as there may be proxy interpreters / middlewares supporting similar hints syntax.

### DIFF
--- a/lib/query/querybuilder.js
+++ b/lib/query/querybuilder.js
@@ -148,6 +148,25 @@ class Builder extends EventEmitter {
     return this;
   }
 
+  // Adds a one or more optimizer hits to the list of "optimizerHints" on the query.
+  optimizerHint() {
+    const hints = helpers.normalizeArr.apply(null, arguments);
+    if (hints.some((hint) => !isString(hint))) {
+      throw new Error('Optimizer hint must be a string');
+    }
+    if (hints.some((hint) => hint.includes('/*') || hint.includes('*/'))) {
+      throw new Error('Optimizer hint cannot include "/*" or "*/"');
+    }
+    if (hints.some((hint) => hint.includes('?'))) {
+      throw new Error('Optimizer hint cannot include "?"');
+    }
+    this._statements.push({
+      grouping: 'optimizerHints',
+      value: hints,
+    })
+    return this;
+  }
+
   // Prepends the `schemaName` on `tableName` defined by `.table` and `.join`.
   withSchema(schemaName) {
     this._single.schema = schemaName;
@@ -1040,6 +1059,7 @@ class Builder extends EventEmitter {
         'with',
         'select',
         'columns',
+        'optimizerHints',
         'where',
         'union',
         'join',
--- a/lib/query/querybuilder.js
+++ b/lib/query/querybuilder.js
@@ -148,20 +148,20 @@ class Builder extends EventEmitter {
     return this;
   }
 
-  // Adds a single hint or an array of hits to the list of "optimizerHints" on the query.
-  optimizerHint(hints) {
+  // Adds a single hint or an array of hits to the list of "hintComments" on the query.
+  hintComment(hints) {
     hints = Array.isArray(hints) ? hints : [hints]
     if (hints.some((hint) => !isString(hint))) {
-      throw new Error('Optimizer hint must be a string');
+      throw new Error('Hint comment must be a string');
     }
     if (hints.some((hint) => hint.includes('/*') || hint.includes('*/'))) {
-      throw new Error('Optimizer hint cannot include "/*" or "*/"');
+      throw new Error('Hint comment cannot include "/*" or "*/"');
     }
     if (hints.some((hint) => hint.includes('?'))) {
-      throw new Error('Optimizer hint cannot include "?"');
+      throw new Error('Hint comment cannot include "?"');
     }
     this._statements.push({
-      grouping: 'optimizerHints',
+      grouping: 'hintComments',
       value: hints,
     })
     return this;
@@ -1059,7 +1059,7 @@ class Builder extends EventEmitter {
         'with',
         'select',
         'columns',
-        'optimizerHints',
+        'hintComments',
         'where',
         'union',
         'join',
--- a/lib/query/querybuilder.js
+++ b/lib/query/querybuilder.js
@@ -148,9 +148,9 @@ class Builder extends EventEmitter {
     return this;
   }
 
-  // Adds a one or more optimizer hits to the list of "optimizerHints" on the query.
-  optimizerHint() {
-    const hints = helpers.normalizeArr.apply(null, arguments);
+  // Adds a single hint or an array of hits to the list of "optimizerHints" on the query.
+  optimizerHint(hints) {
+    hints = Array.isArray(hints) ? hints : [hints]
     if (hints.some((hint) => !isString(hint))) {
       throw new Error('Optimizer hint must be a string');
     }
--- a/lib/query/querycompiler.js
+++ b/lib/query/querycompiler.js
@@ -199,10 +199,21 @@ class QueryCompiler {
     );
   }
 
+  _optimizerHints() {
+    let hints = this.grouped.optimizerHints || [];
+    hints = hints.map((hint) => compact(hint.value).join(' '));
+    hints = compact(hints).join(' ');
+    if (hints) {
+      return `/*+ ${hints} */ `
+    }
+    return ''
+  }
+
   // Compiles the columns in the query, specifying if an item was distinct.
   columns() {
     let distinctClause = '';
     if (this.onlyUnions()) return '';
+    const hints = this._optimizerHints()
     const columns = this.grouped.columns || [];
     let i = -1,
       sql = [];
@@ -234,7 +245,7 @@ class QueryCompiler {
     }
     if (sql.length === 0) sql = ['*'];
     return (
-      `select ${distinctClause}` +
+      `select ${hints}${distinctClause}` +
       sql.join(', ') +
       (this.tableName
         ? ` from ${this.single.only ? 'only ' : ''}${this.tableName}`
--- a/lib/query/querycompiler.js
+++ b/lib/query/querycompiler.js
@@ -199,8 +199,8 @@ class QueryCompiler {
     );
   }
 
-  _optimizerHints() {
-    let hints = this.grouped.optimizerHints || [];
+  _hintComments() {
+    let hints = this.grouped.hintComments || [];
     hints = hints.map((hint) => compact(hint.value).join(' '));
     hints = compact(hints).join(' ');
     return hints ? `/*+ ${hints} */ ` : ''
@@ -210,7 +210,7 @@ class QueryCompiler {
   columns() {
     let distinctClause = '';
     if (this.onlyUnions()) return '';
-    const hints = this._optimizerHints()
+    const hints = this._hintComments()
     const columns = this.grouped.columns || [];
     let i = -1,
       sql = [];
--- a/lib/query/querycompiler.js
+++ b/lib/query/querycompiler.js
@@ -203,10 +203,7 @@ class QueryCompiler {
     let hints = this.grouped.optimizerHints || [];
     hints = hints.map((hint) => compact(hint.value).join(' '));
     hints = compact(hints).join(' ');
-    if (hints) {
-      return `/*+ ${hints} */ `
-    }
-    return ''
+    return hints ? `/*+ ${hints} */ ` : ''
   }
 
   // Compiles the columns in the query, specifying if an item was distinct.
--- a/test/integration/query/selects.js
+++ b/test/integration/query/selects.js
@@ -150,6 +150,78 @@ module.exports = function (knex) {
         });
     });
 
+    it('#4199 - adheres to optimizer hints', async function () {
+      const expectedErrors = {
+        mysql: {
+          code: 'ER_QUERY_TIMEOUT',
+          errno: 3024,
+          sqlMessage: 'Query execution was interrupted, maximum statement execution time exceeded',
+        },
+        mysql2: {
+          errno: 3024,
+          sqlMessage: 'Query execution was interrupted, maximum statement execution time exceeded',
+        },
+      }
+      if (!expectedErrors[knex.client.driverName]) {
+        return this.skip();
+      }
+      const baseQuery = knex('accounts')
+        .select('id', knex.raw('sleep(0.1)'))
+        .limit(2)
+      await expect(
+        baseQuery.clone()
+      ).to.eventually.be.fulfilled.and.to.have.lengthOf(2)
+      await expect(
+        baseQuery.clone().optimizerHint('max_execution_time(10)')
+      ).to.eventually.be.rejected.and.to.deep.include(expectedErrors[knex.client.driverName])
+    });
+
+    it('#4199 - ignores invalid optimizer hints', async function () {
+      return knex
+        .select('id')
+        .orderBy('id')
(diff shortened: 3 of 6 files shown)


types/index.d.ts
```diff
@@ -1000,6 +1001,11 @@ export declare namespace Knex {
     ): QueryBuilder<TRecord, TResult2>;
   }
 
+  interface OptimizerHint<TRecord extends {} = any, TResult extends {} = any> {
+    (...hints: string[]): QueryBuilder<TRecord, TResult>;
+    (hints: readonly string[]): QueryBuilder<TRecord, TResult>;
```
martinmacko47, line 1006: I'm not sure if this typescript rule is correct. Please check.

lib/query/querybuilder.js
```diff
@@ -148,6 +148,25 @@ class Builder extends EventEmitter {
...
+  optimizerHint() {
+    const hints = helpers.normalizeArr.apply(null, arguments);
+    if (hints.some((hint) => !isString(hint))) {
+      throw new Error('Optimizer hint must be a string');
+    }
+    if (hints.some((hint) => hint.includes('/*') || hint.includes('*/'))) {
+      throw new Error('Optimizer hint cannot include "/*" or "*/"');
+    }
+    if (hints.some((hint) => hint.includes('?'))) {
+      throw new Error('Optimizer hint cannot include "?"');
```
martinmacko47, line 161: If there is `?` in the hint it messes up parameter bindings. So I'm forbidding `?` symbols in hints. Later we can somehow  escape `?` in hints or actually allow parameter bindings in hints if they are supported by any dialect.

lib/query/querybuilder.js
```diff
@@ -148,6 +148,25 @@ class Builder extends EventEmitter {
     return this;
   }
 
+  // Adds a one or more optimizer hits to the list of "optimizerHints" on the query.
+  optimizerHint() {
+    const hints = helpers.normalizeArr.apply(null, arguments);
```
kibertoad, line 153: We really should get rid of all the `helpers.normalizeArr.apply(null, arguments);` calls.
Could this be rewritten to be `helpers.normalizeArr(...arguments)` instead? Bonus points if you could avoid using `arguments` at all and just use explicit array as a param.

lib/query/querycompiler.js
```diff
@@ -199,10 +199,21 @@ class QueryCompiler {
     );
   }
 
+  _optimizerHints() {
+    let hints = this.grouped.optimizerHints || [];
+    hints = hints.map((hint) => compact(hint.value).join(' '));
+    hints = compact(hints).join(' ');
+    if (hints) {
+      return `/*+ ${hints} */ `
```
martinmacko47, line 207: Later we may add an optional parameter to configure the `+` mark in the hint if there are some dialects or proxy / server interprets that require different marks. I don't know about any so far though.

lib/query/querybuilder.js
```diff
@@ -148,6 +148,25 @@ class Builder extends EventEmitter {
...
+    if (hints.some((hint) => !isString(hint))) {
+      throw new Error('Optimizer hint must be a string');
+    }
+    if (hints.some((hint) => hint.includes('/*') || hint.includes('*/'))) {
```
kibertoad, line 157: Fastest way to do this check is `indexOf <...> !== -1`

lib/query/querycompiler.js
```diff
@@ -199,10 +199,21 @@ class QueryCompiler {
     );
   }
 
+  _optimizerHints() {
+    let hints = this.grouped.optimizerHints || [];
+    hints = hints.map((hint) => compact(hint.value).join(' '));
+    hints = compact(hints).join(' ');
+    if (hints) {
```
kibertoad, line 206: can be a ternary statement

test/integration/query/selects.js
```diff
@@ -150,6 +150,78 @@ module.exports = function (knex) {
         });
     });
 
+    it('#4199 - adheres to optimizer hints', async function () {
+      const expectedErrors = {
+        mysql: {
```
kibertoad, line 155: can you also implement this at least for PostgreSQL?

test/integration/query/selects.js
```diff
@@ -150,6 +150,78 @@ module.exports = function (knex) {
         });
     });
 
+    it('#4199 - adheres to optimizer hints', async function () {
```
martinmacko47, line 153: I don't know how to test any other hint besides `max_execution_time` in MySQL as they don't alter the result just the alhorithm.

Also I don't know how to test any of the hints used in Oracle.

kibertoad:
Great job!
This also needs documentation update at https://github.com/knex/documentation

test/integration/query/selects.js
```diff
@@ -150,6 +150,78 @@ module.exports = function (knex) {
...
+        .limit(2)
+      await expect(
+        baseQuery.clone()
+      ).to.eventually.be.fulfilled.and.to.have.lengthOf(2)
+      await expect(
+        baseQuery.clone().optimizerHint('max_execution_time(10)')
+      ).to.eventually.be.rejected.and.to.deep.include(expectedErrors[knex.client.driverName])
+    });
+
+    it('#4199 - ignores invalid optimizer hints', async function () {
```
martinmacko47, line 179: Neither Postres, Sqlite3 nor MSSQL support this kind of hints. So I at least test they don't interfere if used.

lib/query/querybuilder.js
martinmacko47: @kibertoad Sure, I just followed the patterns in code arround

test/integration/query/selects.js
kibertoad: Actually, PostgreSQL just doesn't support hints and don't intend to. Could you make it throw an error when you try to use it on PG? Also I think SQLite doesn't support it either.

test/integration/query/selects.js
kibertoad: MSSQL does, I think: https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-query?view=sql-server-ver15

Everything that does not should be throwing an error.

types/index.d.ts
```diff
@@ -1000,6 +1001,11 @@ export declare namespace Knex {
     ): QueryBuilder<TRecord, TResult2>;
   }
 
+  interface OptimizerHint<TRecord extends {} = any, TResult extends {} = any> {
+    (...hints: string[]): QueryBuilder<TRecord, TResult>;
```
kibertoad, line 1005: I would prefer not to support this at all and just force user to pass array always

types/index.d.ts
kibertoad: @lorefnon Could you take a look?

lib/query/querybuilder.js
martinmacko47: @kibertoad I followed this review comment in https://github.com/knex/knex/pull/2815#discussion_r268410986

lib/query/querycompiler.js
martinmacko47: @kibertoad Sure I can change it, I just lookend arround the codebase, and it seems that

```
if (...) return ...
return ...
```

pattern is much more frequent than

```
return ... ? ... : ....
```
pattern. So I followed the more common style.

Should I change it?

lib/query/querybuilder.js
kibertoad: fair enough :)

lib/query/querycompiler.js
kibertoad: codebase, unfortunately, is full of suboptimal code, definitely don't follow it :)

kibertoad: but either is fine, pick the one that looks most readable to you

test/integration/query/selects.js
martinmacko47: @kibertoad  Vanilla PostgreSQL does not suport optimizer hints of this kind. It is supported just in extensions like [EDB Postgres Advanced Server](https://www.enterprisedb.com/edb-docs/d/edb-postgres-advanced-server/user-guides/database-compatibility-for-oracle-developers-guide/11/Database_Compatibility_for_Oracle_Developers_Guide.1.038.html
).

types/index.d.ts
martinmacko47: @kibertoad You mean like either a single hint or an array of hints? Or you mean that even for a single hint an array with this single hint should be passed?

kibertoad: Either would work. Either support a singular argument and wrap it into array inside, or always require an array; main thing I don't want to support is an arbitrary amount of separate arguments being passed and spread later.

test/integration/query/selects.js
martinmacko47: Is it good to throw errors if hints are used for dialects that don't support hints? There are proxies or middlewares that may listen to the hints. Like in #1982 they needed comments for passing hints for their router. Also there are DB extensions like EDB Postgres Advanced Server which support hints while vanilla Postgres does not.

kibertoad: Good point, then we shouldn't throw an error.

lib/query/querybuilder.js
martinmacko47: @kibertoad Changed in 65f488b. I removed variadic arguments, therefore it's now just simple `optimizerHint(hints)`

lib/query/querycompiler.js
martinmacko47: changed in 8414e94

types/index.d.ts
martinmacko47: @kibertoad Changed in 65f488b

test/integration/query/selects.js
martinmacko47: @kibertoad The problem with MSSQL is that it uses OPTION clause instead of `/*+ ... */` comment syntax, which works differently.

For instance we cannot just simply validate the hint for non-existence of `*/` within the hint to prevent a premature end of the hint as mentioned in https://github.com/knex/knex/pull/2815#discussion_r261689009. We would need to parse the hint itself to know whether there are not more `)` than `(` in it. Otherwise any code could leak into the query passing somehting like `.optimizerHint('HASH GROUP) injected sql (...')`.

Also as the OPTION clause is a normal SQL clause, it supports parameter bindings, while the comment hints syntax does not. So I'm not sure if it is good to mingle two so different syntexes into a single method. Possibly we could implement the OPTION clause using a separate method.

OTOH, we may let MSSQL use the OPTION clause for the hints, but without checking for injections within hints. And with forbidden parameter bindings for now, as adding parameter bindings will mess with the comment syntax. Assuming there are no proxy interpreters / middlewares / routers for MSSQL using the comments hint syntax. Otherwise we would prevent users from passing the comment hints to them when using MSSQL dialect.

Perhaps we could rename the `.optimizerHint(hint)` method to something like `.commentHint(hint)` to emphasize it is for hints passed via the `/*+` comment syntax regardless of used dialect.

martinmacko47:
>  This also needs documentation update at https://github.com/knex/documentation

I'll add docs after we consolidate the new API. Number of parameters, comments syntax vs OPTIONS clause for hists, parameter bindings and so.

test/integration/query/selects.js
kibertoad: @martinmacko47 Then it sounds like current validation logic should be moved from generic querybuilder into MySQL querybuilder. Then we could implement it differently for MSSQL. Would that allow to use same `optimizerHint` for both?

martinmacko47: @kibertoad I guess we need to decide whether we want:

a) A semantic method for optimizer hints as they are supported in each dialect. That would be comment-like syntax for MySQL and Oracle, OPTION clause for MSSQL, and unsupported exception for Postgres and SQLite. With no way to pass comment-like hints for proxies / routers / whatever middlewares people may use.

b) A common method for any hints with comment-like syntax regardless they are consumed by the DB itself or any party on the way to DB.

I don't really have an opinion which one is more useful.

kibertoad: It's a tough one to pick. On one heand one of the main goals of knex is to abstract away dialect differences between different DBs, so that swapping them works as transparently as possible. Now in this case one would obviously need to rewrite hints across dbs either way, but using unified syntax for that is still likely to be more familiar for people doing the change.

However, I see how that would be limiting the use of the hints for non-optimizer cases, hence I probably would vote for the second one, with very explicit documentation of what use-case it could be used.

martinmacko47: @kibertoad Then let's keep the comment-like syntax hints for all dialects and we will try to document it properly. Just perhaps let's name the method not `optimizerHint` so users won't confuse it with other types of optimizer hints in other dialects, but somehow else.

martinmacko47:
>  This also needs documentation update at https://github.com/knex/documentation

@kibertoad Docs added in https://github.com/knex/documentation/pull/306

test/integration/query/selects.js
kibertoad: MySQL documentation calls them "optimizer hint comments", so either `optimizerHintComment` or `hintComment` works for me.

martinmacko47: @kibertoad Renamed to `hintComment` in ad8b9a6. The other is too long IMHO.

martinmacko47:
@kibertoad I guess the PR is ready. What else can I do to have it merged and released? Thanks.

kibertoad:
will review in the evening, thanks!

kibertoad (approved):

kibertoad:
@martinmacko47 Since this is a part of a next semver major release which is not yet fully ready, I'm afraid I can't cut a new stable version yet. I see two options if you need it released sooner than in two-three weeks: either you'd have to backport it to 0.21 branch, or I can release a preview NEXT version with this change.

martinmacko47:
> @martinmacko47 Since this is a part of a next semver major release which is not yet fully ready, I'm afraid I can't cut a new stable version yet. I see two options if you need it released sooner than in two-three weeks: either you'd have to backport it to 0.21 branch, or I can release a preview NEXT version with this change.

@kibertoad Sure, I can backport it to 0.21 branch. How to do it? Just to cherrypick the commit to 0.21 branch, resolve conflicts and create a new PR onto 0.21? Or is there anything else not to forget about?

kibertoad:
@martinmacko47 yeah, that would be worth a try, but I expect some conflicts, as master was heavily refactored since then. as long as tests will pass, should be good.

martinmacko47:
@kibertoad Here is the backport PR https://github.com/knex/knex/pull/4258

I didn't find 0.21 base branch for docs. Is there one? Or how to backport docs?

kibertoad:
@martinmacko47 There is no way to do that, unfortunately, so I'll just link to the documentation PR in the changelog.

Travis CI is sloooooooooow, give it a couple of hours to kick in.

PR #4245, opened 2021-01-20
Timestamp UTC Standardization for Migrations

Tolterix:
The rest parameter is for testing purposes.


Tolterix:
Just a note that this is the continuation of the discussion had for #4164.

kibertoad:
Thank you for your contribution! Could you please explain why there is still lots of complex logic and some custom seed files in this PR? I think we previously discussed just always using the UTC, no?

Tolterix:
A lot of the complex logic comes from changing time by the timezone(noted by the offset). As you probably know, changing an hour can potentially change days, months, or even years. Due to this, I had to account for leap years and months passing by by the hours being on a different timezone than UTC given by the computers offset. The custom seed files can be deleted, I thought I deleted them sorry about that. @kibertoad

Tolterix:
As I was running the tests on my local machine it was creating the seeds from the seed:make test but wouldnt delete them properly after running it.

kibertoad:
@Tolterix Why do you need to apply any kind of offset if you could simply instruct JS to generate current UTC timestamp?..

Tolterix:
@kibertoad I suppose you could, guess if you are really trying to crunch down I can do that instead. The initial reason I went this route was if a team wanted to possibly have functionality sometime in the future to give a specific timezone to work in, but if you want me to eliminate that possibility I most certainly can.

Tolterix:
Also since this is doing the same thing as Date.UTC(), it works the same way.

kibertoad:
@Tolterix I believe we already discussed that just using UTC in all cases is perfectly sufficient, timestamp is only used for ordering db resources in chronological order.

lib/migrations/util/timestamp.js
```diff
@@ -1,16 +1,14 @@
-// Get a date object in the correct format, without requiring a full out library
-// like "moment.js".
 function yyyymmddhhmmss() {
-  const d = new Date();
+	const now = new Date(Date.now());
```
kibertoad, line 2: Why use `new Date(Date.now());` together instead of either of these two?

test/cli/seed-make.spec.js
```diff
@@ -395,6 +395,7 @@ development: {
 
     it('Creates a new seed using --timestamp-filename-prefix CLI flag', async () => {
       const seedGlobPath = `${process.cwd()}/seeds/*_somename.js`;
+      console.log(seedGlobPath);
```
kibertoad, line 398: is this necessary?

lib/migrations/util/timestamp.js
Tolterix: Date.now() returns milliseconds from 1970 to now, to my understanding there is no integer prototype for .getUTCSeconds, so you have to convert it to a new Date() object.

test/cli/seed-make.spec.js
Tolterix: nope, thought i got rid of that. apparently not. thanks for the catch, will fix now.

lib/migrations/util/timestamp.js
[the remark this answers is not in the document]
kibertoad: https://stackoverflow.com/questions/31591098/how-do-i-stub-new-date-using-sinon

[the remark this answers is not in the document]
kibertoad: What is the difference between `new Date()` and ` new Date(Date.now());`? Don't both return the present time?

PR #4253, opened 2021-01-27
Migrate to classes

maximelkin:


maximelkin:
tests breaking on 
{code:js}
    function ClientFoobar(config) {
      SqliteClient.call(this, config);
    }
{code}

Do we want to drop old way of clients declaration?

kibertoad:
> Do we want to drop old way of clients declaration?

Yes, definitely.

UPGRADING.md
```diff
@@ -29,6 +29,7 @@ const knexInstance: Knex = knex(config)
 * v8 flags are no longer supported in cli. To pass these flags use [`NODE_OPTIONS` environment variable](https://nodejs.org/api/cli.html#cli_node_options_options).
   For example `NODE_OPTIONS="--max-old-space-size=1536" npm run knex`
 
+* Clients are now classes instead of new-able functions. Please migrate your custom clients to classes.
```
kibertoad, line 32: Can you add an example of how the old code looked like and how the new one would look?

lib/client.js
```diff
@@ -33,151 +33,145 @@ const debug = require('debug')('knex:client');
...
-    this.connectionConfigExpirationChecker = () => true; // causes the provider to be called on first use
-  } else {
-    this.connectionSettings = cloneDeep(config.connection || {});
-    this.connectionConfigExpirationChecker = null;
-  }
-  if (this.driverName && config.connection) {
-    this.initializeDriver();
-    if (!config.pool || (config.pool && config.pool.max !== 0)) {
-      this.initializePool(config);
+    if (this.dialect && !this.config.client) {
```
kibertoad, line 47: I would suggest either removing this altogether or replacing warning with an error.

lib/client.js
```diff
@@ -378,7 +370,17 @@ Object.assign(Client.prototype, {
       );
     }
     return unwrapRaw(value, true, builder, this, formatter) || '?';
-  },
+  }
+}
+
+Object.assign(Client.prototype, {
```
kibertoad, line 376: This really feels like it should be composed in using an object rather than hacked in as a dynamically defined method, but that would be a larger, separate refactoring.

lib/dialects/mssql/index.js
```diff
@@ -369,7 +361,14 @@ Object.assign(Client_MSSQL.prototype, {
       default:
         return response;
     }
-  },
+  }
+}
+
+// until node 12+
```
kibertoad, line 367: what feature from Node 12 could be used here?

lib/dialects/mysql2/index.js
```diff
@@ -1,35 +1,31 @@
...
     }
     return true;
-  },
+  }
+}
+
+// until node 12+
+Object.assign(Client_MySQL2.prototype, {
+  // The "dialect", for reference elsewhere.
+  driverName: 'mysql2',
```
kibertoad, line 28: This can't be simply a getter?

lib/dialects/mssql/index.js
maximelkin: class field properties

maximelkin: oh, they are declared not on prototypes

lib/dialects/mysql2/index.js
maximelkin: Users could override them using assign in constructor, declaring as getter could break their logic

lib/client.js
maximelkin: I think this is not in scope of merge request, i will create later MR with throwing error

UPGRADING.md
maximelkin: done

lib/dialects/mssql/index.js
maximelkin: deleted this lines

kibertoad (approved):

kibertoad:
Thanks, this is awesome!

PR #4256, opened 2021-01-28
Improve join and conflict types

hermansje:
With TypeScript 4.1, literal types can be constructed. This enables better type checking where it wasn't possible before.

What I did:
- Increase minimum TypeScript version to 4.1 to support template literal types
- Type join arguments
- Type onConflict (+ merge and ignore)

### DIFF
--- a/test-tsd/types.test-d.ts
+++ b/test-tsd/types.test-d.ts
@@ -99,3 +99,16 @@ expectAssignable<QueryBuilder>(
     .merge({ x: 'x' })
     .debug(true)
 );
+
+expectType<
+  QueryBuilder<
+    User,
+    DeferredKeySelection<User, 'id', true, {}, true, {}, never>[]
+  >
+>(
+  knexInstance
+    .table<User>('users')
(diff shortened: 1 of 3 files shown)


kibertoad:
@lorefnon What do you think about this?
(TS version bump should be fine, next version is a semver major anyway)

lorefnon:
@hermansje This looks impressive. Thanks. 

I am curious how/if this handles table aliasing (esp. within joins).

sondresj:
Related issue? https://github.com/knex/knex/issues/4246

kibertoad:
@sondresj Can you check if changes in this PR resolve your issue?

sondresj:
@kibertoad The issue mentioned is not reported by me, but I have the same issue. 

Just tested locally in my project using datacamp-engineering/knex.git#jh/upsert-types as knex dependency version, and the types are correct 👍

kibertoad:
@sondresj Thanks! Hopefully we can land this prior to 0.95.0 release.

kibertoad:
merged as #4318

PR #4266, opened 2021-02-02
Primary keys improvements

gpetrov:
Including changes for:

## Improve mySQL rename column to keep auto increment

Auto increment is removed when column is renamed in mySQL.
It has to be specified again so it is kept.

Fixed issue knex#2767

## Implementing new option primaryKey: False for the increments and bigIncrements types.

Usage:

```js
t.increments('id', { primaryKey: false }) => Not primary key
t.increments('id') => Primary key as currently
```

Rebasing of knex#2011

### DIFF
--- a/lib/dialects/mysql/schema/mysql-tablecompiler.js
+++ b/lib/dialects/mysql/schema/mysql-tablecompiler.js
@@ -92,6 +92,10 @@ class TableCompiler_MySQL extends TableCompiler {
               if (column.Default !== void 0 && column.Default !== null) {
                 sql += ` DEFAULT '${column.Default}'`;
               }
+              // Add back the auto increment if the column had it, fix issue #2767
+              if (column.Extra == 'auto_increment') {
+                sql += ` AUTO_INCREMENT`;
+              }
 
               return runner.query({
                 sql,
--- a/test/integration/schema/index.js
+++ b/test/integration/schema/index.js
@@ -1238,6 +1238,51 @@ module.exports = (knex) => {
               });
           });
         });
+
+        it('#2767 - .renameColumn should not drop the auto incremental', () => {
+          const tableName = 'rename_column_test';
+
+          if (knex && knex.client) {
+            if (/mssql/i.test(knex.client.driverName)) {
+              return knex
+                .raw(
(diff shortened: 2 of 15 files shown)


gpetrov:
There you go @kibertoad 

Added also the unit tests and fixed some other problems.

Not sure why some unit tests failing under Node 12 - I just used the main branch.

lib/dialects/mysql/schema/mysql-tablecompiler.js
```diff
@@ -92,6 +92,10 @@ class TableCompiler_MySQL extends TableCompiler {
               if (column.Default !== void 0 && column.Default !== null) {
                 sql += ` DEFAULT '${column.Default}'`;
               }
+              // Add back the auto increment if the column had it, fix issue #2767
+              if (column.Extra == 'auto_increment') {
```
kibertoad, line 96: I think this part is not currently covered by tests, could you add one?

Also shouldn't similar changes be done for other dialects as well?

kibertoad:
@gpetrov See GA logs, syntax is slightly different:
```
       -alter table `users` add column `id` integer not null autoincrement
      +alter table "users" add column "id" integer not null autoincrement
```

Also the change needs to be documented at https://github.com/knex/documentation

lib/dialects/mysql/schema/mysql-tablecompiler.js
gpetrov: It is a mySQL only issue. Will add a tests for it

kibertoad: @gpetrov It would be best to add a proper integration test (see test/integration2) that would try inserting something after manipulating schema and checking that autoincrement works as expected. That would cover all the DBs automatically.

test/integration/schema/index.js
```diff
@@ -1238,6 +1238,51 @@ module.exports = (knex) => {
               });
           });
         });
+
+        it('#2767 - .renameColumn should not drop the auto incremental', () => {
```
kibertoad, line 1242: see my comment. is it not possible to make test db-agnostic?

gpetrov: the tests are very different depending on the db type. So I've added tests per db to validate it doesn't happen in other dbs and to proof that it is only a mysql problem

lib/dialects/mysql/schema/mysql-tablecompiler.js
gpetrov: well I just extended the already available renameColumn tests - there were just fine to check for the the auto incremental afterwards.

kibertoad (approved):

Issue #1368, opened 2016-04-26
Subquery not working with withSchema


dizlexik:
I ran into this issue today and it's blocking me for now :/

This produces the expected query:
`knex.select(1).from(function() { this.select(1).from('mytable') })`
-> `select 1 from (select 1 from [mytable])`

This does not:
`knex.withSchema('myschema').select(1).from(function() { this.select(1).from('mytable') })`
-> `select 1 from [myschema].[function () { this].[select(1)].[from('mytable'); }]`

igor-savin-ht:
@elhigu Any way to contribute to this issue? Would e. g. test that reproduces the issue help?

elhigu:
@igor-savin-ht PR with test + fix would be really helpful. I don't have fixing this issue in my todo list.

devourment77:
I hit this issue today.  It looks like the schema is prepended to the subquery (EX: `<schema>.select`)

kibertoad:
Released in 0.95.0, please reopen if this still does not work for you.

Issue #3654, opened 2020-02-07
leftJoin will prefix schema even when passed a `raw`


DataTables:
# Environment

Knex version: 0.20.8
Database + version: postgres 10
OS: Fedora 29

# Bug

## 1 - Explain what kind of behaviour you are getting and how you think it should do

When using `withSchema()` and `leftJoin` the schema name is prefixed to `knex.raw()` parameters for the joining table name. Since a `raw` statement is being passed in, I would not expect the schema name to be prefixed.

I think this relates to #2054.

This is a problem because it means a left join cannot be used with a CTE (since the CTE alias is not prefixed with the schema name, but the left join 'from' parameter is).

## 2 - Error message

No error message - just unexpected SQL output.

## 3 - Reduced test code

https://runkit.com/embed/gzo8okof4w0g

In case Runkit is offline for any reason:

```js
const Knex = require('knex');
const knexPg = Knex({
  client: 'pg',
});

let query = knexPg
  .withSchema('testSchema')
  .select('id', 'name')
  .from('host')
  .leftJoin(knexPg.raw('joiner'), 'linkA', 'linkB');
  
console.log(query.toSQL().sql);
```

Results in the SQL statement:

```sql
select "id", "name"
from "testSchema"."host"
left join "testSchema"."joiner" on "linkA" = "linkB"
```

I expected:

```sql
select "id", "name"
from "testSchema"."host"
left join "joiner" on "linkA" = "linkB"
```

DataTables:
Looking into this, I believe the issue is [here](https://github.com/knex/knex/blob/589ea747c2388c8ef00afb933da193eb10201735/lib/query/compiler.js#L297):

```js
const table = join.schema ? `${join.schema}.${join.table}` : join.table;
```

If you agree that this is an issue, then I think the fix is:

```js
const table = join.schema && !(join.table instanceof Raw)
  ? `${join.schema}.${join.table}`
  : join.table;
```

Happy to send a PR with this change if you think it is suitable.

mercmobily:
Good news! This issue has been resolved and can be closed. The fix was implemented in the codebase to prevent schema prefixing when using `knex.raw()` in join statements.

When using `withSchema()` along with `leftJoin()` and passing a `knex.raw()` value for the table name, Knex was incorrectly prefixing the schema name to the raw SQL. This was problematic for CTEs (Common Table Expressions) where the CTE alias should not be schema-prefixed.

**Expected behavior**: Raw SQL should not be schema-prefixed
[...]

Issue #2054, opened 2017-05-10
withSchema + with clause doesn't work for innerJoin and such


yossil:
In posgresql using 'withSchema' clause and 'with' clause for CTE,
knex added the schema also to the CTE alias so it failes!

I think it should recognize that the 'table' name is a CTE and then not append the schema

Example query (from docs - modified):
`
knex
**.withSchema('my_schema')**
.with('with_alias', knex.raw('select * from "books" where "author" = ?', 'Test'))
.select('*').from('with_alias')
`

elhigu:
There is no way that `knex` can tell difference that you are actually referring CTE query instead of table, so it cannot be fixed to work perfectly automatically (there can be table and CTE with the same name I suppose). 

In theory it could be made easier if there would be map of all CTEs stored and it would be checked for every identifier (not so straightforward that it sounds and requires huge amount of tests).

Also it is not very well defined, which are all the places where `withSchema()` passed table should be passed (for if inside CTE it should be applied the same way that in upper level... I suppose it should be).

Anyways there should be way to tell that explicitly table name without any conversion for example with raw and at least following should be supported:

```js
knex.withSchema('my_schema').with('with_alias', knex.raw('select * from "books" where "author" = ?', 'Test')).select('*').from(knex.raw('with_alias')).toSQL();

// this doesn't work yet, should be fixed
```

PRs fixing the `.from`to allow passing `knex.raw` without adding schema would be welcome.

yossil:
Thanks for the detailed answer!

For now my solution will be to use a function with schema_name as parameter instead.

jebricker:
Is this similar when using .withSchema and .joins?

```
let q = knex.withSchema('rcubed').select('srt.SUBMISSION_REQ_TMPLT_ID').from({srt:'SUBMISSION_REQ_TMPLT '})
        .join({ptr: 'SUB_REQ_PRODUCT_TYP_REF'}, 'srt.SUBMISSION_REQ_TMPLT_ID', '=', 'ptr.SUBMISSION_REQ_TMPLT_ID') 
.where('ptr.PRODUCT_TYPE_ID',product_type_id)
```
produces 
select srt.SUBMISSION_REQ_TMPLT_ID from rcubed.[object Object] inner join rcubed.[object Object] on srt.SUBMISSION_REQ_TMPLT_ID = ptr.SUBMISSION_REQ_TMPLT_ID where PRODUCT_TYPE_ID = 1

Before adding the join it worked

[...]

PR #4268, opened 2021-02-02
Fix issue with schema usage with FROM clause

Saka7:
Fixes issue: #1368
Related issues: #2054, #3654

# Description
Prevent adding a `schema` (specified by `.withSchema`) to a `.from` caluse containing a `QueryBuilder`, `Raw` or `function`.

### DIFF
--- a/lib/query/querycompiler.js
+++ b/lib/query/querycompiler.js
@@ -340,7 +340,10 @@ class QueryCompiler {
     if (!joins) return '';
     while (++i < joins.length) {
       const join = joins[i];
-      const table = join.schema ? `${join.schema}.${join.table}` : join.table;
+      const table =
+        join.schema && !(join.table instanceof Raw)
+          ? `${join.schema}.${join.table}`
+          : join.table;
       if (i > 0) sql += ' ';
       if (join.joinType === 'raw') {
         sql += this.formatter.unwrapRaw(join.table);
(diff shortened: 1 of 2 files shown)


Saka7:
IMHO, In case this PR is merged we should implement #3654 as well for consistency

lib/query/querycompiler.js
```diff
@@ -1199,7 +1199,13 @@ class QueryCompiler {
       let tableName = this.single.table;
       const schemaName = this.single.schema;
 
-      if (tableName && schemaName) tableName = `${schemaName}.${tableName}`;
+      const isQueryBuilder = tableName instanceof QueryBuilder;
+      const isRawQuery = tableName instanceof Raw;
+      const isFunction = typeof tableName === 'function';
+
+      if (tableName && schemaName && !isQueryBuilder && !isRawQuery && !isFunction) {
```
kibertoad, line 1206: Since last three checks are only relevant if first three pass, I would split checks in two blocks and only resolve these conditions if first two pass.

kibertoad:
@Saka7 I agree. Can you either add the change to this PR or open a separate one?

kibertoad:
Also there are conflicts now.

Saka7:
> @Saka7 I agree. Can you either add the change to this PR or open a separate one?

I've added the fix into this PR

> @Saka7 Also there are conflicts now.

Resolved

lib/query/querycompiler.js
Saka7: Refactored

kibertoad (approved):

kibertoad:
Thank you!

PR #4272, opened 2021-02-03
Simplify reinsert logic when altering a table in SQLite

nickrum:
This is the procedure recommended by the [SQLite Docs](https://sqlite.org/lang_altertable.html#otheralter). It saves a few SQL statements, most importantly copying the data twice.

Unfortunately, SQLite's `RENAME TABLE` adds double quotes around the table name, so I had to alter some of the tests.

### DIFF
--- a/lib/dialects/sqlite3/schema/internal/sqlite-ddl-operations.js
+++ b/lib/dialects/sqlite3/schema/internal/sqlite-ddl-operations.js
@@ -13,20 +13,14 @@ function insertChunked(trx, chunkSize, target, iterator, existingData) {
   return result;
 }
 
-function createTempTable(createTable, tablename, alteredName) {
-  return createTable.sql.replace(tablename, alteredName);
+function createNewTable(sql, tablename, alteredName) {
+  return sql.replace(tablename, alteredName);
 }
 
 // ToDo To be removed
-async function copyData(trx, tableName, alteredName) {
+async function copyData(trx, iterator, tableName, alteredName) {
   const existingData = await trx.raw(`SELECT * FROM "${tableName}"`);
-  return insertChunked(trx, 20, alteredName, identity, existingData);
-}
-
-// ToDo To be removed
-async function reinsertData(trx, iterator, tableName, alteredName) {
-  const existingData = await trx.raw(`SELECT * FROM "${alteredName}"`);
-  return insertChunked(trx, 20, tableName, iterator, existingData);
+  return insertChunked(trx, 20, alteredName, iterator, existingData);
 }
 
 function copyAllData(sourceTable, targetTable) {
@@ -35,10 +29,6 @@ function copyAllData(sourceTable, targetTable) {
 
 function dropOriginal(tableName) {
   return `DROP TABLE "${tableName}"`;
-}
-
-function dropTempTable(alteredName) {
-  return `DROP TABLE "${alteredName}"`;
 }
 
 function renameTable(tableName, alteredName) {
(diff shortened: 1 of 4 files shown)


lib/dialects/sqlite3/schema/internal/sqlite-ddl-operations.js
```diff
@@ -37,12 +31,8 @@ function dropOriginal(tableName) {
   return `DROP TABLE "${tableName}"`;
 }
 
-function dropTempTable(alteredName) {
-  return `DROP TABLE "${alteredName}"`;
-}
-
 function renameTable(tableName, alteredName) {
-  return `ALTER TABLE "${tableName}" RENAME TO "${alteredName}"`;
+  return `ALTER TABLE "${alteredName}" RENAME TO "${tableName}"`;
```
kibertoad, line 35: why was this swapped?

nickrum: Because the temporary altered table has to be renamed to the original table name. The function was only used in `renameTable()` in `ddl.js`, which in turn was unused.

kibertoad: I would then change the passed params, because original naming implies "[original]Name" -> "alteredName", and now it is confusing.

lib/dialects/sqlite3/schema/internal/sqlite-ddl-operations.js
```diff
@@ -13,20 +13,14 @@ function insertChunked(trx, chunkSize, target, iterator, existingData) {
...
-// ToDo To be removed
-async function copyData(trx, tableName, alteredName) {
-  const existingData = await trx.raw(`SELECT * FROM "${tableName}"`);
-  return insertChunked(trx, 20, alteredName, identity, existingData);
+function createNewTable(sql, tablename, alteredName) {
+  return sql.replace(tablename, alteredName);
 }
 
 // ToDo To be removed
 async function reinsertData(trx, iterator, tableName, alteredName) {
```
kibertoad, line 21: Is the name still accurate, if we are using it not for reinsertion, but just for copying?

kibertoad:
A bunch of tests are failing.

lib/dialects/sqlite3/schema/internal/sqlite-ddl-operations.js
nickrum: Not precisely. I can rename the function to `copyData`. But wouldn't that mean that `reinsertMapped` and `generateReinsertCommands` aren't accurate either? Any suggestions for alternative names?

kibertoad: How about `copyMapped` and `generateCopyDataCommands`?

nickrum: I thought of that as well but those methods are doing much more than just copying, especially with PR #4189. What do you think of `alter` (or `alterMapped`) and `generateAlterCommands`?

kibertoad: Sure!

nickrum:
Not sure, why some of the mysql tests are failing.

kibertoad:
@nickrum I think this is GitHub going haywire internally, I'm observing the same on other PRs now. Probably will resolve itself in a couple of hours.

kibertoad:
@nickrum It's green now.

kibertoad (approved):

PR #4277, opened 2021-02-05
Recreate indices when altering a table in SQLite

nickrum:
I tried to separate the parsing and the altering logic as much as possible.

When querying `sqlite_master` for create index statements, it seems like only the first part of those statements is in uppercase. It would be nice if the casing was consistent but when I tried to uppercase the index creation statements in the SQLite TableCompiler, nearly all tests failed with cryptic errors.

### DIFF
--- a/lib/dialects/sqlite3/schema/ddl.js
+++ b/lib/dialects/sqlite3/schema/ddl.js
@@ -9,6 +9,7 @@ const fromPairs = require('lodash/fromPairs');
 const isEmpty = require('lodash/isEmpty');
 const negate = require('lodash/negate');
 const omit = require('lodash/omit');
+const identity = require('lodash/identity');
 const { nanonum } = require('../../../util/nanoid');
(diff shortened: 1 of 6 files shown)


lib/dialects/sqlite3/schema/internal/unparser.js
```diff
@@ -0,0 +1,19 @@
+function unparseCreateIndex(ast, wrap = (v) => v) {
```
kibertoad, line 1: I would suggest to use term `compile` instead of `unparse`

kibertoad:
Great job! Other than minor naming nitpick, looks amazing.

lib/dialects/sqlite3/schema/internal/unparser.js
nickrum: I couldn't think of a better name but I like that suggestion.

nickrum:
Thanks! The naming should be fixed.

kibertoad (approved):

PR #4281, opened 2021-02-08
mssql: add all mssql options to Knex.ConnectionConfig

tylerjwatson:
This patch introduces typings for all the available connection options in `mssql` with the tedious driver.

It includes tsd tests to assert type overloading of the `MsSqlConnectionConfig` union type.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1941,7 +1941,7 @@ interface MsSqlDefaultAuthenticationConfig extends MsSqlConnectionConfigBase {
   /**
    * User name to use for sql server login.
    */
-  userName?: string;
+  user?: string;
   /**
    * Password to use for sql server login.
    */
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1927,8 +1927,137 @@ export declare namespace Knex {
     requestTimeout?: number;
   }
 
-  // Config object for mssql: see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mssql/index.d.ts
-  interface MsSqlConnectionConfig {
+  type MsSqlAuthenticationTypeOptions =
+    | 'default'
+    | 'ntlm'
+    | 'azure-active-directory-password'
+    | 'azure-active-directory-access-token'
+    | 'azure-active-directory-msi-vm'
+    | 'azure-active-directory-msi-app-service'
+    | 'azure-active-directory-service-principal-secret';
+
+interface MsSqlDefaultAuthenticationConfig extends MsSqlConnectionConfigBase {
+  type?: 'default' | never;
+  /**
+   * User name to use for sql server login.
+   */
+  userName?: string;
+  /**
+   * Password to use for sql server login.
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -1927,8 +1927,137 @@ export declare namespace Knex {
...
+    | 'azure-active-directory-msi-vm'
+    | 'azure-active-directory-msi-app-service'
+    | 'azure-active-directory-service-principal-secret';
+
+interface MsSqlDefaultAuthenticationConfig extends MsSqlConnectionConfigBase {
+  type?: 'default' | never;
+  /**
+   * User name to use for sql server login.
+   */
+  userName?: string;
```
kibertoad, line 1944: Is this correct, considering that we are remapping from the usual `knexconfig.user` field?

types/index.d.ts
```diff
@@ -1927,8 +1927,137 @@ export declare namespace Knex {
...
+    | 'azure-active-directory-msi-vm'
+    | 'azure-active-directory-msi-app-service'
+    | 'azure-active-directory-service-principal-secret';
+
+interface MsSqlDefaultAuthenticationConfig extends MsSqlConnectionConfigBase {
+  type?: 'default' | never;
+  /**
+   * User name to use for sql server login.
+   */
+  user?: string;
```
kibertoad, line 1944: since user and password are coming from `MsSqlConnectionConfigBase` already, I think they can be dropped from all the extension types.

kibertoad:
@tylerjwatson New type tests were not failing for me even with wrong values, so I've adjusted to a different assertion style that was failing when it should.
Also after checking documentation, I've changed back types to match mssql ones, not the knex, as this better correlates to `The connection options are passed directly to the appropriate database client` sentiment expressed in the documentation (even though we are still doing our fair share of remapping, which was already the case for mssql, but I'm open to switch to just passing connection params as-is)

tylerjwatson:
@kibertoad Sweet, sounds good.   I didn't expect to have this merged so soon.

kibertoad:
@tylerjwatson I have everything I wanted to have for 0.95.0 merged already, so the plan for release is this now:

1) Wait for the querycontext PR from you;
2) Release preview2;
3) Get confirmation from you that it is working well;
4) Release final 0.95.0

tylerjwatson:
> Get confirmation from you that it is working well;

Cool.  It's going to take a little while to get full confirmation though, our production cycles take a little while but I can absolutely run a `preview2` release in our stack for full confidence.

kibertoad:
@tylerjwatson Thanks!  And do you have any timeline on when you will have time to work on the querycontext part?

tylerjwatson:
Yep will do it most likely on the weekend.

On 10 Feb 2021, at 6:47 pm, Igor Savin <notifications@github.com> wrote:

﻿

@tylerjwatson<https://github.com/tylerjwatson> Thanks! And do you have any timeline on when you will have time to work on the querycontext part?

Issue #4295, opened 2021-02-15
ES Modules broken on npm@7


richardsimko:
# Environment

Knex version: 0.21.17
Database + version: N/A
OS: N/A

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do

When setting `type: 'module'` in `package.json` knex fails to use ESM to load the migration files on `npm >= 7.0.0`. This works on lower versions. The reason being that npm has removed the environment variable which knex uses to determine if it should use ESM or CommonJS (https://github.com/knex/knex/blob/b20a31fe5d0256234e1542c1ec4cbe2082f9970e/lib/migrations/util/import-file.js#L2). See this NPM RFC for more details https://github.com/npm/rfcs/blob/a044b355af60310b05ad1995fdc1cf672243a4e0/implemented/0021-reduce-lifecycle-script-environment.md. If the file is named `knexfile.mjs` it works as expected since the filename check triggers the usage of ESM.

2. Error message

```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /project-root/config/knexfile.js
require() of ES modules is not supported.
require() of /project-root/node_modules/knex/lib/util/import-file.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename knexfile.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /project-root/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1080:13)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at importFile (/project-root/node_modules/knex/lib/util/import-file.js:13:7)
    at openKnexfile (/project-root/node_modules/knex/bin/cli.js:26:22)
    at initKnex (/project-root/node_modules/knex/bin/cli.js:47:13)
    at Command.<anonymous> (/project-root/node_modules/knex/bin/cli.js:170:32)
    at Command.listener [as _actionHandler] (/project-root/node_modules/commander/index.js:426:31)
```

3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real
   database connection to MySQL or PostgreSQL, then single file example which initializes
   needed data and demonstrates the problem.
   
[...]

PR #4296, opened 2021-02-15
Fix for ES Module detection using npm@7 (#4295)

richardsimko:


lib/migrations/util/is-module-type.js
```diff
@@ -0,0 +1,16 @@
...
+    try {
+        if (process.env.npm_package_json) {
+            // npm >= 7.0.0
+            const packageJson = JSON.parse(await readFile(process.env.npm_package_json, 'utf-8'));
+            if (packageJson.type === 'module') {
+                return true;
+            }
+        }
+    } catch (e) {
+        console.warn('Error reading package.json', e);
```
kibertoad, line 13: In which scenarios do you expect this logic to fail?

richardsimko: Now that I think about it it should not be something that can occur, as `npm` would refuse to run if `package.json` had an invalid syntax.

kibertoad:
@richardsimko Bunch of previous tests seem to be failing, see https://github.com/knex/knex/pull/4296/checks?check_run_id=1904682577

richardsimko:
Yeah I don't think my approach to mocking the env vars was good. I'll try another way but if that doesn't work I'm not sure it's possible to add tests for this code in a good way.

kibertoad:
@richardsimko Unit tests are fixed, but it looks like something in migration loading logic might be broken, judging by non-mysql integration tests.

richardsimko:
I think what's most likely happening is that I've uncovered latent issues where `importFile` was not used with `await`. `import(filePath)` is an async function and always was - but this function was only used in the ESM case which doesn't seem to be commonly used. Now `importFile` is always async and this is how that issue manifests. Fixing this would require a lot more in-depth knowledge of knex than what I have but if you have any pointers as to where this might be an issue I'm all ears.

richardsimko:
`importFile` seems to be used here https://github.com/richardsimko/knex/blob/2cc3247d83e17cfb7d5788f6b391ab31faa4758b/lib/migrations/migrate/sources/fs-migrations.js#L84:L84

Which in turn is used in many places without `await`:
https://github.com/knex/knex/blob/2cc3247d83e17cfb7d5788f6b391ab31faa4758b/lib/migrations/migrate/Migrator.js#L493:L493
https://github.com/knex/knex/blob/2cc3247d83e17cfb7d5788f6b391ab31faa4758b/lib/migrations/migrate/Migrator.js#L145:L145
https://github.com/knex/knex/blob/2cc3247d83e17cfb7d5788f6b391ab31faa4758b/lib/migrations/migrate/Migrator.js#L86:L86

kibertoad:
@richardsimko Yeah, noticed same thing: https://github.com/knex/knex/blob/b20a31fe5d0256234e1542c1ec4cbe2082f9970e/lib/migrations/migrate/sources/fs-migrations.js#L84 seems to be the only place where we importFile without awaiting, so I would review all cases where `getMigration` is used to ensure we await properly.

richardsimko:
I can slap `await`s on all the above mentioned locations but in my experience fixing these types of issues is never that easy 😅

kibertoad:
@richardsimko You might need to make few more functions async (e. g. `waterfallBatch`), but that should do it. We have pretty good coverage for migrations at this point, if something breaks down, you'll know immediately.

lib/migrations/migrate/Migrator.js
```diff
@@ -80,12 +80,16 @@ class Migrator {
...
+      !(
+        await Promise.all(
+          migrations.map(async (migration) => {
+            const migrationContents = await this.config.migrationSource.getMigration(
+              migration
+            );
+            return !this._useTransaction(migrationContents);
+          })
+        )
+      ).some((e) => e);
```
kibertoad, line 92: Maybe a slightly more descriptive name could be used? from my understanding the value you'll have there is `isTransactionUsed` or something along these lines

richardsimko:
It worked 🎉  I added a note in the changelog as well. Do you want me to squash the commits?

kibertoad:
@richardsimko Thanks a lot! 
Squashing is not worth the effort, I'll squash on merge anyway.

kibertoad (approved):

richardsimko:
@kibertoad Is there anything I have to do in order to get this backported to 0.21.x?

kibertoad:
@richardsimko There is a 0.21 branch, just open a PR targetting that.

richardsimko:
Thanks, I opened #4308

PR #4300, opened 2021-02-17
Include 'name' property in MigratorConfig

nath:
PR #3416 added the ability to specify a filename for the up and down commands, but that parameter was left out of the MigratorConfig interface, causing compilation errors for those using that property and typescript.


kibertoad (approved):

kibertoad:
Thanks!

PR #4301, opened 2021-02-18
events: introduce queryContext on query-error

tylerjwatson:
This patch introduces support for submitting the `queryContext` on `query-error` events.  

Because of the async nature of some drivers, the root callsite of responses is usually `TCP._onRead`.  Query builders created in the context of an Express request or another mechanism get lost in error events attached to the query builders.  Being able to access the `queryContext` in case of an error is a low-overhead way of relating a query error back to something more high level, enabling meaningful error reporting.

The patch is covered by a unit test.

### DIFF
--- a/lib/execution/runner.js
+++ b/lib/execution/runner.js
@@ -122,6 +122,7 @@ class Runner {
 
     const runner = this;
     let queryPromise = this.client.query(this.connection, obj);
+    const queryContext = this.builder.queryContext();
 
     if (obj.timeout) {
       queryPromise = timeout(queryPromise, obj.timeout);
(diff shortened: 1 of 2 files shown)


kibertoad (approved):

kibertoad:
Thanks! I'll prepare a new `next` version shortly.

kibertoad:
@tylerjwatson OK, 0.95.0-next3 is out and should include all the new features from your branch. Note that there were plenty of breaking changes in this release, see https://github.com/knex/knex/blob/master/UPGRADING.md for the migration guide.
Please let me know how the testing will go!

tylerjwatson:
@kibertoad We have just completed preliminary work to update our stack using `knex@next`, returning upstream.  Most of the work was putting the stack into k8s, and now I have a team of guys performing a full regression test, we should be done in a little while but so far it's looking all green!

I'll let you know soon as I have more information.

kibertoad:
@tylerjwatson Great news!

kibertoad:
@tylerjwatson 0.95.0 was released, please give it a try!

PR #4308, opened 2021-02-22
Fixed #4295

### LINKED ISSUE #4295: ES Modules broken on npm@7

richardsimko:


kibertoad (approved):

kibertoad:
Thank you!

lib/migrate/Migrator.js
```diff
@@ -80,12 +80,16 @@ class Migrator {
...
+      !(
+        await Promise.all(
+          migrations.map(async (migration) => {
+            const migrationContents = await this.config.migrationSource.getMigration(
+              migration
+            );
+            return !this._useTransaction(migrationContents);
+          })
+        )
+      ).some((isTransactionUsed) => isTransactionUsed);
```
falkenhawk, line 92: were the changes in `Migrator` intended in this PR?

falkenhawk: Okay I can see more details in https://github.com/knex/knex/pull/4296

kibertoad: @falkenhawk Did that cause some breaking changes for you?

falkenhawk: I can't tell yet, I noticed this while I was looking through the changes after I got notified by renovatebot about the new knex version 0.21.18... and not having this explicitly listed in changelog confused me and stopped me from upgrading, as it seemed to me to be something extra than just a fix for #4296 - and I could not afford breaking things at that moment 😅

If I decide to upgrade in the following days and will run into some issues, I will let you know then.

kibertoad: It should behave same way for the end user in most cases, and also handle ESM case better, see https://github.com/knex/knex/pull/4296#issuecomment-779330504 for the reason of more methods becoming async. If you are not using ESM, this upgrade won't do anything useful for you.

PR #4313, opened 2021-02-24
Respect KNEX_TEST, support omitting sqlite3 from DB, and reduce outside mssql test db config

jeremy-w:
This PR contains several changes that simplify getting a green test run in a fresh checkout by eliminating some unexpected behavior around KNEX_TEST and DB config.

Before, if DB omitted sqlite3, several of the CLI tests would fail, because only some of them conditionally skipped when sqlite3 config was not available. Now, all of those tests skip when sqlite3 tests are not selected.

Before, some of the DB integration tests did not use the config supplied in KNEX_TEST. Now, all run by test:db do.

Before, the mssql transaction isolation tests required, not just a test DB, but also a test DB with snapshot isolation already enabled. This led to test failures due to having a test DB without that config already set up. Now, that test suite ensures snapshot isolation is enabled upfront before running its tests.

There is no documentation PR; the main intent of this PR is to align test suite behavior with the existing documentation around how the DB and KNEX_TEST environment variables affect test running.

### DIFF
--- a/test/integration2/transaction/set-isolation-level.spec.js
+++ b/test/integration2/transaction/set-isolation-level.spec.js
@@ -10,6 +10,14 @@ describe('Transaction', () => {
         const tableName = 'key_value';
         before(() => {
           knex = getKnexForDb(db);
+
+          if (isMssql(knex)) {
+            // Enable the snapshot isolation level required by certain transaction tests.
+            return knex.raw(
+              `ALTER DATABASE :db: SET ALLOW_SNAPSHOT_ISOLATION ON`,
+              { db: knex.context.client.config.connection.database }
+            );
+          }
         });
 
         after(() => {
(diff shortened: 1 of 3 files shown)


jeremy-w:
Spurious test failure: db:start on Travis failed due to:

> ERROR: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

kibertoad:
@jeremy-w Yeah, ignore the Travis part, it is extremely flaky and probably will stop working altogether at some point.

test/integration2/transaction/set-isolation-level.spec.js
```diff
@@ -10,6 +10,14 @@ describe('Transaction', () => {
         const tableName = 'key_value';
         before(() => {
           knex = getKnexForDb(db);
+
+          if (isMssql(knex)) {
```
kibertoad, line 14: Do we still need the Docker config part then? I think it was originally added precisely because it was not clear how to do that on per-test basis.

kibertoad (approved):

kibertoad:
thanks!

PR #4314, opened 2021-02-25
mssql: don't raise query-error twice

tylerjwatson:
This patch slightly reworks how the query context gets passed from the runner into the dialects.  The query object now has a `context` parameter with the query context to be raised on `query-error` events, this avoids having to re-bubble `query-error` events from the builder instance in the `mssql` dialect.  Test surface has not changed as it gets the same outcome.

### DIFF
--- a/lib/dialects/mssql/index.js
+++ b/lib/dialects/mssql/index.js
@@ -113,10 +113,6 @@ class Client_MSSQL extends Client {
 
   queryBuilder() {
     const b = new QueryBuilder(this);
-    const base = this;
-    b.on('query-error', function () {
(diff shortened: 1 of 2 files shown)


kibertoad (approved):

kibertoad:
@tylerjwatson Do you need next-4 with this change?

kibertoad:
@tylerjwatson How's the testing going? Are we good to release?

tylerjwatson:
@kibertoad no fundamental issues from my side although I cannot test native error reporting because this PR has not been released to npm as far as I can tell.

PR #4317, opened 2021-02-25
fix mssql alter column must have its own query

### LINKED ISSUE #4316: mssql: schema-builder - mutliple altered columns results in syntax error
# Environment

Knex version: 0.21.16
Database + version: mssql azure sql
OS: linux

If issue is about oracledb support, tag @ atiertant. For MSSql tag @smorey2.
If issue is about TypeScript definitions, tag @ lorefnon.

# Bug

When attempting to alter multiple columns at once using the schema builder invalid sql is generated causing an error.

eg:

```js
knex.schema.alterTable('test', (table) => {
  table.integer('col1').notNullable().alter();
  table.integer('col2').notNullable().alter();
});
```

Will generate sql:

```tsql
ALTER TABLE [test] alter column [col1] int not null, [col2] int not null
```

However this is not valid and should produce two distinct queries:

```tsql
ALTER TABLE [test] alter column [col1] int not null;
ALTER TABLE [test] alter column [col2] int not null;
```

See https://stackoverflow.com/a/3465574

2. Error message

```
RequestError: ALTER TABLE [test] alter column [col1] int not null, [col2] int not null - Incorrect syntax near ','
```

```js
exports.up = async (knex) => {
    await knex.schema.createTable('test', (table) => {
        table.increments('id').primary();
        table.integer('col1').nullable();
        table.integer('col2').nullable();
    });
    await knex.schema.alterTable('test', (table) => {
        table.integer('col1').notNullable().alter();
        table.integer('col2').notNullable().alter();
    });
};
```

dhensby:
fixes #4316

As a side effect of my change, the `alter column` casing now respects the `lowerCase` flag on all alter commands

### DIFF
--- a/lib/dialects/mssql/schema/mssql-tablecompiler.js
+++ b/lib/dialects/mssql/schema/mssql-tablecompiler.js
@@ -52,6 +52,21 @@ class TableCompiler_MSSQL extends TableCompiler {
     }
   }
 
+  alterColumns(columns, colBuilder) {
+    // in SQL server only one column can be altered at a time
+    columns.sql.forEach((sql) => {
+      this.pushQuery({
+        sql:
+          (this.lowerCase ? 'alter table ' : 'ALTER TABLE ') +
+          this.tableName() +
+          ' ' +
+          (this.lowerCase ? this.alterColumnPrefix.toLowerCase() : this.alterColumnPrefix) +
+          sql,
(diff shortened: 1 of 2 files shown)


kibertoad (approved):

kibertoad:
Thanks!

PR #4252, opened 2021-01-26
Added merge strategy to allow selecting columns to upsert.

MattGson:
Fixes #4209 
Maybe fixes #4106

New syntax on merge:

```
knex('upsert_tests')
          .insert(
            { email: 'mergedest@example.com', name: 'SHOULD BE USED', age: 12 },
          )
          .onConflict('email')
          .merge(['name', 'age'])
```
Will update only the name and age fields.
Also works for multi-inserts.

PR #4318, opened 2021-02-25
Improve join and conflict types v2

kibertoad:
Rebased version of #4256

With TypeScript 4.1, literal types can be constructed. This enables better type checking where it wasn't possible before.

What was done:

* Increase minimum TypeScript version to 4.1 to support template literal types
* Type join arguments
* Type onConflict (+ merge and ignore)


types/index.d.ts
```diff
@@ -467,7 +467,7 @@ export declare namespace Knex {
   interface OnConflictQueryBuilder<TRecord, TResult> {
     ignore(): QueryBuilder<TRecord, TResult>;
     merge(mergeColumns?: (keyof TRecord)[]): QueryBuilder<TRecord, TResult>;
-    merge(data?: Extract<DbRecord<TRecord>, object>): QueryBuilder<TRecord, TResult>;
+    merge(data?: Extract<DbRecord<ResolveTableType<TRecord, 'update'>>, object>): QueryBuilder<TRecord, TResult>;
```
kibertoad, line 470: @hermansje @lorefnon Original #4256 didn't have `Extract<>` here, which was added later in #4252.
Is it still needed here or it's safe to drop it?

PR #4319, opened 2021-02-26
mssql: schema builder - add predictable constraint names for default values

dhensby:
In SQL server, default values are stored as `CONSTRAINTS` against the table. if no constraint name is provided and just the short syntax of `DEFAULT xxx` is used, then SQL server creates a "random" constraint name. This makes writing predictable schema migrations tricky as the `DEFAULT` constraint needs to be dropped before it can be updated/changed.

The constraint names follow the format `${tableName}_${columnName}_${default}` (lowercased) which follows the convention of other constraints like default index names.

Usage:

```js
knex.schema.alterTable('table', (table) => {
  // new behaviour - will produce a default constraint named table_myint_default
  table.integer('myInt').defaultTo(1);
  // create a custom named constraint
  table.integer('myInt').defaultTo(1, 'DF_table_myInt');
  // no constraint name (SQL generates a random one, current behaviour
  table.integer('myInt').defaultTo(1, '');
});
```

Potential issues:

This *might* be considered a breaking change, but I don't think it is. The produced SQL is different but the end effect to users is the same. A constraint name they don't know the name of needs to be looked up before being able to remove it, this is typically done with a query or stored procedure. Even if users with this library rolled back migrations and then ran them again, their logic would still work as the constraint name would never have been known (due to it's random nature at the moment).

To do / help wanted:

- [x] Update docs around use of this new second parameter in `.defaultTo(value, constraintName)` for mssql
- [x] Add mssql specific test in integration suite covering custom named constraints / unnamed constraints
- [x] Update typings??

### DIFF
--- a/lib/dialects/mssql/schema/mssql-columncompiler.js
+++ b/lib/dialects/mssql/schema/mssql-columncompiler.js
@@ -2,6 +2,7 @@
 // -------
 const ColumnCompiler = require('../../../schema/columncompiler');
 const { toNumber } = require('../../../util/helpers');
+const { formatDefault } = require('../../../formatter/formatterUtils');
 
 class ColumnCompiler_MSSQL extends ColumnCompiler {
   constructor(client, tableCompiler, columnBuilder) {
@@ -61,6 +62,22 @@ class ColumnCompiler_MSSQL extends ColumnCompiler {
   after(column) {
     this.client.logger.warn('Column after modifier not available for MSSQL');
     return '';
+  }
+
+  defaultTo(value, { constraintName } = {}) {
+    const formatedValue = formatDefault(value, this.type, this.client);
+    constraintName =
+      typeof constraintName !== 'undefined'
+        ? constraintName
+        : `${
+            this.tableCompiler.tableNameRaw
+          }_${this.getColumnName()}_default`.toLowerCase();
+    if (!constraintName) {
+      return `DEFAULT ${formatedValue}`;
+    }
+    return `CONSTRAINT ${this.formatter.wrap(
+      constraintName
+    )} DEFAULT ${formatedValue}`;
   }
 
   comment(comment) {
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1833,7 +1833,7 @@ export declare namespace Knex {
     references(columnName: string): ReferencingColumnBuilder;
     onDelete(command: string): ColumnBuilder;
     onUpdate(command: string): ColumnBuilder;
-    defaultTo(value: Value | null): ColumnBuilder;
+    defaultTo(value: Value | null, options?: DefaultToOptions): ColumnBuilder;
     unsigned(): ColumnBuilder;
     notNullable(): ColumnBuilder;
     nullable(): ColumnBuilder;
@@ -2307,6 +2307,11 @@ interface MsSqlConnectionConfigBase {
     existingType?: boolean;
     schemaName?: string;
     enumName: string;
+  }
+
+  interface DefaultToOptions {
+    // only supported by mssql driver
+    constraintName?: string;
   }
 
   //
--- a/test/integration/schema/index.js
+++ b/test/integration/schema/index.js
@@ -473,7 +473,7 @@ module.exports = (knex) => {
               'create index "test_table_one_logins_index" on "test_table_one" ("logins")',
             ]);
             tester('mssql', [
-              "CREATE TABLE [test_table_one] ([id] bigint identity(1,1) not null primary key, [first_name] nvarchar(255), [last_name] nvarchar(255), [email] nvarchar(255) null, [logins] int default '1', [balance] float default '0', [about] nvarchar(max), [created_at] datetime2, [updated_at] datetime2)",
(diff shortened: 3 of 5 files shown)


kibertoad:
@dhensby Could you also add some integration tests to make sure this logic actually works as expected? `test/integration2/dialects` is probably the right place for them.

dhensby:
Looks like I've got my work cut out to get the existing ones passing! 😅

kibertoad:
@dhensby Fair :D. Please let me know if you'll need any help with running tests locally! `db:start:no-oracle` should cover launching mssql if you have Docker configured.

dhensby:
@kibertoad ok - I think it makes sense to put these changes across 2 or 3 PRs so they can be tested/reviewed in isolation.

I've put this back to the original intention of just producing predictable constraint names on columns as well as allowing developer override of them.

lib/dialects/mssql/schema/mssql-columncompiler.js
```diff
@@ -63,6 +65,36 @@ class ColumnCompiler_MSSQL extends ColumnCompiler {
     return '';
   }
 
+  defaultTo(value, constraintName) {
+    if (value === void 0) {
+      return '';
+    } else if (value === null) {
+      value = 'null';
+    } else if (value instanceof Raw) {
```
kibertoad, line 73: There is no need to import raw here, 
```
if (value && value.isRawInstance) {
```
should suffice

lib/dialects/mssql/schema/mssql-columncompiler.js
```diff
@@ -63,6 +65,36 @@ class ColumnCompiler_MSSQL extends ColumnCompiler {
     return '';
   }
 
+  defaultTo(value, constraintName) {
```
kibertoad, line 68: Would be better to add second parameter to all dialects (or at least major ones, such as MySQL and PostgreSQL), so that it works consistently.

lib/dialects/mssql/schema/mssql-columncompiler.js
nickrum: Or use `formatDefault()` from `lib/formatter/formatterUtils.js`

kibertoad: You'd need to pull latest changes from master for that, though

dhensby: I've just lifted this from the base definition. Happy to change this but should it be changed in the base too?

lib/dialects/mssql/schema/mssql-columncompiler.js
dhensby: Do the other dialects name constraints in a similar way? I'd be at a bit of a loss to add it other than in the signature definition of the function

lib/dialects/mssql/schema/mssql-columncompiler.js
kibertoad: Pull from the master, base was adjusted to use `formatDefault` :)

lib/dialects/mssql/schema/mssql-columncompiler.js
kibertoad: I did a quick check, looks like DEFAULT constraint name is pretty much a MSSQL thing, neither MySQL nor PostgreSQL seem to have it. From this perspective it seems like it would be better to have second parameter as an options object with named properties, `constraintName` being supported only for MSSQL. Otherwise when we'll need to add one more option, it'll be an issue.

lib/dialects/mssql/schema/mssql-columncompiler.js
dhensby: Ah, not sure why I wasn't up to date, I've updated the PR to use `formatDefault` now

lib/dialects/mssql/schema/mssql-columncompiler.js
dhensby: Good point, I've updated the signature to expect an object as the second parameter

lib/dialects/mssql/schema/mssql-columncompiler.js
```diff
@@ -63,6 +64,22 @@ class ColumnCompiler_MSSQL extends ColumnCompiler {
     return '';
   }
 
+  defaultTo(value, { constraintName }) {
```
kibertoad, line 67: This needs to be slightly more defensive about argument not being passed, so we probably can't destructure it always

dhensby: This is the problem with making changes at midnight!

I've updated this to provide a default value for the argument; is that acceptable? I've also updated the typings

types/index.d.ts
```diff
@@ -2309,6 +2309,10 @@ interface MsSqlConnectionConfigBase {
     enumName: string;
   }
 
+  interface DefaultToOptions {
+    constraintName?: string;
```
kibertoad, line 2313: adding comment explaining that this only works for mssql might be helpful

lib/dialects/mssql/schema/mssql-columncompiler.js
kibertoad: yup, this looks good! Do you need any help with writing a test?

types/index.d.ts
dhensby: done

kibertoad:
Please ignore the failing flaky SQLite test. This is green and good to go, only documentation update at https://github.com/knex/documentation is missing now.

kibertoad (approved):

kibertoad:
Thank you!

kibertoad:
Released in 0.95.0

PR #4321, opened 2021-02-26
mssql: schema builder - attempt to drop default constraints when changing default value on columns

dhensby:
Depends on #4319 

When changing a default value on a column in mssql, the previous default constraint needs to be dropped before a new one can be assigned. This PR attempts to drop the constraint and then adds the constraint via a separate query:

```tsql
DROP CONSTRAINT [constraint_name]; //does some look up to find the constraint name
ALTER TABLE [table] ALTER COLUMN ...; // other altering of column
ALTER TABLE [table] ADD CONSTRAINT [constraint_name] DEFAULT [defaultVal] FOR [column_name]; // applies default constraint
```

Help required:

I've tried to get a grasp of how the internals of the builder work for this and  implemented this in a way that I think makes the most sense.

Under some circumstances this change can result in a needless alter query being run when the only change to a column is its default value; in this case we should only be running the drop and add constraint queries, but I'm not clear on how best to skip the other query that is produced under these circumstances.

To do:

- [x] Create integration tests to cover this feature specifically

### DIFF
--- a/lib/dialects/mssql/schema/mssql-columncompiler.js
+++ b/lib/dialects/mssql/schema/mssql-columncompiler.js
@@ -65,7 +65,7 @@ class ColumnCompiler_MSSQL extends ColumnCompiler {
   }
 
   defaultTo(value, { constraintName } = {}) {
-    const formatedValue = formatDefault(value, this.type, this.client);
+    const formattedValue = formatDefault(value, this.type, this.client);
     constraintName =
       typeof constraintName !== 'undefined'
         ? constraintName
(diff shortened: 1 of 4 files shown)


kibertoad:
```
      -ALTER TABLE [users] ADD CONSTRAINT [users_foo_default] DEFAULT test FOR [foo]
      +ALTER TABLE [users] ADD CONSTRAINT [users_foo_default] DEFAULT 'test' FOR [foo]
```

Looks like escaping changed.

kibertoad (approved):

kibertoad:
Thanks!

kibertoad:
Released in 0.95.0

PR #4322, opened 2021-02-26
Add support for altering columns to SQLite

nickrum:
The approach I took to parse `CREATE INDEX` statements using regexes was unfortunately not flexible and robust enough to properly parse `CREATE TABLE` statements. I replaced it with a tokenizer and parser combinators. The grammar is based on the SQLite documentation.

Any idea where the `formatDefault()` function would fit? It is currently directly inside the `ColumnCompiler` class which doesn't feel right.

### DIFF
--- a/lib/dialects/sqlite3/schema/internal/parser-combinator.js
+++ b/lib/dialects/sqlite3/schema/internal/parser-combinator.js
@@ -1,0 +1,161 @@
+// Sequence parser combinator
+function s(sequence, post = (v) => v) {
+  return function ({ index = 0, input }) {
+    let position = index;
+    const ast = [];
+
+    for (const parser of sequence) {
+      const result = parser({ index: position, input });
+
+      if (result.success) {
+        position = result.index;
+        ast.push(result.ast);
+      } else {
+        return result;
+      }
+    }
+
+    return { success: true, ast: post(ast), index: position, input };
+  };
+}
+
+// Alternative parser combinator
+function a(alternative, post = (v) => v) {
+  return function ({ index = 0, input }) {
+    for (const parser of alternative) {
+      const result = parser({ index, input });
+
+      if (result.success) {
+        return {
+          success: true,
+          ast: post(result.ast),
+          index: result.index,
+          input,
+        };
+      }
+    }
+
+    return { success: false, ast: null, index, input };
+  };
+}
+
+// Many parser combinator
+function m(many, post = (v) => v) {
+  return function ({ index = 0, input }) {
+    let result = {};
+    let position = index;
+    const ast = [];
+
+    do {
+      result = many({ index: position, input });
+
+      if (result.success) {
+        position = result.index;
+        ast.push(result.ast);
+      }
+    } while (result.success);
+
+    if (ast.length > 0) {
+      return { success: true, ast: post(ast), index: position, input };
+    } else {
+      return { success: false, ast: null, index: position, input };
+    }
+  };
+}
+
+// Optional parser combinator
+function o(optional, post = (v) => v) {
+  return function ({ index = 0, input }) {
+    const result = optional({ index, input });
+
+    if (result.success) {
+      return {
+        success: true,
+        ast: post(result.ast),
+        index: result.index,
+        input,
+      };
+    } else {
+      return { success: true, ast: post(null), index, input };
+    }
+  };
+}
+
+// Lookahead parser combinator
+function l(lookahead, post = (v) => v) {
+  return function ({ index = 0, input }) {
+    const result = lookahead.do({ index, input });
+
+    if (result.success) {
+      const resultNext = lookahead.next({ index: result.index, input });
+
+      if (resultNext.success) {
+        return {
+          success: true,
+          ast: post(result.ast),
+          index: result.index,
+          input,
+        };
+      }
+    }
+
+    return { success: false, ast: null, index, input };
+  };
+}
+
+// Negative parser combinator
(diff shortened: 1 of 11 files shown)


lib/schema/columncompiler.js
```diff
@@ -97,6 +97,26 @@ class ColumnCompiler {
     return modifiers.length > 0 ? ` ${modifiers.join(' ')}` : '';
   }
 
+  formatDefault(value) {
```
kibertoad, line 100: What about lib/formatter/formatterUtils.js or somewhere around that?

lib/dialects/sqlite3/schema/internal/parser-combinator.js
```diff
@@ -0,0 +1,161 @@
+// Sequence parser combinator
+function s(sequence, post = (v) => v) {
```
kibertoad, line 2: Can `s` and `v` be renamed to something more informative?

lib/dialects/sqlite3/schema/internal/parser-combinator.js
```diff
@@ -0,0 +1,161 @@
...
+    if (ast.length > 0) {
+      return { success: true, ast: post(ast), index: position, input };
+    } else {
+      return { success: false, ast: null, index: position, input };
+    }
+  };
+}
+
+// Optional parser combinator
+function o(optional, post = (v) => v) {
```
kibertoad, line 67: Can this also be renamed from `o`?

lib/dialects/sqlite3/schema/internal/parser-combinator.js
```diff
@@ -0,0 +1,161 @@
...
+const e = function ({ index = 0, input }) {
+  return { success: true, ast: null, index, input };
+};
+
+// Finish parser constant
+const f = function ({ index = 0, input }) {
+  return { success: index === input.length, ast: null, index, input };
+};
+
+module.exports = { s, a, m, o, l, n, t, e, f };
```
kibertoad, line 161: same comment about all single-letter methods, they are difficult to understand at-a-glance. is there a reason for such naming?

lib/dialects/sqlite3/schema/internal/tokenizer.js
```diff
@@ -0,0 +1,38 @@
...
+  let index = 0;
+  const ast = [];
+
+  while (index < text.length) {
+    compiledRegex.lastIndex = index;
+    const result = text.match(compiledRegex);
+
+    if (result !== null) {
+      const [type, text] = Object.entries(result.groups).find(
+        ([_, t]) => t !== undefined
```
kibertoad, line 18: is `t` `text` or `token` here?

lib/schema/columncompiler.js
nickrum: Sounds good!

lib/dialects/sqlite3/schema/internal/parser-combinator.js
nickrum: Yes, I deliberately chose single letter names because these are more like "operators" rather than "functions". Also they are heavily used within the grammar and it felt like the grammar is easier to comprehend if there is less overhead from function names. But I can of course rename those if you so desire.

lib/dialects/sqlite3/schema/internal/tokenizer.js
nickrum: `t` is `text` in this case, analogous to `[type, text]`. A token is the whole thing consisting of a type and the text.

lib/dialects/sqlite3/schema/internal/parser-combinator.js
kibertoad: Makes sense, let's keep them as-is

lib/dialects/sqlite3/schema/internal/tokenizer.js
kibertoad: then let's rename this one to `textEntry` or something along these lines, so that it is less confusing

nickrum: I renamed the destructured array to `[name, group]` as those entries are essentially regex named capture groups and their content.

kibertoad:
Unit tests seem to be failing now.

nickrum:
Very strange... This seems to be caused by requiring `Raw` inside `formatterUtils`. Any idea what the actual cause of this is and how to fix it?

kibertoad:
@nickrum I would imagine a circular dependency happening.

lib/formatter/formatterUtils.js
```diff
@@ -16,7 +19,25 @@ function wrapAsIdentifier(value, builder, client) {
   return client.wrapIdentifier((value || '').trim(), queryContext);
 }
 
+function formatDefault(value, type, client) {
+  if (value === void 0) {
+    return '';
+  } else if (value === null) {
+    return 'null';
+  } else if (value instanceof Raw) {
```
kibertoad, line 27: You can use
```
if (value && value.isRawInstance) {
```
for doing the check without requiring Raw.

nickrum:
@kibertoad Good catch, `Raw` is indeed circularly depended upon.

lib/formatter/formatterUtils.js
nickrum: Done, thanks!

kibertoad (approved):

kibertoad:
Thanks!

PR #4324, opened 2021-02-28
Fix timeout method

### LINKED ISSUE #2211: `.timeout(ms, {cancel: true})` not working for transactions
This issue applies only to mysql and mariadb as `.timeout(ms, {cancel: true})` is implemented for them only. Tested with knex 0.13.0

If **in a transaction**, `client.acquireConnection` acquires always the same connection. Therefore, [`const acquiringConn = this.acquireConnection()`](https://github.com/tgriesser/knex/blob/8f7d2e297799d4b70de726b8cfc9d4939fe3923b/src/dialects/mysql/index.js#L148) in `cancelQuery` returns the same connection as is being canceled (`connectionToKill.threadId` and `acquiringConn.threadId` are the same).

As the result the query `KILL QUERY ?` (with `?` bound to `connectionToKill.threadId`) is executed only **after** the original query finishes, as they both use the same connection. The original query is not killed at all, and only after it finishes the kill query just kills itself, as it has the same `threadId`.

**How to reproduce the bug**
```
  return trx.select(trx.raw('SLEEP(47)')).timeout(5000, {cancel: true})
}).then(function(res){
  console.log('OK', res)
}).catch(function(err){
  console.log('ERROR', err)
})
```

*Result*: The query runs for 47 seconds (check with `SHOW PROCESSLIST`) and then (after the 47 seconds) throws error: "Error: After query timeout of 5000ms exceeded, cancelling of query failed."

*Expected result*: The query should be killed after 5 seconds and should throw error: "TimeoutError: Defined query timeout of 5000ms exceeded when running query."

If the query is not executed in a transaction (e.g. `knex.select(knex.raw('SLEEP(47)')).timeout(5000, {cancel: true})`) it is killed as expected.

**Possible fix**

Change [`const acquiringConn = this.acquireConnection()`](https://github.com/tgriesser/knex/blob/8f7d2e297799d4b70de726b8cfc9d4939fe3923b/src/dialects/mysql/index.js#L148) in `cancelQuery` to always get a new connection, even if in a transaction.

martinmacko47:
Based on PR https://github.com/knex/knex/pull/2527 - rebased and added the fix for postgres too:

> KILL commands have to be sent on a seprate connection, and transaction Clients override acquireConnection to always return the same connection.
>
> Furthermore, it is plausible that some applications may never be able to Acquire a pooled connection for cancellation becuase a query which is Exceeding the timeout may coincidentally exhaust the connection pool.
>
> For example, if a DDL operation is blocking writes, a connection pool May be quickly exhaused by write operations.

Timeout cancellation during transaction is fixed by using `acquireRawConnection` and `_query` to bypass the transaction connection and cancel the query from a separate connection.

fixes #2211

### DIFF
--- a/lib/dialects/mysql/index.js
+++ b/lib/dialects/mysql/index.js
@@ -165,7 +165,6 @@ class Client_MySQL extends Client {
     const conn = await this.acquireRawConnection();
     try {
       return await this._query(conn, {
-        method: 'raw',
         sql: 'KILL QUERY ?',
         bindings: [connectionToKill.threadId],
         options: {},
--- a/lib/dialects/mysql/index.js
+++ b/lib/dialects/mysql/index.js
@@ -162,16 +162,16 @@ class Client_MySQL extends Client {
   }
 
   async cancelQuery(connectionToKill) {
-    const conn = await this.acquireConnection();
+    const conn = await this.acquireRawConnection();
     try {
-      return await this.query(conn, {
+      return await this._query(conn, {
         method: 'raw',
         sql: 'KILL QUERY ?',
         bindings: [connectionToKill.threadId],
         options: {},
       });
     } finally {
-      await this.releaseConnection(conn);
+      await this.destroyRawConnection(conn);
     }
   }
 }
--- a/lib/dialects/mysql/index.js
+++ b/lib/dialects/mysql/index.js
@@ -171,6 +171,9 @@ class Client_MySQL extends Client {
       });
     } finally {
       await this.destroyRawConnection(conn);
+      if (conn.__knex__disposed) {
+        this.logger.warn(`Connection Error: ${conn.__knex__disposed}`);
+      }
     }
   }
 }
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -242,7 +242,9 @@ class Client_PG extends Client {
     try {
       return await this._wrappedCancelQueryCall(conn, connectionToKill);
     } finally {
-      await this.destroyRawConnection(conn);
+      await this.destroyRawConnection(conn).catch((err) => {
+        this.logger.warn(`Connection Error: ${err}`);
+      });
     }
   }
   _wrappedCancelQueryCall(conn, connectionToKill) {
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -241,20 +241,20 @@ class Client_PG extends Client {
     // Purposely not putting timeout on `pg_cancel_backend` execution because erroring
     // early there would release the `connectionToKill` back to the pool with
     // a `KILL QUERY` command yet to finish.
-    const conn = await this.acquireConnection();
+    const conn = await this.acquireRawConnection();
 
     try {
       return await this._wrappedCancelQueryCall(conn, connectionToKill);
     } finally {
       // NOT returning this promise because we want to release the connection
       // in a non-blocking fashion
-      this.releaseConnection(conn);
+      this.destroyRawConnection(conn);
     }
   }
   _wrappedCancelQueryCall(conn, connectionToKill) {
-    return this.query(conn, {
+    return this._query(conn, {
       method: 'raw',
-      sql: 'SELECT pg_cancel_backend(?);',
+      sql: 'SELECT pg_cancel_backend($1);',
       bindings: [connectionToKill.processID],
       options: {},
     });
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -237,18 +237,12 @@ class Client_PG extends Client {
   }
 
   async cancelQuery(connectionToKill) {
-    // Error out if we can't acquire connection in time.
-    // Purposely not putting timeout on `pg_cancel_backend` execution because erroring
-    // early there would release the `connectionToKill` back to the pool with
-    // a `KILL QUERY` command yet to finish.
     const conn = await this.acquireRawConnection();
 
     try {
       return await this._wrappedCancelQueryCall(conn, connectionToKill);
     } finally {
-      // NOT returning this promise because we want to release the connection
-      // in a non-blocking fashion
-      this.destroyRawConnection(conn);
+      await this.destroyRawConnection(conn);
     }
   }
   _wrappedCancelQueryCall(conn, connectionToKill) {
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -253,7 +253,6 @@ class Client_PG extends Client {
   }
   _wrappedCancelQueryCall(conn, connectionToKill) {
     return this._query(conn, {
-      method: 'raw',
       sql: 'SELECT pg_cancel_backend($1);',
       bindings: [connectionToKill.processID],
       options: {},
--- a/test/integration/query/additional.js
+++ b/test/integration/query/additional.js
@@ -980,7 +980,7 @@ module.exports = function (knex) {
         });
     });
 
-    it.skip('.timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection', async function () {
+    it('.timeout(ms, {cancel: true}) should cancel slow query even if connection pool is exhausted', async function () {
       // Only mysql/postgres query cancelling supported for now
       if (!isMysql(knex) && !isPostgreSQL(knex)) {
         return this.skip();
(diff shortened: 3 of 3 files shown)


test/integration/query/additional.js
```diff
@@ -869,7 +869,118 @@ module.exports = function (knex) {
         });
     });
 
-    it('.timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection', async function () {
+    it('.timeout(ms, {cancel: true}) should throw TimeoutError and cancel slow query in transaction', function () {
```
martinmacko47, line 872: This test is a copy of ".timeout(ms, {cancel: true}) should throw TimeoutError and cancel slow query" test with just the query enveloped into a transaction.

You can use this test to confirm that canceling a timeouted query does not work inside a transaction on current `master`. The test will fail there. I've made as little as possible changes to the test, to make sure the cause is the transation.

test/integration/query/additional.js
```diff
@@ -869,7 +869,118 @@ module.exports = function (knex) {
...
+                sleepProcess = _.find(processes, {
+                  Info: 'SELECT SLEEP(10)',
+                });
+              }
+              expect(sleepProcess).to.equal(undefined);
+            });
+        });
+    });
+
+    it.skip('.timeout(ms, {cancel: true}) should throw error if cancellation cannot acquire connection', async function () {
```
martinmacko47, line 983: We go around the connection pool for the cancellation query by purpose. As the pool will likely to be exhausted when queries start timeouting, which would prevent us from cancelling the query in the first place.

Therefore this test is contradictory and will fail, as an exhausted pool should not prevent the cancellation.

lib/dialects/mysql/index.js
```diff
@@ -162,16 +162,16 @@ class Client_MySQL extends Client {
   }
 
   async cancelQuery(connectionToKill) {
-    const conn = await this.acquireConnection();
+    const conn = await this.acquireRawConnection();
     try {
-      return await this.query(conn, {
+      return await this._query(conn, {
         method: 'raw',
         sql: 'KILL QUERY ?',
```
martinmacko47, line 169: There is a bit of semantic difference between MySQL and Postgress:

- `KILL QUERY ?` in MySQL stops just the current query, but keeps the transaction active. So the user can send further queries to the same transaction and commit it eventually.
- `SELECT pg_cancel_backend($1)` in Postress stops the current query, and also automaticaly rejects the transaction. So any further queries sent to the same transaction will fail and the transaction will be eventually rolled back.

Should we keep this semantinc difference? Or should we rather manually roll back the transaction in MySQL as well. If we decide either way, it won't be a breaking change, as the query cancellation didn't work in transactions so far anyway.

kibertoad:
@elhigu @maximelkin @briandamaged This needs a careful review.

martinmacko47:
The CI error seems unrelated. When I run the worflow on my fork, CI passed. Perhaps just hit the re-run button (I don't have access to it, so I can't re-run the workflow for the PR).

lib/dialects/postgres/index.js
```diff
@@ -241,20 +241,20 @@ class Client_PG extends Client {
...
-    const conn = await this.acquireConnection();
+    const conn = await this.acquireRawConnection();
 
     try {
       return await this._wrappedCancelQueryCall(conn, connectionToKill);
     } finally {
       // NOT returning this promise because we want to release the connection
       // in a non-blocking fashion
-      this.releaseConnection(conn);
+      this.destroyRawConnection(conn);
```
maximelkin, line 251: this change could cause unhandledPromiseRejection if connection destroy fails, because releaseConnection was non-throwing.

lib/dialects/mysql/index.js
```diff
@@ -162,16 +162,16 @@ class Client_MySQL extends Client {
   }
 
   async cancelQuery(connectionToKill) {
-    const conn = await this.acquireConnection();
+    const conn = await this.acquireRawConnection();
     try {
-      return await this.query(conn, {
+      return await this._query(conn, {
         method: 'raw',
```
maximelkin, line 168: probably extra field (in pg too)

lib/dialects/postgres/index.js
kibertoad: Would you recommend to catch and log here?

lib/dialects/mysql/index.js
kibertoad: What do you mean by "extra"?

lib/dialects/postgres/index.js
martinmacko47: We might also just `await` the promise. As it was for MySQL dialect.

Actually why `this.releaseConnection(conn)` was awaited for MySQL while not for Postres? I would expect it to be symetric.

lib/dialects/mysql/index.js
martinmacko47: Oh, I see `method: 'raw'` field is not used in driver-level `_query`. It was just field used by knex-level `query` layer. I will remove the field. The same for Postres dialect.

martinmacko47: Removed in 0787821

lib/dialects/postgres/index.js
kibertoad: Hard to say. It originates back to `https://github.com/knex/knex/pull/2636/files` and was originally implemented that way, with no explanation for why non-blocking is preferred there.

martinmacko47: It looks like MySQL implementation also originally had un-awaited `releaseConnection`, and the `await` was added here: https://github.com/knex/knex/pull/3666/files#diff-c46d41926fdd59ada0e2bb2f34472b1295493d6dad75e336385ef4d22c28c32cR183

So I guess it should be awaited for Postgres implementation as well.

kibertoad: makes sense!

lib/dialects/postgres/index.js
```diff
@@ -241,20 +241,19 @@ class Client_PG extends Client {
     // Purposely not putting timeout on `pg_cancel_backend` execution because erroring
     // early there would release the `connectionToKill` back to the pool with
     // a `KILL QUERY` command yet to finish.
-    const conn = await this.acquireConnection();
+    const conn = await this.acquireRawConnection();
 
     try {
       return await this._wrappedCancelQueryCall(conn, connectionToKill);
     } finally {
       // NOT returning this promise because we want to release the connection
```
kibertoad, line 249: this comment is no longer accurate

lib/dialects/postgres/index.js
martinmacko47: Added `await` in 51b5191

maximelkin: btw does releaseConnection error should fail cancelQuery? before it doesnt

martinmacko47: @maximelkin IDK, I just think it should do the same for both dialects (MySQL and Postgress). In current version `cancelQuery` fails with `releaseConnection` error for Postgress, and ignores it for MySQL.

I guess it is like that because [this chage] was applied just to MySQL not to all dialects symmetrically.

[this chage]: https://github.com/knex/knex/pull/3666/files#diff-c46d41926fdd59ada0e2bb2f34472b1295493d6dad75e336385ef4d22c28c32cR183

maximelkin: I mean, releaseConnection https://github.com/knex/knex/blob/9c390036dbd98b1e2c37411db0eabdcb1247e9a4/lib/client.js#L306
Doesn't throw, but  destroyRawConnection does

martinmacko47: @maximelkin Ok. I see. So what should I do with such exception?

maximelkin: catch and log error, this will be backward compatible

martinmacko47: @maximelkin By logging the error you mean `this.logger`?

```js
try {
  return await this._wrappedCancelQueryCall(conn, connectionToKill);
} finally {
  this.destroyRawConnection(conn).catch((err) => {
    this.logger.warn(`Connection Error: ${err}`);
  })
}
```

Or something else?

What about MySQL? There I noticed that MySQL `destroyRawConnection` [catches errors itself](https://github.com/knex/knex/blob/master/lib/dialects/mysql/index.js#L73-L74):
```js
  async destroyRawConnection(connection) {
    try {
      const end = promisify((cb) => connection.end(cb));
      return await end();
    } catch (err) {
      connection.__knex__disposed = err;
    } finally {
      // see discussion https://github.com/knex/knex/pull/3483
      defer(() => connection.removeAllListeners());
    }
  }
```
So then I guess for MySQL it is OK as it is, isn't it?

maximelkin: Maybe log __knex_disposed error?

martinmacko47: @maximelkin catched and logged in f93c03f.

test/integration/query/additional.js
kibertoad: We can remove the test then

kibertoad: Do we ignore the pool limits for all dialects now, or test still makes sense for some of them?

martinmacko47: @kibertoad

> Do we ignore the pool limits for all dialects now, or test still makes sense for some of them?

Query cancelling is supported ony for mysql and postgres, so this test was [skipped for other dialects](https://github.com/knex/knex/blob/master/test/integration/query/additional.js#L871-L874) anyway.

```js
      // Only mysql/postgres query cancelling supported for now
      if (!isMysql(knex) && !isPostgreSQL(knex)) {
        return this.skip();
      }
```

> We can remove the test then

Maybe I'd rather turn the test arround that the query is indeed cancelled even if pool is exhausted.

kibertoad: Great idea!

martinmacko47: @kibertoad Test refactored in 944e868

maximelkin (approved):

kibertoad:
Released in 0.95.3.

PR #4330, opened 2021-03-02
SQLite: made the constraint detection case-insensitive

tpatel:
Hey @kibertoad 👋  I'm Thibaut from [Ghost](https://github.com/TryGhost/), we're working on releasing Ghost 4.0 very soon but found a bug in knex that is blocking us, here is the fix for it! The best case scenario for us would be if this fix would land in knex `0.21` ❤️ 

The problem comes when using the `foreign` feature on SQLite.

The limitations of SQLite requires the creation of a temporary table, and the constraints are uppercased due to [this code](https://github.com/knex/knex/blob/0.21/lib/dialects/sqlite3/schema/ddl.js#L478-L489). This means it introduces constraints definitions in uppercase format (like `FOREIGN KEY`).
However the constraint detection is using `startsWith` or `includes` which are both case-sensitive.

This PR fixes this issue by lowercasing the definitions before searching for the keywords.

### DIFF
--- a/lib/dialects/sqlite3/schema/ddl.js
+++ b/lib/dialects/sqlite3/schema/ddl.js
@@ -355,8 +355,8 @@ class SQLite3_DDL {
           .map((line) => line.trim())
           .filter((defLine) => {
             if (
-              defLine.startsWith('constraint') === false &&
-              defLine.includes('foreign key') === false
+              defLine.toLowerCase().startsWith('constraint') === false &&
+              defLine.toLowerCase().includes('foreign key') === false
             )
               return true;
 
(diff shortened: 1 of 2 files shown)


kibertoad:
Thank you for contribution! Could you please add integration test for this?
See https://github.com/knex/knex/blob/master/test/integration2/schema/alter.spec.js for a good example

kibertoad (approved):

kibertoad:
Thanks! Can you also create a backport PR for branch `0.21`? Not sure if it will be straightforward, though, lots of SQLite-related code changed in master.
We are planning to release 0.95.0 in a few days, would that work for you?

tpatel:
Thank you so much for the quick action ❤️  I've created the backport PR here: https://github.com/knex/knex/pull/4332.

kibertoad:
@tpatel Released in 0.95.0

PR #4333, opened 2021-03-02
SQLite parser improvements

nickrum:
The first two commits should be self explanatory.

The third commit adds support for the `NULL` column constraint to the SQLite create table parser and compiler. This constraint is not mentioned in the SQLite documentation because it is a no-op. SQLite only parses it for compatibility reasons. As Knex sets this constraint on columns created with `.nullable()`, we have to parse it as well.
The alternativ would be to not set this constraint on nullable columns and to not parse it. Unfortunately, this would make altering columns on existing databases with nullable columns (and on those created with a raw query using the null constraint) impossible.


test/integration2/schema/alter.spec.js
```diff
@@ -119,7 +122,25 @@ describe('Schema', () => {
...
+            });
+
+            it('removes an existing null constraint if a not null constraint is added to a column', async () => {
+              await knex.schema.alterTable('alter_table', (table) => {
+                table.boolean('column_nullable').notNullable().alter();
+              });
+
+              const tableAfter = (await knex.raw(QUERY_TABLE))[0].sql;
+
+              expect(
```
kibertoad, line 136: this does not need to be awaited?

nickrum: It definitely does! My bad, I think I misunderstood how the Chai Promise library works.

kibertoad (approved):

kibertoad:
Great job as usual!

nickrum:
Thank you, I'm doing my best!

PR #4336, opened 2021-03-03
Fix SQLite not doing rollback when altering columns fails

nickrum:
The statements produced by the statements producer are currently not executed inside a transaction, which means that the changes are not rolled back if an error occurs. This will be fixed by #4189, but I won't find the time to finish that PR before the `0.95.0` release. Once it's done, this change can be reverted.

### DIFF
--- a/lib/dialects/sqlite3/schema/ddl.js
+++ b/lib/dialects/sqlite3/schema/ddl.js
@@ -279,13 +279,9 @@ class SQLite3_DDL {
 
         const newTable = compileCreateTable(parsedTable, this.wrap);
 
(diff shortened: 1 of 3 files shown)


kibertoad:
@nickrum What's your estimate for #4189? We can potentially delay 0.95.0 if it's somewhere in sight.

nickrum:
@kibertoad Finishing #4189 is my next top priority, but I will be busy with personal things for at least 1-2 weeks. So, including the time it takes to finish the PR, I don't think it would be worth it to delay the release.

kibertoad (approved):

PR #4344, opened 2021-03-04
fix: update ts.stub files to new TypeScript namespace

KubaJastrz:
Changed `ts.stub` files for migrations and seeds to reflect changes in 0.95, also documented in https://github.com/knex/knex/blob/master/UPGRADING.md#upgrading-to-version-0950
> `import { Knex } from 'knex' // this is a namespace, and a type of a knex object`

---

Also changed `test/jake-util/knexfile-ts/knexfile_migrations` files for the same reason, including an import path change from `<root>/knex.js` to `<root>`, as the node will automatically resolve the `<root>` import to what's in the `package.json#main` field, allowing TypeScript to do similar for `package.json#types` field.

### DIFF
--- a/lib/migrations/migrate/stub/ts.stub
+++ b/lib/migrations/migrate/stub/ts.stub
@@ -1,4 +1,4 @@
-import * as Knex from "knex";
+import { Knex } from "knex";
 
 <% if (d.tableName) { %>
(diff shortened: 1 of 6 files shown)


kibertoad (approved):

kibertoad:
Thank you!

PR #4348, opened 2021-03-05
package.json changes (order/sort) made by npm@7.6.1

Ardeshir81:
Hi. First time contributing. I just did an `npm i` and I had a diff in `package.json`. So instead of discarding it, or keeping a dirty working tree, I decided to do a PR, so `npm i`s from now on would be cleaner. my `npm` version is `7.6.1`

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,11 +57,11 @@
     "tildify": "2.0.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


kibertoad (approved):

kibertoad:
Thanks!

PR #4350, opened 2021-03-06
Improve ESM import support

### LINKED ISSUE #4346: 0.95 with old type behaviour
# Environment

Knex version: 0.95
mySQL + 5.7:
OS: macOS
Node: 14.16
NPM: 6

When using "type": "module" in package.json, is there any way to get old (0.28) behaviour of types export.

I am getting errors in the VS Code and when run node app

image

image

Here it work, but then then the VS Code does not show completion

image

kibertoad:
fixes #4346

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,3 +116,33 @@ try {
   // error handling
 }
 ```
+
+## Usage as ESM module
+
+If you are launching your Node application with `--experimental-modules`, `knex.mjs` should be picked up automatically and named ESM import should work out-of-the-box.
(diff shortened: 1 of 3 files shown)


kibertoad:
@terkolaka Could you try using knex/knex#fix/4346-types as a dependency version and also change
```
/**
 * @type {knex}
 */
```

into
```
/**
 * @type {Knex}
 */
```
?

This seems to work on your repro repo for me, would be interested to see if it helps you too.

PR #4356, opened 2021-03-07
[WIP] Unify db test helpers

machuga:
Refactoring tests to use the functions in db-helpers rather than checking for the type of driver in `n` different ways (feedback on #4327).

This PR seemed large enough to break out into its own and do it in a standalone fashion before moving on with #4327. 

Also adding constants for driver names

When testing against the `driverName` there seems to be a number of
locations where the value being checked against may not be the actual
`driverName`. This may result in tests passing or failing incorrectly.

Adding a constant that has all current `driverName`s defined so that
matching can be done confidently and with tooling support. It's
effectively an enum.

Updating the tests to use the db-helpers, and having the db-helpers
utilize the drivers for correct naming should ensure all tests fire as 
expected and allow for easier refactoring.

### DIFF
--- a/test/integration/execution/transaction.js
+++ b/test/integration/execution/transaction.js
@@ -7,11 +7,13 @@ const _ = require('lodash');
 const sinon = require('sinon');
 const { KnexTimeoutError } = require('../../../lib/util/timeout');
 const delay = require('../../../lib/execution/internal/delay');
+const { isRedshift, isOracle, isMssql, isPostgreSQL } = require('../../util/db-helpers');
+const { DRIVER_NAMES: drivers } = require('../../util/constants');
 
 module.exports = function (knex) {
   // Certain dialects do not have proper insert with returning, so if this is true
   // then pick an id to use as the "foreign key" just for testing transactions.
-  const constid = /redshift/.test(knex.client.driverName);
+  const constid = isRedshift(knex);
   let fkid = 1;
 
   describe('Transactions', function () {
@@ -37,10 +39,7 @@ module.exports = function (knex) {
 
     it('supports direct retrieval of a transaction without a callback', () => {
       const trxPromise = knex.transaction();
-      const query =
-        knex.client.driverName === 'oracledb'
-          ? '1 as "result" from DUAL'
-          : '1 as result';
+      const query = isOracle(knex) ? '1 as "result" from DUAL' : '1 as result';
 
       let transaction;
       return trxPromise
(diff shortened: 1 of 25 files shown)


test/integration/execution/transaction.js
```diff
@@ -681,9 +676,9 @@ module.exports = function (knex) {
       const mysqlKillConnection = async (connection) =>
         knex.raw('KILL ?', [connection.threadId]);
       const killConnectionMap = {
-        mysql: mysqlKillConnection,
-        mysql2: mysqlKillConnection,
-        pg: async (connection) =>
+        [drivers.MySQL]: mysqlKillConnection,
```
machuga, line 679: Example of using the constants in here to ensure the `driverName` matches. May be overkill, but it is an experiment to gauge interest.

kibertoad: This does look like a great idea!
Will you add separate branch for pg-native here later on? Wonder if we can somehow cover all multi-dialect dbs without multiple entries.

test/integration/query/inserts.js
```diff
@@ -740,7 +747,7 @@ module.exports = function (knex) {
           function () {
             // No errors happen in sqlite3, which doesn't have native support
             // for the enum type.
-            if (knex.client.driverName !== 'sqlite3') {
+            if (isSQLite(knex)) {
```
kibertoad, line 750: Check should be inverted

test/integration/execution/transaction.js
machuga: I've thought about it a bit this afternoon. The idea would also be that this could be used in places like: 
https://github.com/knex/knex/pull/4356/files#diff-f275a3d5b4fde0fe972896d099c1eac3ae2614ff845e49a0f73e276ee925e8b5R985 to ensure it always points toward a valid type. This may be overkill because I think some of these areas check to see if it's a valid strategy. 

As for answering the direct question:

In the situation of a map where the types are implicitly matched, duplicating seems like the only option. Switch statements will offer roughly the same ability. I found utilizing `else if` style, as in some areas like this: https://github.com/knex/knex/pull/4356/files#diff-f275a3d5b4fde0fe972896d099c1eac3ae2614ff845e49a0f73e276ee925e8b5R1482 worked well so that the same helpers could be used without much structure change. This lets `isPgBased` be used for `pg`, `pg-native`, and `pg-redshift`, `isMySQL` be used for those grouping, etc. It's not the most elegant, but it does allow the encapsulation to be maintained.

There are definitely other options, but I'm curious of your thoughts. Additionally we can do nothing further with this PR (or even back them out) and come back to it after some more thought goes into it.

machuga: Given the number of occurrences of the references in the tests, doing a manual conversion any further may not yield the best ROI (given risk and time allocations).

kibertoad (approved):

kibertoad:
Thank you! As for the duplicated checks, need to give it more thought.

PR #4361, opened 2021-03-09
[wip] fixes #4335: warn if negative offset was provided

### LINKED ISSUE #4335: Preventing negative `offset` by default
# Environment

Knex version: 0.21.17
Database + version:  `~#: mysql --version -> mysql  Ver 14.14 Distrib 5.5.62, for debian-linux-gnu (x86_64) using readline 6.3`
OS: Debian

# Bug

1. I mistakenly set a negative offset:
  ```ts
// ...
 page(page: number = 1) {
    const offset = (page - 1) * this.limit;
    this.query.offset(offset);
  }
   ```
I mistakenly passed `0` to this function, and my query became: 
`select * from `` order by `Time` desc limit 20 offset -20`

2. Error message
Screenshot from 2021-03-03 11-34-10
(sorry for the blue marks)

3. Reduced test code:

```ts
import Knex from 'knex';

const knexConnection = Knex({
  client: 'mysql',
  connection: {
    host: '', //...
    user: '', //...
    password: '', //...
    database: '', //...
  },
});
const query = knexConnection('test').limit(20).offset(-20);
```

I believe if no negative offset is allowed in mysql query, the knex query-builder should prevent it. (instead of letting the error happen and generated by mysql)

If it is knex that should implement this, and this is known as fix that needs to be done in knex, I'd love to contribute to code.

Ardeshir81:
HI. I started working on this. I faced some challenges I need advice on.

- First of all, I'm only familiar with `mysql`. Is negative offset invalid for all kind of databases? Should I consider database type?
- Then I found `offset` function in `querybuilder.js` file, and extended an existing condition:
image
- The problem with above solution was that negative offset values resulted in a warn log (orange in my case) and were automatically omitted (or converted to `zero`). I thought this automatic omitting makes bad logic/behavior is user's code hidden or hard-to-debug.
-  Later I  thought maybe it's not a bad idea if we threw an error in case of negative offset
image
- The above solution can be and not be a breaking change. It can be a breaking change, because it did not use to throw an error, but can not be thought of a breaking change because the query would result to an exception throw (by the database) later anyway.
- The final solution I came up with is this:
image

I thought if it would be a nice idea if we warn user that this behavior (providing `negative` value to `offset`) would be deprecated and throw in a future version?

Thanks if you really read all that :hearts:

### DIFF
--- a/lib/query/querybuilder.js
+++ b/lib/query/querybuilder.js
@@ -908,6 +908,8 @@ class Builder extends EventEmitter {
       const val = parseInt(value, 10);
       if (isNaN(val)) {
         this.client.logger.warn('A valid integer must be provided to offset');
+      } else if (val < 0) {
+        throw new Error(`A non-negative integer must be provided to offset.`);
       } else {
         this._single.offset = val;
       }
--- a/test/integration/query/selects.js
+++ b/test/integration/query/selects.js
@@ -156,6 +156,21 @@ module.exports = function (knex) {
         });
     });
 
+    it('#4335 - should throw an error when negative offset provided', function (ok) {
+      try {
+        knex.from('accounts').limit(20).offset(-20);
+        throw new Error('no error was thrown for negative offset!');
+      } catch (error) {
+        if (
+          error.message === 'A non-negative integer must be provided to offset.'
+        ) {
+          ok();
+        } else {
+          throw error;
+        }
+      }
+    });
+
     it('#4199 - adheres to hint comments', async function () {
       const expectedErrors = {
         mysql: {


kibertoad:
@Ardeshir81 Makes sense to make negative value support deprecated! So what happens now if value is < 0? If I understand the code correctly, it is ignored, which is equivalent to passing 0, which is same is it worked before, right?

Could you also add integration test (`test/integration2`) for this?

Ardeshir81:
No it is not the same as it worked before. As stated in issue (https://github.com/knex/knex/issues/4335), it resulted in querying negative offset on database which threw an error by the db itself (not the library).

I believe throwing an error is a good practice, because, in my case it helped me find a bad logic in my code, But automatic ignore/conversion to zero, may make library users' bugs hard-to-find-&-debug.

So yes! You got it correctly, it is ignored, which is equivalent to passing 0, but it log warns the user about it.

I'll add integration tests :+1: . But what is the deprecation process? Is it enough to just warn the user that passing negative values to offset will be deprecated in a future version ?

kibertoad:
> Is it enough to just warn the user that passing negative values to offset will be deprecated in a future version ?

Yup!

Ardeshir81:
Heay @kibertoad . Nice profile picture :+1: 

You answered my question about `Is it enough to just warn the user that passing negative values to offset will be deprecated in a future version ?`. But I read the discussion over and over again here and in the issue (https://github.com/knex/knex/issues/4335) and decided to revert the functionality to throwing an error. I think its a good practice, as like a bug is fixed.

I squashed the commit. And also wrote tests. I had problems understanding and running tests, so wrote one in a different style.

is it ok ? 

thanks

Ardeshir81:
Is the test and code ok @kibertoad ?

kibertoad:
@Ardeshir81 Sorry for the delay, will review today.

kibertoad (approved):

kibertoad:
Released in 0.95.3!

Issue #4204, opened 2021-01-05
MSSQL Comments in SchemaBuilder are not used, using exec sp_updateextendedproperty in lieu of comments


alexanderankin:
# Environment

Knex version: 21.12
Database + version: mssql (just the dialect)
OS: ubuntu

Select applicable template from below.
If issue is about MSSql tag @smorey2.

# Feature discussion / request

1. Explain what is your use case

I would like to use the `db.schema.createTable('mytable', t => t.string('somecol').comment('the col'));` method and have the resulting comment be persisted in the database. The warning is presented to the user if the comment is too big (instead of giving feedback that its not supported) so lets support it.

2. Explain what kind of feature would support this

Here's background on what this is: [web.archive.org http://sqlblog.com/blogs/michael.../2010/01/12/t-sql-tuesday-easy-extended-properties.aspx](https://web.archive.org/web/20170716134332/http://sqlblog.com/blogs/michael_coles/archive/2010/01/12/t-sql-tuesday-easy-extended-properties.aspx). basically the suggested proposal below implements this, which is how you can add comments using the visual tools provided with [SSMS](https://docs.microsoft.com/en-us/sql/ssms/sql-server-management-studio-ssms?view=sql-server-ver15).

3. Give some API proposal, how the feature should work

I would like to add a PR with warnings where the column information is discarded.

Ideally knex would implement the real functionality (not just discarding values) where possible, here's a sneak preview of changes I would like to submit:

<details>
<summary>the diff</summary>

```diff
diff --git a/lib/dialects/mssql/schema/columncompiler.js b/lib/dialects/mssql/schema/columncompiler.js
index c8563b62..90925f37 100644
--- a/lib/dialects/mssql/schema/columncompiler.js
+++ b/lib/dialects/mssql/schema/columncompiler.js
@@ -93,11 +93,14 @@ Object.assign(ColumnCompiler_MSSQL.prototype, {
   comment(comment) {
     if (comment && comment.length > 255) {
       this.client.logger.warn(
         'Your comment is longer than the max comment length for MSSQL'
       );
     }
+
+    this.tableCompiler.addExtendedPropertyToColumn(comment, this.args[0]);
+
     return '';
   },
 });
 
 module.exports = ColumnCompiler_MSSQL;
diff --git a/lib/dialects/mssql/schema/tablecompiler.js b/lib/dialects/mssql/schema/tablecompiler.js
index 6ed8d6af..e7e5f877 100644
[...]

PR #4362, opened 2021-03-10
mssql: Improve estimate of max comment length

jeremy-w:
Before, table comments used a limit of 60 chars (seemingly inherited from the MySQL dialect), while columns warned about exceeding a limit of 255 chars. This led to spurious warnings.

Now, warnings are only emitted if it seems possible for the comment to exceed the actual, documented 7,500 byte limit:

> `[ @value= ] { 'value'}`
> Is the value to be associated with the property. value is `sql_variant`, with a default of `NULL`. **The size of value cannot be more than 7,500 bytes.** ([sp_addextendedproperty (Transact-SQL)](https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-addextendedproperty-transact-sql?view=sql-server-ver15), emphasis added)

That limit proved to be when the `comment.length` exceeded 7500/2, as demonstrated by testing.

(This is a small step towards https://github.com/knex/knex/issues/4204.)

## Questions for Reviewers
- Making the warning more precise:
  - Would it be OK for a schema compiler to run queries to sort out stuff like "what is the database collation"? That would theoretically allow us to actually compute the byte length, not make a pessimistic estimate. (But my guess is "no".)
  - If not, would it be better to use an optimistic estimate of 1 char == 1 byte?
- **Should we even bother warning at all?** When we get around to actually adding the comments, MSSQL will raise an error if the length is exceeded for us:

  > Msg 15097, Level 16, State 1, Server 49df3e21e80c, Procedure sp_updateextendedproperty, Line 16
  > The size associated with an extended property cannot be more than 7,500 bytes.


kibertoad:
@jeremy-w Considering that `The size associated with an extended property cannot be more than 7,500 bytes.` doesn't mention comments anywhere, warning is still helpful to point people experiencing the issue in the right direction. Could you also add two integration tests, one of them being slightly below the threshold, and another one being above the threshold and actually causing MSSQL to fail?

jeremy-w:
@kibertoad would the warning still be useful if we trapped the error during adding the comment and converted it into a more knex-relevant error? or would that be preferable?

since the "length exceeded" error has Level 16 (greater than 10, less than 20), it looks like [try…catch](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/try-catch-transact-sql?view=sql-server-ver15) would work.

kibertoad:
@jeremy-w While we already have plenty of magic for handling MSSQL edge-cases, I would prefer to avoid adding even more unless strictly necessary. Just improving error messages doesn't sound like a big enough reason to complicate generated queries, so I would still prefer just a knex warning in this case.

jeremy-w:
@kibertoad Updated with the requested tests. Also added a test to validate our "7500/4" estimate for when to warn is a good one.

This also gets us most of the way to being able to generate comments as part of the schema builder, so things are looking good for #4204. :)

test/integration2/dialects/mssql.spec.js
```diff
@@ -0,0 +1,179 @@
...
+          // (They also happen to need 4 in UTF-8, so this estimate is a useful worst-case for a UTF-8 collation, as well.)
+          const N = 7500 / 4;
+          it(`worst-case allows at most ${N} characters in an nvarchar`, async () => {
+            const astralPlaneCharacter = '\u{1D306}';
+            const worstCaseComment =
+              "N'" + astralPlaneCharacter.repeat(N) + "'";
+            await expect(comment(knex, 'add', worstCaseComment, columnTarget))
+              .to.eventually.be.fulfilled;
+
+            const oneMoreByte = 'X';
```
jeremy-w, line 143: ❌ this is actually 2 more bytes in UTF-16

test/integration2/dialects/mssql.spec.js
```diff
@@ -0,0 +1,179 @@
...
+            await expect(
+              comment(knex, 'update', byteCountOver7500, columnTarget)
+            ).to.eventually.be.rejectedWith(
+              'The size associated with an extended property cannot be more than 7,500 bytes.'
+            );
+          });
+
+          // Characters in the supplementary plane need 4 bytes in UTF-16.
+          // (They also happen to need 4 in UTF-8, so this estimate is a useful worst-case for a UTF-8 collation, as well.)
+          const N = 7500 / 4;
```
jeremy-w, line 135: ❗ i'm not confident our estimation here based on string length is right, though, because JS will treat the astral plane character i'm using as having a length of 2. we can probably instead allow the string.length to be 7500/2 as a result!

**TODO:** confirm this by asserting against the test comment raw string length, & revise the estimate if it proves out.

jeremy-w (commented):
thought of a couple issues overnight. i bet we can allow a 2x longer char count before warning due to how JS counts character length when characters outside the BMP get involved.

jeremy-w:
OK, I addressed the issues I thought of, and I was able to provide a lot more headroom - comments have to be longer than 3,750 characters before comment length begins to be a concern!

kibertoad (approved):

kibertoad:
Thank you!

kibertoad:
Released in 0.95.3.

PR #4363, opened 2021-03-10
add dropForeign arg type for single column

capaj:


kibertoad:
@capaj 
```
  types/index.d.ts:1862:29
  ✖  1862:29  readonly type modifier is only permitted on array and tuple literal types.  
```

capaj:
@kibertoad should be good now

kibertoad (approved):

kibertoad:
Thanks!

kibertoad:
Released in 0.95.3.

PR #4366, opened 2021-03-11
fix TypeScript migration stub after 0.95.0 changes

santialbo:
Release 0.95.0 changed how types have to be imported when using TypeScript. This PR fixes the migration stub so new migrations are created correctly.

### DIFF
--- a/lib/migrations/migrate/stub/ts.stub
+++ b/lib/migrations/migrate/stub/ts.stub
@@ -1,8 +1,8 @@
-import * as Knex from "knex";
+import { Knex } from "knex";
 
 <% if (d.tableName) { %>
 export async function up(knex: Knex): Promise<Knex.SchemaBuilder> {
-    return knex.schema.createTable("<%= d.tableName %>", (t: Knex.AlterTableBuilder) => {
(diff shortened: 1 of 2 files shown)


lib/migrations/migrate/stub/ts.stub
```diff
@@ -1,8 +1,8 @@
-import * as Knex from "knex";
+import { Knex } from "knex";
 
 <% if (d.tableName) { %>
 export async function up(knex: Knex): Promise<Knex.SchemaBuilder> {
-    return knex.schema.createTable("<%= d.tableName %>", (t: Knex.AlterTableBuilder) => {
+    return knex.schema.createTable("<%= d.tableName %>", (t) => {
```
kibertoad, line 5: Is there a reason why `Knex.AlterTableBuilder` was removed? Is it redundant now?

santialbo: Yes, it's redundant, TypeScript infers the type which is actually `Knex.CreateTableBuilder` not `Knex.AlterTableBuilder`.

kibertoad (approved):

PR #4367, opened 2021-03-11
[TypeScript] Allow config.client to be non-client instance

thomaschaaf:
TypeScript makes it very hard to get typescript to correctly check the `instanceof Client` part of the statement.

See: https://github.com/Microsoft/TypeScript/wiki/FAQ#why-doesnt-extending-built-ins-like-error-array-and-map-work


kibertoad (approved):

kibertoad:
Thanks!

kibertoad:
Released in 0.95.3.

PR #4376, opened 2021-03-15
Rework dropForeign to use new tokenized structure

### LINKED ISSUE #4369: [SQLite] `dropForeign` is silently ignored
# Environment

Knex version: 0.95.2
Database + version: SQLite (any)
OS: macOS

# Bug

When using [`.dropForeign()`](http://knexjs.org/#Schema-dropForeign) in SQLite, it's silently ignored. When inspecting the queries made, it correctly creates a new temp table for the `alterTable` call, but fails to strip out the `FOREIGN KEY` line in the create table sql.

2. Error message

n/a

3. Reduced test code

```js
require('sqlite3');
const Knex = require('knex');

const sqlite = Knex({
  client: 'sqlite',
  connection: ':memory:',
});

(async function run() {
  await sqlite.schema.createTable('test', (t) => {
    t.increments('id').primary();
  });
  
  await sqlite.schema.createTable('related', (t) => {
    t.increments('id').primary();
    t.integer('fk').references('id').inTable('test');
  });
  
  console.log(await getCreateTableSQL('related'));
  // Contains the `FOREIGN KEY` as expected
  
  await sqlite.schema.alterTable('related', (t) => {
    t.dropForeign(['fk']);
  });
  
  console.log(await getCreateTableSQL('related'));
  // Still contains the `FOREIGN KEY`. Should've been removed.
})();

async function getCreateTableSQL(table) {
    const result = await sqlite
        .select('sql')
        .from('sqlite_master')
        .where({ name: table })
        .first();
    
    return result && result.sql;
}
```

https://runkit.com/rijkvanzanten/knex-sqlite-drop-foreign-example

(@nickrum, before I start chopping away in your amazing work from the other day, do you have any instructions on how to (re-)implement `dropForeign` using the new tokenized approach you set up?)

rijkvanzanten:
Fix dropForeign not working as expected in SQLite.  Uses the new setup introduced by @nickrum .

Fixes #4369, fixes https://github.com/directus/directus/issues/4489

### DIFF
--- a/lib/dialects/sqlite3/schema/ddl.js
+++ b/lib/dialects/sqlite3/schema/ddl.js
@@ -334,49 +334,40 @@ class SQLite3_DDL {
     );
   }
 
-  async dropForeign(columns, indexName) {
+  async dropForeign(columns, foreignKeyName) {
     return this.client.transaction(
       async (trx) => {
         this.trx = trx;
 
         const { createTable, createIndices } = await this.getTableSql();
 
-        const oneLineSql = createTable.replace(/\s+/g, ' ');
-        const matched = oneLineSql.match(/^CREATE TABLE\s+(\S+)\s*\((.*)\)/);
+        const parsedTable = parseCreateTable(createTable);
 
-        const defs = matched[2];
-
-        if (!defs) {
-          throw new Error('No column definitions in this statement!');
+        if (!foreignKeyName) {
+          parsedTable.columns = parsedTable.columns.map((column) => ({
+            ...column,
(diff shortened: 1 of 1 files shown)


kibertoad:
Can we also fix test to fail properly?

rijkvanzanten:
We did! @nickrum figured out that the reason the test "failed successfully" was because of the fact that `dropForeign` would delete _all_ foreign keys when a foreign key name was passed. This would cause the test to pass for the wrong reason 👍🏻

PR #4377, opened 2021-03-16
nguymin4/update typings

nguymin4:
Update typings for
- `TypePreservingAggregation` with options  `as`. Issue https://github.com/knex/knex/issues/4165
```typescript
await knexInstance<User>('users')
    .min('age', {as: 'a'})
    .max('age', {as: 'b'});
```
- `stream()` AsyncInterable<T>
```typescript
const stream =  knexInstance<User>('users').select('*').stream()
for await (const user of stream) {
    // console.log(user)
}
```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1737,6 +1737,7 @@ export declare namespace Knex {
     connection(connection: any): this;
     debug(enabled: boolean): this;
     transacting(trx: Transaction): this;
+    stream(): AsyncIterable<ArrayMember<T>>;
     stream(handler: (readable: stream.PassThrough) => any): Promise<any>;
     stream(
       options: Readonly<{ [key: string]: any }>,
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -1430,7 +1440,7 @@ export declare namespace Knex {
         // causes rest of the keys to be dropped and only first to be considered
         [K in keyof TAliases]?: K extends keyof TRecord ?
           TRecord[K] :
-          TValue
+          UnknownToAny<ResolveTableType<TRecord[TAliases[K]]>>
```
nguymin4, line 1443: The test currently fails with `Type TAliases[K] cannot be used to index type TRecord. ` Does anyone know what could be the best way to handle this case?
The idea is 
```typescript
  // $ExpectType ({ dep: number; } & { a?: number; } & { b?: number; })[]
  await knexInstance<User>('users')
    .select({dep: 'departmentId'})
    .min({a: 'age'} as const)
    .max({b: 'age'} as const);
```

kibertoad (approved):

kibertoad:
Thank you!

kibertoad:
Released in 0.95.3.

nguymin4:
Thank you

PR #4400, opened 2021-03-26
fix: mistyping of stream

monotykamary:
Reverts a breaking type change to `stream` from https://github.com/knex/knex/pull/4377. The `stream` method returns an
instance of `stream.Passthrough` or allows a callback of an instance `stream.Passthrough` that returns a `Promise` as per documentation. This is not interchangeable with `AsyncIterable` as `Readable` already implements `AsyncIterableIterator`.


kibertoad:
@nguymin4 Could you comment?

kibertoad:
@monotykamary Is it possible to adjust type instead of removing it?

nguymin4:
@kibertoad wihout AsyncIterable the `user` type in the code below will be `any` this is bad practice in my opinion
```
  const stream = knexInstance<User>('users').select('*').stream();
  for await (const user of stream) {
    // user type is any
  }
```

monotykamary:
@nguymin4 Breaking the type is also bad practice, especially since the stream returns a `Duplex` and there may be cases where a user needs to read and abort the stream; although I haven't considered that the original pull request was to cover instance types over iterables.

> @monotykamary Is it possible to adjust type instead of removing it?

In that case, it might just make sense by intersecting both types:
```ts
    stream(): stream.PassThrough & AsyncIterable<ArrayMember<T>>;
```

kibertoad:
@monotykamary Could you adjust the PR to do the intersection?

nguymin4:
The intersection will not work. It doesn't resolve the type of `user` correctly and return `any`. But yeah if people are using loosely typing then we shouldn't make this change.
From API level point of view, I think we are trying to do too much at the same time `stream` which works in Javascript but not nicely in Typescript. In a sense I feel like the `pipe` method should be the only one return stream.Passthrough. Or alternative solution I'm thinking about is we can have `asIterable` method but maybe too much work and not worth the effort.

monotykamary:
> The intersection will not work.

It seems to work fine on my end:
image

Does the repository have a specific typescript version target that this doesn't work on?

nguymin4:
Doesn't seems to work for me unfortunately :thinking: 
Screenshot from 2021-03-26 19-29-24

nguymin4:
Ok seems to work now. So we can merge those two streams method into one.
`stream(options?: Readonly<{ [key: string]: any }>): stream.PassThrough & AsyncIterable<ArrayMember<T>>;`

@monotykamary Could you also add a test for the `stream.Passthrough` as well? Probably something like
```
// $ExpectType PassThrough
knexInstance<User>('users').select('*').stream().pipe(new PassThrough());
```
I forgot to add this in previous PR which caused the issue.

types/test.ts
```diff
@@ -132,7 +132,7 @@ const main = async () => {
       [ null ]
   );
 
-  // $ExpectType AsyncIterable<User>
+  // $ExpectType stream.PassThrough & AsyncIterable<User>
   knexInstance<User>('users').select('*').stream();
```
kibertoad, line 136: can we also add
```
// $ExpectType PassThrough
knexInstance<User>('users').select('*').stream().pipe(new PassThrough());
```
?

types/index.d.ts
```diff
@@ -1737,7 +1737,7 @@ export declare namespace Knex {
     connection(connection: any): this;
     debug(enabled: boolean): this;
     transacting(trx: Transaction): this;
-    stream(): AsyncIterable<ArrayMember<T>>;
+    stream(): stream.PassThrough & AsyncIterable<ArrayMember<T>>;
```
nguymin4, line 1740: Let's merge this line and line 1746. They're basically the same.

nguymin4 (approved):
Looks good to me!

Please fix the test. Probably missing import.

kibertoad (approved):

kibertoad:
Thank you!

PR #4416, opened 2021-04-06
Fix cancelling timeouted queued queries in transactions for MySQL

### LINKED ISSUE #4415: `.timeout(ms, {cancel: true})` for with queued queries in transaction
# Environment

Knex version: current `master`
Database + version: MySQL - any version
OS: any

# Bug

If we queue multiple slow queries with timeouts in a transaction, not all of them will get cancelled after timeouting. E.g.:

```
  return Promise.all([
    firstSlowQuery(trx).timeout(5000, {cancel: true}),
    secondSlowQuery(trx).timeout(5000, {cancel: true}),
    ...
    lastSlowQuery(trx).timeout(5000, {cancel: true}),
  ])
})
```

As the queries are in the transaction, they get queued in the driver which sends them to DB one by one. As all queries will timeout at approximately the same time, all their `KILL QUERY ?` commands will be also sent to DB at approximately the same time. Therefore they will kill only the first query. After that the driver sends the second query to DB which will be never killed, as its timeout already elapsed and its `KILL QUERY ?` command was already sent.

The problem occurs only on MySQL. Postgres marks the transaction failed after killing the query, therefore any further queries in the same transaction will immediately fail.

I guess we should simulate the same behaviour on MySQL as well. Either by removing the timeouted queries from the driver queue, or by rolling back the transaction with the killed timeouted query.

I've written a test to replicate this bug. It passes for Postgres and fails for MySQL: https://github.com/martinmacko47/knex/pull/4/files#diff-ef380392f7c99984959538380cd3a01631d49da7eb768b02fd71695bfb2a3febR981-R1052

martinmacko47:
Attempt to fix #4415 by removing the timeouted queries from the driver queue.

### DIFF
--- a/lib/dialects/mysql/index.js
+++ b/lib/dialects/mysql/index.js
@@ -172,6 +172,7 @@ class Client_MySQL extends Client {
   async cancelQuery(connectionToKill, queryToKill) {
     if (this._cancelQueuedQuery(connectionToKill, queryToKill)) return;
 
+    connectionToKill.pause();
     const conn = await this.acquireRawConnection();
     try {
       return await this._query(conn, {
@@ -180,6 +181,7 @@ class Client_MySQL extends Client {
         options: {},
       });
     } finally {
+      connectionToKill.resume();
       await this.destroyRawConnection(conn);
       if (conn.__knex__disposed) {
         this.logger.warn(`Connection Error: ${conn.__knex__disposed}`);
--- a/test/integration/query/additional.js
+++ b/test/integration/query/additional.js
@@ -1111,6 +1111,70 @@ module.exports = function (knex) {
       });
     });
 
+    it('.timeout(ms, {cancel: true}) should not cancel subsequent queries in a transaction if timeouted query completes while being cancelled', async function () {
+      // Only mysql/postgres query cancelling supported for now
+      // Postgres automatically aborts the entire transaction if one query is canceled
+      if (!isMysql(knex)) {
+        return this.skip();
+      }
+
+      const getProcessesForDriver = {
+        pg: async () => {
+          const results = await knex.raw('SELECT * from pg_stat_activity');
+          return _.map(_.filter(results.rows, {state: 'active'}), 'query');
+        },
+        mysql: async () => {
+          const results = await knex.raw('SHOW PROCESSLIST');
+          return _.map(results[0], 'Info');
+        },
+        mysql2: async () => {
+          const results = await knex.raw('SHOW PROCESSLIST');
+          return _.map(results[0], 'Info');
+        },
+      };
+
+      const driverName = knex.client.driverName;
+      if (!Object.prototype.hasOwnProperty.call(getProcessesForDriver, driverName)) {
+        throw new Error('Missing test query for driverName: ' + driverName);
+      }
+
+      const getProcesses = getProcessesForDriver[driverName];
+
+      return knex.transaction(async (trx) => {
+        const queryA = trx.raw('SELECT SLEEP(1), "A"');
+        const queryB = trx.raw('SELECT SLEEP(1), "B"');
+
+        const promise = Promise.all([
+          queryA.timeout(500, { cancel: true }),
+          queryB,
+        ])
+
+        await delay(100);
+        const processesBeforeTimeout = await getProcesses();
+        expect(processesBeforeTimeout).to.include(queryA.toString());
+        expect(processesBeforeTimeout).to.not.include(queryB.toString());
+
+        const origAcquireRawConnection = trx.client.acquireRawConnection.bind(trx.client);
+        trx.client.acquireRawConnection = async () => {
+          await delay(600);
+          return origAcquireRawConnection();
+        }
+
+        await expect(promise).to.eventually.be.rejected.and.to.deep.include({
+          timeout: 500,
+          name: 'KnexTimeoutError',
+          message: 'Defined query timeout of 500ms exceeded when running query.',
+        });
+
+        trx.client.acquireRawConnection = origAcquireRawConnection;
+
+        await delay(100);
+        const processesAfterTimeout = await getProcesses();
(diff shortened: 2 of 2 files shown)


lib/dialects/mysql2/index.js
```diff
@@ -20,6 +21,18 @@ class Client_MySQL2 extends Client_MySQL {
...
     return true;
   }
+
+  _cancelQueuedQuery(connectionToKill, queryToKill) {
+    const index = findIndex(
+      connectionToKill._commands.toArray(),
+      (cmd) => cmd.__knexQueryUid === queryToKill.__knexQueryUid
+    );
+    if (index >= 0) {
+      connectionToKill._commands.removeOne(index);
```
martinmacko47, line 31: This quite an invasion into the driver internals. Perhaps we could eventually ask the driver upstream to implement canceling of queued queries. But for now, there is no other way how to remove a query from the queue.

maximelkin: >This is quite an invasion into the driver internals
Dont really like this idea. Can we implement it in drivers first?

maximelkin: Wait, implement this probably bad idea, can we try to make things same for pg and mysql instead?  
Also some related thread was here https://github.com/knex/knex/pull/4324

martinmacko47: @maximelkin 

> Can we implement it in drivers first?

I can definitely try to make a PR to drivers. But first we should be sure this solution is acceptable for us. 

> Wait, implement this probably bad idea, can we try to make things same for pg and mysql instead?

Can you please explain what you mean by making things the same for pg and mysql? How would we make things the same for them?

In MySQL there are two commands:

- `KILL QUERY ?` stops the current query, and keeps both the transaction and the connection active.

- `KILL CONNECTION ?` stops the current query, rolls back the transaction and closes the connection.

In Postgres there are also two commands:

- `pg_cancel_backend(?)` stops the current query, marks the transaction aborted, bug keeps the connection active. The user can roll back to a savepoint and continue with the transaction, or roll back the transaction completly.

- `pg_terminate_backend(?)` stops the current query, rolls back the transaction and closes the connection.

We probably don't want to use `KILL CONNECTION ?` nor `pg_terminate_backend(?)` because that would be a breaking change for timeouts outside transactions. Timeouts outside transactions worked for years already, so there will be a lot of code expecting connections won't be closed after a query times out.

So we have to use `KILL QUERY ?` and `pg_cancel_backend(?)` which have different semantics. We can't make them do the same. We just need to use them somehow to make the timeouts working reliably also in transactions.

Also IMHO it is good to keep the connection open, so the user may still handle and resurrect the transation if one of their queries times out. In postgres they may want to roll back to some savepoint they created before running the slow query, in MySQL they may want to just catch the timeout exception and continue with another query.

lib/dialects/mysql/index.js
```diff
@@ -132,6 +133,7 @@ class Client_MySQL extends Client {
           resolver(obj);
         }
       );
+      query.__knexQueryUid = obj.__knexQueryUid;
```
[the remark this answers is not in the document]
martinmacko47: Alternatively we could keep a reference to driver's query object:
```
obj.__query = connection.query(...)
```
And then use this reference to identify the driver's query by identity. E.g.:
```
query === queryToKill.__query
```

lib/dialects/mysql2/index.js
maximelkin: I suggest adding connection.removeQuery() to mysql/mysql2, which accepts driver Query, and which responses with one of three states: not executed (and cancelled), currently executing (no actions, we should cancel it), not found (probably executed or cancelled by some other way).

currently executing - this case is dangerous, because cancel initialization could take too many time, so next query will be cancelled instead.

Maybe we need lock mechanism to prevent sending any commands to mysql before query die.


Also todo: check concurrent manual cancel + timeout cancel.

martinmacko47: @maximelkin I filed the feature request to both drivers: https://github.com/mysqljs/mysql/issues/2478, https://github.com/sidorares/node-mysql2/issues/1316.

Please provide every possible endorsment for this feature request in the drives on behalf of Knex.

martinmacko47: @maximelkin @kibertoad I have created draft implementations of `Query.dequeue()` for both drivers `mysql` and `mysql2`:
- https://github.com/mysqljs/mysql/pull/2480
- https://github.com/martinmacko47/node-mysql2/pull/1 (I will create an upstream PR for this one after somebody from mysql2 team responds to the issue)

Also I have adapted this PR to use `Query.dequeue()` instead of driver internals. I created a separate PR for it, so we can have both versions arround until `Query.dequeue()` gets merged into the drivers: https://github.com/knex/knex/pull/4444

martinmacko47: @maximelkin 
> currently executing - this case is dangerous, because cancel initialization could take too many time, so next query will be cancelled instead.
> 
> Maybe we need lock mechanism to prevent sending any commands to mysql before query die.

I managed to simulate this race condition in a test. And it seems `connectionToKill.pause()`/`resume()` indeed fixes it. Even thought it was supposedly meant for streams, it actually pauses processing packets from server for non-stream queries as well. As a consequence the driver after `pause()` won't start any new queries until we `resume()` it. Therefore even if the timeouted query happens to complete before we manage to kill it, no new query will start in the transaction, until we `resume()` the connection.

See the test: https://github.com/martinmacko47/knex/pull/7

I have put it into a separate branch until we decide it is working, so we won't intermingle two problems here.

martinmacko47: > I have put it into a separate branch until we decide it is working, so we won't intermingle two problems here.

I've merged martinmacko47#7 both to this PR and to #4444. I've also backported all additional commits from #4444 except the commit using dequeque methods from drivers. So we have both versions ready in case it will take too long to get the dequeque methods upstream to drivers.

Issue #1767, opened 2016-11-02
.fromRaw API variant missing


elhigu:
This could be added for consistency, since most of the other builder methods already have their `.xxxRaw` variants already.

mittalyashu:
Any updates on this feature?

kibertoad:
@mittalyashu Would you like to send a PR?

LucasHFS:
@kibertoad I would like to send a PR. This would be my fist PR, some advice?

kibertoad:
Thank you!

LucasHFS:
I'm new to the lib & this would be my first PR, I'm sutdying the lib, can I consider **.fromRaw** a convenience helper for `.from(knex.raw(query))` like the **.whereRaw** is to **.where** case? I'm accepting advices 😃

LucasHFS:
Made a PR #4429  to solve this, if there's some change needed let me know.

kibertoad:
Released in 1.0.1.

PR #4429, opened 2021-04-12
Creating .fromRaw API & its tests

LucasHFS:
Attemp to solve #1767 . I'm new to the lib and this is my first PR. Any changes or things to add let me know 😃.

### DIFF
--- a/lib/query/querybuilder.js
+++ b/lib/query/querybuilder.js
@@ -203,6 +203,17 @@ class Builder extends EventEmitter {
     return this;
   }
 
+  fromRaw(sql, bindings) {
+    //Verify if is rawInstance
+    const raw = sql.isRawInstance ? sql : this.client.raw(sql, bindings);
+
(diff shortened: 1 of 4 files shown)


kibertoad:
@elhigu What do you think?

lib/query/querybuilder.js
```diff
@@ -203,6 +203,17 @@ class Builder extends EventEmitter {
     return this;
   }
 
+  fromRaw(sql, bindings) {
+    //Verify if is rawInstance
```
elhigu, line 207: why this is not implemented just 

```
const raw = sql.isRawInstance ? sql : this.client.raw(sql, bindings);
return this.from(raw)
```

kibertoad: @LucasHFS Ping?..

OlivierCavadenti:
Superseded by #4781

PR #4444, opened 2021-04-18
Fix cancelling timeouted queued queries in transactions for MySQL using driver dequeue method

### LINKED ISSUE #4415: `.timeout(ms, {cancel: true})` for with queued queries in transaction

martinmacko47:
Resolves #4415 

Addaptation of PR #4416 using draft implementations of `Query.dequeue()` for `mysql` and `mysql2` drivers:
- https://github.com/mysqljs/mysql/pull/2480
- https://github.com/martinmacko47/node-mysql2/pull/1 (I will create an upstream PR for this one after somebody from `mysql2` team responds to the issue)


test/integration/query/additional.js
```diff
@@ -978,6 +978,79 @@ module.exports = function (knex) {
...
+      };
+
+      const driverName = knex.client.driverName;
+      if (!Object.prototype.hasOwnProperty.call(testQueries, driverName)) {
+        throw new Error('Missing test query for dialect: ' + driverName);
+      }
+
+      const query = testQueries[driverName]();
+
+      const getProcessesForDriver = {
```
maximelkin, line 1012: Maybe merge tests about timeout in describe block & put common into it?

maximelkin:
Nice
now waiting for driver maintainers for approve & .pause decision

test/integration/query/additional.js
```diff
@@ -978,6 +978,79 @@ module.exports = function (knex) {
...
+      return knex.transaction(async (trx) => {
+        const promise = Promise.all(_.range(5).map(async () => {
+          await query.timeout(50, { cancel: true }).transacting(trx);
+        }));
+
+        await delay(10);
+        const processesBeforeTimeout = await getProcesses();
+        expect(processesBeforeTimeout).to.include(query.toString());
+
+        await expect(promise).to.eventually.be.rejected.and.to.deep.include({
```
maximelkin, line 1042: not sure about this part, I feel like this test will be green for current master too. Maybe execute 5 queries & ensure what from n-th query all queries (which amount >= 2) are cancelled?  

Also, please add test from this https://github.com/knex/knex/issues/4415#issuecomment-814945828

test/integration/query/additional.js
martinmacko47: @maximelkin Totaly agreed. Tests for timeout need a refactor. But let's postpone it into a separate PR after merging this PR. So we can keep track what is new in this PR and what is just a refactor.

test/integration/query/additional.js
martinmacko47: @maximelkin You are rignt that check `await expect(promise).to.eventually.be.rejected.and.to.deep.include({...})` will pass on the current master as well.

The important check in this test is bellow:

```
        await delay(10);
        const processesAfterTimeout = await getProcesses();
        expect(processesAfterTimeout).to.not.include(query.toString());
```

Which is ensuring that a few milliseconds after the timeout *all* queries are cancelled, i.e. none of those 5 queries is running in DB any more. If you ran this test on the current master it will fail on this later check.

> Also, please add test from this #4415 (comment)

Yep.

martinmacko47:
> now waiting for driver maintainers for approve & .pause decision

@maximelkin @kibertoad Hi. Do you think, is there anything we can do to have those timeouts in transactions on MySQL fixed? Diver maintainers did not respond to my PR https://github.com/mysqljs/mysql/pull/2480, besides the first https://github.com/mysqljs/mysql/issues/2478#issuecomment-819808583. Thanks

PR #4446, opened 2021-04-19
Add support for file open flags for sqlite3 

Boelensman1:
With this pull request it should be possible to open a shared memory database without recompiling sqlite, which has quite a lot of use cases, especially in testing. It should for example allow [this](https://github.com/knex/knex/issues/1871#issuecomment-452342526) solution work even if sqlite3 is not compiled with the SQLITE_USE_URI=1 flag.

I set the default mode selection to the default of node-sqlite3 so all existing code should still work.

Link to documentation PR: https://github.com/knex/documentation/pull/320


kibertoad (approved):

kibertoad:
Thank you!

PR #4451, opened 2021-04-21
Additional tests for JSON (de)serialization

kibertoad:


maximelkin:
Looks good, but why tests failed?

myndzi:
@kibertoad I hope you don't mind, I processed this like any of the other PRs and pushed a couple changes. The first commit makes the tests pass, but the second one actually alters the test to remove sqlite from being tested (details written in the commit, but the TLDR is that it doesn't make sense for this test to run against sqlite)

Integration tests passed locally, just waiting on CI. If you're cool with the changes it's ready to merge, otherwise they should be easy to back out. Assigning to you to either merge or send my way for alterations - I can't request a review from you since you're the author.

test/integration2/util/knex-instance-provider.js
```diff
@@ -118,6 +118,12 @@ const testConfigs = {
       user: 'testuser',
       password: 'testpassword',
       charset: 'utf8',
+      typeCast: function (field, next) {
+        if (field.type == 'JSON') {
+          return JSON.parse(field.string());
+        }
+        return next();
+      },
```
myndzi, line 127: The `mysql` client does allow for automatic parsing of json fields, but it's opt-in only via the typeCast argument. (see https://github.com/mysqljs/mysql/issues/1267) This means that _users_ who want this behavior will also have to do it this way. We could conceivably make this the default, but I seem to recall you had a preference against doing that sort of thing.

kibertoad:
@myndzi shoudl CI be failing?

myndzi:
@kibertoad looks like it was just a flake. I'm not sure what happened between the time I sent it for merge and now - I wouldn't have sent it with failing checks. I guess they got re-run at some point?

Issue #2265, opened 2017-10-11
Database agnostic where ILIKE 


esatterwhite:
As mentioned in Issue #233 @ErisDS  there are no way to do case insensitive string comparison (which would work the same way that ilike in postgres). Would be nice to have one.

EDIT: made description to be about just one single feature (original issue had a lot more discussion around the topic)

esatterwhite:
Generally speaking there is no consistent way to specify the case sensitivity of your queries through knex. 
for example w/ mysql it is case insensitive. and doing `where('a' 'ilike' 'b')` doesn't work, you have to send something like

```javascript
whereRaw(util.format("%s LIKE BINARY '%s'",field, '%' + term + '%'));
```
swapping DB engines at this point is not possible. The query parser need to better handle what `ilike` means.

perhaps just a function ? 

```javascript
knex.ilike(<field>, <value>)
```

elhigu:
I removed copy-paste from original issue and changed the title, because it just would have made this issue to be as unclear about its purpose that the original was.

Since this seems to be just about having `.where` variant for case insensitive query filterin, then the API should be the same way with other `.where` methods. 

`.whereILike(...)` and `.orWhereILike(...)` (probably there is no need for ends with / starts with variants, because they should trivial to implement with ilike and %).

esatterwhite:
Yes, the additional methods aren't strictly needed, I'm pretty sure the wild card character is the same across the board. 

You will need `whereLike` and `orWhereLike` variants though.

elhigu:
Those are kind of different issue not related to case sensitivity, but it will be ok to add them at the same time if one wants to have them as special methods. I've been always just using .where(col, 'like', something) since there is no compatibility problems with that between dbs (well actually I always target postgres only, but anyways I would use it that way even if I had to support multiple dbs).

esatterwhite:
Sent from my BlackBerry 10.                                                                                                                                                                                                                From: Mikael LepistöSent: Wednesday, October 11, 2017 3:46 PMTo: tgriesser/knexReply To: tgriesser/knexCc: esatterwhite; AuthorSubject: Re: [tgriesser/knex] Database agnostic where ILIKE (#2265)Those are kind of different issue not related to case sensitivity, but it will be ok to add them at the same time if one wants to have them as special methods. I've been always just using .where(col, 'like', something) since there is no compatibility problems with that between dbs.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
[...]

PR #4457, opened 2021-04-27
Implementing .whereLike and .whereILike api

LucasHFS:
Attempt to solve #2265 .

### DIFF
--- a/lib/dialects/mssql/query/mssql-querycompiler.js
+++ b/lib/dialects/mssql/query/mssql-querycompiler.js
@@ -479,6 +479,22 @@ class QueryCompiler_MSSQL extends QueryCompiler {
     }
     return offset;
   }
+
+  whereLike(statement) {
+    return `[${
+      statement.column
+    }] collate SQL_Latin1_General_CP1_CS_AS ${this._not(statement, 'like ')}'${
+      statement.value
+    }'`;
+  }
+
+  whereILike(statement) {
+    return `[${
+      statement.column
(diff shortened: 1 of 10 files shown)


kibertoad:
Thank you! Could you also please add some integration tests that actually retrieve some data? see `integration2` directory for good examples.

lib/dialects/mssql/query/mssql-querycompiler.js
```diff
@@ -479,6 +479,22 @@ class QueryCompiler_MSSQL extends QueryCompiler {
     }
     return offset;
   }
+
+  whereLike(statement) {
+    return `[${
```
kibertoad, line 484: Shouldn't this need some escaping for values? I think for columns we typically use `formatter.columnize`, and for a single value `client.parameter`. See `C:\sources\node\knex\lib\query\querycompiler.js` for an example.

lib/dialects/postgres/query/pg-querycompiler.js
```diff
@@ -218,6 +218,18 @@ class QueryCompiler_PG extends QueryCompiler {
   distinctOn(value) {
     return 'distinct on (' + this.formatter.columnize(value) + ') ';
   }
+
+  whereLike(statement) {
```
kibertoad, line 222: Do we still need client-specific implementations of this functions if we use standard knex escaping mechanism? I assume there are no other differences?

OlivierCavadenti:
Superseded by https://github.com/knex/knex/pull/4779

xpuu:
Hey @OlivierCavadenti , thanks for your effort! What about orWhereLike and orWhereILike?

PR #4464, opened 2021-05-02
Fixed error when running in browser. Dissabled Migrator and Seeder when running in browser.

### LINKED ISSUE #4441: Uncaught TypeError: The "original" argument must be of type Function
# Environment

Knex version:0.95.4
Database + version:pg+
OS:win10 + chrome  89.0.4389.114

# Bug

1. Can't initialize knex. In fact, when you open site http://knexjs.org, the console will print this error too.

2. Error message

```
util.js?3022:602 Uncaught TypeError: The "original" argument must be of type Function
    at promisify (util.js?3022:602)
    at eval (fs.js?eba2:8)
    at Object../node_modules/knex/lib/migrations/util/fs.js (chunk-vendors.js:6026)
    at __webpack_require__ (main.js:849)
    at fn (main.js:151)
    at eval (template.js?8f9c:3)
    at Object../node_modules/knex/lib/migrations/util/template.js (chunk-vendors.js:6059)
    at __webpack_require__ (main.js:849)
    at fn (main.js:151)
    at Object.eval (MigrationGenerator.js?0437:2)
```

3. Just simple code:
```
  client: 'pg',
  version: '7.2',
  connection: {
    host : '127.0.0.1',
    user : 'your_database_user',
    database : 'myapp_test'
  }
});
```

hananf11:
Knex will now work correctly with the browser, fixes #4441.
make-knex.js checks if window is undefined before adding migrate and seed to KNEX_PROPERTY_DEFINITIONS.
closes #4441

### DIFF
--- a/lib/knex-builder/make-knex.js
+++ b/lib/knex-builder/make-knex.js
@@ -1,7 +1,5 @@
 const { EventEmitter } = require('events');
 
-const { Migrator } = require('../migrations/migrate/Migrator');
-const Seeder = require('../migrations/seed/Seeder');
 const FunctionHelper = require('./FunctionHelper');
 const QueryInterface = require('../query/method-constants');
 const merge = require('lodash/merge');
@@ -45,20 +43,6 @@ const KNEX_PROPERTY_DEFINITIONS = {
     configurable: true,
   },
 
-  migrate: {
-    get() {
-      return new Migrator(this);
-    },
-    configurable: true,
-  },
-
-  seed: {
-    get() {
-      return new Seeder(this);
-    },
-    configurable: true,
-  },
-
   fn: {
     get() {
       return new FunctionHelper(this.client);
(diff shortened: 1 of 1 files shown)


rathboma:
+1 can a maintainer weigh in here? It'd be great to restore working browser behavior.

kibertoad:
This is a draft and also looks somewhat messy. Is it possible to structure it in a somewhat less hacky way?

hananf11:
This pull request is a draft because I after I 'fixed' the problem I realised this solution has a few issues.

This solution is not ideal because it just checks if window is undefined. I can imagine this would lead to problems, for example the user for some reason decides they want to use the migrator or seeder in the browser. 
I think that a much cleaner solution would be to add options to disable the migrator and/or seeder when creating a knex instance, I think this would also mean that the migrator and seeder code could be removed all together when tree shaking. 
The other issue is that yes, it is hacky, and I can't work out how to write it better.
I don't have a great deal of knowledge about JavaScript, and even less about how knex works.

PR #4465, opened 2021-05-02
CI: add Node 16 to version matrix

stscoundrel:
Node 16 was released short time ago, and became the "current" version. Knex readme states that all Node versions 10+ are supported, but CI is not currently run for Node 16.

Adds Node 16 to version matrixes. Everything seems to pass just fine, as seen [here](https://github.com/stscoundrel/knex/pull/1)

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


kibertoad:
@stscoundrel Can you also remove Node 15?

stscoundrel:
Sure thing @kibertoad, removed 15.

kibertoad (approved):

kibertoad:
Thanks!

Issue #2544, opened 2018-03-25
Inconsistencies between `select` and `returning`


migueloller:
# Environment

Knex version: v0.14.4
Database + version: pg@7.4.1
OS: macOS v10.13.3

# Feature discussion / request

1. Explain what is your use case

When using `returning` it currently returns an array of values or an array of objects depending on whether one or more columns are specified. To be consistent with `select`, it should always return an array of objects.

2. Explain what kind of feature would support this

All features that support `returning`.

3. Give some API proposal, how the feature should work

The API would stay the same, but instead of `returning('column_name')` returning an array of values, it would return an array of objects, each with the key `column_name` and its value—just like `select`.

wubzz:
I think this is a good idea, especially with #2571 in mind. This is also a breaking change.

```javascript
const userId = await knex(table).insert({name: 'knex'}).returning('id');
//Old: 1
//New: {id: 1}
```

Also optimal to at least support:
```javascript
const ids = await knex(table).update({value: 1}).returning('id').map('id');
//[1, 2 , 3, 4, 5]
```

elhigu:
I don't have strong opinion about this one. This issue is kind of related too https://github.com/tgriesser/knex/issues/2682 to keep in mind.

Tuupertunut:
Yes please do this. It has causes me some hard-to-debug issues.

kibertoad:
@Tuupertunut Would you consider submitting a PR for this?

Tuupertunut:
No I don't know the codebase well enough, at least yet.
For Postgres I think the relevant code is here: https://github.com/knex/knex/blob/e37aeaa31c8ef9c1b07d2e4d3ec6607e557d800d/lib/dialects/postgres/index.js#L261-L266
I think you could find a similar piece of code for other database engines as well, but I'm not sure.

mercmobily:
Closing this in favour of #6254 which closes the exact same issue

Issue #2337, opened 2017-11-17
multiple .returning can't be used, unlike .select


caub:
knex: 0.14
OS: linux
DB: postgresql 9.6

**Problem**:
`.returning` overwrites a previous one:
```js
knex('bla').returning('*').returning({foo: 'bar'}).insert({ foo: 'bar' })+''
// 'insert into "bla" ("foo") values (\'bar\') returning "bar" as "foo"'
```
I expected it to work like `.select`:
```js
knex('bla').select('*').select({foo: 'bar'})+''
// 'select *, "bar" as "foo" from "bla"'
```
`*` is preserved there

elhigu:
That make sense for consistency. Also merge style of functionality should be also documented.

elhigu:
But this is breaking change so I wouldnt like to get this implemented until few more months.

caub:
thanks ok, no worry

kurtzilla:
not sure if this is relevant to your exact use-case, but I found this to be helpful to alias some return columns...
```
knex('messages')
  .column('*', { createdAt: 'created_at' })
  .where({ some_id: someId })
  .orderBy('created_at', 'ASC');
```
returns all of the columns with an additional aliased "createdAt" column

stephenhuh:
i can pick this up, definitely something i wanted in my use of knex.

duzymaju:
Talking about object mappings in `returning` for postgres, the only thing to do to let it works as described is to modify this conditional statement:

https://github.com/knex/knex/blob/e37aeaa31c8ef9c1b07d2e4d3ec6607e557d800d/lib/dialects/postgres/index.js#L261-L266

adding ` || _.isPlainObject(returning)` (or more sophisticated condition) to it.

It would then allow to create
```
knex('table').insert({ example_id: 123 }).returning({ id: 'example_id' })
```
receiving
```
[{ id: 123 }]
```
instead of
```
['']
```

I can prepare proper pull request if only there is a will to introducing such functionality. What do you think?

kibertoad:
@duzymaju Sure, that would be great!

Issue #2571, opened 2018-04-16
How to use raw() in returning?


qwang07:
# Environment

Knex version: 0.14.6
Database + version: PostgreSQL 10.0
OS: Debian 9

How should I use `knex.raw()` in `returning()` function? The output of those two statements are the same:
```
knex('users').insert({ id: 1, firstname: 'test', lastname: 'test' }).returning(['id', 'firstname', 'lastname']).toString()
knex('users').insert({ id: 1, firstname: 'test', lastname: 'test' }).returning(knex.raw('"id", "firstname", "lastname"')).toString()
```
However, the first one can return the data correctly, the second will return `[ undefined ]`. Because I want to use SQL function `concat_ws()` in returning statement, I have to use `raw()` to create the query.

elhigu:
You are using it correctly (except from using ?? bindings for quoting column names) and since output of the queries is the same they do return the same result. 

You have some other problem in your code, please provide complete reproduction code for your problem (with db initialization etc.). https://github.com/tgriesser/knex/blob/master/CONTRIBUTING.md#what-is-minimal-code-to-reproduce-bug-and-why-i-have-to-provide-that-when-i-can-just-tell-whats-the-problem-is

qwang07:
My `.env` file:
```
PGHOST=localhost
PGPORT=5432
PGUSER=qwang
PGPASSWORD=qwang123
PGDATABASE=test_db

ADMIN_USER=db_admin
ADMIN_PASS=Admin123
ADMIN_MOBILE=0000000000
ADMIN_EMAIL=qwang0617@gmail.com
```

Here is the test file:
```
const { createHash, createSalt, head, prop } = require('lib/utils')
const db = require('knex')({
  client: 'pg',
  connection: {
    host: process.env.PGHOST,
    user: process.env.PGUSER,
    password: process.env.PGPASSWORD,
    database: process.env.PGDATABASE,
    port: process.env.PGPORT
  },
})

const salt = createSalt()
const user1 = {
  username: 'test',
  hash: createHash('pwd', salt),
  salt,
  type: 1,
  store: null,
  contact: 1,
}

const user2 = {
  username: 'test2',
  hash: createHash('pwd', salt),
  salt,
  type: 1,
  store: null,
  contact: 1,
}

const s = async () => {
  console.log(db('users')
    .insert(user1)
    .returning(db.raw(`??`, [
      [
        'id',
        'username',
        'type',
        'locked',
      ]
    ])).toString())

[...]

PR #4471, opened 2021-05-06
Inconsistencies between .select and .returning

LucasHFS:
attempt to solve #2544 and make the #2337 functionality work


LucasHFS:
Yes, I was searching for something to contribute with. I started by this #2544. Then I saw #2337 after a few days that this PR covers.

LucasHFS:
Searching for issues to solve, This PR solves #2571 as well

fdevibe:
I would really appreciate it if this was merged. As described in #2571, this also prevents returning geometry in a usable format, like with `ST_AsGeoJSON()` (AFAICT).

kibertoad:
I'll look into this in the upcoming weeks. Considering that Objection still hasn't updated to the latest semver major, I think there is still space to squeeze in another breaking change.

kibertoad:
@OlivierCavadenti Can you take this one?

kibertoad:
@OlivierCavadenti That happens if you rerun migration after deleting some migration files that were run previously. Consider deleting knex changelog table

kibertoad:
[object] thing looks very sus, though, probably we broke logging after changing migration file structure, but maybe it comes from other branch being used before

kibertoad:
ah, so ci has it too. guess we broke migrator logic for fetching previously ran migrations

OlivierCavadenti:
> ah, so ci has it too. guess we broke migrator logic for fetching previously ran migrations

... in fact I miss to report one test correction with the rebase... now seems good (excepted Lint issue)

kibertoad (approved):

kibertoad:
@OlivierCavadenti Incredible work, thank you so much! I will fix the linting issue tomorrow myself.
Do we need to update documentation after this change?

OlivierCavadenti:
> @OlivierCavadenti Incredible work, thank you so much! I will fix the linting issue tomorrow myself. Do we need to update documentation after this change?

Sure : https://github.com/knex/documentation/pull/378

kibertoad:
Released in 1.0.1.

PR #4480, opened 2021-05-11
Remove peerDependencies to avoid auto-install on npm 7

### LINKED ISSUE #4437: remove peerDependencies and keep peerDependenciesMeta only.
The npm v7 will install peerDependencies automaticly, it will install mysql,mysql2,pg,sqlite3,tedious, so just keep peerDependenciesMeta will be suitable.

kibertoad:
fixes #4437

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,13 +56,6 @@
     "tarn": "^3.0.1",
     "tildify": "2.0.0"
   },
-  "peerDependencies": {
-    "mysql": "^2.18.1",
-    "mysql2": "^2.2.5",
-    "pg": "^8.5.1",
-    "sqlite3": "^5.0.0",
-    "tedious": "^11.0.5"
-  },
   "peerDependenciesMeta": {
     "tedious": {
       "optional": true


B4nan:
This seems wrong to me, NPM 7 should actually respect `peerDependenciesMeta`, but without the real `peerDependencies` I don't think it has any meaning. 

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

I'd expect that doing this might result in wrong dependency resolutions (the peer deps won't be considered as peer deps). In other words, I'd expect things like yarn pnp to be broken as knex should not see those dependencies at all under strict environments. cc @merceyz maybe I got it wrong or I missed something.

kibertoad:
@B4nan `Marking a peer dependency as optional ensures npm will not emit a warning if the soy-milk package is not installed on the host.`

Does it actually prevent npm7 from autoinstalling the dependency, though?

https://betterprogramming.pub/the-step-by-step-guide-to-understanding-and-adopting-npm-7-914504f7090f says nothing about respecting the meta.

B4nan:
I believe it does, see https://github.com/npm/rfcs/discussions/289

> peerOptional deps are deps that are listed in peerDependencies and also have an entry in peerDependenciesMeta containing {"optional": true}. These sorts of dependencies are not installed by default, but if installed, must be resolved to a version that satisfies the optional peer dependency relationship.

kibertoad:
@B4nan thanks, I'll gve it a test today and will release a hotfix.

merceyz:
> I'd expect that doing this might result in wrong dependency resolutions (the peer deps won't be considered as peer deps). In other words, I'd expect things like yarn pnp to be broken as knex should not see those dependencies at all under strict environments. cc @merceyz maybe I got it wrong or I missed something.

Yarn treats `peerDependenciesMeta.<name>.optional: true` as an implicit `peerDependencies.<name>: "*"` if one isn't specified https://github.com/yarnpkg/berry/blob/0f7644f6d9cd219b532816ab6250f6cb63426486/packages/yarnpkg-core/sources/Configuration.ts#L1533-L1543

PR #4484, opened 2021-05-12
Add RegExp to match unescaped single quotes

### LINKED ISSUE #4481: MySQL Back-end does not appear to escape ' in enum values.
# Environment

Knex version: 0.95.4
Database + version: MariaDB 10.5.9
OS: Windows 10

# Bug

When building tables with enum columns, providing a value list that contains `'` within the values results in a SQL Syntax error triggered by the `'`.  Taking the generated query and escaping the values results in a functional query.

## Resulting error: 
Standard SQL Syntax error
> ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near [then the text immediately following the `'`].

## Reproducing code sample
With a valid connection:

```
(async function run() {
  await knexSqlite.schema.createTable('test', (t) => {
    t.enu('foo', ["inavlid's value", "valid value"]);
  });
})();
```

ramos-ph:
**Context**

There is a case where the MySQL dialect doesn't parse the query correctly. Here's an example:

```js
await knex.schema.createTable("test", (t) => {
  t.enu("foo", ["invalid's value", "valid value"]);
});
``` 

The lack of parsing on this context where a single quote is wrapped between double quotes was causing an SQL syntax error. This PR fixes this.

**How it was done**

Add a loop to match and escape single quotes to prevent SQL syntax errors.

Created a RegExp that matches only the single quotes that:

- Doesn't come after a parentheses
- Doesn't come before a parentheses
- Doesn't come before a comma
- Doesn't come before a space

If desired, please refer to https://regexr.com/5spjb to validate the behavior of the RegExp.

**Aditional Info**

It seems to be specific to the MySQL dialect. Reading the code for other dialects, I couldn't see any lack of validation for this specific scenario. 

Fixes #4481

### DIFF
--- a/lib/dialects/mysql/schema/mysql-tablecompiler.js
+++ b/lib/dialects/mysql/schema/mysql-tablecompiler.js
@@ -17,6 +17,19 @@ class TableCompiler_MySQL extends TableCompiler {
       ? 'create table if not exists '
       : 'create table ';
     const { client } = this;
+
+    // Fixes possible single quotes on columns
+    for (let i = 0, l = columns.sql.length; i < l; i++) {
+      // Matches all of the single quotes that:
+      // - Doesn't come after a parentheses
+      // - Doesn't come before a parentheses
+      // - Doesn't come before a comma
+      // - Doesn't come before a space
+      // https://regexr.com/5spjb
+      const re = /(?<![(\s])(')(?![,)])/g;
+      columns.sql[i] = columns.sql[i].replace(re, "\\'");
+    }
+
     let conn = {};
     let sql =
       createStatement + this.tableName() + ' (' + columns.sql.join(', ') + ')';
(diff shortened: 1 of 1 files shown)


acrosman:
I don't know the larger context of this code path, what automated tests may exist for it, nor general coding standards of the project, so some of the following feedback maybe invalid but here are my thoughts looking at the suggested code:

The regex only checks for `'`, but as I understand the escaping issues here there are several other characters that also need to be checked (see [this SO answer](https://stackoverflow.com/a/7760578/24215) for reference).

The new code also assumes that the values being provided as an array of strings. But I given the existing code an array of integers (and several other numeric primitive types) would probably have worked previously and now fail. So type checking before calling replace is probably in order.

Those things said, it seems likely to me that Knex already has some form as string escaping included elsewhere in the project. So I would assume those functions should be leveraged to handle these values.

ramos-ph:
Thank you for your feedback! I didn't know that the MySQL escaping were deeper.
Before carrying on, I would like to see the feedback from the maintainers as well. For what I could notice, the query object at this level have a specific format:
```
`foo` enum('inavlid's value', 'valid value')
```
Therefore I didn't saw the need to check types, since it comes that way from about two or three function calls.

kibertoad:
hey! is this still relevant? I can potentially merge it today

PR #4491, opened 2021-05-19
Include options object in count(Distinct) typings

### LINKED ISSUE #4165: Typings of aggregate functions (count, max etc.) are missing options object ("as" alias)
# Environment

Knex version: 0.21.13
Database + version: N/A
OS: N/A

Issue is about TypeScript definitions, @lorefnon 

# Bug

As stated in the documentation, `count`, `min` etc. take an optional options object, with the key `as` specifying an alias for the aggregate value. However, the types `AsymmetricAggregation` and `TypePreservingAggregation` only have the `...columnNames` parameter; the options object is missing from the typings.

tkalliom:
Fixes #4165

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1413,6 +1413,13 @@ export declare namespace Knex {
   // is significant.
 
   interface AsymmetricAggregation<TRecord = any, TResult = unknown[], TValue = any> {
+    <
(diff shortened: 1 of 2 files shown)


kibertoad (approved):

kibertoad:
Thank you!

PR #4502, opened 2021-05-26
Allow empty onConflict to not specify conflicting columns

santialbo:
Both [postgres](https://www.postgresql.org/docs/9.5/sql-insert.html) and [sqllite](https://sqlite.org/lang_upsert.html) allow to not pass the conflicting columns. Right now knex doesn't allow this use case: if you pass nothing to onConflict a query like the following gets generated
```
... on conflict (true) ...
```
which of course fails because it expects a column named "true" to exist.

This PR adds support for this missing use case.

Note: TS types already allow not passing parameters to `onConflict`

CC'ing @nicoburns since he was the main author of this feature.

### DIFF
--- a/lib/dialects/postgres/query/pg-querycompiler.js
+++ b/lib/dialects/postgres/query/pg-querycompiler.js
@@ -76,10 +76,20 @@ class QueryCompiler_PG extends QueryCompiler {
   }
 
   _ignore(columns) {
-    return ` on conflict (${this.formatter.columnize(columns)}) do nothing`;
+    let sql = ' on conflict';
+    if (columns.length > 0) {
+      sql += ` (${this.formatter.columnize(columns)})`;
+    }
+    sql += ' do nothing';
+    return sql;
   }
 
   _merge(updates, columns, insert) {
+    if (columns.length === 0) {
+      throw new Error(
+        'A conflict target must be set when using ON CONFLICT DO UPDATE'
+      );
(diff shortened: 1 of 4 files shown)


nicoburns:
Oh interesting. I was under the impression that specifying the columns (or unique constraint name) was mandatory.

What is the behaviour when the columns are not specified? It uses all unique constraints? Or just rhw primary key?

The documentation probably ought to be updated along with this PR, as it currently specifies that the columns are mandatory for postgres and sqlite. The typescript types should be checked too.

santialbo:
Hi @nicoburns and thanks for your comment. If no columns are specified it uses all unique indices and constraints.
Now that I pay more attention, the docs for postgres say
> For ON CONFLICT DO NOTHING, it is optional to specify a conflict_target; when omitted, conflicts with all usable constraints (and unique indexes) are handled. For ON CONFLICT DO UPDATE, a conflict_target must be provided.

For sqlite aparently it's ok to `ON CONFLICT DO UPDATE`.

I've updated my PR to handle this but I'm unsure this would be the preferred *knexy* way of doing it.

Typescript types already allow not passing any arguments to `.onConflict`

kibertoad:
@santialbo Is this a duplicate for #4557?

santialbo:
@kibertoad it is

kibertoad:
Released in 0.95.7

PR #4508, opened 2021-05-28
adding with clause in mysql dialect

### LINKED ISSUE #4428: with() doesn't work with update statements?
# Environment

Knex version: 0.95.2
Database + version: MySQL 8.0.22
OS: WSL Ubuntu 20

# Bug

When using `with()` with `update()`, the `with` statement doesn't appear in the query.

2. Error message

None

```
knex('foo')
  .with('foo2', builder => builder.select('*').from('foo'))
  .update({ col: 1 });
```

Expected query:

```
with `foo2` as (select * from `foo`) update `foo` set `col` = 1
```

Actual query:

```
update `foo` set `col` = 1
```

**Edit:**

It looks like this can be fixed by adding `this.with()` to `QueryCompiler_MySQL.update`

LucasHFS:
attempt to fix #4428


kibertoad:
Thanks! Could you add integration test for this?

LucasHFS:
Ok, WIP.

OlivierCavadenti:
note for me: need to add integration update with test...

kibertoad (approved):

OlivierCavadenti:
@kibertoad seems one test fail in Mssql, I will fix it later

Issue #4132, opened 2020-11-24
Error: stream destroyed


giovanni-bertoncelli:
# Environment

Knex version: 0.21.12
Nodejs version: 14.15.1
Database + version: MariaDB 10.3.27
OS: Centos 7

# Bug

I encountered this issue after updating knex from version 0.16.3 to version 0.21.12 and NodeJS from 6.9.0 to 14.15.1, before that I have never see this error.

1. I'm trying to retrieve a large amount of data from a SELECT statement by using the stream API:
```javascript
    return new Promise((res) => {
        let query = knex(knex.ref('table').as('l'))
            .distinct('l.*');
        query = addQueries1(query);
        query = query.where(whereFilter);
        query = addQueries2(query);
        lib.log.info('QUERYING table with: ' + query.toString());
        return query.stream(res);
    });
```
After that I simply listen on events like `stream.on('data', (chunk) => {})` or `'error'` or `'end'`. I'd like to specify that there no other call on the stream rather than the `on` listeners (neither `destroy()` or `pipe()`).
2. After some random time the stream triggers an error on the `on('error')` listener, that goes like this:
```
nov 24 10:38:55 dev2.eplate.local npm[31646]: [CHILD][PID: 32545] - [INFO] [logger] - Error: stream destroyed
nov 24 10:38:55 dev2.eplate.local npm[31646]: at Transform.destroy [as _destroy] (/opt/nodejs/eplate_nodejs/node_modules/knex/lib/runner.js:91:31)
nov 24 10:38:55 dev2.eplate.local npm[31646]: at Transform.destroy (internal/streams/destroy.js:39:8)
nov 24 10:38:55 dev2.eplate.local npm[31646]: at finish (_stream_writable.js:670:14)
nov 24 10:38:55 dev2.eplate.local npm[31646]: at processTicksAndRejections (internal/process/task_queues.js:80:21)
```
3. Demo code:
```javascript
const promise = new Promise(res => {
        let query = knex(knex.ref('table').as('l'))
            .distinct('l.*');
        // ensure a large set of data
        query = query.where({id: {lte: 100000}});
        lib.log.info('QUERYING table with: ' + query.toString());
        return query.stream(res);
});
promise.then(s => {
s.on('error', e => {console.error(e);});
s.on('data', c => {console.log(c);});
s.on('end', () => {console.log('end!');});
});
```

# Other details
I'm not able to reproduce this issue on a different OS...

maximerichrd:
I had the same error : 
[...]

PR #4532, opened 2021-06-21
Fix error message when stream automatically destroyed

blinkzz:
This PR addreses #4132 and simillar issues

Starting from 14.0.0 Node.js has autoDestroy: true for all readable and writable streams by default and this causes error when stream is getting destoyed after finish, because Knex was always passing error message despite this natural behaviour

All kudos to @leedm777 for solution

### DIFF
--- a/lib/execution/runner.js
+++ b/lib/execution/runner.js
@@ -68,13 +68,14 @@ class Runner {
       transform: (chunk, _, callback) => {
         callback(null, this.client.postProcessResponse(chunk, queryContext));
       },
-      destroy() {
+      destroy(err) {
         // For some reason destroy is not available for mssql on Node 14. Might be a problem with tedious: https://github.com/tediousjs/tedious/issues/1139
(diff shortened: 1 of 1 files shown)


kibertoad:
@blinkzz Thank you! Could you add a test for this? I can help with running them locally if needed.

blinkzz:
@kibertoad, i can see that there is already some stream related streams defined in [integration/query/selects.js](https://github.com/knex/knex/blob/master/test/integration/query/selects.js), is there any reason that they are not running as part of CI process, i wonder if they have shown this regression in first place

kibertoad:
@blinkzz What makes you think they are not running?
See `test/integration/suite.js` -> `require('./query/selects')(knex);`

Maybe you have some specific test in mind that is currently skipped for PostgreSQL? If so, feel free to remove the skipping part in this PR.

OlivierCavadenti:
@kibertoad can you give me help about how to test this ?

PR #4534, opened 2021-06-23
log error message

ysosiak:
When `stack` and `detail` properties are not available, only `undefined` is printed in the console which makes it extremely difficult to diagnose the issue. This change logs `messsage` property if set

Can be replicated when using `envvar` lib  and  [UnsetVariableError](https://github.com/plaid/envvar/blob/master/index.js#L17  ) is [thrown](https://github.com/plaid/envvar/blob/master/index.js#L39)


kibertoad (approved):

kibertoad:
Thank you!

kibertoad:
Released in 0.95.7

PR #4536, opened 2021-06-25
Add DPI error codes for Oracle

vpeltola:
Add DPI-1010 and DPI-1080 error codes required for oracledb >=5.2.0

### DIFF
--- a/lib/dialects/oracle/utils.js
+++ b/lib/dialects/oracle/utils.js
@@ -43,6 +43,8 @@ ReturningHelper.prototype.toString = function () {
 // mark the connection as failed
 function isConnectionError(err) {
   return [
+    'DPI-1010',  // not connected
+    'DPI-1080',  // connection was closed by ORA-%d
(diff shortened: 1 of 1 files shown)


kibertoad:
@cjbj Does this look correct?

cjbj:
@kibertoad by eyeball, it seems fine.  The full error number list is a bit over-the-top (and wouldn't be needed for new code that targets node-oracledb 5.2+), but if this case statement is only checked on error conditions, there isn't any harm.  

I would question whether NJS-040 should be on the list - there is no connection in this case.

kibertoad:
@vpeltola Can you drop the NJS-040?

vpeltola:
Removed NJS-040 as requested.

kibertoad (approved):

kibertoad:
Thanks!

kibertoad:
Released in 0.95.8

PR #4550, opened 2021-07-01
fix migrate:up for migrations disabling transactions

michaelbrewerdavis:
knex migrate:up <migration>

does not respect configuration

    export const config = { transaction: false }

due to treating the return value of getMigration as a value rather than a promise.

I could use advice on where to add tests for this behavior.

### DIFF
--- a/lib/migrations/migrate/Migrator.js
+++ b/lib/migrations/migrate/Migrator.js
@@ -113,7 +113,7 @@ class Migrator {
         }
         return value;
       })
-      .then(([all, completed]) => {
+      .then(async ([all, completed]) => {
         const newMigrations = getNewMigrations(
           this.config.migrationSource,
           all,
@@ -142,7 +142,7 @@ class Migrator {
           useTransaction:
             !migrationToRun ||
             this._useTransaction(
-              this.config.migrationSource.getMigration(migrationToRun)
+              await this.config.migrationSource.getMigration(migrationToRun)
             ),
         };
       })


jeff-permiso:
Bump

kibertoad:
@jeff-permiso Sorry for the delay.
Mixing promise chain with an async/await is not the best solution. Could you please rewrite the whole block to use async/await?
Also this change is missing a test.

michaelbrewerdavis:
@kibertoad Sorry for the delay -- I've added tests around single up/down migrations without transactions, and refactored the method as requested.

michaelbrewerdavis:
@kibertoad Any feedback on the tests or refactoring?

Thanks much.

test/integration/migrate/test_single_per_migration_trx_disabled/down.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const { isPostgreSQL } = require('../../../util/db-helpers');
+
+exports.up = async function () {};
+
+exports.down = async function (knex) {
+  if (isPostgreSQL(knex)) {
```
kibertoad, line 8: this doesn't seem to be the right place for skipping the test, could you move it into a test itself? there are plenty of examples of how db-specific tests are handled

test/integration/migrate/migration-integration-tests.js
```diff
@@ -665,71 +672,122 @@ module.exports = function (knex) {
...
+          await knex.migrate.rollback(
+            {
+              directory:
+                'test/integration/migrate/test_single_per_migration_trx_disabled',
+            },
+            true
+          );
+        });
+
+        it('should run', async () => {
```
kibertoad, line 777: does this test those migrations to simply not fail with an error? can we also assert that they actually did what we expect them to?

test/integration/migrate/migration-integration-tests.js
```diff
@@ -665,71 +672,122 @@ module.exports = function (knex) {
...
+                    .select('*')
+                    .then((data) => {
+                      expect(data).to.have.length(0);
+                    });
+                });
+            });
+        });
+
+        it('should not error if all migrations have already been undone', function () {
+          return knex.migrate
```
kibertoad, line 752: can this be rewritten into async/await?

test/integration/migrate/migration-integration-tests.js
michaelbrewerdavis: The migrations don't actually do anything other than throw an error if called within a transaction.  Added a comment to that effect.

test/integration/migrate/test_single_per_migration_trx_disabled/down.js
michaelbrewerdavis: Moved to test

test/integration/migrate/migration-integration-tests.js
michaelbrewerdavis: Updated

michaelbrewerdavis (commented):
Updated PR with requests

kibertoad:
What is PostgreSQL-specific for these tests, btw? Aren't transactions pretty universal?

michaelbrewerdavis:
> What is PostgreSQL-specific for these tests, btw? Aren't transactions pretty universal?

The method I was using to test being in a transaction wasn't.  Rewrote to make more universal.

kibertoad (approved):

kibertoad:
Thanks a lot!

PR #4553, opened 2021-07-05
Add camelCase to timestamps method

Vilkas67:


lib/schema/tablebuilder.js
```diff
@@ -46,15 +46,29 @@ class TableBuilder {
...
-  timestamps() {
-    const method = arguments[0] === true ? 'timestamp' : 'datetime';
-    const createdAt = this[method]('created_at');
-    const updatedAt = this[method]('updated_at');
-    if (arguments[1] === true) {
-      const now = this.client.raw('CURRENT_TIMESTAMP');
-      createdAt.notNullable().defaultTo(now);
-      updatedAt.notNullable().defaultTo(now);
+
+  timestamps(param1, param2, useCamelCase) {
```
kibertoad, line 50: Could you please rename `param1` and `param2` to how they are called in knex documentation, and also make `useCamelCase` a part of general params object, that would be the third function argument? We shouldn't keep adding positional parameters, so introducing a general param wrapping object would be helpful here.

lib/schema/tablebuilder.js
```diff
@@ -46,15 +46,29 @@ class TableBuilder {
...
+    const camel = useCamelCase === true;
```
kibertoad, line 51: Why do you need additional variable here?

lib/schema/tablebuilder.js
```diff
@@ -46,15 +46,29 @@ class TableBuilder {
...
+    if (camel) {
+      const createdAt = this[method].camelCase('created_at');
+      const updatedAt = this[method].camelCase('updated_at');
+      if (param2 === true) {
+        const now = this.client.raw('CURRENT_TIMESTAMP');
+        createdAt.notNullable().defaultTo(now);
+        updatedAt.notNullable().defaultTo(now);
+      }
     }
+    {
```
kibertoad, line 62: This is too much duplication for no good reason. The only different part is the name of the fields, and you can actually camelCase them manually

kibertoad:
Tests are missing. Check https://github.com/knex/knex/tree/master/test/integration2 for examples of how to write multi-db tests. There is also a npm script available in package.json for running databases locally (pick the no-oracle one).

I think that `dropTimestamps` method should also be updated to support camelCase option.

OlivierCavadenti:
I will make another PR...

Issue #4552, opened 2021-07-02
on postgres .onConflict().ignore() should be able to take no columns


calvinmetcalf:
# Environment

Knex version:0.95.4
Database + version: postgres 12.7
OS: OSX

# Bug
running
```js
knex('table').insert({something:1}).onConflict().ignore();
```

generates

```sql
insert into "table" (something) values($1) on conflict ("true") do nothing;
```

when it should generate

```
insert into "table" (something) values($1) on conflict do nothing;
```

it looks like a fairly simply fix [around here](https://github.com/knex/knex/blob/228b49f4afb7d2cb03f4219b6cb0d1bb854d60f8/lib/dialects/postgres/query/pg-querycompiler.js#L79) which I might be able to open a pr for next week, amusing my wife doesn't give birth this weekend.

calvinmetcalf:
update: she hasn't so I opened up a pull #4557

kibertoad:
Released in 0.95.7

PR #4557, opened 2021-07-06
add ability to omit columns on an onConflict().ignore()

calvinmetcalf:
pull request to handle #4552. Adds the ability to omit the column list for onConflict in postgres and sqlite3 (mysql currently only supports omitting the columns).

the column list is used as a proxy for whether an onConflict exists, so that's why it's true if there are no columns, we could try making it something more explicit like a symbol if we think there could be a problem.


calvinmetcalf:
no updates to the types are necessary as columns were already optional in mysql

kibertoad (approved):

kibertoad:
Thank you!

Issue #873, opened 2015-06-22
Is it impossible to 'delete join'?


vivaladiem:
Hi
I want to use delete join
but this query doesn't work.

'''''''
knex('user_answer').join('answers', 'answers.id', '=', 'user_answer.answer_id')
            .where('user_answer.user_id', userId).andWhere('answers.question_id', questionId).del();
'''''''
this ignore join statement.

Is it impossible? should I use raw query?

elhigu:
Can you write that query directly with SQL? If that works, then check what `toString()` outputs for your knex query to see if there is a problem in knex generated SQL.

marcbachmann:
In postgres you can use the `USING` keyword in combination with a where statement.
Joins aren't supported in a delete statement: http://www.postgresql.org/docs/8.2/static/sql-delete.html

```
knex
.delete()
.from(knex.raw('user_answer USING answers'))
.where(knex.raw('answers.id = user_answer.answer_id'))
.andWhere('user_answer.user_id', userId)
.andWhere('answers.question_id', questionId)
```

chrisbroome:
You can also use a sub-select to delete rows. For example, if I have a one-to-many between `book` and `chapter`, and I want to remove all books that don't have chapters, then the following would work:

``` JavaScript
const qb = knex.from('book')
  .whereNotIn('id', function() {
    return this.from('chapter').distinct('book_id');
  })
  .delete();

console.log(qb.toString());
```
Output:
``` SQL
delete from `book` where `id` not in (select distinct `book_id` from `chapter`)
```

twardnw:
any chance this can get some attention? This type of query is perfectly valid

nfantone:
Any progress here? Issue is five years old now. This looks like a very reasonable use case to me.

kibertoad:
@nfantone 
Is it more than reasonable, it's just that noone implemented it yet, I think. Any help would be welcome.

elhigu:
@nfantone is there some case that cannot be done with subquery instead of join? That might be a reason why anyone hasn't bothered to implement this.

mnahkies:
MySQL at least it isn't able to use indexes effectively for my use case when doing a 
```
DELETE FROM x WHERE fk_id IN (SELECT id FROM y WHERE z = 'foo')
```

Causing a sequential scan of a large table, making the query slow and causing knock on effects to other queries due to excessive locking.

[...]

PR #4568, opened 2021-07-11
#873 | implement delete with join

Egor-Koldasov:
Fix for the issue https://github.com/knex/knex/issues/873
The issue is that joins are not included in delete queries.
Examples can be found in tests: https://github.com/knex/knex/compare/master...Egor-Koldasov:fix/873-delete-join?expand=1#diff-9a54bb8c91fbb672179c076a18af9db40b73bc5c4a3e3ea7cc89c48ea8ebeadfR9944

Notes on the solution:
1. Solution does not include the ability to delete multiple tables. To me, it is not clear, how the interface should be expanded to include this option. It would be great to have a recommendation from collaborators: https://github.com/knex/knex/issues/873#issuecomment-877761010
It does solve other issues, allowing to delete values from a table by filtering values from another joined table.
It also helps with unexpected queries and potential data loss mentioned here: https://github.com/knex/knex/issues/873#issuecomment-808735961

2. Postgres, SQLite and Oracle do not support joins on delete. I did not include any custom error throws for them, instead, I allowed the query to build the same way and let the database report the error. Tests for error are included.

3. The project has a linter added to the commit hook. It updated the whole file: https://github.com/knex/knex/compare/master...Egor-Koldasov:fix/873-delete-join?expand=1#diff-9a54bb8c91fbb672179c076a18af9db40b73bc5c4a3e3ea7cc89c48ea8ebeadf
I'm not sure why previous commits have not done this, since it is automatic.

4. I'm having issues with successfully running 1 test: https://github.com/knex/knex/blob/master/test/integration/query/selects.js#L1361
The timeout error is not being caught properly leading to a general "Timeout of 5000ms exceeded." error.
It is not related to this update, but it's better to let know.

Thanks.

### DIFF
--- a/test/integration/query/deletes.js
+++ b/test/integration/query/deletes.js
@@ -1,6 +1,8 @@
 'use strict';
 
+const { expect } = require('chai');
 const { TEST_TIMESTAMP } = require('../../util/constants');
+const { isSQLite, isPostgreSQL, isOracle } = require('../../util/db-helpers');
 
 module.exports = function (knex) {
   describe('Deletes', function () {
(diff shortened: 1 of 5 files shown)


test/integration/query/deletes.js
```diff
@@ -82,5 +84,61 @@ module.exports = function (knex) {
...
         });
     });
+
+    describe('Delete with join', function () {
+      it('should handle basic delete with join', async function () {
+        const query = knex('test_table_two')
+          .join('accounts', 'accounts.id', 'test_table_two.account_id')
+          .where({ 'accounts.email': 'test3@example.com' })
+          .del();
+        if (isSQLite(knex) || isPostgreSQL(knex) || isOracle(knex)) {
```
kibertoad, line 94: Do you see downside of using this approach for PostgreSQL - https://www.postgresqltutorial.com/postgresql-delete-join/?

kibertoad:
> I'm having issues with successfully running 1 test: https://github.com/knex/knex/blob/master/test/integration/query/selects.js#L1361
The timeout error is not being caught properly leading to a general "Timeout of 5000ms exceeded." error.
It is not related to this update, but it's better to let know.

This is a known flaky test, feel free to ignore.

test/integration/query/deletes.js
Egor-Koldasov: @kibertoad  I did not think about it since it has a different syntax.
The best solution for PostgreSQL I think would be to add direct support for `USING` syntax like a `.using` method, it can use the same interface with the current `.join`.
Overall, internally transforming from `.join` might be convenient and helpful too. Among downsides, maybe a small confusion from query transformation.

kibertoad: Fair. Converting joins to non-joins is likely to be confusing.

kibertoad (approved):

kibertoad:
Thanks!

test/integration/query/deletes.js
elhigu: In this case it would seem to be quite trivial crossdb compatibility thing to implement... though, maybe native syntax should be allowed too. 

Also tests for error cases where people tries to create .join queries for unsupported databases seem to be missing.

test/unit/query/builder.js
```diff
@@ -10497,31 +9918,78 @@ describe('QueryBuilder', () => {
...
+        mssql: {
+          sql: 'delete [users] from [users] inner join [photos] on [photos].[id] = [users].[id] where [user].[email] = ?;select @@rowcount',
+          bindings: ['mock@test.com'],
+        },
+        oracledb: {
+          sql: 'delete "users" from "users" inner join "photos" on "photos"."id" = "users"."id" where "user"."email" = ?',
+          bindings: ['mock@test.com'],
+        },
+        pg: {
+          sql: 'delete "users" from "users" inner join "photos" on "photos"."id" = "users"."id" where "user"."email" = ?',
```
elhigu, line 9965: Is this valid SQL or just generates error from database? If feature is not supported by database, knex should probably throw an error and tell about it or otherwise we will get bug reports about this feature not working.

test/integration/query/deletes.js
kibertoad: @Egor-Koldasov Could you add these in a follow-up PR?

kibertoad: @OlivierCavadenti Delete join support with `USING` syntax for PG at this point is probably the largest gap in functionality still missing for 1.0.0. If you'll have some spare time, any chance you could take a look at it?

OlivierCavadenti: @kibertoad yeah I can work on it soon. As I understand, and to be sure, we go on a syntax from :

```
knex
	.select('xyz')
	.from('my_table')
	.leftJoin('users', 'users.user_id', 'my_table.user_id')
        .where('my_table.criteria', '<', 10)
	.delete();
```

to in PG (no hidden converting left join to using ?):

```
knex
	.select('xyz')
	.from('my_table')
	.using('users')
        .where('my_table.criteria', '<', 10)
        .andWhere('users.user_id', 'my_table.user_id')
	.delete(); 
```

kibertoad: Yes, exactly!

PR #4571, opened 2021-07-12
Initial infrastructure for CockroachDB

kibertoad:
Includes major integration tests refactoring to decouple them from each other and make them easier to debug in isolation, which is crucial for fixing failures on CockroachDB.


kibertoad:
I'm merging this prematurely since CI passes and there are sweeping changes all over the integration tests that will cause ridiculous amount of conflicts if not merged.

Issue #4452, opened 2021-04-24
Returning supported by lastest MariaDB, but not working in Knex


khalilst:
# Environment

Knex version: 0.95.4
Database + version: MariaDB 10.5.8
OS: Ubuntu 20.04

@smorey2

# Bug

1. MariaDB's latest version is supporting `returning`, but calling `.returning` using `knex` has no effect.

2. Error message:
```js
.returning() is not supported by MySQL and will not have any effect.
```

kibertoad:
@khalilst Would you consider submitting a PR for this?

khalilst:
@kibertoad Sure. I will create a PR today.

intech:
@khalilst could you do? Did not find PR.

khalilst:
@intech Yes. I will submit it.

mercmobily:
Closing this in favour of #6254 which closes the exact same issue

PR #4572, opened 2021-07-15
feat: add support for returning in mariadb

khalilst:
This PR removes the `.returning()` support warning in MariaDB in the following of issue #4452.

Documentation for `returning` statement:

- [MariaDB](https://mariadb.com/kb/en/insertreturning)


kibertoad:
@khalilst I don't think this is correct, linked documentation for MySQL says nothing about `returning` being supported and warning is still correct for it. We should make this warning conditional, based on client version.
Here is an example of such conditional logic: https://github.com/knex/knex/blob/c335fda65723e73c9de5d1a92fe7444f0cfb4071/lib/dialects/postgres/schema/pg-columncompiler.js#L125

We probably need to add version parsing logic that would determine if DB is Maria or MySQL, with possibility to do manual override

khalilst:
@kibertoad 
Thank you for the point.
Let me add the conditional warning. if you don't mind.

elhigu:
> @kibertoad We probably need to add version parsing logic that would determine if DB is Maria or MySQL, with possibility to do manual override..

Automatic parsing that was in psql dialect did cause lots of problems with postgresql when people were using different postgresql service propietary versions and different enterprise versions etc. and version numbers were not compatible.

We ended up allowing users to pass version number manually and that passed version tells to which official non enterprise psql version the server that is being connected is compatible with. So if server is MyCloudPsql 3.1 and service tells that it is compatible with psql 11.0, then version 11.0 is passed to knex.

So I believe that it is easier and more robust just to tell knex to which kind of mysql server we are connecting to.

elhigu:
Warning could also mention, that user needs to pass version to knex to enable that feature.

khalilst:
@kibertoad 
Sorry for the late answer.
The required changes are applied to the PR.
Please let me know if there is anything else.

lib/dialects/mysql/query/mysql-querycompiler.js
```diff
@@ -9,12 +9,13 @@ class QueryCompiler_MySQL extends QueryCompiler {
     super(client, builder, formatter);
 
     const { returning } = this.single;
-    if (returning) {
+    if (returning && (!client.version || parseFloat(client.version) < 10.5)) {
```
kibertoad, line 12: I would extract the version check in a separate function and write tests for it. We need to support both the `10.5` and the `10.6.3` format, and I don't think that current solution does.
In order to avoid reparsing on every query, I think we could check it once when creating a client instance, and then set a flag `isReturningSupported` that will be reused for checks.

kibertoad, line 12: since mysql does not support `returning`, only maria does, we need to check not just version. Probably we can introduce another field alongside `client` and `version` that would indicate specific db type, and include it in the check.

kibertoad:
This needs to be documented at https://github.com/knex/documentation

lib/dialects/mysql/query/mysql-querycompiler.js
khalilst: You are right, but at the moment it does NOT result in the wrong warning.
Because the latest version of `MySQL` is much smaller than `10.5.0`.

kibertoad: It's still hacky and confusing, and should MySQL ever go up to version 10, things will suddenly break. Let's handle it from the very beginning.

khalilst: All right. I suggest putting the `type` field as a new field alongside `client` and `version`.

kibertoad: >  I suggest putting the type field as a new field alongside client and version.

Sure, that works.

asdofindia:
Can I work on this? @kibertoad @khalilst ?

marvin-wtt:
Any update on this?

kibertoad:
if someone pushes this over the finish line, I'm happy to merge

jclaveau:
This seems to be a little change for a lot of value. Any update?

tgriesser:
This only changes the warning, which I cleaned up a bit since #6415 landed, however we still need to ensure the `returning` clause gets applied properly which is the bigger lift / more important since with this change, the warning will be gone but `returning` will do nothing.

Adding some commits for that in here

/cc @erulabs

lib/dialects/mariadb/query/mariadb-querycompiler.js
```diff
@@ -22,6 +22,77 @@ class QueryCompiler_MariaDB extends QueryCompiler_MySQL {
...
+    if (version < 10.5) {
+      return false;
+    }
+    if (method === 'update') {
+      return version >= 13 && (this.grouped.join || []).length === 0;
+    }
+    if (method === 'del') {
+      return (this.grouped.join || []).length === 0;
+    }
+    return version >= 10.5;
```
tgriesser:
```suggestion
    return true;
```

PR #4576, opened 2021-07-16
nguymin4/improve types

nguymin4:
- Improve type support for analytic functions (See tests for more detail)
- Extract common interfaces for tsd

Important note: I intended to migrate analytic-function-types testing to tsd but I found a problem with the tsd library
For example
```typescript
  expectType<Pick<User, 'id'>>(
    await knex<User>('users').select('foo')
  );
```
I expect this test fails but the test somehow passes. I think this is a tsd issue. Thus we need to be more careful before migrating from `dtslint` to `tsd`. I will try to debug this if I have time but in my opinion `dtslint` is still doing a good job for us

### DIFF
--- a/test-tsd/common.ts
+++ b/test-tsd/common.ts
@@ -6,3 +6,31 @@ export const clientConfig: Knex.Config = {
     filename: './mydb.sqlite',
   },
 };
+
+export interface User {
+  id: number;
+  age: number;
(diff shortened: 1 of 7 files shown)


kibertoad (approved):

kibertoad:
Released in 0.95.8

Issue #4577, opened 2021-07-16
Not passing useTz in datetime column creation results in "datetime without timezone


armandAkop:
# Environment

Knex version: ^0.21.17
Database + version: postgres 13
OS: MAC OS Catalina 10.15.7
Additional Info: Running this with docker-compose using the latest postgres image (v13), node 14-alpine

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do

I'm not sure if this is a bug with knex, or just a [documentation bug](https://knexjs.org/#Schema-datetime). I created a datetime column and did not provide a "useTz". According to the documentation above, it should default to true (and by my understanding, column data type should be `datetime with timezone`

> Adds a datetime column. By default PostgreSQL creates column with timezone (timestamptz type). This behaviour can be overriden by passing the useTz option (which is by default true for PostgreSQL).

I expected to get a column with type `datetime with timezone` but instead I received `datetime without timezone`.

2. Error message

N/A

3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real
   database connection to MySQL or PostgreSQL, then single file example which initializes
   needed data and demonstrates the problem.

Here is the migration file
```node
exports.up = async (knex) => {
  await knex.schema
    .createTable('my_table', function (table) {
      table.datetime('modified_date_useTz_not_provided', { precision: 6 }).defaultTo(knex.fn.now(6));
      table.datetime('modified_date_useTz_true', { precision: 6, useTz: true }).defaultTo(knex.fn.now(6));
      table.datetime('modified_date_useTz_false', { precision: 6, useTz: false }).defaultTo(knex.fn.now(6));
    });

};

exports.down = async (knex) => {
  await knex.schema.dropTable('my_table');
};

  
exports.config = { transaction: false };

```

When I query my table schema + type
```sql
SELECT column_name, data_type
FROM information_schema.columns
WHERE table_name = 'my_table';
```

I see this
![Screen Shot 2021-07-16 at 3 47 52 PM](https://user-images.githubusercontent.com/3230752/126015447-97486594-d203-478a-88d2-ccb8f09d660a.png)

elhigu:
Looks like a bug, PR welcome.

zeotuan:
Hi i want to work on this. Do i need to get assigned or i can just work and create a PR?

kibertoad:
@zeotuan Go ahead!

zeotuan:
[...]

PR #4578, opened 2021-07-18
Fixing posgres datetime and timestamp column created with wrong format

zeotuan:
- Why:
See issue #4577 

- How: 
1. Change default useTz option of datetime and timezone to true in pg column compiler. 
2. Adding additional test to check when empty option object or option object wihout timezone is passed.

### DIFF
--- a/test/unit/schema-builder/postgres.js
+++ b/test/unit/schema-builder/postgres.js
@@ -1155,11 +1155,37 @@ describe('PostgreSQL SchemaBuilder', function () {
     );
   });
 
+  it('adding default datetime', () => {
+    tableSql = client
+      .schemaBuilder()
+      .table('users', (table) => {
+        table.datetime('foo');
+      })
+      .toSQL();
+    equal(1, tableSql.length);
+    expect(tableSql[0].sql).to.equal(
+      'alter table "users" add column "foo" timestamptz'
+    );
+  });
+
   it('adding timestamp with timezone', () => {
     tableSql = client
       .schemaBuilder()
       .table('users', (table) => {
-        table.timestamp('foo', false);
+        table.timestamp('foo', true);
+      })
+      .toSQL();
+    equal(1, tableSql.length);
+    expect(tableSql[0].sql).to.equal(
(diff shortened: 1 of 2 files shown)


kibertoad:
> Doesn't this change default values and hence is a breakibg change?

kibertoad:
Isn't this still a breaking change, as it will invert values set by users already?

zeotuan:
@kibertoad sorry the naming of variables was bad. Yeah i think so. I can change it so that it is not a breaking change anymore.

kibertoad:
It is cleaner, yes, just a bit too late to change it, would require a semver major and still likely to break stuff for people who don't read docs.

zeotuan:
Alright i Changed it. It shouldn't invert value already set by user anymore.

test/unit/schema-builder/postgres.js
```diff
@@ -1168,6 +1181,19 @@ describe('PostgreSQL SchemaBuilder', function () {
     );
   });
 
+  it('adding datetime with timezone', () => {
+    tableSql = client
+      .schemaBuilder()
+      .table('users', (table) => {
+        table.datetime('foo', false);
```
armandAkop, line 1188: Maybe I'm missing some context but wouldn't passing `false` here result in timestamp WITHOUT timezone? Why does it seem inverted?

zeotuan: - It's Because this is the original design 
image

- And this is the original test for timestamp() which do the same thing as datetime():
image

I did changed it to withTz so it does not seem inverted anymore. But as @kibertoad said, it will cause breaking change so i  just add the bug fix and sometest without changing the parameter. Maybe we could request this as feature for future update because i agree that it is not intuiitive. 
Also it is only mentioned in the documentation that you can use ```optiton: {useTz: boolean, precision: number}``` so i don't know if many people create timezone like that

zeotuan:
@kibertoad I changed it so it fixed the bug without making breaking changes. Is there any other change that i need to do?

kibertoad (approved):

kibertoad:
Thank you!

Issue #4292, opened 2021-02-13
Please support "DEFERRABLE INITIALLY DEFERRED"


yodhcn:
# Feature request
Is there any chance to have a Knex native way to set "DEFERRABLE INITIALLY DEFERRED" on a table with foreign key constraints ?

```
knex.schema.table('users', function (table) {
  table.integer('user_id').unsigned()
  table.foreign('user_id').references('Items.user_id_in_items').deferred().
})
```

kibertoad:
@yodhcn PR for this would be most welcome, and if you need any help with finding the right spot in source or writing tests, would be more than happy to help.

Alucard17:
hey @kibertoad  can I take this issue?

Alucard17:
hey @yodhcn can you provide the compiled SQL query of above mentioned example.

kibertoad:
Released in 0.95.8

PR #4584, opened 2021-07-23
Add deferrable support for constraint

zeotuan:
Address #4292 : Add Deferrable Support
- Foreign: ```deferrable([deferrable])``` can be chained off ```table.foreign()``` or ```column.references()```
- allow option object to be used for primary/unique:
```column.unique(indexName | options={[indexName: string], [deferrable: string]});```  
```table.unique(columns,indexName | options={[indexName: string], [deferrable: string]});``` 
```column.primary(constraintName | options={[constraintName : string], [deferrable: string]});```  
```table.primary(columns,constraintName | options={[constraintName : string], [deferrable: string]});```

### CODE REVIEW

**Comment by zeotuan** [comment]
@kibertoad this test seem to be flaky:
```Integration Tests
       postgresql | pg
         knex.migrate.latest in parallel
           is not able to run two migrations in parallel when transactions are disabled:

      AssertionError: expected false to equal true
      + expected - actual

      -false
      +true
      
      at /home/runner/work/knex/knex/test/integration/migrate/migration-integration-tests.js:789:33
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)```

File: lib/dialects/oracle/schema/oracle-tablecompiler.js [hunk 1]
```diff
@@ -86,14 +87,20 @@ class TableCompiler_Oracle extends TableCompiler {
     );
   }
 
-  primary(columns, constraintName) {
+  primary(columns, constraintName, deferrable) {
+    if (isObject(columns) && !(columns instanceof Array)) {
```
kibertoad, line 91: I think idiomatic JS way is Array.isArray()

lib/dialects/oracle/schema/oracle-tablecompiler.js
```diff
@@ -86,14 +87,20 @@ class TableCompiler_Oracle extends TableCompiler {
     );
   }
 
-  primary(columns, constraintName) {
+  primary(columns, constraintName, deferrable) {
```
kibertoad, line 90: Generally we try to avoid adding many positional parameters, as these become very confusing very fast.
Could you add options object as a third parameter and put `deferrable` there?

lib/dialects/oracle/schema/oracle-tablecompiler.js
```diff
@@ -125,15 +132,22 @@ class TableCompiler_Oracle extends TableCompiler {
     this.pushQuery(`drop index ${indexName}`);
   }
 
-  unique(columns, indexName) {
+  unique(columns, indexName, deferrable) {
```
kibertoad, line 135: same comments from `primary` apply here

test/integration2/transaction/set-deferrable-constraint.spec.js
```diff
@@ -0,0 +1,120 @@
...
+          });
+        });
+
+        afterEach(async () => {
+          await knex.schema.dropTable(tableName);
+          await knex.schema.dropTable(tableName1);
+        });
+
+        it('deferrable initially immediate unique constraint all row are checked at end of update', async () => {
+          if (isPostgreSQL(knex) || isOracle(knex) || isOracle(knex)) {
```
kibertoad, line 42: Why double isOracle check?

kibertoad, line 42: generally it's better to explicitly skip unsupported cases rather than execute empty test for them, that results in cleaner test suite output

lib/dialects/oracle/schema/oracle-tablecompiler.js
zeotuan: Are you suggesting having it like this? ```primary(columns,consntraintName,options={[deferrable:string]}```
Or should it be ```primary(columns,constraintName|options={[constrantName:string],[deferrable:string]})```
This way we can still make sure that code using older version still work and allow user to provide ```deferrable``` without having to specify ```constraintName```
Then maybe in the future we can switch to have only options```primary(columns,options={[constrantName:string],[deferrable:string]})```

kibertoad: Yes, it is important to maintain backwards compatibility with the existing code. Second approach seems cleaner, since it doesn't make sense to require constraintName to be specified in order to be able to use options

test/integration2/transaction/set-deferrable-constraint.spec.js
zeotuan: My bad gonna just skip all test like you said

zeotuan: Got it

kibertoad (approved):

kibertoad:
Great job! Could you also create doc PR at https://github.com/knex/documentation?

zeotuan:
> Great job! Could you also create doc PR at https://github.com/knex/documentation?

Documentation added. Can you check it out.

PR #4589, opened 2021-07-27
Improve Typing: add typescript support for deferrable, new Primary/Unique syntax

### LINKED ISSUE #3944: Inconsistent camelCase use for datetime in schema
# Environment

Knex version: 0.21.2

# Bug

The docs describe creating a `datetime` column by using `table.datetime()`: http://knexjs.org/#Schema-datetime

However, the TypeScript definition and tests seem to indicate it should be `table.dateTime()`:

https://github.com/knex/knex/blob/b59161f9944acf858b149a6c2901b14fcd4ba8aa/types/index.d.ts#L1560

https://github.com/knex/knex/blob/b59161f9944acf858b149a6c2901b14fcd4ba8aa/test/integration/builder/selects.js#L743

as a matter of fact, I was only able to find a single reference that seems to use `datetime` in schema building:

https://github.com/knex/knex/blob/b59161f9944acf858b149a6c2901b14fcd4ba8aa/test/unit/schema/mysql.js#L832

Are the docs wrong, or is there something bigger here I'm missing? (Also, how does that last test pass when using `datetime()` instead of `dateTime()`? Are both valid? @lorefnon

zeotuan:
also addressing missing datetime missing in 
close #3944

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2405,7 +2405,7 @@ interface MsSqlConnectionConfigBase {
   }
 
   interface SeederConfig<V extends {} = any> {
-    extension?: string;cd
+    extension?: string;
     directory?: string | readonly string[];
     loadExtensions?: readonly string[];
     specific?: string;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1910,9 +1910,8 @@ export declare namespace Knex {
     comment(val: string): TableBuilder;
     specificType(columnName: string, type: string): ColumnBuilder;
     primary(columnNames: readonly string[], options?: Readonly<{constraintName?: string, deferrable?: deferrableType}>): TableBuilder;
-    /**@deprecated*/
+    /** @deprecated */
     primary(columnNames: readonly string[], constraintName?: string): TableBuilder;
-    
     index(
       columnNames: string | readonly (string | Raw)[],
       indexName?: string,
@@ -1965,7 +1964,7 @@ export declare namespace Knex {
     withKeyName(keyName: string): ColumnBuilder;
     after(columnName: string): ColumnBuilder;
     first(): ColumnBuilder;
-    deferrable(type: deferrableType);
+    deferrable(type: deferrableType): ColumnBuilder;
   }
 
   interface ForeignConstraintBuilder {
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1884,6 +1884,7 @@ export declare namespace Knex {
     boolean(columnName: string): ColumnBuilder;
     date(columnName: string): ColumnBuilder;
     dateTime(columnName: string, options?: Readonly<{useTz?: boolean, precision?: number}>): ColumnBuilder;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -1908,12 +1909,16 @@ export declare namespace Knex {
     uuid(columnName: string): ColumnBuilder;
     comment(val: string): TableBuilder;
     specificType(columnName: string, type: string): ColumnBuilder;
+    primary(columnNames: readonly string[], options?: Readonly<{constraintName?: string, deferrable?: deferrableType}>): TableBuilder;
```
zeotuan, line 1912: New type for primary using options object

types/index.d.ts
```diff
@@ -1908,12 +1909,16 @@ export declare namespace Knex {
...
+    /** @deprecated */
     primary(columnNames: readonly string[], constraintName?: string): TableBuilder;
     index(
       columnNames: string | readonly (string | Raw)[],
       indexName?: string,
       indexType?: string
     ): TableBuilder;
+    unique(columnNames: readonly (string | Raw)[], options?: Readonly<{indexName?: string, deferrable?: deferrableType}>): TableBuilder;
```
zeotuan, line 1920: Added type for unique using options object

types/index.d.ts
```diff
@@ -1908,12 +1909,16 @@ export declare namespace Knex {
     uuid(columnName: string): ColumnBuilder;
     comment(val: string): TableBuilder;
     specificType(columnName: string, type: string): ColumnBuilder;
+    primary(columnNames: readonly string[], options?: Readonly<{constraintName?: string, deferrable?: deferrableType}>): TableBuilder;
+    /** @deprecated */
```
zeotuan, line 1913: add deprecate tag for old syntax

zeotuan:
@kibertoad Hi i added type support for the the new deferrable support that was added. Also added a datetime type

types/index.d.ts
```diff
@@ -1884,6 +1884,7 @@ export declare namespace Knex {
     boolean(columnName: string): ColumnBuilder;
     date(columnName: string): ColumnBuilder;
     dateTime(columnName: string, options?: Readonly<{useTz?: boolean, precision?: number}>): ColumnBuilder;
+    datetime(columnName: string, options?: Readonly<{useTz?: boolean, precision?: number}>): ColumnBuilder;
```
kibertoad, line 1887: what is it for? isn't this overload fully equivalent to the previous one?

zeotuan: I think that the problem was mentioned in #3944 where the same method is aliased as ```dateTime```and ```datetime``` but only the definition for ```dateTime``` is available in TS defs while only the ```datetime``` is mentioned in documentation so i added ```datetime``` to the TS defs.

kibertoad: ah, gotcha, I missed the difference in casing. thank you

kibertoad (approved):

PR #4594, opened 2021-07-29
Fix autoincrement for Oracle with schema

### LINKED ISSUE #4592: table.increments() errors on Oracle when .withSchema()
# Environment

**Knex version:** `0.95.6`
**Database + version:** Oracle 18c (@atiertant)
**OS:** Ubuntu Linux

# Bug

1. **Explain what kind of behavior you are getting and how you think it should do**

When using Oracle and attempting to create an incrementing column in a table with a non-default schema Knex crashes with an internal error.

2. **Error message**

> Indents added by issue filer for easier reading

```sql
Error:
DECLARE PK_NAME VARCHAR(200);
BEGIN  EXECUTE IMMEDIATE ('CREATE SEQUENCE "my_table_seq"');
SELECT cols.column_name INTO PK_NAME
  FROM all_constraints cons, all_cons_columns cols
  WHERE cons.constraint_type = 'P'
    AND cons.constraint_name = cols.constraint_name
    AND cons.owner = cols.owner
    AND cols.table_name = 'my_table';
execute immediate
  ('create or replace trigger "my_table_autoinc_trg"
    BEFORE INSERT on "my_table" for each row
    declare checking number := 1;
    begin
      if (:new."' || PK_NAME || '" is null) then
        while checking >= 1 loop
          select "my_table_seq".nextval into :new."' || PK_NAME || '" from dual;
          select count("' || PK_NAME || '") into checking from "my_table"
          where "' || PK_NAME || '" = :new."' || PK_NAME || '";
        end loop;
      end if;
    end;');
END;
- ORA-00942: table or view does not exist
ORA-06512: at line 1
```

3. **Reduced test code, for example in https://npm.runkit.com/knex or if it needs real
   needed data and demonstrates the problem.**

Minimal migration file:

```ts
import { Knex } from 'knex'

const schemaName = 'my_schema'
const tableName = 'my_table'

exports.up = async function (knex: Knex): Promise<void> {
  // Create schema
  const sanitizedRaw = knex.raw('CREATE USER ?? IDENTIFIED BY ??', [schemaName, schemaName]).toQuery()
  await knex.schema.raw(sanitizedRaw)
  // Create table
  await knex.schema.withSchema(schemaName).createTable(tableName, (table) => {
    table.increments('my_column')
  })
}

exports.down = async function (knex: Knex): Promise<void> {
  // left empty for simplest example
}
```

4. **Additional thoughts.**

    1. A quick search through the codebase seems to indicate that the following function is the one which generates the failed code: https://github.com/knex/knex/blob/1744c8c2655a8362260ba987d706ec3473681a78/lib/dialects/oracle/schema/internal/trigger.js#L46
    2. Oracle has had support for `IDENTITY` columns since 12c, if that's a cleaner way to solve this. Though that doesn't help for backwards compatibility pre 12c. Reference: https://chartio.com/resources/tutorials/how-to-define-an-auto-increment-primary-key-in-oracle/#identity-columns

kibertoad:
fixes #4592


kibertoad:
@codeape I don't have an Oracle instance at hand to test, could you check if this thing actually works? If not, would appreciate suggestions on how expected SQL output should actually look.

code-ape:
@kibertoad will check immediately and report back! Thank you.

code-ape:
Found that we needed to match `cons.owner` on the given schema and also create the trigger and sequence in the given sequence. Code may need a bit of cleaning, but appears to work!

<details>
<summary>View code</summary>

```ts
createAutoIncrementTrigger: function (logger, tableName, schemaName) {
    const tableQuoted = `"${tableName}"`;
    const tableUnquoted = tableName;
    const schemaQuoted = schemaName ? `"${schemaName}".` : '';
    const constraintOwner = schemaName ? `'${schemaName}'` : 'cols.owner';
    const triggerName = utils.generateCombinedName(
      logger,
      'autoinc_trg',
      tableName
    );
    const sequenceNameUnQuoted = utils.generateCombinedName(logger, 'seq', tableName);
    const sequenceNameQuoted = `"${sequenceNameUnQuoted}"`
    return (
      `DECLARE ` +
      `PK_NAME VARCHAR(200); ` +
      `BEGIN` +
      `  EXECUTE IMMEDIATE ('CREATE SEQUENCE ${schemaQuoted}${sequenceNameQuoted}');` +
      `  SELECT cols.column_name INTO PK_NAME` +
      `  FROM all_constraints cons, all_cons_columns cols` +
      `  WHERE cons.constraint_type = 'P'` +
      `  AND cons.constraint_name = cols.constraint_name` +
      `  AND cons.owner = ${constraintOwner}` +
      `  AND cols.table_name = '${tableUnquoted}';` +
      `  execute immediate ('create or replace trigger ${schemaQuoted}"${triggerName}"` +
      `  BEFORE INSERT on ${schemaQuoted}${tableQuoted}` +
      `  for each row` +
      `  declare` +
      `  checking number := 1;` +
      `  begin` +
      `    if (:new."' || PK_NAME || '" is null) then` +
      `      while checking >= 1 loop` +
      `        select ${schemaQuoted}${sequenceNameQuoted}.nextval into :new."' || PK_NAME || '" from dual;` +
      `        select count("' || PK_NAME || '") into checking from ${schemaQuoted}${tableQuoted}` +
      `        where "' || PK_NAME || '" = :new."' || PK_NAME || '";` +
      `      end loop;` +
      `    end if;` +
      `  end;'); ` +
      `END;`
    );
  }
```
</details>

PR #4599, opened 2021-08-01
Improve connection string handling for SQLite3 and Postgresql.

rustyconover:
Allow the use null as default option for sqlite3 to be specified in a connection string
by including it in the query parameters.  An example is:

`sqlite3:///:memory:?useNullAsDefault=true`

Additionally, improve PostgreSQL connection string handling by allowing
the searchPath to be specified in the connection string.  An example is:

`postgresql:///?searchPath=pg_temp`

Without these changes, it is impossible to allow users to specify these options
using connection strings.

### DIFF
--- a/lib/knex-builder/internal/parse-connection.js
+++ b/lib/knex-builder/internal/parse-connection.js
@@ -32,24 +32,45 @@ module.exports = function parseConnectionString(str) {
 
   const isPG = ['postgresql', 'postgres'].includes(protocol);
 
-  return {
+  const result = {
     client: protocol,
-    connection: isPG ? parsePG(str) : connectionObject(parsed),
+    connection: isPG ? parsePG(str) : connectionObject(protocol, parsed),
   };
+  if (
+    result.client === 'sqlite3' &&
+    result.connection != null &&
+    result.connection.useNullAsDefault
+  ) {
+    delete result.connection['useNullAsDefault'];
+    result['useNullAsDefault'] = true;
+  }
+  if (
+    result.client === 'postgresql' ||
+    (result.client === 'postgres' && result.connection.searchPath != null)
+  ) {
+    const path = result.connection.searchPath;
+    delete result.connection.searchPath;
+    result.searchPath = path;
+  }
+  return result;
 };
 
 /**
  * @param {URL} parsed
  * @returns {{}}
  */
-function connectionObject(parsed) {
+function connectionObject(protocol, parsed) {
   const connection = {};
   let db = parsed.pathname;
   if (db[0] === '/') {
     db = db.slice(1);
   }
 
-  connection.database = db;
+  if (protocol === 'sqlite3') {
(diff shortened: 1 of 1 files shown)


lib/knex-builder/internal/parse-connection.js
```diff
@@ -32,24 +32,45 @@ module.exports = function parseConnectionString(str) {
...
+  if (
+    result.client === 'sqlite3' &&
+    result.connection != null &&
+    result.connection.useNullAsDefault
+  ) {
+    delete result.connection['useNullAsDefault'];
+    result['useNullAsDefault'] = true;
+  }
+  if (
+    result.client === 'postgresql' ||
```
kibertoad, line 48: is this condition correct? logical operators and parentheses seem off

kibertoad:
Could you please add tests for new logic?

lib/knex-builder/internal/config-resolver.js
```diff
@@ -42,9 +40,11 @@ function resolveConfig(config) {
 
   // If config connection parameter is passed as string, try to parse it
   if (typeof parsedConfig.connection === 'string') {
-    resolvedConfig = Object.assign({}, parsedConfig, {
-      connection: parseConnection(parsedConfig.connection).connection,
-    });
+    resolvedConfig = Object.assign(
+      {},
+      parsedConfig,
+      parseConnection(parsedConfig.connection)
```
maximelkin, line 46: Maybe
`parseConnection(parsedConfig.connection).connection`
?

maximelkin, line 46: `parseConnection().client` goes into `Dialect` variable in previous code fragments

lib/knex-builder/internal/parse-connection.js
```diff
@@ -32,24 +32,45 @@ module.exports = function parseConnectionString(str) {
 
   const isPG = ['postgresql', 'postgres'].includes(protocol);
 
-  return {
+  const result = {
     client: protocol,
-    connection: isPG ? parsePG(str) : connectionObject(parsed),
+    connection: isPG ? parsePG(str) : connectionObject(protocol, parsed),
   };
+  if (
```
maximelkin, line 39: This is mixing of parameters for db and parameters for knex, its make harder to understand where options actually will be applied.  
Also `useNullAsDefault` is option to configurate via code, not via connection string config, because code could rely on it. (I'm implying that connection string goes from external config, which is common case. In case of rest options such as `useNullAsDefault`, they should be set from code, because depends on it)

maximelkin:
Btw where is kinda strange syntax for postgres search_path:
```
postgresql://user@localhost:5433/mydb?options=-c%20search_path%3Dyour_schema
```
basically `{options: 'search_path=your_schema'}`, urlencoded

kibertoad:
@rustyconover why delete it? if comments are addressed, this can be merged

PR #4600, opened 2021-08-01
fix typing deferrable and withkeyName should not be in ColumnBuilder

zeotuan:
I came across this PR: https://github.com/knex/knex/pull/3556. since it hasn't been actived for a long time so i want to conttinue it and also to fix my last Pull Request where i also put ```deferrable``` inside ```ColumnBuilder```

### DIFF
--- a/test-tsd/tables.test-d.ts
+++ b/test-tsd/tables.test-d.ts
@@ -1,6 +1,6 @@
 import { knex, Knex } from '../types';
 import { clientConfig, User, Department, Article } from './common';
-import { expectType } from 'tsd';
+import { expectType, expectAssignable } from 'tsd';
 
 const knexInstance = knex(clientConfig);
 
@@ -65,4 +65,17 @@ const main = async () => {
   expectType<Pick<User, 'id'> | undefined>(
     await knexInstance.first('id').from('users_composite')
   );
+
+  //These tests simply check if type work by showing that it does not throw syntax error
+
+  knexInstance.schema.createTable('testTable',(table) => {
+    table.foreign('fkey_three').references('non_exist.id').withKeyName('non_for1').deferrable('deferred');
+    table.foreign('fkey_threee').references('non_exist.id').deferrable('deferred').withKeyName('non_for2');
+    table.integer('num').references('non_exist.id').deferrable('immediate').withKeyName('non_for3');
+    table.integer('num').references('non_exist.id').withKeyName('non_for4').deferrable('deferred').onDelete('CASCADE');
+    table.integer('num').references('non_exist.id').withKeyName('non_for5').deferrable('deferred').onDelete('CASCADE');
+    table.integer('num').references('id').inTable('non_exist').withKeyName('non_for6').deferrable('deferred').onDelete('CASCADE');
+    table.integer('num').references('id').withKeyName('non_for7').deferrable('deferred').inTable('non_exist').onDelete('CASCADE');
+  })
 };
+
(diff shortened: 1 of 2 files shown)


kibertoad:
Can you also add a test for it?

types/index.d.ts
```diff
@@ -1981,6 +1979,8 @@ export declare namespace Knex {
 
   interface ReferencingColumnBuilder extends ColumnBuilder {
     inTable(tableName: string): ColumnBuilder;
+    deferrable(type: deferrableType): ColumnBuilder;
+    withKeyName(keyName: string): ColumnBuilder;
```
zeotuan, line 1983: Put This two in ```ReferencingColumnBuilder``` instead of just ```ForeignConstraintBuilder & MultikeyForeignConstraintBuilder``` since they need to be both accessible through ```references``` and ```foreign``` method

zeotuan:
> Can you also add a test for it?

Yeah sure

zeotuan:
@kibertoad I need help with writing test though. how should i write tsd test for this?

kibertoad:
@zeotuan I would simply invoke a builder chain in a test that used to throw a type error in the past. Since from my understanding this particular PR adds support for this method for more different builder cases, using several different chains would be the best.
I don't think there will be any actual assertions in new test code, just some code that wouldn't have compiled in the past.

types/index.d.ts
```diff
@@ -1979,8 +1976,10 @@ export declare namespace Knex {
     index(indexName?: string, indexType?: string): ColumnBuilder;
   }
 
-  interface ReferencingColumnBuilder extends ColumnBuilder {
-    inTable(tableName: string): ColumnBuilder;
+  interface ReferencingColumnBuilder extends ColumnBuilder{
+    inTable(tableName: string): ReferencingColumnBuilder;
+    deferrable(type: deferrableType): ReferencingColumnBuilder;
+    withKeyName(keyName: string): ReferencingColumnBuilder;
```
zeotuan, line 1982: extended this ```interface``` and also changed the type of ```inTable``` to ```ReferencingColumnBuilder``` so that all these function can be chained in any order

test-tsd/tables.test-d.ts
```diff
@@ -65,4 +65,17 @@ const main = async () => {
...
+
+  knexInstance.schema.createTable('testTable',(table) => {
+    table.foreign('fkey_three').references('non_exist.id').withKeyName('non_for1').deferrable('deferred');
+    table.foreign('fkey_threee').references('non_exist.id').deferrable('deferred').withKeyName('non_for2');
+    table.integer('num').references('non_exist.id').deferrable('immediate').withKeyName('non_for3');
+    table.integer('num').references('non_exist.id').withKeyName('non_for4').deferrable('deferred').onDelete('CASCADE');
+    table.integer('num').references('non_exist.id').withKeyName('non_for5').deferrable('deferred').onDelete('CASCADE');
+    table.integer('num').references('id').inTable('non_exist').withKeyName('non_for6').deferrable('deferred').onDelete('CASCADE');
+    table.integer('num').references('id').withKeyName('non_for7').deferrable('deferred').inTable('non_exist').onDelete('CASCADE');
+  })
```
zeotuan, line 79: @kibertoad test added. Should this test be here or should we add a different test file for this type of test?

types/index.d.ts
zeotuan: @kibertoad should i also add onUpdate and onDelete to this ```interface``` ? since right now ```onUpdate``` and ```onDelete``` always need to be at the end of the ```foreign/references``` chain

kibertoad: if it wouldn't be too hard!

test-tsd/tables.test-d.ts
kibertoad: here is good!

zeotuan:
@kibertoad This should be in a seperate issue but i think we only support ```onDelete``` and ```onUpdate``` with ```foreign``` and ```references``` but the documentation and our type suggest that we support ```onUpdate``` and ```onDelete``` on ```ColumnBuilder```.

kibertoad (approved):

kibertoad:
Thank you! Could you create an issue for the remaining problem?

PR #4602, opened 2021-08-02
Fix Analytic orderBy and partitionBy to follow the SQL documentation

wolfcomp:
This is a little annoyance I found while writing query calls using `rowNumber` as the typescript types shows you are allowed to do this but if you try to use an array of objects the SQL request fail with `column ?? does not exist` where ?? is the array of objects columnized.

### DIFF
--- a/lib/query/analytic.js
+++ b/lib/query/analytic.js
@@ -1,5 +1,4 @@
 const assert = require('assert');
-const { isObject } = require('../util/is');
 
 // Analytic
 // -------
--- a/lib/query/analytic.js
+++ b/lib/query/analytic.js
@@ -19,7 +19,7 @@ class Analytic {
     this.grouping = 'columns';
   }
 
-  partitionBy(column) {
+  partitionBy(column, direction) {
     assert(
       Array.isArray(column) || typeof column === 'string',
       `The argument to an analytic partitionBy function must be either a string
@@ -29,7 +29,7 @@ class Analytic {
     if (Array.isArray(column)) {
       this.partitions = this.partitions.concat(column);
     } else {
-      this.partitions.push(column);
+      this.partitions.push({ column: column, order: direction });
     }
     return this;
   }
--- a/lib/query/analytic.js
+++ b/lib/query/analytic.js
@@ -1,4 +1,5 @@
 const assert = require('assert');
+const { isObject } = require('../util/is');
 
 // Analytic
 // -------
@@ -33,7 +34,7 @@ class Analytic {
     return this;
   }
 
-  orderBy(column) {
+  orderBy(column, direction) {
     assert(
       Array.isArray(column) || typeof column === 'string',
       `The argument to an analytic orderBy function must be either a string
@@ -41,11 +42,15 @@ class Analytic {
     );
 
     if (Array.isArray(column)) {
-      this.order = this.order.concat(column);
+      this.order = this.order.concat(this._orderByArray(column));
     } else {
-      this.order.push(column);
+      this.order.push(column + (direction ? " " + direction : ""));
     }
     return this;
+  }
+  
+  _orderByArray(columnDefs){
+    return columnDefs.map(c => isObject(c) ? c.column + (c.order ? " " + c.order : "") : c).join(", ");
   }
 }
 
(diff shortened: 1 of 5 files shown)


kibertoad:
Thank you for your contribution! Could you please add integration test for this?

kibertoad:
@wolfcomp Can I help you with writing a test?

wolfcomp:
Currently writing the tests. Im also adding it into the `partition by` statement at the same time, as they use the exact same operations in the SQL documentation.

wolfcomp:
@kibertoad if you could start the workflows believe it should work as i cant see it in my terminal window due to me not being able to setup sql server connections on the network im on currently

wolfcomp:
Im unsure to why its saying that the formatter is undefined for columnize as the ones that fails seems to have the formatter set.

kibertoad:
formater -> formatter

wolfcomp:
🤦

wolfcomp:
Now it should work surprised that I haven't had any errors my self.

wolfcomp:
Actually there seems to still be an error as in the project that I use this I did function and not object assign on the parameter.

wolfcomp:
@kibertoad now it should all be fixed from what I can see in my terminal window

lib/query/analytic.js
```diff
@@ -18,7 +19,7 @@ class Analytic {
     this.grouping = 'columns';
   }
 
-  partitionBy(column) {
+  partitionBy(column, direction) {
```
kibertoad, line 22: This needs to be documented. Could you create a PR for https://github.com/knex/documentation?

lib/query/querycompiler.js
```diff
@@ -1035,13 +1035,23 @@ class QueryCompiler {
         sql += 'partition by ';
         sql +=
           map(stmt.partitions, function (partition) {
-            return self.formatter.columnize(partition);
+            if (isString(partition)) return self.formatter.columnize(partition);
```
kibertoad, line 1038: Please add missing {}

lib/query/querycompiler.js
```diff
@@ -1035,13 +1035,23 @@ class QueryCompiler {
...
+                self.formatter.columnize(partition.column) +
+                (partition.order ? ' ' + partition.order : '')
+              );
           }).join(', ') + ' ';
       }
 
       sql += 'order by ';
       sql += map(stmt.order, function (order) {
-        return self.formatter.columnize(order);
+        if (isString(order)) return self.formatter.columnize(order);
```
kibertoad, line 1049: Please add missing {}

test/unit/query/builder.js
```diff
@@ -7854,6 +7854,29 @@ describe('QueryBuilder', () => {
     );
   });
 
+  it('row_number with object', function () {
```
kibertoad, line 7857: can you also add an integration test?

wolfcomp: I can try to look into it, haven't really written integration tests before, probably will just copy one of the other ones and get it to fit the function

wolfcomp: @kibertoad after looking through the `test/integration` folder, I can't find anything previously for `rowNumber`, `denseRank`, `rank` in any of the files, so I'm unsure of how I should do it in that case.

lib/query/analytic.js
wolfcomp: https://github.com/knex/documentation/pull/333

test/unit/query/builder.js
kibertoad: You are right, we only have unit tests for analytic functions now

kibertoad:
@wolfcomp Are TS types correct now? Could you add a tsd test for this case?

wolfcomp:
I can check if the ts files are correct if nor correct them fast

wolfcomp:
There seems to be no tsd tests for analytic functions at all from what I can find.

wolfcomp:
@kibertoad this is what the best I can do from what I can find in the files when it comes to already existing code for testing and documenting [knex/documentation#333](https://github.com/knex/documentation/pull/333) the added and corrected functionality

kibertoad (approved):

kibertoad:
Released in 0.95.11

PR #4609, opened 2021-08-08
fix pluck typing issue when CompositeTableType is used

aq1018:
When a `CompositeTableType` is used, `keyof TRecord` will result in `base`, `insert`, and `update`, instead of the real database column names. This change fixes this issue.

Table definitions:

```typescript
import { Knex } from 'knex'

declare module 'knex/types/tables' {
  interface User {
    id: string
    name: string
    createdAt: Date
    updatedAt: Date
  }

  interface UserInsert {
    id?: string
    name: string
    createdAt?: Date
    updatedAt?: Date
  }

  interface UserUpdate {
    id?: string
    name?: string
    createdAt?: Date
    updatedAt?: Date
  }

  interface Tables {
    users:  Knex.CompositeTableType<User, UserInsert, UserUpdate>
  }
}
```

When using:

```typescript
import knex as 'knex'
import knexConfig from './knexfile'

const db = knex(knexConfig)

/* 
  *  Currently shows type as `{}`, 
  *  and autocomplete suggests: `base`, `insert`, `update`.
  *  This is because those are the keys of `CompositeTableType`.
  * 
  *  With the fix, it shows the type as `string[]`,
  *   and autocomplete suggests `id`, `name`, `createdAt`, `updatedAt`.
  *  This is because `ResolveTableType` can pull out `User` interface.
  */
db('users').pluck('id') 

```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -640,9 +640,9 @@ export declare namespace Knex {
     // Others
     first: Select<TRecord, DeferredKeySelection.AddUnionMember<UnwrapArrayMember<TResult>, undefined>>;
 
-    pluck<K extends keyof TRecord>(
+    pluck<K extends keyof ResolveTableType<TRecord>>(
       column: K
-    ): QueryBuilder<TRecord, TRecord[K][]>;
+    ): QueryBuilder<TRecord, ResolveTableType<TRecord>[K][]>;
     pluck<TResult2 extends {}>(column: string): QueryBuilder<TRecord, TResult2>;
 
     insert(


kibertoad:
could you please add tsd test for this?

aq1018:
@kibertoad I'll try. I've never done this before. Some hints would be nice.

aq1018:
@kibertoad after some research I think you are referring to adding tests under `./test-tsd` directory? I have some issues running the tests though. Maybe you can help. Here is what I did:

```
npm install
npx tsd
```

The process runs a bit, with no output and just stops.  I then tried to modify a few tests and see if it fails, and ran the `npx tsd` command again, and same result, no output or any indications.

Am I doing something wrong?

kibertoad:
@aq1018 I usually run it using `npm run lint:types`, but `npx tsd` should work fine as well. Unless it actually fails with an error, I think there is no output from it.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "knex",
-  "version": "1.0.0",
+  "version": "0.95.14",
```
kibertoad, line 3: why did you change the version? we can backport to 0.95 branch later, but master is for 1.0.0

Blargel: I merged to the 0.95.14 tag to update this fork because our application uses this fork for now. However, that tag has 1.0.0 in the package.json file for some reason. That version inconsistency was causing some weird sub-dependency issues that was causing type errors so I did this to fix it.

Blargel: In retrospect, perhaps our application should be looking at a different branch so as to not interfere with this PR. Apologies for the mess. I had forgotten this PR existed while I was trying to get our app working.

kibertoad:
You are targetting master.

package.json
kibertoad: ah, I see. No worries :)

Blargel:
I set the state of this branch back to before I messed with it a few hours ago. I'll try to figure out how to write tests so this later on this weekend I guess.

Blargel:
When I added the tests, I tried removing the changes that @aq1018 made while running them and expected to see a failure. However, when I tried that, they actually still passed for some reason, even though this exact use case failed in our own application. I spent some time investigating that, but I couldn't figure out what was different between our application and the test. At the very least, this change doesn't break existing typing, though.

myndzi:
The problem here was that the resolved type with the incorrect typing was resolving to `any`, which passed `tsd`'s `expectType` assertion. It behaved this way even on the latest version. I chose to add `expect-type` with the intention to migrate to that library.

PR #4623, opened 2021-08-17
fix: use sys info function instead of connection db name

sifer:
There is an edge case when getting proxied by for example PgBouncer that could result in client connection not being the same as the current database.
Using current_database() seems like a safer approach.

### DIFF
--- a/lib/dialects/postgres/query/pg-querycompiler.js
+++ b/lib/dialects/postgres/query/pg-querycompiler.js
@@ -186,8 +186,8 @@ class QueryCompiler_PG extends QueryCompiler {
     }
 
     let sql =
-      'select * from information_schema.columns where table_name = ? and table_catalog = ?';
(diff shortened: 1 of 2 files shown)


kibertoad:
can you add test for this?

kibertoad:
@rijkvanzanten Does this make sense for you?

rijkvanzanten:
Yes it does! 

Like @sifer mentioned, it is a bit of an edge case, but you can in fact tell PgBouncer to expose a different database name than the one used in the actual database, effectively creating an alias for the database. By relying on the connection config in knex, it'll use the name of the alias in the SQL query instead of the name of the actual database, which will fail.

It does make me wonder: does the same problem occur with other databases and load balancer combinations?

kibertoad (approved):

kibertoad:
Released in 0.95.11

PR #4629, opened 2021-08-20
Add support for mysql/mariadb-client JSON parameters in connectionURIs

### LINKED ISSUE #4628: Knex doesn't parse JSON values in connectionURIs causing incompatibility with mysql / node-mysql2
# Environment

Knex version: 0.95.9, 0.21.21
Database + version: MariaDB 10.3, `mysql` package 2.18.1 but also affects node-mysql2
OS: MacOS Monterey

# Bug

## 1. Explain what kind of behaviour you are getting and how you think it should do

I am trying to pass in a connectionURI that contains JSON in the `searchParams`.

```
mariadb://foo:bar@foobar.com/foo?ssl={\"ca\": \"--- BEGIN CERTIFICATE --- ...\"}
```

This might seem funky at first, but it is actually supported by both `mysql` and `node-mysql2`

Here's the `node-mysql2` implementation:

https://github.com/sidorares/node-mysql2/blob/d74558b605162156b813248a024f7559785de6fb/lib/connection_config.js#L234-L254

and here's the `mysql` one:

https://github.com/mysqljs/mysql/blob/3430c513d6b0ca279fb7c79b210c9301e9315658/lib/ConnectionConfig.js#L179-L209

As you can see, the parser first tries to parse the search/query parameters with `JSON.parse` and then fallbacks on just adding the value as string.

`knex`s parser, however, just assigns them as strings:

https://github.com/knex/knex/blob/1744c8c2655a8362260ba987d706ec3473681a78/lib/knex-builder/internal/parse-connection.js#L70-L73

## 2. Error message

The SSL configuration always gets parsed as a string, resulting (for example in `mysql`) in the following:

```
TypeError: Unknown SSL profile '{"ca":"-----BEGIN CERTIFICATE-----
```

## 3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real

Codesandbox:
https://codesandbox.io/s/dazzling-burnell-7gsyj?file=/src/index.js:0-197

```js
const knex = require("knex")({
  client: "mysql",
  connection:
    'mariadb://foo:bar@foobar.com/foo?ssl={"ca": "--- BEGIN CERTIFICATE --- ..."}'
});

console.log(knex.client.connectionSettings);
```

Result:

```js
{
  database: 'foo',
  host: 'foobar.com',
  user: 'foo',
  password: 'bar',
  ssl: '{"ca": "--- BEGIN CERTIFICATE --- ..."}'
}
```

Expected:

```js
{
  database: 'foo',
  host: 'foobar.com',
  user: 'foo',
  password: 'bar',
  ssl: : {
    ca: "--- BEGIN CERTIFICATE --- ..."
  }
}
```

petetnt:
This PR fixes #4628 by adding support for parsing JSON-like URL parameters for mysql/mariadb-clients (mysqljs/mysql, node-mysql2)


kibertoad:
Thanks! Can you also add a very simple integration test that would ensure MySQL actually works when such connection params are passed to it?

kibertoad:
@petetnt CI is failing, unfortunately.

petetnt:
Dang, Ill look into that on monday! Tests were passing locally, but only ran a subset of them as installing stuff on M1 Macs was a bit of a pain :)

kibertoad:
@petetnt There is docker-compose config available for all the dbs, check scripts in package.json

petetnt:
I was using that, I meant that eg. `mysql` doesn't have an ARM64 build at all, so one needs to specify `platform: linux/x86_64` on the docker-compose file etc. to pull those images and run them with Rosetta and so on.

The failing test is 

```

  Failed Tests: There were 2 failures

    #4628, supports mysql / mariadb client JSON parameters

      ✖ should be loosely deeply equivalent
      ✖ should be loosely deeply equivalent
```

I _think_ that the prettier fixing removing those "extra" `/`s as unnecessary escapes broke the equivalency checks there, should be an easy fix when I get pack to my computer 👍

petetnt:
The actual issue was port 0000 being parsed to 0, which was fixed iin https://github.com/knex/knex/pull/4629/commits/4fda93e1266d86ec842770e37d6b4463b14d090e :)

kibertoad (approved):

PR #4633, opened 2021-08-21
fix(mssql): columnInfo() now works with case-sensitive database [fixes #4573]

### LINKED ISSUE #4573: SQL Server: columnInfo() fails for a case-sensitive database
# Environment

Knex version: Encountered in v0.21.12, but review of the code for HEAD shows it should still repro there.
Database + version: MS SQL Server 2019 Developer Edition
OS: macOS 11

For MSSql tag: @smorey2

# Bug

Attempting to retrieve column info on a database using a case-sensitive collation fails:

```
    err: {
      "type": "RequestError",
      "message": "select * from information_schema.columns where table_name = @p0 and table_catalog = @p1 and table_schema = 'dbo' - Invalid object name 'information_schema.columns'.",
      "stack":
          RequestError: select * from information_schema.columns where table_name = @p0 and table_catalog = @p1 and table_schema = 'dbo' - Invalid object name 'information_schema.columns'.
              at createQueryBuilder (SRCROOT/node_modules/knex/lib/util/make-knex.js:313:26)
              at knex (SRCROOT/node_modules/knex/lib/util/make-knex.js:99:12)
              at QueryBuilder.columnInfo (SRCROOT/node_modules/objection/lib/queryBuilder/QueryBuilder.js:829:29)
              at fetchTableMetadata (SRCROOT/node_modules/objection/lib/model/modelTableMetadata.js:47:8)
              at Function.fetchTableMetadata (SRCROOT/node_modules/objection/lib/model/Model.js:541:12)
              at Object.mapper (SRCROOT/node_modules/objection/lib/queryBuilder/join/RelationJoiner.js:64:34)
      "code": "EREQUEST",
      "originalError": {
        "type": "Error",
        "message": "Invalid object name 'information_schema.columns'.",
        "stack":
            Error: Invalid object name 'information_schema.columns'.
                at handleError (SRCROOT/node_modules/mssql/lib/tedious/request.js:364:19)
                at Connection.emit (events.js:314:20)
                at Parser.<anonymous> (SRCROOT/node_modules/tedious/lib/connection.js:832:12)
                at Parser.emit (events.js:314:20)
                at Parser.<anonymous> (SRCROOT/node_modules/tedious/lib/token/token-stream-parser.js:37:14)
                at Parser.emit (events.js:314:20)
                at addChunk (SRCROOT/node_modules/readable-stream/lib/_stream_readable.js:298:12)
                at readableAddChunk (SRCROOT/node_modules/readable-stream/lib/_stream_readable.js:280:11)
                at Parser.Object.<anonymous>.Readable.push (SRCROOT/node_modules/readable-stream/lib/_stream_readable.js:241:10)
                at Parser.Object.<anonymous>.Transform.push (SRCROOT/node_modules/readable-stream/lib/_stream_transform.js:139:32)
        "info": {
          "name": "ERROR",
          "event": "errorMessage",
          "number": 208,
          "state": 1,
          "class": 16,
          "message": "Invalid object name 'information_schema.columns'.",
          "serverName": "0f6f0cffb249",
          "procName": "",
          "lineNumber": 1
        }
      }
```

3. This needs a set-up MSSQL database to repro. Create it with a case-sensitive collation e.g. `CREATE DATABASE CaseSensitiveDemo COLLATE Latin1_General_100_CS_AI_SC_UTF8`. Then you can try running the query.

The issue proves to be that SQL Server requires the `INFORMATION_SCHEMA.COLUMNS` part be capitalized exactly like that.

But at present, the dialect is capitalizing the column names it is selecting appropriately, but not the info schema + columns part it's selecting from. Fixing this should be a matter of capitalizing that. Automating testing the fix OTOH might require a dedicated test database, or switching the existing test database to use a case-sensitive collation so this doesn't sneak through again (probably the better move, TBH; any query that works on case-sensitive will also work on case-insensitive). (Tedious presently can't handle varchar UTF-8 columns properly https://github.com/tediousjs/tedious/issues/1143, so probably don't use quite that collation in knex testing just yet.)

Line at issue, in the latest tagged release: https://github.com/knex/knex/blob/0.95.5/lib/dialects/mssql/query/mssql-querycompiler.js#L417

And in HEAD as of just now: https://github.com/knex/knex/blob/c335fda65723e73c9de5d1a92fe7444f0cfb4071/lib/dialects/mssql/query/mssql-querycompiler.js#L432

You can see this at work in this `sqlcmd` session transcript with interspersed commentary:

Here's the select target as written, with the error direct from SQL Server:

```
1> select * from information_schema.columns where table_name = 'whatever'
2> go
Msg 208, Level 16, State 1, Server 49df3e21e80c, Line 1
Invalid object name 'information_schema.columns'.
```

What if we capitalize just one part or the other? Still no luck:

```
1> select * from INFORMATION_SCHEMA.columns where table_name = 'whatever'
2> go
Msg 208, Level 16, State 1, Server 49df3e21e80c, Line 1
Invalid object name 'INFORMATION_SCHEMA.columns'.
1> select * from information_schema.COLUMNS where table_name = 'whatever'
2> go
Msg 208, Level 16, State 1, Server 49df3e21e80c, Line 1
Invalid object name 'information_schema.COLUMNS'.
```

What if we capitalize both parts, like this is FORTRAN or something? Oh look, it works, finally:

```
1> select * from INFORMATION_SCHEMA.COLUMNS where table_name = 'whatever'
2> go
TABLE_CATALOG TABLE_SCHEMA TABLE_NAME COLUMN_NAME ORDINAL_POSITION COLUMN_DEFAULT IS_NULLABLE DATA_TYPE CHARACTER_MAXIMUM_LENGTH CHARACTER_OCTET_LENGTH NUMERIC_PRECISION NUMERIC_PRECISION_RADIX NUMERIC_SCALE DATETIME_PRECISION CHARACTER_SET_CATALOG CHARACTER_SET_SCHEMA CHARACTER_SET_NAME COLLATION_CATALOG COLLATION_SCHEMA COLLATION_NAME DOMAIN_CATALOG DOMAIN_SCHEMA DOMAIN_NAME
------------- ------------ ---------- ----------- ---------------- -------------- ----------- --------- ------------------------ ---------------------- ----------------- ----------------------- ------------- ------------------ --------------------- -------------------- ------------------ ----------------- ---------------- -------------- -------------- ------------- -----------
(0 rows affected)
```

jeremy-w:
Before, the SQL used to implement `columnInfo()` for the mssql driver referenced `information_schema.columns`.

But in a database with a case-sensitive collation, both of those parts must be capitalized, as `INFORMATION_SCHEMA.COLUMNS`, or you just get an error [#4573]. (It doesn't seem sensitive to how you capitalize the column names, though. 🤷 )

Now, that capitalized form is used when interacting with mssql.

# Question for Reviewer: Should I just switch the default mssql test database to use a case-sensitive collation?
In order to test this change, I needed to create a new database with a case-sensitive collation, then tear it down. This required digging around to grab the database config used in the integration test suite.

It would be cleaner to change the default test database for MSSQL to be created with a case-sensitive collation, in which case, every test would be a test of case-sensitive behavior. This does create a risk of writing SQL that relies on case-sensitivity in its function, which then fails under case-insensitive collations, but I consider that rather less likely than case-insensitive code that fails when case-sensitivity is enabled.

If we did change it to use a case-sensitive collation, I _think_ we could remove the dedicated test suite I added, though each of those tests did turn out to be required to beat this functionality into shape, because I repeatedly missed various ways this could go wrong:

- "I'll just addproperty": fails during alter that updates the comment
- "I'll base add/update on whether we're creating or altering": fails when a table/column is created without any comment, and the alter needs to add, not update

Finally I came around to, "We just have to check and add/update based on what we find each time," which is the implementation you see here.

### DIFF
--- a/lib/dialects/mssql/query/mssql-querycompiler.js
+++ b/lib/dialects/mssql/query/mssql-querycompiler.js
@@ -429,7 +429,8 @@ class QueryCompiler_MSSQL extends QueryCompiler {
       schema = this.client.customWrapIdentifier(schema, identity);
     }
 
-    let sql = `select [COLUMN_NAME], [COLUMN_DEFAULT], [DATA_TYPE], [CHARACTER_MAXIMUM_LENGTH], [IS_NULLABLE] from information_schema.columns where table_name = ? and table_catalog = ?`;
+    // GOTCHA: INFORMATION_SCHEMA.COLUMNS must be capitalized to work when the database has a case-sensitive collation. [#4573]
+    let sql = `select [COLUMN_NAME], [COLUMN_DEFAULT], [DATA_TYPE], [CHARACTER_MAXIMUM_LENGTH], [IS_NULLABLE] from INFORMATION_SCHEMA.COLUMNS where table_name = ? and table_catalog = ?`;
     const bindings = [table, this.client.database()];
 
     if (schema) {
--- a/test/integration/query/additional.js
+++ b/test/integration/query/additional.js
@@ -300,7 +300,7 @@ module.exports = function (knex) {
         [drivers.SQLite]: "SELECT name FROM sqlite_master WHERE type='table';",
         [drivers.Oracle]: 'select TABLE_NAME from USER_TABLES',
         [drivers.MsSQL]:
-          "SELECT table_name FROM information_schema.tables WHERE table_schema='dbo'",
+          "SELECT table_name FROM INFORMATION_SCHEMA.TABLES WHERE table_schema='dbo'",
       };
       return knex
         .raw(tables[knex.client.driverName])
@@ -417,7 +417,7 @@ module.exports = function (knex) {
           );
           tester(
             'mssql',
-            "select [COLUMN_NAME], [COLUMN_DEFAULT], [DATA_TYPE], [CHARACTER_MAXIMUM_LENGTH], [IS_NULLABLE] from information_schema.columns where table_name = ? and table_catalog = ? and table_schema = 'dbo'",
+            "select [COLUMN_NAME], [COLUMN_DEFAULT], [DATA_TYPE], [CHARACTER_MAXIMUM_LENGTH], [IS_NULLABLE] from INFORMATION_SCHEMA.COLUMNS where table_name = ? and table_catalog = ? and table_schema = 'dbo'",
             ['datatype_test', 'knex_test'],
             {
               enum_value: {
@@ -493,7 +493,7 @@ module.exports = function (knex) {
           );
           tester(
             'mssql',
-            "select [COLUMN_NAME], [COLUMN_DEFAULT], [DATA_TYPE], [CHARACTER_MAXIMUM_LENGTH], [IS_NULLABLE] from information_schema.columns where table_name = ? and table_catalog = ? and table_schema = 'dbo'",
+            "select [COLUMN_NAME], [COLUMN_DEFAULT], [DATA_TYPE], [CHARACTER_MAXIMUM_LENGTH], [IS_NULLABLE] from INFORMATION_SCHEMA.COLUMNS where table_name = ? and table_catalog = ? and table_schema = 'dbo'",
             null,
             {
               defaultValue: null,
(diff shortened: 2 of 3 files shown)


kibertoad (approved):

kibertoad:
Thank you!

kibertoad:
Released in 0.95.11

Issue #4514, opened 2021-06-06
Expand withRecursive database support via supporting column name declarations


code-ape:
# Environment

**Knex version:** 0.95.5
**Database + version:** Sqlite 3.x, Oracle 18c, Postgres 13.x, SQL Server 2016
**OS:** Ubuntu Linux, Windows Server

# Feature discussion / request

1. Explain what is your use case

I need support for recursive `WITH` syntax on Oracle DB and SQL Server in addition to Sqlite and Postgres. All four databases support recursive CTE `WITH` statements so this is a question of how to best implement this in Knex (assuming it's agreeable to do so).

All four databases support column declaration (`WITH temp_table (column1, column2, ...) AS (...)`) and for two of them column declaration enables CTE `WITH` functionality. Of those four, two support the `RECURSIVE` syntax word and two don't. For Sqlite and Postgres the `RECURSIVE` operator is supported and for Oracle and MSSQL the functionality is enabled by simply declaring the column names. Here is a quick breakdown.

| Database | Supports `RECURSIVE` | Supports column names | Reference |
|----------|-------------------------|--------------------------|-----------|
| Sqlite | Yes | Yes | [Link](https://sqlite.org/lang_with.html) |
| Postgres | Yes | Yes | [Link](https://www.postgresql.org/docs/13/queries-with.html) |
| Oracle | No | Yes | [Link](https://livesql.oracle.com/apex/livesql/file/tutorial_GQMLEEPG5ARVSIFGQRD3SES92.html) |
| MSSQL | No | Yes |  [Link](https://docs.microsoft.com/en-us/sql/t-sql/queries/with-common-table-expression-transact-sql?view=sql-server-ver15) |

Since all databases support this and support the column names this seems like a reasonable abstraction to provide a recursive with over top of all of them.

2. Explain what kind of feature would support this

Changing `knex.withRecursive` to support column names.

3. Give some API proposal, how the feature should work

```ts
knex.withRecursive('with_alias', ['column1', 'column2'], (qb) => {...}).select('*').from('with_alias')
// On Sqlite and Postgres
//  WITH RECURSIVE `with_alias` (`column1`, `column2`) AS (...) SELECT * FROM `with_alias`
// On Oracle and MSSQL
//  WITH `with_alias` (`column1`, `column2`) AS (...) SELECT * FROM `with_alias`
```

jeremy-w:
[...]

PR #4639, opened 2021-08-23
fix(mssql): Generate valid SQL for withRecursive() [#4514]

jeremy-w:
mssql forbids use of the RECURSIVE keyword required by some other
dialects.

The approach taken here maximizes subclass reuse and leaking
changes to the query outside `with` generation. If we skipped
restoring the recursive labels using `undoList`, the SQL would
generate fine here, but the test suite fails:
The next-tested driver fails its test, because clone() preserves the
top-level statement's `recursive = false` change, so it's shy one
`RECURSIVE` in the generated SQL.

This addresses the mssql portion of [#4514], but not the oracledb issues.
If this approach to omitting RECURSIVE looks good, I can apply the same to the
oracledb dialect. (Supporting a column alias list as separate argument, as Oracle
requires for recursive CTEs, would require deeper changes.)

### DIFF
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -9367,13 +9367,16 @@ describe('QueryBuilder', () => {
         .select('*')
         .from('secondWithClause'),
       {
+        // mssql does not allow the RECURSIVE keyword.
         mssql:
-          'with recursive [firstWithClause] as (with recursive [firstWithSubClause] as ((select [foo] from [users]) as [foz]) select * from [firstWithSubClause]), [secondWithClause] as (with recursive [secondWithSubClause] as ((select [bar] from [users]) as [baz]) select * from [secondWithSubClause]) select * from [secondWithClause]',
+          'with [firstWithClause] as (with [firstWithSubClause] as ((select [foo] from [users]) as [foz]) select * from [firstWithSubClause]), [secondWithClause] as (with [secondWithSubClause] as ((select [bar] from [users]) as [baz]) select * from [secondWithSubClause]) select * from [secondWithClause]',
         sqlite3:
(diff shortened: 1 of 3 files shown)


test/unit/query/builder.js
```diff
@@ -9367,13 +9367,16 @@ describe('QueryBuilder', () => {
         .select('*')
         .from('secondWithClause'),
       {
+        // mssql does not allow the RECURSIVE keyword.
         mssql:
-          'with recursive [firstWithClause] as (with recursive [firstWithSubClause] as ((select [foo] from [users]) as [foz]) select * from [firstWithSubClause]), [secondWithClause] as (with recursive [secondWithSubClause] as ((select [bar] from [users]) as [baz]) select * from [secondWithSubClause]) select * from [secondWithClause]',
+          'with [firstWithClause] as (with [firstWithSubClause] as ((select [foo] from [users]) as [foz]) select * from [firstWithSubClause]), [secondWithClause] as (with [secondWithSubClause] as ((select [bar] from [users]) as [baz]) select * from [secondWithSubClause]) select * from [secondWithClause]',
```
kibertoad, line 9372: can  you also add integration test for this to ensure we don't break it again in the future?

jeremy-w: i've pushed a revised version with integration tests.

jeremy-w:
first pass at integration test had errors on a couple platforms i wasn't testing locally:

- postgres: the format of returning must vary across drivers - pg was returning [{id: 1}] rather than just [1]
- mysql: its increments() column is apparently unsigned, and so it refused to let a (signed) integer column reference the id column.

i revised it to omit the keys and fkeys (they weren't the point of the test - just superfluous "realism"), and everything passed the next time around. 👍

kibertoad (approved):

kibertoad:
Thanks a lot! Could you also add same for Oracle?

kibertoad:
Released in 0.95.11

PR #4640, opened 2021-08-23
Add indexType support to MySQL unique() schema

glensc:
Implements index type for https://github.com/knex/knex/issues/4637


kibertoad:
What's the status on this?

PR #4644, opened 2021-08-24
Fix for comment escaping in postgres

### LINKED ISSUE #4641: Regex Error in Safari / WebKit browsers (including Gnome Web)
The comment regex in the Postgres column compiler has a look-behind. This isn't supported by Safari or WebKit, and so causes any web bundled app to crash.

This line:

```js
const commentEscapeRegex = /(?<!')'(?!')/g;
```
This file:
https://github.com/knex/knex/blob/master/lib/dialects/postgres/schema/pg-columncompiler.js#L7

Here's an error output from `Gnome Web (Midori)`:
image

rathboma:
The current implementation will not escape multiple sequences of single quotes, but it should. Examples:

`COMMENT ON table IS 'This is a '''comment'` <- All three need escaping
`COMMENT ON table IS 'Double '' quotes'` <- both need escaping here, or else `''` turns into `'` when returned to the user.

Actually also fixes #4641


kibertoad:
Could you please add both unit and integration test for this?

OlivierCavadenti:
Any news on this PR ? @rathboma

rathboma:
sorry, let me jump back on this.

OlivierCavadenti:
I update tests.
By the way, pre-quoted strings are not possible because all ' are escaped, it's what we want ? 
@kibertoad

rathboma:
@OlivierCavadenti @kibertoad Thanks for adding the tests.

Personal opinion: I think escaping all ' characters is the correct behavior, and this is the right way to do it.

@kibertoad hoping we can merge this!

kibertoad:
@rathboma Can you also add integration tests that would actually execute SQL generated in such cases and ensure it works correctly in real queries?

kibertoad:
@rathboma ping

OlivierCavadenti:
Rebased and extended the fix to cover table comments in postgres, mysql, and oracle, which were also vulnerable to unescaped single quotes.

test/unit/schema-builder/mysql.js
```diff
@@ -1297,7 +1325,7 @@ module.exports = function (dialect) {
 
       equal(1, tableSql.length);
       expect(tableSql[0].sql).to.equal(
-        "create table `test` (`column1` text comment 'The table\\'s first column and it\\'s escaped')"
+        "create table `test` (`column1` text comment 'The table\\\\\\'s first column and it\\\\\\'s escaped')"
```
tgriesser, line 1332: Is this change correct? It seems like previously we allowed escaping the quotes with backslashes and now we don't?

tgriesser (commented):
@OlivierCavadenti I don't think I'd consider this a patch change, as it changes the behavior given the same input - if you're pre-escaping your strings it seems it's going to output differently than it currently does. Would you agree?

test/unit/schema-builder/mysql.js
OlivierCavadenti: Hum...
The previous regex (?<!\\) assumed users would pre-escape their quotes manually, which was fragile, for example it didn't handle \\\' correctly.
CodeQL alerts on this PR confirm backslashes weren't escaped at all in table comments (https://github.com/knex/knex/security/code-scanning/27)
The new approach lets knex handle all escaping: first backslashes, then quotes which is safer but yeah, that will break code where anybody was already pre escaping their comment strings.

OlivierCavadenti:
@tgriesser agreed, not a patch. Pre-escaped comments now get double-escaped, so it's breaking. Set the `breaking` label.

PR #4648, opened 2021-08-25
Clean up SQLite DDL handling and move all operations to the parser-based approach

nickrum:


kibertoad:
Why was a bunch of tests dropped?

nickrum:
The dropped tests were all testing the functionality of `_doReplace()` which I removed as it is no longer needed. The cases that were tested should all be covered by the parser/compiler unit tests.

kibertoad (approved):

PR #4649, opened 2021-08-25
Fix adding a column with a foreign key constraint in SQLite

nickrum:
If more than one query is passed to queryArray(), the statement producer of every query has to be executed.


test/integration2/schema/foreign-keys.spec.js
```diff
@@ -221,6 +220,25 @@ describe('Schema', () => {
...
+
+            await knex.schema.alterTable('foreign_keys_table_one', (table) => {
+              table
+                .integer('fkey_new')
+                .unsigned()
+                .notNull()
+                .references('foreign_keys_table_two.id');
+            });
+
+            const fks = await knex.raw(
```
kibertoad, line 237: why not make this test dialect-indepentent by triggering constraint violation?

lib/execution/runner.js
```diff
@@ -238,7 +238,7 @@ class Runner {
 
     const results = [];
     for (const query of queries) {
-      results.push(await this.query(query));
+      results.push(await this.queryArray([query]));
     }
```
kibertoad, line 242: what's the benefit of calling queryArray with single element?

test/integration2/schema/foreign-keys.spec.js
nickrum: The problem is that SQLite is the only dialect that uses `statementsProducer()`. When adding a column and adding a foreign key constraint at the same time, the statements producer to add the constaint will simply not execute. Thus it is not possible to trigger a constraint violation, as there is no constraint to be violated. That's why the test is checking for the presence of the constraint.

lib/execution/runner.js
nickrum: This ensures that the statements producer is executed. Maybe we should move the statements producer handling to `runner.query()`.

kibertoad:
Released in 0.95.11

PR #4652, opened 2021-08-27
fix(oracledb): withRecursive: omit invalid RECURSIVE keyword, include column list [#4514]

jeremy-w:
Before, the Oracle query compiler included an erroneous RECURSIVE keyword, and you could not include a column list, as required in Oracle SQL when writing an rCTE.

Now, both are fixed.

# TODO
- [x] Agree `with` args validation function with how they are used in `withWrapped`. Right now you could pass validation but fail during usage, because I'm not interpreting the 2 vs 3 arg versions the same in both places.
- [ ] Open docs PR with corresponding changes

### DIFF
--- a/lib/dialects/oracledb/query/oracledb-querycompiler.js
+++ b/lib/dialects/oracledb/query/oracledb-querycompiler.js
@@ -201,6 +201,29 @@ class Oracledb_Compiler extends Oracle_Compiler {
     return sql;
   }
 
+  with() {
+    // WITH RECURSIVE is a syntax error in Oracle SQL.
+    // So mark all statements as non-recursive, generate the SQL, then restore.
+    // This approach ensures any changes in base class with() get propagated here.
+    const undoList = [];
+    if (this.grouped.with) {
+      for (const stmt of this.grouped.with) {
+        if (stmt.recursive) {
+          undoList.push(stmt);
+          stmt.recursive = false;
+        }
+      }
+    }
+
+    const result = super.with();
+
+    // Restore the recursive markings, in case this same query gets cloned and passed to other drivers.
+    for (const stmt of undoList) {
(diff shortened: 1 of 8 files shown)


kibertoad:
@jeremy-w API proposed in https://github.com/knex/knex/issues/4514 looks sensible. Do we need anything beyond that?

jeremy-w:
That should do it. It’s the three-arg variant with a column array in between the alias & builder/raw/callback variant I included in the list. Just wanted confirmation on the intended API changes before proceeding with QueryBuilder changes. Thanks!

jeremy-w:
The integration test for oracledb is failing with master (b5225179ce8760c61418b4bd1c2cbb1a742c662a) for me right now:

```
  320 passing (24s)
  47 pending
  2 failing

  1) Schema
       alter nullable
         oracledb
           setNullable
             sets column to be nullable:
     Error: SELECT 1 - ORA-00923: FROM keyword not found where expected
  

  2) Schema
       alter nullable
         oracledb
           dropNullable
             sets column to be not nullable:
     Error: SELECT 1 - ORA-00923: FROM keyword not found where expected
```

I ran the integration test suite with my changes, and they do not introduce any further breakage. 😬

kibertoad (approved):

kibertoad:
Released in 0.95.11

kibertoad:
@jeremy-w Could you also update TS types for this change?

jeremy-w:
Type updates were included. https://github.com/knex/knex/pull/4652/commits/d79fa9e89635eb889237664aab3a10e3649f796e#diff-093ad82a25aee498b11febf1cdcb6546e4d223ffcb49ed69cc275ac27ce0ccce

I think the types might erroneously include a withRaw function that has been removed, though.

PR #4656, opened 2021-08-28
drop onDelete/onUpdate for ColumnBuilder

### LINKED ISSUE #4620: Missleading Typescript Type For onDelete on onUpdate Function
# Environment

Knex version:
Latest release knex 95.8

@ lorefnon.

# Bug

1. At the Moment, Knex only support onDelete and onUpdate with foreignKey constraint but typescript suggest that onDelete and onUpdate can be chained off Column 

image

maximelkin:
fixes #4620
This breaks wrong code, when onDelete/onUpdate called not in context of foreign constraints.
But also breaks valid code, i.e.:
```ts
table.integer('num').references('non_exist.id').index('num_idx').onDelete('CASCADE');
 ```
(index() method return `ColumnBuilder` interface, not `ReferencingColumnBuilder`)

This probably could be fixed with rewriting code to extending previous interface with new methods instead of current approach, idk is it worth it?

@lorefnon

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1952,8 +1952,6 @@ export declare namespace Knex {
     /** @deprecated */
     unique(indexName?: string): ColumnBuilder;
     references(columnName: string): ReferencingColumnBuilder;
-    onDelete(command: string): ColumnBuilder;
-    onUpdate(command: string): ColumnBuilder;
     defaultTo(value: Value | null, options?: DefaultToOptions): ColumnBuilder;
     unsigned(): ColumnBuilder;
     notNullable(): ColumnBuilder;

### CODE REVIEW

**Comment by kibertoad** [comment]
@zeotuan Any thoughts?

**Comment by zeotuan** [comment]
if we change the `index(indexName?:string): ColumnBuilder` to `index(indexName?:string): ReferrencingColumnBuilder` It would temporarily solve this problem right? But i am not sure if it would be a good solution though. @maximelkin @lorefnon what do you guys think?

**Comment by maximelkin** [comment]
Now this is more close to knex js implementation, but this is kinda hack.

**Comment by maximelkin** [comment]
@zeotuan can you check PR?

**Comment by kibertoad** [comment]
@zeotuan Ping?

**Comment by zeotuan** [comment]
I tried this code and got the same result as your @maximelkin. What is the advantage of yours compared to this?  

```js
type ReferencingColumnBuilder = {
    [K in keyof ColumnBuilder]: (...args) =>  ReferencingColumnBuilder
  } & {
    inTable(tableName: string): ReferencingColumnBuilder;
    deferrable(type: deferrableType): ReferencingColumnBuilder;
    withKeyName(keyName: string): ReferencingColumnBuilder;
    onDelete(command: string): ReferencingColumnBuilder;
    onUpdate(command: string): ReferencingColumnBuilder;
  };
```
Also if you believe your solution is better than the one above can you make it this way so it's easier to understand

```js
type ReferencingColumnBuilder = {
    [K in keyof ColumnBuilder]: ColumnBuilder[K] extends (...args: infer TArgs) => any 
      ? (...args: TArgs) => ReferencingColumnBuilder 
      : ColumnBuilder[K];
  } & {
    inTable(tableName: string): ReferencingColumnBuilder;
    deferrable(type: deferrableType): ReferencingColumnBuilder;
    withKeyName(keyName: string): ReferencingColumnBuilder;
    onDelete(command: string): ReferencingColumnBuilder;
    onUpdate(command: string): ReferencingColumnBuilder;
  };
```

maximelkin:
@zeotuan first variant erases args types

zeotuan:
@maximelkin You are right! What do you think about this one? 
```js
 // patched ColumnBuilder methods to return ReferencingColumnBuilder with new methods
  // relies on ColumnBuilder returning only ColumnBuilder
  type ReferencingColumnBuilder = {
    [K in keyof ColumnBuilder]: (...arg: Parameters<ColumnBuilder[K]> ) => ReferencingColumnBuilder
  } & {
    inTable(tableName: string): ReferencingColumnBuilder;
    deferrable(type: deferrableType): ReferencingColumnBuilder;
    withKeyName(keyName: string): ReferencingColumnBuilder;
    onDelete(command: string): ReferencingColumnBuilder;
    onUpdate(command: string): ReferencingColumnBuilder;
  };
```

maximelkin:
@zeotuan this is nice, updated PR

PR #4659, opened 2021-09-03
fix(oracledb): Support setNullable, dropNullable

jeremy-w:
Before, these functions would fail when run, as indicated by test suite failures when testing against oracledb.

Now, they work.

The existing tests were also not checking the error message as intended, so I have corrected that. (I originally thought to include a "make sure errorMessage is a string" check, but since only a fraction of the drivers had spec'd error messages, I backed that out.)


jeremy-w:
This fixes the existing test failures I discovered while working on https://github.com/knex/knex/pull/4652.

kibertoad (approved):

kibertoad:
Thank you!

Issue #4665, opened 2021-09-06
Filtered catch not transmited after a transaction (for Bluebird promises)


Plaffont:
# Environment

Knex version: 0.95.8
Database + version: PostgreSQL 10.6
OS: Ubuntu

# Bug

When using a [Bluebird filtered catch](http://bluebirdjs.com/docs/api/catch.html#filtered-catch) after a transaction only the first parameter is transmited to the actual promise. This parameter is then considered has the handler by bluebird and then causing distinct troubles.
For instance, with a filtered catch on a Bookshelf error like a [NoRowsDeletedError](https://bookshelfjs.org/api.html#section-Model-static-NoRowsDeletedError) an unhandled error is thrown:
`Cannot set property 'message' of undefined`

With a simple custom error ther is an error trying to call the constructor: 
`TypeError: Class constructor FunctionalError cannot be invoked without 'new'` 

Here is a simple test code:
https://runkit.com/plaffont/613625df7c945d000988a650

To fix that we had to change the definition of the catch function in the transaction class (passing all parameters).

KristjanTammekivi:
Knex removed bluebird a long time ago, the Promise returned from knex.transaction is vanilla

elhigu:
merged, the change did seem to be implemented backwards compatible way.

PR #4666, opened 2021-09-06
Passing all arguments to _promise.catch function #4665

Plaffont:
This is a pull request to fix the issue concerning the use of filtered catch after a transaction (see [issue](https://github.com/knex/knex/issues/4665))


kibertoad (approved):

kibertoad:
Thank you!

PR #4677, opened 2021-09-14
Maybe return rows.affectedRows

FrankFang:


kibertoad:
Could you please add tests for this?

kibertoad:
Is it possible to implement this for other dbs as well?

elhigu:
Also this might be breaking change.

FrankFang:
@kibertoad 
eg. user Alice creates/updates10 users, but 2 of them fail.
The expected message is "8 users are created, but 2 not".

affectedRows helps me to get the number of succeed items. 
I could not figure out any other methods to get the number.

FrankFang:
yes, I add some tests. But how could I trigger a ci test ?

OlivierCavadenti:
> yes, I add some tests. But how could I trigger a ci test ?

I trigger them, you have some issues on tests

test/integration2/query/insert/inserts.spec.js
```diff
@@ -1157,6 +1161,11 @@ describe('Inserts', function () {
             );
           })
           .then(function (rows) {
+            if (isMysql(knex)) {
+              expect(rows.length).to.equal(2);
+            } else {
+              expect(rows.length).to.equal(1);
+            }
```
OlivierCavadenti, line 1168: You have a duplicate "expect(rows.length).to.equal(1);" below

OlivierCavadenti:
> Is it possible to implement this for other dbs as well?

I push a commit to fix test. Do you want I check for other dbs ? Maybe I can finish this for 1.0.0 ?

FrankFang:
No, I am not familiar with other dbs.

kibertoad:
If it is feasible to also have it for other dbs, that would be strongly preferable.

Issue #4681, opened 2021-09-17
[Oracle] Cannot make text column "notNullable"


aidenfoxx:
# Environment

Knex version: 0.95.11
Database + version: OracleDB 18
OS: Ubuntu

# Bug

When trying to alter a text field to be "notNullable" Oracle throws the error:

```
Error: alter table "test_table" modify "test_column" clob not null - ORA-22296: invalid ALTER TABLE option for conversion of LONG datatype to LOB
```

This can be reproduced with the following test case:

```
    await knex.schema.createTable('test_table', function (table) {
      table.text('test_column');
    });

    await knex.schema.table('test_table', function (table) {
      table.text('test_column').notNullable().alter();
    })
```

There is no data in the table, so it's not an issue of existing data containing null values.

@atiertant

aidenfoxx:
This isn't the first time I've had similar issues with knex. This would be easily solved if we had some way to communicate that we don't want to update the underlying column type; just the constraints. For example, the following query works correctly:

`ALTER TABLE "test_table" MODIFY "test_column" DEFAULT NULL;`

It would be nice to have some kind of way to only target constraints with a syntax like:

```
table('test_column').notNullable().alter();

OR

table.column('test_column').notNullable().alter();
```

kibertoad:
@aidenfoxx Funny that you mention that. 0.95.11 introduced exactly that, `table.setNullable()`. Can you give it a try?

aidenfoxx:
Good timing! Yes, that absolutely works. Sorry for creating the unnecessary issue!

kibertoad:
No worries!

aidenfoxx:
@kibertoad Is there any particular reason why `dropNullable` isn't supported for SQLite?

kibertoad:
@aidenfoxx Because SQLite itself doesn't support that operation, I think, when using `.alter()` approach the whole table gets rebuilt.
If you could implement a working approach for SQLite, would be more than happy to accept PR for that.

aidenfoxx:
If I get a moment this weekend I'll look into adding it. I think the `dropNullable` function is great for overall compatibility, so it would be nice to have it working for all vendors. 😄

kibertoad:
@aidenfoxx That would be amazing, thank you!

PR #4684, opened 2021-09-18
Added setNullable support to SQLite

aidenfoxx:
As discussed in #4681 I've added support for `setNullable` and `dropNullable` to the SQLite dialect.

Hopefully I didn't miss anything!

### DIFF
--- a/lib/dialects/sqlite3/schema/ddl.js
+++ b/lib/dialects/sqlite3/schema/ddl.js
@@ -372,7 +372,7 @@ class SQLite3_DDL {
 
         const newTable = compileCreateTable(parsedTable, this.wrap);
 
-        return this.alter(newTable, createIndices);
+        return this.generateAlterCommands(newTable, createIndices);
       },
       { connection: this.connection }
     );
--- a/lib/dialects/sqlite3/schema/ddl.js
+++ b/lib/dialects/sqlite3/schema/ddl.js
@@ -344,6 +344,40 @@ class SQLite3_DDL {
     );
   }
 
+  setNullable(column, isNullable) {
+    return this.client.transaction(
+      async (trx) => {
+        this.trx = trx;
+
+        const { createTable, createIndices } = await this.getTableSql();
+
+        const parsedTable = parseCreateTable(createTable);
+        const parsedColumn = parsedTable.columns.find((c) =>
+          isEqualId(column, c.name)
+        );
+
+        if (!parsedColumn) {
+          throw new Error(
+            `.setNullable: Column ${column} does not exist in table ${this.tableName()}.`
+          );
+        }
+
+        parsedColumn.constraints.notnull = isNullable
+          ? null
+          : { name: null, conflict: null };
+
+        parsedColumn.constraints.null = isNullable
+          ? parsedColumn.constraints.null
+          : null;
+
+        const newTable = compileCreateTable(parsedTable, this.wrap);
+
+        return this.alter(newTable, createIndices);
+      },
+      { connection: this.connection }
+    );
+  }
+
   async alter(newSql, createIndices, columns) {
     await this.createNewTable(newSql);
     await this.copyData(columns);
--- a/lib/dialects/sqlite3/schema/sqlite-tablecompiler.js
+++ b/lib/dialects/sqlite3/schema/sqlite-tablecompiler.js
@@ -266,8 +266,16 @@ class TableCompiler_SQLite3 extends TableCompiler {
   }
 
   _setNullableState(column, isNullable) {
-    const fnCalled = isNullable ? '.setNullable' : '.dropNullable';
-    throw new Error(`${fnCalled} is not supported for SQLite.`);
+    const compiler = this;
+
+    this.pushQuery({
+      sql: `PRAGMA table_info(${this.tableName()})`,
+      output(pragma) {
+        return compiler.client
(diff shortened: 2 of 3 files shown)


lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -344,6 +344,40 @@ class SQLite3_DDL {
     );
   }
 
+  setNullable(column, isNullable) {
+    return this.client.transaction(
```
kibertoad, line 348: does sqlite even support transactions?

lib/dialects/sqlite3/schema/sqlite-tablecompiler.js
```diff
@@ -266,8 +266,16 @@ class TableCompiler_SQLite3 extends TableCompiler {
   }
 
   _setNullableState(column, isNullable) {
-    const fnCalled = isNullable ? '.setNullable' : '.dropNullable';
-    throw new Error(`${fnCalled} is not supported for SQLite.`);
+    const compiler = this;
+
+    this.pushQuery({
+      sql: `PRAGMA table_info(${this.tableName()})`,
```
kibertoad, line 272: We've been aiming to make SQLite support in knex more friendly towards use-cases where one might want to use knex as a pure query builder and just generate SQL; to that end it is preferable to not have this kind of pseudo-sql.

`foreign` is a better example of multi-statement operation, see how it utilizes `statementsProducer`

lib/dialects/sqlite3/schema/ddl.js
```diff
@@ -344,6 +344,40 @@ class SQLite3_DDL {
...
+          ? null
+          : { name: null, conflict: null };
+
+        parsedColumn.constraints.null = isNullable
+          ? parsedColumn.constraints.null
+          : null;
+
+        const newTable = compileCreateTable(parsedTable, this.wrap);
+
+        return this.alter(newTable, createIndices);
```
kibertoad, line 375: It is preferable to avoid directly executing SQL inside `setNullable` and return an array of generated SQL statements instead.

lib/dialects/sqlite3/schema/ddl.js
kibertoad: It does support read transactions, which should be sufficient for an operation which would merely generate SQL statements

aidenfoxx: The function is just a modification of the `alterColumn` function in the same file, which also uses transactions, so one would hope so!

lib/dialects/sqlite3/schema/sqlite-tablecompiler.js
aidenfoxx: Sure, I can look into that. As mentioned, the implementation was based on the `alterColumn` functions which seemed more analogous to what I was trying to achieve. 👍

lib/dialects/sqlite3/schema/ddl.js
aidenfoxx: I see this as an extension of your point on `statementsProducer` so I'll mark this one as resolved.

lib/dialects/sqlite3/schema/sqlite-tablecompiler.js
aidenfoxx: @kibertoad Does all look correct to you now?

kibertoad: Could you also add an assertion to your test, similar to the one here https://github.com/knex/knex/blob/a75d0194cf42dceb95019ea22c6a289afb1f3397/test/integration2/schema/foreign-keys.spec.js#L65 ? To check that we are generating correct SQL statements.

aidenfoxx: I notice the foreign key test only checks postgres and sqlite SQL generation. Is there tests elsewhere that checks generic SQL generation, or should I add an assertion for the other providers (MySQL, Oracle, MSSQL etc...)?

kibertoad: sqlite would be plenty, postgre a nice bonus, but optional, everything else an overkill :)

kibertoad (approved):

aidenfoxx:
Just a question for @nickrum when he gets a chance:

In [#ed0e8a](https://github.com/knex/knex/commit/ed0e8a53df1bfa09cc24b16d5c44e92038f028a3) I notice you updated the `alterColumn` DDL to use `this.alter` rather than `this.generateAlterCommands`. Judging by the commit message, there is a good reason for the change. However, @kibertoad encouraged me to use `this.generateAlterCommands` since it returns the distinct queries of the function.

Anyway, my point is, which is correct way of doing things since there is clearly issues with both ways?

kibertoad:
@aidenfoxx Judging by skipped test, it looks more like a hack than a proper solution. I still believe that `generateAlterCommands` is a better choice.

nickrum:
@aidenfoxx The problem with `generateAlterCommands` was that the generated commands weren't executed in a transaction, so if there was an error while executing one of the alter statements, the database might be left in some in-between-state without a way to revert back. This was IMO worse than not being able to generate the commands.

Now that #4189 has been merged, using `generateAlterCommands` should be as safe as using `alter`, so I reverted the hotfix in #4896.

PR #4710, opened 2021-10-08
mssql: Support table.primary, table.unique variant with options object

jeremy-w:
Before, the mssql driver only supported `table.primary(cols, 'literalString')`, and likewise for `table.unique`. But the TypeScript types have that variant marked deprecated, which pushed you to try to write `table.primary(cols, {constraintName: 'literalString'})` / `table.unique(cols, {indexName: 'literalString'})`, which promptly failed at runtime.

Now, both variants work. If a `deferrable` configuration is provided other than `'not deferred'`, a warning will be omitted, but schema building will proceed.


jeremy-w:
Apparently mysql & sqlite3 also were not updated to cope with the non-deprecated flavor. 🤔

kibertoad:
Thank you!

Issue #3667, opened 2020-02-16
Support for 'is null' in 'order by'


sripberger:
I'd like to request a feature, which I would be happy to implement myself since it doesn't seem like it would be all that big of a change. Feel free to correct me if I'm wrong.

1. Explain what is your use case

Needing to specify the sort order of nulls within a nullable column is fairly common use case. I see that this is discussed [here][1] many years back, and it seems like the conclusion was to just use raw SQL, with `orderByRaw` being added as a shortcut for `orderBy(knex.raw('...'))`.

I see this being said there, but I really don't follow.

> This is fairly important.  You could also add an orderByRaw();  the problem is that you add tick marks around the field name in order by, so we can't provide the necessary formula, e.g. order by" ('field1' is NULL)" , "desc" or "(field1' is NOT NULL)", "asc";
> 
> _Originally posted by @mibi-r in https://github.com/knex/knex/issues/282#issuecomment-47250471_

If this is the reason that this feature was never implemented, and that reason still stands, some further explanation would be very helpful as I'm rather confused by this comment.

Using raw SQL is fine in a lot of use cases, but I'm trying to write an abstraction around [Objection][2] where it is common to use [snake case conversion][3] so that database columns and tables are named in snake case (idiomatic in SQL) while those same entities can be referred to in camel case (idiomatic in JavaScript).

Raw queries obviously don't get the snake case mappings, so using a raw query in my case would require me to extract the snake case mappers and run them on the column names myself, before building the raw SQL. Either that, or it would require users of my abstraction who are also using the mappers to be aware of the raw query, and specify their column names as they appear in database.

It would be much nicer for the regular `orderBy` to support this.

2. Explain what kind of feature would support this

Basically just an additional option in the `orderBy` method.

3. Give some API proposal, how the feature should work.

Something like this would be great:

```js
knex('users').orderBy('name', 'desc', 'last');
```
Outputs:
```sql
order by (`name` is null) asc , `name` desc
```

```
knex('users').orderBy([ { column: 'name', order: 'desc', nulls: 'last' }, 'id' ]);
```

Outputs:
```sql
order by (`name` is null) asc, `name` desc, `id` asc
```

Of course, the reason for not using NULLS LAST here is because it would not work in all of the knex-supported databases. I believe this calculated criteria approach would, though.

[1]: https://github.com/knex/knex/issues/282
[2]: https://vincit.github.io/objection.js/
[...]

PR #4720, opened 2021-10-10
Support for 'is null' in 'order by' (#3667)

OlivierCavadenti:
- Attempt to resolve issue #3667
- add support to order nulls in order by with 'first' and 'last' option.
- 'first' add 'is not null' in order by clause (=> put nulls in first position)
- 'last' add 'is null' in order by clause (=> put nulls in last position)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1485,7 +1485,7 @@ export declare namespace Knex {
       TResult2 = AggregationQueryResult<TResult, {[x in TAlias]: number}>
       >(alias: TAlias,
         orderBy: TKey | TKey[] | { columnName: TKey, order?: 'asc' | 'desc', nulls?: 'first' | 'last' },
-        partitionBy?: TKey | TKey[] | { columnName: TKey, order?: 'asc' | 'desc', nulls?: 'first' | 'last' }):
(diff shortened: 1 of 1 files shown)


kibertoad:
TS types need to be updated, and we also need documentation updated for this.

kibertoad:
@OlivierCavadenti It appears that MSSQL doesn't support that, and one would need to implement hacks to get that working: https://stackoverflow.com/questions/5886857/sql-server-order-by-date-and-nulls-last

If you feel like implementing custom solution to get that working with MSSQL, that would be great, but if that's too much work, throwing an error when trying to use that param on MSSQL is also sufficient.

OlivierCavadenti:
> @OlivierCavadenti It appears that MSSQL doesn't support that, and one would need to implement hacks to get that working: https://stackoverflow.com/questions/5886857/sql-server-order-by-date-and-nulls-last
> 
> If you feel like implementing custom solution to get that working with MSSQL, that would be great, but if that's too much work, throwing an error when trying to use that param on MSSQL is also sufficient.

I will see what I can do to get that working with MSSQL. Thanks for the link.

kibertoad:
Thanks a lot! Can you also add documentation for this?

PR #4738, opened 2021-10-14
Handles new socket error introduced in Tedious 13.1.0

jamescammarano:
In Tedious 13.1.0 they introduced a new socket error that was not being caught by Knex in the acquireRawConnection function.

I changed the `.on('error')` handler to the beginning of the acuireRawConnection function rather than immediately rejecting because Tedious uses event emitters for their errors and the error was never hitting the `.on('error').`


kibertoad:
@jaycammarano Could you please add test for this? Also it's not possible to handle it on mssql dialect level?

Issue #4719, opened 2021-10-10
Inconsistent listing of migrations in the Knex.Migrator API


marceliwac:
# Environment

Knex version: 0.21.21
Database + version: (irrelevant)
OS: (irrelevant)

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do.
When using the `Knex.Migrator` API to run migrations programmatically, listing the migrations returns the completed and pending migrations in two different formats. This could be pinpointed to the `migration-list-resolver.js` which defines this behaviour in `listCompleted` function, i.e.:

```js
// Lists all migrations that have been completed for the current db, as an
// array.
async function listCompleted(tableName, schemaName, trxOrKnex) {
  await ensureTable(tableName, schemaName, trxOrKnex);
  const completedMigrations = await trxOrKnex
    .from(getTableName(tableName, schemaName))
    .orderBy('id')
    .select('name');

  return completedMigrations.map((migration) => {
    return migration.name;
  });
}
```

Running the actual `list` function of the Migrator API with one completed migration (i.e. `01-completed.js`) and one pending one (`02-pending.js`) results in inconsistent listing of the functions that looks something like:
```js
 [
   [ 'completed.js' ],
   [
     {
       file: 'pending.js',
       directory: './migrations'
     }
   ]
]
```
as opposed to:
```js
 [
   [
     {
       file: 'completed.js',
       directory: './migrations'
     } 
   ],
   [
     {
       file: 'pending.js',
       directory: './migrations'
     }
   ]
]
```

2. Error message
n/a

3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real
   database connection to MySQL or PostgreSQL, then single file example which initializes
   needed data and demonstrates the problem.
n/a

4. Potential solution *
Seeing as the only places where the `migration-list-resolver.js/listCompleted` is used is the Migrator API (Migrator.js), the fix should be relatively easy to implement though might be a breaking change (i.e. output no longer just presented as an array of string but array of objects instead).

kibertoad:
@marceliwac We are planning to release another semver major this year, so this might make the cut. Would you consider sending PR for this?

kibertoad:
Released in 1.0.1.

PR #4745, opened 2021-10-16
Refactor Migrator code and API to return list of migrations as objects consistently.

marceliwac:
Fix to issue #4719.


kibertoad:
Does this affect any types from index.d.ts?

PR #4755, opened 2021-10-20
Support SELECT .. FOR NO KEY UPDATE / KEY SHARE row level locking clauses in Postgres

### LINKED ISSUE #4754: SELECT .. FOR NO KEY UPDATE locking clause in Postgres
# Environment

Knex version: all
Database + version: Postgres 9.3 and above

Knex currently supports the `FOR UPDATE` and `FOR SHARE` locking clauses on select queries. Postgres has two additional lock strengths, `FOR NO KEY UPDATE` and `FOR KEY SHARE`, see https://www.postgresql.org/docs/9.6/sql-select.html#SQL-FOR-UPDATE-SHARE. Could support be added for those as e.g. `.forNoKeyUpdate()` and `forKeyShare()`?

Not sure if other DBs have an equivalent feature, or if the support would be Postgres-specific. I can contribute a PR for Postgres if needed.

domkck:
In addition to `SELECT .. FOR UPDATE` and `SELECT .. FOR SHARE`, Postgres supports `FOR NO KEY UPDATE` and `FOR KEY SHARE` row level locking clauses https://www.postgresql.org/docs/14/explicit-locking.html#LOCKING-ROWS.

This adds support as `.forNoKeyUpdate()` and `forKeyShare()`.

I don't know if there's an equivalent feature in MySQL and MSSQL - how should that be handled? 

Fixes #4754

### DIFF
--- a/lib/dialects/postgres/query/pg-querycompiler.js
+++ b/lib/dialects/postgres/query/pg-querycompiler.js
@@ -147,37 +147,26 @@ class QueryCompiler_PG extends QueryCompiler {
     return sql.join(', ');
   }
 
-  forUpdate() {
+  _lockingClause(lockMode) {
     const tables = this.single.lockTables || [];
 
-    return (
-      'for update' + (tables.length ? ' of ' + this._tableNames(tables) : '')
-    );
+    return lockMode + (tables.length ? ' of ' + this._tableNames(tables) : '');
+  }
+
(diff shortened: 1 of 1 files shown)


kibertoad:
Looks good! Could you please also submit documentation PR to https://github.com/knex/documentation?

PR #4767, opened 2021-10-22
feat(cockroachdb): add support upsert

intech:
Add support `upsert()` for CockroachDB
Docs: [upsert](https://www.cockroachlabs.com/docs/stable/upsert.html#parameters)
Related: https://github.com/knex/knex/issues/2002


kibertoad:
@intech Just add it in text description that it only works with crdb.

PR #4768, opened 2021-10-23
Implement partial index support

### LINKED ISSUE #657: Conditional index in a migration
Is it possible to specify a condition index when creating table in a migration? Something that would result in the following index:

```sql
create index category_score_id on products(category, score desc, id desc) where active;
```

Thank you.

OlivierCavadenti:
- Rebase of https://github.com/knex/knex/pull/2401 (i lost the git history unfortunaly) to finish the feature implemented and unfinished (I cant commit to the original repo).
- Attempt to Fixes #657

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2026,10 +2026,7 @@ export declare namespace Knex {
   type storageEngineIndexType = 'hash' | 'btree';
 
   interface ColumnBuilder {
-    index(
-      indexName?: string,
-      options?: Readonly<{indexType?: string, storageEngineIndexType?: storageEngineIndexType, predicate?: QueryBuilder}>
-    ): ColumnBuilder;
+    index(indexName?: string): ColumnBuilder;
     primary(options?: Readonly<{constraintName?: string, deferrable?: deferrableType}>): ColumnBuilder;
     /** @deprecated */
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -1975,11 +1975,11 @@ export declare namespace Knex {
     index(
       columnNames: string | readonly (string | Raw)[],
       indexName?: string,
-      indexType?: string
+      options?: Readonly<{indexType?: string, storageEngineIndexType?: storageEngineIndexType, predicate?: QueryBuilder}>
```
B4nan, line 1978: fyi this change was breaking for TS users as the third parameter no longer accepts plain string (the code still supports it, just a typings issue)

OlivierCavadenti: Thanks for the report.
You'r right, I can make an hotfix today (at the end of day) if necessary. @kibertoad

kibertoad: @OlivierCavadenti Would be great, could you also backport it to 0.95? I'd make a release today then.

kibertoad: @B4nan @OlivierCavadenti Fix was published in 0.95.13

@OlivierCavadenti, thank you for the rapid fix!

PR #4779, opened 2021-10-29
Support whereLike and whereILike

OlivierCavadenti:
- Finish https://github.com/knex/knex/pull/4457
- I made some mutualization to avoid duplicate code.
- test classes can be very big so I create new where.spec.js file and I move all 'where' tests in this file.
- Closes https://github.com/knex/knex/issues/2265


test/integration2/query/select/where.spec.js
```diff
@@ -0,0 +1,903 @@
...
+                    balance: 0,
+                    about: 'Lorem ipsum Dolore labore incididunt enim.',
+                    created_at: TEST_TIMESTAMP,
+                    updated_at: TEST_TIMESTAMP,
+                    phone: null,
+                  },
+                ]
+              );
+              tester(
+                'pgnative',
```
kibertoad, line 316: If I remember correctly, you can pass an array as first param, so that you can cover several identical dbs at once.

would be nice to also add cockroach if it supports this

OlivierCavadenti: I rework all tests in where.spec.js to avoid duplicates. It's more concise now.

kibertoad: Great job!

OlivierCavadenti: I fix all tests. Maybe I can disable "is not able to run two migrations in parallel when transactions are disabled", make a TODO and an issue to fix it later ? At this time, it make a lot of noise in all tests.

kibertoad: yes, please! there might be existing issue for it

kibertoad (approved):

test/integration2/query/select/where.spec.js
OlivierCavadenti: I add the issue here : https://github.com/knex/knex/issues/4789 but it seems I made a mistake when I rebase/merge... I have duplicate commits xd.

OlivierCavadenti: done

PR #4780, opened 2021-10-29
Fix overzealous warning on use of whereNot with "in" or "between"

willheslam:
I ran into this with a `whereNot('id', 'in')` clause producing a warning despite it only intending to occur when 'in' is an operator, not a value.

Please let me know if the test format isn't right, I'd be happy to change it!

### DIFF
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -1050,6 +1050,36 @@ describe('QueryBuilder', () => {
     }
   });
 
+  it('where not should not throw warning when used with "in" or "between" as equality', function () {
+    testquery(
+      clientsWithCustomLoggerForTestWarnings.pg
+        .queryBuilder()
+        .select('*')
+        .from('users')
+        .whereNot('id', 'in'),
+      {
+        mysql: "select * from `users` where not `id` = 'in'",
+        pg: 'select * from "users" where not "id" = \'in\'',
+        'pg-redshift': 'select * from "users" where not "id" = \'in\'',
+        mssql: "select * from [users] where not [id] = 'in'",
+      }
+    );
+
+    testquery(
+      clientsWithCustomLoggerForTestWarnings.pg
+        .queryBuilder()
+        .select('*')
+        .from('users')
+        .whereNot('id', 'between'),
+      {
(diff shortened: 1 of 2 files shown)


kibertoad:
@OlivierCavadenti Could you take a look at this fix?

test/unit/query/builder.js
```diff
@@ -1050,6 +1050,36 @@ describe('QueryBuilder', () => {
     }
   });
 
+  it('where not should not throw warning when used with "in" or "between" as equality', function () {
+    testquery(
+      clientsWithCustomLoggerForTestWarnings.pg
```
OlivierCavadenti, line 1055: Here you select only one client (PG) but test query with multiple db providers.
You can replace clientsWithCustomLoggerForTestWarnings.pg.queryBuilder() with only 'qb()'

willheslam: This test was largely copied from the test above:
```

  it('where not should throw warning when used with "in" or "between"', function () {
    try {
      clientsWithCustomLoggerForTestWarnings.pg
      ...
```

so I kept it the same for consistency, but I appreciate the assertion checks against all of them, which is itself inconsistent. I'm unfamiliar with the test suite so I wasn't sure exactly what the assertion should look like.

Regarding `qb()` - is this safe?
Looking at the definition of `clientsWithCustomLoggerForTestWarnings` it uses this custom logger config
```
  log: {
    warn: function (message) {
      throw new Error(message);
    },
  },
  ```
  which the test implicitly relies on - expecting the error not to be thrown.
  
If I use the results of the `qb` function, will the error throwing behaviour of the logger be present?

Having thought about this, the implicit expectation that the test *could* fail but shouldn't makes me feel like the body of this test might actually better belong in the test for "where not should throw warning when used with "in" or "between"" - i.e. as a counter example.

What do you think? That would make the implicit dependence on a failing logger more obvious, as it's paired with a test that *does* expect the logger to throw.

OlivierCavadenti: Yeah it's seems your are right. At the first read, it's just strange for me to use a querybuilder from a specific client dans check with all clients, but any that works for me.

OlivierCavadenti (approved):

PR #4785, opened 2021-10-30
Allowing for alternate naming of Pg CLI version reporting

adamjgrant:
My entire directus.io install fails because of this line. It's trying to read

```
psql --version
psql (PostgreSQL) 12.5 (Ubuntu 12.5-1.pgdg20.04+1)
```

### DIFF
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -127,7 +127,7 @@ class Client_PG extends Client {
   }
 
   _parseVersion(versionString) {
-    return /^PostgreSQL (.*?)( |$)/.exec(versionString)[1];
+    return /^(PostgreSQL|psql) (.*?)( |$)/.exec(versionString)[1];
   }
 
   // Position the bindings for the query. The escape sequence for question mark


kibertoad:
Can you add unit test for this?

adamjgrant:
Hi, I'm still here. Sorry, need to work out a quirk with my setup so I can get Knex running, write a test and ensure it properly fails.

Also the regex I provided here is terrible. I'll update that too so it's more flexible. 

[Backlink](https://github.com/directus/directus/discussions/9268#discussioncomment-1583930)

lib/dialects/postgres/index.js
```diff
@@ -127,7 +127,7 @@ class Client_PG extends Client {
   }
 
   _parseVersion(versionString) {
-    return /^PostgreSQL (.*?)( |$)/.exec(versionString)[1];
```
adamjgrant, line 130: Note to self: it's worse than I thought: `versionString` comes back as

`CockroachDB CCL v21.2.0-beta.4-20-g3061813685 (x86_64-unknown-linux-gnu, built 2021/10/15 07:02:45, go1.16.6)`

So the actual result of `pg --version` would still not match but we're not even getting that in this function.

Reading out `resp.rows` where this string comes from, I get 

~~~
[{"version":"CockroachDB CCL v21.2.0-beta.4-20-g3061813685 (x86_64-unknown-linux-gnu, built 2021/10/15 07:02:45, go1.16.6)"}]
~~~

kibertoad:
@adamjgrant You should be using CockroachDB dialect then, it already should work correctly. Try latest version.

adamjgrant:
@kibertoad Sorry, just saw your comment. I must be doing something wrong from the directus end (client) in that case. Will check...

adamjgrant:
Closing this for now as I'm not in the right file and am not describing the issue correctly anyway.

kibertoad:
@adamjgrant CockroachDB dialect is brand new, you'd need to update knex to the latest version if you want to use it. Or that is not intentional and it should be PostgreSQL?

PR #4800, opened 2021-11-03
Support Joins and Using syntax for PostgreSQL Delete Statement (#4591)

### LINKED ISSUE #4591: pg: incorrect "delete + join" query since 0.95.8
# Environment

Knex version: 0.95.8
Database + version: postgresql 12.7
OS: linux/debian 11

# Bug

I get an incorrect `delete` postgres query, containing the table name in the `delete` statement:
```sql
delete "my_table" from "my_table" left join "users" on "users"."user_id" = "my_table"."user_id" where "my_table"."criteria" < 10
```
It should produce this code instead:
```sql
delete from "my_table" left join "users" on "users"."user_id" = "my_table"."user_id" where "my_table"."criteria" < 10
```
2. Error message
error from postgres:
```
syntax error at or near ""my_table""
```

```javascript
	client: 'pg',
	connection: { /* redacted */ }
});

const query = knex
	.select('xyz')
	.from('my_table')
	.leftJoin('users', 'users.user_id', 'my_table.user_id')
        .where('my_table.criteria', '<', 10)
	.delete();

console.log(query.toString());
```

this code produces the following query in 0.95.8:
```sql
delete "my_table" from "my_table" left join "users" on "users"."user_id" = "my_table"."user_id" where "my_table"."criteria" < 10
```

and in 0.95.7:
```sql
delete from "my_table" where "my_table"."criteria" < 10
```
The 0.95.7 result is not correct either, but at least it generates a valid PostgreSQL statement.

I'm not sure the way knex is used in this example is correct, but I found an actual use of knex like this one in my codebase.

I hope this information is relevant, and if not, sorry for the noise.

Best Regards

OlivierCavadenti:
- PostgreSQL dont support Join in Delete (https://www.postgresqltutorial.com/postgresql-delete-join/) so I implements the support of using to do the same thing in two king : implicit and explicit methods.

So implicit method, with usual join methods work: 

```javascript
 qb()
    .del()
    .from('users')
    .join('photos', 'photos.id', 'users.id')
    .where({ 'user.email': 'mock@example.com' })
```
and will generate the query with 'using' and 'where' to work in PostgreSQL.
It will general the same query that explicit form:

```javascript
 qb()
  .del()
  .from('users')
  .using('photos')
  .whereRaw('photos.id = users.id')
  .where({ 'user.email': 'mock@example.com'})
```
- ~~I dont want to go in intrusive and complex implementation (no new kind of joins...) so I just add small using method in querybuilder and make all the specific logic in PG files.~~ Finally I implements both implicit joins conversion and explicit form.
- ~~I choose to make a specific implement of Delete that dont allow joins in Delete and throw an error when user try to make the joins with Delete statement => "Joins with delete query isn't valid in PostgreSQL, use 'using' with 'where' functions instead."~~ Specific implement of Delete for PostgreSQL that convert joins in 'using' + 'wheres'
- Fixes #4591

### DIFF
--- a/lib/dialects/postgres/query/pg-querycompiler.js
+++ b/lib/dialects/postgres/query/pg-querycompiler.js
@@ -105,6 +105,7 @@ class QueryCompiler_PG extends QueryCompiler {
               column: clause.column,
               operator: '=',
               value: clause.value,
+              asColumn: true,
             })
           );
         }
@@ -117,13 +118,10 @@ class QueryCompiler_PG extends QueryCompiler {
       }
     }
 
-    // When using the clause 'using' we delete the "from" table name (like joins)
-    const deleteSelector = using ? tableName + ' ' : '';
+    // With 'using' syntax, no tablename between DELETE and FROM.
     const sql =
       withSQL +
-      `delete ${deleteSelector}from ${
-        this.single.only ? 'only ' : ''
-      }${tableName}` +
+      `delete from ${this.single.only ? 'only ' : ''}${tableName}` +
       (using ? ` ${using}` : '') +
       (wheres ? ` ${wheres}` : '');
     const { returning } = this.single;
--- a/lib/dialects/postgres/query/pg-querycompiler.js
+++ b/lib/dialects/postgres/query/pg-querycompiler.js
@@ -57,9 +57,44 @@ class QueryCompiler_PG extends QueryCompiler {
     };
   }
 
-  // Compiles an `update` query, allowing for a return value.
+  using() {
+    const usingTables = this.single.using;
+    if (!usingTables) return;
+    let sql = 'using ';
+    if (Array.isArray(usingTables)) {
+      sql += usingTables
+        .map((table) => {
+          return this.formatter.wrap(table);
+        })
+        .join(',');
+    } else {
+      sql += this.formatter.wrap(usingTables);
+    }
+    return sql;
+  }
+
+  // Compiles an `delete` query, allowing for a return value.
   del() {
-    const sql = super.del(...arguments);
+    // Make sure tableName is processed by the formatter first.
+    const { tableName } = this;
+    const withSQL = this.with();
+    const wheres = this.where();
+    // Joins in PG with delete it's not valid, we throw an error and ask to use 'using'
+    if (this.join().length > 0) {
+      throw new Error(
+        "Joins with delete query isn't valid in PostgreSQL, use 'using' with 'where' functions instead."
+      );
+    }
+    const using = this.using();
+    // When using the clause 'using' we delete the "from" table name (like joins)
+    const deleteSelector = using ? tableName + ' ' : '';
+    const sql =
+      withSQL +
+      `delete ${deleteSelector}from ${
+        this.single.only ? 'only ' : ''
+      }${tableName}` +
+      (using ? ` ${using}` : '') +
+      (wheres ? ` ${wheres}` : '');
     const { returning } = this.single;
     return {
       sql: sql + this._returning(returning),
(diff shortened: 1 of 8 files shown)


lib/dialects/postgres/query/pg-querycompiler.js
```diff
@@ -57,9 +57,44 @@ class QueryCompiler_PG extends QueryCompiler {
...
+
+  // Compiles an `delete` query, allowing for a return value.
   del() {
-    const sql = super.del(...arguments);
+    // Make sure tableName is processed by the formatter first.
+    const { tableName } = this;
+    const withSQL = this.with();
+    const wheres = this.where();
+    // Joins in PG with delete it's not valid, we throw an error and ask to use 'using'
+    if (this.join().length > 0) {
```
kibertoad, line 83: converting joins to using transparently is not feasible?

lib/query/querybuilder.js
```diff
@@ -269,6 +269,11 @@ class Builder extends EventEmitter {
     return this;
   }
 
+  using(tables) {
```
kibertoad, line 272: what if someone tries using it outside of pg?

OlivierCavadenti: Nothing. I dont see how I can restrict function access in query builder. Or maybe I put a warning ?

kibertoad: I would probably make default implementation throw an error, and override it for pg.

lib/dialects/postgres/query/pg-querycompiler.js
OlivierCavadenti: You mean use the joins logic by add 'using' joins types with the 'using()' syntax ?
I think about that but with that implementation I cant throw an error here.
Also we have a join with using syntax inside and I quite afraid about some misunderstand.

lib/query/querybuilder.js
OlivierCavadenti: Oh, yes.... I will make that.

t1a2l:
@elhigu - are you going to review this?

t1a2l:
ok i just added the fix to local project.

lib/dialects/postgres/query/pg-querycompiler.js
kibertoad: I'm still on the fence about this, TBH. Considering that many multi-db systems use knex specifically to abstract away db differences, transforming joins to `using` here would allow them to avoid having some conditional logic. Is it possible to always transform `join` into `using`, or you see certain case where that would fail and user would be forced to use `using` explicitly anyway?

OlivierCavadenti: I see the argument. I start on 'using' function because of an argument in the original PR about avoid silent converting operations. But in fact, we always do "silent" operations in Knex, especially with oracle, when converting limit/offset functions...
I will rework it.

kibertoad: I wonder if we should also keep `using` syntax for those who want to be explicit.

OlivierCavadenti: I push with implicit joins to 'using' + 'where's syntax and I also keep the actual using syntax too.

kibertoad:
Would also be helpful to update existing `Delete with join` tests, so that they would no longer skip PostgreSQL. They also currently miss assertions that only correct data is deleted, which would be great to add too.

OlivierCavadenti:
> Would also be helpful to update existing `Delete with join` tests, so that they would no longer skip PostgreSQL. They also currently miss assertions that only correct data is deleted, which would be great to add too.

I push with more integration tests + one unit test (mixed joins + using notation).
By the way I  made fixes on querycompiler ('where' value as column value for ex).
I see that my example in this PR + unit test are not good (cannot use id columns with 'where' use 'whereRaw' instead).

kibertoad (approved):
Outstanding work on this one!

Issue #4688, opened 2021-09-21
Remove extension from migration names


glensc:
So, having your migrations in TypeScript, running them via `ts-node` locally works fine, but when deploying production the migrations are compiled to `.js` files via `tsc`.

So migrations recorded in the production database have `.js` extension, and when such database dump is attempted to be used in the development, you'll have name conflict as seems like migrations not applied.

Workaround is this:
```sql
update knex_migrations set name=replace(name, '.js', '.ts');
```

But I was thinking, as the project is heading towards 1.0, perhaps make the breaking change and rename the extension from migration names?

elhigu:
I agree that not having migration file extension in the database would be a good thing.

glensc:
implementation wise, the code could fetch name from database, strip extension, then compare. i.e not do `select * from knex_migrations where name='123_111.js'` but rather do `select` without `where`.

kibertoad:
@glensc Would you consider sending a PR for this?

elhigu:
I thinkg that implementation sounds also good. It should be backwards compatible. I'm not sure about the case where the same directory has migration_name.ts and migration_name.js available and both extensions enabled in configuration. 

Should that throw an error or just silently select either one of those to run. In any case it is probably good to have complete file name in the database and just ignore extension in comparison 👍 That will make distinction in those cases which file actually was applied to DB if there are many with the same base name.

mercmobily:
Closing this in favour of #2756 which covers the same issue

PR #4802, opened 2021-11-04
migration without file extension

### LINKED ISSUE #4028: Ignore file extension for migrations
# Environment

Knex version: 0.21.1
Database + version: MySQL
OS: Linux

Developing locally on TypeScript but running tests with compiled JavaScript on AWS Lambda
Because for local development and integration tests we use same DB.
It causes that when we run migrations locally it will add `migration.ts` as name in table but when it's run on Lambda it won't find any `.ts` files as they'll be compiled to `.js`

So we get this error
```
Error: The migration directory is corrupt, the following files are missing: migration.ts
```

Store in DB only migration filename without extension. That is `migration` and not `migration.ts`.
Also imagine case when we're working in JS but later decide to switch to TypeScript then with current behavior we wouldn't be able to change those migration files. So I would say extension doesn't really matter it's the same migration and I don't think anyone would have `migration.ts` and `migration.js` which does different things...

Fydon:
This fixes #4028.

This is a breaking change as it currently expects all migrations to not have a file extension. It may be possible to have it instead check if the migration ends with one of the supported extensions and remove it before comparing.

There is still a failing test at [test\integration2\migrate\migration-integration.spec.js:827](https://github.com/Fydon/knex/blob/feature/migrationWithoutFileExtension/test/integration2/migrate/migration-integration.spec.js#L827). I'm not sure what the cause of the change is.

I've created test\unit\migrations\migrate\sources\fs-migrations.js to specifically test the output of the two, although this is being test with test\integration2\migrate\migration-integration.spec.js. I could add more tests.

### DIFF
--- a/lib/migrations/migrate/Migrator.js
+++ b/lib/migrations/migrate/Migrator.js
@@ -52,7 +52,7 @@ class Migrator {
       this.knex.client.logger
     );
     this._activeMigration = {
-      fileName: null,
+      name: null,
     };
   }
 
@@ -121,7 +121,13 @@ class Migrator {
     );
 
     let migrationToRun;
-    const name = this.config.name;
+    const name =
+      this.config.name != null
+        ? this.config.migrationSource.getMigrationName({
+            file: this.config.name,
+            directory: this.config.directory,
+          })
+        : null;
     if (name) {
       if (!completed.includes(name)) {
         migrationToRun = newMigrations.find((migration) => {
@@ -400,9 +406,9 @@ class Migrator {
             "lock manually by running 'knex migrate:unlock'"
         );
       } else {
-        if (this._activeMigration.fileName) {
+        if (this._activeMigration.name) {
           this.knex.client.logger.warn(
-            `migration file "${this._activeMigration.fileName}" failed`
+            `migration name "${this._activeMigration.name}" failed`
           );
         }
         this.knex.client.logger.warn(
@@ -492,7 +498,7 @@ class Migrator {
     const log = [];
     migrations.forEach((migration) => {
       const name = this.config.migrationSource.getMigrationName(migration);
-      this._activeMigration.fileName = name;
+      this._activeMigration.name = name;
       const migrationContent =
         this.config.migrationSource.getMigration(migration);
 
@@ -500,7 +506,7 @@ class Migrator {
       current = current
         .then(async () => await migrationContent) //maybe promise
         .then((migrationContent) => {
-          this._activeMigration.fileName = name;
+          this._activeMigration.name = name;
           if (
             !trx &&
             this._useTransaction(migrationContent, disableTransactions)
@@ -564,7 +570,7 @@ function validateMigrationList(migrationSource, migrations) {
   const diff = getMissingMigrations(migrationSource, completed, all);
   if (!isEmpty(diff)) {
     throw new Error(
-      `The migration directory is corrupt, the following files are missing: ${diff.join(
+      `The migration directory is corrupt, the following migrations are missing: ${diff.join(
         ', '
       )}`
     );
--- a/lib/migrations/migrate/sources/fs-migrations.js
+++ b/lib/migrations/migrate/sources/fs-migrations.js
@@ -72,7 +72,7 @@ class FsMigrations {
   }
 
   getMigrationName(migration) {
-    return migration.file;
+    return path.parse(migration.file).name;
   }
 
   getMigration(migration) {
--- a/lib/migrations/migrate/sources/fs-migrations.js
+++ b/lib/migrations/migrate/sources/fs-migrations.js
@@ -59,14 +59,12 @@ class FsMigrations {
       // return the migrations fully qualified
       if (this.sortDirsSeparately) {
         return filterMigrations(
-          this,
           migrations,
           loadExtensions || this.loadExtensions
         );
       }
 
       return filterMigrations(
-        this,
         sortBy(migrations, 'file'),
         loadExtensions || this.loadExtensions
       );
@@ -85,10 +83,9 @@ class FsMigrations {
   }
 }
 
-function filterMigrations(migrationSource, migrations, loadExtensions) {
+function filterMigrations(migrations, loadExtensions) {
   return migrations.filter((migration) => {
-    const migrationName = migrationSource.getMigrationName(migration);
-    const extension = path.extname(migrationName);
+    const extension = path.extname(migration.file);
     return loadExtensions.includes(extension);
   });
 }
--- a/test/cli/migrate.spec.js
+++ b/test/cli/migrate.spec.js
@@ -53,7 +53,7 @@ describe('migrate:latest', () => {
       });
     });
     const row = await getPromise;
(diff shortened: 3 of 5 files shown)


Fydon:
Fixed failing linting tests caused by this change. Sorry I should have run that before.

When I run `npm run lint:everything` I see `The 'no-redundant-jsdoc' rule threw an error in 'knex/types/index.d.ts'` which is to do with the @deprecated JSDocs in [knex/types/index.d.ts](https://github.com/knex/knex/blob/master/types/index.d.ts).

OlivierCavadenti:
> Fixed failing linting tests caused by this change. Sorry I should have run that before.
> 
> When I run `npm run lint:everything` I see `The 'no-redundant-jsdoc' rule threw an error in 'knex/types/index.d.ts': Error: Unexpected tag kind: JSDocDeprecatedTag` which is to do with the @deprecated JSDocs in [knex/types/index.d.ts](https://github.com/knex/knex/blob/master/types/index.d.ts). Although that test here completed without errors.
> 
> Any assistance with the failing test in migration-integration.spec.js would be appreciated.

For Error: Unexpected tag kind: JSDocDeprecatedTag it seems that is not blocking error.
On the other point, by the way I unfortunaly cant help you one this point... @kibertoad  ?

OlivierCavadenti:
Also, I assume this will fix also https://github.com/knex/knex/issues/4688  ?

kibertoad:
@Fydon 

> There is still a failing test at test\integration2\migrate\migration-integration.spec.js:827. I'm not sure what the cause of the change is.

To me it seems to indicate that partial execution is broken after one of the changes, and the down script is not executed at all. Do you need help debugging this issue?

Fydon:
Thank you. I see now that I can run a single test. Stepping through it, I realised that it was trying to get a migration name from a file name when executing the down migration. I think that they should all pass now.

Fydon:
> Also, I assume this will fix also #4688 ?

@OlivierCavadenti Yes that should also be closed with this change. It could also be closed as a duplicate of the other.

kibertoad:
Btw, since this is a breaking change, which migration path for existing systems do you see?

Fydon:
Potential ideas to not make this breaking:

1. We could try to detect if the migration name ends with an extension and if so remove the extension. Unfortunately this means that this code to adjust the migration stays, which is not ideal.
2. Add a new column to the migrations table to identify its "version", defaulting it to some value "X". If the "version" is value "X" then remove extension. This would allow for future changes where the structure of the migration table or the format of its data changes. I don't know the code well enough to know if this is possible though.

Alternatively we keep it as is and request that people update the migration names in their database prior to using the latest version. There may be better ways to manage this.

Fydon:
Sorry I also meant to say that I created test\unit\migrations\migrate\sources\fs-migrations.js but I wasn't sure where to add it in the test suite. You may be happy with the tests that you already have to test this class. Should I delete this test class or where should I add it? It is currently two unit tests, which I took from [migration-list-resolver.js](https://github.com/knex/knex/blob/master/test/unit/migrations/migrate/migration-list-resolver.js#L12-L95).

kibertoad:
@Fydon I would say that anything that requires users to fix their migrations manually is no-go. At worst we should provide them with a knex command which could do the conversion, at best it should happen automatically when migrations are executed. This compatibility bridge could be removed few major versions later.

Fydon:
> Sorry I also meant to say that I created test\unit\migrations\migrate\sources\fs-migrations.js but I wasn't sure where to add it in the test suite. You may be happy with the tests that you already have to test this class. Should I delete this test class or where should I add it? It is currently two unit tests, which I took from [migration-list-resolver.js](https://github.com/knex/knex/blob/master/test/unit/migrations/migrate/migration-list-resolver.js#L12-L95).

Could you please also confirm what you would prefer me to do regarding the above?

Fydon:
Hi @kibertoad and @OlivierCavadenti. Thank you for your work on this project. Please look at my previous two comments when you have a chance.

kibertoad:
@OlivierCavadenti Could you please take a look?

lib/migrations/migrate/Migrator.js
```diff
@@ -222,7 +228,13 @@ class Migrator {
...
 
         let migrationToRun;
-        const name = this.config.name;
+        const name =
+          this.config.name != null
+            ? this.config.migrationSource.getMigrationName({
+                file: this.config.name,
+                directory: this.config.directory,
+              })
+            : null;
```
OlivierCavadenti, line 237: duplicate code with above

OlivierCavadenti (changes requested):
Thanks for the PR.
By the way I read the code in separate branch but it's seems a bit complicated for what we want to do... It will be used for other use cases ? 
 I am quite more in the "be capable of still handling extensions when they are present." solution if it's possible.

Fydon:
Thank you for the review. Yes my hope is that the versioning of each migration will be used for any further alterations to the internal tables. I could not think of a simpler solution that did not break all current databases or require manual steps to patch existing databases.

Are you requesting that I perform the updated every time Knex is initialised? I can do that in another branch. I use this for a small app, but I wonder if this will cause issues with database load if someone has an app with many instances and it crashes on load after Knex is initialised.

lib/migrations/migrate/Migrator.js
Fydon: Created a function to contain the duplicate code

Fydon:
Sorry for the delay. I got COVID.

I've created [another branch](https://github.com/Fydon/knex/pull/4/files) which I think performs what you requested. Please check if this is better.

I'm struggling to work out how best to test calling `ensureTable` in `lib\migrations\migrate\table-creator.js`. I assume like `test\unit\migrations\migrate\Migrator.js` I would use sqlite, but how would I setup the database so that migrations table is setup to have specific migration names?

Fydon:
I've managed to create a test in [that branch](https://github.com/Fydon/knex/pull/4/files). Using breakpoints during the test, the code appears to be working as expected. However the test is showing the table as though it is not being updated. Any ideas?

Fydon:
I found the issue. Are you happy with the implementation? If not, how can I improve it?

How should I include `test\unit\migrations\migrate\table-creator.js` in the suite of tests?

Should I include `test\unit\migrations\migrate\sources\fs-migrations.js` in the suite of tests or delete it?

Fydon:
@kibertoad @OlivierCavadenti I'm sure you both are really busy. Any time you could spare to move this along is appreciated. Please see my last comment.

Fydon:
> I found the issue. Are you happy with [the implementation](https://github.com/Fydon/knex/pull/4/files)? If not, how can I improve it?
> 
> Instead of including a new database, I could try using `test\test.sqlite3` and setting it up with the required data.
> 
> How should I include `test\unit\migrations\migrate\table-creator.js` in the suite of tests?
> 
> Should I include `test\unit\migrations\migrate\sources\fs-migrations.js` in the suite of tests or delete it? If I should, where should I include it?

@OlivierCavadenti @kibertoad When you get a chance, please see my comment quoted here. If you are happy with the implementation, I can merge it into this pull request causing it to be ready for whenever you want to include it.

PR #4803, opened 2021-11-04
Add camel case to timestamps method

OlivierCavadenti:
- I remake a PR to finish previous work of @Vilkas67 (I make bad rebase on https://github.com/knex/knex/pull/4553, sry for inconvenance).
- Doc : https://github.com/knex/documentation/pull/358

### DIFF
--- a/lib/schema/tablebuilder.js
+++ b/lib/schema/tablebuilder.js
@@ -51,14 +51,9 @@ class TableBuilder {
 
   // The "timestamps" call is really just sets the `created_at` and `updated_at` columns.
 
-  timestamps() {
-    let useTimestamps, defaultToNow, useCamelCase;
+  timestamps(useTimestamps, defaultToNow, useCamelCase) {
     if (isObject(arguments[0])) {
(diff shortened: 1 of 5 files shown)


lib/schema/tablebuilder.js
```diff
@@ -50,16 +50,25 @@ class TableBuilder {
...
 
   // The "timestamps" call is really just sets the `created_at` and `updated_at` columns.
+
   timestamps() {
-    const method = arguments[0] === true ? 'timestamp' : 'datetime';
-    const createdAt = this[method]('created_at');
-    const updatedAt = this[method]('updated_at');
-    if (arguments[1] === true) {
+    let useTimestamps, defaultToNow, useCamelCase;
+    if (isObject(arguments[0])) {
```
kibertoad, line 56: can you replace arguments with proper function parameters?

kibertoad: there is still reference to arguments. can we use useTimestamps instead?

kibertoad (approved):

PR #4804, opened 2021-11-05
initial commit Japanese translation of readme

yukitomeow:
Hello. There is no Japanese translation on Knex. So I made Japanese version of README.


OlivierCavadenti:
@kibertoad maybe we can plan other translation of README ?
If yes, I can do in French and we can add links in README.md

njivaolafpro (approved):
lgtm

PR #4807, opened 2021-11-05
0.95

bipinkumar01:
this code change made on queryCompiler.js file inside whereIn methods to handle max number of expression in a list more 1000 in query while executing query which is more 1000 expression in Oracle DB.
Oracle DB has restrictions for more than 1000 expression in query.

So in order to handle it need to implement after where in ........ OR condition in .......


OlivierCavadenti:
Duplicate with https://github.com/knex/knex/pull/4808

PR #4808, opened 2021-11-05
Fix 4783 and 4790

Himanshu-Singhal:
Fix for the whereIn condition to work with oracledb 11g constraint (ORA 01795) when the number of records is more than 1000.

### DIFF
--- a/lib/query/querycompiler.js
+++ b/lib/query/querycompiler.js
@@ -892,6 +892,34 @@ class QueryCompiler {
         this.bindingsHolder
       )})`;
     } else {
+      function chunk(arr, start, amount) {
+        var result = [],
+          i,
+          start = start || 0,
+          amount = amount || 500,
+          len = arr.length;
+
+        do {
+          //console.log('appending ', start, '-', start + amount, 'of ', len, '.');
+          result.push(arr.slice(start, start + amount));
+          start += amount;
+        } while (start < len);
+
+        return result;
+    }
+
+    if (Array.isArray(statement.column))
+        return this.multiWhereIn(statement);
+
+    if(statement.value.length > 1000) {
+        let values = chunk(statement.value, 0, 1000);
+        return '(' + values.map(function(array) {
+            statement.value = array;
+            return this.whereIn(statement);
+        }.bind(this)).join(' OR ') + ')';
+    }
+
+
       columns = wrap_(
         statement.column,
         undefined,


OlivierCavadenti:
- Can you a comprehensive title ?
- Put in PR comment : Fixes #XXXX  where #XXXX is the number of the issue. That will make a link.
- It's oracle specific change, so I better see it in the oracle compiler file.

lib/query/querycompiler.js
```diff
@@ -892,6 +892,34 @@ class QueryCompiler {
...
+        } while (start < len);
+
+        return result;
+    }
+
+    if (Array.isArray(statement.column))
+        return this.multiWhereIn(statement);
+
+    if(statement.value.length > 1000) {
+        let values = chunk(statement.value, 0, 1000);
```
OlivierCavadenti, line 915: Lodash already have a chunk function that do that:
https://www.geeksforgeeks.org/lodash-_-chunk-method/

lib/query/querycompiler.js
```diff
@@ -892,6 +892,34 @@ class QueryCompiler {
...
+          start += amount;
+        } while (start < len);
+
+        return result;
+    }
+
+    if (Array.isArray(statement.column))
+        return this.multiWhereIn(statement);
+
+    if(statement.value.length > 1000) {
```
OlivierCavadenti, line 914: 1000 is a magic number, that need to be explain and documented. The reason is here: 
https://docs.oracle.com/en/database/oracle/oracle-database/12.2/sqlrf/Expression-Lists.html#GUID-5CC8FC75-813B-44AA-8737-D940FA887D1E

"A comma-delimited list of expressions can contain no more than 1000 expressions. A comma-delimited list of sets of expressions can contain any number of sets, but each set can contain no more than 1000 expressions."

lib/query/querycompiler.js
```diff
@@ -892,6 +892,34 @@ class QueryCompiler {
...
+        let values = chunk(statement.value, 0, 1000);
+        return '(' + values.map(function(array) {
+            statement.value = array;
+            return this.whereIn(statement);
+        }.bind(this)).join(' OR ') + ')';
```
OlivierCavadenti, line 919: All new addition need to be careful tested (unit test + integration test if applicable).

OlivierCavadenti (changes requested):

lib/query/querycompiler.js
```diff
@@ -892,6 +892,34 @@ class QueryCompiler {
...
+
+        return result;
+    }
+
+    if (Array.isArray(statement.column))
+        return this.multiWhereIn(statement);
+
+    if(statement.value.length > 1000) {
+        let values = chunk(statement.value, 0, 1000);
+        return '(' + values.map(function(array) {
```
maximelkin, line 916: you can use arrow function instead, to not .bind(this)

OlivierCavadenti:
I change the base to master, it's to 0.95 which is not correct. Rebase you commit on master.

PR #4809, opened 2021-11-05
Resolves #4583 Add JSDoc (TS flavor) to stub files

### LINKED ISSUE #4583: knex js snippet (auto complete) doesn't work?
in previous version I hadn't problem in knex.js, but in latest version when I wanna write code snippet(auto complete) doesn't work?

thanks

windupbird144:
Documents the types of the cli-generated files for init, migrate:make and seed:make. Uses the TS-specific flavor of JSDoc described in the [TypeScript](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) Handbook. This enables code hinting. Typescript users can use the [checkJs](https://www.typescriptlang.org/tsconfig/#checkJs) in their project to raise a compilation error e.g. if they forget to return a Promise from the migration function. So this provides an alternative to making the cli generate TS directly with the `-x ts` option. Being JSDoc, it does not assume any specific module system on the user's machine to import the Knex.

Some screenshots with IntelliSense in Visual Studio Code

Seed, current
seed-before

Seed, PR
seed-after

Knexfile, current
knexfile-before

Knexfile, PR
knexfile-after

Migrations, current
migrations-before

Migrations , PR
migrations-after

### DIFF
--- a/lib/migrations/migrate/stub/js.stub
+++ b/lib/migrations/migrate/stub/js.stub
@@ -1,4 +1,7 @@
-
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
 exports.up = function(knex) {
   <% if (d.tableName) { %>
   return knex.schema.createTable("<%= d.tableName %>", function(t) {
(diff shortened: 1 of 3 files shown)


OlivierCavadenti (approved):
seems ok for me, thanks for the contribution !

PR #4812, opened 2021-11-08
Mysql2 validate connection fix #4794

### LINKED ISSUE #4794: MYSQL2 Dialect - Connection not disposed after a connection error during query execution
# Environment

Knex version: 0.21.21 (same issue on 0.95.12)
Database + version: MYSQL 5.7
OS: Tested on Windows 10 and Linux Ubuntu 20.04.2

# Bug

1. When there is a connection issue (closed connection) during a query execution, the connection does not get removed from the pool and continues to be used, causing all queries executed after to fail with connection is closed error. The only way to recover from the issue is to restart the app. Note this does not happen on mysql dialect, only on Mysql2 dialect.

MYSQL2 does not emit an error event, when there is a connection issue while executing a query, but returns a fatal=true property in the exception. The dialect should handle this as a fatal error and set the __knex__disposed ?

2. Error: Can't add new command when connection is in closed state

3. Example Code (needs a mysql server)
```
    client: 'mysql2',
    connection: {
      host : '127.0.0.1',
      port : 3306,
      user : 'your_database_user',
      database : 'myapp_test'
    },
    pool: { min: 0, max: 7 }
});

const main = async function()
{
    try {
        await knex.raw("select sleep(100)");
        // !! Manually KILL Connection from MYSQL to simulate connection error => KILL <pid>
    } catch(err) {
        console.log(err);
    }

    // all queries from now on will return error when reusing the same connection in the pool
    try {
        await knex.raw("select 1");
    } catch(err) {
        console.log(err);
    }
    process.exit();
}
main();
```

OlivierCavadenti:
- Fix mysql2 validate connection.
- Closes https://github.com/knex/knex/issues/4794#issuecomment-963568255

### DIFF
--- a/lib/dialects/mysql/index.js
+++ b/lib/dialects/mysql/index.js
@@ -89,13 +89,9 @@ class Client_MySQL extends Client {
   }
 
   validateConnection(connection) {
-    if (
-      connection.state === 'connected' ||
-      connection.state === 'authenticated'
-    ) {
-      return true;
-    }
-    return false;
+    return (
+      connection.state === 'connected' || connection.state === 'authenticated'
+    );
   }
 
   // Grab a connection, run the query via the MySQL streaming interface,
--- a/lib/dialects/mysql2/index.js
+++ b/lib/dialects/mysql2/index.js
@@ -15,10 +15,13 @@ class Client_MySQL2 extends Client_MySQL {
     return require('mysql2');
   }
   validateConnection(connection) {
-    if (connection._fatalError) {
-      return false;
-    }
-    return true;
+    return (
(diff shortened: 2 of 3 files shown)


kibertoad (approved):

kibertoad:
Could you also backport this to 0.95?

OlivierCavadenti:
> Could you also backport this to 0.95?

Yeah put I push without a PR (I push on the upstream instead of fork unfortunaly): 
https://github.com/knex/knex/commits/0.95
Hope is good for you

kibertoad:
sure, np!
will finally publish today

Issue #4819, opened 2021-11-10
Support filetype .cjs for knexfile


VladRabota:
# Environment

Knex version: ^0.95.14
Database + version: mysql2
OS: Ubuntu 20.04
Node version: 16.13.0

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do
Hi, i am pretty new to programming, so i apologize if this issue didn't make sense.
So, at my current project i need to use ESModules and i have `"type": "module",` in my package.json file.
When i run `npx knex init` all fine, knexfile.js is created. But when i run `npx knex migrate:make test-migration` i get this message: 
```
ReferenceError: module is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '/home/vlad/nodejs/main-project-side/prototype1/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```
So i followed message and renamed knexfile.js to knexfile.cjs. I run `npx knex migrate:make test-migration` and get this message instead: 
```
Failed to resolve config file, knex cannot determine where to generate migrations
The "path" argument must be of type string. Received undefined
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:371:5)
    at validateString (node:internal/validators:119:11)
    at Object.resolve (node:path:1098:7)
    at /home/vlad/nodejs/main-project-side/prototype1/node_modules/knex/lib/migrations/migrate/MigrationGenerator.js:77:19
    at Array.map (<anonymous>)
    at MigrationGenerator._absoluteConfigDirs (/home/vlad/nodejs/main-project-side/prototype1/node_modules/knex/lib/migrations/migrate/MigrationGenerator.js:71:24)
    at MigrationGenerator._ensureFolder (/home/vlad/nodejs/main-project-side/prototype1/node_modules/knex/lib/migrations/migrate/MigrationGenerator.js:28:23)
    at MigrationGenerator.make (/home/vlad/nodejs/main-project-side/prototype1/node_modules/knex/lib/migrations/migrate/MigrationGenerator.js:20:16)
    at Migrator.make (/home/vlad/nodejs/main-project-side/prototype1/node_modules/knex/lib/migrations/migrate/Migrator.js:314:27)
    at Command.<anonymous> (/home/vlad/nodejs/main-project-side/prototype1/node_modules/knex/bin/cli.js:218:10)
```
[...]

PR #4824, opened 2021-11-11
Added cjs file extension to filetypes. Linked to #4819

VladRabota:
This PR should solve the problem which I encountered in project, that haves `"type": "module"` on the `package.json`. Details in [#4819](https://github.com/knex/knex/issues/4819). In short, knexfile.cjs didn't worked, so I added cjs to filetypes and then I able to use knexfile with cjs extension. I run test locally before and after change and got same test passed percent.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -64,7 +64,7 @@ async function initKnex(env, opts) {
 }
 
 function invoke() {
-  const filetypes = ['js', 'coffee', 'ts', 'eg', 'ls'];
+  const filetypes = ['js', 'coffee', 'ts', 'eg', 'ls', 'cjs'];
 
   const cwd = argv.knexfile
     ? path.dirname(path.resolve(argv.knexfile))


OlivierCavadenti (commented):
Hi,
I say an existing unfinished PR about this problem : https://github.com/knex/knex/pull/3638
I see it also add the extension in fs-migrations.js.
Also an integration test can useful (https://github.com/knex/knex/tree/master/test/cli)

VladRabota:
Hi, the extension in fs-migrations.js already there: https://github.com/knex/knex/pull/4382.
As I said, I'm new to programming and at this moment didn't write a single line of tests. I looked at tests, that you send me and the only thing thing that I can spot, that in [https://github.com/knex/knex/blob/master/test/cli/esm-interop.spec.js](https://github.com/knex/knex/blob/master/test/cli/esm-interop.spec.js) .cjs knexfiles was tested with `--esm` agrument, but I run migrations without `--esm`. Maybe, if someone tell me, what exactly I need to write, I can try to make tests.

VladRabota:
And also, I find from that issue, that I can run migrations like this `npx knex migrate:latest --knexfile knexfile.cjs` specifying path to knexfile. So this PR is not really necessary.

OlivierCavadenti:
So I can close it ?
Maybe could be nice to put in doc the command:
`npx knex migrate:latest --knexfile knexfile.cjs`

VladRabota:
Yep, we can close this PR. As far as I can understand, .cjs files not really in use.

PR #4833, opened 2021-11-18
let user can use customized String format function like push "N" in t…

qwerty30112:
this is related to the issue
https://github.com/knex/knex/issues/4582

first time, user can config like below:
```
const knex = require('./knex')({
  client: 'mssql', stringFormatter: (str) => {
    if (escape(str).indexOf("%u") > -1)
      return `N${str}`;
    return `${str}`;
  }});
```


qwerty30112:
Does anyone know how to merge the PR?

kibertoad:
@qwerty30112 This needs a test, types update and documentation.

PR #4836, opened 2021-11-18
Add binary uuid option #1778

### LINKED ISSUE #1778: knex should use binary(16) for UUID outside of PG
Currently, on databases other than PostgreSQL, knex is creating UUID tables as char(36), wasting a lot of space (and making it inconvenient to use those columns as primary keys).

knex should store UUIDs as binary(16) instead, as each UUID is a 128bit (=16 byte) number and it's only represented with 36 hex characters for convenience.

OlivierCavadenti:
- Add useBinaryUuid option in 'uuid' function.
- If this option is 'true' use 'binary(16)' type (other than PG).
- Closes #1778

### DIFF
--- a/lib/dialects/mssql/schema/mssql-columncompiler.js
+++ b/lib/dialects/mssql/schema/mssql-columncompiler.js
@@ -143,7 +143,9 @@ ColumnCompiler_MSSQL.prototype.json = 'nvarchar(max)';
 // TODO: mssql supports check constraints as of SQL Server 2008
 // so make enu here more like postgres
 ColumnCompiler_MSSQL.prototype.enu = 'nvarchar(100)';
-ColumnCompiler_MSSQL.prototype.uuid = 'uniqueidentifier';
+ColumnCompiler_MSSQL.prototype.uuid = ({ useBinaryUuid = false } = {}) =>
+  useBinaryUuid ? 'binary(16)' : 'uniqueidentifier';
+
 ColumnCompiler_MSSQL.prototype.datetime = 'datetime2';
 ColumnCompiler_MSSQL.prototype.bool = 'bit';
 
--- a/lib/dialects/redshift/schema/redshift-columncompiler.js
+++ b/lib/dialects/redshift/schema/redshift-columncompiler.js
@@ -2,6 +2,7 @@
 // -------
 
 const ColumnCompiler_PG = require('../../postgres/schema/pg-columncompiler');
+const ColumnCompiler = require('../../../schema/columncompiler');
 
 class ColumnCompiler_Redshift extends ColumnCompiler_PG {
   constructor() {
@@ -54,7 +55,7 @@ ColumnCompiler_Redshift.prototype.mediumblob = 'varchar(16777218)';
 ColumnCompiler_Redshift.prototype.set = 'text';
 ColumnCompiler_Redshift.prototype.text = 'varchar(max)';
 ColumnCompiler_Redshift.prototype.tinyblob = 'varchar(256)';
-ColumnCompiler_Redshift.prototype.uuid = 'char(36)';
+ColumnCompiler_Redshift.prototype.uuid = ColumnCompiler.prototype.uuid;
 ColumnCompiler_Redshift.prototype.varbinary = 'varchar(max)';
 ColumnCompiler_Redshift.prototype.bigint = 'bigint';
 ColumnCompiler_Redshift.prototype.bool = 'boolean';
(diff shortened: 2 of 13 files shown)


mishase (changes requested):
You should create function to transform uuid-s before inserting and after get. You can use something like uuid-parse package.
For example, currently i'm are using query transformer to automatically parse uuid in readable strings for mysql db like this:
```
function typeCast(field, next) {
	if (field.type === "STRING" && field.length === 16) {
		const buf = field.buffer();
		return buf ? uuidParse.unparse(buf) : null;
	} else {
		return next();
	}
}
```
this function makes easier to work with uuid-s in backend, but limits some functionality because it will trigger to any varchar(16) columns

kibertoad (approved):

OlivierCavadenti:
Ok, I see the point.
Knex include already uuid package (https://www.npmjs.com/package/uuid#uuidparsestr) so I will check later what I can do.

OlivierCavadenti:
> You should create function to transform uuid-s before inserting and after get. You can use something like uuid-parse package. For example, currently i'm are using query transformer to automatically parse uuid in readable strings for mysql db like this:
> 
> ```
> function typeCast(field, next) {
> 	if (field.type === "STRING" && field.length === 16) {
> 		const buf = field.buffer();
> 		return buf ? uuidParse.unparse(buf) : null;
> 	} else {
> 		return next();
> 	}
> }
> ```
> 
> this function makes easier to work with uuid-s in backend, but limits some functionality because it will trigger to any varchar(16) columns

As I can see it's not so easy to make that in Knex.
I want to make something generic, where the logic of parse/unparse is defined with column definition (in ColumnCompiler).
When you prepare Insert in querycompiler you dont have any information about the column type and you cant convert your uuid at this time.
Same problem in runner.js when you process the query.
If you have hints about how to get the type of column inside insert compiler and runner I am open @kibertoad.
If it's not possible, I can also make utils to manually convert uuid's before and after inserts.

kibertoad:
@OlivierCavadenti Knex should never have any info about column it is inserting to. I would suggest adding util method to knex-utils instead.

OlivierCavadenti:
> @OlivierCavadenti Knex should never have any info about column it is inserting to. I would suggest adding util method to knex-utils instead.

One other option is to provide utils in knex (like knex.raw...) to convert in binary with database specific SQL (like BIN_TO_UUID in MYSQL: https://stackoverflow.com/a/45186851/1535159).  Need more work btw.

kibertoad:
yeah, that works too

intech:
In cockroachdb, pg types are used and I have to fix the answer with hexadecimal data, I temporarily did this:
```js
const { types } = require("pg");
types.setTypeParser(17, value => Buffer.from(value.replace("\\x", ""), "hex"));
```
You need to think about a more beautiful solution.

OlivierCavadenti:
> In cockroachdb, pg types are used and I have to fix the response with hexadecimal data, I temporarily did this:
> 
> ```js
> const { types } = require("pg");
> types.setTypeParser(17, value => Buffer.from(value.replace("\\x", ""), "hex"));
> ```
> 
> We need to think about a more beautiful solution.
> 
> For insertion the uuid format can be any hex, or formatted with a separator, or guid in curly braces, all of these formats are converted automatically to the specified column type, or text representation, or binary (hex)

Thanks for comments. I need to finish another task before this PR, but as I can see, a lot of people have workaround and wait this. I will find a clean solution to make the use as smooth as possible :D.

kibertoad:
@OlivierCavadenti BTW, can you add yourself to the list of maintainers in package.json?

OlivierCavadenti:
I choose to add two functions in FunctionHelper.js to avoid dependencies and SQL db specific generation.

To use : 

Create table with option actived on uuid column:

```js
await knex.schema.createTable('uuid_table', (t) => {
  t.uuid('uuid_col_binary', { useBinaryUuid: true });
});
```

Insert your uuid in binary format with uuidToBin function: 

```js
await knex('uuid_table').insert({
  uuid_col_binary:  knex.fn.uuidToBin('3f06af63-a93c-11e4-9797-00505690773f'),
});
```

Retreive it and convert it:

```js
const uuid = await knex('uuid_table').select('uuid_col_binary');
knex.fn.binToUuid(uuid[0].uuid_col_binary)
```

OlivierCavadenti:
@mishase @intech the solution is good for you (see comment above) ?

kibertoad (approved):

PR #4843, opened 2021-11-23
Recreate client on `knex.initialize` to use new config

### LINKED ISSUE #3828: Manually initializing a destroyed connection pool did not change database schema name
# Summary

I am setting up test script to create multiple database schema to utilize [AVA] parallel test run. As a result, I need to create the schema first, then switch the application connection pool to use the new created schema. However I encountered a bug in doing so, and would like to ask the official proper way to change the database schema name.

# Environment

Knex version: 0.21.0
Database + version: MySQL 5.7
OS: Windows 10

# Bug

According to the documentation:
> If you ever need to explicitly teardown the connection pool, you may use `knex.destroy([callback])`. You may use `knex.destroy` by passing a callback, or by chaining as a promise, just not both. To manually initialize a destroyed connection pool, you may use `knex.initialize([config])`, if no config is passed, it will use the first knex configuration used.

I expect I can change the database name by `knex.destroy` and `knex.initialize(newConfig)` with new configuration. This does not work.

Reduced test case as follows:

```js
const knex = require('knex')

function getConfig () {
  return {
    client: 'mysql',
    connection: {
      host: 'localhost',
      user: 'root',
      password: ''
    }
  }
}

const db = knex(getConfig()).on('query', function (data) {
  console.log(data.sql)
})

const method = 1

;(async () => {
  // Randomly (in my case it is based on server listening port) assigned a database name for AVA parallel testing
  const dbName = 'TMP_DB_' + (Math.floor(Math.random() * 999) + 9001)
  try {
    await db.raw(`CREATE DATABASE ${dbName}`)
    await db.raw(`USE ${dbName}`)
    await db.schema.createTable('tmp1', (table) => {
      table.increments()
      table.string('val1', 10).notNullable()
    })
    if (method === 1) {
      // I tried the documented way, but the `connection.database` is not changed with the following code
      await db.destroy()
      const newConfig = getConfig()
      newConfig.connection.database = dbName
      db.initialize(newConfig)
    } else if (method === 2) {
      // Is this the proper way to set the `connection.database`?
      db.client.connectionSettings.database = dbName
    }
    // This will result in `Error: ER_NO_DB_ERROR: No database selected` if database name is not set
    await Promise.all([
      db('tmp1').insert({ val1: 'a' }),
      db('tmp1').insert({ val1: 'b' }),
      db('tmp1').insert({ val1: 'c' })
    ])
  } finally {
    try {
      await db.raw(`DROP DATABASE ${dbName}`)
    } catch (ignore) {}
    await db.destroy()
  }
  console.log('End')
})().catch(e => console.log(e.stack))
```

# Question

My additional question together with this bug, is how to change the database schema of the connection pool? Is the `const method = 2` the proper way to do so? Or it can only be treated as undocumented way and may change in the future?

[AVA]: https://github.com/avajs/ava

MehdiSaffar:
This PR is an attempt to close #3828. When calling `knex.destroy` then `knex.initialize([config])`, the current version of knex does not take into account the passed new config. This PR makes `knex.initialize` replace the `knex.client` with a new `Client` based on the `config` using `resolveConfig`, similar to what happens in `makeKnex`

# Notes
- I couldn't use `npm run db:start:postgres` due to the following error that I couldn't fix: 
```
Creating scripts_postgres_1 ... error
ERROR: for scripts_postgres_1  Cannot start service postgres: cgroups: cgroup mountpoint does not exist: unknown
ERROR: for postgres  Cannot start service postgres: cgroups: cgroup mountpoint does not exist: unknown
ERROR: Encountered errors while bringing up the project.   
```
- In my own local test database, I have two roles `postgres` and `postgres2`. I did roughly the following:
```js
const knex = require("knex")({
  client: 'pg',
  connection: {
    user: 'postgres',
    password: 'postgres',
    host: 'localhost',
    database: 'test',
  }
})

await knex.raw('SELECT "current_user"()') // returns "postgres"
await knex.destroy()

knex.initialize({
  client: 'pg',
  connection: {
    user: 'postgres2',
    password: 'postgres2',
    host: 'localhost',
    database: 'test',
  }
})
await knex.raw('SELECT "current_user"()') // returns "postgres2"
```
this proves that knex effectively used the second config.
- This is the first time I contribute to open source so please be kind :)

### DIFF
--- a/lib/knex-builder/make-knex.js
+++ b/lib/knex-builder/make-knex.js
@@ -7,6 +7,7 @@ const QueryInterface = require('../query/method-constants');
 const merge = require('lodash/merge');
 const batchInsert = require('../execution/batch-insert');
 const { isObject } = require('../util/is');
+const { resolveConfig } = require('./internal/config-resolver');
 
 // Javascript does not officially support "callable objects".  Instead,
 // you must create a regular Function and inject properties/methods
@@ -164,7 +165,14 @@ function initContext(knexFn) {
 
     // Typically never needed, initializes the pool for a knex client.
     initialize(config) {
-      return this.client.initializePool(config);
+      if (this.client.pool) {
+        this.client.logger.warn('The pool has already been initialized');
+        return;
+      }
+
+      const { resolvedConfig, Dialect } = resolveConfig(config);
+      const client = new Dialect(resolvedConfig);
+      this.client = client;
     },
 
     // Convenience method for tearing down the pool.


kibertoad:
Thank you! Can you please add integration test for this?

MehdiSaffar:
I am having trouble using the docker-compose for setting up the database, so I kinda "blindly" created an integration test. It takes cares of setting up a dummy role to which we can switch to by providing a new config. It also cleans up the role after it. I hope this is good enough.

kibertoad:
Can you push that test?

MehdiSaffar:
Sorry I thought I pushed ahah, I pushed now ;)

kibertoad:
You can fix test using sqlite to reproduce, that one shouldn't require Docker.

PR #4856, opened 2021-11-26
Fix for createViewOrReplace and its support for sqlite3

thefoxie:
Related issues:
https://github.com/knex/knex/issues/4854
https://github.com/knex/knex/issues/4855

### DIFF
--- a/test/integration2/schema/views.spec.js
+++ b/test/integration2/schema/views.spec.js
@@ -143,6 +143,13 @@ describe('Views', () => {
                   "create or replace view `view_test` (`a`, `b`) as select `a`, `b` from `table_view` where `b` > '10'",
                 ]
               );
+              tester(
+                ['sqlite3'],
+                [
+                  'drop view if exists `view_test`',
+                  "create view `view_test` (`a`, `b`) as select `a`, `b` from `table_view` where `b` > '10'",
+                ]
+              );
               tester('mssql', [
                 "CREATE OR ALTER VIEW [view_test] ([a], [b]) AS select [a], [b] from [table_view] where [b] > '10'",
               ]);
@@ -175,9 +182,16 @@ describe('Views', () => {
                 ]
               );
               tester(
-                ['sqlite3', 'mysql'],
+                ['mysql'],
                 [
                   "create or replace view `view_test` as select `a`, `b` from `table_view` where `b` > '10'",
+                ]
+              );
+              tester(
+                ['sqlite3'],
+                [
+                  'drop view if exists `view_test`',
+                  "create view `view_test` as select `a`, `b` from `table_view` where `b` > '10'",
                 ]
               );
               tester('mssql', [
--- a/test/integration2/schema/views.spec.js
+++ b/test/integration2/schema/views.spec.js
@@ -122,6 +122,81 @@ describe('Views', () => {
             });
         });
 
+        it('create or replace view', async () => {
+          await knex.schema
+            .createViewOrReplace('view_test', function (view) {
+              view.columns(['a', 'b']);
+              view.as(
+                knex('table_view').select('a', 'b').where('b', '>', '10')
+              );
+            })
+            .testSql((tester) => {
+              tester(
+                ['pg', 'pg-redshift', 'cockroachdb', 'oracledb'],
+                [
+                  'create or replace view "view_test" ("a", "b") as select "a", "b" from "table_view" where "b" > \'10\'',
+                ]
+              );
+              tester(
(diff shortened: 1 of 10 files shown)


test/integration2/schema/views.spec.js
```diff
@@ -122,6 +122,95 @@ describe('Views', () => {
             });
         });
 
+        it('create or replace view', async () => {
+          await knex.schema
```
OlivierCavadenti, line 126: can you create view with same name before, test it exists, then make this test to really test that view is really replaced.

test/unit/schema-builder/sqlite3.js
```diff
@@ -110,14 +110,30 @@ describe('SQLite SchemaBuilder', function () {
...
+    });
+
+    it('create view or replace without columns', async function () {
+      const viewSql = client
+        .schemaBuilder()
+        .createViewOrReplace('adults', function (view) {
+          view.as(knexSqlite3('users').select('name').where('age', '>', '18'));
+        })
+        .toSQL();
+      equal(1, viewSql.length);
```
OlivierCavadenti, line 133: use 'expect' instead of 'equal' (its old pratice)

OlivierCavadenti (commented):
Thanks for the report + PR ! Tests fails

test/unit/schema-builder/sqlite3.js
```diff
@@ -110,14 +110,30 @@ describe('SQLite SchemaBuilder', function () {
...
-          .toSQL();
-      }).to.throw('rename column of views is not supported by this dialect.');
+      const viewSql = client
+        .schemaBuilder()
+        .createViewOrReplace('adults', function (view) {
+          view.columns(['name']);
+          view.as(knexSqlite3('users').select('name').where('age', '>', '18'));
+        })
+        .toSQL();
+      equal(1, viewSql.length);
```
OlivierCavadenti, line 120: You have two queries, so == 2

OlivierCavadenti (changes requested):

OlivierCavadenti:
@kibertoad maybe could be nice to add this in 0.95.15 ?

kibertoad:
sure, sounds good

OlivierCavadenti (dismissed):

OlivierCavadenti:
@kibertoad I push returns, if you can take a look.

thefoxie:
@OlivierCavadenti sorry, I was busy and then sick. Have you sorted it out? BTW I was thinking whether the functionality could behave the same as in sqlite3 across all dbs. In postgres, when you use the builtin functionality, it needs to have the same columns in same order and new ones at the end. I ended up using drop and create anyway.
```js
await Model.knex().schema.dropViewIfExists(dBViewName);

dbView = await Model.knex().schema.createView(
    dBViewName,
    (view) => {
        view.as(Model.knex().raw(query));
    }
);
```

OlivierCavadenti:
> @OlivierCavadenti sorry, I was busy and then sick. Have you sorted it out? BTW I was thinking whether the functionality could behave the same as in sqlite3 across all dbs.

No problem ^^.

> In postgres, when you use the builtin functionality, it needs to have the same columns in same order and new ones at the end.

I am afraid I dont understand, you need to create new view but with same columns in same order when you use createViewOrReplace ? I dont test that, but it throws an error ?

OlivierCavadenti:
> BTW I was thinking whether the functionality could behave the same as in sqlite3 across all dbs.

> In postgres, when you use the builtin functionality, it needs to have the same columns in same order and new ones at the end.

By the way, I assume this limitation in PostgreSQL (need replace with same columns, same order) it's to dont break all existing queries that query the view you replace.

In my opinion, it's better to dont change existing queries 'create or replace view' and so have this safe behavior, and put in Knex documentation what you are saying : the difference between sqlite3 and other dbs.

@kibertoad what do think about that ?

kibertoad:
@OlivierCavadenti I agree, handling it via documentation is a better path.

OlivierCavadenti:
@kibertoad we will put this one on 1.0.0 (with backport in 0.95.X) or we will merge it after the 1.0.0 tag ?

kibertoad:
@OlivierCavadenti I'm OK with merging it in 1.0.0 if you prefer it that way.

kibertoad (approved):

kibertoad:
@OlivierCavadenti Any last minute changes, or I can start releaseing 1.0.0 in an hour or so?

kibertoad:
Released in 1.0.1

PR #4875, opened 2021-12-05
feat: Checks Constraints Support for CockroachDB

intech:
Related: [4874](https://github.com/knex/knex/pull/4874)

### DIFF
--- a/lib/dialects/cockroachdb/crdb-tablecompiler.js
+++ b/lib/dialects/cockroachdb/crdb-tablecompiler.js
@@ -7,6 +7,68 @@ class TableCompiler_CRDB extends TableCompiler {
     super(client, tableBuilder);
   }
 
+  addColumns(columns, prefix, colCompilers) {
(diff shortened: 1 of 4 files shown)


lib/dialects/cockroachdb/crdb-tablecompiler.js
```diff
@@ -7,6 +7,68 @@ class TableCompiler_CRDB extends TableCompiler {
     super(client, tableBuilder);
   }
 
+  addColumns(columns, prefix, colCompilers) {
+    if (prefix === this.alterColumnsPrefix) {
```
OlivierCavadenti, line 11: Can you put similar code in function in pg-tablecompiler that add queries for alter columns, call the function in pg and crdb file, then make cockroachdb specific pushQuery ? This code is duplicated.

OlivierCavadenti (changes requested):
changes for duplicate code

OlivierCavadenti (approved):

Issue #4825, opened 2021-11-11
MSSQL unique creates index instead of column constraint


edwardwu-servian:
# Environment

Knex version: ^0.95.14
Database + version: Microsoft SQL Azure (RTM) - 12.0.2000.8
OS: macOS 11.6.1

@smorey2 

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do
```javascript
exports.up = function (knex, Promise) {
  return knex.schema
    .withSchema("meta")
    .createTable("constraint_artefact_type", function (table) {
      table.increments("constraint_artefact_type_id");
      table.string("value").notNullable();
      table.unique("value");
    });
};
```
https://github.com/knex/knex/blob/27ade6f8814a5601916489a96474b55649e15462/lib/dialects/mssql/schema/mssql-tablecompiler.js#L307
According to this code, this will generate 
```SQL
CREATE TABLE [meta].[constraint_artefact_type] ([constraint_artefact_type_id] int identity(1,1) not null primary key, [value] nvarchar(255) not null;
CREATE UNIQUE INDEX [constraint_variable_object_type_value_unique] ON [meta].[constraint_variable_object_type] ([value]) WHERE [value] IS NOT NULL;
```
However, this filtered unique index has made it impossible to use `[value]` column in subsequent Foreign Key constraints. Below statement will fail creating foreign key. This is highlighted in current implementation (https://stackoverflow.com/questions/767657/how-do-i-create-a-unique-constraint-that-also-allows-nulls/767702#767702)
```javascript
exports.up = function (knex, Promise) {
  return knex.schema
    .withSchema("meta")
    .createTable("artefact", function (table) {
      table
        .string("type")
        .notNullable();
      table
        .foreign("type")
        .references("value")
        .inTable("meta.constraint_artefact_type")
        .onDelete("SET NULL");
      table.unique(["name", "source"]);
    });
};
```

Expected behaviour here is ` table.unique("value")` would be generating a UNIQUE constraint on `[value]` column (which would be similar to postgres)
```SQL
CREATE TABLE [meta].[constraint_artefact_type] ([constraint_artefact_type_id] int identity(1,1) not null primary key, [value] nvarchar(255) not null UNIQUE;
```
or
```SQL
CREATE TABLE [meta].[constraint_artefact_type] ([constraint_artefact_type_id] int identity(1,1) not null primary key, [value] nvarchar(255) not null;
[...]

PR #4887, opened 2021-12-12
Added support for MSSQL unique constraint

mlevit:
MSSQL supports both unique indexes and unique constraints. One of the issues we found was unique index (current implementation) cannot be used with foreign key relationships. We must use unique constraints instead.

This is all based on issue #4825

I thought I'd give it a go and update the way MSSQL handles `table.unique`. This is my first PR here so please let me know if there's anything missing, which I'm sure there is.

I created a new test which seems to work fine 🤞

### DIFF
--- a/test/integration2/dialects/mssql.spec.js
+++ b/test/integration2/dialects/mssql.spec.js
@@ -170,7 +170,7 @@ describe('MSSQL dialect', () => {
           });
         });
 
-        describe('unique table constraint with options object', () => {
+        describe('unique table index with options object', () => {
           const tableName = 'test_unique_index_options';
           before(async () => {
             await knex.schema.createTable(tableName, function () {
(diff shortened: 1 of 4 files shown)


test/integration2/dialects/mssql.spec.js
```diff
@@ -199,6 +199,38 @@ describe('MSSQL dialect', () => {
...
+
+          it('accepts indexName and constraint in options object', async () => {
+            const indexName = `UK_${tableName}_x_y`;
+            await knex.schema.alterTable(tableName, function () {
+              this.unique(['x', 'y'], {
+                indexName: indexName,
+                constraint: true,
+              });
+            });
+            expect(
```
kibertoad, line 223: shouldn't this line be awaited?

kibertoad:
Can you create a PR for https://github.com/knex/documentation with documentation?
Also TypeScript types need to be updated

OlivierCavadenti (commented):
You can push tests with add another one in mssql.js file (after "test adding unique key" test) and check the generated SQL.

mlevit:
@kibertoad could you point me in the right direction? I'm no TS developer, so not really sure what to look at.

test/integration2/dialects/mssql.spec.js
mlevit: Copied the test above "unique table index with options object". Looks like other `except` tests are awaited. Want me to change?

kibertoad: yes, please!

kibertoad:
@mlevit I think somewhere around here: https://github.com/knex/knex/blob/2763e47129837111fc5fd4dd5fb19996580bc01d/types/index.d.ts#L2078

test/integration2/dialects/mssql.spec.js
mlevit: Done 7a9de41

kibertoad (approved):

mlevit:
@kibertoad @OlivierCavadenti are we happy with the argument name `constraint`? Is it descriptive enough or should we change it?

kibertoad:
@mlevit Something more explicitly boolean `isContraint` or `useConstraint` might be less surprising, but I defer to @OlivierCavadenti here.

mlevit:
My only issue is the argument is only usable for MSSQL... so not really sure what the best play here is. Would `createMssqlConstraint` be too verbose? `isConstraint` is good, but sort of doesn't make sense for other DBs.

kibertoad:
@mlevit We don't do db-specific naming for db-specific parameters. Note in documentation is sufficient

mlevit:
Documentation updated https://github.com/knex/documentation/pull/373

OlivierCavadenti:
> @kibertoad @OlivierCavadenti are we happy with the argument name `constraint`? Is it descriptive enough or should we change it?

useConstraint seems nice in my opinion, more descriptive.

mlevit:
> > @kibertoad @OlivierCavadenti are we happy with the argument name `constraint`? Is it descriptive enough or should we change it?
> 
> useConstraint seems nice in my opinion, more descriptive.

Done.

OlivierCavadenti:
Thanks !

PR #4903, opened 2021-12-20
Create multiple PKs with increments

OlivierCavadenti:
- Resolve the case when a increments column is created in same time that primary keys. Now you can specify a increments columns with or without primary keys: 
   * If a primary key is specified, the increment column is added to composite key,
   * If a primary key is not specified, the increment column is set to primary key.

This two cases works :

```js
knex.schema.createTable('table_multiple_keys', function (t) {
  t.primary(['id', 'second_id', 'other_col']);
  t.increments('id');

  t.string('second_id', 16).notNullable();
  t.integer('other_col').notNullable();
});
```

```js
knex.schema.createTable('table_multiple_keys', function (t) {
  t.primary(['second_id', 'other_col']);
  t.increments('id');

  t.string('second_id', 16).notNullable();
  t.integer('other_col').notNullable();
});
```
- Solve tricky cases : 
     * SQLite : it's not possible to specify autoincrement with composite key, autoincrement need to be primary key. I solve it by transform primary keys in unique index if autoincrement with primary keys is specified.
     * MYSQL don't support autoincrement without primary keys so it's not possible to add autoincrement column without primary key and alter table to add primary key in second query. I solve it by add classic int column, add composite primary key and transform the column to autoincrement.

- Closes https://github.com/knex/knex/issues/385

### DIFF
--- a/lib/dialects/mssql/schema/mssql-columncompiler.js
+++ b/lib/dialects/mssql/schema/mssql-columncompiler.js
@@ -126,12 +126,22 @@ class ColumnCompiler_MSSQL extends ColumnCompiler {
     });
     return '';
   }
+
+  increments(options = { primaryKey: true }) {
+    return (
+      'int identity(1,1) not null' +
+      (this.tableCompiler._canBeAddPrimaryKey(options) ? ' primary key' : '')
+    );
+  }
+
+  bigincrements(options = { primaryKey: true }) {
+    return (
+      'bigint identity(1,1) not null' +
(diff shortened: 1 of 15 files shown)


OlivierCavadenti:
@kibertoad CI seems KO :X

kibertoad (approved):

kibertoad:
@OlivierCavadenti Great job, as always! Do you think this needs additional documentation?

PR #4909, opened 2021-12-24
Add type to typescript knexfile

tharrison4815:


lib/migrations/migrate/stub/knexfile-ts.stub
```diff
@@ -1,7 +1,8 @@
-// Update with your config settings.
+import type { Knex } from "knex";
 
-module.exports = {
+// Update with your config settings.
 
+const config: { [key: string]: Knex.Config } = {
```
kibertoad, line 5: why [key: string]?

tharrison4815: It's an object with strings as the property key names and config objects as the property values. Is that not what knex expects?

kibertoad: isn't just Knex.Config sufficient?

kibertoad: oh, wait, this is for knexfile. Right

kibertoad (approved):

kibertoad:
Released in 1.0.1.

PR #4915, opened 2021-12-30
Enable wrapIdentifier for SQLite .hasTable

### LINKED ISSUE #4898: hasTable method doesn't transform the name into snake_case
# Environment

Knex version: 0.95.14
Database + version: SQLite3@5.0.2
OS: MacOS 10.15.7

# Bug

I have defined postProcessResponse and wrapIdentifier but it seems that with the hasTable method it doesn't work.

```js
db.schema.hasTable('infoBoxes').toString()
```
```sql
select * from sqlite_master where type = 'table' and name = 'infoBoxes'
```

```js
  db.schema.createTable('infoBoxes', (t) => {
    t.increments('id').primary()
    t.string('text').notNullable()
    t.string('textColor')
    t.string('borderStyle')
    t.string('borderWidth')
    t.string('borderColor')
    t.string('backgroundColor')
    t.timestamp('createdAt').defaultTo(db.fn.now());
  }).toString()
```
```sql
create table `info_boxes` (`id` integer not null primary key autoincrement, `text` varchar(255) not null, `text_color` varchar(255), `border_style` varchar(255), `border_width` varchar(255), `border_color` varchar(255), `background_color` varchar(255), `created_at` datetime default CURRENT_TIMESTAMP)
```

ramos-ph:
Currently, the `wrapIdentifier` option does not apply for SQLite `db.schema.hasTable`.

For example, even if we set `wrapIdentifier` to parse identifiers to snake case, we'd get the following:

```js
db.schema.hasTable('infoBoxes').toString()
> select * from sqlite_master where type = 'table' and name = 'infoBoxes'
```

This PR fixes this by enabling the wrap for SQLite Compiler:

```js
db.schema.hasTable('infoBoxes').toString()
> select * from sqlite_master where type = 'table' and name = 'info_boxes'
```

Fix #4898

### DIFF
--- a/test/integration2/schema/misc.spec.js
+++ b/test/integration2/schema/misc.spec.js
@@ -1577,6 +1577,12 @@ describe('Schema (misc)', () => {
             expect(resp).to.equal(false);
           }));
 
+        it('should not parse table name if wrapIdentifier is not specified', () => {
+          knex.schema.hasTable('testTableTwo').then((resp) => {
+            expect(resp).to.equal(false);
+          });
+        });
+
         it('should parse table name if wrapIdentifier is specified', () => {
           knex.client.config.wrapIdentifier = (value, origImpl, queryContext) =>
             origImpl(_.snakeCase(value));
--- a/test/integration2/schema/misc.spec.js
+++ b/test/integration2/schema/misc.spec.js
@@ -1578,18 +1578,22 @@ describe('Schema (misc)', () => {
           }));
 
         describe('sqlite only', () => {
-          if (!isSQLite(knex)) {
-            return Promise.resolve();
-          }
+          it('should not parse table name if wrapIdentifier is not specified', async function () {
+            if (!isSQLite(knex)) {
+              return this.skip();
+            }
 
-          it('should not parse table name if wrapIdentifier is not specified', async () => {
             knex.client.config.wrapIdentifier = null;
 
             const resp = await knex.schema.hasTable('testTableTwo');
-            expect(resp).to.equal(false);
+            expect(resp).to.equal(true);
           });
 
-          it('should parse table name if wrapIdentifier is specified', async () => {
+          it('should parse table name if wrapIdentifier is specified', async function () {
+            if (!isSQLite(knex)) {
+              return this.skip();
+            }
+
             knex.client.config.wrapIdentifier = (
               value,
               origImpl,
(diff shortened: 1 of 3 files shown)


OlivierCavadenti:
Can you add tests for this plz ?

test/integration2/schema/misc.spec.js
```diff
@@ -1576,6 +1576,21 @@ describe('Schema (misc)', () => {
...
+          });
+        });
+
+        it('should parse table name if wrapIdentifier is specified', () => {
+          knex.client.config.wrapIdentifier = (value, origImpl, queryContext) =>
+            origImpl(_.snakeCase(value));
+
+          knex.schema.hasTable('testTableTwo').then((resp) => {
+            expect(resp).to.equal(true);
+          });
```
talyssonoc, line 1592: Since these methods return promises we need to let the test know that its result is async by returning the promise from it or using async/await (as it is done in other tests of this file):

```suggestion
        it('should not parse table name if wrapIdentifier is not specified', async () => {
          const resp = await knex.schema.hasTable('testTableTwo')
          expect(resp).to.equal(false);
        });

        it('should parse table name if wrapIdentifier is specified', async () => {
          knex.client.config.wrapIdentifier = (value, origImpl, queryContext) =>
            origImpl(_.snakeCase(value));

          const resp = await knex.schema.hasTable('testTableTwo')
          expect(resp).to.equal(true);
```

ramos-ph: Hey, thanks for your suggestion!
I thought of using async/await, but I saw that the other test cases of `hasTable` were using `.then`, so I based myself on that.
https://github.com/knex/knex/blob/master/test/integration2/schema/misc.spec.js#L1564

ramos-ph: Thanks for your suggestion! I haven't noticed before :smile:

ramos-ph:
Hey, I came across a few bug on the test file:
```js
describe('sqlite only', () => {
            if (!isSQLite(knex)) {
              return Promise.resolve();
            }
```
This will never work because the `if` statement runs before the `describe` which causes `knex` to be undefined - never returning true on the function output.
I double checked other workflows to make sure it wasn't a problem with my branch.
Should a new issue be considered?

kibertoad:
> Should a new issue be considered?

Yes, please! Could you create one?

ramos-ph:
> Yes, please! Could you create one?

Sure! Leave it to me.

kibertoad:
@OlivierCavadenti Could you review this?

OlivierCavadenti (approved):

kibertoad:
Released in 1.0.1.

PR #4928, opened 2022-01-11
remove outdated warning message

beenotung:
Returning is supported by sqlite3 in the current version.

Reference: https://sqlite.org/lang_returning.html
Discussion: https://stackoverflow.com/a/67644622/3156509

### DIFF
--- a/lib/dialects/sqlite3/query/sqlite-querycompiler.js
+++ b/lib/dialects/sqlite3/query/sqlite-querycompiler.js
@@ -22,12 +22,6 @@ class QueryCompiler_SQLite3 extends QueryCompiler {
 
     const { returning } = this.single;
 
-    if (returning) {
-      this.client.logger.warn(
-        '.returning() is not supported by sqlite3 and will not have any effect.'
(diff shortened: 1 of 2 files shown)


kibertoad (approved):

OlivierCavadenti (commented):
"should warn to user when use `.returning()` function in SQLite3"

unit test need to be removed.

@kibertoad lint seems to throw an error since yesterday, I dont reproduce in local.

kibertoad:
I'll take a look at lint error, probably new versions of something got released

kibertoad:
@OlivierCavadenti were returning tests enabled for sqlite before?

kibertoad:
@OlivierCavadenti I think this might have been a premature change. I've tried enabling tests for SQLite, and I don't think that `returning` is working as expected. When performing inserts, I don't see SQLite driver actually returning anything, even when I add the knex part for handling it.

kibertoad:
Released in 1.0.1, supported only by better-sqlite3 for now.

Issue #4922, opened 2022-01-06
Some tests are not being executed


ramos-ph:
# Environment

Knex version: 0.95.14

# Bug

On the `/test/integration2/schema/misc.spec.js` file, there are some statements like this:

```js
describe('sqlite only', () => {
  if (!isSQLite(knex)) {
    return Promise.resolve();
  }
```
This will never work because the `if` statement runs before the `describe`, which causes `knex` to be undefined, hence not executing anything inside the `describe` block because the function will always return `false`.

What we should be doing is skipping the test cases, or finding another way to skip those tests.

```js
describe('sqlite only', () => {
  it('should do something', async function() {
    if (!isSQLite(knex)) {
      return this.skip();
    }
    
    // assertions
  });
});
```

PR #4930, opened 2022-01-12
Change skip tests

perchespierri:


perchespierri:
Closes #4922

test/integration2/schema/misc.spec.js
```diff
@@ -2191,24 +2210,34 @@ describe('Schema (misc)', () => {
...
               .where(fieldName, 'something')
               .then((rows) => {
                 expect(rows.length).to.equal(0);
-              }));
+              });
+          });
 
-          it('should throw when referencing a non-existent column', () =>
-            knex(tableName)
+          it('should throw when referencing a non-existent column', () => {
```
ramos-ph, line 2226:
```suggestion
          it('should throw when referencing a non-existent column', function () {
```
We missed that one :smile:

test/integration2/schema/misc.spec.js
```diff
@@ -543,19 +551,19 @@ describe('Schema (misc)', () => {
...
-          if (!isPgBased(knex)) {
-            return Promise.resolve();
-          }
-
           afterEach(() =>
             knex.schema
               .dropTableIfExists('native_enum_test')
               .raw('DROP TYPE IF EXISTS "foo_type"')
               .raw('DROP SCHEMA IF EXISTS "test" CASCADE')
           );
```
ramos-ph, line 559: We'll have to add a PG check here too. There are a few tests breaking because of SQL syntax error.

test/integration2/schema/misc.spec.js
```diff
@@ -543,19 +551,20 @@ describe('Schema (misc)', () => {
         });
 
         describe('enum - postgres', () => {
-          if (!isPgBased(knex)) {
-            return Promise.resolve();
-          }
+          afterEach(async () => {
+            if (!isPgBased(knex)) {
```
ramos-ph, line 555:
```suggestion
            if (isPgBased(knex)) {
```
Wrong logic here. We must run when it is PG based

kibertoad:
Thank you so much!

PR #4934, opened 2022-01-14
Fix tests for RETURNING and enable them for SQLite.

kibertoad:
Functionality doesn't seem to work yet


kibertoad:
@OlivierCavadenti Could you take a look at this and see if I'm missing something and there is actually a way to get necessary data from SQLite? The way I see it, 
`connection[callMethod](obj.sql, obj.bindings, function (err, response) {`
part of SQLite dialect is returning `undefined` response and we don't actually get anything back. Wonder if we need to do something additionally to make SQLite to return data for inserts.

kibertoad:
Related ticket: https://github.com/microsoft/vscode-node-sqlite3/issues/15

OlivierCavadenti:
> @OlivierCavadenti Could you take a look at this and see if I'm missing something and there is actually a way to get necessary data from SQLite? The way I see it, `connection[callMethod](obj.sql, obj.bindings, function (err, response) {` part of SQLite dialect is returning `undefined` response and we don't actually get anything back. Wonder if we need to do something additionally to make SQLite to return data for inserts.

Like you I made some tests and I dont find anything to get back the results... maybe something missing in the library.

kibertoad:
@OlivierCavadenti I will restore warnings for 1.0.0 then, and hopefully we'll get some response from vscode sqlite team
I'll also try with better-sqlite, maybe that one works.

OlivierCavadenti:
@kibertoad this can be reimplemented since we have the old-new sqlite3 package :D.

kibertoad:
Will take a look if they expose necessary data now

PR #4940, opened 2022-01-17
Support of MATERIALIZED and NOT MATERIALIZED with WITH/CTE

OlivierCavadenti:
- Add withMaterialized and withNotMaterialized functions for PG and SQLite to add 'materialized' and 'not materialized' keywords with WITH/CTE (see https://www.postgresql.org/docs/12/queries-with.html).
- Close https://github.com/knex/knex/issues/4470


kibertoad:
Can you also add an integration test which would actually run the sql?

santacruz123:
#4948 alternative version

kibertoad (approved):

OlivierCavadenti:
@kibertoad  I push a fix of unit test :D :

image

PR #4948, opened 2022-01-19
materialized CTE support for Postgres

santacruz123:
`knex.withMaterialized('t',knex('users')).from('t')` for `with 't' as materialized (select * from users) select * from t`

and 

`knex.withNotMaterialized('t',knex('users')).from('t')` -> `with 't' as not materialized (select * from users) select * from t`

https://www.postgresql.org/docs/current/queries-with.html#id-1.5.6.12.7

I will add documentation changes if this is approved

### DIFF
--- a/lib/query/querybuilder.js
+++ b/lib/query/querybuilder.js
@@ -124,6 +124,20 @@ class Builder extends EventEmitter {
     return this.withWrapped(alias, statementOrColumnList, nothingOrStatement);
   }
 
+  withMaterialized(alias, statementOrColumnList, nothingOrStatement) {
+    validateWithArgs(alias, statementOrColumnList, nothingOrStatement, 'with');
+    this.withWrapped(alias, statementOrColumnList, nothingOrStatement);
+    this._statements[this._statements.length - 1].materialized = true;
+    return this;
+  }
+
+  withNotMaterialized(alias, statementOrColumnList, nothingOrStatement) {
+    validateWithArgs(alias, statementOrColumnList, nothingOrStatement, 'with');
+    this.withWrapped(alias, statementOrColumnList, nothingOrStatement);
+    this._statements[this._statements.length - 1].materialized = false;
+    return this;
+  }
+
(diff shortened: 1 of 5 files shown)


santacruz123:
#4948 related

santacruz123:
```
  1) Selects
       postgres
         MATERIALIZED CTE:
     TypeError: knex.withMaterialized is not a function
      at Context.<anonymous> (test/integration2/query/select/selects.spec.js:1108:41)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

Why it's not visible?  Should I build something special for integration test?

lib/query/querybuilder.js
```diff
@@ -124,6 +124,20 @@ class Builder extends EventEmitter {
     return this.withWrapped(alias, statementOrColumnList, nothingOrStatement);
   }
 
+  withMaterialized(alias, statementOrColumnList, nothingOrStatement) {
```
OlivierCavadenti, line 127: You implement the functionnality for all database, but only available PG and Sqlite3, so no need here.

lib/query/querycompiler.js
```diff
@@ -1160,6 +1160,16 @@ class QueryCompiler {
       this.client,
       this.bindingsHolder
     );
+
+    let sqlMaterialized = '';
+    const canMaterialized = this.client.config.client === 'pg';
```
OlivierCavadenti, line 1165: we dont do check like this in generic files.

OlivierCavadenti (commented):
BTW, I will push integration2 + doc tests on my PR today (add integration tests).

lib/query/querybuilder.js
santacruz123: Do you imply this should be added to PG dialect only?

kibertoad: it should throw an error on dialects which do not support it. so base method should throw, pg and sqlite should override it to actually do something useful

PR #4954, opened 2022-01-20
Ensure consistent `integer` vs `bigInteger` use in CRDB

### LINKED ISSUE #4953: [CockroachDB] No difference between `integer` and `bigInteger`
# Environment

Knex version: 0.95.15+
Database + version: CockroachDB v21.1.13
OS: macOS Monterrey

# Bug

Both `integer` and `bigInteger` both result in a `int8` column being created in CockroachDB, whereas Postgres creates them as `int4` and `int8` respectively. This becomes an issue, as `int8`s are always returned as a String by the `pg` connection library, causing unexpected incompatibilities between the database vendors.

```js
knex.schema.createTable('example', (table) => {
  table.increments();

  table.integer('example_int');
  table.bigInteger('example_bigint');
});
```

Note how both columns are created as an `int8` in CockroachDB.

I believe this is due to the fact that CockroachDB defaults to `int8` when using `.. ADD COLUMN example INTEGER`, which is [the default from here](https://github.com/knex/knex/blob/63980987a6f50529f37b96c127c05e6d87abc7d6/lib/schema/columncompiler.js#L282-L285) (AFAICT as the PG dialect doesn't have a specific `integer` in the columnbuilder).

rijkvanzanten:
Forces the CockroachDB dialect to use `int4` for `.integer()` use to align with Postgres and differentiate from `.bigInteger()`.

Fixes #4953

### DIFF
--- a/lib/dialects/cockroachdb/crdb-columncompiler.js
+++ b/lib/dialects/cockroachdb/crdb-columncompiler.js
@@ -1,0 +1,11 @@
+/* eslint max-len: 0 */
+
+const ColumnCompiler = require('../postgres/schema/pg-columncompiler');
+
+class ColumnCompiler_CRDB extends ColumnCompiler {
+  integer() {
+    return 'int4';
+  }
+}
+
+module.exports = ColumnCompiler_CRDB;
--- a/lib/dialects/cockroachdb/index.js
+++ b/lib/dialects/cockroachdb/index.js
@@ -6,6 +6,7 @@ const QueryCompiler = require('./crdb-querycompiler');
 const TableCompiler = require('./crdb-tablecompiler');
 const ViewCompiler = require('./crdb-viewcompiler');
 const QueryBuilder = require('./crdb-querybuilder');
+const ColumnCompiler = require('./crdb-columncompiler');
 
 // Always initialize with the "QueryBuilder" and "QueryCompiler"
 // objects, which extend the base 'lib/query/builder' and
@@ -21,6 +22,10 @@ class Client_CockroachDB extends Client_PostgreSQL {
 
   tableCompiler() {
     return new TableCompiler(this, ...arguments);
+  }
+
+  columnCompiler() {
+    return new ColumnCompiler(this, ...arguments);
   }
 
   viewCompiler() {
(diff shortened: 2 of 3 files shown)


kibertoad (approved):

kibertoad:
@OlivierCavadenti Technically this is a very breaking change for CockroachDB users since migrations will suddenly do a very different thing. Should we release this as 2.0.0?

OlivierCavadenti (commented):
Some tests breaks (int out of range, and others like : "accepts table names starting with numeric values" that need to swap "integer" to "int4" in SQL generated.)

rijkvanzanten:
> @OlivierCavadenti Technically this is a very breaking change for CockroachDB users since migrations will suddenly do a very different thing. Should we release this as 2.0.0?

Curious what you end up deciding for this. We often find ourselves debating this on the Directus side as well: Is it a Major SemVer breaking change release if you're patching a bug that corrects usage that was intended from the start? It technically is a breaking change for people relying on the broken behavior, but by that logic technically every bug fix could in turn be a breaking change for those relying on the wrong behavior that's corrected

> Some tests breaks

Yeah, my first test file approach was more a quick 'n dirty POC to see if you'd be interested in having it. Just pushed up 2 changes that should actually make it run properly 👍🏻

OlivierCavadenti:
> @OlivierCavadenti Technically this is a very breaking change for CockroachDB users since migrations will suddenly do a very different thing. Should we release this as 2.0.0?

Yeah, in other hands, users will continue to have migrations with this bug and 2.0.0 seems far ? Maybe an intermediate version with this breaking change can be released to prevent that earlier ?

kibertoad:
@OlivierCavadenti I don't mind releasing 2.0.0 early, semver seems more important for me than ensuring that every single major version has lots of stuff to it.
@rijkvanzanten In this case, imagine that you already ran your migrations in production and they generated tables of a different type. You update knex and suddenly migrations generate something different and your local environment is no longer consistent with what you have everywhere else. This is not the kind of a change that should be unleashed on users in semver minor.

rijkvanzanten:
> @OlivierCavadenti I don't mind releasing 2.0.0 early, semver seems more important for me than ensuring that every single major version has lots of stuff to it.
> @rijkvanzanten In this case, imagine that you already ran your migrations in production and they generated tables of a different type. You update knex and suddenly migrations generate something different and your local environment is no longer consistent with what you have everywhere else. This is not the kind of a change that should be unleashed on users in semver minor.

Totally! Fully agree.

rijkvanzanten:
Before we merge this, lets also look into the difference in `increments()` and `bigIncrements()`, as that seems to have the same problem (`serial` in CockroachDB defaults to `int8`)

PR #4965, opened 2022-01-24
bugfix: use logger.warn for redshift transaction dialect

stigerland:
The redshift transaction dialect previously assumed that the logger
was a function that could be used directly instead of using a specific
logger method. The logger interface might actually have been a single
function at some point in time, but now you at least need to use a
specific logger function indicating the appropriate log level.


stigerland:
I pushed a bit early. I'll reopen it with a test also.

PR #4967, opened 2022-01-25
Add alterType and update index.d.ts for alter function

OlivierCavadenti:
- Fix https://github.com/knex/knex/pull/4730/files#r791157891
- And I think we are good for an 1.0.2 (doc here : https://github.com/knex/documentation/pull/385)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2210,7 +2210,7 @@ export declare namespace Knex {
     notNullable(): ColumnBuilder;
     nullable(): ColumnBuilder;
     comment(value: string): ColumnBuilder;
-    alter(): ColumnBuilder;
+    alter(options : Readonly<{alterNullable?: boolean}>): ColumnBuilder;
     queryContext(context: any): ColumnBuilder;
     after(columnName: string): ColumnBuilder;
     first(): ColumnBuilder;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2210,7 +2210,7 @@ export declare namespace Knex {
     notNullable(): ColumnBuilder;
     nullable(): ColumnBuilder;
     comment(value: string): ColumnBuilder;
-    alter(options: Readonly<{alterNullable?: boolean}>): ColumnBuilder;
+    alter(options: Readonly<{alterNullable?: boolean, alterType?: boolean}>): ColumnBuilder;
     queryContext(context: any): ColumnBuilder;
     after(columnName: string): ColumnBuilder;
     first(): ColumnBuilder;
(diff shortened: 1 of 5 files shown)


OlivierCavadenti:
@intech do you think I can do something for cockroachdb test ? Or I need to revert for now ?

intech:
> @intech do you think I can do something for cockroachdb test ? Or I need to revert for now ?

Quickly studied the error and can not say how to get around it. I'll ask the cockroachdb engineers for advice and get back.

intech:
@OlivierCavadenti This column change is not yet supported https://github.com/cockroachdb/cockroach/issues/47636
And https://github.com/cockroachdb/cockroach/issues/49329#issuecomment-1022120446 for other test case

kibertoad:
wait, wasn't the whole idea of the change to avoid altering nullability? or crdb doesn't support such change even without toucing nullability?

intech:
@kibertoad I don't really understand the real application of this functionality, which I wrote about in the original issue on another project using knex.
Now the error occurs on this request:
```sql
alter table "primary_table_null" alter column "id_not_nullable_primary" type integer using ("id_not_nullable_primary"::integer);
```
Why do we need to change the column type to int if we already have int when creating the table?
```sql
create table "primary_table_null" ("id_not_nullable_primary" integer not null);
```

@OlivierCavadenti @kibertoad Could you please explain to me the application of these queries?

kibertoad:
@intech I think what happens there is that since knex does not know what is the current type of the column, it uses the one specified in alter command. To address that we probably need additional parameter alterType, which omits type change if set.

types/index.d.ts
```diff
@@ -2210,7 +2210,7 @@ export declare namespace Knex {
     notNullable(): ColumnBuilder;
     nullable(): ColumnBuilder;
     comment(value: string): ColumnBuilder;
-    alter(): ColumnBuilder;
+    alter(options: Readonly<{alterNullable?: boolean}>): ColumnBuilder;
```
kibertoad, line 2213: looks like we need another option `alterType`

kibertoad (approved):

types/index.d.ts
```diff
@@ -2210,7 +2210,7 @@ export declare namespace Knex {
     notNullable(): ColumnBuilder;
     nullable(): ColumnBuilder;
     comment(value: string): ColumnBuilder;
-    alter(): ColumnBuilder;
+    alter(options: Readonly<{alterNullable?: boolean, alterType?: boolean}>): ColumnBuilder;
```
B4nan, line 2213: this is breaking for TS users. the parameter was not there, now its required.

OlivierCavadenti: Yeah, sorry, just fixed and merged here : https://github.com/knex/knex/pull/4996

kibertoad: will publish hotfix shortly

PR #4976, opened 2022-01-27
move deferrable to after on update/on delete

bflyblue:
Postgresql requires that deferrable appear after on update/on delete, the current order fails with:

```
error: alter table "smtp_servers" add constraint "smtp_servers_email_channel_foreign" foreign key ("email_channel") references "email_channels" ("id") deferrable initially deferred  on update cascade on delete cascade - syntax error at or near "on"
```

Moving the deferred clause to the end works as intended.

### DIFF
--- a/test/integration2/query/insert/on-conflict-deferred.spec.js
+++ b/test/integration2/query/insert/on-conflict-deferred.spec.js
@@ -9,7 +9,7 @@ describe('Insert', () => {
     getAllDbs().forEach((db) => {
       describe(db, () => {
         let knex;
-        before(() => {
+        before(function () {
           knex = getKnexForDb(db);
           if (!isPostgreSQL(knex)) {
             this.skip('This test is PostgreSQL only');
--- a/test/integration2/query/insert/on-conflict-deferred.spec.js
+++ b/test/integration2/query/insert/on-conflict-deferred.spec.js
@@ -1,0 +1,76 @@
+const {
+  getAllDbs,
+  getKnexForDb,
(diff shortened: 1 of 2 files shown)


kibertoad:
can you add a test that fails without this change?

bflyblue:
@kibertoad I'm new to the ecosystem so the test might not be great, but I hope it's sufficient to demonstrate the issue.

It fails with `error: alter table "table_a" add constraint "table_a_b_id_foreign" foreign key ("b_id") references "table_b" ("id") deferrable initially deferred  on update cascade on delete cascade - syntax error at or near "on"` before the change.

test/integration2/query/insert/on-conflict-deferred.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+describe('Insert', () => {
+  describe('onConflict deferred', () => {
+    getAllDbs().forEach((db) => {
+      describe(db, () => {
+        let knex;
+        before(() => {
+          knex = getKnexForDb(db);
+          if (!isPostgreSQL(knex)) {
+            this.skip('This test is PostgreSQL only');
```
kibertoad, line 15: this.skip is not available in arrow function, you'd need to change it to `function()`

bflyblue: Thank you, I learned something new today. :)

kibertoad (approved):

kibertoad:
Released in 1.0.2

PR #4998, opened 2022-02-03
Fix type definition for getMigration in MigrationSource

akg1:
FsMigrations returns a promise, and all calling code expects a promise, so this updates the type definition to match that.


OlivierCavadenti (approved):

kibertoad:
Released in 1.0.3

PR #5005, opened 2022-02-07
Use async / await syntax in seeds as default

vencho-mdp:


lib/migrations/seed/stub/js.stub
```diff
@@ -2,15 +2,13 @@
...
-        {id: 3, colName: 'rowValue3'}
-      ]);
-    });
+  await knex('table_name').del()
+  await knex('table_name').insert([
+    {id: 1, colName: 'rowValue1'},
+    {id: 2, colName: 'rowValue2'},
+    {id: 3, colName: 'rowValue3'}
+  ]);
+  return knex
```
kibertoad, line 13: is returning knex even necessary here?

vencho-mdp: Good point, I don't think it is, but I'm not really sure

kibertoad: Original was just returning the promise from seed, so this should not be needed.

kibertoad (approved):

kibertoad:
thank you!

kibertoad:
Released in 1.0.3

PR #5009, opened 2022-02-07
Add withMaterialized and withNotMaterialized to method-constants

### LINKED ISSUE #5008: withMaterialized and withNotMaterialized are not present on knex object
# Environment

Knex version: 1.0.2
Database + version: postgres 12 and 13
OS: macOS and Linux

# Bug

The new `.withMaterialized` and `.withNotMaterialized` functions added in #4470 are not present on the `knex` object, despite the example in the docs showing them used this way.

The [docs page](https://knexjs.org/#Builder-withMaterialized) shows an error for these methods as well

2. Error message
`knex.withMaterialized is not a function`

See [the example on the docs page](https://knexjs.org/#Builder-withMaterialized):
```
knex.withMaterialized('with_alias', knex.raw('select * from "books" where "author" = ?', 'Test')).select('*').from('with_alias')
Error:
knex.withMaterialized is not a function
```

The solution is to add `withMaterialized` and `withNotMaterialized` to method-constants in https://github.com/knex/knex/blob/master/lib/query/method-constants.js

jhiesey:
Fixes #5008

### DIFF
--- a/lib/query/method-constants.js
+++ b/lib/query/method-constants.js
@@ -3,6 +3,8 @@
 module.exports = [
   'with',
   'withRecursive',
+  'withMaterialized',
+  'withNotMaterialized',
   'select',
   'as',
   'columns',


OlivierCavadenti (approved):
thanks, @kibertoad go for 1.0.3 or you think about another thing to add ?

PR #5011, opened 2022-02-08
Fix whereJsonPath queries for postgres and cockroachdb

stvsh:
This PR fixes 2 bugs I've found working with where clauses build with 'whereJsonPath' for postgres database:

1.  If a passed value is a number, it is casted to integer or float on postgres side. However, is casted also if value only starts with a number but contains also letters or other characters. This is because **parseInt** and **parseFloat** are used for value check and they just ignore all non-numeric characters.
2. Queries using 'orWhereJsonPath' don't have proper binding for value. 

For 1st I've added additional check if value is not a NaN and for 2nd I've fixed arguments passed in query builder method.

### DIFF
--- a/lib/dialects/cockroachdb/crdb-querycompiler.js
+++ b/lib/dialects/cockroachdb/crdb-querycompiler.js
@@ -35,9 +35,9 @@ class QueryCompiler_CRDB extends QueryCompiler_PG {
 
   whereJsonPath(statement) {
     let castValue = '';
-    if (parseInt(statement.value)) {
+    if (!isNaN(statement.value) && parseInt(statement.value)) {
       castValue = '::int';
-    } else if (parseFloat(statement.value)) {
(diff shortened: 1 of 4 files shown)


stvsh:
I'm a little confused why 'orWhereJsonPath' doesn't work for some dialects (sqlite, mysql) after my changes. I'll try to figure this out and create a new PR (or I'll report a bug if couldn't) so this one include fix only for 1st issue I've mentioned ;)

OlivierCavadenti:
> I'm a little confused why 'orWhereJsonPath' doesn't work for some dialects (sqlite, mysql) after my changes. I'll try to figure this out and create a new PR (or I'll report a bug if couldn't) so this one include fix only for 1st issue I've mentioned ;)

Ok, I will merge this one. Can you make an issue for the second point ? I am take a look.

OlivierCavadenti (approved):

kibertoad:
Released in 1.0.3

PR #5021, opened 2022-02-12
[types] Make Knex's default generic parameter types be assignable to knex's (TypeScript 4.6)

ide:
The default generic parameters of `Knex` didn't match the generic parameter types of `knex`. This becomes relevant with TypeScript 4.6 for `k: Knex = knex()` to work.

We can't change `knex`'s declaration from `TResult = unknown[]` to `TResult = Record<string, any>[]` because other types like `DeferredKeySelection.ReplaceBase` specifically look for `unknown`. Instead, this commit changes `Knex` to use `TResult = any[]`.

This works with TypeScript 4.6. Added both tsd and dtslint tests (tsd doesn't support TS 4.6 yet). Also tested this change in a larger codebase and confirmed that nearly all TS 4.6 errors were addressed.

Follow-up to https://github.com/knex/knex/pull/4933.

### DIFF
--- a/test-tsd/types.test-d.ts
+++ b/test-tsd/types.test-d.ts
@@ -19,6 +19,18 @@ expectType<Knex<any, unknown[]>>(knexCjsImport.knex({}));
 expectType<KnexTimeoutError>(new knex.KnexTimeoutError());
 expectType<KnexTimeoutError>(new knex.KnexTimeoutError());
 
+// Knex instances need to be assigned first so their generic types aren't inferred
+const k1 = knex({});
+expectAssignable<Knex>(k1);
+const k2 = knexStar.default({});
+expectAssignable<Knex>(k2);
+const k3 = knexStar.knex({});
+expectAssignable<Knex>(k3);
+const k4 = knexCjsImport.default({});
+expectAssignable<Knex>(k4);
+const k5 = knexCjsImport.knex({});
+expectAssignable<Knex>(k5);
+
 // eslint-disable-next-line
 expectType<any>(knexCjs({}));
 // eslint-disable-next-line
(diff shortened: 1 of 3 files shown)


kibertoad:
Can you add a tsd test for this?

kibertoad:
@ide Do you think it's worth it to preserve unknown in higher-order types? Wonder if it brings value over Record.

ide:
I don't have a strong opinion on that since I don't directly use Knex's generic parameters (my data access layer above Knex is strictly typed, and the interface between this layer and Knex is covered by tests). My original motivation was to make `k: Knex = knex({})` work with the TS 4.6 release candidate.

ide:
Revisiting this since TypeScript 4.6.2 has been released. I understand if you are on the fence and don't want to commit to taking this PR as the maintainer of the repo.

kibertoad:
@OlivierCavadenti thoughts?

kibertoad (approved):

KoltesDigital:
Thanks for the PR, I was having the same problem and started a fork too before seeing this one.

I've introduced two types `DefaultRecord` and `DefaultResult` to unify these types across the typings file. I think this is still relevant. But I'm wondering what should these types be?

I don't know all API functions, but if `TResult` is supposed to always be an object, they should all be changed to `Record<string, any>`, instead of reverting this one back to `any`.

Besides, TypeScript doc encourages the use of `unknown` instead of `any`. This should not be a problem for users as they should override `knex/types/tables`. This could even encourage the ones which didn't override to do it. At least, it took a long time for me to find out the override ability, and maybe I would have found it sooner if the compiler was blocking me instead of letting me freely.

Thoughts?

kibertoad:
Released in 1.0.5

PR #5022, opened 2022-02-12
Fix orWhereJsonPath clause

OlivierCavadenti:
- Closes https://github.com/knex/knex/issues/5020

### DIFF
--- a/lib/query/querybuilder.js
+++ b/lib/query/querybuilder.js
@@ -1505,8 +1505,8 @@ class Builder extends EventEmitter {
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


OlivierCavadenti:
> this is not a breaking change?

No, a parameter is missing in whereJsonPath so orWhereJsonPath doesn't work.

kibertoad:
are types correct?

OlivierCavadenti:
> are types correct?

Yes, interfaces are the same for all 3 functions: 

image

kibertoad:
Released in 1.0.4

PR #5030, opened 2022-02-19
Rework Union Wrapping

OlivierCavadenti:
- Rework union construction to work with order by, limit, offset, etc.
- Before this PR, union is done before group by, limit, ... so statements position make syntax errors.
- With this PR, union is done after other statements but, as exposed in https://github.com/knex/knex/issues/913#issuecomment-334481807, is not only an order problem.
I also wrapped all queries in parenthesis to avoid syntax error  in PG.
By the way, reordering components (as intended in the issue) dont work also because the order of bindings is not changed.
- Closes https://github.com/knex/knex/issues/913
- Closes https://github.com/knex/knex/issues/5047


OlivierCavadenti:
I change wrap logic (main query is wrapped is union query are wrapped) but I still need to investigate how to generate valid SQL with limit, group by, etc. in unions (not so trivial).

kibertoad (approved):

PR #5031, opened 2022-02-21
Fix types for "returning" methods

### LINKED ISSUE #4942: Failing types on "returning" methods
I recently updated to the latest version of knex and I faced what I think is one of the breaking changes from the release notes:

> Changed data structure from RETURNING operation to be consistent with SELECT;

```
const ids =  await knex.from("table").where(...).update({...}, "id");
```
used to return a `string[]` but now returns a `{id: string}[]`, fair enough.

However some of the overloads still reflect the previous behavior. I managed to fix it with patch-package with the following diff but I'm not 100% this is the correct solution.

Am I on the right track?

```diff
diff --git a/node_modules/knex/types/index.d.ts b/node_modules/knex/types/index.d.ts
index d9bf6e5..b50c579 100644
--- a/node_modules/knex/types/index.d.ts
+++ b/node_modules/knex/types/index.d.ts
@@ -692,7 +692,7 @@ export declare namespace Knex {
     ): QueryBuilder<TRecord, DeferredKeySelection<TRecord, never>[]>;
     insert<
       TKey extends StrKey<ResolveTableType<TRecord>>,
-      TResult2 = DeferredIndex.Augment<
+      TResult2 = DeferredKeySelection.Augment<
         UnwrapArrayMember<TResult>,
         ResolveTableType<TRecord>,
         TKey
@@ -720,7 +720,7 @@ export declare namespace Knex {
     ): QueryBuilder<TRecord, TResult2>;
     insert<
       TKey extends string,
-      TResult2 = DeferredIndex.Augment<
+      TResult2 = DeferredKeySelection.Augment<
         UnwrapArrayMember<TResult>,
         TRecord,
         TKey
@@ -828,7 +828,7 @@ export declare namespace Knex {
     update<
       K1 extends StrKey<ResolveTableType<TRecord, 'update'>>,
       K2 extends StrKey<ResolveTableType<TRecord>>,
-      TResult2 = DeferredIndex.Augment<
+      TResult2 = DeferredKeySelection.Augment<
         UnwrapArrayMember<TResult>,
         ResolveTableType<TRecord>,
         K2
@@ -870,7 +870,7 @@ export declare namespace Knex {
     ): QueryBuilder<TRecord, DeferredKeySelection<TRecord, never>[]>;
     update<
       TKey extends StrKey<ResolveTableType<TRecord>>,
-      TResult2 = DeferredIndex.Augment<
+      TResult2 = DeferredKeySelection.Augment<
         UnwrapArrayMember<TResult>,
         ResolveTableType<TRecord>,
         TKey
@@ -925,7 +925,7 @@ export declare namespace Knex {
     returning(column: '*', options?: DMLOptions): QueryBuilder<TRecord, DeferredKeySelection<TRecord, never>[]>;
     returning<
       TKey extends StrKey<ResolveTableType<TRecord>>,
-      TResult2 = DeferredIndex.Augment<
+      TResult2 = DeferredKeySelection.Augment<
         UnwrapArrayMember<TResult>,
         ResolveTableType<TRecord>,
         TKey
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

santialbo:
Closes #4942
With the release of v1, the methods for `returning` changed a bit. However, the type definitions were not updated.

Reviewing the rest of the types I found that the `onConflict` clause was setting the Result type with the passed conflicting columns. This looked very strange and at least when using postgres as the backend, this is not the case.

Also, (this is another issue) when not using the "returning", the types specify that the result is `number[]`. However (at least using pg) a Result object is returned instead.

```ts
  const users = await knex
    .from<Contact>("contact").insert([{
      email: "xxxx@gmail.com",
      first_name: "jeff"
    }])
  // users is suposed to be a number[];
  console.log(users)
  // Result {
  //   command: 'INSERT',
  //   rowCount: 2,
  //   oid: 0,
  //   rows: [],
  //   fields: [],
  //   ...
  // }
```

### DIFF
--- a/test-tsd/types.test-d.ts
+++ b/test-tsd/types.test-d.ts
@@ -94,12 +94,7 @@ expectAssignable<QueryBuilder>(
     .debug(true)
 );
 
-expectType<
-  QueryBuilder<
-    User,
-    DeferredKeySelection<User, 'id', true, {}, true, {}, never>[]
-  >
->(
+expectAssignable<QueryBuilder>(
   knexInstance
     .table<User>('users')
     .insert({ id: 10, active: true })
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -284,27 +284,6 @@ type AggregationQueryResult<TResult, TIntersectProps2> = ArrayIfAlready<
   : TIntersectProps2
 >;
 
-// Convenience alias and associated companion namespace for working
-// with DeferredSelection having TSingle=true.
-//
-// When TSingle=true in DeferredSelection, then we are effectively
-// deferring an index access operation (TBase[TKey]) over a potentially
-// unknown initial type of TBase and potentially never initial type of TKey
-
-type DeferredIndex<TBase, TKey extends string> = DeferredKeySelection<TBase, TKey, false, {}, true>;
-
-declare namespace DeferredIndex {
-  type Augment<
-    T,
-    TBase,
-    TKey extends string,
-    TAliasMapping = {}
-    > = DeferredKeySelection.SetSingle<
-      DeferredKeySelection.AddKey<DeferredKeySelection.SetBase<T, TBase>, TKey>,
-      true
-    >;
-}
-
 // If we have more categories of deferred selection in future,
 // this will combine all of them
 type ResolveResult<S> = DeferredKeySelection.Resolve<S>;
@@ -694,7 +673,7 @@ export declare namespace Knex {
     ): QueryBuilder<TRecord, DeferredKeySelection<TRecord, never>[]>;
     insert<
       TKey extends StrKey<ResolveTableType<TRecord>>,
-      TResult2 = DeferredIndex.Augment<
+      TResult2 = DeferredKeySelection.Augment<
         UnwrapArrayMember<TResult>,
         ResolveTableType<TRecord>,
         TKey
@@ -722,7 +701,7 @@ export declare namespace Knex {
     ): QueryBuilder<TRecord, TResult2>;
     insert<
       TKey extends string,
-      TResult2 = DeferredIndex.Augment<
+      TResult2 = DeferredKeySelection.Augment<
         UnwrapArrayMember<TResult>,
         TRecord,
         TKey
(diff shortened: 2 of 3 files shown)


test-tsd/types.test-d.ts
```diff
@@ -94,17 +94,13 @@ expectAssignable<QueryBuilder>(
...
-    DeferredKeySelection<User, 'id', true, {}, true, {}, never>[]
-  >
->(
+expectAssignable<QueryBuilder>(
   knexInstance
     .table<User>('users')
     .insert({ id: 10, active: true })
     .onConflict('id')
     .ignore()
+    .debug(true)
```
kibertoad, line 103: why was debug added?

kibertoad:
Type tests are failing: https://github.com/knex/knex/runs/5271463571?check_suite_focus=true

test-tsd/types.test-d.ts
santialbo: it was a copy paste thing from the previous test, it's removed

santialbo:
> Type tests are failing: https://github.com/knex/knex/runs/5271463571?check_suite_focus=true

fixed, sorry I missed those.

test-tsd/types.test-d.ts
```diff
@@ -94,12 +94,7 @@ expectAssignable<QueryBuilder>(
     .debug(true)
 );
 
-expectType<
-  QueryBuilder<
-    User,
-    DeferredKeySelection<User, 'id', true, {}, true, {}, never>[]
-  >
->(
+expectAssignable<QueryBuilder>(
```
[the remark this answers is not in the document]
santialbo: It is possible but the type would be "wrong" anyways due to the issue that I mentioned before.
Insert/update methods without "returning" still have pre v1 signature of returning a `number[]` but they are not actually returning that. They return a "Result" object. I'm not sure if this is something that happens with just pg or it happens with every backend db.
```
  const users = await knex
    .from<Contact>("contact").insert([{
      email: "xxxx@gmail.com",
      first_name: "jeff"
    }])
  // users is suposed to be a number[];
  console.log(users)
  // Result {
  //   command: 'INSERT',
  //   rowCount: 2,
  //   oid: 0,
  //   rows: [],
  //   fields: [],
  //   ...
  // }
  ```

kibertoad (approved):

kibertoad:
thank you!

KristjanTammekivi:
Can we have this released? It's a bit of a blocker for our knex upgrade

kibertoad:
I will release it tomorrow

esemeniuc:
This is still broken on `knex.raw()`

PR #5039, opened 2022-02-26
Add cast functions to support 'cast as' statements

### LINKED ISSUE #501: How to cast types in SELECT
```
knex('users').sum('products')

Outputs:
select sum(`products`) from `users`
```

How do you cast types in this instance, e.g.:

```
select sum(`products`)::BIGINT from `users`
```

### LINKED ISSUE #2587: Expose knex.cast function

To be able to cast values and references to specific types.

Exposing a new function (or rather a set of functions)

Disclaimer: Mostly inspiried by objectionjs as per original discussion in https://github.com/tgriesser/knex/issues/2507

```javascript
knex.cast(value, type);
knex.castText(value);
knex.castInt(value);
knex.castBigInt(value);
knex.castFloat(value);
knex.castDecimal(value);
knex.castReal(value);
knex.castBool(value);
knex.castJson(value);
knex.cast*(value).as(alias)
```

Intended to support any "valid-knex" input:
```javsacript
knex.castText(1); //CAST '1' AS text
knex.castText(knex.raw(1)); //CAST 1 AS text
knex.castText(knex.ref('id')); // CAST "id" AS text
```

Example implementation
```javascript
'use strict';
import Raw from './raw';

class Cast extends Raw {
  constructor(client) {
    super(client);

    this.value = null;
    this.to    = '';
    this.alias = '';
  }

  cast(value, to) {
    this.value = value;
    this.to    = to;

    return this;
  }

  as(alias) {
    this.alias = alias;

    return this;
  }

  static get TYPES() {
    return {
      Text:    'text',
      Integer: 'integer',
      BigInt:  'bigint',
      Float:   'float',
      Decimal: 'decimal',
      Real:    'real',
      Bool:    'boolean',
      Json:    'json',
    };
  }

  castText(value) {
    return this.cast(value, Cast.TYPES.Text);
  }

  castInt(value) {
    return this.cast(value, Cast.TYPES.Integer);
  }

  castBigInt(value) {
    return this.cast(value, Cast.TYPES.BigInt);
  }

  castFloat(value) {
    return this.cast(value, Cast.TYPES.Float);
  }

  castDecimal(value) {
    return this.cast(value, Cast.TYPES.Decimal);
  }

  castReal(value) {
    return this.cast(value, Cast.TYPES.Real);
  }

  castBool(value) {
    return this.cast(value, Cast.TYPES.Bool);
  }

  castJson(value) {
    return this.cast(value, Cast.TYPES.Json);
  }

  toSQL() {
    const {value, to, alias} = this;

    const asAlias = alias ? `AS ${this.client.formatter(this).wrap(alias)}` : '';

    const cast = to === Cast.TYPES.Json
      ? `to_jsonb(?) ${asAlias}`
      : `CAST(? AS ${to}) ${asAlias}`;

    this.set(cast, [value]);

    return super.toSQL(...arguments);
  }
}

export default Cast
```

Extending `Raw` _(same as `Ref`)_ makes it a valid input across the knex API and adds automatic support for existing `Raw` functions such as `.wrap`.

Creating this for discussion purposes, I'm not sure what would be the most optimal API or implementation, but I believe this is something knex currently lacks.

OlivierCavadenti:
- First draft to add knex.cast functions on Knex, like knex.raw and knex.ref but needs some work to be finished (I push to have comments, if any).
- Postgres integration test is ok, other dialects seems have issues. Need to check:
    - [x] Sqlite3
    - [x] OracleDB
    - [x] MSSQL
    - [x] Mysql
    - [x] CoachroachDB
- [x] Index.d.ts
- [ ] Documentation
- Closes https://github.com/knex/knex/issues/2587
- Closes https://github.com/knex/knex/issues/501


lib/cast.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  toString() {
+    return this.toQuery();
+  }
+
+  _getColType(type) {
+    let columnCompiler;
+
+    switch (this.client.config.client) {
```
kibertoad, line 36: why not implement getCompiler method on this.client instead? not sure we don't have one yet, actually

lib/client.js
```diff
@@ -141,6 +143,47 @@ class Client extends EventEmitter {
   ref() {
     return new Ref(this, ...arguments);
   }
+
+  cast(value, to, alias) {
+    return new Cast(this).cast(value, to, alias);
```
kibertoad, line 148: if new Cast is only supposed to be called from client, then you can actually pass correct compiler from here as a constructor param

OlivierCavadenti: I try quickly but it's seems I cant use columnBuilder function because it's depends of args from tableBuilder, type, args. But maybe you want getCompiler function (overrided in each Client) that return the columnCompiler prototype ?

OlivierCavadenti:
I have strange behavior with MSSQL when I cast binary:

image

Can't explain that, if anyone have clues...

OlivierCavadenti:
> I have strange behavior with MSSQL when I cast to binary:
> 
> image
> 
> Can't explain that, if anyone have clues...
> 
> (cast to varbinary(max))

this is quite complex....

MSSQL drivers bind string parameters as nvarchar, which uses UTF-16LE encoding (2 bytes per ASCII character). When directly casting to varbinary(max), the raw UTF-16 bytes are preserved...... and resulting in null bytes (0x00) interleaved between each character byte !

The workaround is a double cast: first to varchar(max) (single-byte encoding), then to varbinary(max):
cast(cast(? as varchar(max)) as varbinary(max))

https://learn.microsoft.com/en-us/sql/t-sql/data-types/binary-and-varbinary-transact-sql?view=sql-server-ver17
https://gist.github.com/sevaa/f084a0a5a994c3bc28e518d5c708d5f6

lib/cast.js
```diff
@@ -0,0 +1,116 @@
...
+    };
+  }
+}
+
+// Workaround to avoid circular dependency between wrappingFormatter.unwrapRaw and rawFormatter
+Cast.prototype.isRawInstance = true;
+
+// Allow the `Cast` object to be utilized with full access to the relevant
+// promise API.
+augmentWithBuilderInterface(Cast);
```
tgriesser: Why does `Cast` require the full promise API?

lib/cast.js
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  toString() {
+    return this.toQuery();
+  }
+
+  _getColType(type) {
+    const columnCompiler = this.client.getColumnCompiler();
+    const typeName =
+      columnCompiler.prototype[type] ?? columnCompiler.prototype[type + 'eger'];
```
tgriesser:
```suggestion
columnCompiler.prototype[type + 'eger'];
```
Feels kinda odd, I didn't see anywhere that this is doing `_getColType('int')` which is what I presume this is attempting to do?

tgriesser (commented):
Nice idea, a few comments

lib/cast.js
OlivierCavadenti: Good catch, it doesn't. `Cast` extends `Raw`, and `Raw` already calls `augmentWithBuilderInterface` / `addQueryContext` and sets `isRawInstance`, so re-applying them here was pure redundancy. I removed them.

lib/cast.js
OlivierCavadenti: You're right that it's confusing ! fixed

tgriesser (approved):

lib/dialects/mysql/index.js
```diff
@@ -344,6 +349,32 @@ class Client_MySQL extends Client {
...
+  }
+
+  castBinary(value, alias) {
+    return new Cast(this).cast(value, 'binary', alias);
+  }
+
+  castText(value, alias) {
+    throw new Error(
+      `Cast to Test is not possible in MySQL, use castChar function instead.`
+    );
```
tgriesser: Should we just have this `castChar` - would seem more consistent with knex's behavior elsewhere, to sort of smooth over cross-dialect inconsistencies

types/index.d.ts
```diff
@@ -3365,6 +3376,7 @@ declare namespace Knex {
     runner(builder: any): any;
     transaction(container: any, config: any, outerTx: any): Transaction;
     raw(...args: any[]): any;
+    castText(...args: any[]): any;
```
tgriesser:
```suggestion
```

Do we want to put this on the `client` interface? If so should we add all of the `cast*` fns?

tgriesser (changes requested):

lib/dialects/mysql/index.js
OlivierCavadenti: Agreed, that's more consistent. castText now produces a CHAR cast on MySQL instead of throwing, so it behaves like the other dialects.

types/index.d.ts
OlivierCavadenti: Removed it. The cast helpers are already typed on the public interface around line 2429, so castText alone on the internal Client type was redundant.

test/unit/query/builder.js
```diff
@@ -12208,4 +12208,125 @@ describe('QueryBuilder', () => {
...
+        sqlite3: 'select cast(`age` as integer) as `val` from `t`',
+      });
+    });
+
+    it('cast without alias', () => {
+      testCastSql((c) => c.castInt('42'), {
+        pg: 'select cast(? as integer) from "t"',
+        mysql: 'select cast(? as unsigned) from `t`',
+        mssql: 'select cast(? as int) from [t]',
+      });
```
tgriesser, line 12389: Can we add `mariadb` to the tests here?

OlivierCavadenti: Added mariadb to the cast tests. It matches mysql except for castJson and castJsonb, which map to char, and castReal, which maps to double, reflecting the mariadb overrides.

PR #5042, opened 2022-03-01
Add missing reject on stream error (Closes #4453)

### LINKED ISSUE #4453: Stream method does not throw on error
# Environment

Knex version: 0.95.4
Database + version: 
OS: ubuntu (popos)

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do;

I maintain mock-knex module and on this update to 0.92, one of my tests started to fail;

```js
it("should catch errors on stream", (done) => {
      tracker.on("query", (query) => {
        throw new Error("Third Error");
      });

      db.select("columnA", "columnB", "columnC")
        .from("field")
        .where({
          columnA: true,
        })
        .stream(_.noop)
        .then(function () {
          console.log(arguments);
        })
        .catch((err) => {
          expect(err).to.be.an.instanceof(Error);
          expect(err.message).to.equal("Third Error");
          done();
        });
    });
```

In this test the catch is never executed, from the screenshot below it seems that it is caught but never re-thrown to allow it to bubble up.   Looking at the documentation it looks like streaming queries should throw if there is an error. I can confirm the stream error event is emitted and in my abstraction the promise is rejected properly.

image

danielgindi:


OlivierCavadenti (commented):
Hi, thx for the PR.
Can you add test for this ?

danielgindi:
> Hi, thx for the PR. Can you add test for this ?

Added.

However, I'm not entirely sure that the issue is actually an issue. It depends on semantics:

1. Does the `Promise` returned by `stream(cb)` actually needs to reject on a stream error?
2. Does the `Promise` mean "wait for acquiring the stream" or "wait for stream completion"?
3. If this PR is not merged, then people will still have to wrap the call to `stream` with a `Promise`, to avoid a forever-pending Promise.  
4. If the PR does get merge, then if people call the `stream` function with a callback but without awaiting the result - they may get a warning about an unhandled rejection.

kibertoad:
can we add a check to avoid unhandled rejection if callback is provided?

danielgindi:
> can we add a check to avoid unhandled rejection if callback is provided?

I think you missed the point. Hi his function only returns a Promise if callback *is* provided.
So yes, the semantics here are confusing from the beginning.

AlexandrMov:
Hi there!

Any updates here? Was the issue solved?

PR #5043, opened 2022-03-01
Destroy should not trigger an error (Closes #4132)

### LINKED ISSUE #4132: Error: stream destroyed

danielgindi:
…ed by the transform.

Closes #4132

### DIFF
--- a/lib/execution/runner.js
+++ b/lib/execution/runner.js
@@ -61,22 +61,12 @@ class Runner {
     Transform = Transform || require('stream').Transform;
 
     const queryContext = this.builder.queryContext();
-    let queryStream;
 
     const stream = new Transform({
       objectMode: true,
       transform: (chunk, _, callback) => {
         callback(null, this.client.postProcessResponse(chunk, queryContext));
-      },
-      destroy() {
-        // For some reason destroy is not available for mssql on Node 14. Might be a problem with tedious: https://github.com/tediousjs/tedious/issues/1139
-        if (queryStream && queryStream.destroy) {
-          queryStream.destroy(new Error('stream destroyed'));
-        }
-      },
-    });
-    stream.on('pipe', (qs) => {
-      queryStream = qs;
+      }
     });
 
     const connectionAcquirePromise = this.ensureConnection(


kibertoad:
@dhensby Would it be safe to do so?

kibertoad:
@OlivierCavadenti @elhigu Could use second opinion

dhensby:
Hey, I don't have a super in-depth knowledge around this with tedious.

Though from my understanding of how streams work, it can be removed.

danielgindi:
> Hey, I don't have a super in-depth knowledge around this with tedious.
> 
> Though from my understanding of how streams work, it can be removed.

Not only "can", but should. Destroy should not throw an error. It is a normal node event. There's an "error" event that is fired when there's an error. "destroy" is not that...

dprgarner:
This commit has been cherry-picked and merged in https://github.com/knex/knex/pull/5243 .

PR #5044, opened 2022-03-01
Add whereLike functions

### LINKED ISSUE #2265: Database agnostic where ILIKE

OlivierCavadenti:
- Add andWhereLike / andWhereILike / orWhereLike / orWhereILike.
- Closes #2265


kibertoad (approved):

kibertoad:
we probably also needs docs?

PR #5055, opened 2022-03-11
createTableLike callback should be optional

tharrison4815:
As per the documentation the callback is supposed to be optional. Leaving the callback as undefined works fine functionally but TypeScript marks it as invalid.

This update is to correct the types file.


kibertoad (approved):

kibertoad:
Thank you!

kibertoad:
Released in 1.0.4

PR #5060, opened 2022-03-14
Fix completed migration listing

### LINKED ISSUE #4951: npx knex migrate:list doesn't print a readable list of completed migrations files
# Environment

Knex version: 1.0.1
Database + version: PostgreSQL 12.7
OS: macOS 11.6.2

# Bug

1.

```shell
npx knex migrate:list
```

produces 

```shell
[object Object]
[object Object]
[object Object]
No Pending Migration files Found.
```

Previous version (0.95) listed the filenames of completed migrations:

```shell
20220119154150_x.ts 
20220119155840_y.ts 
20220120081602_z.ts 
No Pending Migration files Found.
```

2. Error message

No error message, just a bit inconvenient to read :) 

Run `npx knex migrate:list` with at least one completed migration.

I think the fix is to say `completed[i].name` here https://github.com/knex/knex/blob/1744c8c2655a8362260ba987d706ec3473681a78/bin/utils/migrationsLister.js#L16

santialbo:
Fixes #4951

### DIFF
--- a/bin/utils/migrationsLister.js
+++ b/bin/utils/migrationsLister.js
@@ -6,7 +6,7 @@ function listMigrations(completed, newMigrations) {
   let message = '';
 
   if (completed.length === 0) {
-    message += color.red('No Completed Migration files Found. \n');
+    message += color.red('No Completed Migration files Found.\n');
   } else {
     message = color.green(
       `Found ${completed.length} Completed Migration file/files.\n`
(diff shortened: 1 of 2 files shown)


OlivierCavadenti (commented):
thx for the PR. No tests are changed ? Can you add one that fail before the change and success after ?

santialbo:
> thx for the PR. No tests are changed ? Can you add one that fail before the change and success after ?

Done! Updated an existing test for better coverage. This test without the change would have failed.

kibertoad (approved):

kibertoad:
Thank you!

Issue #4447, opened 2021-04-19
Cannot import ESM modules when using TypeScript


vvo:
# Environment

Knex version: 0.95.4
Database + version: sqlite3 5.0.2
OS: macOS Big Sur 11.1

@lorefnon (since related to TypeScript)

Hi there, I spent quite a bunch of time on this without finding the proper way to resolve it.

# Bug

1. Explain what kind of behavior you are getting and how you think it should do

When trying to `import randomInt from "random-int"` from a TypeScript migration file we get an error.

2. Error message

```txt
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/[redacted]/knex-typescript-migrations-esm/node_modules/random-int/index.js
require() of ES modules is not supported.
require() of /Users/[redacted]/knex-typescript-migrations-esm/node_modules/random-int/index.js from /Users/[redacted]/knex-typescript-migrations-esm/migrations/20210419131555_test.ts is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/[redacted]/knex-typescript-migrations-esm/node_modules/random-int/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1080:13)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/Users/[redacted]/knex-typescript-migrations-esm/migrations/20210419131555_test.ts:2:1)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Module.m._compile (/Users/[redacted]/knex-typescript-migrations-esm/node_modules/ts-node/src/index.ts:1056:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/[redacted]/knex-typescript-migrations-esm/node_modules/ts-node/src/index.ts:1059:12)
```

Running with --esm sort of works but breaks TypeScript:

```txt
export async function up(knex: Knex): Promise<void> {
                             ^

SyntaxError: Invalid or unexpected token
[...]

PR #5071, opened 2022-03-18
Fix isModuleType() for yarn (#4447)

lehni:
This relates to the issues #4447, #4295, #4158

I encounter the same problem as #4447 in an ESM project that uses yarn instead of NPM.

~~My proposed PR keeps the check for `process.env.npm_package_type` because it's faster than `await getPackageType(filepath)) === 'module'` and to avoid possible breakage on projects that somehow rely on it, but in a future one could move entirely away from that and it should still work.~~

**Update:** I found out it's best to not rely on `process.env.npm_package_type` at all and use a very simple, sturdy implementation that also solves more scenarios in which `--esm` previously lead to failures:

```js
async function isModuleType(filepath) {
  return (
    filepath.endsWith('.mjs') ||
    (!filepath.endsWith('.cjs') &&
      (await getPackageType(filepath)) === 'module')
  );
};
```


kibertoad:
@lehni Some unit tests are failing

lehni:
@kibertoad I'm looking into it right now. As far as I can tell by now, it looks like my change makes the `--esm` CLI argument redundant and correctly figures out how to import files without it. So the tests that are expected to fail don't fail anymore.

How shall we deal with it?

lehni:
@kibertoad it doesn't make `--esm` fully redundant, but it does make more situations work, which I've now reflected in the updated tests. Please take a look.

lehni:
@kibertoad all green! 🎉

kibertoad (approved):

kibertoad:
Thanks a lot!

lehni:
@kibertoad you're welcome, happy to help! Thank you for merging so quickly!

andrewchilds:
Hi @lehni and @kibertoad! After upgrading Knex from 1.0.2 to 1.0.7, I was having trouble running migrations locally, and tracked it down to this PR. I use yarn, ESM, and `.js` file extensions in my project.

Here's the command I've been using:

```
node_modules/knex/bin/cli.js migrate:up --esm --knexfile ./path/to/knexfile.js
```

Knex would evaluate the next migration file, but not actually execute the `up` command. I finally realized that by removing the `--esm` flag, the migration would run. This seems at odds with the knexjs documentation at https://knexjs.org/#esm-interop. What is the behavior of `--esm` flag in Knex 1.0.5 and newer? I'm happy to help update the documentation, but not sure what the new behavior is.

btw - @lehni I've been a fan of your work for many years! And @kibertoad, thank you for maintaining this extremely useful library.

lehni:
@andrewchilds I may be wrong, but I believe `--esm` is now only needed to support esm migrations from a non-ESM project. It's been a while since I wrote this PR and I already forgot the details, which are rather complex :)

I'd recommend creating a separate issue to make sure this gets addressed in the docs...

PS: Thanks for the kind words on my work!

lehni:
Yes it's indeed like this. This code here is from the tests:

```js
const fixture = [
  /** MIGRATIONS */
  {
    title: 'migrates esm modules',
    testCase: 'knexfile-esm',
    knexArgs: ['migrate:latest', '--esm'],
    dropDb: true,
    expectedOutput: 'Batch 1 run: 1 migrations',
    expectedSchema: [
      'knex_migrations',
      'sqlite_sequence',
      'knex_migrations_lock',
      'xyz',
    ],
  },
  {
    title: 'migrates esm modules from a module',
    testCase: 'knexfile-esm-module',
    knexArgs: ['migrate:latest'],
    expectedOutput: 'Batch 1 run: 1 migrations',
    dropDb: true,
    expectedSchema: [
      'knex_migrations',
      'sqlite_sequence',
      'knex_migrations_lock',
      'xyz',
    ],
  },
```

See how the first fixture (`knexfile-esm`) needs it, and the second (`knexfile-esm-module`) doesn't?  It's the second that provides `"type": "module"` in its `package.json` file.

andrewchilds:
Thank you @lehni for the detailed reply! I'll open a new issue about clarifying the docs related to the `--esm` flag. Cheers.

Issue #1630, opened 2016-08-16
Having in a unionAll creates a misplaced having clause


ianmaddox:
Example code

```
var foo = knex.select('id').from("mytable").having('val', '>', '0');
var bar = foo.clone();
bar.unionAll(foo.clone());
bar.unionAll(foo.clone());
console.log(bar.toString());
```

Output:

```
select `id` from `mytable` 
union all
select `id` from `mytable` having `val` > '0'
union all
select `id` from `mytable` having `val` > '0' having `val` > '0'
```

It appears that the union all from the first subquery is applied at the very end.

Tested on version 0.11.10

spruce-bruce:
I've been looking at this a little bit. Seems related to https://github.com/tgriesser/knex/issues/913#issuecomment-240234262

ianmaddox:
Looking at both now, I concur. Guess that helps narrow the source of the problem down to the union logic!

kibertoad:
Released in 1.0.5

PR #5072, opened 2022-03-18
Wrapped Unions Fixes

### LINKED ISSUE #5061: Syntax errors when using union with first/limit in 1.0.4
# Environment

Knex version: 1.0.4
Database + version: PostgreSQL 13.4
OS: Debian

# Bug

In 1.0.4 the behavior (order and wrapping) of union changed causing queries that previously worked to generate syntax errors. 

One example is:
```ts
const results = await knex('accounts')
  .select('last_name')
  .unionAll(function () {
    this.select('last_name').from('accounts');
  })
  .first();
expect(results).to.eql({ last_name: 'User' });
```

Which in 1.0.4 gives `error: select "last_name" from "accounts" limit $1 union all select "last_name" from "accounts" - syntax error at or near "union"`.

But in 1.0.3 it generates `select "last_name" from "accounts" union all select "last_name" from "accounts" limit ?`.

Another query that fails is
```ts
const results = await knex
  .unionAll(
    function () {
      this.select('last_name').from('accounts');
    },
    function () {
      this.select('last_name').from('accounts');
    },
    true
  )
  .first();
expect(results).to.eql({ last_name: 'User' });
```

with `error: (select "last_name" from "accounts") union all (select "last_name" from "accounts") (limit $1) - syntax error at or near "("`

I believe this change is introduced in https://github.com/knex/knex/pull/5030. Is this new behavior intended and we should rewrite our union queries?

This following is working, so it's not that hard to update the queries but it seems like a breaking change?
```ts
const results = await knex
  .unionAll(
    function () {
      this.select('last_name').from('accounts');
    },
    function () {
      this.select('last_name').from('accounts');
    }
  )
  .first();
```

OlivierCavadenti:
- Fix errors with wrapped unions.
- Closes #5061
- Related #1630

### DIFF
--- a/lib/query/querycompiler.js
+++ b/lib/query/querycompiler.js
@@ -125,16 +125,27 @@ class QueryCompiler {
     let sql = this.with();
 
     let unionStatement = '';
-    // Compute all statements to main query
-    const statements = components.map((component) => {
+
+    const firstStatements = [];
+    const endStatements = [];
+
+    components.forEach((component) => {
       const statement = this[component](this);
       // We store the 'union' statement to append it at the end.
       // We still need to call the component sequentially because of
       // order of bindings.
-      if (component === 'union') {
-        unionStatement = statement;
-      } else {
-        return statement;
+      switch (component) {
+        case 'union':
+          unionStatement = statement;
+          break;
+        case 'columns':
+        case 'join':
+        case 'where':
+          firstStatements.push(statement);
+          break;
+        default:
+          endStatements.push(statement);
+          break;
       }
     });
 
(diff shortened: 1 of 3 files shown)


kibertoad (approved):

kibertoad:
@OlivierCavadenti Are you planning any additional fixes in the nearby future, or can I cut new version after this is merged?

OlivierCavadenti:
> @OlivierCavadenti Are you planning any additional fixes in the nearby future, or can I cut new version after this is merged?

No, in the pipe, (in this order) : finish async / await all query integration files, finish all my opened PR (cast + override connection), then I will dig in open PR to see what we can finish.

felixmosh:
When this fix will be published?

kibertoad:
I'll try to publish new version today.

kibertoad:
Released in 1.0.5

PR #5075, opened 2022-03-19
Async/await in tests : all query integration files

OlivierCavadenti:
- Convert all query integration test files to async / await notation.
- https://github.com/knex/knex/issues/3170

### DIFF
--- a/test/integration2/query/misc/additional.spec.js
+++ b/test/integration2/query/misc/additional.spec.js
@@ -35,7 +35,6 @@ const {
 const logger = require('../../../integration/logger');
 const { insertAccounts } = require('../../../util/dataInsertHelper');
 const sinon = require('sinon');
-const JSONStream = require('JSONStream');
 
 describe('Additional', function () {
   getAllDbs().forEach((db) => {
--- a/test/integration2/query/misc/additional.spec.js
+++ b/test/integration2/query/misc/additional.spec.js
@@ -139,8 +139,7 @@ describe('Additional', function () {
             response = res;
           });
           stream.on('finish', () => {
-            expect(response.callCount).to.equal(1);
-            done();
+            return expect(response.callCount).to.equal(1);
           });
         });
 
--- a/test/integration2/query/misc/additional.spec.js
+++ b/test/integration2/query/misc/additional.spec.js
@@ -35,6 +35,7 @@ const {
 const logger = require('../../../integration/logger');
 const { insertAccounts } = require('../../../util/dataInsertHelper');
 const sinon = require('sinon');
+const JSONStream = require('JSONStream');
 
 describe('Additional', function () {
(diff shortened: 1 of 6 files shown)


test/integration2/query/misc/additional.spec.js
```diff
@@ -143,24 +129,18 @@ describe('Additional', function () {
...
-        it('should process response done through a stream', (done) => {
-          knex('accounts')
-            .truncate()
-            .then(() => {
-              return insertAccounts(knex, 'accounts');
-            })
-            .then(() => {
-              let response;
-              const stream = knex('accounts').limit(1).stream();
+        it('should process response done through a stream', async (done) => {
```
kibertoad, line 132: should it both be async and have `done`?

test/integration2/query/misc/additional.spec.js
```diff
@@ -143,24 +129,18 @@ describe('Additional', function () {
...
-              stream.on('finish', () => {
-                expect(response.callCount).to.equal(1);
-                done();
-              });
-            });
+          stream.on('data', (res) => {
+            response = res;
+          });
+          stream.on('finish', () => {
+            return expect(response.callCount).to.equal(1);
```
kibertoad, line 142: we need to add assertion that this point was reached at all

test/integration2/query/misc/additional.spec.js
```diff
@@ -135,13 +135,15 @@ describe('Additional', function () {
...
           const stream = knex('accounts').limit(1).stream();
 
+          let count = 0;
           stream.on('data', (res) => {
+            count++;
             response = res;
           });
           stream.on('finish', () => {
-            return expect(response.callCount).to.equal(1);
+            expect(count).to.equal(1);
```
kibertoad, line 144: this would still pass if finish is never invoked, wouldn't it? best approach would be to still await stream and then assert this count after that.

OlivierCavadenti: I just realise that I can't await a stream unfortunaly ^^ https://stackoverflow.com/a/33599789/1535159

kibertoad: Yeah, I was unsure it would work as expected, hence the suggestion to add more explicit assertions.

OlivierCavadenti: I have a test that is valid on my machine but always fail (/home/runner/work/knex/knex/test/integration2/query/update/updates.spec.js:124:34) with transactions, do you have suggestions about how to fix it ?

kibertoad:
@OlivierCavadenti Pretty sure that it is failing due to data leakage, probably one of the tests is removing or deleting row that test depends upon. Can you modify it to always delete all entries in the table and populate it with everything it needs?

kibertoad (approved):

Issue #5002, opened 2022-02-05
Knex Throwing SQLITE3 error when SQLITE3 Is Installed, 


Jerrardsmith215:
# Environment

Knex version: 1.0.2
Database + version: sqlite3 @ latest
OS: Windows 10

# Bug

1. Hi. I am experiencing an issue where I am having an error thrown while trying to use knex. If I migrate:latest I get the error below. I have tried reinstalling sqlite3 until I recognized that it asked for a @vscode-sqlite3 package. I installed that package and still the failures. 

2. Error message

Knex: run
$ npm install sqlite3 --save
Cannot find module '@vscode/sqlite3'
Require stack:
- C:\Users\jsmith1\Desktop\PPMUniversity\api\node_modules\knex\lib\dialects\sqlite3\index.js
- C:\Users\jsmith1\Desktop\PPMUniversity\api\node_modules\knex\lib\knex-builder\internal\config-resolver.js
- C:\Users\jsmith1\Desktop\PPMUniversity\api\node_modules\knex\lib\knex-builder\Knex.js
- C:\Users\jsmith1\Desktop\PPMUniversity\api\node_modules\knex\lib\index.js
- C:\Users\jsmith1\Desktop\PPMUniversity\api\node_modules\knex\knex.js
- C:\Users\jsmith1\Desktop\PPMUniversity\api\knexfile.js
- C:\Users\jsmith1\AppData\Roaming\npm\node_modules\knex\lib\migrations\util\import-file.js
- C:\Users\jsmith1\AppData\Roaming\npm\node_modules\knex\bin\cli.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:92:18)
    at Client_SQLite3._driver (C:\Users\jsmith1\Desktop\PPMUniversity\api\node_modules\knex\lib\dialects\sqlite3\index.js:33:12)
    at Client_SQLite3.initializeDriver (C:\Users\jsmith1\Desktop\PPMUniversity\api\node_modules\knex\lib\client.js:190:26)
    at new Client (C:\Users\jsmith1\Desktop\PPMUniversity\api\node_modules\knex\lib\client.js:75:12)
[...]

PR #5081, opened 2022-03-21
Fix @vscode-sqlite3 error message

OlivierCavadenti:
- Fix the error message when sqlite3 driver don't exist.
- Relates #5002
- If I remove my vscode/sqlite3 install, I have the error. If I reinstall it via "npm install @vscode/sqlite3 --save" it's ok for me. I miss something ?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -189,10 +189,9 @@ class Client extends EventEmitter {
     try {
       this.driver = this._driver();
     } catch (e) {
-      let driverName = this.driverName;
-      if (driverName === 'sqlite3') {
-        driverName = '@vscode/' + driverName;
-      }
+      const driverName = this.aliasDriverName
+        ? this.aliasDriverName
+        : this.driverName;
       const message = `Knex: run\n$ npm install ${driverName} --save`;
       this.logger.error(`${message}\n${e.message}\n${e.stack}`);
       throw new Error(`${message}\n${e.message}`);
--- a/lib/client.js
+++ b/lib/client.js
@@ -189,7 +189,11 @@ class Client extends EventEmitter {
     try {
       this.driver = this._driver();
     } catch (e) {
-      const message = `Knex: run\n$ npm install ${this.driverName} --save`;
+      let driverName = this.driverName;
+      if (driverName === 'sqlite3') {
+        driverName = '@vscode/' + driverName;
+      }
+      const message = `Knex: run\n$ npm install ${driverName} --save`;
       this.logger.error(`${message}\n${e.message}\n${e.stack}`);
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -189,7 +189,11 @@ class Client extends EventEmitter {
     try {
       this.driver = this._driver();
     } catch (e) {
-      const message = `Knex: run\n$ npm install ${this.driverName} --save`;
+      let driverName = this.driverName;
+      if (driverName === 'sqlite3') {
```
kibertoad, line 193: I would suggest more generic solution of driverName override map or, even better, make Client instances aware of their driverName

OlivierCavadenti: I introduce an alias driver name in sqlite3 index.js, I can't change the driver name by the way (is used by all tests that use testSql...).
By the way, I retest to delete the package, and run the command "npm install @vscode/sqlite3 --save" and I don't have any problem. I can't see more to do here to fix the problem in the related issue.

kibertoad (approved):

lib/client.js
kibertoad: I think current solution is perfect

tompsherman:
I've just spent a day or so chasing this bug in my code... and this thread as well as [Knex Throwing SQLITE Error](https://github.com/knex/knex/issues/5002) helped me fix it. I found two solutions that seem to work:

either the error message needs to be changed to `npm install @vscode/sqlite3 --save` because this fixes the issue but currently the error message reads `npm install sqlite3 --save` which does not fix the issue no matter how many times you install sqlite3
--OR-- 
the driver on line 32  `knex/lib/dialects/sqlite3/index` needs to be changed from `return require('@vscode/sqlite3');` to `return require('sqlite3');`

daniellockyer:
@OlivierCavadenti/@kibertoad: Any indication when this will be released? 🙂 The error message is tripping people up when `@vscode/sqlite3` fails to compile

kibertoad:
Released in 1.0.5

daniellockyer:
Thank you!

thefonso:
I have a small crazy question....Why is Knex dependent on @vscode/sqlite3 ?

I'm currently working a postgres project....no sqlite on my system...at all. Sooo. 

Why is Knex dependent on @vscode/sqlite3 ?

kibertoad:
@thefonso Do you mean migration creation or something else?

mrondin1:
Similar to @thefonso comment above, I'm getting the @vscode/sqlite3 error when I'm creating migrations on an exclusively postgres project. Is that expected?

kibertoad:
No. See #5106
1.0.6 should help in some cases, #5106 should help with the rest.

PR #5095, opened 2022-03-30
Update knex types for TS 4.7

weswigham:
Mostly, this involves adding a lot of missing type parameter constraints - in older versions of TS, we'd let an unconstrained type parameter be used where a type parameter extending `{}` is expected - in 4.7, we fix this.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -193,7 +193,7 @@ declare namespace DeferredKeySelection {
     ? DeferredKeySelection<TBase, TKeys | TKey, true, TAliasMapping, TSingle, TIntersectProps, TUnionProps>
     : DeferredKeySelection<unknown, TKey, true>;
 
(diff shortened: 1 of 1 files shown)


kibertoad:
Tests are failing.

weswigham:
That should do it, I hope - the explicit `{}` constraints require some of the `any`-swapping logic done for `unknown` to also need to apply to exactly `{}` (since now when inference fails for those parameters, they'll be `{}` instead of `unknown`).

weswigham:
I only changed the types, so I can't really answer why the `(12.x, pgnative)` CI build is failing - kinda doubt it has to do with this.

kibertoad:
Released in 1.0.5

glensc:
Does this mean that downstream projects need also to update their typescript to 4.7?

kibertoad:
@glensc I think it should work with earlier version too.

PR #5111, opened 2022-04-12
remove default client for migrations

sl-ffx:
1. the option "client" can not be set in the cli, because the "client" option is missing in the commander call and the resulting `config.client` is always "sqlite3"
2. all the other commands also have no default client

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -210,7 +210,6 @@ function invoke() {
     )
     .action(async (name) => {
       const opts = commander.opts();
-      opts.client = opts.client || 'sqlite3'; // We don't really care about client when creating migrations
       const instance = await initKnex(env, opts, true);
       const ext = getMigrationExtension(env, opts);
       const configOverrides = { extension: ext };


OlivierCavadenti:
> 1. the option "client" can not be set in the cli, because the "client" option is missing in the commander call and the resulting `config.client` is always "sqlite3"
> 2. all the other commands also have no default client

In the case of this command, as unit test show, you don't have client setted in this case and an error is thrown.

sl-ffx:
> config.client is always to sqlite3 hence when configured in knexfile.js ?
>
> Like :
>
> module.exports = {
>   client: 'pg',
>   connection: 'postgres://localhost/test'
> }

It does not work with a knexfile.js containing the client 'pg'.

> Maybe the solution is to add client option in the commander call also.

With the client option in the commander call works. But I think this is not ideal, because the client is already set in the knexfile.js

I'm just wondering: Why does the "migrate:make" have a test without a knexfile.js, but the "migrate:latest" haven't.

OlivierCavadenti:
Fix in 1.0.7 by https://github.com/knex/knex/pull/5106

PR #5116, opened 2022-04-13
FIX 5114: Allow `null` as a valid value for `.enu`/`.enum`

### LINKED ISSUE #5114: Types for `.enu` should accept `null`
# Environment

Knex version: 1.0.5
Typescript: 4.6.3

@lorefnon.

# Bug

The `.enum`/`.enu` functions of `TableBuilder` should accept `null` as a value input in the typings, as is the example in the documentation.

3. Error message
```
error TS2345: Argument of type 'null' is not assignable to parameter of type 'readonly Value[]'
```
4. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real

```typescript
///@ts-check
/**param { import('knex').Knex} Knex */
export async function up(knex) {
   await knex.schema.createTable("mytable", (table) => {
     // typescript error occurs here:
      table.enu('column', null, { useNative: true, existingType: true, enumName: 'foo_type' })
   });
}
```

Naddiseo:
When using the `useNative` and `existingType` options, the value isn't utilized.
The documentation uses `null` as an example, this commit allows said example
to pass in typescript

close #5114

### DIFF
--- a/test-tsd/tables.test-d.ts
+++ b/test-tsd/tables.test-d.ts
@@ -101,6 +101,12 @@ const main = async () => {
     table.integer('num').references('id').withKeyName('non_for7').deferrable('deferred').inTable('non_exist').onDelete('CASCADE');
     table.integer('num').references('id').inTable('non_exist').onDelete('CASCADE').withKeyName('non_for6').deferrable('deferred');
     table.integer('num').references('id').withKeyName('non_for7').onDelete('CASCADE').deferrable('deferred').inTable('non_exist');
+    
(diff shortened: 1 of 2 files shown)


Naddiseo:
I have no idea why the lint stage throws, but also doesn't fail the test:
```
The 'no-redundant-jsdoc' rule threw an error in '/home/runner/work/knex/knex/types/index.d.ts':
Error: Unexpected tag kind: JSDocDeprecatedTag
```

OlivierCavadenti (approved):

PR #5120, opened 2022-04-14
Add 'validate' pool options. Resolves #5119.

### LINKED ISSUE #5119: Validate and/or expire connections
# Environment

Knex version: all
Database + version: mysql, postgres
OS: linux

In our cloud environment, we can add and remove database instances. We would like, if we add a new database instance, to see load spread out onto the new instances.

At present, connections can potentially be very very long lived. There's no setting or configuration possible (without a warning) to alter this behavior. We do not see connections tend to migrate towards being evenly distributed.

The most general solution would be to expose [Tarn.js's `validate`](https://github.com/vincit/tarn.js/#usage), which would let consumers perform arbitrary validation of their connections before use.

Our specific use case would also be well served by adding a `maxConnectionLifetime` or some such, which automatically expired connections after a certain time. Note that a good implementation ought include some fuzz time: we have 50 instances of foo-service, which all start up about the same time. That's ok, but we don't want them to (for example), every 5 minutes, all reconnect at the same time. We imagine each connection expiring after 2-4 minutes, randomly selected.

For a general solution, we could add `validate` to the [`POOL_CONFIG_OPTIONS` constant](https://github.com/knex/knex/blob/939d8a219c432a7d7dcb1ed1a79d1e5a4686eafd/lib/constants.js#L36), so users may provide their own pool.validate function.

Expiring/settings a maximum connection lifetime could be a more explicit pool configuration item. I don't know what color we'd want to paint this bikeshed.

# Alternatives

[`afterCreate`](http://knexjs.org/#Installation-pooling-afterCreate) could also be used to self-destruct a connection after a time!

Packages like https://www.npmjs.com/package/knex-dynamic-connection monkeypatch Knex's behavior, to allow connections to be dynamically defined. Potentially we could create connections which expire/close themselves after a time with this.

rektide:
In #5119, I described an issue where knex & the underlying tarn.js pool of connections does not roll off old connections, and so new database instances added behind a load-balancer end up going unused.

This PR proposes the simplest possible solution to #5119, which is to allow consumers to also define a `pool.validate` function.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -269,6 +269,10 @@ class Client extends EventEmitter {
           return false;
         }
 
+        if (poolConfig.validate && !poolConfig.validate(connection)) {
+          return false;
+        }
+
         return this.validateConnection(connection);
       },
     });
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -43,6 +43,7 @@ const POOL_CONFIG_OPTIONS = Object.freeze([
   'numTestsPerRun',
   'softIdleTimeoutMillis',
   'Promise',
+  'validate',
 ]);
 
 /**


OlivierCavadenti (changes requested):
tests fails (and could be nice to add some others)

jmealo:
@OlivierCavadenti @rektide This would be very helpful. There's no good way to handle when an Aurora failover occurs and the pool is connected to a READONLY instance. (Doesn't appear to be a way to globally handle errors, which has a reference to the connection, and can either return it to the pool or destroy the connection (not the pool)? )

mercmobily:
@rektide You still around? If not all good

mercmobily:
@kibertoad It would make me feel a lot better if you could have a look at these changes.

mercmobily:
@myndzi @kibertoad I would LOVE to know what you think about this one. I have established a workflow so that I won't have to have, in the future, quite so many failed CI tests...

lib/client.js
```diff
@@ -273,15 +305,66 @@ class Client extends EventEmitter {
...
+          await updatePoolConnectionSettingsFromProvider(true);
+
+          if (
+            this.connectionConfigExpirationChecker &&
+            this.connectionConfigExpirationChecker()
+          ) {
+            // If it would still report expired after a refresh, disable it to avoid an infinite refresh/drop loop.
+            // Without this guard we could spin forever: checker -> refresh -> still expired -> drop -> repeat.
+            // Examples: provider returns already expired data, or the checker is misconfigured to always return true.
+            this.connectionConfigExpirationChecker = null;
```
myndzi: This behavior is a bit surprising, though I understand why you chose it.

I think it would be preferable, when encountering a case like this, to just fail rather than "decide to do something different than what the user asked for". After all, this isn't our problem, it's the caller's problem. Ideally, we could reject the query that asked for the connection with a clear message.

A throttle on acquiring new config is a straightforward solution to the "expired loop" problem, limiting knex from behaving badly if the user makes an error. If we request new config no more frequently than once per X time, it's not possible to have an "expired loop" impair our ability to acquire connections at all. Then, so long as the config is valid, we will continue to request new configuration for as long as the caller tells us we should.

If a user tells us the min config lifetime is 0 and also always returns expired = true, then it's well and truly on them and their queries should be failing with a clear explanation, not failing with an unclear explanation (which is what will happen if we stop asking for config at all, but the config has expired)

lib/client.js
```diff
@@ -208,10 +208,10 @@ class Client extends EventEmitter {
   }
 
   getPoolSettings(poolConfig) {
-    poolConfig = defaults({}, poolConfig, this.poolDefaults());
+    const rawPoolConfig = defaults({}, poolConfig, this.poolDefaults());
```
myndzi: I can see that you took my comments on discord into account, but what I should have done was given an example of what I meant, so I'll leave that here and you can consider if you want to use it or not.

By my understanding, this function receives "what the user gave to us", layers it with defaults, plucks off some stuff, and gives Tarn a subset of those options. A way to tackle this cleanly is:

```js
const {known, tarn, options, ...userConfig} = defaults({}, poolConfig, this.poolDefaults());
const {maxConnectionLifetimeMillis, etc} = userConfig;

const tarnPoolConfig = {
  known,
  tarn,
  options,
  create: () => ...,
  validate: () => ...
};
```

This is the kind of thing I was trying to describe; it ensures that, now and in the future, Tarn never receives unexpected properties that may not work. Instead of "remove the things we know shouldn't be there" it's "use only the things we know _should_ be there".

It also makes it directly easy to read, in the code, what it's expected to be doing, instead of "tarn gets _something_ that comes from _somewhere_ that's the result of merging _some stuff_".

Finally, it removes the need to reassign "poolConfig" over the input arguments and also to have a parallel version (rawPoolConfig), because we're being explicit about our data. We can see what we are telling tarn to do, and where those instructions are expected to come from, which makes the code more understandable.

lib/client.js
```diff
@@ -273,15 +305,66 @@ class Client extends EventEmitter {
...
+        if (
+          this.connectionConfigExpirationChecker &&
+          this.connectionConfigExpirationChecker()
+        ) {
+          await updatePoolConnectionSettingsFromProvider(true);
+
+          if (
+            this.connectionConfigExpirationChecker &&
+            this.connectionConfigExpirationChecker()
+          ) {
```
myndzi: Since we're already defining function-local helpers, it might be simpler to define this logic in the same place, up above, and then the validate function can look like:

```js
validate: async (connection) => {
  if (connectionIsExpired()) return false;
  if (configWasRefreshed()) return false;
  if (!userValidate(connection)) return false;
  return this.validateConnection(connection);
}
```

This makes the high-level logic clear, and avoids splitting the logic of e.g. "checking for expired config" between "this code down here" and "that code up there".

Instead of checking "does a function exist || does the function return something?" everywhere, we define a function that always exists, but it's just a noop if the config isn't there:

```js
const configWasRefreshed = async () => {
  if (!userConfig.connectionConfigExpirationChecker) return false;
  await updatePoolConnectionSettingsFromProvider(true);
  return true;
}
```

lib/client.js
```diff
@@ -233,15 +233,40 @@ class Client extends EventEmitter {
...
-    const updatePoolConnectionSettingsFromProvider = async () => {
+    const maxConnectionLifetimeMillis =
+      rawPoolConfig.maxConnectionLifetimeMillis;
+    const maxConnectionLifetimeJitterMillis =
+      rawPoolConfig.maxConnectionLifetimeJitterMillis;
+    const userValidate = rawPoolConfig.validate;
+
+    const hasMaxConnectionLifetime =
+      typeof maxConnectionLifetimeMillis === 'number' &&
+      maxConnectionLifetimeMillis > 0;
```
myndzi, line 247:
```suggestion
    const hasMaxConnectionLifetime =
      typeof maxConnectionLifetimeMillis === 'number' &&
      Number.isFinite(maxConnectionLifetimeMillis) &&
      maxConnectionLifetimeMillis > 0;
```

lib/client.js
```diff
@@ -233,15 +233,40 @@ class Client extends EventEmitter {
...
+    const generateLifetimeLimit = () => {
+      if (!hasMaxConnectionLifetime) {
+        return null;
+      }
+      const jitter =
+        typeof maxConnectionLifetimeJitterMillis === 'number' &&
+        maxConnectionLifetimeJitterMillis > 0
+          ? Math.floor(Math.random() * maxConnectionLifetimeJitterMillis)
+          : 0;
+      return Date.now() + maxConnectionLifetimeMillis + jitter;
```
myndzi:
```suggestion
      return performance.now() + maxConnectionLifetimeMillis + jitter;
```

lib/client.js
```diff
@@ -273,15 +305,66 @@ class Client extends EventEmitter {
...
           return false;
         }
 
+        if (hasMaxConnectionLifetime) {
+          if (!connection.__knexLifetimeLimit) {
+            connection.__knexLifetimeLimit = generateLifetimeLimit();
+          }
+          if (
+            connection.__knexLifetimeLimit &&
+            Date.now() > connection.__knexLifetimeLimit
```
myndzi:
```suggestion
            performance.now() > connection.__knexLifetimeLimit
```

test/unit/client/pool-config.js
```diff
@@ -0,0 +1,117 @@
...
+    expect(userValidate.calledOnce).to.be.true;
+    expect(result).to.be.false;
+  });
+
+  it('rejects connections past max lifetime', async () => {
+    const client = new TestClient();
+    const poolConfig = client.getPoolSettings({
+      maxConnectionLifetimeMillis: 1,
+    });
+    const connection = { __knexLifetimeLimit: Date.now() - 1000 };
```
myndzi:
```suggestion
    const connection = { __knexLifetimeLimit: -1 };
```
(in line with the performance.now() change; but for testing it just has to always be expired so a static value is fine)

test/unit/client/pool-config.js
```diff
@@ -0,0 +1,117 @@
...
+  it('logs and rejects when user validate throws', async () => {
+    const client = new TestClient();
+    const warnSpy = sinon.spy(client.logger, 'warn');
+    const userValidate = sinon.stub().rejects(new Error('boom'));
+
+    const poolConfig = client.getPoolSettings({ validate: userValidate });
+    const result = await poolConfig.validate({});
+
+    expect(userValidate.calledOnce).to.be.true;
+    expect(warnSpy.calledOnce).to.be.true;
```
myndzi, line 76: It'd be good to match the message here too; it's easy to have tests that falsely pass on things like this when a warning is unexpectedly generated elsewhere

test/unit/client/pool-config.js
```diff
@@ -0,0 +1,117 @@
...
+    sinon.stub(Math, 'random').returns(0.5); // halfway jitter
+    const poolConfig = client.getPoolSettings({
+      maxConnectionLifetimeMillis: 1000,
+      maxConnectionLifetimeJitterMillis: 1000,
+    });
+    const connection = {};
+
+    const result = await poolConfig.validate(connection);
+
+    expect(connection.__knexLifetimeLimit).to.equal(1000 + 1000 + 500);
```
myndzi: Rather than test implementation, we should test behavior. We're already using sinon to mock time, so advance the time to a point between the min and the extra jitter, expect it to work; then, advance the time past the jitter, expect it to fail

test/unit/client/pool-config.js
```diff
@@ -0,0 +1,117 @@
...
+    expect(result).to.be.false;
+  });
+
+  it('does not warn when pool.validate is provided', () => {
+    const client = new TestClient();
+    const warnSpy = sinon.spy(client.logger, 'warn');
+
+    client.getPoolSettings({ validate: () => true });
+
+    expect(warnSpy.called).to.be.false;
```
myndzi, line 64: Not fully sure why this test exists, but it should also verify some subset of the warning message, to distinguish between the warning you expect not to be here and some unrelated warning

lib/client.js
mercmobily: I thought about this hard. I ended up working to please the test... I agree, this is too surprising. There is no point for the user to return "true" all of the time. I changed the code to throw, since there is no legit reason why it wouldn't be like this.
The most painful part was reworking the tests. I am still not 100% happy, but "happy enough"

lib/client.js
mercmobily: Alright I see your point. I think I went far enough improving this... Whitelisting only Tarn keys would be clearer but adds maintenance, so I'd keep the current approach.

lib/client.js
mercmobily: I think I've done this. You are right, the code is muchhhhhhhhhhhhhhhh nicer...

test/unit/client/pool-config.js
mercmobily: I believe this is solved with the changed tests.

test/unit/client/pool-config.js
mercmobily: Again, the new tests should be doing this now.

test/unit/client/pool-config.js
mercmobily: Again addressed with changed tests.

mercmobily:
> Since this is my first review here, I'm leaning towards rigor and improving things for the future. We'll see how things go and I can always adjust to be either more or less detailed.

Wow, you did absolutely review this, no doubts there :D

I *think* I addressed everything. I didn't do a final super-check -- I would prefer doing that once you agree that I haven't missed anything.

docs/src/guide/index.md
```diff
@@ -254,7 +254,7 @@ const knex = require('knex')({
 });
 ```
 
-By default, the configuration object received via a function is cached and reused for all connections. To change this behavior, an `expirationChecker` function can be returned as part of the configuration object. The `expirationChecker` is consulted before trying to create new connections, and in case it returns `true`, a new configuration object is retrieved. For example, to work with an authentication token that has a limited lifespan:
+By default, the configuration object received via a function is cached and reused for all connections. To change this behavior, an `expirationChecker` function can be returned as part of the configuration object. The `expirationChecker` is consulted both when creating new connections and before reusing a pooled connection. If it returns `true`, Knex refreshes the connection settings from your provider and drops the current connection so the pool can create a new one with the updated configuration. If the checker still reports expired after a refresh, it is disabled to avoid repeated churn.
```
myndzi: Docs should be updated to reflect the new behavior :)

docs/src/guide/index.md
```diff
@@ -367,6 +384,43 @@ const knex = require('knex')({
 });
 ```
 
+#### Custom validation (`pool.validate`)
+
+You can provide a `pool.validate` function to decide if a pooled connection should be reused. It can be sync or async; return/resolve `false` (or throw) to veto reuse and force the pool to create a fresh connection.
```
myndzi:
```suggestion
You can provide a `pool.validate` function to decide if a pooled connection should be reused. It can be sync or async; return/resolve `false` (or throw) to discard the connection and force the pool to replace it.
```

This is super minor, but the phrase "... to veto reuse and ..." could be misunderstood; the noun and verb forms have slightly differing connotations (the noun, as in "to veto the reuse of", is the intended meaning here, but the verb, as in "to reuse", means the opposite)

lib/client.js
myndzi:
```suggestion
      return performance.now() + maxConnectionLifetimeMillis + jitter;
```

Previous suggestion got dismissed

lib/client.js
```diff
@@ -233,15 +239,34 @@ class Client extends EventEmitter {
 
     // acquire connection timeout can be set on config or config.pool
     // choose the smallest, positive timeout setting and set on poolConfig
-    poolConfig.acquireTimeoutMillis = Math.min(...timeouts);
+    const acquireTimeoutMillis = Math.min(...timeouts);
 
-    const updatePoolConnectionSettingsFromProvider = async () => {
+    const hasMaxConnectionLifetime =
+      typeof maxConnectionLifetimeMillis === 'number' &&
+      maxConnectionLifetimeMillis > 0;
```
myndzi, line 247:
```suggestion
      typeof maxConnectionLifetimeMillis === 'number' &&
      Number.isFinite(maxConnectionLifetimeMillis) &&
      maxConnectionLifetimeMillis > 0;
```

lib/client.js
```diff
@@ -256,13 +281,58 @@ class Client extends EventEmitter {
...
+    const connectionIsOverLifetime = (connection) => {
+      if (!hasMaxConnectionLifetime) {
+        return false;
+      }
+      if (!connection.__knexLifetimeLimit) {
+        connection.__knexLifetimeLimit = generateLifetimeLimit();
+      }
+      return (
+        connection.__knexLifetimeLimit &&
+        Date.now() > connection.__knexLifetimeLimit
```
myndzi:
```suggestion
        performance.now() > connection.__knexLifetimeLimit
```

docs/src/guide/index.md
mercmobily: Ah I had missed this one. Done.

lib/client.js
myndzi: Github still being a pain.  This needs to also use `performance.now()` so that it agrees with line 302

Image

zahid492:
I think types are missing in `index.d.ts`

>     maxConnectionLifetimeMillis?: number;
>     maxConnectionLifetimeJitterMillis?: number;

andersnm:
@mercmobily  We just tested a stress test run with this config:
```
  pool: {
    maxConnectionLifetimeMillis: 4 * 60000, // 4 minutes
    maxConnectionLifetimeJitterMillis: 60000,
```

Alas, the load was not distributed to the autoscaled replica. I believe @myndzi pointed out the correct underlying cause above where the life checks mix up `performance.now()` and `Date.now()`.

Here the CPU goes up, it autoscales and adds a replica, so the average CPU goes down, but then response time and average DB CPU stays the same - expected response time to go down and DB CPU to go up:
image

Also confirmed in the logs when measuring the average, only one instance was used.

We can try again once this and the failing tests are fixed.

mercmobily:
@myndzi Tests pass, please have a look, I am very ready to mark this as ready to merge

mercmobily:
@andersnm Apologies, I fixed it. Could you try now? Sorry...

mercmobily:
> I think types are missing in `index.d.ts`
> 
> > ```
> > maxConnectionLifetimeMillis?: number;
> > maxConnectionLifetimeJitterMillis?: number;
> > ```

Well caught. @myndzi I have bounced this back onto my list, I will apply the final fix (TS types missing) and will send it back to you for final check.

mercmobily:
@zahid492 Would you like to lend a hand in general with Knex? Development is picking up, we would love to have more hands on deck

andersnm:
@mercmobily Thanks, works as expected! Here we are using `maxConnectionLifetimeMillis: 4 * 60000, // 4 minutes` and no jitter:

image

mercmobily:
@myndzi I added the missing types. PLEASE CHECK... If everything is OK, and you are happy with everything, please mark it as to be merged since it should be all done.

docs/src/guide/index.md
```diff
@@ -254,7 +254,7 @@ const knex = require('knex')({
 });
 ```
 
-By default, the configuration object received via a function is cached and reused for all connections. To change this behavior, an `expirationChecker` function can be returned as part of the configuration object. The `expirationChecker` is consulted before trying to create new connections, and in case it returns `true`, a new configuration object is retrieved. For example, to work with an authentication token that has a limited lifespan:
+By default, the configuration object received via a function is cached and reused for all connections. To change this behavior, an `expirationChecker` function can be returned as part of the configuration object. The `expirationChecker` is consulted both when creating new connections and before reusing a pooled connection. If it returns `true`, Knex refreshes the connection settings from your provider and drops the current connection so the pool can create a new one with the updated configuration. If the checker still reports expired after a refresh, Knex throws to avoid repeated churn.
```
myndzi:
```suggestion
By default, the configuration object received via a function is cached and reused for all connections. To change this behavior, an `expirationChecker` function can be returned as part of the configuration object. 
```

It looks like this section is duplicated below (line 279), presumably to elaborate more clearly with the new example. Did you intend to delete it from here?

kibertoad (approved):

kibertoad:
@myndzi can you resolve the conflict?

myndzi:
@kibertoad done :)

shurane:
Is there documentation on how to use `validate`? The only thing I see on how to use this is at  https://github.com/vincit/tarn.js/#usage

mercmobily:
@shurane This has been documented in the official documentation. I recommend you look at the diff to see specifically what was added: https://github.com/knex/knex/pull/5120/changes#diff-d7c624c97a458e85cfb151e6f82c7c53f06621ca26d14f0c1b38baa2181ce6b4

shurane:
Oh ok thanks @mercmobily! I see it here: https://github.com/knex/knex/blob/master/docs/src/guide/index.md#custom-validation-poolvalidate

Btw, is there a more recent release that pulls in this (and other changes)? It seems like the last one was 3.1.0 during Dec 2023.

mercmobily:
We are releasing VERY SOON. Along with the release, we will also add a workflow to prevent drift between documentation and releases. Till then (I am talking days, not weeks), I actually recommend you use master!

PR #5121, opened 2022-04-15
Migrate documentation to vitepress

stafyniaksacha:
This pull request is related to work made on https://github.com/knex/documentation/issues/391 and https://github.com/knex/documentation/issues/337

Preview 🪄 https://stafyniaksacha.github.io/knex/

It migrates back the documentation on this repository (with github action configured to ignore changes on the `docs` so it will not trigger them) so we benefit from the latest changelog and have one source of truth. It's opinionated, but I think it will encourage people to update documentation when they make changes (we can move this back to the other repository if needed).

All the documentation and wiki content has been migrated to markdown with latest version (v1.0.7)

- I do not have made anything related with dialect SQL output selection, but we can do it easily with custom vue components and localstorage (as on the vuejs website for the composition/option api)
- We may want to discuss the documentation sections as well!
- We can also add a great search experience easily with https://docsearch.algolia.com/apply/ 
- A lot of small improvement can be made!


OlivierCavadenti:
Amazing ! I will read carefull read this PR as soon as possible, but the result seems wonderful.

vencho-mdp:
Oh, it looks awesome!

stafyniaksacha:
@vencho-mdp @OlivierCavadenti just pushed a darkmode :)

vencho-mdp:
I'd like to try to add the SQL dropdown we have in the current Knex docs. It's quite nice (just as the composition and options api in vue).

vencho-mdp:
Hi! I'm having quite a lot of problems with the installation of dependencies. In my opinion, if the docs and knex source code will be in the same repo, each should have a separate `package.json` because of this, dependencies issues. What's your opinion?

stafyniaksacha:
> Hi! I'm having quite a lot of problems with the installation of dependencies. In my opinion, if the docs and knex source code will be in the same repo, each should have a separate `package.json` because of this, dependencies issues. What do you think?

hum, that's a good question. I think since we are not in a monorepo we should keep only one package.json. Se we can either bring back the doc to the other repo or keep the doc dependencies lean. (we may just need vueuese I think, but we can add tons of vite plugins)

vencho-mdp:
Oh, there's already a repo for knex docs. https://github.com/knex/documentation

intech:
### Excellent work! Let me suggest a few improvements:

**1. The color of the names is too pale. It would be best if you made it more contrast (darker)**

Screenshot from 2022-04-18 16-00-54

**2. The width of the right content column is too narrow. Why not make it fullscreen**

Screenshot from 2022-04-18 16-01-27

stafyniaksacha:
> Excellent work! Let me suggest a few improvements:
> 
> **1. The color of the names is too pale. It would be best if you made it more contrast (darker)**
> 
> Screenshot from 2022-04-18 16-00-54

I agree with your feedback! So I updated a bit the contrast:

image

> **2. The width of the right content column is too narrow. Why not make it fullscreen**
> 
> Screenshot from 2022-04-18 16-01-27

For the width, I just augmented the max-width. As fullwidth make sentences and code samples to wide: 
image


@intech @vencho-mdp should I move this pr to https://github.com/knex/documentation ?

OlivierCavadenti:
I think is better to keep the two repos splitted to not blend dependencies and configuration (like webpack). @kibertoad  ?

kibertoad:
I agree, let's keep them separate.

stafyniaksacha:
Can you create an empty `main` branch on the `documentation` repo so I can create the PR from https://github.com/stafyniaksacha/documentation/tree/main

kibertoad:
@stafyniaksacha Empty `main` created.

stafyniaksacha:
@kibertoad @OlivierCavadenti @intech @vencho-mdp done in https://github.com/knex/documentation/pull/400 I close this PR

PR #5123, opened 2022-04-15
Fix driver options specified with `.options()` method being ignored for oracledb dialect

Kytech:
### Bug Description

Recently, when working on an application using oracle databse, it became necessary to specify options to the oracle database driver/client on a per-query basis. According to the knex documentation, this is facilitated using the `options()` method. However, the options provided would not be passed to the oracle database client, even though all options were returned from calling `queryBuilder.toSQL().options`. In my case, I was supplying the `fetchInfo` option to oracle, though I would find it's options would not be applied to the query.

The version of the oracle database client node package used was `4.2.x`, though the same behavior exists within versions `5.x` as they did not alter the API surface dealing with type mappings.

The below example illustrates the expected behavior and the actual behavior:

```js
const knex = require('../path/to/instantiated/knex/import');
const oracledb = require('oracledb');

// NUMERIC_COLUMN is any column in Oracle that has a numeric column type
const result = await knex('TEST_TABLE').select('NUMERIC_COLUMN').options({ fetchInfo: { "NUMERIC_COLUMN": { type: oracledb.STRING } } })

// This returns { fetchInfo: { "NUMERIC_COLUMN": { type: oracledb.STRING } } } as expected
knex('TEST_TABLE').select('NUMERIC_COLUMN').options({ fetchInfo: { "NUMERIC_COLUMN": { type: oracledb.STRING } } }).toSQL().options

// Expected value for result
[
  {
    "NUMERIC_COLUMN": "5"
  }
]

// Actual value for result
[
  {
    "NUMERIC_COLUMN": 5
  }
]
```

### Underlying Cause

After a bit of debugging, I found the cause of the issue to be [line 172 in `lib/dialects/oracledb/index.js`](https://github.com/knex/knex/blob/939d8a219c432a7d7dcb1ed1a79d1e5a4686eafd/lib/dialects/oracledb/index.js#L172), where a predefined set of options is passed to the oracle client without reading or merging in `obj.options`, resulting in anything being supplied to `.options()` on the query builder being ignored when using the oracledb dialect.

### Fix Details

The fix was to merge the user-provided options with the set of pre-defined options required by knex (as a bit of digging indicated knex does not support autoCommit on oracle), allowing the pre-defined options to overwrite anything the user may specify. This resulted in the oracle client receiving the expected parameters and the expected return value from the query.

### Final Comments

For almost every other dialect, if not all others, proper support for the `options()` query builder method is working. This seems to suggest that this was unintended behavior. If this is intended behavior, this causes issues for cases where the query result type mapping needs to be altered for a specific query [in common conditions detailed here by Oracle](https://oracle.github.io/node-oracledb/doc/api.html#typemap) (section 17.1.7.2 provides a good example). To do this on a per-query basis, this must be done using the [`fetchInfo`](https://oracle.github.io/node-oracledb/doc/api.html#--42635-fetchinfo) option provided to `connection.execute()` as all other options control this setting on a global level.


OlivierCavadenti:
Can you add some tests (you can use your example) ?

Kytech:
@OlivierCavadenti Sure thing. I can write some tests based on my example.

Kytech:
@OlivierCavadenti I've added an integration test based on the example I gave in this PR. It should be ready for review.

test/integration2/dialects/oracledb.spec.js
```diff
@@ -1,5 +1,6 @@
 const { expect } = require('chai');
 const { getAllDbs, getKnexForDb } = require('../util/knex-instance-provider');
+const oracledb = require('oracledb');
```
OlivierCavadenti, line 3: why need this ? you have the db below, in "dbs" variable

OlivierCavadenti (changes requested):
Tests fails:

```
   Error: DPI-1047: Cannot locate a 64-bit Oracle Client library: "libclntsh.so: cannot open shared object file: No such file or directory". See https://oracle.github.io/node-oracledb/INSTALL.html for help
Node-oracledb installation instructions: https://oracle.github.io/node-oracledb/INSTALL.html
You must have 64-bit Oracle Client libraries configured with ldconfig, or in LD_LIBRARY_PATH.
If you do not have Oracle Database on this computer, then install the Instant Client Basic or Basic Light package from 
https://www.oracle.com/database/technologies/instant-client/linux-x86-64-downloads.html
```

test/integration2/dialects/oracledb.spec.js
Kytech: @OlivierCavadenti correct me if I'm wrong, but from what I can tell, `getAllDbs`, which is what populates the `dbs` variable, just returns an array of strings indicating the names of all databases that have been specified to test (since it reads process.env.DB or a default array of all database client names). I'm not sure how that would give me what I am needing in this test. What I'm needing here is the oracle database driver package and not a string for the driver's name since I need to import some constants from the `node-oracledb` database driver package (exposed under the `oracledb` import name) itself since many of oracle's database driver options require use of constants defined in the package. (Line 62 shows where those constants are used.) As far as I can tell, `dbs` does not give me access to these constants.

Let me know if that clears up your question or if I'm perhaps misunderstanding something about the `dbs` variable.

Kytech:
> Tests fails:
> 
> ```
>    Error: DPI-1047: Cannot locate a 64-bit Oracle Client library: "libclntsh.so: cannot open shared object file: No such file or directory". See https://oracle.github.io/node-oracledb/INSTALL.html for help
> Node-oracledb installation instructions: https://oracle.github.io/node-oracledb/INSTALL.html
> You must have 64-bit Oracle Client libraries configured with ldconfig, or in LD_LIBRARY_PATH.
> If you do not have Oracle Database on this computer, then install the Instant Client Basic or Basic Light package from 
> https://www.oracle.com/database/technologies/instant-client/linux-x86-64-downloads.html
> ```

This is an issue with the Oracle Instant Client not being installed or detected when this test is running. This error gets thrown by the `node-oracledb` package and is external to the test code. I've seen this one a lot when helping developers setup a new workstation at my workplace. Maybe there is an issue with how the `integration2` tests are being run, considering how all oracle tests in `integration` run queries against oracle without problem. The only reason I can think of as to why it runs fine on my workstation is because I have instant client setup on the machine and its system-level install is always detected by any nodejs app I run.

test/integration2/dialects/oracledb.spec.js
OlivierCavadenti: It seems it's that require that made the tests fails as I can see, can we avoid that ?

kibertoad: Since Oracle driver by itself requires properly configured environment, every test that depends on it automatically becomes integration one. I would suggest extracting that particular test for Oracle into integration2

Kytech: @kibertoad I currently have this test in the `integration2` directory. Is there any additional configuration I need to do so that this is running as a proper integration test? @OlivierCavadenti I could rewrite this test to pass a different option, such as setting the `extendedMetaData` boolean option instead. It would not require importing any constants from the oracledb package, but it would still result in different behavior for the test query I am using and would still confirm that options set using `.options()` eventually make it to the oracle database driver for that specific query.

Kytech:
Just realized that the error was occurring in the non-oracledb tests. Looks like I will need to adjust those test files to make sure that test isn't being run for other DBs.

test/integration2/dialects/oracledb.spec.js
Kytech: @OlivierCavadenti @kibertoad I just realized that this test file was being included/executed even when other databases were specified to run, contrary to my expectation. Moving the test inside the `dbs.forEach` -> `describe` block prevented these tests from executing when oracle is not included in the databases to test for that run. I also moved the require call to the oracledb driver into the body of the test that needs it so that we don't run into that error about not having an oracle environment setup when that test is not being executed. I do need to `require` the oracle driver at some point since most of oracle's options require referencing some pre-defined values included in the driver for proper use. The driver name to require is also being resolved off of the knex instance as extra insurance so that the test won't include a driver package that is not installed.

Kytech:
@OlivierCavadenti or @kibertoad could I get a re-run of the test suite? I've implemented some changes that should resolve the oracle instant client error that was occurring, but I'll want to verify the fix against the CI results.

baileymatthewr:
@Kytech : could you check on this again, please?

Kytech:
> @Kytech : could you check on this again, please?

Not really anything else I can do other than wait on that re-review. I'll check and see if the knex maintainers have been doing anything lately and if so, I'll try opening an issue that links this PR to see if that gets better visibility.

Kytech:
> @Kytech : could you check on this again, please?

Not really anything else I can do other than wait on that re-review. Knex maintainers seem to have looked over the PRs they have recently, so I'll see if @OlivierCavadenti has anything else they want changed or if the adjustments I made for those failing tests addressed everything needed.

OlivierCavadenti:
> > @Kytech : could you check on this again, please?
> 
> Not really anything else I can do other than wait on that re-review. Knex maintainers seem to have looked over the PRs they have recently, so I'll see if @OlivierCavadenti has anything else they want changed or if the adjustments I made for those failing tests addressed everything needed.

Yeah, after a long hiatus, I try to reviews as many as I can :D. Sorry for the delay

OlivierCavadenti (approved):

PR #5134, opened 2022-04-20
Added a dump option to the knex CLI

StanleyMasinde:
## Goal
To have a database dump option which is useful for things like back ups. This commit only changes the knex CLI by adding the `dump:db` option.

I've been able to test it on mysql and Sqlite because those are the DBs I currently have access to. 

### Test it out 
1. Run `npm i stanleymasinde/knex`
2. `npx knex schema:dump`

That should work okay unless you are using esm like in my case where you'll have to specify all the options e.g `npx knex schema:dump --esm --knexfile knexfile.cjs`

Your database (The one in config) will be available as <database-name>.sql in the root of your folder. You can specify the filename with --output <file-name>


StanleyMasinde:
Hey @OlivierCavadenti could you please review this so that I can know where to make further changes?

kibertoad:
Can you please add full integration test?

StanleyMasinde:
> Can you please add full integration test?

Sure let me do that

PR #5137, opened 2022-04-21
Migrate Husky from 4 to 7

OlivierCavadenti:
- Close https://github.com/knex/knex/pull/4758


kibertoad:
We probably also need to bump husky version?

kibertoad (approved):

PR #5144, opened 2022-04-24
Add development support Typescript source files

code-ape:
> Provided as draft PR for feedback from @kibertoad based on discussion about potentially adding support for Typescript source files.

## Change(s)

1. Add support for Typescript source files in `lib/` which can co-exist alongside the existing `.js` files.
2. Add add `tsconfig.json` file for configuring typescript compilation process.
3. Added dependencies for Typescript compiler in `package.json`
4. Added build commands for Typescript in `package.json`
5. Added `scripts/update_gitignore_for_tsc_output.sh` to generate `lib/.gitignore` file for generated files from Typescript compiler.
6. Documented new development experience with Typescript in `CONTRIBUTING.md`
7. Updated CI to include new build step.
8. Converted `lib/dialects/index.js` to `lib/dialects/index.ts` as small example.


package.json
```diff
@@ -215,6 +219,8 @@
   "files": [
     "bin/*",
     "lib/*",
+    "!lib/**/*.ts",
+    "!lib/.gitigore",
```
kibertoad, line 223: I assume this is a typo

package.json
```diff
@@ -8,6 +8,9 @@
     "node": ">=12"
   },
   "scripts": {
+    "build": "npm run build:gitignore && npm run build:ts",
+    "build:ts": "tsc",
+    "build:gitignore": "./scripts/update_gitignore_for_tsc_output.sh run",
```
kibertoad, line 13: this is fairly hacky and also is not OS-independent. would it be possible to replace it with a JS script?

package.json
```diff
@@ -8,6 +8,9 @@
     "node": ">=12"
   },
   "scripts": {
+    "build": "npm run build:gitignore && npm run build:ts",
+    "build:ts": "tsc",
```
kibertoad, line 12: will generated typings correctly merge with prepackaged knex ones?

package.json
code-ape: No, this was intentional. I put them there so that nothing but the compiled `.js` files would be bundled into the NPM package when published.

package.json
code-ape: Done!

package.json
code-ape: Intention for this is that for Knex 2.x the `.ts` file types are build time only and not included in the published npm package. Presumably a breaking change with Knex 3.x would be to remove the hand maintained `.d.ts` files and start including the ones from the compiler.

package.json
kibertoad: gitigore? not gitignore?

code-ape: Ohhhhh! Yeah 😅 that is a typo. Fixed.

package.json
```diff
@@ -8,6 +8,9 @@
     "node": ">=12"
   },
   "scripts": {
+    "build": "npm run build:gitignore && npm run build:ts",
```
kibertoad, line 11: can you also add `prepublishOnly` hook for this?

code-ape: Done! Added `"prepublishOnly": "npm run build"`.

kibertoad (approved):

OlivierCavadenti:
Hi,
I have an error in index.ts:
image

And I can't launch unit tests, do I need to update something with ts ?

code-ape:
Hi @OlivierCavadenti, only change should be compiling the Typescript files via `npm run build`. This only has to be done when `.ts` files are changed. So if you're coding `.js` files you should only need to run that once after each branch change. Let me know if you did that prior to running tests and are still encountering that issue!

kibertoad:
@code-ape We probably should update readme nothing this, and also automatically do first build on npm install

code-ape:
@kibertoad happy to do that. I added docs in `CONTRIBUTING.md` since there didn't seem to be a "Developer" section of the `README.md`, but happy to start one if you'd like?

PR #5148, opened 2022-04-27
fix: mysql hasColumn Error.(hasColumn `a_id` but hasColumn('a_Id') is false)

KlausBai:
I find this problem when i use strapi who will try to create column `a_id` although i already have a column `a_Id` on mysql.
This is my first pull request, if there are any error, please  give me a change to correct it.

### DIFF
--- a/lib/dialects/mysql/schema/mysql-compiler.js
+++ b/lib/dialects/mysql/schema/mysql-compiler.js
@@ -48,8 +48,8 @@ class SchemaCompiler_MySQL extends SchemaCompiler {
       output(resp) {
         return resp.some((row) => {
           return (
-            this.client.wrapIdentifier(row.Field) ===
-            this.client.wrapIdentifier(column)
+            this.client.wrapIdentifier(row.Field.toLowerCase()) ===
+            this.client.wrapIdentifier(column.toLowerCase())
           );
         });
       },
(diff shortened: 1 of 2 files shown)


OlivierCavadenti:
We are sure we want a case insensitive check ? Why it's bug ?

KlausBai:
Sorry, i can not get the reason of insensitive check, would  you tell we why.
In my mind, mysql Column is not case-sensitive on any platform, and insensitive check  is  not really matching the performance of mysql.For mysql, has a column 'id' means has a column named 'Id' or 'ID' or 'iD' or 'id'.
And there is my problem.
when i want to create a program to automatically add some column to my table, i should not add column when it is already exist to avoid error. If i check the existence of a column named 'a_id' by `hasColumn`, i will get a 'false', but when i try to add a column named 'a_id' to table, mysql will throw an error for had having  a column named 'a_ID'.

KlausBai:
just like this issue about sqlite 
https://github.com/knex/knex/issues/3434

KlausBai:
Sorry, i had not get your point, knex's hasColumn of mysql is case sensitive now

KlausBai:
And sqlite has checked 'hasColumn' by `this.client.wrapIdentifier(col.name.toLowerCase()) ===
            this.client.wrapIdentifier(column.toLowerCase())` why mysql not

OlivierCavadenti (commented):
Ok ! By the way, can you do an integration test for this case ?

KlausBai:
Sure.I have finished it.

KlausBai:
Sure.I have finished it now.

kibertoad:
Released in 2.1.0

PR #5152, opened 2022-05-01
Change string column default value

### LINKED ISSUE #4409: Allow .string() to create varchar field with unlimited length if length is not specified on PostgreSQL
# Environment

Knex version: 0.95.4
Database + version: PostgreSQL 13.2
OS: MacOS 10.15.7

Currently, there is no way to create `varchar` field with unlimited length with the help of `.string()`.

I would expect to get `varchar` with *unlimited length*, if I **do not specify** length like this:

```ts
table.string("name");
```

and `varchar` with *specific length*, if I **specify** length like this:

```ts
table.string("name", 300);
```

which will create `varchar` field with length of `300`.

Currently, Knex gives a default length of `255` for `varchar`, if length is not specified in `.string()`.

Reference: http://knexjs.org/#Schema-string

image

OlivierCavadenti:
- Remove the optional string value default by create an unlimited varchar column instead of 255 default length.
- Closes #4409 
- Doc https://github.com/knex/documentation/pull/401


kibertoad (approved):

OlivierCavadenti:
Still one test that fail, I don't know why (seems count order change...)

```
 1) Integration Tests
       postgresql | pg
         Aggregate
           support the groupBy function:

      AssertionError: expected [ { count: '2' }, { count: '6' } ] to deeply equal [ { count: '6' }, { count: '2' } ]
      + expected - actual

       [
         {
      -    "count": "2"
      +    "count": "6"
         }
         {
      -    "count": "6"
      +    "count": "2"
         }
       ]
      
```

OlivierCavadenti:
@kibertoad @tgriesser rebased on master and tests fixed, CI is green now. Ready for another look.

tgriesser:
@OlivierCavadenti I think the change makes sense, but it'd probably be considered breaking since it changes the behavior of existing migrations. Which is probably not a huge deal but worth considering

Mind if we wait to group into a 4.0 release?

OlivierCavadenti:
Agreed, let's hold it for 4.0 since it changes existing behavior.

PR #5167, opened 2022-05-09
Fix #4869, oracledb correctly INSERTS Buffer

### LINKED ISSUE #4869: Inserting Buffers for Oracle CLOBs silently fails
## Environment

**Knex version:** 0.95.11
**Database + version:** Oracle 18c
**OS:** Ubuntu Linux

If issue is about oracledb support, tag @atiertant. 

# Bug

## Behavior

> Explain what kind of behavior you are getting and how you think it should do

The `oracle-node` driver clearly documents that the best way to insert a large string for a CLOB is to simply hand a buffer of that string to the driver for the insert parameter (reference: https://blogs.oracle.com/opal/post/node-oracledb-112-working-with-lobs-as-string-and-buffer-connection-pinging). However, today when you do this Knex silently replaces the `Buffer` reference with the query string `EMPTY_BLOB()`. Effectively causing a silent failure which inserts no data! 😢

## Error message

None, silent failure.

## Reduced test code

```ts
// Generate insert SQL
const b = Buffer.from('hello', 'utf-8')
const query = knex('table1').insert({ value: b })
const queryText = query.toSQL().toNative()
console.log(query)
```

## Suggested fixes

Initial tests seem to indicate that this can easily be resolved by changing this code:

https://github.com/knex/knex/blob/bfdece3cdc9a029589cee1c8907d1930edace2cc/lib/dialects/oracledb/index.js#L106-L118

It appears that some work has been done to support this but never completed. A quick test by myself found that the following code worked:

```ts
  parameter(value, builder, formatter) {
    if (typeof value === 'function') {
      ...
    } else if (value instanceof BlobHelper) {
      formatter.bindings.push(value.value);
      return '?';
    }
    ...
  }
```

It's unclear to me whether or not this is an elegant solution. I suspect you may want to tinker with the `Buffer` object slightly so that it "pretty prints" better when the query is logged like I did for the example code to reproduce above. Currently it just prints the u8 array which should probably just be replaced by something like `[ Buffer of length ${X}]`. 

## Additional information

Of important note is that the fallback for using `Buffer` is `to_clob([4000 chars])||to_clob([4000 chars])||...` which is around 2,000% slower that using `Buffer` for inserting a 2MB payload.

code-ape:
## Related Issues and PRs
1. #4869

## Change(s)
1. Fixes `parameter` function for Oracledb dialect to properly handle Blobs instead of replacing them with `EMPTY_BLOB()` in `lib/dialects/oracledb/index.js`.
2. Created new test file to verify new behavior in `test/integration2/dialects/oracledb.spec.js`.

### DIFF
--- a/lib/dialects/oracledb/index.js
+++ b/lib/dialects/oracledb/index.js
@@ -111,7 +111,8 @@ class Client_Oracledb extends Client_Oracle {
         this
       );
     } else if (value instanceof BlobHelper) {
-      return 'EMPTY_BLOB()';
+      formatter.bindings.push(value.value);
+      return '?';
     }
     return unwrapRaw(value, true, builder, this, formatter) || '?';
   }
--- a/test/integration2/dialects/oracledb.spec.js
+++ b/test/integration2/dialects/oracledb.spec.js
@@ -1,0 +1,46 @@
+const { expect } = require('chai');
+const { getAllDbs, getKnexForDb } = require('../util/knex-instance-provider');
+
+describe('Oracledb dialect', () => {
+  describe('Connection configuration', () => {
+    const dbs = getAllDbs().filter((db) => db === 'oracledb');
+
+    dbs.forEach((db) => {
+      describe(db, () => {
+        let knex;
+        before(() => {
+          knex = getKnexForDb(db, {
+            connection: {
+              user: 'user',
+              password: 'password',
+              connectString: 'connect-string',
+              externalAuth: true,
+              host: 'host',
+              database: 'database',
+            },
+          });
+        });
+
+        after(() => {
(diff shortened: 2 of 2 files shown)


code-ape:
Hey @kibertoad and @OlivierCavadenti, hope your weeks are going well! Just wanted to drop a quick message in case this has slipped through your inbox. Would love to get this in this week if possible, as we need it for a production fix. Thank you both 🙏

OlivierCavadenti:
@code-ape sorry for the late reply, I will check today. By the way, coverage decrease because it's count lines in test... strange

code-ape:
No worries, thanks so much @OlivierCavadenti 🙏

PR #5174, opened 2022-05-12
Use correct paths in package.json browser field

jrobeson:
The seeder entry is the only one really required to easily build knex for
the browser without shimming `fs`, but I updated the migrations path to
match.

However, there's also `lib/migrations/seed/seed-stub.js`, which actually
seems to provide errors when attempting to use the seed functionality in
a browser, but isn't actually used anywhere

I'm not exactly sure which way is intended.

If the seed stub is intended, then I'll change it to use that, otherwise it should probably be removed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,9 +201,9 @@
     }
   ],
   "browser": {
-    "./lib/migrate/Migrator.js": "./lib/util/noop.js",
+    "./lib/migrations/migrate/Migrator.js": "./lib/util/noop.js",
     "./lib/bin/cli.js": "./lib/util/noop.js",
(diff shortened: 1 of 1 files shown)


kibertoad (approved):

kibertoad:
Thank you!

jrobeson:
but what about the seed stub?

kibertoad:
It was introduced a long time ago, I'm not entirely sure. Probably updating it to be on a safe side makes sense.

jrobeson:
what do you mean updating it?  It should either be used in place of the noop stub or deleted shouldn't it?

Issue #5154, opened 2022-05-02
`dropNullable` gives incorrect SQL on MariaDB 10.2.7+


daniellockyer:
# Environment

Knex version: 2.0.0
Database + version: `10.5.15-MariaDB-0+deb11u1`
OS: Linux

# Bug

refs https://github.com/TryGhost/Ghost/issues/14634

We've ran into a problem in Ghost where we're unable to drop nullable from a column. It specifically seems to be an issue on MariaDB and after investigation, it seems MariaDB 10.2.7+ returns the default value as a string (`COLUMN_DEFAULT` in https://mariadb.com/kb/en/information-schema-columns-table/), which Knex uses for the column info and appends to the column modify query.

[This](https://github.com/knex/knex/blob/2670c2f8b563bb2ddcbf292172af0c57ed02c16f/lib/schema/tablecompiler.js#L328-L337) block of code seems to append `default: 'NULL'`.

My column was created by Knex and the SQL was:

```sql
alter table `newsletters` add `created_at` datetime null
```

after we populate it, we try and drop the nullable using `.dropNullable()`:

```sql
alter table `newsletters` modify  `created_at` datetime not null default 'NULL'
```

Note: `default 'NULL'` is incorrect and causing the issue.

To test, I printed out `val` here: https://github.com/knex/knex/blob/2670c2f8b563bb2ddcbf292172af0c57ed02c16f/lib/dialects/mysql/query/mysql-querycompiler.js#L132

```
{
  TABLE_CATALOG: 'def',
  TABLE_SCHEMA: 'ghost',
  TABLE_NAME: 'newsletters',
  COLUMN_NAME: 'created_at',
  ORDINAL_POSITION: 21,
  COLUMN_DEFAULT: 'NULL',
...
}
```

I would expect the SQL to be:

```sql
alter table `newsletters` modify  `created_at` datetime not null
```

iBotPeaches:
Took a stab at this. Fixed my Maria install and will attempt to get it upstream and solve whatever I need to get it merged.

https://github.com/knex/knex/pull/5181

PR #5181, opened 2022-05-16
Fixes Maria issue with 'NULL' returned instead of NULL on MariaDB 10.2.6+

iBotPeaches:
Fixes: #5154

---

If we look at https://jira.mariadb.org/browse/MDEV-13132, a bug was fixed to properly quote default value to be able to determine expressions vs null. This meant due to the strict typing that `'NULL'` was not detected as null and thus attempted to be included as a default value even against a not-nullable table here - https://github.com/knex/knex/blob/master/lib/schema/tablecompiler.js#L333

I've never worked with this project, but my Ghost blog stopped working and while Ghost is dropping Maria and I'm sad about that. I cannot complain seeing how I paid $0, so I'm attempting to fix to resolve this issue.

I added a test which I believe resolves issue, but to truly replicate you need Maria 10.2.6+ and I don't even see that in the integration suite (the docker). So while I pointed my test to my own Maria and confirmed this works as well as "monkey patching" this fix into my live environment as well. So I am setting this to draft to figure out how you recommend I solve this in terms of "proving" it was solved.

* Should I bring in Maria into integration suite, which might expose a few more issues?
* How should I handle this?

---
(proof of Ghost working after patch)

```
Message: Ghost was able to start, but errored during boot with: alter table `newsletters` modify  `created_at` datetime null default 'NULL' - Invalid default value for 'created_at'

Debug Information:
    OS: Ubuntu, v20.04.4 LTS
    Node Version: v14.19.2
    Ghost Version: 4.47.3
    Ghost-CLI Version: 1.18.1
    Environment: production
    Command: 'ghost restart'

Additional log info available in: xxx

connor@dune:/var/www/connor/blog$ ghost start
+ sudo systemctl start ghost_connortumbleson-com
✔ Starting Ghost
```

### DIFF
--- a/lib/dialects/mysql/query/mysql-querycompiler.js
+++ b/lib/dialects/mysql/query/mysql-querycompiler.js
@@ -131,7 +131,7 @@ class QueryCompiler_MySQL extends QueryCompiler {
       output(resp) {
         const out = resp.reduce(function (columns, val) {
           columns[val.COLUMN_NAME] = {
-            defaultValue: val.COLUMN_DEFAULT,
+            defaultValue: val.COLUMN_DEFAULT === 'NULL' ? null : val.COLUMN_DEFAULT,
             type: val.DATA_TYPE,
             maxLength: val.CHARACTER_MAXIMUM_LENGTH,
             nullable: val.IS_NULLABLE === 'YES',
--- a/test/integration2/query/misc/additional.spec.js
+++ b/test/integration2/query/misc/additional.spec.js
@@ -423,6 +423,27 @@ describe('Additional', function () {
           expect(expectedUuid).to.equal(originalUuid);
         });
 
+        it ('#5154 - should properly mark COLUMN_DEFAULT as null', async function () {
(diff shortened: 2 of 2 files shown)


OlivierCavadenti (approved):
thx for the PR !

kibertoad:
Unfortunately, you can't. This version of knex was not released yet, but even if it was, Ghost uses bookshelf, which still does not support latest knex versions: https://github.com/bookshelf/bookshelf/pull/2120

daniellockyer:
@kibertoad Ghost (I) just decided to run with the latest Knex anyway. It means we're ignoring the peer dependencies but Bookshelf is pretty dead and we're looking at other Knex-based ORMs anyway. I considered forking Bookshelf but what we did was the quickest route.

Once Knex is released, I can bump within Ghost

kibertoad:
Got it, I'll release today then.

kibertoad:
Released in 2.1.0

daniellockyer:
Thanks!

PR #5182, opened 2022-05-16
Fix sqlite returning

Gokaysim:
**Issue**
Client sqlite3 does not return results. It only returns last inserted id as number array. It always return last inserted 

**Cause**
when method is insert, sqlite3 client sends query with 'run' call method. The call method does not allow to return any value.

**Fix**
When method is insert and returning exists, call method is set to 'all' instead of 'run'

### DIFF
--- a/lib/dialects/sqlite3/index.js
+++ b/lib/dialects/sqlite3/index.js
@@ -116,6 +116,12 @@ class Client_SQLite3 extends Client {
     let callMethod;
     switch (method) {
       case 'insert':
+        // if there is any returning column
+        // query should return data
+        if(obj.returning){
+          callMethod = 'all';
+          break;
+        }
       case 'update':
       case 'counter':
       case 'del':


kibertoad:
@Gokaysim Could you please check why CLI tests are failing now?

lib/dialects/sqlite3/index.js
```diff
@@ -116,6 +116,12 @@ class Client_SQLite3 extends Client {
     let callMethod;
     switch (method) {
       case 'insert':
+        // if there is any returning column
+        // query should return data
+        if (obj.returning) {
+          callMethod = 'all';
+        }
+        break;
```
code-ape, line 126: Friendly suggestion. May want to check the version number on the Sqlite driver object is greater than or equal to `3035000` (i.e. 3.35.0) if so do `RETURNING` but otherwise log the error message. This should allow for support of the feature properly regardless of which version of the Sqlite C library the NPM package was built with 😄

Gokaysim: Thanks for suggestion I will check if it is needed then implement it

test/integration2/query/insert/inserts.spec.js
```diff
@@ -126,7 +126,7 @@ describe('Inserts', function () {
                 1,
                 TEST_TIMESTAMP,
               ],
-              [1]
```
tgriesser, line 129: We'll either want this to be opt-in, or need to classify this as a breaking change rather than a fix, requiring a major version bump. 

Since it's only available in newer sqlite versions, I think it might be better handled as a config option where we can opt-in to this behavior for now, and then we can make this a breaking change later. 

I'll see what @kibertoad thinks though since I've been away from the project for quite a bit.

kibertoad: It would be great to make this consistent across all dialects, I think MySQL is also doing it the other way around now.
Config option would require support across all dialects; I prefer a simpler solution, and don't mind going for 3.0.0; not like we are running out of numbers any time soon.

PR #5191, opened 2022-05-24
support singleValue in postgres jsonExtract

ethanj:
I dont have time to create tests, etc, but I figured I would jump start the PR and let you guys finish the details.


mercmobily:
I am commenting on this since an issue was reported matching exactly this PR (#6308) and this PR didn't have an issue attached to it.

Obviously, thanks for opening this. The change is heading in the right direction, but it is not ready to merge yet (as you mentioned it wouldn't). It addresses the core problem, but it is a breaking change and needs more tests, docs, and a couple of fixes.

- In PostgreSQL, jsonExtract now honors `singleValue` and uses `jsonb_path_query_first` when true; otherwise jsonb_path_query.
- Some test expectations are updated to use jsonb_path_query_first.

The change makes sense because:
- MSSQL uses `JSON_VALUE` vs `JSON_QUERY`.
- Oracle uses `json_value` vs `json_query`.
- PostgreSQL already uses `jsonb_path_query_first` in WHERE and JOIN ON comparisons to get a single value for casting and comparison, so aligning `SELECT` behaviour is consistent and avoids SRF problems.

Main concerns:
- Backward compatibility. This is my main one, but it's easy enough to fix. The builder defaults `singleValue` to `true` today, but PostgreSQL ignores it. With this PR, many existing PG queries will silently switch from `jsonb_path_query` to `jsonb_path_query_first` (!), which breaks use cases that relied on SRF behaviour to explode rows or drop unmatched rows. 
- Tests. Yes, OP mentioned that tests were incomplete. so, there are no targeted tests that demonstrate missing path preserves the row with NULL when `singleValue=true`; array paths like `$.data[*]` return only the first element with `singleValue=true` and explode to many rows with `singleValue=false`; passing `singleValue=false` restores SRF behaviour; nesting `jsonExtract` inside `jsonb_set/jsonb_insert `still produces a single jsonb value when `singleValue=true`.
- Redshift expectations. Some updated expectations for pg-redshift now show `jsonb_path_query_first`. Redshift’s compiler actually uses `json_extract_path_text` for extraction , not PG `jsonb_path_*` functions. These expectations should be revisited.
- Docs and types. The docs currently say the `singleValue` flag is for `Oracle` and `MSSQL`. If PG starts honouring it, the docs and type definitions need to be updated, and the change in behavior must be clearly called out.
- Return type differences across dialects. Even with `singleValue=true`, PostgreSQL returns `jsonb` from `jsonb_path_query_first`, while `MSSQL` and `Oracle` return scalars from `JSON_VALUE`/`json_value`. This is fine if documented, but users will often need `#>> '{}'` or explicit casts on PG to get text or numbers. NOTE: In general, I would LOVE to see Knex actually uniform the type of data return, but this is a 4.0 conversation.

Here is what I think should happen with this PR:

- Safer non-breaking option: only switch PG to `jsonb_path_query_first` when the caller explicitly provided the `singleValue` flag as `true`. That likely requires tracking whether the fourth argument was provided instead of defaulting it to `true` in the builder. If `singleValue` was not explicitly provided, keep the current `SRF` behaviour for backward compatibility. We might change things for a 4.0 release
- Add tests that cover missing path with `NULL` and row preserved when `singleValue=true`; array paths behaviour with `true` vs `false`; nesting inside `jsonb_set`/`jsonb_insert`; and ensure Redshift expectations match its compiler.
- Update the `jsonExtract` docs to state that PG honors `singleValue`, clarify return types per dialect, and include examples for casting on PG.

Again, I am here thanks to (because of? :D) #6308 .

PR #5197, opened 2022-05-26
Support Object Names Greater than 30 Characters for Oracle DB Versions 12.2 and Greater

### LINKED ISSUE #3914: Limit is hard coded when it has changed since oracle 12.2
https://github.com/knex/knex/blob/e37aeaa31c8ef9c1b07d2e4d3ec6607e557d800d/lib/dialects/oracle/utils.js#L3

As of their 12.2 release they now support long identifiers up to 128 bytes.

https://docs.oracle.com/en/database/oracle/oracle-database/12.2/newft/new-features.html#GUID-ACC0E96C-8D2D-468F-92E0-2013C7F13455

The tricky part could be that depending on the language a character may be multiple bytes.

https://docs.oracle.com/cd/E11882_01/server.112/e10729/ch2charset.htm#NLSPG171

Thanks!

Kytech:
Since oracle database 12.2, database object names longer than 30 characters has been supported. This PR dynamically sets the maximum character limit for names generated in oracledb based on either the version specified using the `version` attribute in the knex configuration, or it will auto-detect the version number if it is not specified (using a similar technique for auto-detection as is being done in postgres). I added a bit of extra version format processing since oracle version numbers come in both the major.minor format as well as an edition format (such as 9i).

Additional unit and integration tests have been added for the new auto-version-detection features as well as the name max length resolution. I have also updated existing integration and unit tests that relied on the hashed names to expect the standard format since the test oracle database is using version 18.

For older oracle versions, the old hashed name method is still used.

One aspect of this change that may or may not need additional consideration is in the scenario that someone using an older knex version, prior to this change, on oracle 12.2 or greater upgrades to the newest version of knex after this feature is released. Knex will no longer try to convert to the hashed name format when performing schema operations. This would be problematic when a "legacy-named" object is altered or deleted in a migration after upgrading. One potential work-around is to set the version manually to `12.1` in the knex configuration. The other option is providing a means for renaming the old format to the new format. This may make this a breaking change as it may require configuration changes in some use cases.

When running the oracle integration tests, I had several failing, even on master before any of these changes. I have gone through to make sure that all tests that were failing after these changes were the same that were failing before. Unit test results appear to be completing without issue before and after changes.

Closes #3914


lib/dialects/oracle/utils.js
```diff
@@ -1,25 +1,42 @@
...
-      .digest('base64')
-      .replace('=', '');
+class NameHelper {
+  constructor(oracleVersion) {
+    this.oracleVersion = oracleVersion;
+
+    const versionParts = oracleVersion
+      .split('.')
+      .map((versionPart) => parseInt(versionPart));
+    if (
```
kibertoad, line 8: comment explaining the difference would be helpful

lib/dialects/oracledb/index.js
```diff
@@ -165,6 +180,58 @@ class Client_Oracledb extends Client_Oracle {
     return connection.release();
   }
 
+  // In Oracle, we need to check the version to dynamically determine
+  // certain limits.
+  checkVersion(connection) {
```
kibertoad, line 185: why not make this method async?

Kytech: Probably makes sense to mark it async, even though I do still need to wrap the call to execute in a promise due to node-oracledb's use of a callback-style of interface. Makes that top portion a bit more readable and it makes it more obvious that the whole function has an async behavior. I'll add that change.

lib/dialects/oracle/utils.js
Kytech: Let me know if the comment I added is in line with what you were thinking @kibertoad

lib/dialects/oracledb/index.js
```diff
@@ -165,6 +180,58 @@ class Client_Oracledb extends Client_Oracle {
...
+        ? versionString
+        : /^Oracle Database .* ((?:\d+\.){4}\d+).*$/.exec(versionString)[1];
+      // We only care about first two version components at most
+      const versionParts = version.split('.').slice(0, 2);
+      // Strip off any character suffixes in version number (ex. 12c => 12, 12.2c => 12.2)
+      versionParts.forEach((versionPart, idx) => {
+        versionParts[idx] = versionPart.replace(/\D/, '');
+      });
+      return versionParts.join('.');
+    } catch {
```
OlivierCavadenti, line 229: missing error variable:

`} catch(err) {
`

OlivierCavadenti (changes requested):

lib/dialects/oracledb/index.js
Kytech: @OlivierCavadenti This has been fixed.

Kytech:
Just corrected the lint error in the test file I created. Will keep an eye on this for any further feedback since I recently altered the version detection method used in this PR.

erwan-becquet:
Hi there, any idea when this will be included in knex ? thanks in advance.

Kytech:
> Hi there, any idea when this will be included in knex ? thanks in advance.

I do need to update this PR to resolve a few merge conflicts before this can be merged upstream, which I'll be able to do once I am able to finish the current project that demands most of my time at the moment, but other than that, I've addressed all the current feedback from the maintainers, so it ultimately depends on when they will take another look at this PR once I resolve the merge conflicts.

Kytech:
@OlivierCavadenti and @kibertoad I've resolved the merge conflict and it looks like the checks that are failing are the same as those on the current master branch as of when I resolved the merge conflict. Let me know if anything further needs to be done to push this over the finish line.

baileymatthewr:
@Kytech: Perhaps we can get this reviewed again.

Kytech:
> @Kytech: Perhaps we can get this reviewed again.

Looks like the maintainers saw your comment, so I take it that they are aware of the current state of this PR. Just waiting to see if any further changes are needed.

kibertoad:
@Kytech I'm OK to land the change after conflicts are resolved

Kytech:
> @Kytech I'm OK to land the change after conflicts are resolved

Great! I'll get conflicts resolved within the next hour.

Kytech:
@kibertoad merge conflicts have been resolved. Looks like the CI failures are the same ones as on the master branch, I didn't spot anything new. Let me know if you need me to do anything else to get this wrapped up.

kibertoad (approved):

PR #5201, opened 2022-05-30
Explicit jsonb support for custom pg clients

viqueen:
I ran into an issue where my `jsonb` columns are created as `text` when using `pg-mem` in tests and aurora friendly clients like `knex-data-api-client`

It resulted in a lot of `cast blah as jsonb` workarounds, but really, the issue is that under the hood, knex just falls back to `text` data type when it cannot resolve a valid pg version.

This change lets custom clients like `knex-data-api` explicitly state they support `jsonb` instead of passing in a potentially wrong postgres version (think serverless setup)

### DIFF
--- a/lib/dialects/postgres/schema/pg-columncompiler.js
+++ b/lib/dialects/postgres/schema/pg-columncompiler.js
@@ -139,8 +139,8 @@ function jsonColumn(client, jsonb) {
   if (
     !client.version ||
     client.config.client === 'cockroachdb' ||
-    parseFloat(client.version) >= 9.2 ||
-    client.config.jsonbSupport === true
+    client.config.jsonbSupport === true ||
+    parseFloat(client.version) >= 9.2
   ) {
     return jsonb ? 'jsonb' : 'json';
   }
--- a/lib/dialects/postgres/schema/pg-columncompiler.js
+++ b/lib/dialects/postgres/schema/pg-columncompiler.js
@@ -138,7 +138,6 @@ ColumnCompiler_PG.prototype.uuid = 'uuid';
 function jsonColumn(client, jsonb) {
   if (
     !client.version ||
-    client.config.client === 'cockroachdb' ||
     parseFloat(client.version) >= 9.2 ||
     client.config.jsonbSupport === true
   ) {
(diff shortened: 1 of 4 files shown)


OlivierCavadenti:
Ready for review ?

viqueen:
@OlivierCavadenti it's ready , though I think I should leave the check on `cockroachdb` client in place to avoid breaking somebody's world

viqueen:
@kibertoad documentation added over here https://github.com/knex/documentation/pull/423
I added a unit test only.
a meaningful integration test would be with a serverless db like aurora postgres, I'm thinking to wire it up with localstack, but can't sink more time this week.
unit test sounds like a good middle ground for now , what do you think ?

viqueen:
and another ping @kibertoad , is there more to do documentation wise ?
I've already opened this https://github.com/knex/documentation/pull/423 , but it's also not reviewed yet

kibertoad:
sorry for the delay, I was swamped a bit
will take a look over the weekend

kibertoad:
> unit test sounds like a good middle ground for now , what do you think ?
sure, let's go with a unit test

lib/dialects/postgres/schema/pg-columncompiler.js
```diff
@@ -139,7 +139,8 @@ function jsonColumn(client, jsonb) {
   if (
     !client.version ||
     client.config.client === 'cockroachdb' ||
-    parseFloat(client.version) >= 9.2
+    parseFloat(client.version) >= 9.2 ||
+    client.config.jsonbSupport === true
```
kibertoad, line 143:
```suggestion
    client.config.jsonbSupport === true ||
    parseFloat(client.version) >= 9.2
```

test/integration2/util/knex-instance-provider.js
```diff
@@ -132,6 +132,7 @@ const testConfigs = {
     pool,
     migrations,
     seeds,
+    jsonbSupport: true,
```
kibertoad, line 135:
```suggestion
```

kibertoad (approved):

kibertoad:
Released in 2.3.0

PR #5212, opened 2022-06-06
fix(postgresql): add primaryKey option for uuid

### LINKED ISSUE #5211: CockroachDB inserting additional `rowid` column when using `.primary()`
# Environment

Knex version: 2.1.0
Database + version: CockroachDB v22.1 (v21 affected as well)
OS: macOS 12.3.1

# Bug

## Behavior

When using CockroachDB, creating a primary key of type UUID:

```js
await database.schema.createTable("example_a", (table) => {
	table.uuid("id").notNullable().primary();
});
```

will cause a secondary "rowid" column to be autogenerated. This is most likely due to the fact that `knex` generates two SQL queries to create this table+id combo:

```js
db.schema
	.createTable("example_a", (table) => {
		table.uuid("id").notNullable().primary();
	})
	.toSQL();

// [
//   {
//     sql: 'create table "example_a" ("id" uuid not null)',
//     bindings: []
//   },
//   {
//     sql: 'alter table "example_a" add constraint "example_a_pkey" primary key ("id")',
//     bindings: []
//   }
// ]
```

The first SQL query executed in Cockroach will generate the default "rowid" column, as cockroach requires a primary key to exist in the table. This means the issue (and hopefully the fix) is very similar to https://github.com/knex/knex/issues/4141 / https://github.com/knex/knex/pull/5017.

A temporary workaround is to use a `specificType`, for example:

```js
await database.schema.createTable("example_b", (table) => {
	table.specificType("id", "UUID PRIMARY KEY");
});
```

as that generates SQL that correctly creates the primary key in the first query in CRDB:

```js
database.schema
	.createTable("example_b", (table) => {
		table.specificType("id", "UUID PRIMARY KEY");
	})
	.toSQL();

// [
//   {
//     sql: 'create table "example_b" ("id" UUID PRIMARY KEY)',
//     bindings: []
//   }
// ]
```

## Reproduction

https://github.com/rijkvanzanten/knex-crdb-primary

hairmare:
Adds support for creating a uuid field as primary key directly in a create statement using `uuid('id', { primaryKey: true })` for PostgreSQL dialects.

For now the option is only implemented for PostgreSQL dialects since it should work for all of them and it's only really an issue for CockroachDB.

As far as i can tell, using `uuid('id', { primaryKey: true }).primary()` would still work if you need to support other dialects (the `primary()` would just generate a no-op alter statement).

It's getting late here, so i'll add a docs PR tomorrow. Please let me know if the integration test is too hacky and i'll improve on it.

Fixes #5211

### DIFF
--- a/test/integration2/schema/misc.spec.js
+++ b/test/integration2/schema/misc.spec.js
@@ -482,6 +482,47 @@ describe('Schema (misc)', () => {
           });
         });
 
+        describe('uuid types - postgres', () => {
+          before(async () => {
+            await knex.schema.createTable('uuid_column', (table) => {
+              table.uuid('id', { primaryKey: true });
+            });
+          });
+
+          it('#5211 - creates an uuid column as primary key', function () {
+            if (!isPgBased(knex)) {
+              return this.skip();
+            }
+
+            const table_name = 'uuid_columns_test';
+            const expected_column = 'id';
+
+            return knex
+              .raw(
+                'SELECT c.oid FROM pg_catalog.pg_class c WHERE c.relname = ?',
+                [table_name]
+              )
+              .then((res) => {
+                const column_oid = res.rows[0].oid;
+
+                return knex
+                  .raw(
+                    `select c.column_name
+                     from INFORMATION_SCHEMA.COLUMNS c
+                     join INFORMATION_SCHEMA.KEY_COLUMN_USAGE cu
+                     on (c.table_name = cu.table_name and c.column_name = cu.column_name)
+                     where c.table_name = ? and cu.constraint_name like '%_pkey`,
+                    table_name
+                  )
+                  .then((res) => {
+                    const column_name = res.rows[0].column_name;
+
+                    expect(column_name).to.equal(expected_column);
+                  });
+              });
+          });
+        });
+
         describe('increments types - mysql', () => {
           before(() =>
             Promise.all([
--- a/lib/dialects/postgres/schema/pg-columncompiler.js
+++ b/lib/dialects/postgres/schema/pg-columncompiler.js
@@ -124,6 +124,13 @@ class ColumnCompiler_PG extends ColumnCompiler {
       (this.tableCompiler._canBeAddPrimaryKey(options) ? ' primary key' : '')
     );
   }
+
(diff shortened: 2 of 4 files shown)


intech:
The cockroach dialect needs to be able to set the default value to `gen_random_uuid()`. The full SQL line looks like this:
```sql
UUID PRIMARY KEY DEFAULT gen_random_uuid()
```

test/integration2/schema/misc.spec.js
```diff
@@ -482,6 +482,47 @@ describe('Schema (misc)', () => {
...
+
+            const table_name = 'uuid_columns_test';
+            const expected_column = 'id';
+
+            return knex
+              .raw(
+                'SELECT c.oid FROM pg_catalog.pg_class c WHERE c.relname = ?',
+                [table_name]
+              )
+              .then((res) => {
```
OlivierCavadenti, line 505: Use instead await notation

test/integration2/schema/misc.spec.js
```diff
@@ -482,6 +482,47 @@ describe('Schema (misc)', () => {
...
+                const column_oid = res.rows[0].oid;
+
+                return knex
+                  .raw(
+                    `select c.column_name
```
OlivierCavadenti, line 510: Can we select and test tolumn type too ?

hairmare:
I switched to  using await and implemented the column_type check in 3cb450b8aba794946242f9f32fd759ea4e944e29 and added `default gen_random_uuid()` for the CockroachDB dialect in 9010b3f73022c9156850683af76a71e852f6d5f6.

The tests worked locally, can i have another workflow approval here to see if they still work on GitHub Actions?

hairmare:
I also added docs in https://github.com/knex/documentation/pull/419

kibertoad (changes requested):
TS types also need to be updated

hairmare:
@kibertoad thanks for the hint, types have been added in 46b8a699479f7d836268ae46593d4c94f54966ba

I also realised that i had forgotten to register the new CockroachDB unit tests which has now also been fixed.

kibertoad (approved):

kibertoad:
Thank you! I'll publish new version tonight.

PR #5239, opened 2022-06-25
fix: orderBy nulls

### LINKED ISSUE #5135: orderBy nulls argument not correctly applied
# Environment

Knex version: 1.0.7
Database + version: Postgres 12.7
OS: OS X, RDS

# Bug

From the docs: 
```
orderBy — .orderBy(column|columns, [direction], [nulls])
Adds an order by clause to the query. column can be string, or list mixed with string and object. nulls specify where the nulls values are put (can be 'first' or 'last').
```

Problem:
Using `.orderBy('score', 'desc', 'first')` or `.orderBy('score', 'desc', 'last')` does not perform the query the same way as `.orderByRaw('score desc nulls first')` or `.orderByRaw('score desc nulls last)`.   Different results are obtained by the former than the later.   

e.g.
`select * from table order by (score is null) desc` is not equivalent to `select * from table order by score desc nulls last`

The later, `orderByRaw`, provides the correct behavior for my use-case.

OlivierCavadenti:
- Fix orderBy nulls implementation
- Close #5135


rluvaton (commented):
Why existing tests should be changed, did they contain bug as well?

OlivierCavadenti:
> Why existing tests should be changed, did they contain bug as well?

I (re)check them and yes, the existing tests were validating the buggy SQL output....
The old code was only ordering by the null-check boolean expression (e.g. ORDER BY (foo is not null) desc) without including the actual column, so values within each null/non-null group were not sorted at all. 
Also, the null positioning was inverted when using desc direction.
I just rebased on master, and all tests pass.

tgriesser:
@OlivierCavadenti there was another PR opened here related to order by's - I assume this one would supersede that one? https://github.com/knex/knex/pull/6437

OlivierCavadenti:
> @OlivierCavadenti there was another PR opened here related to order by's - I assume this one would supersede that one? #6437

this one is more complete for tests + cases I think (need to check if the other one cover other cases...).

OlivierCavadenti:
Checked it: this one is the most complete. It also covers MSSQL (which #6437 doesn't) and the tests assert the actual row ordering on every dialect. I've ported #6437's guards over too, so nothing is lost, so #6437 could be closed.

@tgriesser

tgriesser:
Seems good to me, wondering if since it technically changes existing behavior if we should group into 4.0?

OlivierCavadenti:
Yes, makes sense to hold it for 4.0 !

PR #5253, opened 2022-07-05
Builder extend() options w/ force

agiangrant:
This PR is to add an Options object to each of the builder.extend functions. Further down the road, the options object can expand without creating breaking changes.

The main motivation for the specific changes was to mitigate breaking changes when new functionalities are incomplete (i.e. the upsert function), but also to allow users to overwrite any of the functionalities if they feel comfortable doing so.

The perfect use case:

knex community adds a new upsert function in the latest version. You have extended the QueryBuilder and created your own version of upsert. When the time comes to update packages, your upsert extension breaks. You can now pass in an options object with `{ force: true }` as the third optional parameter of `knex.QueryBuilder.extend()`. No large refactoring is necessary.

### DIFF
--- a/lib/knex-builder/Knex.js
+++ b/lib/knex-builder/Knex.js
@@ -26,33 +26,35 @@ knex.Client = Client;
 knex.KnexTimeoutError = KnexTimeoutError;
 
 knex.QueryBuilder = {
-  extend: function (methodName, fn) {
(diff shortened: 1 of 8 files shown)


agiangrant:
@OlivierCavadenti please let me know if anything looks out of place for this pull request.

Issue #5269, opened 2022-07-19
Transactor missing property declared in `Knex.Transaction`


ConnorDuvalS:
# Environment

Knex version:
PGSQL + 0.95.15:
OS: macOS 12.4

If issue is about TypeScript definitions, tag @lorefnon.

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do

In the `Knex.Transaction` interface, a 'query' method is listed.

```
interface Transaction<TRecord extends {} = any, TResult = any[]>
    extends Knex<TRecord, TResult> {

    query<TRecord extends {} = any, TResult = void>(
      conn: any,
      sql: any,
      status: any,
      value: any
    ): QueryBuilder<TRecord, TResult>;
}
```

This method does not exist on the transaction objects and is not declared where they're created in `execution/transaction.js`'s `makeTransactor` function.

This method would be useful for either performing injection on the function or just running some direct functionality yourself, though I know of no use for the latter. The design I am attempting to implement essentially needs to run a hook after a transaction has been committed and having access to the transaction's 'query' method would be optimal for this.

**A simple fix** would be to add a `transactor.query = (conn,sql,status,value)=>trx.query(conn,sql,status,value);` anywhere after the `makeKnex` call in the `makeTransactor` function. An adverse solution that would only make the situation worse would be to remove the method declaration from the interface.

2. Error message

JavaScript error, `TypeError: Cannot read properties of undefined (reading 'query')`.

3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real
   database connection to MySQL or PostgreSQL, then single file example which initializes
   needed data and demonstrates the problem.

Example of erroneous behaviour:
```
// in practise, these would be valid values but since the function trying to be called is undefined, it won't matter
const someArguments = ["argument for connection", "argument for sql", "argument for status", "argument for value"];

await knex.transaction(async (transaction) => {
  console.log(transaction.query); // prints undefined
  await transaction.query(...someArguments);
})
```

mercmobily:
Thank you for reporting this issue. This has been added to the TypeScript META issue #6248 where we are tracking all TypeScript-related improvements and fixes. Your issue will be addressed as part of our comprehensive TypeScript support improvements. We are consolidating all TypeScript issues to avoid fragmentation and to tackle the underlying problems systematically.

PR #5270, opened 2022-07-19
Include property that is missing in transaction.js

ConnorDuvalS:
Relevant issue: #5269


kibertoad:
@ConnorDuvalS Could you please add integration test making use of this field in some way?

PR #5272, opened 2022-07-21
fix: js date inserted as a number instead of DATETIME for sqlite3

### LINKED ISSUE #5240: JS Date is inserted as a number instead of DATETIME for SQLite
# Environment

Knex version: 2.1.0
Database + version: Sqlite3 + 5.0.8
OS: Windows 10 Home 21H2 (WSL: 5.10.16.3-microsoft-standard-WSL2)

# Bug

I have a Sqlite3 database with a `datetime` column called `timestamp`. I am performing a batch insert using an array of objects, where the field `timestamp` is a JS Date object:
image

Instead of inserting this as a Sqlite date, Knex seems to insert the data as a number instead (top 2 are Knex, bottom 2 are Sequelize):
image

I have tried using `Date().getTime()` and `Date().toISOString()` just to see if that would fix the issue, but both do not result in a `datetime` type to be inserted into the database.

3. Error message
N/A

5. Reduced test code
I have included a small program that uses both Knex and Sequelize to explain the issue. All of the code is documented and clicking run will print the results to the console:
https://replit.com/@Vandesm14/Knex-DATETIME

Hulxv:
fix #5240

That's the first contribution for me, I hope I did everything right.

### DIFF
--- a/lib/dialects/sqlite3/query/sqlite-querycompiler.js
+++ b/lib/dialects/sqlite3/query/sqlite-querycompiler.js
@@ -9,6 +9,7 @@ const reduce = require('lodash/reduce');
 const QueryCompiler = require('../../../query/querycompiler');
 const noop = require('../../../util/noop');
 const { isString } = require('../../../util/is');
+const { dateToString } = require('../../../util/string');
 const {
   wrapString,
   columnize: columnize_,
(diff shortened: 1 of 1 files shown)


Hulxv:
The errors in tests happen because they compare between js Date object and formated date to string.
So, I think the tests need to some changes to fix it 😅.

ciriousjoker:
Can one of the maintainers take a look at this?

briosheje:
FYI - This seems to be already merged somewhere in main, but the issue I have encountered is that for generic date columns other than timestamp it does work upon inserting, but not upon updating.

When inserting, the value will be converted to string, but when updating a record it will fallback to a timestamp.

tgriesser:
@OlivierCavadenti this seems like it's changing existing behavior - I'd classify this as being part of the next major version bump. Would you agree?

OlivierCavadenti:
> @OlivierCavadenti this seems like it's changing existing behavior - I'd classify this as being part of the next major version bump. Would you agree?

yes sure !
that can lead to inconsistent sort/comparison results on date columns unless users migrate their existing data with this fix....

PR #5292, opened 2022-08-07
Implement mapBinding mssql dialect option

kellyrbourg:
### Changes
- todo

### DIFF
--- a/lib/dialects/mssql/index.js
+++ b/lib/dialects/mssql/index.js
@@ -333,14 +333,24 @@ class Client_MSSQL extends Client {
   _typeForBinding(binding) {
     const Driver = this._driver();
 
+    if (
+      this.connectionSettings.options &&
+      this.connectionSettings.options.mapBinding
+    ) {
+      const result = this.connectionSettings.options.mapBinding(binding);
+      if (result) {
+        return [result.value, result.type];
+      }
+    }
+
     switch (typeof binding) {
       case 'string':
-        return Driver.TYPES.NVarChar;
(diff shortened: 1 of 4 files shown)


lib/dialects/mssql/index.js
```diff
@@ -350,25 +360,25 @@ class Client_MSSQL extends Client {
...
+          return [binding, Driver.TYPES.BigInt];
         }
 
-        return Driver.TYPES.Int;
+        return [binding, Driver.TYPES.Int];
       }
       default: {
         // if (binding === null || typeof binding === 'undefined') {
         //   return tedious.TYPES.Null;
         // }
```
OlivierCavadenti, line 371: can we remove this comment (not in your PR, but can be good to clean up)

OlivierCavadenti (commented):
Why only MsSQL dialect ?

lib/dialects/mssql/index.js
kellyrbourg: > Why only MsSQL dialect ?

Each dialect within knex has it's own sauce for how to go from nodejs types to database specific types. Some even defer to the external library for handling the mapping itself (AFAIK). Executing this PR for all dialects would require each knex dialect/client to own this mapping itself. Would need someone that knows how to do that mapping for the other dialects to help take that on. I'm not sure this is as big of an issue with other dialects either.

> can we remove this comment (not in your PR, but can be good to clean up)
Will do.

kellyrbourg:
Documentation PR created. https://github.com/knex/documentation/pull/440

kibertoad (approved):

PR #5303, opened 2022-08-15
add similar to operator

rachelleannmorales:
To cater the new matching operator: **SIMILAR TO**
https://www.postgresql.org/docs/current/functions-matching.html

### DIFF
--- a/lib/formatter/wrappingFormatter.js
+++ b/lib/formatter/wrappingFormatter.js
@@ -28,6 +28,7 @@ const operators = transform(
     'regexp',
     'not regexp',
     'match',
+    'similar to',
     '&',
     '|',
     '^',


OlivierCavadenti (commented):
Can you add some test for this operator ?

myndzi:
This is just an allow list of "operators we're willing to pass to the database", so I'm okay with not testing that "we put the string in the string". This also matches the current tests: we don't have test coverage that "everything that exists in this list gets sent to the database", but we have implicit coverage that "things that are allowed to be sent do get sent and work". So, I don't think it's fruitful to add tests here.

(If the operator is invalid for the database / version of the database being queried, the database will respond with an error and that's okay)

mercmobily:
LGTM

mercmobily:
@myndzi This needs rebasing and merging. @kibertoad if you approve this, we will rebase/merge.

mercmobily:
Wasn't there an issue that was waiting for this PR to get merged? I can't find it...

mercmobily:
@kibertoad Merged. Trivial change with expected outcome.

PR #5311, opened 2022-08-24
feat(postgres): default datetime/timestamp precision setting added

fabis94:
closes https://github.com/knex/knex/issues/5308

### DIFF
--- a/lib/.gitignore
+++ b/lib/.gitignore
@@ -7,4 +7,4 @@
 **/*.js.map
 
 # Do not include .js files from .ts files
-dialects\index.js
+dialects/index.js
--- a/test/integration2/dialects/postgresql.spec.js
+++ b/test/integration2/dialects/postgresql.spec.js
@@ -1,0 +1,119 @@
+const { expect } = require('chai');
+const {
+  getAllDbs,
+  getKnexForDb,
+  Db,
+  getDbTestConfig,
+} = require('../util/knex-instance-provider');
+
+describe('PostgreSQL dialect', () => {
+  const dbs = getAllDbs().filter((db) => db === Db.PostgresSQL);
+
+  dbs.forEach((db) => {
+    describe(db, () => {
+      let knex;
+      before(() => {
+        knex = getKnexForDb(db);
+      });
+
+      after(async () => {
+        await knex.destroy();
+      });
+
+      describe('when schema building', () => {
+        describe('with default datetime precision specified', () => {
+          const CONFIGURED_DEFAULT_DATETIME_PRECISION = 2;
+
+          let knex;
+          before(() => {
+            const testConfig = getDbTestConfig(db);
+            testConfig.connection.defaultDateTimePrecision =
+              CONFIGURED_DEFAULT_DATETIME_PRECISION;
+            knex = getKnexForDb(db, testConfig);
+          });
+
+          after(async () => {
+            await knex.destroy();
+          });
+
+          const checkColMetadata = async (tableName) => {
+            const result = await knex.schema.raw(`
+            SELECT * FROM information_schema.columns WHERE table_name = '${tableName}'
+            `);
+
+            const keyedResults = {};
+            for (const row of result.rows || []) {
+              keyedResults[row['column_name']] = row;
+            }
+
+            return keyedResults;
+          };
+
+          it('if default datetime precision specified, it gets used when creating datetime/timestamp fields', async () => {
+            const tableName = 'default_datetime_precision_test_' + Date.now();
+            const colTimestamp = 'timestamp_field';
+            const colDatetime = 'datetime_field';
+            const colCreatedAt = 'created_at';
+            const colUpdatedAt = 'updated_at';
+
+            await knex.schema.createTable(tableName, (table) => {
+              table.timestamp(colTimestamp);
+              table.datetime(colDatetime);
+              table.timestamps();
+            });
+
+            const metaData = await checkColMetadata(tableName);
+            expect(metaData[colTimestamp]['datetime_precision']).to.eq(
+              CONFIGURED_DEFAULT_DATETIME_PRECISION
+            );
+            expect(metaData[colDatetime]['datetime_precision']).to.eq(
+              CONFIGURED_DEFAULT_DATETIME_PRECISION
+            );
+            expect(metaData[colCreatedAt]['datetime_precision']).to.eq(
+              CONFIGURED_DEFAULT_DATETIME_PRECISION
+            );
+            expect(metaData[colUpdatedAt]['datetime_precision']).to.eq(
+              CONFIGURED_DEFAULT_DATETIME_PRECISION
+            );
+          });
+
+          it('if default datetime precision specified, it gets used when updating datetime/timestamp fields', async () => {
+            const tableName = 'default_datetime_precision_test_' + Date.now();
+            const colTimestamp = 'timestamp_field';
+            const colDatetime = 'datetime_field';
+
+            const initialPrecision = 5;
+
+            await knex.schema.createTable(tableName, (table) => {
+              table.timestamp(colTimestamp, {
+                precision: initialPrecision,
+              });
+              table.datetime(colDatetime, { precision: initialPrecision });
+            });
+
+            const metaData = await checkColMetadata(tableName);
+            expect(metaData[colTimestamp]['datetime_precision']).to.eq(
+              initialPrecision
+            );
+            expect(metaData[colDatetime]['datetime_precision']).to.eq(
+              initialPrecision
+            );
+
+            await knex.schema.alterTable(tableName, (table) => {
+              table.timestamp(colTimestamp).alter();
+              table.datetime(colDatetime).alter();
+            });
+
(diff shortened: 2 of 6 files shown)


fabis94:
I dunno why, but locally i'm getting random files that I never touched being changed and added to commits: lib/.gitignore and dialects/index.js

I've removed them (hence the multiple commits), but if I try to rebase the branch or anything they appear again. Is this some magic configured together with git hooks?

test/integration2/dialects/postgresql.spec.js
```diff
@@ -0,0 +1,119 @@
+const { expect } = require('chai');
+const {
+  getAllDbs,
+  getKnexForDb,
+  Db,
+  getDbTestConfig,
+} = require('../util/knex-instance-provider');
+
+describe('PostgreSQL dialect', () => {
+  const dbs = getAllDbs().filter((db) => db === Db.PostgresSQL);
```
kibertoad, line 10: so uh, why do you get a list and filter it down to a single value, instead of just starting with that single value??

test/integration2/dialects/postgresql.spec.js
```diff
@@ -0,0 +1,119 @@
+const { expect } = require('chai');
+const {
+  getAllDbs,
+  getKnexForDb,
+  Db,
+  getDbTestConfig,
+} = require('../util/knex-instance-provider');
+
+describe('PostgreSQL dialect', () => {
```
kibertoad, line 9: does this not work with Cockroach?

fabis94: no idea, i have never used Cockroach and don't know much about it

test/integration2/dialects/postgresql.spec.js
fabis94: good point, I was mostly just repeating this procedure the same way it was done in other integration tests

fabis94: @kibertoad so should I just fix this part and then this is ready for merge?

test/integration2/dialects/postgresql.spec.js
kibertoad: can you try ensbling tests for Cockroach too and see if it passes?
crdb extends pg dialect

test/integration2/dialects/postgresql.spec.js
kibertoad: let's try enabling tests for CockroachDB as well and see if that would also pass

lib/.gitignore
```diff
@@ -7,4 +7,4 @@
 **/*.js.map
 
 # Do not include .js files from .ts files
-dialects\index.js
+dialects/index.js
```
fabis94, line 10: this is done automatically by one of the npm scripts, in fact I can't even get rid of this change cause it gets added back every time I revert it

fabis94:
@kibertoad i'd appreciate it if we could get this merged sometime soon 🙌

fabis94:
@kibertoad I see some tests are failing, but the failures don't make sense to me - There are some connection refused errors and errors in tests that don't really seem to have anything to do with my changes, so could it be a test runner issue as well? The tests did run through successfully on my end

OlivierCavadenti:
> @kibertoad I see some tests are failing, but the failures don't make sense to me - There are some connection refused errors and errors in tests that don't really seem to have anything to do with my changes, so could it be a test runner issue as well? The tests did run through successfully on my end

I will check the tests later, but it's seems you do some strange init of DB instances...

fabis94:
> > @kibertoad I see some tests are failing, but the failures don't make sense to me - There are some connection refused errors and errors in tests that don't really seem to have anything to do with my changes, so could it be a test runner issue as well? The tests did run through successfully on my end
> 
> I will check the tests later, but it's seems you do some strange init of DB instances...

Thank you! This is a new & foreign repo for me, so I just tried to look at how other tests are written to write mine

fabis94:
@OlivierCavadenti @kibertoad Hi, any news on this? If there are any changes I need to do, let me know, but please also provide me some providers on the correct way to do things, cause this repo is new to me

fabis94:
Anything I can do to push this through quicker? I'm fine making changes, if anyone can give me pointers :)

test/integration2/dialects/postgresCockroach.spec.js
```diff
@@ -0,0 +1,126 @@
...
+            );
+            expect(metaData[colUpdatedAt]['datetime_precision']).to.eq(
+              CONFIGURED_DEFAULT_DATETIME_PRECISION
+            );
+          });
+
+          // cockroachdb doesn't support altering columns at this point in time? https://github.com/cockroachdb/cockroach/issues/49329
+          if (db !== Db.CockroachDB) {
+            it('if default datetime precision specified, it gets used when altering datetime/timestamp fields', async () => {
+              const tableName = 'default_datetime_precision_test_' + Date.now();
```
rluvaton, line 87: I think it would be cleaner if you could do the skip inline:
```suggestion
         it('if default datetime precision specified, it gets used when altering datetime/timestamp fields', async function () {
              if(db === Db.CockroachDB) {
              	this.skip();
              	return;
              }
              
              const tableName = 'default_datetime_precision_test_' + Date.now();
```

rluvaton (commented):
left a few comments, great job! 
let's resolve those and get this thing merged

@kibertoad what about docs update?

fabis94:
@rluvaton I've resolved both of the issues you brought up

rluvaton (approved):
LGTM

@kibertoad can you please merge

kibertoad:
@fabis94 Tests are failing

fabis94:
@kibertoad @rluvaton The test failures don't make sense to me, it's a bunch of "Error: connect ECONNREFUSED ::1:26257" errors, and I didn't change any of the connection settings so it doesn't feel very related 🤔 https://github.com/knex/knex/actions/runs/7098240971/job/19422356671

kibertoad:
@fabis94 Only new tests are failing, though. See the end of logs at https://github.com/knex/knex/actions/runs/7098240971/job/19422352541?pr=5311

Are they passing locally for you?

test/integration2/dialects/postgresCockroach.spec.js
```diff
@@ -0,0 +1,129 @@
...
+            expect(metaData[colCreatedAt]['datetime_precision']).to.eq(
+              CONFIGURED_DEFAULT_DATETIME_PRECISION
+            );
+            expect(metaData[colUpdatedAt]['datetime_precision']).to.eq(
+              CONFIGURED_DEFAULT_DATETIME_PRECISION
+            );
+          });
+
+          // cockroachdb doesn't support altering columns at this point in time? https://github.com/cockroachdb/cockroach/issues/49329
+          it('if default datetime precision specified, it gets used when altering datetime/timestamp fields', async () => {
```
rluvaton, line 85:
```suggestion
          it('if default datetime precision specified, it gets used when altering datetime/timestamp fields', async function() {
```

rluvaton:
It fails for me as well

rluvaton:
Thank you for the feedback, do you mind opening an issue with all the thing you wish was better?

Are the `with default datetime precision specified` tests are failing in master as well?

myndzi:
Rebased, cleaned up / removed the git mess created by `npm build` to preserve history and avoid merge conflicts, fixed up crdb tests. Ensured `pgnative` is covered in these integration tests and that only the selected databases (with `DB=` env variable) are run.

I lurked in a change to use a deep merge the config with the overrides in `getKnexForDb` to avoid the weird dance to set a connection property in the config override, and to avoid accidental shallow replacement when using config overrides. This should be generally useful in other tests so that they can more directly say what they mean.

Opened https://github.com/knex/knex/pull/6355 to prevent problems with `npm build` checkins in the future too

~~Since this was approved before, I'm expecting the changes to be noncontroversial, but since I changed the code I'm putting it out for review again.~~

Edit: PR still needs docs, and `config.connection` is not the correct place to specify this setting. However, the typescript types don't support dialect-specific options either. I'm going to have to consider whether the lift of making that work is worth it, or if we just live with yet another dialect/client specific config option on the global config bag for now.

lib/.gitignore
myndzi: This wasn't your fault; you followed the instructions for contributions, but they were out of date. I've cleaned up the PR and put a separate PR out to fix the `npm build` command

test/integration2/dialects/postgresql.spec.js
myndzi: They do pass with a caveat: cockroachdb asynchronously applies alter schema (at least in the old version we're testing on) in such a way that sometimes the schema alterations are not "done" in time. I added a delayed, limited retry loop to the section of the test where this mattered and updated the docker-compose file to use an in-memory database ([as recommended in the cockroachdb docs](https://www.cockroachlabs.com/docs/stable/local-testing.html#use-a-local-single-node-cluster-with-in-memory-storage)) to hopefully reduce the likelihood that it matters. If the schema alterations do not converge in ~ 1 second, the test will fail. In local testing I was able to hit 2-3 retries occasionally but not the full 5.

kibertoad (approved):

PR #5316, opened 2022-08-31
feat: support partial unique indexes

### LINKED ISSUE #4870: Feature request: partial index inside unique method
# Environment

Knex version: 0.95.14
Database + version: postgres 12.5
OS: mac

I saw that there is this feature request created [#2525](https://github.com/knex/knex/issues/2525) but was closed. I have also checked that this PR [#2401](https://github.com/knex/knex/pull/2401) add predicate options only for `.index` method. As postgres supports the unique partial index, is it possible to add the predicate inside the `.unique` method too?

abalabahaha:
Adds MSSQL, PostgreSQL, and SQLite support for partial unique indexes, heavily based on #4768.

The Postgres driver defaults to creating a unique constraint instead of an index, so Postgres support for `useConstraint` was added, defaulting to `true` without predicate and `false` with predicate. The alternative would be to always default to `useConstraint: true`, but then `predicate` would error by default unless `useConstraint: false` is explicitly specified. Dunno which is better.

Docs: knex/documentation#449

Fixes #4870

### DIFF
--- a/lib/dialects/mssql/schema/mssql-tablecompiler.js
+++ b/lib/dialects/mssql/schema/mssql-tablecompiler.js
@@ -280,19 +280,23 @@ class TableCompiler_MSSQL extends TableCompiler {
    * Create a unique index.
    *
    * @param {string | string[]} columns
-   * @param {string | {indexName: undefined | string, deferrable?: 'not deferrable'|'deferred'|'immediate', useConstraint?: true|false }} indexName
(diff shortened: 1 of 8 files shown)


kibertoad:
@abalabahaha unit tests are failing

abalabahaha:
@kibertoad sorry about that, I must have missed those since random tests were timing out on my computer. I fixed the failed tests

kibertoad (approved):

enchorb:
Any idea when this will be released?

kibertoad:
This was released in 2.4.0, sorry for the delay

PR #5334, opened 2022-09-21
fix: add missing type for 'expirationChecker' on PgConnectionConfig

sam-super:
expirationChecker currently works with PgConnectionConfig if you ignore the type error.

### DIFF
--- a/test-tsd/pg-config.test-d.ts
+++ b/test-tsd/pg-config.test-d.ts
@@ -3,7 +3,7 @@ import { expectAssignable } from 'tsd';
 import {Knex} from '../types';
 import * as stream from "stream";
 
-const connectionConfig: Knex.PgConnectionConfig = {
+expectAssignable<Knex.PgConnectionConfig>({
   user: '',
   database: '',
   password: '',
@@ -25,6 +25,4 @@ const connectionConfig: Knex.PgConnectionConfig = {
   },
   options: '',
   expirationChecker: () => {return true;},
-};
-
-expectAssignable<Knex.PgConnectionConfig>(connectionConfig);
+});
--- a/test-tsd/pg-config.test-d.ts
+++ b/test-tsd/pg-config.test-d.ts
@@ -1,0 +1,30 @@
+import { expectAssignable } from 'tsd';
+
+import {Knex} from '../types';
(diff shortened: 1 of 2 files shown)


LeoFalco (approved):

types/index.d.ts
```diff
@@ -3024,6 +3024,7 @@ export declare namespace Knex {
     connectionTimeoutMillis?: number;
     types?: PgCustomTypesConfig;
     options?: string;
+    expirationChecker?(): boolean;
   }
```
kibertoad, line 3028: can you add tsd test for this?

sam-super: I'll try. If i'm honest, i didn't do it because I didn't see those tests for the types that already define `expirationChecker`, so this maintains the current standard. I do appreciate it's hard to improve things if we only maintain the current standards tho.

sam-super: @kibertoad i've not done TSD tests before - is this right? https://github.com/knex/knex/pull/5334/commits/2e7fca0093ee1662ede628c7c5dbb72e31f9a7f6

test-tsd/pg-config.test-d.ts
```diff
@@ -0,0 +1,30 @@
...
+  application_name: '',
+  connectionTimeoutMillis: 5,
+  types: {
+    getTypeParser: () => {},
+  },
+  options: '',
+  expirationChecker: () => {return true;},
+};
+
+expectAssignable<Knex.PgConnectionConfig>(connectionConfig);
```
kibertoad, line 30:
```suggestion
expectAssignable<Knex.PgConnectionConfig>({
  user: '',
  database: '',
  password: '',
  port: 1,
  host: '',
  connectionString: '',
  keepAlive: true,
  stream: new stream.Duplex(),
  statement_timeout: false,
  parseInputDatesAsUTC: false,
  ssl: true,
  query_timeout: 2,
  keepAliveInitialDelayMillis: 3,
  idle_in_transaction_session_timeout: 4,
  application_name: '',
  connectionTimeoutMillis: 5,
  types: {
    getTypeParser: () => {},
  },
  options: '',
  expirationChecker: () => {return true;},
});
```

Issue #5341, opened 2022-10-01
bigincrements does not work with composite keys (MySQL)


davidf84:
# Environment

Knex version: 2.3.0
Database + version: MySQL 5.7
OS: Windows

# Bug

1. bigincrements does not work with a composite primary key. This is because the ALTER table command run after the CREATE to add the auto_increment only runs when you define an "increments". Logic is missing for bigincrements.

2. 

3. Example 
```

// this works (using increments)
knex.schema.createTable("test_table", table => {
        table.primary(['userId', 'name']);
        table.increments('userId');
        table.string('name');
}


// OUTPUT
[
  {
    sql: 'create table `test_table` (`userId` int unsigned not null, `name` varchar(255), primary key (`userId`, `name`))',
    bindings: []
  },
  {
    sql: 'alter table `test_table` modify column `userId` int unsigned not null auto_increment',
    bindings: []
  }
]

```
```

// this does not work (using bigincrements)
knex.schema.createTable("test_table", table => {
        table.primary(['userId', 'name']);
        table.bigincrements('userId');
        table.string('name');
}

// OUTPUT - missing the alter table command
[
  {
    sql: 'create table `test_table` (`userId` bigint unsigned not null, `name` varchar(255), primary key (`userId`, `name`))',
    bindings: []
  }
]


```

kibertoad:
Released in 3.1.0

mercmobily:
Thank you for reporting this. This bug has been fixed and released in 3.1.0, as noted in the thread. The MySQL table compiler now applies the AUTO_INCREMENT alter for both `increments` and `bigincrements` when a composite primary key is added after create.

PR #5343, opened 2022-10-02
Fixed issue with bigincrements not working with composite key in MySQL - #5341

davidf84:
Fixes issue with bigincrements not working when having a composite keys since the alter table is not present for bigincrements as descriped in #5341 .

Also fixes issue when using bigincrements and the field was not added as a primary key. The below currently fails since logic to handle bigincrements in the primary key function is missing.

```
knex.schema.createTable("test_table", table => {
        table.primary(['name']);
        table.bigincrements('userId');
        table.string('name');
}
```

### DIFF
--- a/lib/dialects/mysql/schema/mysql-tablecompiler.js
+++ b/lib/dialects/mysql/schema/mysql-tablecompiler.js
@@ -160,6 +160,14 @@ class TableCompiler_MySQL extends TableCompiler {
             }
           });
         }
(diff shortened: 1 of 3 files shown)


rluvaton (approved):
LGTM

rluvaton:
@kibertoad it's looking good

kibertoad:
thanks!

PR #5353, opened 2022-10-12
Fix `migrate:unlock` failing when using `wrapIdentifier` knexfile option

Kytech:
This PR contains a small fix that resolves failure of the `migrate:unlock` command when used with wrapIdentifier. The error reported by `migrate:unlock` in such instances indicates that the lock table does not exist when there is a valid migration lock table detected in the database, as evidenced by a migrate operation failing due to the lock being set.

This issue is caused by failing to disable `wrapIdentifier` processing when running `migrate:unlock`, where other migration operations disable `wrapIdentifier` processing. This creates issues when `wrapIdentifier` alters the casing of an identifier if the database exhibits case-sensitive behavior (ex. such as Oracle when identifiers are quoted inside the query, as is standard practice in knex), resulting in the database treating the non-formatted and formatted identifier versions as two separate tables.

I attempted to add a unit test for this in `test/unit/migrations/migrate/Migrator.js`, though I was having issues getting the tests to run (test would freeze up and timeout), even after going back through, re-adding that test to the unit test config and resolving a few other smaller errors that existed in the file when I attempted to work with it. I didn't dig further since I wasn't sure if this small change warranted additional tests. If we feel a need to add tests for this scenario, I'd be happy to add them if someone could point me in the right direction regarding where would make most sense for testing such functionality.


baileymatthewr:
@Kytech : Could you assign some reviewers to this PR, please?

Kytech:
> @Kytech : Could you assign some reviewers to this PR, please?

I don't have the permissions to add reviewers to a PR, only the maintainers can do that. They probably just haven't triaged this PR to assign it out for review. If needed, I can open an issue and reference this PR to try to get some visibility on this.

Kytech:
Since this was a very simple change, I'm not sure if we need an additional test or not, but if so, I'd be happy to add one if needed.

OlivierCavadenti (approved):

PR #5357, opened 2022-10-14
[ISSUE-5350] Create EXCEPT as a valid keyword

dvsweeney9:
Addresses https://github.com/knex/knex/issues/5350 
Except is a valid SQL keyword https://www.tutorialspoint.com/sql/sql-except-clause.htm 
This is functionality that is blocking my project.

## Summary
*  adds `.except()` as a function to the query builder. `.except()` works exactly like `.union()` or `.intersect()`.
* updates `.intersect()` to use reusable code. 
* Adds unit tests for `.except()` that mimic those of `.intersect()`.
* Updates the knex types.

The documentation PR is available here: https://github.com/knex/documentation/pull/468

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -714,7 +714,6 @@ export declare namespace Knex {
     intersect: Intersect<TRecord, TResult>;
     except: Except<TRecord, TResult>;
 
-
     // Having
     having: Having<TRecord, TResult>;
     andHaving: Having<TRecord, TResult>;


roy-kim-33:
is there any progress on this?

kibertoad:
thank you for the ping, I'll take a look on Friday

lib/query/querybuilder.js
```diff
@@ -816,30 +816,12 @@ class Builder extends EventEmitter {
     return this._union('union all', args);
   }
 
-  // Adds an intersect statement to the query
```
kibertoad, line 819: why was this whole section deleted?

kibertoad (dismissed):
This breaks a lot of tests by removing all assertions from them.

Please add assertions to all new tests and restore them in prior ones.

lib/query/querybuilder.js
dvsweeney9: This section was an exact duplication of the `_union` method, so I refactored it to just call `_union` instead of duplicating the code.

dvsweeney9:
> This breaks a lot of tests by removing all assertions from them.
> 
> Please add assertions to all new tests and restore them in prior ones.

This PR doesn't edit any previous tests, it only indents them to meet linting standards. I'm happy to add new assertions to the old tests to make them usable but was unsure if this was expected.

kibertoad:
@dvsweeney9 Ah, apologies, tests shifted visually a lot, so old tests without assertions look like they replaced the other ones which had assertions.

No need to fix old ones (although that would also be welcome), but if you could add assertions to new tests, that should suffice, and this PR could be merged then.

dvsweeney9:
@kibertoad I ended up adding assertions to the old tests as well as the new tests and restructured them to match the `.intersect()` testing style

dvsweeney9:
@kibertoad quick ping

OlivierCavadenti (approved):

PR #5368, opened 2022-10-25
fix: remove sqlite warning for default value

ylc395:
sqlite support default values. see https://www.sqlite.org/lang_createtable.html#the_default_clause

### DIFF
--- a/lib/dialects/sqlite3/index.js
+++ b/lib/dialects/sqlite3/index.js
@@ -28,14 +28,6 @@ class Client_SQLite3 extends Client {
           '(see docs https://knexjs.org/guide/#configuration-options)'
       );
     }
-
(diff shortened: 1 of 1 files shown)


kibertoad:
Can you enable tests with defaults for sqlite?

ylc395:
Sorry, I misunderstood the meaning of `useNullAsDefault` in knex, which means `Insert into employee(id, name, doj) values(1, ’Aarav’, NULL)` instead of `Insert into employee(id, name, doj) values(1, ’Aarav’, DEFAULT)` when `doj` is `undefined` in javascript.

Sqlite doesn't support `default` in insert statement, indeed. If we want to use the default value in a inserted row, we can only omit that field in insert statement.

PR #5371, opened 2022-10-31
add additional typing for column.index

1mike12:
https://knexjs.org/guide/schema-builder.html#index-1

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2581,7 +2581,15 @@ export declare namespace Knex {
   type lengthOperator = '>' | '<' | '<=' | '>=' | '!=' | '=';
 
   interface ColumnBuilder {
-    index(indexName?: string): ColumnBuilder;
+    index(indexName?: string, indexType?:string): ColumnBuilder;
+    index(
+      indexName?: string,
+      options?: Readonly<{
+        indexType?: string,
+        storageEngineIndexType?: storageEngineIndexType,
+        predicate?: QueryBuilder
+      }>
+    ): ColumnBuilder;
     primary(
       options?: Readonly<{
         constraintName?: string;


OlivierCavadenti:
It seems you have conflict types with SqlLite and Mysql column builder types

OlivierCavadenti (commented):
It seems you have conflict types with SqlLite and Mysql column builder types

myndzi:
It's very abstract. When you call `column.index(...)` it passes through to the TableBuilder method of the same name, prepending the column name:

https://github.com/knex/knex/blob/c8d3eb595bfa81b360488452bc5256a88fcd96f5/lib/schema/columnbuilder.js#L79-L89

This essentially records "the arguments we _will_ use at the compile step", and during compilation those arguments are applied to the specific TableCompiler methods.

From there, _some_ of the dialect-specific TableCompiler `index` methods treat the third argument (`options`) with a kind of fallback / migration behavior where they assume, if it's a string, that it means one specific thing; otherwise, if it's an object, they attempt to destructure the thing they care about from that object:

https://github.com/knex/knex/blob/c8d3eb595bfa81b360488452bc5256a88fcd96f5/lib/dialects/postgres/schema/pg-tablecompiler.js#L241-L245

On the whole, the semantic meaning of "specifying a string as the options argument" varies by dialect, so it's not really reasonable to give it a specific meaning here. The code also smacks of "change over time" - that is, at some point in the past, there was only "one thing to pass here" so it was a string. Then, there became a need to pass "more than one thing", and it was turned into an object - but the string syntax was kept for backwards compatibility.

As far as our publicly-exposed interface goes, we should accept and document only the object-form, and this should not cause problems for users. We can eventually deprecate or remove the string-form.

PR #5373, opened 2022-11-08
[TypeScript] Fix usage of `object` type that is too broad.

VanTanev:
This PR was motivated by knex.raw() not throwing a TS error when passed an `undefined` value.
```ts
knex.raw("?", [undefined]) // expected TS error
```

The reason was the `Value` type including `object`
```ts
  type Value =
    | string
    | number
    | boolean
    | null
    | Date
    | Array<string>
    | Array<number>
    | Array<Date>
    | Array<boolean>
    | Buffer
    | object  // <- culprit!
    | Knex.Raw;
```    

In typescript, `object` means `any JS object, including Arrays`. So in practice, the `Value` type also includes `Array<any>`. This broke the `knex.raw()` inference of bindings passed as an array.

The type has been replaced with `Record<string, any>`, which is conceptually the same, but fixes inference issues for the `RawBuilder`.

I have also fixed one other instance of `object`.

Additionally, I have fixed 3 instances of `Object`.

`Object` is different from `object`.  `Object` means `any non-nullish value`, so it includes strings and numbers. [Example TS payground](https://www.typescriptlang.org/play?#code/MYGwhgzhAECC0G9oF8BQAzArgO2AFwEsB7badIogHgBVoBTADzzuwBMYiAjAKzvwD4AFAAcwAJzABbAFzRqASkQoMOfMVKdxNekxbtoAeR588Q0RJlzFSNBgqC8YzHXkBuVJrEOnLu0UEAcmAB8h7igcGh5P4AjG5hXnFAA)

In the positions it appeard in, it was clear that the intent was not to allow strings and numbers.

----

Many thanks to @vramana on the Matt's TS Wizards discord server who diagnosed the original issue!


myndzi (approved):
Thanks for the detailed explanation, and good callout!

`object` (and `Object`) are definitely... objectionable here 😄

I spent a little time digging into the `merge` case to understand what `Extract<DbRecord<ResolvableTableType<TRecord, 'update'>>, object>` was _actually_ expected to do so that I could assert that the change is safe. I believe it is; as expected, the type here should be an "object of the shape of a database row, when we have the types for it".

I want to leave a note in the issue history to mention that `Value`'s type here probably needs to be split up into "something that is acceptable as a specific value in an SQL query" and, separately, "containers that represent a set of column/value pairs". That's most likely a task for a future TS conversion of the codebase itself, rather than trying to maintain the types in parallel with the code.

kibertoad (approved):

PR #5386, opened 2022-11-17
feat(pg): implement QueryBuilder#updateFrom

wms:
Documentation PR: https://github.com/knex/documentation/pull/476

Introduces the `updateFrom` querybuilder method for PostgreSQL. This allows the creation of `UPDATE <table> SET <data> FROM <name>` queries, which may reference relations or CTEs outside of the target table.

I opted for creating a new QB method (`updateFrom`) to avoid conflict or weird interactions or unexpected changes to behaviour with the target table already being inferred from `knex(table)`, `knex.table(table)`, etc.

### DIFF
--- a/lib/dialects/postgres/query/pg-querybuilder.js
+++ b/lib/dialects/postgres/query/pg-querybuilder.js
@@ -1,6 +1,11 @@
 const QueryBuilder = require('../../../query/querybuilder.js');
 
 module.exports = class QueryBuilder_PostgreSQL extends QueryBuilder {
+  updateFrom(name) {
+    this._single.updateFrom = name;
+    return this;
+  }
+
   using(tables) {
     this._single.using = tables;
     return this;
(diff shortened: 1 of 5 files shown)


wms:
Sorry, looks like I've not fully understood how the testsuite works; its attempting to run pg-specific QB methods against other DB engines.

Is there documentation on how to correctly place my tests so that they only run against the correct DB engine(s)?

OlivierCavadenti (commented):
It's only for PG so in tests you need to skip the test for each other non-PG db.
BY the way, we are sure only PG support update from ? Cannot be added for other DB?
```js
if (!isPostgreSQL(knex)) {
  return this.skip();
}
```

wms:
> BY the way, we are sure only PG support update from ? Cannot be added for other DB?

I've not done an exhaustive search, but I know that (and am in need of support for) PostgreSQL handles this.

OlivierCavadenti (approved):

PR #5417, opened 2022-12-22
1227: add assertion for basic where clause values

antoniopuero:
Meant to mitigate https://github.com/knex/knex/issues/1227

Checked also `whereJsonObject` and it throws the binding error if nested object is given. So this case is handled, but in other way. I can add the other assert there as well to do smth like `Object.values(object).some(isObject)` so the interface is unified somewhat.


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 # 2.3.0 - 31 August, 2022
 
+### Bug fixes
+- Add assertion for basic where clause not to be object or array #1227
+
```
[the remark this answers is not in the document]
antoniopuero: indeed, addressed

test/unit/query/builder.js
```diff
@@ -848,6 +848,66 @@ describe('QueryBuilder', () => {
     });
   });
 
+  it('basic wheres should not accept array or object as a value #1227', () => {
+    testquery(qb().select('*').from('users').where('id', '=', 1), {
+      mysql: 'select * from `users` where `id` = 1',
+      pg: 'select * from "users" where "id" = 1',
+      'pg-redshift': 'select * from "users" where "id" = 1',
+      mssql: 'select * from [users] where [id] = 1',
```
r2dev2, line 856: Instead of "= 1", it should be "= \`1\`" because for mysql, if `= 0` is used, it may match other values. I'm not sure if it should be this way for other databases though, can someone versed in the other databases comment on this?

r2dev2 (changes requested):

lib/query/querybuilder.js
```diff
@@ -441,6 +441,11 @@ class Builder extends EventEmitter {
       }
     }
 
+    assert(
+      !isObject(value),
+      'The values in where clause must not be object or array.'
+    );
```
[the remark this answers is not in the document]
lauripiisang: boolean value `false` , if not passed as a string, yields the same result (talking about MySQL instead of knex here) https://www.db-fiddle.com/f/w8CJ5SAYEKwJqAXyoYPSwC/2

test/unit/query/builder.js
lauripiisang: I tested this on db-fiddle, postgre does not allow comparisons between types with `=` operator and expectedly returns an error.

lib/query/querybuilder.js
```diff
@@ -527,6 +532,12 @@ class Builder extends EventEmitter {
   // Adds a raw `where` clause to the query.
   whereRaw(sql, bindings) {
     const raw = sql.isRawInstance ? sql : this.client.raw(sql, bindings);
+
+    assert(
+      !raw.bindings.some(isObject),
```
OlivierCavadenti, line 537: raw.bindings can be null

OlivierCavadenti (changes requested):
Thanks for the PR !
Check CI + tests

test/unit/query/builder.js
antoniopuero: Yeah, this is scary but it is rather a MySQL issue than a library concern.  Or at least this is a separate case with the type coercion in MySQL while this pr handles the invalid usage of the where clause with objects as plain values.

lib/query/querybuilder.js
antoniopuero: thanks, updated the assert and tests

antoniopuero: btw I decided to move the code to MySQL-specific logic since there was a test suite for pg that actually allows arrays as values in bindings https://github.com/antoniopuero/knex/blame/master/test/unit/query/builder.js#L9974 and PG should handle the error correctly. Not sure about other DB providers tho.

test/unit/query/builder.js
```diff
@@ -9269,7 +9269,7 @@ describe('QueryBuilder', () => {
     }).to.throw('Hint comment must be a string');
   });
 
-  it('Any undefined binding in a SELECT query should throw an error', () => {
+  it.only('Any undefined binding in a SELECT query should throw an error', () => {
```
lauripiisang, line 9272: This looks like dev change made it to branch, you probably want to clear up the `only`

test/unit/query/builder.js
r2dev2: > Yeah, this is scary but it is rather a MySQL issue than a library concern.

I disagree that this is not a library concern. The purpose of knex is to abstract away database-specific quirks to allow safe query-building. Since mysql has the quirk of "= 0" matching all values, knex is responsible for escaping it to "= \`0\`" to represent the developer's intent.

> Or at least this is a separate case with the type coercion in MySQL while this pr handles the invalid usage of the where clause with objects as plain values.

Should adding backticks around values for mysql be added in a different pr then? Two prs addressing the fundamental issue of knex not expecting non-strings through assertions and escaping seems unnecessarily complex, error-prone, and time-consuming compared to calling `.toString()` on inputs.

test/unit/query/builder.js
antoniopuero: Indeed

test/unit/query/builder.js
lauripiisang: > Since mysql has the quirk of "= 0" matching all values, knex is responsible for escaping it to "= `0`" to represent the developer's intent.

@r2dev2 there are some issues with this.
1 - backticks for values is invalid syntax in MySQL, backticks should only be used for identifiers.
2 - when we consider quotes instead of backticks, evne this is a bit contentious, I thought it would break with integers, but MySQL is full of surprises. `= 0` is perfectly valid in case of integers, and for reasonable types, probably should be used. 
Here are the tests related tests: https://www.db-fiddle.com/f/vytxD1kFa3xBiPf7TBADxq/4

image

r2dev2: > 1 - backticks for values is invalid syntax in MySQL, backticks should only be used for identifiers

Ah, it appears I got a bit confused with previous experiments. It is the quote that should be used for values, not backtick. I added a quote escaping test to your fiddle (so it does `= '0'` instead of `= 0` for secret column) and that works properly.

Fiddle: https://www.db-fiddle.com/f/vytxD1kFa3xBiPf7TBADxq/5

sofalse (approved):
Thanks for fixing this!

test/unit/query/builder.js
antoniopuero: The change is not simple toString. In raw queries you need to convert bindings to string as well, while the collection of bindings could be either array or object (named bindings)

r2dev2: > The change is not simple toString. In raw queries you need to convert bindings to string as well, while the collection of bindings could be either array or object (named bindings)

Yea, I tried calling `.toString()` at various places in the codebase and it seems like calling `.toString()` on all the bindings is not a solution as a binding could map to many different situations such as `limit = ?`

Perhaps only calling `.toString()` for `where =` clauses and other operators where the value can be quoted is a solution? One dirty way I did this was by adding

```js
whereBasic(statement) {
  if (['number', 'boolean', 'array', 'object'].includes(typeof statement.value)) {
    statement.value = statement.value.toString();
  }
  return super.whereBasic(statement);
}
```

to `lib/dialects/mysql/query/mysql-querycompiler.js`

However, I'm sure there is a better way of doing this which someone more familiar with the knex codebase than me can think of.

nathanburgess:
What's the timeline on having this merged in?

TimeTerminal:
Merging is blocked until @OlivierCavadenti is able to review it.

test/unit/dialects/mysql.js
```diff
@@ -0,0 +1,91 @@
...
+    client: 'mysql',
+    connection: {
+      port: 1433,
+      host: '127.0.0.1',
+      password: 'yourStrong(!)Password',
+      user: 'sa',
+    },
+  });
+
+  it('should pass with plain values or with emtpy raw bindings', () => {
```
TimeTerminal, line 16: Nit: This should say "empty"

OlivierCavadenti:
@kibertoad maybe we could release a version now ?

kibertoad:
Sure, I'll release it in the morning

piyushchauhan2011:
Cool

sgarbesi:
@kibertoad Thank you.

Ockejanssen:
The only with breaks is that where('column', new Date())

PR #5422, opened 2022-12-29
Custom Naming Convention For Migration Files

sgarbesi:
Documentation: https://github.com/knex/documentation/pull/484

Purpose: Provide the ability to have a custom naming convention for migration files.

Reasoning: Some projects may want to use a different naming convention other than the default for what this library provides.

### DIFF
--- a/lib/migrations/migrate/MigrationGenerator.js
+++ b/lib/migrations/migrate/MigrationGenerator.js
@@ -49,10 +49,10 @@ class MigrationGenerator {
     );
   }
 
-  _getNewMigrationPath(name, getNewMigrationName) {
+  _getNewMigrationPath(name) {
     const fileName = (
-      typeof getNewMigrationName === 'function'
-        ? getNewMigrationName
+      typeof this.config.getNewMigrationName === 'function'
+        ? this.config.getNewMigrationName
         : this._getNewMigrationName.bind(this)
     )(name);
     const dirs = this._absoluteConfigDirs();
@@ -62,8 +62,8 @@ class MigrationGenerator {
 
   // Write a new migration to disk, using the config and generated filename,
   // passing any `variables` given in the config to the template.
-  async _writeNewMigration(name, getNewMigrationName) {
-    const migrationPath = this._getNewMigrationPath(name, getNewMigrationName);
+  async _writeNewMigration(name) {
+    const migrationPath = this._getNewMigrationPath(name);
     await writeJsFileUsingTemplate(
       migrationPath,
       this._getStubPath(),
--- a/lib/migrations/migrate/MigrationGenerator.js
+++ b/lib/migrations/migrate/MigrationGenerator.js
@@ -18,7 +18,10 @@ class MigrationGenerator {
       );
     }
     await this._ensureFolder();
-    const createdMigrationFilePath = await this._writeNewMigration(name);
+    const createdMigrationFilePath = await this._writeNewMigration(
+      name,
+      this.config.getNewMigrationName
+    );
     return createdMigrationFilePath;
   }
 
@@ -41,11 +44,17 @@ class MigrationGenerator {
 
   _getNewMigrationName(name) {
     if (name[0] === '-') name = name.slice(1);
-    return yyyymmddhhmmss() + '_' + name + '.' + this.config.extension.split('-')[0];
+    return (
+      yyyymmddhhmmss() + '_' + name + '.' + this.config.extension.split('-')[0]
+    );
   }
 
-  _getNewMigrationPath(name) {
-    const fileName = this._getNewMigrationName(name);
+  _getNewMigrationPath(name, getNewMigrationName) {
+    const fileName = (
+      typeof getNewMigrationName === 'function'
+        ? getNewMigrationName
+        : this._getNewMigrationName.bind(this)
+    )(name);
     const dirs = this._absoluteConfigDirs();
     const dir = dirs.slice(-1)[0]; // Get last specified directory
     return path.join(dir, fileName);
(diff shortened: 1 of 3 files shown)


mercmobily:
This feels like feature creep. The current timestamp-based naming works well and follows industry standards. Custom naming adds complexity for minimal benefit. Most teams needing special naming can simply rename files after generation or wrap the Knex CLI.

acostalima:
@mercmobily 

> or wrap the Knex CLI.

How? `knex migrate:make <migration_name>` always prepends the timestamp to `migration_name`. It does not seem to be possible to override the migration name generation logic.

mercmobily:
This shows two practical ways to generate custom‑named migration files (e.g., `0001_create_users.js` instead of timestamped) while keeping the standard Knex CLI flow. A live, working demo is here:

- StackBlitz: https://stackblitz.com/~/github.com/mercmobily/knex-named-migrations?file=README.md

The simplest, most ergonomic approach is a tiny runtime shim (“money patch”) you preload when running the CLI. You still pass the migration name exactly as usual; only the prefix logic changes.

## Option A: Preload a tiny shim (keeps the CLI shape)

1) Add this file to your project root as `patch-knex-naming.js`:

```js
// patch-knex-naming.js
// Runtime patch: override Knex MigrationGenerator naming to use sequential prefixes (0001_, 0002_, ...)

const path = require('path');

try {
  // Resolve installed knex package root
  const knexPkgPath = require.resolve('knex/package.json', { paths: [process.cwd()] });
  const knexRoot = path.dirname(knexPkgPath);
  const genPath = path.join(knexRoot, 'lib', 'migrations', 'migrate', 'MigrationGenerator.js');
  const MigrationGenerator = require(genPath);

  if (
    MigrationGenerator &&
    MigrationGenerator.prototype &&
    typeof MigrationGenerator.prototype._getNewMigrationName === 'function'
  ) {
    const orig = MigrationGenerator.prototype._getNewMigrationName;
    MigrationGenerator.prototype._getNewMigrationName = function (name) {
      const fs = require('fs');
      const p = require('path');

      const dir = dirs.slice(-1)[0]; // last configured directory

      try {
        const files = fs.readdirSync(dir);
        const seqRe = /^(\d{4})_.*\.(js|ts|cjs|mjs)$/;
        let max = 0;
        for (const f of files) {
          const m = f.match(seqRe);
          if (m) max = Math.max(max, parseInt(m[1], 10));
        }
        const next = String(max + 1).padStart(4, '0');
        if (name[0] === '-') name = name.slice(1);
        return `${next}_${name}.${this.config.extension.split('-')[0]}`;
      } catch (e) {
        // Fallback to original timestamped name on any error
        return orig.call(this, name);
      }
    };
    console.log('[knex patch] Custom migration naming active');
  }
} catch (e) {
  console.warn('[knex patch] Naming override not applied:', e && e.message);
}
```

2) Ensure your `knexfile.js` sets a migrations directory (so the CLI knows where to write):

```js
// knexfile.js
module.exports = {
  development: {
    migrations: {
      directory: './migrations',
    },
  },
};
```

3) Run the CLI with the shim preloaded.

- POSIX/macOS:
  - NODE_OPTIONS="--require ./patch-knex-naming.js" npx knex migrate:make <name>

- Windows PowerShell:
  - $env:NODE_OPTIONS="--require ./patch-knex-naming.js"; npx knex migrate:make <name>

- Windows cmd (using cross‑env):
  - npx cross-env NODE_OPTIONS="--require ./patch-knex-naming.js" knex migrate:make <name>

Result:
- Files are created under `./migrations` as `0001_<name>.js`, `0002_<name>.js`, … with your `<name>` exactly as passed. Only the prefix logic is changed.

Tip: If you prefer not to type NODE_OPTIONS every time, add an NPM alias:

```json
{
  "scripts": {
    "knex": "NODE_OPTIONS='--require ./patch-knex-naming.js' knex"
  }
}
```

Then run:
- npm run knex -- migrate:make <name>

## Option B: Wrapper script (no patch)

If you’d rather avoid touching internals, you can create a small script that calls `knex.migrate.make(<name>)` and then renames the new file to `000N_<name>.js`.

Minimal variant:

```js
// scripts/mk-migration.js
const fs = require('fs').promises;
const path = require('path');
const knex = require('knex');

(async () => {
  const name = process.argv[2];
  if (!name) {
    console.error('Usage: node scripts/mk-migration.js <name>');
    process.exit(1);
  }

  const migrationsDir = path.resolve(process.cwd(), 'migrations');
  await fs.mkdir(migrationsDir, { recursive: true });

  // No DB driver required if you omit "connection"
  const db = knex({ client: 'sqlite3', migrations: { directory: migrationsDir } });

  const createdPath = await db.migrate.make(name); // produces timestamped file
  const files = await fs.readdir(migrationsDir);
  const seqRe = /^(\d{4})_.*\.(js|ts|cjs|mjs)$/;
  let max = 0;
  for (const f of files) {
    const m = f.match(seqRe);
    if (m) max = Math.max(max, parseInt(m[1], 10));
  }
  const next = String(max + 1).padStart(4, '0');
  const ext = path.extname(createdPath);
  const target = path.join(migrationsDir, `${next}_${name}${ext}`);
  await fs.rename(createdPath, target);
  console.log('Created', target);

  await db.destroy();
})().catch((e) => {
  console.error(e);
  process.exit(1);
});
```

Run:
- node scripts/mk-migration.js create_users

---

## Notes and caveats

- Migration order: Knex runs files in lexicographic order. Numeric prefixes ensure deterministic ordering.
- Your `<name>` is set exactly as usual: it’s the positional argument to `migrate:make`.
- The runtime shim touches a private API (small, surgical). If you prefer zero shims, use the wrapper or a `patch-package` on your installed `knex`.
- If you want “no prefix” instead of `0001_`, change the return value in the shim to `${name}.${...}`.

---

## Demo

- Full working example (including both approaches and an in‑memory runner with pg‑mem):  
  https://stackblitz.com/~/github.com/mercmobily/knex-named-migrations?file=README.md

This has been fully tested end‑to‑end.

PR #5439, opened 2023-01-12
Fix Postgres Malformed array literal 2.4.0 Regression

loren138:
Reverts https://github.com/knex/knex/pull/5321
Fix https://github.com/knex/knex/issues/5365
Fix https://github.com/knex/knex/issues/5430

Adds new test, which I confirmed fails without the changes to `lib/client.js`


capaj (approved):
Thank you @loren138

kibertoad (approved):

Issue #5320, opened 2022-09-02
Can't insert data to json column when array are used as the top-level


mingoing:
# Environment

Knex version:2.3.0
Database + version:pg14
OS:MacOS 12.3.1

# Bug
Array is legal as top-level JSON-text, but I get an error when inserting an array into a JSON column:

``` javascript
await knex.schema.createTable('test', table => {
  table.increments('id').primary()
  table.jsonb('data')
})

// insert failed
await knex('test').insert({
  id: '1000',
  data: [
    { address: '123 Main St', city: 'Anytown', state: 'CA', zip: '12345' },
    { address: '456 Main St', city: 'Anytown', state: 'CA', zip: '12345' }
  ]
})

// insert successfully
await knex('test').insert({
  id: '1001',
  data: { address: '123 Main St', city: 'Anytown', state: 'CA', zip: '12345' }
})
```

mingoing:
PR: https://github.com/knex/knex/pull/5321

mercmobily:
Hi Mingo, thanks for the concise reproduction and for opening the PR.

This has been resolved; support for inserting a top‑level array into a JSON/JSONB column shipped in 2.4.0.

Your PR fixed JSON array binding so that a top‑level array is serialised correctly for Postgres. The CHANGELOG lists “Insert array into json column #5321” under 2.4.0 (CHANGELOG.md:100–120). PR #5321 was merged on 2022‑09‑02 and released in 2.4.0. Upgrading from 2.3.0 to 2.4.0 or later resolves the failure in your example.

Please upgrade to ≥ 2.4.0. If you still see an error after upgrading, share the exact SQL and bindings from `.toSQL()` for us to investigate.

PR #5444, opened 2023-01-13
fix: only stringify JSON if insert value is an array containing plain object

minh-hoang-trinh:
Related to https://github.com/knex/knex/issues/5430 regression

The regression is introduced in [PR](https://github.com/knex/knex/pull/5321) follow from a feature requested [5320](https://github.com/knex/knex/issues/5320)

this PR is to add an extra check for value, the check is now also check if the value is an array and containing at least one plain object. 
the PR is trying to refine the feature implemented by @mingoing (rather than rollback the commit)

also add some tests heavily influenced by the PR from @loren138 (https://github.com/knex/knex/pull/5439)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -401,7 +401,8 @@ class Client extends EventEmitter {
       if (i > 0) str += ', ';
       let value = values[i];
       // json columns can have object in values.
-      if (isPlainObject(value) || Array.isArray(value)) {
+      // check for plain object and array containing at least one plain object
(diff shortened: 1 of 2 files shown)


kibertoad:
Why was this closed?

minh-hoang-trinh:
@kibertoad, closed following the comment by @caseywebdev here https://github.com/knex/knex/issues/5430#issuecomment-1382077189

at the beginning I want to submit a PR to fix a regression introduced by #5320 in our codebase (cannot insert array of string to `text array`) but seem to me that my changes will not work with some other cases.

Issue #5455, opened 2023-01-18
Executable node/r not found on Linux


ldrick:
# Environment

Knex version: 2.4.1
Database + version: any
OS: Ubuntu 

# Bug

1. Running knex on Linux fails pretty early with the same behavior of this bug: #4614 . If I manually change within node_modules/.bin/knex from CRLF to LF it runs smoothly.

2. Error message
`/usr/bin/env: ‘node\r’: No such file or directory
error Command failed with exit code 127.`

kibertoad:
thanks, need to reconfigure new rig

rannooo:
Ouch, this just broke our CICD production release flow. As a workaround I installed `sudo apt install dos2unix` and converted DOS line breaks to Unix line breaks with `dos2unix node_modules/knex/bin/cli.js`

alizeegtd:
I also just got the issue 4 hours ago during migration and seed. Maybe a bug coming from a Knex dependency?

kibertoad:
will fix tonight, it's a result of a different git configuration on my new computer :-/

davesalomon:
A different (temporary) fix until a new version is published:

`package.json` before: `knex migrate:latest`
`package.json` after: `npx --yes crlf --set=LF node_modules/knex/bin/cli.js && knex migrate:latest`

falkenhawk:
@kibertoad v2.4.1 has been released with CRLF (windows line endings) in all files. Line endings should be reverted to LF in all files, not only `cli.js`. renovatebot shows it clearly as all lines have been modified in all files.
https://app.renovatebot.com/package-diff?name=knex&from=2.4.0&to=2.4.1

ripecosta:
@kibertoad Adding a `prettier --lint-different` step to the CI pipeline would catch these changes and avoid them in the future.

yovasx2:
i am having 2.4.1 and still having the crlf problem

kibertoad:
@ripecosta Thank you for the tip, added.
@ldrick can you please check if 2.4.2 solves the issue for you?

ldrick:
@kibertoad Tahnk you very much, it works like a charm.

beenotung:
This issue could be closed?

dawson-conway:
When I download 2.4.2 this is resolved, but when I install 2.5.0 the issue returns and `bin/cli.js` has CRLF endings again.

apoorv254:
[...]

PR #5460, opened 2023-01-22
Improve linting

ripecosta:
Improves linting configs. Runs linting on the repo.

Details:
- Prettier now covers every file in the repo
- On commit hook now runs both Prettier and ESLint
- CI now fails on both Prettier and ESLint errors
- ESLint now uses cache - runs faster locally

Note: The ESLint config currently has a blind spot on `.mjs` and `.ts` files. Fixing that blindspot is out of scope of this PR.


kibertoad (approved):

kibertoad:
it's failing, workflow yamls shouldn't be modified

ripecosta:
> it's failing, workflow yamls shouldn't be modified

I think the issue is the use of "[[" in the npm scripts. Fix incoming.

package.json
```diff
@@ -12,14 +12,13 @@
...
     "build:ts": "tsc",
     "build:gitignore": "node scripts/update_gitignore_for_tsc_output.js run",
-    "format": "prettier --write \"{lib,bin,scripts,test}/**/*.js\"",
-    "format:check-difference": "prettier --list-different \"{lib,bin,scripts,test}/**/*.js\"",
+    "format": "prettier $([ ! $CI ] && echo '--write') --list-different .",
     "debug:test": "mocha --inspect-brk --exit -t 0 test/all-tests-suite.js",
     "debug:tape": "node --inspect-brk test/tape/index.js",
     "coveralls": "nyc report --reporter=lcov",
-    "lint": "eslint \"lib/**/*.js\" \"test/**/*.js\" \"bin/**/*.js\"",
+    "lint": "eslint --cache $([ ! $CI ] && echo '--fix') '**/*.js'",
```
kibertoad, line 19: I don't think echo is cross-platform and is likely not work correctly on Windows. Can you add separate command specifically for CI?

package.json
```diff
@@ -12,14 +12,13 @@
...
     "lint:types": "tsd && dtslint types",
-    "lint:everything": "npm run lint:types && npm run lint",
+    "lint:everything": "npm run format && npm run lint && npm run lint:types",
```
kibertoad, line 21: linting and formatting are different things, they should not be combined

package.json
```diff
@@ -12,14 +12,13 @@
     "clean": "node scripts/clean.js",
     "build:ts": "tsc",
     "build:gitignore": "node scripts/update_gitignore_for_tsc_output.js run",
-    "format": "prettier --write \"{lib,bin,scripts,test}/**/*.js\"",
-    "format:check-difference": "prettier --list-different \"{lib,bin,scripts,test}/**/*.js\"",
+    "format": "prettier $([ ! $CI ] && echo '--write') --list-different .",
```
kibertoad, line 15: separate command for CI, please

package.json
```diff
@@ -58,7 +57,7 @@
     "stress:test": "node scripts/stress-test/knex-stress-test.js | grep -A 5 -B 60 -- '- STATS '",
     "stress:destroy": "docker-compose -f scripts/stress-test/docker-compose.yml stop",
     "prepare": "husky install && npm run clean && npm run build",
-    "prepublishOnly": "npm run format:check-difference && npm run lint && npm run clean && npm run build"
+    "prepublishOnly": "CI=true npm run lint:everything && npm run clean && npm run build"
```
kibertoad, line 60: this will fail on Windows.
also why is this CI? it's run locally

package.json
ripecosta: This is true but since prettier does a little linter I figured it wouldn't be a problem to combine. Do you prefer that I need it separate?

package.json
ripecosta: Can do 👍

package.json
ripecosta: I'll split it out so it's clearer.

package.json
ripecosta: Will do 👍

package.json
kibertoad: yes, otherwise there is a blind spot if CI preformats something and then says oh yeah, it all passes linting

ripecosta: I'm happy to separate it but just to the clear the setup in this PR doesn't have that blind spot. It runs prettier in check mode so it can fail PRs with things like files using the wrong line breaks (like it happened in #5455).

package.json
ripecosta: Done.

package.json
ripecosta: Done.

package.json
ripecosta: Done.

package.json
```diff
@@ -12,14 +12,16 @@
...
     "build:gitignore": "node scripts/update_gitignore_for_tsc_output.js run",
-    "format": "prettier --write \"{lib,bin,scripts,test}/**/*.js\"",
-    "format:check-difference": "prettier --list-different \"{lib,bin,scripts,test}/**/*.js\"",
+    "format": "prettier --write' --list-different .",
+    "format:check": "prettier --list-different .",
     "debug:test": "mocha --inspect-brk --exit -t 0 test/all-tests-suite.js",
     "debug:tape": "node --inspect-brk test/tape/index.js",
     "coveralls": "nyc report --reporter=lcov",
-    "lint": "eslint \"lib/**/*.js\" \"test/**/*.js\" \"bin/**/*.js\"",
+    "lint": "eslint --cache --fix '**/*.js'",
```
kibertoad, line 20: lint by definition is about checking, it would be better to have explicit `lint:fix` command for fixing.
And CI shouldn't be running it.

package.json
ripecosta: Done.

package.json
ripecosta: Sure. But you'll end up with two different patterns for similar things. For ESLint you'll need ":fix" to make changes but not for Prettier. And for Prettier you'll need ":check" for check-only but not for ESLint.

Given the formatter and linter will always both run (in check-mode for CI and fix-mode locally) does it not make more sense to a)  follow the same script naming pattern; and b) be joined the same command ?

PR #5466, opened 2023-01-27
fix() manually close source stream

s-montigny-desautels:
In case where the knex stream is closed manually, the source stream is not closed. This is due to the fact that close event is not propagated up the pipe chain.

The proposed change is to add another close event handler, but only if the stream get pipped to. This close handler manually closes the source stream if it's not already closed.

For the integration test, I don't know if there is a better way to do it. Is there another way to validate that the connection is released from the database? It seems to work fine, since removing the line 81 (`sourceStream.destroy()`) fails with the timeout.

### DIFF
--- a/lib/execution/runner.js
+++ b/lib/execution/runner.js
@@ -72,6 +72,24 @@ class Runner {
       this.client.releaseConnection(this.connection);
     });
 
+    // If the stream is manually destroyed, the close event is not
+    // propagated to the top of the pipe chain. We need to manually verify
+    // that the source stream is closed and if not, manually destroy it.
+    stream.on('pipe', (sourceStream) => {
+      const cleanSourceStream = () => {
+        if (!sourceStream.closed) {
+          sourceStream.destroy();
+        }
+      };
+
+      // Stream already closed, cleanup immediately
+      if (stream.closed) {
+        cleanSourceStream();
+      } else {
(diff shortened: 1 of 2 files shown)


s-montigny-desautels:
I've fixed the tests for mysql and mssql.

The issue that I fixed seems to only present for Postgres, since its the only client using a cursor for the streaming interface.

AlCabohne (approved):

konradpe (approved):

OlivierCavadenti (approved):

pineapplemachine:
I made a repro case here when I ran into issues with query streaming with postgres: https://github.com/pineapplemachine/knex-streaming-error-repro

When I changed the `package.json` Knex dependency to use this branch, like so:

``` js
"knex": "s-montigny-desautels/knex#fix/stream-destroy-not-closing-connection",
```

The script would no longer hang at the same point that it used to, when simulating an error, but would run through completely. Meaning that the stream was closed and the connection was released as expected.

I would very much like to see this fix merged and included in a release!

teobmg:
@OlivierCavadenti can you create a new release with this fix?

OlivierCavadenti:
> @OlivierCavadenti can you create a new release with this fix?

Released in [2.5.0](https://github.com/knex/knex/releases/tag/2.5.0)

PR #5471, opened 2023-01-31
fix(sqlite): append RETURNING statement when insert empty row

ylc395:
when on sqlite:

`knex(tableName).insert({}).returning(['field1', 'field2'])`

No `returning` sql statement was appended.


OlivierCavadenti:
Can you add a test for this case ?

ylc395:
> Can you add a test for this case ?

Done

ylc395:
@OlivierCavadenti Please approve another workflows

ylc395:
@OlivierCavadenti sorry, but another workflows again😭

ylc395:
@OlivierCavadenti hello, what should we do to continue?

PR #5482, opened 2023-02-10
fix: precision lost with bigint in sqlite3

KlausBai:
add supportBigNumbers, bigNumberStrings options like mysql to support sqlite can read bigint by string

### DIFF
--- a/lib/dialects/better-sqlite3/index.js
+++ b/lib/dialects/better-sqlite3/index.js
@@ -1,5 +1,7 @@
 // better-sqlite3 Client
 // -------
+const mapValues = require('lodash/mapValues');
+const isArray = require('lodash/isArray');
 const Client_SQLite3 = require('../sqlite3');
 
 class Client_BetterSQLite3 extends Client_SQLite3 {
@@ -31,7 +33,31 @@ class Client_BetterSQLite3 extends Client_SQLite3 {
     const bindings = this._formatBindings(obj.bindings);
 
     if (statement.reader) {
-      const response = await statement.all(bindings);
+      if(this.connectionSettings.supportBigNumbers) {
+        statement.safeIntegers(true)
+      }
+      let response = await statement.all(bindings);
+      if(this.connectionSettings.supportBigNumbers) {
+        const bigNumberStrings = this.connectionSettings.bigNumberStrings
+        const processBigintModeRow = (row) => {
+          return mapValues(row,v=>{
+            if(typeof v === 'bigint') {
+              const numV = Number(v);
+              if(numV == v) return numV;
+              else {
+                if(bigNumberStrings) return v.toString()
+                else return v
+              }
+            }
+            else return v
+          })
+        }
+        if(isArray(response)) {
+          response = response.map(r=>processBigintModeRow(r))
+        } else {
+          response = processBigintModeRow(response)
+        }
+      }
       obj.response = response;
       return obj;
     }


KlausBai:
In my case, i had a table named 'addresses' with schema like this
image
with data
image
When i try to get this row by below code , I find that the precision of 'test' is just lost.

`const knex = require('../knex')

const instance = knex({
    client: 'better-sqlite3',
    connection: {
      filename: './data.db',
    },
    useNullAsDefault: true
  })

const conn = instance('addresses');

conn.select('*')
conn.first();
conn.where('id',1);
conn.then(res=>console.log({res}))`

image

so I try to add supportBigNumbers, bigNumberStrings options to support knex return bigint value by 'string' , like knex's mysql module do.

In my version, user can just read bigint data by add "supportBigNumbers", "bigNumberStrings" to connection, to get precise data.

`const knex = require('../knex')

const instance = knex({
    client: 'better-sqlite3',
    connection: {
      filename: './data.db',
      supportBigNumbers: true,
      bigNumberStrings: true,
    },
    useNullAsDefault: true
  })

const conn = instance('addresses');

conn.select('*')
conn.first();
conn.where('id',1);
conn.then(res=>console.log({res}))`
image

KlausBai:
@OlivierCavadenti

KlausBai:
@kibertoad can you make a review for me ?

lib/dialects/better-sqlite3/index.js
```diff
@@ -31,7 +33,31 @@ class Client_BetterSQLite3 extends Client_SQLite3 {
...
 
     if (statement.reader) {
-      const response = await statement.all(bindings);
+      if(this.connectionSettings.supportBigNumbers) {
+        statement.safeIntegers(true)
+      }
+      let response = await statement.all(bindings);
+      if(this.connectionSettings.supportBigNumbers) {
+        const bigNumberStrings = this.connectionSettings.bigNumberStrings
+        const processBigintModeRow = (row) => {
```
OlivierCavadenti, line 42: Can be simplified with:

```js
const processBigintModeRow = (row) => {
  return mapValues(row, v => {
    if (typeof v === 'bigint') {
      const numV = Number(v);
      if (numV === v) {
        return numV;
      } else if (bigNumberStrings) {
        return v.toString()
      }
    }
    return v
  })
}
```

OlivierCavadenti (changes requested):

lib/dialects/better-sqlite3/index.js
KlausBai: You are right, this function can be simplified. However, there may be a problem with 'numV === v' because 1n === 1 is false.

myndzi: Using a cast-and-equal check here to make the decision about whether to represent a value as a number or a bigint is undesirable IMO for the following reasons:

- order of coercion is unclear (number == bigint is a type error in typescript for this reason)
- it will return numbers instead of bigints for some values > Number.MAX_SAFE_INT but representable as 64-bit integers (and same for negatives)
- it will return numbers instead of bigints for some values > 2**64 (and same for negatives)

While it _might_ be desirable to use Numbers "when we can", this would be acceptable if-and-only-if the user has no intent to use the values in mathematical expressions. For example, it's okay for ids but is likely to lead to unexpected results when the value is used in a JS mathematical expression. The obvious interpretation of "the library can give me a number or a bigint" is, IMO, "it's a number when it's safe to use as a number, and it's a bigint when it's not safe to use as a number".

If the user is using a bigint column, they should of course be cautious to ensure any math they perform on values that "could need to be bigints" is correct, but the lurking footgun is using "typeof <value>" to make this determination.

If we instead use an inequality to check the range of the number to make the decision about whether to return a bigint or a number, we can define clear semantics for when values will be numbers vs bigints; the semantics should be documented. A bonus is that it's more performant to do it this way than cast-and-check. My preference would be to use the min/max safe integer range.

(an additional note: "processBigintModeRow" should be defined outside of the `_query` method; define it once and allow it to get optimized, rather than define and discard it every query.)

Here's some code demonstrating both the problem and the performance of the two approaches:

```ts
import { run, bench, summary, do_not_optimize } from "mitata";

const byCast = (v: bigint): bigint | number => {
  const numV = Number(v);
  // @ts-ignore
  if (numV == v) return numV;
  return v;
};

const minSafeBI = BigInt(Number.MIN_SAFE_INTEGER);
const maxSafeBI = BigInt(Number.MAX_SAFE_INTEGER);
const byComp = (v: bigint): bigint | number => {
  if (v >= minSafeBI && v <= maxSafeBI) {
    return Number(v);
  }
  return v;
};

// Number(unsafe) == unsafe
// unsafe > Number.MAX_SAFE_INTEGER
const unsafe = 9007199254741002n;

// Number(lucky) != lucky
// lucky > Number.MAX_SAFE_INTEGER - happens to work
const lucky = 9007199254741001n;

// Number(toobig) == toobig
// toobig > 2**64
const toobig = 18446744073709551616n;

const vals = [-1000n, 0n, 1000n, lucky, unsafe, toobig];
const exp = [-1000, 0, 1000, lucky, unsafe, toobig];

import { equal } from "node:assert/strict";
[byCast, byComp].forEach((cast) => {
  // we only verify up to index 4 here, because "unsafe" and "toobig" incorrectly return Numbers
  for (let i = 0; i < 4; i++) {
    const actual = cast(vals[i]!);
    const expected = exp[i]!;
    equal(
      actual,
      expected,
      `${cast.name}/${i}: got ${typeof actual}, expected ${typeof expected}`
    );
  }
});

summary(() => {
  bench("byCast", function* () {
    yield {
      [0]() {
        return vals;
      },
      bench(vals: any[]) {
        return do_not_optimize(vals.map(byCast));
      },
    };
  });
  bench("byComp", function* () {
    yield {
      [0]() {
        return vals;
      },
      bench(vals: any[]) {
        return do_not_optimize(vals.map(byComp));
      },
    };
  });
});

await run();
```

Benchmark results:
```
clk: ~4.68 GHz
cpu: AMD Ryzen 9 6900HX with Radeon Graphics
runtime: node 24.7.0 (x64-linux)

benchmark                   avg (min … max) p75 / p99    (min … top 1%)
------------------------------------------- -------------------------------
byCast                       133.83 ns/iter 132.50 ns  █
                    (129.77 ns … 305.07 ns) 167.58 ns  █
                    (  2.85  b … 383.52  b) 144.97  b ▄█▆▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
                   3.99 ipc ( 96.77% cache)    0.01 branch misses
         625.86 cycles   2.50k instructions    5.16 c-refs    0.17 c-misses

byComp                        67.54 ns/iter  66.33 ns  █
                     (63.38 ns … 204.85 ns) 100.61 ns  █▃
                    (  7.13  b … 373.42  b)  97.13  b ▃██▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
                   3.92 ipc ( 97.16% cache)    0.01 branch misses
         317.74 cycles   1.24k instructions    3.46 c-refs    0.10 c-misses

summary
  byComp
   1.98x faster than byCast
```

myndzi:
As an additional comment: sqlite3 does not actually have a bigint datatype (see: https://sqlite.org/datatype3.html). Rather, any integer column can store values up to 8 bytes depending on its magnitude. Defining a column as an integer does not enforce that the value is within the bigint range, so it might be prudent to apply this bigint logic to any integer column, not just a column defined as a bigint.

myndzi:
Since I got assigned this PR, I dug into the details here and am leaving a state dump for when I come back to this

- Only `better-sqlite3` has bigint support; `sqlite3` does not appear to have a mechanism for doing this
- `better-sqlite3`'s behavior here is to return _all_ numeric values (as known by the database) as either numbers or bigints. This makes sense, because sqlite3 itself isn't supplying any information about its stored datatype and/or affinity in its response to a query(?)
- `better-sqlite3` can _receive_ bigints just fine, and knex doesn't have a problem with sending them, but the Typescript types are complaining in various ways
- It's pretty straightforward to use Knex's `options` method and `connection.options` option-bags to enable bigint mode by default or for a specific query:

     ```diff
     +++ b/lib/dialects/better-sqlite3/index.js
     @@ -10,11 +10,14 @@ class Client_BetterSQLite3 extends Client_SQLite3 {
     // Get a raw connection from the database, returning a promise with the connection object.
     async acquireRawConnection() {
          const options = this.connectionSettings.options || {};
     -
     -    return new this.driver(this.connectionSettings.filename, {
     +    const db = this.driver(this.connectionSettings.filename, {
          nativeBinding: options.nativeBinding,
          readonly: !!options.readonly,
          });
     +    if (typeof options.safeIntegers === 'boolean') {
     +      db.defaultSafeIntegers(options.safeIntegers);
     +    }
     +    return db;
     }

     // Used to explicitly close a connection, called internally by the pool when
     @@ -32,7 +35,13 @@ class Client_BetterSQLite3 extends Client_SQLite3 {
          throw new Error('No connection provided');
          }

     +    const opts = obj.options ?? {};
     +
          const statement = connection.prepare(obj.sql);
     +    if (typeof opts.safeIntegers === 'boolean') {
     +      statement.safeIntegers(opts.safeIntegers);
     +    }
     +
     ```

- It might be preferable to create a specific query builder method to make it more discoverable, but this method would only function for a single client, so may not be desirable
- This pull request is implementing the specific behavior of converting bigints to "either a number or a string" when a bigint is _received_ in a query result. I'm not sure that this behavior belongs in knex itself; it's pretty specific to the desires of the application using knex to perform queries. I'm leaning towards wiring things up so that the types and the library can support a user's ability to receive bigint values, but it's up to the user to perform any conversions they want on them.

I'll come back to this and figure out what to do with it after I've had a chance to talk it over with folks

mercmobily:
@myndzi If you decide to take over this PR, and assuming that the submitter is gone, you should set yourself as DRI here. Note that the PR has: "Maintainers are allowed to edit this pull request.". This effectively means that you can push over the PR and continue on. Right now the developer's dashboard will tell you that you are reviewing it (rightly so!)

PR #5509, opened 2023-03-07
[Sqlite3] Fix can not get table sql because of case sensitive

zydmayday:
## Example

Here is an example how to reproduce the issue.
https://github.com/zydmayday/learn-knex/blob/main/knex/knex.test.ts#L34-L44

## Details

Use sqlite3 as database for testing, and has some migration scripts for setting up tables.

If we have one script create a table called "some_table(https://github.com/zydmayday/learn-knex/blob/main/knex/migrations/1_create_table.ts)",
and another script alter table but using table name "Some_Table(https://github.com/zydmayday/learn-knex/blob/main/knex/migrations/2_add_columns.ts)".
Then after running jest, we will have the following error logs.

```
TypeError: Cannot read properties of undefined (reading 'sql')

      at client.transaction.connection (node_modules/knex/lib/dialects/sqlite3/schema/ddl.js:56:13)
```

## How to fix

Because the related SQL script is case sensitive, change it to run in a case insensitive way.

### DIFF
--- a/test/integration2/migrate/migration-integration.spec.js
+++ b/test/integration2/migrate/migration-integration.spec.js
@@ -81,6 +81,20 @@ describe('Migrations', function () {
             });
         });
 
+        it.only('should not fail alter-table-ignore-case-sensitive', async () => {
+          return knex.migrate
+            .latest({
+              directory:
+                'test/integration2/migrate/alter-table-ignore-case-sensitive',
+            })
+            .then(() => {
+              return knex.migrate.rollback({
+                directory:
+                  'test/integration2/migrate/alter-table-ignore-case-sensitive',
+              });
+            });
+        });
+
         if (isPostgreSQL(knex)) {
           it('should not fail drop-and-recreate-column operation when using promise chain and schema', () => {
             return knex.migrate
--- a/test/integration2/migrate/migration-integration.spec.js
+++ b/test/integration2/migrate/migration-integration.spec.js
@@ -81,7 +81,7 @@ describe('Migrations', function () {
             });
         });
 
-        it.only('should not fail alter-table-ignore-case-sensitive', async () => {
+        it('should not fail alter-table-ignore-case-sensitive', async () => {
           return knex.migrate
             .latest({
               directory:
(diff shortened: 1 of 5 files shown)


zydmayday:
@OlivierCavadenti Thanks for the label, i added the tests.

kibertoad:
If SQLite is case-sensitive, why should knex hide that?
does it behave differently between DDL operations and querying? would same code work case-insensitively for querying?

zydmayday:
> If SQLite is case-sensitive, why should knex hide that? does it behave differently between DDL operations and querying? would same code work case-insensitively for querying?

Hi @kibertoad 
Thanks for replying.
Actually it is not because sqlite3 is case sensitive.
It is because the query in function `getTableSql` is case sensitive.

When we try to get some tables DDL, we don't care if we pass the UPPERCASE or lowercase of table name.
In fact, sqlite3 is not case sensitive as I searched.

kibertoad:
got it, thanks!

test/integration2/migrate/migration-integration.spec.js
```diff
@@ -81,6 +81,20 @@ describe('Migrations', function () {
             });
         });
 
+        it.only('should not fail alter-table-ignore-case-sensitive', async () => {
+          return knex.migrate
```
kibertoad, line 85: * please rewrite it to use async await
* please add assertion, which would only pass if both migrations completed successfully

zydmayday: Thanks, updated test case as requested

OlivierCavadenti:
@zydmayday I will check coveralls configuration before merge that. It's seems coverage decreased computation take into account not only coverage of code, but also coverage of tests... I will not ask you to make a test for test :D.

OlivierCavadenti (approved):

zydmayday:
> @zydmayday I will check coveralls configuration before merge that. It's seems coverage decreased computation take into account not only coverage of code, but also coverage of tests... I will not ask you to make a test for test :D.

@OlivierCavadenti
Thanks! Good to know the reason! haha

PR #5517, opened 2023-03-10
migration status command doesn't attempt to create any tables

nampas:
Addresses suggestion [here](https://github.com/knex/knex/issues/5516).

### DIFF
--- a/lib/migrations/migrate/migration-list-resolver.js
+++ b/lib/migrations/migrate/migration-list-resolver.js
@@ -1,5 +1,5 @@
 const { getTableName } = require('./table-resolver');
-const { ensureTable } = require('./table-creator');
+const { knexTablesExist } = require('./table-creator');
 
 // Lists all available migration versions, as a sorted array.
 function listAll(migrationSource, loadExtensions) {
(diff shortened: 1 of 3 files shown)


lib/migrations/migrate/migration-list-resolver.js
```diff
@@ -9,12 +9,14 @@ function listAll(migrationSource, loadExtensions) {
...
 async function listCompleted(tableName, schemaName, trxOrKnex) {
-  await ensureTable(tableName, schemaName, trxOrKnex);
-
-  return await trxOrKnex
-    .from(getTableName(tableName, schemaName))
-    .orderBy('id')
-    .select('name');
+  // If the internal tables (migrations and locks) don't exist, then
+  // we haven't run any migrations at all.
+  return (await knexTablesExist(tableName, schemaName, trxOrKnex))
```
OlivierCavadenti, line 14: I think this breaks migrations because of removing of ensureTable.
That not only check tables but create them with specific schema and insert a default lock row.
I think we still need to default creation tables

OlivierCavadenti (changes requested):

PR #5530, opened 2023-03-28
Add readonly option to Better-SQLite3

undees:
Better-SQLite3 [supports a `readonly` option](https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#new-databasepath-options) on initialization. This PR lets Knex users set that option.

(The automatic code formatter wanted to reflow some lines unrelated to the new code. I've split those automated changes out into a separate commit for clarity and to preserve author attribution on the changed lines.)

Docs PR is in knex/documentation#505.

### DIFF
--- a/test/unit/dialects/better-sqlite3.js
+++ b/test/unit/dialects/better-sqlite3.js
@@ -12,12 +12,18 @@ describe('better-sqlite3 unit tests', () => {
         connection: {
           filename: ':memory:',
           options: {
-            nativeBinding: path.resolve(__dirname, '../../../node_modules/better-sqlite3/build/Release/better_sqlite3.node'),
+            nativeBinding: path.resolve(
+              __dirname,
+              '../../../node_modules/better-sqlite3/build/Release/better_sqlite3.node'
+            ),
           },
         },
       });
 
-      const result = await knexInstance.select(knexInstance.raw('2 + 2 as answer')).first().catch(err => err);
+      const result = await knexInstance
+        .select(knexInstance.raw('2 + 2 as answer'))
+        .first()
+        .catch((err) => err);
       expect(result.answer).to.equal(4);
     });
 
@@ -33,7 +39,10 @@ describe('better-sqlite3 unit tests', () => {
         },
       });
 
-      const result = await knexInstance.select(knexInstance.raw('2 + 2 as answer')).first().catch(err => err);
+      const result = await knexInstance
+        .select(knexInstance.raw('2 + 2 as answer'))
+        .first()
+        .catch((err) => err);
       console.dir(result);
       expect(result).to.be.an('error');
       expect(result.code).to.equal('MODULE_NOT_FOUND');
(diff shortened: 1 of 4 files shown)


test/unit/dialects/better-sqlite3.js
```diff
@@ -33,10 +39,44 @@ describe('better-sqlite3 unit tests', () => {
...
+        connection: {
+          filename: '/Users/erindees/src/knex/test/test.sqlite3',
+          options: {
+            readonly: true,
+          },
+        },
+      });
+
+      const connection = await knexInstance.client.acquireConnection();
+      expect(connection.readonly).to.equal(true);
```
OlivierCavadenti, line 66: can you add a clause that try to execute a query that modify data to check readonly option is take into account ?

test/unit/dialects/better-sqlite3.js
```diff
@@ -33,10 +39,44 @@ describe('better-sqlite3 unit tests', () => {
...
     });
   });
+
+  describe('readonly', () => {
+    it('should initialize the DB with the passed-in `readonly` option', async () => {
+      const knexInstance = knex({
+        client: 'better-sqlite3',
+        useNullAsDefault: true,
+        connection: {
+          filename: '/Users/erindees/src/knex/test/test.sqlite3',
```
OlivierCavadenti, line 58: filename doesn't exists so unit test fails

OlivierCavadenti (changes requested):

test/unit/dialects/better-sqlite3.js
undees: Yes, definitely. I originally wrote a test case for this, but deleted it because I thought it might be too "integration-y" for a unit test. But it is a good idea to check this, so I've restored it. Thank you for the feedback!

test/unit/dialects/better-sqlite3.js
undees: Whoops, meant to change those paths to relative before committing. That's what I get for pushing code when I'm tired. 😅 Fixed.

OlivierCavadenti:
Actions seems broken and we have not checks, I will try to fix that before merging.

OlivierCavadenti:
Hi,

I have this error in tests, it's seems I can create a database with readonly: false (like before) but with readonly : true;

```
SqliteError: unable to open database file
    at new Database (node_modules\better-sqlite3\lib\database.js:63:26)
    at Client_BetterSQLite3.acquireRawConnection (lib\dialects\better-sqlite3\index.js:14:12)
    at create (lib\client.js:254:39)
```
@undees

undees:
Hi, @OlivierCavadenti. Sorry about this! Looks like the tests were only passing intermittently, depending on whether a previous test case created `test/unit/test.sqlite3` before the read-only case looked for it. I've opened #5608 to fix the broken test case, and just for good measure I (re-)verified manually in Node that Better-SQLite3 respects the `readonly` option.

Issue #2983, opened 2019-01-06
Migrations: Add lifecycle callbacks


jimsimon:
# Environment

Knex version: 0.16.3
Database + version: all
OS: Mac OSX

# Feature discussion / request

1. Explain what is your use case
I'd like to be able to interact with the connection and/or transaction before any migrations are run.  This would allow for better customization of migration behavior when using them programmatically.  In my specific case, I'd like to be able to run `SET SCHEMA xyz` on the transaction before migrations are run so that I can more easily update a schema based multitenant database.  In my database, each tenant gets their schema, and each schema has the same exact tables.  So I need to run the migrations once per schema.

2. Explain what kind of feature would support this
The ability to specify a callback function that is called just after the transaction is created (or connection if transactions are disabled) as part of the config object that can be passed to `knex.migrate.latest()`, `knex.migrate.rollback()`, etc.

3. Give some API proposal, how the feature should work
```javascript
async function migrateSchemas () {
  const schemas = ['abc', 'def', 'ghi']
  for (const schema of schemas) {
    await knex.migrate.latest({
      schemaName: schema,  // Needed for knex to find the right migrations table
      beforeMigrations: async function (knexOrTrx) {
        await knexOrTrx.raw('SET SCHEMA ??', [schema])
      }
      // Other lifecycle function ideas might be "beforeEachMigration", "afterEachMigration", "afterMigrations"
    })
  }
}
```

jimsimon:
My current workaround for not being able to set the schema prior to migrations running is to create a new knex instance with a custom `searchPath` for each schema migration.  While this approach works, it's a lot clunkier and eats up extra resources.

elhigu:
If I understood correctly this use case was already covered with https://github.com/tgriesser/knex/issues/2799

jimsimon:
This isn't quite what I'm looking for.  In my case, I don't want the migration author to have to manually specify the schema as it would be pretty easy to forget to add the `.withSchema` call.  The goal is to abstract away the tenant concept so that the migrations appear as though they were written for a single database.  I proposed generic lifecycle methods as a solution so that they may be useful for other use cases aside from mine (or for other considerations I may have like setting the user's role before migrations).

kibertoad:
@jimsimon Can't you just use knex.withUserParams for that? (https://github.com/knex/documentation/pull/135)

jimsimon:
[...]

PR #5541, opened 2023-04-12
Add migration lifecycle hooks

timorthi:
This is an attempt at addressing the use cases described in #2983 . 

This adds the ability for the user to define a `beforeAll`, `afterAll`, `beforeEach`, `afterEach` hook during migrations.


lib/migrations/migrate/Migrator.js
```diff
@@ -583,12 +602,9 @@ function getNewMigrations(migrationSource, all, completed) {
   });
 }
 
-function checkPromise(logger, migrationPromise, name, commitFn) {
+function checkPromise(logger, migrationPromise, name) {
   if (!migrationPromise || typeof migrationPromise.then !== 'function') {
     logger.warn(`migration ${name} did not return a promise`);
-    if (commitFn) {
-      commitFn();
-    }
```
timorthi, line 591: I removed the explicit commit so that we can run the `afterEach` hook in the same transaction (line 527)

test/integration2/migrate/migration-integration.spec.js
```diff
@@ -527,7 +528,131 @@ describe('Migrations', function () {
           });
         });
 
+        describe('knex.migrate.latest - lifecycle hooks', function () {
```
rluvaton, line 531: Could you please add tests that:
1. if one of the hooks fails the migration is rolled back
2. If one of the hook fails the next hooks/migrations are not executed:
    1. if `beforeAll` failed, the `beforeEach`, the migration itself, `afterEach` and `afterAll` are not called
    2. if `beforeEach` failed, the migration itself, `afterEach` and `afterAll` are not called
    3. if the migration itself failed, the `afterEach` and `afterAll` are not called
    4. if the `afterEach` failed, `afterAll` is not called
3. test with all the hooks and check execution order - `beforeAll` -> `beforeEach` -> `afterEach` -> `afterAll`
4. no hooks are being called if no pending migrations

rluvaton, line 531: I think this should be in a separate file for life cycle hooks

test/integration2/migrate/migration-integration.spec.js
```diff
@@ -527,7 +528,131 @@ describe('Migrations', function () {
...
+            });
+          });
+        });
+
         describe('knex.migrate.rollback', function () {
+          before(() => {
+            return knex.migrate.latest({
+              directory: 'test/integration2/migrate/test',
+            });
+          });
```
rluvaton, line 654: Why is this needed all of a sudden?

rluvaton (commented):
Looking good! thanks for the great job!
could you please also update the types?

left few comments and I also think docs should be updated as well (Cc @kibertoad)


Also few questions:
1. If you have 2 migrations, and the second one failed, should you run the `afterAll` should you rollback the `beforeAll`?
2. In case transaction are disabled for migrations, should we allow hooks?

test/integration2/migrate/migration-lifecycle-integration.spec.js
```diff
@@ -0,0 +1,317 @@
...
+              'afterEach-20131019235242_migration_1.js',
+              'beforeEach-20131019235306_migration_2.js',
+              'afterEach-20131019235306_migration_2.js',
+              'afterAll',
+            ]);
+          });
+        });
+
+        // NOTE (PR #5541) This is disabled pending further discussion
+        describe.skip('when there are no pending migrations', function () {
```
timorthi, line 287: This test is currently failing because there appears to be no short-circuiting check on the length of new migrations returned by `getNewMigrations`. https://github.com/knex/knex/blob/e2bc612179a8971d167bbb0d571f51fe4e602eac/lib/migrations/migrate/Migrator.js#L74

Is there a specific reason we don't short-circuit, or was this an oversight?

test/integration2/migrate/migration-integration.spec.js
timorthi: Great suggestions, thank you!

timorthi:
@rluvaton 


> 1. If you have 2 migrations, and the second one failed, should you run the afterAll should you rollback the beforeAll?

In this scenario, I feel like it may be appropriate to abandon the afterAll and roll back the beforeAll. 

> 2. In case transaction are disabled for migrations, should we allow hooks?

I can't personally think of a situation in which someone would want to run these hooks without also running their migrations in a transaction, though I wouldn't be surprised if that's a use case for someone out there. Maybe for the initial release, we can require transactions to run hooks, and see if anyone raises an issue with that.

test/integration2/migrate/migration-lifecycle-integration.spec.js
rluvaton: I think without the hooks it did not matter, but now it does... you should add short circus

timorthi: Done!

timorthi: Done! I've also relocated the `beforeAll`/`afterAll` hook calls to a place that's more DRY.

test/integration2/migrate/migration-lifecycle-integration.spec.js
```diff
@@ -0,0 +1,303 @@
...
+
+            await knex.migrate
+              .latest({
+                directory: 'test/integration2/migrate/test',
+                beforeEach,
+                afterEach,
+                afterAll,
+              })
+              .catch((error) => {
+                expect(error.message).to.equal('force beforeEach hook failure');
```
rluvaton, line 158: If the function won't reject this won't be called, saved the error in a value as well as if rejected or not

test/integration2/migrate/migration-lifecycle-integration.spec.js
```diff
@@ -0,0 +1,303 @@
...
+                hasSecondTestTable: false,
+              },
+              {
+                hasFirstTestTable: true,
+                hasSecondTestTable: false,
+              },
+            ]);
+          });
+
+          it('does not run the migration or the afterEach/afterAll hooks if the hook fails', async function () {
```
rluvaton, line 143:
```suggestion
          it('does not run the migration and the afterEach/afterAll hooks if the hook fails', async function () {
```

test/integration2/migrate/migration-lifecycle-integration.spec.js
```diff
@@ -0,0 +1,303 @@
...
+const { getAllDbs, getKnexForDb } = require('../util/knex-instance-provider');
+
+describe('Migrations Lifecycle Hooks', function () {
+  getAllDbs().forEach((db) => {
+    describe(db, () => {
+      let knex;
+
+      // Force clean slate before each test
+      beforeEach(async () => {
+        rimraf.sync(path.join(__dirname, './migration'));
```
rluvaton, line 19: Why sync?

rluvaton (commented):
Could you please add test when transaction are disabled for migration?

lib/migrations/migrate/Migrator.js
```diff
@@ -496,30 +502,40 @@ class Migrator {
...
           ) {
             this.knex.enableProcessing();
-            return this._transaction(
-              this.knex,
-              migrationContent,
-              direction,
-              name
-            );
+            return await this.knex.transaction(async (trx) => {
+              await beforeEach(trx, [migration]);
```
timorthi, line 519: The migration that's passed into the hook is wrapped in an array here to maintain consistent types between the `each` and `all` hooks (instead of passing in an array in `all` but a migration object in `each`)

test/integration2/migrate/migration-lifecycle-integration.spec.js
timorthi: Not 100% sure - this setup block was copied from `migration-integration.spec.js` and I chose not to change anything.

timorthi:
@rluvaton 

> Could you please add test when transaction are disabled for migration?

I took a closer look at the code and it appears that the user is allowed to customize whether each migration is run in a transaction or not: https://github.com/knex/knex/blob/f7a31477e203dd9b00c36b841de204c5d109f6c1/lib/migrations/migrate/Migrator.js#L480-L490

Because of this, I think it may be simpler to just run the lifecycle hooks all the time, even if transactions are disabled. Otherwise, we will be dealing with a bunch of edge cases regarding when or when not to run lifecycle hooks. What do you think and would you still like me to test the non-transaction case if we choose to always run hooks?

rluvaton:
Can you please update the types

timorthi:
@rluvaton I've updated `index.d.ts`, did I miss any other type definitions?

For CI, it looks like this particular step (Legacy Types Linting) has been failing for the last 3 weeks, unrelated to this PR https://github.com/knex/knex/actions/workflows/linting-legacy.yml

rluvaton (approved):
Thank you for your contribution

michaelwiles:
Please can this be merged? I need this as well.

m33ch:
Hello and thank you all for your contribution! I really needed this feature.

I noticed that the change has been merged into master but I see that a new version hasn't been pubilshed to npm.
Approximately, when is a new release on npm expected?

Thanks a lot!

PR #5559, opened 2023-05-01
fix(security): avoid password leaks on query logs

### LINKED ISSUE #5560: DB passwords can be accidentally leaked through logs
**Disclaimer:** I'm reporting this problem here because there are no public policies on how to report security problems. I would have preferred doing this privately.

# Environment

Knex version: 2.4.2
Database + version: MySQL, PostgreSQL, MSSQL (I didn't test with others. DB version does not have any effect)
OS: Mac OS X, Linux

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do:
   ```typescript
   const query = knex('myTable').select('*');
   logger.debug({ query }) // passwords are leaked here
   
   // For a more realistic case:
   try {
     const result = await query
     return result
   } catch (err) {
     logger.error({ err, query }, 'Unknown Error while performing query') // passwords are leaked here
   }
   ```
   The previous code should not leak passwords through the logs, but it does 😢 .
   In our case, the logger was Pino, this problem doesn't necessarily happen with other loggers (as they might use different
   serialisation methods).

3. Error message: There is no error message

4. Reduced test code:
   ```typescript
   // This new test could be placed in `test/integration2/query/misc/additional.spec.js`
   it('should not leak passwords when logged', async function () {
      const util = require('util')
      const query = knex('test_table_two').select('*');
      const fakeLog = util.inspect(query, { depth: null });
      const passwordMatches = fakeLog.match(
        /(knextest|testpassword|S0meVeryHardPassword|testrootpassword)/g
      );

      expect(passwordMatches).to.be.null;
   });
   ```

castarco:
## Motivation:
We detected that under certain conditions, passwords could be leaked through logs, when executing code such as:
```typescript
const query = knex('myTable').select('*');
logger.debug({ query }) // passwords are leaked here
```
In our case, the logger was `Pino`, this problem doesn't necessarily happen with other loggers (as they might use different
serialisation methods).

## Fixes:
- password fields are not leaked anymore. Fixes #5560 .
- accepts configuration objects where the password field exists but it is non-enumerable (this might happen when lib users are trying to prevent the same previous problem, but at a higher level).

## Changes:
- marks the password field as non-enumerable when possible, to avoid potential leaks through logs.
- adapts tests
- add new regression test

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -44,8 +44,8 @@ class Client extends EventEmitter {
     this.config = config;
     this.logger = new Logger(config);
 
-    if (config.connection.password) {
-      setHiddenProperty(config.connection);
+    if (this.config.connection && this.config.connection.password) {
+      setHiddenProperty(this.config.connection);
     }
 
     //Client is a required field, so throw error if it's not supplied.
@@ -74,7 +74,7 @@ class Client extends EventEmitter {
       this.connectionConfigExpirationChecker = () => true; // causes the provider to be called on first use
     } else {
       this.connectionSettings = cloneDeep(config.connection || {});
-      if (config.connection.password) {
+      if (config.connection && config.connection.password) {
         setHiddenProperty(this.connectionSettings, config.connection);
       }
       this.connectionConfigExpirationChecker = null;
--- a/lib/client.js
+++ b/lib/client.js
@@ -31,6 +31,7 @@ const { POOL_CONFIG_OPTIONS } = require('./constants');
 const ViewBuilder = require('./schema/viewbuilder.js');
 const ViewCompiler = require('./schema/viewcompiler.js');
 const isPlainObject = require('lodash/isPlainObject');
+const { setHiddenProperty } = require('./util/security.js');
 
 const debug = require('debug')('knex:client');
 
@@ -44,10 +45,7 @@ class Client extends EventEmitter {
     this.logger = new Logger(config);
 
     if (config.connection.password) {
-      Object.defineProperty(config.connection, 'password', {
-        enumerable: false,
-        value: config.connection.password,
-      });
+      setHiddenProperty(config.connection);
     }
 
     //Client is a required field, so throw error if it's not supplied.
@@ -77,10 +75,7 @@ class Client extends EventEmitter {
     } else {
       this.connectionSettings = cloneDeep(config.connection || {});
       if (config.connection.password) {
-        Object.defineProperty(this.connectionSettings, 'password', {
-          enumerable: false,
-          value: config.connection.password,
-        });
+        setHiddenProperty(this.connectionSettings, config.connection);
       }
       this.connectionConfigExpirationChecker = null;
     }
@@ -320,24 +315,14 @@ class Client extends EventEmitter {
       debug('acquired connection from pool: %s', connection.__knexUid);
       if (connection.config) {
         if (connection.config.password) {
-          Object.defineProperty(connection.config, 'password', {
-            enumerable: false,
-            value: connection.config.password,
-          });
+          setHiddenProperty(connection.config);
         }
         if (
           connection.config.authentication &&
           connection.config.authentication.options &&
           connection.config.authentication.options.password
         ) {
-          Object.defineProperty(
-            connection.config.authentication.options,
-            'password',
-            {
-              enumerable: false,
-              value: connection.config.authentication.options.password,
-            }
-          );
+          setHiddenProperty(connection.config.authentication.options);
         }
       }
       return connection;
--- a/lib/client.js
+++ b/lib/client.js
@@ -43,6 +43,13 @@ class Client extends EventEmitter {
     this.config = config;
     this.logger = new Logger(config);
 
+    if (config.connection.password) {
+      Object.defineProperty(config.connection, 'password', {
+        enumerable: false,
+        value: config.connection.password,
+      });
+    }
+
     //Client is a required field, so throw error if it's not supplied.
     //If 'this.dialect' is set, then this is a 'super()' call, in which case
     //'client' does not have to be set as it's already assigned on the client prototype.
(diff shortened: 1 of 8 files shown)


castarco:
Hi @tgriesser , I'm sorry for adding a direct mention, but I have the feeling that this PR fixes a serious issue, and it can be easy that this PR gets lost among the dozens of automated PRs produced by bots.

lib/client.js
```diff
@@ -43,6 +43,13 @@ class Client extends EventEmitter {
     this.config = config;
     this.logger = new Logger(config);
 
+    if (config.connection.password) {
+      Object.defineProperty(config.connection, 'password', {
+        enumerable: false,
+        value: config.connection.password,
+      });
```
OlivierCavadenti, line 50: Put that inside a global function and use it everywhere please

OlivierCavadenti:
@castarco Thanks for the PR by the way, the error is on the "master" branch and not in your PR.

castarco:
Thanks @OlivierCavadenti  😃 , I was investigating now what I might have broken. It will save me some time!

lib/util/security.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Sets a hidden (non-enumerable) property on the `target` object, copying it
+ * from `source`.
+ *
+ * This is useful when we want to protect certain data to be accidentally leaked
```
castarco, line 5: I'll fix the grammatical error here

test/integration2/query/misc/additional.spec.js
```diff
@@ -1400,6 +1422,16 @@ describe('Additional', function () {
             expect(knexDb.client.pool.destroyed).to.equal(false);
           });
         });
+
+        it('should not leak passwords when logged', async function () {
+          const query = knex('test_table_two').select('*');
+          const fakeLog = util.inspect(query, { depth: null });
+          const passwordMatches = fakeLog.match(
+            /(knextest|testpassword|S0meVeryHardPassword|testrootpassword)/g
```
castarco, line 1430: Should I add a comment pointing to the origin of these passwords?

lib/util/security.js
castarco: Done

castarco:
Any news on whether this will be approved/rejected/merged? From our side it's not urgent, because we are using `yarn patch` in our project, but this PR fixes a vulnerability that can affect other people...

Eomm:
@OlivierCavadenti any update on this?

OlivierCavadenti (approved):

OlivierCavadenti:
@Eomm sorry for the delay.
I will release a new version as soon as possible (this week), but I need time to also check documentation.
Stay tuned :D

daniellockyer:
Is it possible that this has broken auth in 2.5.0? We've started to get auth issues for the 2.4.0 -> 2.5.0 Knex bump in Ghost: https://github.com/TryGhost/Ghost/actions/runs/5505546016/jobs/10033054604?pr=17254#step:11:283

```
  Error: Access denied for user 'root'@'172.17.0.1' (using password: NO)
      at /home/runner/work/Ghost/Ghost/node_modules/knex-migrator/lib/database.js:176:39
      at Packet.asError (/home/runner/work/Ghost/Ghost/node_modules/mysql2/lib/packets/packet.js:728:17)
      at ClientHandshake.execute (/home/runner/work/Ghost/Ghost/node_modules/mysql2/lib/commands/command.js:29:26)
      at Connection.handlePacket (/home/runner/work/Ghost/Ghost/node_modules/mysql2/lib/connection.js:489:32)
      at PacketParser.onPacket (/home/runner/work/Ghost/Ghost/node_modules/mysql2/lib/connection.js:94:12)
      at PacketParser.executeStart (/home/runner/work/Ghost/Ghost/node_modules/mysql2/lib/packet_parser.js:75:16)
      at Socket.<anonymous> (/home/runner/work/Ghost/Ghost/node_modules/mysql2/lib/connection.js:101:25)
      at Socket.emit (node:events:390:28)
      at addChunk (node:internal/streams/readable:315:12)
      at readableAddChunk (node:internal/streams/readable:289:9)
      at Socket.Readable.push (node:internal/streams/readable:228:10)
      at TCP.onStreamRead (node:internal/stream_base_commons:199:23)
```

OlivierCavadenti:
Hi, can you check this @castarco ? https://github.com/knex/knex/issues/5629

castarco:
I'll check. It could be related.

OlivierCavadenti:
> I'll check. It could be related.

Thanks !

elszczepano:
> Is it possible that this has broken auth in 2.5.0? We've started to get auth issues for the 2.4.0 -> 2.5.0 

Yes, it looks exactly the same as in my case 😄

castarco:
I didn't had time this past night to fix it as I was lower on energy, but I was walking through the code and I already have some idea of how to fix it. Hopefully today I'll be able to post a patch.

OlivierCavadenti:
> I didn't had time this past night to fix it as I was low on energy, but I was walking through the code and I already have some idea of how to fix it. Hopefully today I'll be able to post a patch.
> 
> Again, sorry for the disturbance.

No problem, we all deal with our personal life, do as you can

idandagan1:
Hi, 
I used to pass the db connection password using the package "config" with custom variables from outside the app.

But after upgrading to v2.5.0, it fails with the error:
```
Cannot redefine property: password
TypeError: Cannot redefine property: password
    at Function.defineProperty (<anonymous>)
    at setHiddenProperty 
```
So what's your suggestion for injecting the password?

Fryuni:
Just as a note, this broke our integrations tests since we now cannot reuse the same object with the configurations between scenarios.

We have to clone the object for each instantiation of knex otherwise we get `SCRAM-SERVER-FIRST-MESSAGE: client password must be a string` from all but the first instantiation.

Easy fix on our side, but just registering here so other might know.

bertho-zero:
This is a bad PR, the original object is modified and the password is lost.

PR #5585, opened 2023-05-24
fix: config.log bind functions to config.log

Naymi:
problem:
image

image

cause:
image

Solution:
Just bind custom log function to config.log context


Naymi:
Sorry...
Updated

Naymi:
Fail not from me?

PR #5590, opened 2023-05-30
All I get are timeouts when using dropNullable and setNullable for mysql

DavidTanner:
I can run the raw command to add/drop nullable for a column, but when running using the utility method I get a timeout.


OlivierCavadenti (changes requested):
some unit test fails

DavidTanner:
I'm not really sure why this was failing in my setup.  The situation is more complex than I thought, and this PR as is won't cut it.

PR #5608, opened 2023-06-26
Ensure test DB exists at correct path in better-sqlite3 tests

undees:
Follow-up to #5530. Better-SQLite3 tests weren't passing on all platforms due to a path / sequencing issue. This PR addresses the issue by using `path.resolve()` instead of constructing paths by string addition, and by ensuring that an empty database exists no matter what order the tests run in.


undees:
Hi, @OlivierCavadenti! Sorry about the test issue; this PR should get them passing again.

OlivierCavadenti (approved):

PR #5617, opened 2023-07-01
Add uuid helper function

arbezerra:
Meant to be used in the schema creation:
```javascript
await knex.schema
    .createTable('users', table => {
      table.uuid('id').defaultTo(knex.fn.uuid());
      table.string('user_name');
    })
```

Can someone confirm that there is no uuid function in Redshift?

### DIFF
--- a/test/integration2/query/misc/additional.spec.js
+++ b/test/integration2/query/misc/additional.spec.js
@@ -392,6 +392,25 @@ describe('Additional', function () {
           expect(knex.fn.now(6).toQuery()).to.equal('CURRENT_TIMESTAMP(6)');
         });
 
+        it('should allow using .fn-methods to generate a uuid with select', async function () {
+          const uuid = await knex.select(knex.raw(knex.fn.uuid() + ' as uuid'));
+          expect(uuid[0].uuid).to.match(
+            /^[0-9A-F]{8}-[0-9A-F]{4}-[1-5][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i
+          );
+        });
+
+        it('should allow using .fn-methods to be a default value', async function () {
+          await knex.schema.dropTableIfExists('default_uuid_table');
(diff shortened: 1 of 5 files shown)


test/integration2/query/misc/additional.spec.js
```diff
@@ -390,6 +390,25 @@ describe('Additional', function () {
...
+        it('should allow using .fn-methods to be a default value', async function () {
+          await knex.schema.dropTableIfExists('default_uuid_table');
+          await knex.schema.createTable('default_uuid_table', (t) => {
+            t.uuid('uuid').defaultTo(knex.fn.uuid());
+          });
+          await knex('default_uuid_table').insert({});
+          const uuid = await knex('default_uuid_table').select('uuid');
+          expect(uuid[0].uuid).to.match(
+            /^[0-9A-F]{8}-[0-9A-F]{4}-[1-5][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i
+          );
```
OlivierCavadenti, line 409: Can you add cases that test the piece of query generated for each dialect ?

OlivierCavadenti (changes requested):
It's not easy in Amazon Redshift unfortunaly:
http://datasamurai.blogspot.com/2016/06/creating-uuid-function-in-redshift.html

arbezerra:
The coverage decreased because oracle and redshift aren't tested. Should I make a test to cover those lines or put a comment to ignore coverage?

OlivierCavadenti:
> The coverage decreased because oracle and redshift aren't tested. Should I make a test to cover those lines or put a comment to ignore coverage?

Can you add test for OracleDB (same that Oracle) and for Redshift that test that the uuid function throw an Error with "Redshift does not have a uuid function" and it's seems good !

Can you also add a documentation ? 
https://github.com/knex/documentation

arbezerra:
I think everything is ok now.
Opened PR knex/documentation#527

OlivierCavadenti (approved):

PR #5641, opened 2023-07-24
Updates the type of ColumnBuilder.index to match the documentation

aman-atakama:


atlanteh:
In addition to this I think `index(indexName: string, indexType: string)` should also be added

Issue #4910, opened 2021-12-25
.foreign & .withSchema ?


ash0080:
# Environment

Knex version:  0.95.15
Database + version: 
OS:
PostgreSQL 12.7 (Debian 12.7-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do
```
k.schema
	.withSchema('test1')
	.createTable('u_test', t => {
		t.increments('id').primary()
		t.foreign('test_id').references('id').inTable('test')
			.onDelete('CASCADE')
			.onUpdate('NO ACTION')
		t.timestamps()
	})
```
when I use .withSchema(), of course I want the foreign key to refer to test.id under the **same** schema, however, 42P01 appears here because the actual references are default **public** schema

although I could fix it like this,
```.inTable('test1.test')```
But it doesn't seem to be compliant
I think it would be more syntactic to either modify the presentation of .withSchema() or to provide a .foreign() sibling modifier.

2. Error message
42P01

3. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real
   database connection to MySQL or PostgreSQL, then single file example which initializes
   needed data and demonstrates the problem.

rajivkr:
Facing the same issue, any updates on this ? it would be great if we the callback to the `createTable` function picks the schema from the `withSchema` method

ir33k:
Agree, I think that `foreign()` should use schema from `createTable()`.  Also it should be possible to alter schema name using `withSchema()` on `foreign()` if you need to reference table from different schema (which is the problem I'm facing).

mercmobily:
Thank you for reporting this issue. This is a duplicate of #1614, which has already been triaged and labeled as an NG_bug.

Both issues describe the same problem: when using `withSchema()` with foreign key references, the schema prefix is not automatically applied to the referenced table, causing PostgreSQL to look for the table in the default (public) schema instead of the specified schema.

The issue will be resolved by PR #6169, which has been updated to include automatic schema inheritance. When you use `withSchema('test1')` and create a foreign key reference like `table.foreign('test_id').references('id').inTable('test')`, the schema context will now be automatically applied, correctly generating a reference to `test1.test.id` rather than `public.test.id`.

[...]

PR #5651, opened 2023-08-03
Fix .foreign to utilize withSchema setting

ryanrobertsname:
When using `table.foreign()` it does not utilize the schema from `knex.schema.withSchema()`. As a result, the foreign key ends up referencing a table with no schema specified.

This seems like fragile behavior and I'm not sure what the preferred work around would be.

Also see #4910.

This PR makes `table.foreign()` utilize the schema from `knex.schema.withSchema()` (if set).

### DIFF
--- a/lib/schema/tablecompiler.js
+++ b/lib/schema/tablecompiler.js
@@ -100,7 +100,11 @@ class TableCompiler {
         : this._indexCommand('foreign', this.tableNameRaw, foreignData.column);
       const column = this.formatter.columnize(foreignData.column);
       const references = this.formatter.columnize(foreignData.references);
-      const inTable = this.formatter.wrap(foreignData.inTable);
+      const inTable = this.formatter.wrap(
+        foreignData.schema
+          ? `${foreignData.schema}.${foreignData.inTable}`
+          : foreignData.inTable
+      );
       const onUpdate = foreignData.onUpdate
         ? (this.lowerCase ? ' on update ' : ' ON UPDATE ') +
           foreignData.onUpdate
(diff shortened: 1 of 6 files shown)


EduardJoy:
Could someone take a look at this one?

lib/schema/tablecompiler.js
```diff
@@ -100,7 +100,11 @@ class TableCompiler {
         : this._indexCommand('foreign', this.tableNameRaw, foreignData.column);
       const column = this.formatter.columnize(foreignData.column);
       const references = this.formatter.columnize(foreignData.references);
-      const inTable = this.formatter.wrap(foreignData.inTable);
+      const inTable = this.formatter.wrap(
+        foreignData.schema
+          ? `${foreignData.schema}.${foreignData.inTable}`
```
rluvaton, line 105: @kibertoad is it ok to do this string concatenation (security-wise) I looked at the `wrap` function and it looks ok, just wanna double check

rluvaton (commented):
Please instead of adding tests to existing test, create a new ones

mercmobily:
@ryanrobertsname Are you still around to work on this?

mercmobily:
See #6169

mercmobily:
@ryanrobertsname Thank you for identifying this issue and providing a solution. Since you haven't been active recently and PR #6169 provides a complementary fix with an active author, we're going to proceed with that PR and incorporate your schema inheritance idea into it.

Your contribution helped identify the need for automatic schema inheritance, which will be included in the final solution. Thank you for your work on this.

Closing in favour of #6169.

Issue #5358, opened 2022-10-15
Knex typings incompatible with TS 4.7's moduleResolution: NodeNext


ethanresnick:
# Environment

Knex version: 2.3.0
Database + version: pg 13
OS: Mac OS

@lorefnon.

# Bug

In Typescript 4.7, there was [a new mode added](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#esm-nodejs) to support the package.json `exports` keyword. With this mode enabled in a project that uses knex, TS [can no longer correctly find](https://github.com/microsoft/TypeScript/issues/49160) the type declarations for knex.

### Reproduction

Using `"module": "NodeNext"` and `"moduleResolution": "NodeNext"` in `tsconfig.json`, the following code gives an error:

```ts
import knex from "knex";

const x = knex(validConnectionSettings);
```

2. Error message

```sh
error TS2349: This expression is not callable.
  Type 'typeof import("PATH_TO_PROJECT/node_modules/knex/types/index")' has no call signatures.
```

armanatz:
@ethanresnick I found a workaround for this issue that might help until there is a fix.

Where ever you import knex from, you can change that statement to: `import * as knexPkg from 'knex';`

And then reference it in your code as such: `const { knex } = knexPkg.default;`

Hope that helps!

miku86:
Any updates to do this without the workaround?

Or should this workaround get added to the docs?

patrickshipe:
I opened a pull request to fix this: https://github.com/knex/knex/pull/5659

fara-nak:
any update?

mercmobily:
Hello @ethanresnick,

This issue has been resolved in Knex version 3.0.0 and later.

The problem you encountered was due to how Knex's TypeScript type declarations were structured. Prior to version 3.0.0, the type declarations used `export default` syntax as if Knex were an ES module, even though it's actually a CommonJS module. This worked fine in most cases, but when TypeScript 4.7 introduced the `moduleResolution: "NodeNext"` mode (which more strictly follows Node.js's ESM/CJS interop rules), TypeScript correctly detected Knex as a CJS module and expected the default export to be on `module.exports.default` rather than `module.exports` itself, leading to the error you saw.

[...]

PR #5659, opened 2023-08-12
Fix knex d.ts to work with Node16 module mode

patrickshipe:
Currently, the knex type declarations use `export default` as if it were an ES module. This has not mattered up until recently with Node's increased ESM support and TypeScript's `node16` module setting, which is the TypeScript's team [recommended setting](https://github.com/microsoft/TypeScript/issues/52529#issuecomment-1672036026) for Node users.

Bear with me - this is a bit convoluted. Node's [behavior when importing CJS from an ES module](https://nodejs.org/api/esm.html#import-statements) is to make `module.exports` the default export. TypeScript adheres to this, which leads to some weird quirks when we look at knex's type declarations. TS properly detects knex as a CJS module, and when it sees a default export in the declarations, it assumes that the `knex` function is set to `module.exports.default` (which it _also_ is) but _not_ to `module.exports`.  So to TS, under `node16` module mode, the default export from knex is simply an object with a `default` property. It cannot be called as a function, even though this is perfectly legal at runtime.

What's the fix here? It's quite simple. Use `export =` syntax to describe the CJS export, and declare the circular references [set in `knex.js`](https://github.com/knex/knex/blob/5ececd35c9520dd2a006ee27721178dc465a8a8c/knex.js#L20-L21). This works perfectly in both the new world and old, since we are explicitly telling TS the shape of the proper CJS export rather than having it make assumptions.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -13,8 +13,8 @@ import ResultTypes = require('./result');
 
 import { Tables } from './tables';
 
-import { ConnectionOptions } from 'tls';
 import { Stream } from 'stream';
+import { ConnectionOptions } from 'tls';
 
 // # Generic type-level utilities
 
@@ -374,7 +374,7 @@ interface DMLOptions {
   includeTriggerModifications?: boolean;
 }
 
-export interface Knex<TRecord extends {} = any, TResult = any[]>
+interface Knex<TRecord extends {} = any, TResult = any[]>
   extends Knex.QueryInterface<TRecord, TResult>,
     events.EventEmitter {
   <TTable extends Knex.TableNames>(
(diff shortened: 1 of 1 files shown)


patrickshipe:
@OlivierCavadenti - would really appreciate a look at this one. The TypeScript team is really pushing Node users to move to module `node16` and the current typings make this very difficult. Let me know what else I can do to help!

patrickshipe:
https://arethetypeswrong.github.io/?p=knex%402.5.1

image

kibertoad (changes requested):
Thanks! Is it possible to add additional tests which would have been failing under old setup?

patrickshipe:
@kibertoad I have not found a way to configure the type definition testing tool to test using module node16/nodenext in addition to the current settings. The tool seems to only allow for one tsconfig configuration in the package.json. If someone has a way to do this I am happy to give it a shot.

valtlai:
Is there plans to release a new patch version soon that includes this fix?

patrickshipe:
@kibertoad will this be published this week? Thank you :)

Jiralite:
Does this not resolve #5358?

patrickshipe:
It does @Jiralite ! We need a new version published with this code though.

Jiralite:
Awesome. Yeah, I was more pointing out that issue should ideally be closed then.

kibertoad:
Sorry for the delay. 3.0.0 is now out! Please let me know if it works as expected.

valtlai:
@kibertoad Thank you, it’s now working as expected.

PR #5670, opened 2023-08-21
feat(transaction): add method to execute functions after transaction

roggervalf:
When using transactions, people can have some functions that must be executed if transaction is committed, but these function could also depend on results of db operations. It could be useful to save these functions on the transaction itself and only at the end of its execution to be able to execute all the expected async logic to be executed too.

### DIFF
--- a/test/integration/execution/transaction.js
+++ b/test/integration/execution/transaction.js
@@ -395,6 +395,27 @@ module.exports = function (knex) {
         });
     });
 
+    it('should execute functions after transaction', function () {
+      let change1 = false;
+      let change2 = false;
+      return knex
+        .transaction(function (trx) {
+          trx.addFunctionOnAfterTransaction(async () => {
+            change1 = true;
+          });
+          trx.addFunctionOnAfterTransaction(async () => {
+            change2 = true;
+          });
+          trx.debugging = true;
+          return Promise.resolve(null);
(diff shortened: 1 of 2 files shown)


roggervalf:
hi @OlivierCavadenti, please when you get some chance :eyes:

roggervalf:
hi @kibertoad, please when you get some time :eyes:

ityifan:
Very good, I happen to need this feature, but I don't know how long I'll have to wai

ityifan (approved):

roggervalf:
Pls @rluvaton if you get some time

test/integration/execution/transaction.js
```diff
@@ -395,6 +395,30 @@ module.exports = function (knex) {
...
+            change1 = true;
+          });
+          trx.addFunctionOnAfterTransaction(async () => {
+            change2 = true;
+          });
+          trx.debugging = true;
+          return Promise.resolve(null);
+        })
+        .then(async function (result) {
+          await Promise.all(afterTransactionCollector.map((fn) => fn()));
```
roggervalf, line 415: this way, users can decide what to do with these methods after transaction finishes

roggervalf:
Please @kibertoad if you get some time 👀

PR #5681, opened 2023-09-07
feat: set error.cause for tarn acquire connection error

s-sathish:
Log actual error from tarn during acquire connection before conversion. The underlying error can be anything.

It is better to surface out the actual error returned by tarn before converting it.

Also, not sure if instead of converting, why not send back the actual error itself ?


s-sathish:
I have added the actual error from the tarn in the thrown exception incase of TimeoutError

s-sathish:
@kibertoad Ping

s-sathish:
@kibertoad Ping

owlas:
This would help to debug (if not solve) a lot of open issues, it's clearly a stumbling block for most developers running knex in prod:

https://github.com/knex/knex/issues/6034
https://github.com/knex/knex/issues/6051
https://github.com/knex/knex/issues/5696
https://github.com/knex/knex/issues/4932
https://github.com/knex/knex/issues/5658

s-sathish:
@tgriesser @kibertoad Ping

gothraven:
@owlas it's been years now .. why is this not moving forward ? I've spent three days already trying to understand whats going on when i'm following the docs to use transactions, but it keeps giving me `KnexTimeoutError` for no obvious reason.

tgriesser:
Let's change this to set the actual error that's thrown to the [`Error.cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) property so we can release it as a minor change, and then we'll do a better job on the error message here in 4.0

test/integration/execution/transaction.js
```diff
@@ -527,7 +527,7 @@ module.exports = function (knex) {
               'select * FROM accounts WHERE username = ?'
             );
             expect(error.message).to.equal(
-              'Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?'
+              'Knex: Timeout acquiring a connection ("operation timed out for an unknown reason"). The pool is probably full. Are you missing a .transacting(trx) call?'
```
tgriesser:
```suggestion
              'Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?'
```

tgriesser (approved):

PR #5683, opened 2023-09-09
fix andWhereNotJsonObject calling wrong function

FKasa:
I have no time to write tests but the issue and the fix seem pretty obvious.
This is also my first time contributing so please review my change.
I found this bug myself while using the method at work.


kibertoad (approved):

kibertoad:
thanks!

kibertoad:
Released in 3.1.0

PR #5684, opened 2023-09-09
Update package.json to get over Node 14

### LINKED ISSUE #5631: Dropping support for old Node versions in favor of Node 20

Hello. I tried to add Node 20 to to CI / GH actions version matrix. That quickly leads to some incompatibility issues, as Knex currently supports all Node versions 12 and above.

This makes it a bit tricky to support current Node versions. For example, adding Node 20 means we'd need to update `better-sqlite3` to 8.X.X version. Updating `better-sqlite3` remedies Node 20 situation, but also drops the support for Node 12 & 14. There may be other similar issues, but this was the first one I faced.

As currently only Node versions 16, 18 and 20 are maintained, I'd suggest that Knex also drops support for versions 12 and 14. It would allow us to bump some dependencies to their latest versions and include Node 20 support. There already is some glue-code to make CI run with Node 12 and more will come as other libraries drop support for unmaintained Node versions. 

However, as this is technically breaking change and could require major version bump, I thought I'd open a discussion about it. How are we feeling about dropping the old versions in favor of the new?

stevefan1999-personal:
Fixes #5631

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "knex",
   "types": "types/index.d.ts",
   "engines": {
-    "node": ">=12"
+    "node": ">14"
   },
   "scripts": {
     "build": "npm run build:gitignore && npm run build:ts && npm run format",


kibertoad (approved):

kibertoad:
@stevefan1999-personal Could you please drop the part of CI that is no longer relevant? Also if the idea is to drop support for Node 14, could you please set the minimum version to Node 16?

stevefan1999-personal:
@[**kibertoad**](https://github.com/kibertoad) Actually...Node 16 is also getting EOL'd on 11th September 2023. alas, which is ironically today at the date on comment (no offense). 

If we started strictly pushing up the versions, we are quickly limiting the potential users. 

There are still many users on Node 16 and below, and I think we need to be lenient about the engine version, so should we still keep the version on 16 first, or just straight to 18?

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "knex",
   "types": "types/index.d.ts",
   "engines": {
-    "node": ">=12"
+    "node": ">14"
```
kibertoad, line 8:
```suggestion
    "node": ">=16"
```

PR #5691, opened 2023-09-18
Fix where in query with raw column

### LINKED ISSUE #5690: Bug: `.whereIn` query build is broken with raw column binding
# Environment

Knex version: 2.5.1
Database + version: PostgreSQL 15.4
OS: Linux

# Bug

The `.whereIn` query builder doesn't handle raw _column_ values with value bindings correctly.

For example accessing a `key` in a json `column` will break the query builder.

```js
const assert = require("assert");
const knex = require("knex")({ client: "pg" });
const query = knex
    .select("*")
    .from("table")
    .whereIn(knex.raw("?? ->> ?", ["column", "key"]), ["a", "b", "c"])
    .toString();

// => `select * from "table" where "column" ->> 'a' in ('b', 'c', 'key')`

assert.equal(
    query,
    `select * from "table" where "column" ->> 'key' in ('a', 'b', 'c')`
);
```

_This used to work correctly but broke at some point, but not quite sure when._

**Edit:** I bisected the repository from the version I knew it worked in to the current version.
It seems the bug was introduced in this commit [b97d49b](https://github.com/knex/knex/commit/b97d49b1847791b3e1c437710af514d9fbc545fd).

dertieran:
Fixes #5690 

In [b97d49b](https://github.com/knex/knex/commit/b97d49b1847791b3e1c437710af514d9fbc545fd) the order in which the `columns` and `values` where build changed.
But the order is important if the raw `columns` statement contained value bindings.
Because the value bindings from column statements are now added after the values.

This PR fixes this by building the `columns` before the `values` again.


dertieran:
@OlivierCavadenti because you committed the commit where this broke would you mind taking a look at this?
I just want to make sure that this change wasn't intentional or that my change will break anything else :)

kibertoad:
@dertieran Thank you for the PR. Could you please also add an integration test for this, that actually executes the query?

dertieran:
@kibertoad Just wanted to follow up on this PR and check if the tests I added are the way you expected them.

myndzi:
(picking this up)

mercmobily:
Closing this one in favour of #6323 which is rebased and with fixed tests.  But, thank you for the submission which was 99% there!

PR #5703, opened 2023-09-29
Add create and drop schema for mssql

ecsbjunior:
Add createSchema/createIfNotExistsSchema and dropSchema/dropSchemaIfExists to mssql
Fix mssql test, when running 2x or more integration tests "npm run test:mssql" one error occurred

### DIFF
--- a/lib/dialects/mssql/schema/mssql-compiler.js
+++ b/lib/dialects/mssql/schema/mssql-compiler.js
@@ -80,6 +80,26 @@ class SchemaCompiler_MSSQL extends SchemaCompiler {
       `and object_id = object_id(${formattedTable})`;
     this.pushQuery({ sql, output: (resp) => resp.length > 0 });
   }
+
+  createSchema(schemaName) {
+    this.pushQuery(`CREATE SCHEMA ${this.formatter.wrap(schemaName)}`);
+  }
+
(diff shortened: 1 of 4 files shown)


lib/dialects/mssql/schema/mssql-compiler.js
```diff
@@ -80,6 +80,26 @@ class SchemaCompiler_MSSQL extends SchemaCompiler {
...
     this.pushQuery({ sql, output: (resp) => resp.length > 0 });
   }
+
+  createSchema(schemaName) {
+    this.pushQuery(`CREATE SCHEMA ${this.formatter.wrap(schemaName)}`);
+  }
+
+  createSchemaIfNotExists(schemaName) {
+    this.pushQuery(
+      `IF schema_id('${schemaName}') IS NULL EXEC('CREATE SCHEMA ${this.formatter.wrap(
```
rluvaton, line 90: @kibertoad shouldn't the first schema name be wrapped as well?

rluvaton (commented):
Great job!

Docs should be updated as well to note that now MSSQL is also supported

ecsbjunior:
@rluvaton i'm added in documentation my new feature for MSSQL and updated what was changed

Issue #3112, opened 2019-03-18
Using .raw with an escaped question mark ('\\?') produce a syntax error


jajm:
# Environment

Knex version: 0.16.3
Database + version: mysql  Ver 14.14 Distrib 5.7.25, for Linux (x86_64) using  EditLine wrapper
OS: Debian buster

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do

Using .raw with an escaped question mark (`'\\?'`) and other methods (limit, where, ...), database server returns a syntax error

2. Error message

```
Unhandled rejection Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '?' at line 1
    at Query.Sequence._packetToError (/home/urungi/urungi/node_modules/mysql/lib/protocol/sequences/Sequence.js:47:14)
    at Query.ErrorPacket (/home/urungi/urungi/node_modules/mysql/lib/protocol/sequences/Query.js:77:18)
    at Protocol._parsePacket (/home/urungi/urungi/node_modules/mysql/lib/protocol/Protocol.js:278:23)
    at Parser.write (/home/urungi/urungi/node_modules/mysql/lib/protocol/Parser.js:76:12)
    at Protocol.write (/home/urungi/urungi/node_modules/mysql/lib/protocol/Protocol.js:38:16)
    at Socket.<anonymous> (/home/urungi/urungi/node_modules/mysql/lib/Connection.js:91:28)
    at Socket.<anonymous> (/home/urungi/urungi/node_modules/mysql/lib/Connection.js:502:10)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
    at Socket.Readable.push (_stream_readable.js:208:10)
    at TCP.onread (net.js:607:20)
    --------------------
    at Protocol._enqueue (/home/urungi/urungi/node_modules/mysql/lib/protocol/Protocol.js:144:48)
    at Connection.query (/home/urungi/urungi/node_modules/mysql/lib/Connection.js:200:25)
    at /home/urungi/urungi/node_modules/knex/lib/dialects/mysql/index.js:144:18
    at Promise._execute (/home/urungi/urungi/node_modules/knex/node_modules/bluebird/js/release/debuggability.js:313:9)
[...]

Issue #5189, opened 2022-05-19
Postgres JSONB operator ? interpreted as binding


llouka:
# Environment

Knex version: v2.0.0
Database + version:  Postgres v12
OS: Linux

# Bug

1. Explain what kind of behaviour you are getting and how you think it should do
JSONB operator `?` in where is interpreted as binding
`.where('jsonbColumn', '?', 'jsonbKey')` replaces the `?` with a binding value, if replaced with `\\?` the error is that the operator `\\?` does not exist

Portur:
Trying to find a workaround, the following doesnt work

```javascript
   pg.raw("select * from table where a::jsonb ? 'david'", [pg.raw("?")])
```

```javascript
   pg.raw("select * from table where a::jsonb ?? 'david'"])
```

```javascript
   pg.raw("select * from table where a::jsonb //? 'david'"])
```

thomasgwatson:
I'm also having this issue. What workarounds are folks using?

rubengmurray:
@mercmobily please assign this one to me 👍

mercmobily:
Done. But I think I would prefer to see #6236 assigned to you. That will be closed once all of these regep ones are closed, so it will encompass all of them!

rubengmurray:
They're not all the same root cause. I'm happy to have the meta issue assigned as a general theme but they are only similar issues, not 'the same' 👍

mercmobily:
I know I know. That's why I didn't follow the same policy as the typescript metaissue, where I closed the MANY MIRIAD of little requests all to do with TS... If you don't like being assigned such a glorious and fun issue, though, I will definitely understand :D

rubengmurray:
I'm happy to oversee it for now. I've already dug in to a few of the specific issues, and there is definitely a DRY RegEx review required as general task.

PR #5733, opened 2023-11-07
Fix raw sql replacer to correctly escape question marks as documented

ladal1:
Raw formatter was using incorrect regex to correctly parse binding escapes, as it could never select `\\?` so its escape if on lines 14,15 and 16 would never trigger

Could not test fully because of working on arm platform currently, so please check

### DIFF
--- a/lib/formatter/rawFormatter.js
+++ b/lib/formatter/rawFormatter.js
@@ -10,7 +10,7 @@ function replaceRawArrBindings(raw, client) {
   const values = raw.bindings;
   let index = 0;
 
-  const sql = raw.sql.replace(/\\?\?\??/g, function (match) {
+  const sql = raw.sql.replace(/\\?\\?\?\??/g, function (match) {
     if (match === '\\?') {
       return match;
     }


ladal1:
Fixes: #3112, #5189 and probably more I didn't see
ping @rijkvanzanten as you offered to test the fixes another contributor was promising

ladal1:
I messed up, will find a fix later, this issue is quite impactful

klyondx:
@ladal1 thanks for your investigation so far. I see that this this is now closed without a follow-up merge or discussion about what was messed up. Is the fix still in the works? Or if you aren't able to provide a fix at this time, could you by chance just jot down what you found that you 'messed up'? Thanks

ladal1:
Well after testing it seems the issue lays somewhere else, as when I wrote additional tests and went over the existing ones, they work exactly as they should. So either the issue was already fixed, or happens in such a way I couldn't replicate it in testing even with the reported issues.

Somehow it also resolved in my project (be it through mikro-orm v6 or me simply forgetting to test the correct solution before jumping here).

klyondx:
Appreciated for the quick response. I've recently been handed a legacy project and there seemingly are the kinds of issues you describe. However with your hints, I'll first have a go at adding some more test coverage and firm up the lib versions before jumping to any conclusions. Thanks again

PR #5738, opened 2023-11-14
Missing comments on delete, update and insert

sebastienroul:
Few monthes ago comments where included in query, but only in select statement. Here is the code to make delete, update and insert compatible with comments


kibertoad:
@sebastienroul could you please add tests?

sebastienroul:
@kibertoad  Hi, yes sure, but can you give me pointer to the test written for the select statement please, I cant find them (µI'd like to use them as template). For your info PR is working like a charm on 10 of our projects :)

sebastienroul:
@kibertoad : Finally, unit test + integration test are there.
I clear my first attempt to modify all dialect, and focus only on MYSQL. Everything is green... let pull :)

kibertoad:
thanks!

PR #5761, opened 2023-12-02
Supporing returning on delete with sqlite3

mohammedbalila:
This PR is an attempt to fix delete('*') doesn't return deleted rows mentioned in here https://github.com/knex/knex/issues/5757
- It's almost done, but the following test case is failing and I'm investigating it.
`should allow returning for deletes in postgresql and mssql` the error message is clear
```bash
      -    "id": 2
      +    "id": "2"
```
I'm trying to figure out why the id is returned as a number and not a string like other dialects


mustafabalila:
The tests are passing but I'm not 100% sure about this code block, all other databases in the test block return the id as a string `id: '2'` and the test was failing because sqlite returns it as a number, I tried inspecting the code to find why this is happening but didn't have much luck. So I updated the test the expect a numeric id instead, if someone could explain to me why sqlite was different in this case I'd be grateful!
```js
tester(
  'sqlite3',
  'delete from `accounts` where `id` = ? returning *',
  [2],
  [
    {
      id: 2,
      first_name: 'Test',
      last_name: 'User',
      email: 'test2@example.com',
      logins: 1,
      balance: 0,
      about: 'Lorem ipsum Dolore labore incididunt enim.',
      created_at: TEST_TIMESTAMP,
      updated_at: TEST_TIMESTAMP,
      phone: null,
    },
  ]
);
```

code-ape:
Anything I can do to help here? Would be great to get this in sometime in the near future 😄

kibertoad:
@code-ape Can you investigate concern raised in https://github.com/knex/knex/pull/5761#issuecomment-1837587680 and report back whether or not this is something that needs addressing?

mohammedbalila:
Hey @code-ape
Any updates on this? Let me know if I need to change anything. Thanks

PR #5767, opened 2023-12-09
Update index.d.ts

hzgotb:
for below example will tell me `'this' implicitly has type 'any' because it does not have a type annotation.ts(2683)`
```ts
const b = knex.select().from(function () {
        return this.select().as('a');
    }).toQuery();
```


flleeppyy:
Bump.

PR #5771, opened 2023-12-16
Handle Vendor Prefixed Postgres Version Strings

vijaynyaya:
The existing regex pattern expects the version string to start with 'PostgresSQL'. Unfortunately, some vendors like Oracle are getting creative with version strings: 'OCI Optimized PostgreSQL 14.9, 64-bit'

### DIFF
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -132,7 +132,7 @@ class Client_PG extends Client {
   }
 
   _parseVersion(versionString) {
-    return /^PostgreSQL (.*?)( |$)/.exec(versionString)[1];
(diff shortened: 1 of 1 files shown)


lib/dialects/postgres/index.js
```diff
@@ -132,7 +132,7 @@ class Client_PG extends Client {
   }
 
   _parseVersion(versionString) {
-    return /^PostgreSQL (.*?)( |$)/.exec(versionString)[1];
+    return /PostgreSQL (.*?)( |$)/.exec(versionString)[1];
```
kibertoad, line 135: could you please add unit trst for this change?

PR #5774, opened 2023-12-18
fix: recover from broken connection

beltschatsar:


lib/dialects/oracle/utils.js
```diff
@@ -93,6 +93,7 @@ function isConnectionError(err) {
     'ORA-56600', // an illegal OCI function call was issued
     'NJS-024',
     'NJS-003',
+    'NJS-500',
```
kelvinfloresta, line 96: Maybe we can include NJS-521 as well, because I'm receiving this error together with NJS-500:

```
select * from table(***.***(:1, :2)) - NJS-500: connection to the Oracle Database was broken

NJS-521: connection to host *** port ***received end-of-file on communication channel. (CONNECTION_ID=***)
```

beltschatsar: Thanks for reviewing.

Only the start of `err.message` is looked up here (see line 98 below). Hence, NJS-521, as it is not in the beginning of the string, would never be matched here.

That said, I didn't find other missing `NJS-` error code that should be added here, this [search](https://github.com/search?q=repo%3Aoracle/node-oracledb%20errors.getErr&type=code) lead me to think there are none missing.

kelvinfloresta: Yes, never mind. Check [this](https://github.com/oracle/node-oracledb/blob/faa59030b5a5aee62f203f67fee9b5e0e930026d/lib/thin/sqlnet/ntTcp.js#L327), the error will always be wrapped with NJS-500, so you can check if NJS-500 is present

kelvinfloresta (approved):

beltschatsar:
@OlivierCavadenti or @kibertoad is there anything needed to merge this ?
Not having this fix can lead to a backend not responding at all, because some events (ex: connection kill) would lead to broken connection clogging the pool.

beltschatsar:
@kibertoad Kindly ask again to merge this. As said before, not having this fix can lead to a backend not responding at all, because some events (ex: alter system kill session) would prevent to dispose an unusable connection when [validating](https://github.com/VilledeMontreal/knex/blob/4ca3dd5bc28e0665c5bed55026fac2ec45489d81/lib/client.js#L277) this connection .

lib/dialects/oracle/utils.js
vpeltola: Reviewed: adding the NJS-500 is a big improvement to the current situation, and so far no one has reported other valid codes to add.  Thank you!

vpeltola (approved):

robmcguinness:
It would be great to add `NJS-521` and `NJS-500` as part of PR. We hit this same issue when connecting to OCI. As a workaround, leveraged `pnpm patch <pkg name>@<version>` to apply patch and seems to work very well.

Happy to offer my services to help push this along. 

Thank you maintainers!

enov (approved):
We're affected by this and the change looks good :+1: 
A commented out description would be nice, like the following, but not a blocker:
```
    'NJS-500', // connection to the Oracle Database was broken; usually followed by NJS-521 or DPI-1080
```
Thank you for taking the time to make this pull request.

kibertoad:
@rluvaton Could you please review?

kibertoad (approved):

kibertoad:
This looks good, thank you!

enov:
Can we please have a patch tag/release and push a new npm package for this change to be released? Thank you!

PR #5780, opened 2023-12-27
sqlite: insert use normal values syntax, fix returning

ricochet1k:
Knex was losing the `returning` clause when doing `insert default values` even though [SQLite supports](https://www.sqlite.org/lang_insert.html) it, and arguably the returning clause is most important in the default values case.

Also while I was there, I noticed SQLite was generating a weird `select ... union all select` syntax for multi-row inserts which is not actually necessary as [SQLite supports](https://www.sqlite.org/lang_insert.html) the same `values (), ()` syntax all the others do.

I find it frustrating that if the `returning` clause isn't supported (such as by redshift) that Knex just silently ignores it and returns unexpected values.

Tests pass locally.

### DIFF
--- a/lib/.gitignore
+++ b/lib/.gitignore
@@ -5,3 +5,6 @@
 
 # Do not include tsc source maps
 **/*.js.map
+
+# Do not include .js files from .ts files
+dialects/index.js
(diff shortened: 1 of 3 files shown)


agnauck:
looking for this feature as well with sqlite. Any chance this will be merged soon?

laurenskling:
I just found that this fixes my issues in the Strapi CMS, where I cannot save something with over 500 relations because it [crashes on  too many terms in compound SELECT](https://github.com/strapi/strapi/issues/20402), while using the `select ... union all select` syntax.

This PR fixes that issue. <3

Issue #5769, opened 2023-12-13
Sqlite multi-insert fails to convert Object to JSON string but does for single insert


code-ape:
# Environment

**Knex version:** 3.0.1
**Database + version:** `sqlite3` version 5.1.6
**OS:** MacOS 14.1
**Node:** 20.3.1

# Bug

1. **Explain what kind of behavior you are getting and how you think it should work.**

When using the SQLite driver with a JSON column, the behavior between inserting a single record and multiple records is different.

```js
await knex.from('my_table').insert([
   { my_column: { a: 1 }},
])
// Results in conversion being done via JSON.stringify:
//   sql: 'insert into `my_table` (`my_column`) values (?)',
//   bindings: [ '{"a":1}' ]

await knex.from('my_table').insert([
   { my_column: { b: 2 }},
   { my_column: { c: 3 }},
])
// Results in conversion NOT being done:
//   sql: 'insert into `my_table` (`my_column`) select ? as `my_column` union all select ? as `my_column`',
//   bindings: [ { b: 2 }, { c: 3 } ]
// This causes the string `[object Object]` to be inserted as `.toString()` is called by the SQLite3 driver.
```

3. **Error message**

None, this occurs silently.

4. **Reduced test code, for example in https://npm.runkit.com/knex or if it needs real database connection to MySQL or PostgreSQL, then single file example which initializes needed data and demonstrates the problem.**

```js
import Knex from 'knex'

async function main() {
  console.log('Creating in-memory database ...')

  const knex = Knex({
    client: 'sqlite3',
    connection: ':memory:',
    useNullAsDefault: true,
  })

  console.log('Creating table `my_table` ...')
  await knex.schema.createTable('my_table', (table) => {
    table.json('my_column').notNullable()
  })

  console.log('Doing single insert ...')
  const singleInsert = knex.from('my_table').insert([
    { my_column: { a: 1 }},
  ])
  console.log('Single insert SQL:', singleInsert.toSQL().toNative())
  await singleInsert

  console.log('Doing multi insert ...')
  const multiInsert = knex.from('my_table').insert([
    { my_column: { b: 2 }},
    { my_column: { c: 3 }},
  ])
  console.log('Multi insert SQL:', multiInsert.toSQL().toNative())
  await multiInsert

  console.log('Getting all rows from table...')
  const rows = await knex.from('my_table').select('*')
  console.log(rows)

  await knex.destroy()
}

main()

// Output:
//   Creating in-memory database ...
[...]

PR #5782, opened 2023-12-29
Issue #5769: Sqlite support multi-value JSON insert

code-ape:
**Issues:**
1. https://github.com/knex/knex/issues/5769

**Changes:**
1. TODO, will populate this once draft status removed


lib/.gitignore
```diff
@@ -5,3 +5,7 @@
 
 # Do not include tsc source maps
 **/*.js.map
+
+# Do not include .js files from .ts files
+util/version.js
+dialects/index.js
```
kibertoad, line 11: this causes files not to be published either, I'd rsther have them committed

code-ape: Ohhhh! Good to know. There's a script generating these `.gitignore` items . I can update it stop doing that and check in the generated code.

kibertoad: it's ok to generate them prepublish too, to protect against someone forgetting to run it

lib/dialects/sqlite3/query/sqlite-querycompiler.js
```diff
@@ -82,62 +84,34 @@ class QueryCompiler_SQLite3 extends QueryCompiler {
...
-      if (returning) {
-        sql += this._returning(returning);
-      }
-
-      return {
-        sql,
-        returning,
-      };
+    if (
+      !this.client._satisfiesVersion([3, 7, 11]) &&
```
kibertoad, line 88: I'd rather do this check once on driver instantiation and set a feature boolean flag for subsewuent chrcks

lib/dialects/better-sqlite3/index.js
```diff
@@ -7,6 +7,19 @@ class Client_BetterSQLite3 extends Client_SQLite3 {
...
+   * Get `Version` from client or return `[0, 0, 0]` on failure.
+   *
+   * @returns {[number, number, number]}
+   */
+  _clientVersion() {
+    // hard coding 3.25.1 as better-sqlite3 offers no way to get Sqlite3 version it was built with
+    // and better-sqlite3 1.0.0 shipped with 3.25.1 in 2016.
+    //
+    // Issue opened for this: https://github.com/WiseLibs/better-sqlite3/issues/1122
+    return [3, 25, 1];
```
kibertoad, line 20: can we require minimum better-sqlite with support for multi-insert via peerDependencies? I'd rather force people to upgrade rather than lock out them out of the beneficial feature

test/integration2/query/insert/inserts.spec.js
```diff
@@ -2315,6 +2315,72 @@ describe('Inserts', function () {
               assertJsonEquals(result[0].content, arrayOfObject);
             });
         });
+
+        describe(`#5769 sqlite multi-insert uses standard syntax and supports JSON (does NOT include better-sqlite3)`, function () {
+          it(`should JSON.stringify Sqlite multi-insert of Objects`, async function () {
+            if (!isSQLite3(knex)) {
```
kibertoad, line 2321: do neweer better-sqlite support this feature?

lib/dialects/sqlite3/query/sqlite-querycompiler.js
code-ape: @kibertoad what do you want to do if we initialize with the driver with a version below the 3.7.11 threshold? Throw an error and halt, forcing at least that version? Something else?

lib/dialects/better-sqlite3/index.js
code-ape: So, two thoughts:

1. If `better-sqlite3` users are using the bundled `sqlite3` build then they're fine. The 1.0.0 version of `better-sqlite3` satisfies the 10+ year old requirement of being at least 3.11.7.
2. Technically the users of `better-sqlite3` (and `sqlite3` for that matter) can use a different C compilation of the sqlite3 library. So the NPM package version doesn't guarantee the version of `sqlite3.c` they've compiled.

What would you like to do?

test/integration2/query/insert/inserts.spec.js
code-ape: I'll update the test and check. I think so, when I was originally authoring this I was just trying to focus on making sure that `sqlite3` worked.

lib/dialects/sqlite3/query/sqlite-querycompiler.js
kibertoad: Wouldn't it break anyway if someone attempts a multi-value JSON insert anyway? Do we need to force-break anything?

lib/dialects/better-sqlite3/index.js
kibertoad: After https://github.com/WiseLibs/better-sqlite3/pull/1124 is merged we can do a more explicit check, but for older versions I would let user proceed by default and let them crash if their version is too old.

PR #5791, opened 2024-01-19
Fix FOR UPDATE OF with explicit schema

### LINKED ISSUE #5053: BUG - FOR UPDATE must specify unqualified relation names
# Environment

Knex version: 1.0.3
Database + version: PostgreSQL 13
OS: Node 14

# Bug

The error comes up when you build a query like knex('users').withSchema('myschema').where('id','xyz').forUpdate('users')
It seems incompatible to use withSchema and forUpdate in the same query because withSchema add the schema also in front of the forUpdate table name, but should not.

Error message: FOR UPDATE must specify unqualified relation names

ErinCall:
When calling `.withSchema(...)....forUpdate('someTable')`, knex would produce incorrect sql syntax—it included the schema name in the table name, prompting a "FOR UPDATE must specify unqualified relation names" error from postgresql. 

Fixes #5053.

### DIFF
--- a/lib/dialects/postgres/query/pg-querycompiler.js
+++ b/lib/dialects/postgres/query/pg-querycompiler.js
@@ -205,9 +205,16 @@ class QueryCompiler_PG extends QueryCompiler {
   _lockingClause(lockMode) {
     const tables = this.single.lockTables || [];
 
-    return lockMode + (tables.length
-      ? ' of ' + tables.filter(Boolean).map(table => this.formatter.wrap(table)).join(', ')
-      : '');
+    return (
+      lockMode +
+      (tables.length
+        ? ' of ' +
+          tables
+            .filter(Boolean)
+            .map((table) => this.formatter.wrap(table))
+            .join(', ')
+        : '')
+    );
   }
 
   _groupOrder(item, type) {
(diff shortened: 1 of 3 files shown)


lib/dialects/postgres/query/pg-querycompiler.js
```diff
@@ -197,29 +197,12 @@ class QueryCompiler_PG extends QueryCompiler {
     }
   }
 
-  // Join array of table names and apply default schema.
-  _tableNames(tables) {
```
ErinCall, line 201: When the `_tableNames` helper was written (#2835) there were multiple methods hitting it, but they've since been merged into the single `_lockingClause` method. Given that the only really interesting thing it was doing was prepending the schema name--which is the incorrect behavior this commit fixes--it didn't seem worth keeping at this point.

test/integration2/query/select/selects.spec.js
```diff
@@ -993,7 +993,7 @@ describe('Selects', function () {
         try {
           await knex.transaction((trx) => {
             // select all from two test tables and lock only one table
-            return trx('test_default_table')
+            return trx.withSchema('public').from('test_default_table')
```
ErinCall, line 996: I'm not sure you'd say `.withSchema('public')` in a real query but it lets me exercise the behavior in question without having to learn enough about knex's test environment to stand up a discrete schema 😁

OlivierCavadenti (approved):

PR #5804, opened 2024-02-21
feature/add-default-schema-from-config

webselect:


webselect:
Add the avaibilty to add an default schema for every database call from the connexion configuration

PR #6025, opened 2024-03-06
Fixes _setNullableState not respecting schema

### LINKED ISSUE #6024: Knex' default implementation of _setNullableState fails for tables in non-standard schemas
# Environment

Knex version: 3.1.0
Database + version: Mssql 2019-latest
OS: N/A

If issue is about oracledb support, tag @ atiertant. For MSSql tag @ smorey2.
If issue is about TypeScript definitions, tag @ lorefnon.

# Bug

1. What I get: 
I have an MsSql-table in a non-standard schema calle 'testSchema'. 
The table has a non-nullable column 'id_not_nullable'.
I execute  ```await knex.schema.withSchema("testSchema").table(tableName, (table) => {
      table.setNullable('id_not_nullable');
    });``` and it fails.
2. What it should do:
Above code removes the not_nullable flag in the table's column.

4. Error message:

6. Reduced test code, for example in https://npm.runkit.com/knex or if it needs real

OliverHenrichs:
closes #6024

Fix seems obvious to me.

I checked whether I can add tests for using a non-standard schema.
Unfortunately `createSchema` and `dropSchema` are only available for Postgres. (=> Could make a test for pg)
Also the tests in *set-nullable.spec.js* are not really set up to deal with schemas. 

Regarding the fix:
If `this.schemaNameRaw` is undefined, the PR's changes will not affect the query, so everything that used to work still works.
For tables in other schemas the code never worked, now it could (and when I use my fork in our project, it does...) => seems better to me.


OliverHenrichs:
ohhh, we have movement :-) 

<3 @mercmobily

mercmobily:
Lots of it. We are going through the existing PRs as it is. I worked on the web site a lot mainly because I needed myself a better feel of what Knex actually is and does. The improvements in the guide are incredible. 
We are moving on to deal with pretty much all open PRs at the moment, one way or another. This is important as we are planning a lot of improvements which will make the PRs harder to merge.

OlivierCavadenti (approved):

Issue #6021, opened 2024-03-04
suggestion: Force rownum (Oracle dialect) to be a number with the function to_number()


S200124:
Knex version: 3.1.0
Database + version: Oracle 19c EE
OS: Red Hat Enterprise Linux Server 7.9

### Suggestion/improvement to cover Oracle bug (?)

When building a query using the limit() function, in Oracle it is translate to the clause "where rownum <= ?" (which is correct). But sometimes, inconsistently and for unknown reasons, Oracle raise the error "ORA-01722: invalid number".
This can be solved forcing rownum to be a number -> "where to_number(rownum) <= ?"

Error message:
_Error: select * from (select "COL1", "COL2", "COL3", "COL4", "COL5", "COL6" from "TABLE" order by "COL3" ASC) where rownum <= :1 - ORA-01722: invalid number_

This is not a Knex bug, and for what I can see, it is not a consistent error in Oracle. I use the same limit() function with the same setup on other tables/query, and it works without problem. However, changing  "where rownum <= ?" to "where to_number(rownum) <= ?" should cover all cases, also the special ones.

mercmobily:
I would like to dig into this a little deeper. Are you still around @S200124 ?

myndzi:
I tried to get a deeper understanding of this, but was limited to asking LLMs without being able to verify the "answers" via Oracle's docs. However, enough of it made sense that it got me asking the right questions, and I definitely don't think this is the correct fix.

Meaningfully, applying `to_number` to _rownum_ makes no actual sense and potentially has a negative impact on performance. ROWNUM [is a number](https://docs.oracle.com/en/database/oracle/oracle-database/26/sqlrf/ROWNUM-Pseudocolumn.html#GUID-2E40EC12-3FCF-4A4F-B5F2-6BC669021726) and the binding parameter (:1) is certainly the source of the coercion error.

The reason the suggestion "fixes" your problem in practice is indeterminate, but is probably some complicated nuanced thing like altering the optimization plan in such a way that the predicate does not wind up resolving (essentially _deferring_ the error into a context where it doesn't wind up mattering).

I could have seen `to_number(:1)` as a fix, but that's better handled in the JS runtime (just pass it a number to begin with) -- and knex definitely ensures that only integers are passed as the limit value.

The most likely culprit here is that the bindings don't match the query -- which could be a bug in the original code or in knex, it's unclear (not enough information in the OP).

mercmobily:
[...]

PR #6026, opened 2024-03-07
Change "to_number(rownum)" to prevent ORA-01722

S200124:
This fix resolves the issue/suggestion #6021


myndzi:
Closing:
https://github.com/knex/knex/issues/6021#issuecomment-3704123215

There's not enough information to justify merging this as-is, but if you are still experiencing the issue or able to reproduce and gather more information (in particular, the query _and_ bindings associated with a query that failed), I'm happy to dig deeper

Issue #6028, opened 2024-03-07
Query doesn't work with `.stream()` if options not exists


gabolera:
# Environment

Knex version: 3.1.0
Database + version: Oracle 18c
OS: Windows (WSL Ubuntu 22.04)

# Bug

1. On make a query using stream method the knex force me passing a parameter in options! 

Example: (with first case, never join `.on('data')`)
```js
// This dosen't work
const stream = db.raw(`select * from users`).stream()

stream.on('data', (chunk) => {
  console.log(chunk)
})



// But... This work
const stream = db.raw(`select * from users`).stream({})

stream.on('data', (chunk) => {
  console.log(chunk)
})
```

If is optional parameters is should not a impeditive for use `.pipe()` `.on()`

rluvaton:
Can you create a pr?

gabolera:
Shure 🙂

gabolera:
Finished tests, and okay for oracledb driver (PR #6040) but a question arose here. 

In this case i solved only for oracledb driver! I see example used on mysql
https://github.com/knex/knex/blob/aedba5e49ccbe4e44bf1a57f51a63d8b1ec0e6e6/lib/dialects/mysql/index.js#L104-L107


But maybe it's interesting fix this on here where stream options is {} with default for all drivers?
https://github.com/knex/knex/blob/aedba5e49ccbe4e44bf1a57f51a63d8b1ec0e6e6/lib/client.js#L157-L160

```javascript
stream(connection, queryParam, stream, options = {}) {
``` 

or this is a bad idea?

Thanks :) @rluvaton

mercmobily:
Hello! Thank you for reporting this issue and for the excellent reproduction case. You're absolutely right that stream options should be optional, and this is indeed a legitimate bug.

Also thank you for fixing it with  PR #6040, which adds a default value for the options parameter in the Oracle dialect's streaming method. The fix is straightforward and follows the same pattern used by other dialects like MySQL.

Regarding your question about whether to fix this in the base Client class rather than just the Oracle dialect - that's a good observation! The current fix in the Oracle dialect is sufficient and follows established patterns in the codebase, but if you wanted to refactor it to the base class for consistency across all dialects, that would be a nice improvement.

If you don't, we will consider it once we get to the PR review.

We're currently going through a comprehensive triaging process for all issues and PRs, and we'll get to reviewing and merging #6040 once that's complete. Apologies for the time it's taking - we appreciate your patience and your contribution to fixing this bug!

PR #6040, opened 2024-03-21
fix: add default value on stream query `oracledb` (#6028)

gabolera:


gabolera:
@rluvaton you can review this PR?

rluvaton:
did you really fix the bug if no tests exist that failed before and passed after? ;)

asbelljc:
Any updates on this one? I ran into this issue today as well. Thankfully passing `{}` works alright for now.

myndzi:
The root cause here is that `oracledb` checks `arguments.length` in its handling of the `queryStream` method, so it treats "the third argument is undefined" differently than "no third argument given". This caused it to throw an error, which was never surfaced to the caller.

Out of scope for this PR is that there appears to be no way to exercise this in the promise-generating variant of `.stream()` because `Client.stream()` only takes one parameter (_either_ options _or_ a handler), though all the subsequent functions and methods seem to support having both at once. `options` _does_ get defaulted to `{}` at the Runner level, though I chose not to take this option:

https://github.com/knex/knex/issues/6028#issuecomment-2012286967

... since "options" here is semantically distinct for each driver. While it's almost certainly an options bag, it's not required to be, so rather than dig a deeper hole, I'm just keeping things as they are.

I added a test to ensure that, instead of throw and swallow the error. The other test was modified into a slightly cleaner shape but has no impact unless/until options are passable in the "handler" form of `.stream()`

PR #6047, opened 2024-03-26
Chore: clean old CI versions & exceptions

stscoundrel:
There were some leftovers from when Node 20 was added to CI and 12/14 were dropped. 

- Remove Node 12 specific dependency install. It is not used anymore.
- Run individual one-version-only Node 16 and Node 18 tasks in latest LTS (=20). They were the odd ones out.
- Update supported Node versions in readme. It suggested everything from 12 onwards were still supported.


stscoundrel:
What do you think @kibertoad ? Seems the failed CI runs are unrelated to this change.

Issue #5646, opened 2023-07-31
Onconflict / Upsert in MSSQL


code-cream:
# Environment

Knex version: 2.4.2
Database + version: MSSQL
OS: Windows

Select applicable template from below.
If issue is about oracledb support, tag @ atiertant. For MSSql tag @ smorey2.
Rest of dialects don't need tags.
If issue is about TypeScript definitions, tag @ lorefnon.

# Feature discussion / request

1. Explain what is your use case
I'm struggling to find examples of how I can implement upsert functionality for MSSQL using Knex.  Any advice for those trying to implement a pattern similar to `onConflict().Merge()` , but for MSSQL?  Are there any plans to include `onConflict` or `merge` support for MSSQL eventually?

I dug into the implementation thread for pg/mysql and it seems like it was close but some tests were not passing.  I'd be happy to attempt to contribute, especially if i'm not missing some larger reason why it isnt supported.

@smorey2 thank you so much for the assistance

3. Explain what kind of feature would support this
Ideally `onConflict` and `merge` would be available for MSSQL so that they can be used in a db agnostic way.

5. Give some API proposal, how the feature should work

Same way it does for postgres and mysql

KrisLau:
did you ever figure this out?

mercmobily:
Hello! Thank you for opening this feature request.

Good news - there is already an open pull request (#6050) that implements exactly what you're looking for! The PR adds `onConflict()` and `merge()` support for MSSQL using the standard ANSI SQL `MERGE` statement, which would make upsert functionality database-agnostic across PostgreSQL, MySQL, SQLite, and MSSQL.

The implementation in PR #6050 has been quite thoroughly tested - all database-specific tests pass, including MSSQL tests across multiple Node versions (16.x, 18.x, 20.x, 22.x). There's only a minor linting/types check failure that needs addressing before it can be merged.

However, there are some important considerations about using MERGE in MSSQL that you should be aware of. As discussed in the PR comments, MERGE statements in MSSQL have some well-documented issues with concurrency, locking, and triggers that may not be suitable for all use cases, particularly high-concurrency systems. Several articles highlight these concerns:

- https://michaeljswart.com/2021/08/what-to-avoid-if-you-want-to-use-merge/
- https://www.mssqltips.com/sqlservertip/3074/use-caution-with-sql-servers-merge-statement/
- https://sqlblog.org/merge

[...]

PR #6050, opened 2024-03-31
feat: MSSQL onConflict/merge support

dustryder:
Ref: #5646, #3186

### What
Support for onConflict/merge for MSSQL

### Why
Because there there is no support for this in Knex for MSSQL. You'd have to either decide on an upsert approach and wrap it in raw or use individual statements to achieve the similar behaviour

### How
Implementing (most) of the current onConflict/ignore/merge API via a subset of standard ansi sql merge statements. .raw inside onConflict has not been added.
 
### Anything else
Because this approach uses standard syntax, other dialects should also be able implement this in a similar manner (given they too support standard syntax).

### DIFF
--- a/lib/dialects/mssql/query/mssql-querycompiler.js
+++ b/lib/dialects/mssql/query/mssql-querycompiler.js
@@ -28,11 +28,6 @@ class QueryCompiler_MSSQL extends QueryCompiler {
   constructor(client, builder, formatter) {
     super(client, builder, formatter);
 
-    const { onConflict } = this.single;
-    if (onConflict) {
-      throw new Error('.onConflict() is not supported for mssql.');
-    }
-
     this._emptyInsertValue = 'default values';
   }
 
@@ -70,6 +65,9 @@ class QueryCompiler_MSSQL extends QueryCompiler {
   // Compiles an "insert" query, allowing for multiple
   // inserts using a single query statement.
   insert() {
+    if (this.single.onConflict) {
+      return this._insertWithMerge();
+    }
     if (
       this.single.options &&
       this.single.options.includeTriggerModifications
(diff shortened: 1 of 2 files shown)


000dry:
Is anything happening with this? Seems like a no-brainer and would be really helpful to MSSQL users

KrisLau:
@elhigu @kibertoad

KrisLau:
@dustryder Been testing it for a little bit and everything works well except for unique composite indexes.  Seems to have been the issue with the original PR as well: https://github.com/knex/knex/pull/3763#issuecomment-717650068. 

Did you manage to get that part working? Would really appreciate any guidance on getting that working since it seems to be stuck on my end there as well

KalebMatthews:
The code in the commit linked related to triggers is only hit if you add the includeTriggerModifications option in your insert or update statement. This just bypasses the output using a slower methodology but it's a simple way to get around the trigger issues for return values. You could use the same ideology for the upset I suspect though. If you need further clarifications let me know.

KrisLau:
@KalebMatthews Sorry for the late reply! I've tried a bunch of different fixes but none of them seem to work. Would you be able to provide some guidance on the exact changes to make?

KrisLau:
Also been researching and trying to figure out if maybe I can replicate this: https://github.com/sequelize/sequelize/blob/0b7c86d063bfb43fd3d513640456a63304934231/packages/mssql/src/query-interface.js#L50

from sequelize

KalebMatthews:
@KrisLau So you're issue is that you are treating the output like a return statement. As the error indicates if a table has a trigger the output statement must be pushed into a temp table of some sort. The main knex maintainers did not like the weight that this put on the normal query (which I agree with) which is why I added a modification option to enable to trigger specific insert, update, del returns. You can add the return feature only if the return is being utilized and generate the temp table and separate select statement (and delete temp table statement) if a return is specified. If one is not used then you can just return the count or exclude the output completely.

That is my suggestion.

KrisLau:
@KalebMatthews Sorry for all the trouble I'm rather new to this! I looked through the `_insertWithMerge` function but it seems to be doing the same thing that you had implemented in the `insert` methods unless I'm misunderstanding the code. Specifically this part seems to be replicating the behaviour you mentioned:
```js
const returningSql = returning
  ? ` ${this._returning('insert', returning, options.includeTriggerModifications)}`
  : ''

// other code

sql += returningSql;

if (options && options.includeTriggerModifications) {
  sql = this._buildTempTable(returning) + sql + this._buildReturningSelect(returning);
}
```
So is the issue with my code have to do with me having to pass true to the `includeTriggerModication` option?

The only other difference I could see was the return statement returning just the sql like this in this PR: `return sql` instead of an object with both sql and returning like this: 
```js
return {
    sql,
    returning
}
```

KalebMatthews:
According to your testing scripts you are expecting results like 
```merge into [upsert_tests] using (values (?, ?)) as tsource([email], [name]) on [upsert_tests].[email] = tsource.[email] when not matched then insert ([email], [name]) values (tsource.[email], tsource.[name]) output inserted.[email];```

The last output is where your error string is referring to.

**Comment by KrisLau** [comment]
@KalebMatthews Managed to get rid of the error turns out I hadn't noticed that I fixed the error and was getting a different one related to something else (managed to fix that too), thanks so much for the help! Now it just seems to be failing silently and I'm not sure what to do about that haha (no error, no update in the db). Any way to debug a silent failure in knex?

**Comment by KrisLau** [comment]
@KalebMatthews Managed to get rid of the error turns out I hadn't noticed that I fixed the error and was getting a different one related to something else (managed to fix that too), thanks so much for the help!

**Comment by KalebMatthews** [comment]
> @KalebMatthews Managed to get rid of the error turns out I hadn't noticed that I fixed the error and was getting a different one related to something else (managed to fix that too), thanks so much for the help!

Glad I could be of some help.

**Comment by KrisLau** [comment]
@kibertoad Saw that this was merged, thanks! Just wanted to add that I ended up not using the method awhile back when I was researching and thought I'd add some notes for anyone in case they might want to use the method.

It seems at least from some articles and StackOverflow posts I read that the usage of MERGE is not necessarily safe? From my understanding it causes locks and some other issues that would be bad for systems that have a lot of concurrent users as well as a number of other issues. Most articles seem to suggest against using it overall:

- https://michaeljswart.com/2021/08/what-to-avoid-if-you-want-to-use-merge/ (the most cited one when i read forums)
- https://www.mssqltips.com/sqlservertip/3074/use-caution-with-sql-servers-merge-statement/
- https://sqlblog.org/merge

I think the method would still be very useful to a subset of people but a note on the method somewhere in the documentation and maybe a comment on the method specifying that mssql's implementation has some drawbacks on the MSSQL side.

What I ended up implementing (using `objectionjs`'s inside the BaseModel) instead in case this might be helpful to anyone else:
```js
// trx is just me passing in the transaction
// created is a list of created records that's formatted as an object array
static async insertMerge(trx, created) {
    try {
      await this.query(trx)
        .insert(created) 
        .onError(async (error, _query) => {
          // 2627 is a primary key conflict error
          if (error.number === 2627) {
            const createdIDs = created.map(l => l.id);
            const conflicts = await this.query(trx).whereIn('id', createdIDs);
            const inserts = _.differenceBy(created, conflicts, 'id');
            const updates = _.differenceBy(created, inserts, 'id');

            if (inserts.length > 0) {
              await this.query(trx).insert(inserts);
            }

            if (updates.length > 0) {
              for (const x of updates) {
                const updateData = _.omit(x, ['id']); // you should also omit any columns that auto update here like timestamps
                await this.query(trx).where('id', x.id).patch(updateData);
              }
            }
          } else {
            return Promise.reject(error);
          }
        });
    } catch (e) {
      return Promise.reject(e);
    }
  }
```

kibertoad:
@KrisLau would yiu recommend reverting feature overall, or adding documentation?

KrisLau:
@kibertoad I dont think I have enough expertise to give great advice haha but imo probably just add documentation? most of my current knowledge was just those forums & articles

rafagsiqueira:
This looks close. Looking forward to it.

mordam:
will this be merged in sometime soon? would benefit from this functionality today.

PR #6054, opened 2024-04-06
Add `NULLS NOT DISTINCT`

langovoi:
Postgres 15+ allows to add `NULLS NOT DISTINCT` for unique indexes and constraints.

I added `nullNotDistinct: boolean` option for `unique` in `TableBuilder`, to make possible to use this feature.

https://www.postgresql.org/about/featurematrix/detail/392/


rluvaton:
Thank you for the PR, please throw an error for other dialects when they try to use it

langovoi:
Done! @rluvaton does it look well?

lib/dialects/sqlite3/schema/sqlite-tablecompiler.js
```diff
@@ -133,8 +133,14 @@ class TableCompiler_SQLite3 extends TableCompiler {
   unique(columns, indexName) {
     let deferrable;
     let predicate;
+    let nullsNotDistinct;
     if (isObject(indexName)) {
-      ({ indexName, deferrable, predicate } = indexName);
+      ({ indexName, deferrable, predicate, nullsNotDistinct } = indexName);
+    }
+    if (nullsNotDistinct !== undefined) {
```
rluvaton, line 140: allow null as well

```suggestion
    if (nullsNotDistinct != null) {
```

lib/dialects/oracle/schema/oracle-tablecompiler.js
```diff
@@ -156,8 +156,12 @@ class TableCompiler_Oracle extends TableCompiler {
 
   unique(columns, indexName) {
     let deferrable;
+    let nullsNotDistinct;
     if (isObject(indexName)) {
-      ({ indexName, deferrable } = indexName);
+      ({ indexName, deferrable, nullsNotDistinct } = indexName);
+    }
+    if (nullsNotDistinct !== undefined) {
```
rluvaton, line 163: allow null as well
```suggestion
    if (nullsNotDistinct != null) {
```

lib/dialects/mysql/schema/mysql-tablecompiler.js
```diff
@@ -345,14 +345,19 @@ class TableCompiler_MySQL extends TableCompiler {
...
-      ({ indexName, deferrable, storageEngineIndexType } = indexName);
+      ({ indexName, deferrable, storageEngineIndexType, nullsNotDistinct } =
+        indexName);
     }
     if (deferrable && deferrable !== 'not deferrable') {
       this.client.logger.warn(
         `mysql: unique index \`${indexName}\` will not be deferrable ${deferrable} because mysql does not support deferred constraints.`
       );
     }
+    if (nullsNotDistinct !== undefined) {
```
rluvaton, line 358: allow null as well

```suggestion
    if (nullsNotDistinct != null) {
```

lib/dialects/mssql/schema/mssql-tablecompiler.js
```diff
@@ -298,6 +300,9 @@ class TableCompiler_MSSQL extends TableCompiler {
     if (useConstraint && predicate) {
       throw new Error('mssql cannot create constraint with predicate');
     }
+    if (nullsNotDistinct !== undefined) {
```
rluvaton, line 303: allow null as well

```suggestion
    if (nullsNotDistinct != null) {
```

rluvaton (commented):
Please add tests for the other dialects that they are throwing and integration test for postgres that it actually works

atlanteh:
@rluvaton
Can this please be merged?

rluvaton:
@atlanteh there are still unresolved comments

mercmobily:
Alright a few notes about this one.

1) Rebasing
First of all, it needs rebasing. Right now it's getting "[Linting and Types / Linting and Types (22.x) (pull_request)](https://github.com/knex/knex/actions/runs/20723557538/job/59493819464?pr=6054)Failing after 39s" -- but that problem was fixed ages ago elsewhere in the codebase.

2) Redshift
Also note that Redshift still shows SQL (including nulls not distinct) and does not error. That’s because lib/dialects/redshift/schema/redshift-tablecompiler.js inherits Postgres' `unique()` and doesn’t block `nullsNotDistinct`.
I really don't think Redshift supports this. I am for adding a guard like this:

```
  unique(columns, indexName) {
    if (indexName && typeof indexName === 'object') {
      if ('nullsNotDistinct' in indexName) {
        throw new Error(
          'redshift cannot make unique with nullsNotDistinct option'
        );
      }
    }
    return super.unique(columns, indexName);
  }
```

3) Versioning guard for Postgres

```
  if (nullsNotDistinct !== undefined) {
    const version = this.client.version;
    if (version && parseFloat(version) < 15) {
      throw new Error(
        'postgres < 15 does not support nulls [not] distinct unique indexes'
      );
    }
    nullsNotDistinct = nullsNotDistinct
      ? ' nulls not distinct'
      : ' nulls distinct';
  } else {
    nullsNotDistinct = '';
  }
```

Please note that I _really_ dislike the version checking above, and my next issue will be about unifying how it works. However, we need to land this this way first -- that issue/PR will then harmonise every versioning comparison. I want every dialect to call the same function to check version, and have the "I can't parse === any feature" to apply automatically.

myndzi:
I expect this is going to crop up multiple times while we're knocking down the PR backlog. I'm in favor of just deferring it so long as it doesn't actively cause an issue. In this case, for example, if a user uses this feature and the DB doesn't support it -- the DB will just throw an error. We can throw a clearer error, but doing it on the fly for every single PR might be wasted effort when we're already expecting to come back at feature detection more holistically.

On the other hand, if you think it's required to merge these PRs, then perhaps we should build out the feature detection infrastructure now. I'm reasonably certain I could do that in a way that won't cause merge conflicts with pending PRs, e.g. by creating a utility class that we can just call from dialect clients in an isolated way

mercmobily:
>  On the other hand, if you think it's required to merge these PRs, then perhaps we should build out the feature detection infrastructure now. I

I am for merging this as-is, and create NOW a versioning infrastructure (I am doing it right now) to merge ASAP. That PR will convert the existing code (which is only in 3 spots right now); this will give me multiple use cases to check that it works and it's good.

I am working right now on a documentation workflow (an actual PR) so that we don't publish features that aren't actually landed, and on a PR that will include a versioning infrastructure and will do away with ad-hoc checks. Again, I like having some spots where this happens, as it will make me feel more certain that the versioning functions are generic enough to cover different cases.

mercmobily:
And I agree, we need an automatic way so that if I write a comment in a specific format, like "over to @myndzi", it assigns to you automatically.

mercmobily:
> I am for merging this as-is, and create NOW a versioning infrastructure (I am doing it right now) to merge ASAP. That PR will convert the existing code (which is only in 3 spots right now); this will give me multiple use cases to check that it works and it's good.

As discussed,

* throw on the redshift subclass (ugh, subclasses)
* flag for future inclusion in feature gate infrastructure (`#GATING`?)

myndzi:
OK. Should be able to squeeze this in easily enough

PR #6059, opened 2024-04-18
Knex Playground

dgadelha:
knex-playground integrated into Knex's official docs

| Desktop | Mobile |
| ------- | ------ |
|   |  |

> [!WARNING]
> I currently need help on making light mode work with the editor

Any code feedback is also welcome, I have almost zero knowledge with Vue

### DIFF
--- a/docs/.vitepress/theme/Playground.vue
+++ b/docs/.vitepress/theme/Playground.vue
@@ -1,0 +1,29 @@
+<script>
+import Split from 'split.js';
+
+export default {
+  mounted() {
+    Split(['#editor', '#output']);
+  },
+};
+</script>
+
+<template>
+  <div class="playground split">
+    <div id="editor">
+      Editor
+    </div>
+
+    <div id="output">
+      Output
+    </div>
+  </div>
+</template>
+
+<style scoped>
+.playground {
+  width: 100vw;
+  height: 100vh;
+  padding-top: var(--header-height);
+}
+</style>
--- a/docs/.vitepress/theme/Playground.vue
+++ b/docs/.vitepress/theme/Playground.vue
@@ -8,15 +8,73 @@ export default {
 };
 </script>
 
+<script setup>
+import { shallowRef } from 'vue';
+import { useDialect } from './dialect';
+
+const { dialect } = useDialect();
+const editorRef = shallowRef();
+const outputRef = shallowRef();
+
+const code = `// Knex code\nknex("table").select()\n`;
+const sql = `--- generated SQL code\nselect\n  *\nfrom\n  "table"\n`;
+
+const editorOptions = {
+  scrollBeyondLastLine: false,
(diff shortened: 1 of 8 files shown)


docs/.vitepress/theme/Playground.vue
rluvaton, line 1: 1. Does it support multi instances?
2. Can you provide option for having prefilled text + specific language? I want to integrate it in the docs for each example

docs/.vitepress/theme/Playground.vue
```diff
@@ -0,0 +1,152 @@
...
+
+<template>
+  <div
+    class="playground split"
+    :class="{ 'split-vertical': isVertical, 'split-horizontal': !isVertical }"
+  >
+    <vue-monaco-editor
+      id="editor"
+      language="typescript"
+      theme="vs-dark"
```
rluvaton, line 128: This is only dark mode...

rluvaton (commented):
Thanks, will continue review later

docs/.vitepress/theme/Playground.vue
```diff
@@ -0,0 +1,152 @@
...
+    });
+}
+
+watch([code, dialect], () => {
+  let output = '--- generated SQL code\n';
+
+  try {
+    const knex = Knex({ client: dialect.value });
+
+    output += eval(code.value).toQuery();
```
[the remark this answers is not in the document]
dgadelha: We receive the code as string, and we need to call the `toQuery()` method in the code to get the resulting query, I see this method as the only way to achieve this

docs/.vitepress/theme/Playground.vue
dgadelha: yeah, I need help with syncing the light/dark mode from the docs, had no success with that, not enough vue experience

docs/.vitepress/theme/Playground.vue
dgadelha: it should probably support multi instances, but you should probably attempt to measure page performance, because it instantiates two monaco-editors for each <Playground> instance

rluvaton:
I've fixed the PR to support dark mode + light mode

docs/.vitepress/theme/Playground.vue
rluvaton: fixed

docs/.vitepress/theme/Playground.vue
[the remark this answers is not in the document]
rluvaton: Too risky, lets try find another way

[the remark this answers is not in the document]
rluvaton: You might wanna look how https://github.com/wirekang/kysely-playground does it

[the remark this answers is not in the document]
rluvaton: They are using iframe which is better approach

PR #6060, opened 2024-04-19
Do not overwrite cwd if knexfile and cwd is passed

petar-lazarov:
When having both `--cwd` argument and `--knexfile` argument, the knexfile basepath overwrites the process cwd which is strange IMO.

If we pass `--cwd` it should have higher precedence than the basepath of the `--knexfile` for process cwd selection.

My use case for knex is strange. I'm trying to integrate it in firebase project and I want the configuration values for password and username to be loaded by firebase runtime configuration.

The file structure looks like this
```
.
├── src
│   ├── migrations
│   │   └── ...
│   ├── index.ts
│   ├── knexfile.ts
├── .runtimeconfig.json
```

I wanted the `knexfile.ts` and `migrations` to be under `src` directory but the `.runtimeconfig.json` should be in the parent directory and it contains the values for those variables.

When I execute from `.`:
```
knex --knexfile src/knexfile.ts --cwd . migrate:status
```
the working directory is initially `.` but after that it is overwritten by `./src/`.

You cannot tell to firebase where to find its .runtimeconfig.json. It always search for it in the current working directory I think.

I found two pull requests that are related to that topic and based on this PR #3613 and the table which was created by @briandamaged (https://docs.google.com/spreadsheets/d/19TMouhxdIosYGzubO-2lbxm8Ykx4jR-Iu74HXohFYMg/edit#gid=0) I realised that this might be a bug which I think was introduced by PR #4122.

This PR addresses #6103

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -87,7 +87,7 @@ async function initKnex(env, opts, useDefaultClientIfNotSpecified) {
 function invoke() {
   const filetypes = ['js', 'mjs', 'coffee', 'ts', 'eg', 'ls'];
 
-  const cwd = argv.knexfile
+  const cwd = argv.knexfile && !argv.cwd
     ? path.dirname(path.resolve(argv.knexfile))
     : process.cwd();
 
--- a/test/cli/knexfile-test.spec.js
+++ b/test/cli/knexfile-test.spec.js
@@ -164,20 +164,40 @@ module.exports = {
               }
             );
           });
+
+          it("does not change the process's cwd", function () {
+            return execCommand(
+              `node ${KNEX} migrate:latest --cwd=test/jake-util/knexfile --knexfile=subdir/knexfile.js`,
+              {
+                notExpectedOutput: 'Working directory changed to',
+              }
+            );
+          });
         });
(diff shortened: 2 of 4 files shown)


Emanuel-Palestino:
Yeah, it doesn't work :c

petar-lazarov:
> Yeah, it doesn't work :c

Hey @Emanuel-Palestino, can you please describe your use case with those flags? What does not work for you?

patrick99e99:
I just ran into this problem..

I am using dotenv, which expects to find the `.env` config file within the project root.

my npm script is basically doing
`ts-node -- ../node_modules/.bin/knex --knexfile ./path/to/some/where/migration.ts --cwd . migrate:latest &&  some-other-script`

when the migration runs I see in the log:
```
Working directory changed to /path/to/some/where
```
And then my code blows up because a file loaded from `some-other-script` is calling dotenv.config() and it is trying to find a `.env` file from inside `/path/to/some/where` where it should have been in the original cwd `.` ...

shouldn't knex be changing back to its prior directory once it finishes what it was doing?

PR #6063, opened 2024-04-22
Updated sqlite3 and better-sqlite3 to use more accurate numeric types

mohammedbalila:
Fixing `sqlite3` and `better-sqlite3` numeric types. When trying to use `knex.decimal()` or `knex.double()` with sqlite3 it turns them to `float`. Issue https://github.com/knex/knex/issues/5789

### DIFF
--- a/test/unit/schema-builder/better-sqlite3.js
+++ b/test/unit/schema-builder/better-sqlite3.js
@@ -755,6 +755,7 @@ describe('BetterSQLite3 SchemaBuilder', function () {
     equal(tableSql[0].sql, 'alter table `users` add column `foo` float');
   });
 
+
   it('adding double', function () {
     tableSql = client
       .schemaBuilder()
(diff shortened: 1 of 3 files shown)


mohammedbalila:
Check this for more context on what the PR does  https://github.com/knex/knex/issues/5789#issuecomment-2041200867

rluvaton (commented):
It would be breaking change even though it's correct, @kibertoad WDYT?

test/unit/schema-builder/better-sqlite3.js
```diff
@@ -755,6 +755,7 @@ describe('BetterSQLite3 SchemaBuilder', function () {
     equal(tableSql[0].sql, 'alter table `users` add column `foo` float');
   });
 
+
```
rluvaton, line 758: Please remove

rluvaton (commented):
Can you also add another tests in integration so we are sure this is working...

kibertoad:
@mohammedbalila integration2 has the modern structure for tests, it is the one that should be used. we probably should rename the other one

mohammedbalila:
@kibertoad should I keep working on this? since @rluvaton mentioned it's a breaking change

kibertoad:
@mohammedbalila it's fine, next version is semver major anyway

gpetrov:
So what happened here? This seems like a no-brainer to me and should be directly merged @mercmobily

mercmobily:
This is something we will get to. However, we are not there yet. If you look at the issue list, you will notice that we are going through a huge process of triaging/deduping everything. We are prioritising that right now. However, I marked this as a bug, so that is will be prioritised once we get the triaging done. Sorry for the wait.

mohammedbalila:
I thought this will never make it

image

Issue #5979, opened 2023-04-14
MariaDB - Migration ALTER statements fail if system versioning is used


ghost:
**System information**
* MariaDB v10.3.38
* Ubuntu v20.04.1
* Knex CLI version: 2.4.2
* Knex Local version: 2.4.2

**Problem**
Any schema changes fail if any tables that have with system versioning are part of the migration. That is, the generated scripts are correct but all ALTER TABLE statements will fail without the proper flag.

**Reproduce issue**
Run a migration on any table using with system versioning that generates an ALTER TABLE statement. Scripts will be correct but the execution will fail if SET @@system_versioning_alter_history = 1; or SET @@system_versioning_alter_history = KEEP; isn't at the beginning of the transaction.
Here is a post about this issue: https://mariadb.com/resources/blog/use-cases-for-mariadb-data-versioning/

**Error**
migration failed with error: alter table `applications` modify `description` text, modify `app_name` varchar(100) - UNKNOWN_CODE_PLEASE_REPORT: Not allowed for system-versioned `test_db`.`applications`. Change @@system_versioning_alter_history to proceed with ALTER.

mercmobily:
Hello and thank you for reporting this issue!

First of all, I will leave this issue open as we will need to decide whether we might add this to the official documentation, or maybe even have a mechanism to automate this based on the DB/Version of the DB. 

This is a legitimate issue when working with MariaDB's system-versioned tables. The good news is that Knex already provides a solution through migration lifecycle hooks that allow you to execute database-specific setup commands before your migrations run.

MariaDB system-versioned tables indeed require the `@@system_versioning_alter_history` variable to be set before any ALTER TABLE operations can be performed. Without this setting, you'll encounter the error you've described. This is a MariaDB-specific requirement for tables that have temporal versioning enabled.

Knex provides migration lifecycle hooks (introduced in version 0.20.0) that allow you to execute custom code before or after migrations. For your use case, you should use the `beforeAll` hook to set the session variable before any migrations in a batch are run. Here's how to do it:

```javascript
// In your knexfile.js or when calling migrate.latest()
module.exports = {
  development: {
    client: 'mysql2',
    connection: {
      // your connection config
    },
    migrations: {
      directory: './migrations',
      beforeAll: async (knex) => {
        // This runs once before all migrations in the batch
        await knex.raw('SET @@system_versioning_alter_history = 1');
      }
    }
  }
};
```

[...]

Issue #5963, opened 2023-01-12
MariaDB sequence create and nextValue


vsedyshev-intermedia:
Implement by raw knex wrapper sequence under MaridDB since 10.3 sequence mechanism.
Could you please see posiblities add kind of code in next releases:
```
function nextSequenceValue(knex, sequenceName) {
    return knex.raw(`NEXTVAL(${sequenceName})`);
};
```
and
```
function createSequence(knex, sequenceName, {startWith = undefined, incrementBy = undefined, minValue = undefined, maxValue = undefined} = {}) {
    const parts = [
        `CREATE SEQUENCE ${sequenceName}`,
        startWith !== undefined ? `START WITH ${startWith}` : '',
        incrementBy !== undefined ? `INCREMENT BY ${incrementBy}` : '',
        minValue !== undefined ? `MINVALUE=${minValue}` : '',
        maxValue !== undefined ? `MAXVALUE=${maxValue}` : '',
    ];
    return knex.raw(parts.join(' '))
}
```
Thanks.

mercmobily:
Thanks for the feature request! You've identified a useful capability that MariaDB added in version 10.3 - native sequence support using `CREATE SEQUENCE` and `NEXTVAL()` or `NEXT VALUE FOR`.

Currently, Knex doesn't have native support for MariaDB sequences, and implementing this feature would require some foundational work first. Let me explain the current situation and what would need to happen.

Knex currently treats MariaDB as MySQL, using the same dialect for both databases. When you connect to a MariaDB instance, you use `client: 'mysql'` or `client: 'mysql2'` in your configuration, and Knex applies the MySQL dialect to all operations. This approach works well for standard SQL operations that are common to both databases, but it means MariaDB-specific features like sequences aren't available through Knex's schema builder.

The sequence functionality you've described - `CREATE SEQUENCE` with options like `START WITH`, `INCREMENT BY`, `MINVALUE`, and `MAXVALUE`, plus the `NEXTVAL()` function - is specific to MariaDB 10.3 and later. MySQL doesn't have this feature, which is why Knex's MySQL dialect doesn't support it.

To implement sequence support properly, Knex would first need a dedicated MariaDB dialect. This is actually being discussed in issue #5331, which requests native support for the `mariadb` npm package (MariaDB's official Node.js connector). Once Knex has a proper MariaDB dialect, MariaDB-specific features like sequences could be added to it.

[...]

Issue #5637, opened 2023-07-20
MariaDB UUID type not used


AdamSGit:
# Environment

Knex version: 2.5.1
Database + version: MariaDB 11.2
OS: Macos
Adapter: mysql2 

# Feature discussion / request

MariaDB support native UUID data type since 10.7. When creating a schema with knex as follow : 
```sql
table.uuid('id').primary().defaultTo(knex.fn.uuid());
```
Without `useBinaryUuid` option, the created field is a `char(36)`. With it, its a `binary(16)`. The behavior seems to be the same as with mysql. There might be 2 potentials issues here :  

- Default UUID maria type is not used
- Since not used, when making it a binary field with `useBinaryUuid`, the default UUID() function trigger an error (`Data too long for column`), as in mariadb it seems you can use this function on `char(36)` or `uuid` types, but not on `binary(16)`.
- Please note that for the same reason, `UUID_TO_BIN` and `BIN_TO_UUID` functions doesn't exists anymore in mariadb.

I've looked around the codebase, and located the place the change should take place, it would be in `ColumnCompiler_MySQL`. The thing is, `Client_MySQL` doesn't seems to have a `version()` method like the postgres client have, that would also identify mysql or mariadb.

So, this change would require to : 
- Add a `version()` method in `Client_MySQL` to determine type (mariadb or mysql) and version.
- Add an `uuid` method of class `ColumnCompiler_MySQL` that would use `uuid` type if maria >= 10.7.

Is this looking good ? And do you want a PR for it ?

On a side note, did you noticed that there is a mariadb nodejs adapter that seems faster than mysql and mysql2, and also have some additionals features such as batch insert ? Do you plan to support it ? I could take care of it as well. _Ps : I did read in contributing guide that adapters should be implemented in separate packages, but in this case, this adapter is very close and use the same api as mysql and mysql2 with better performance, so it make sense to have it by default in knex._

mercmobily:
This is a MySQL/MariaDB-specific issue. There are multiple MySQL/MariaDB issues that may be consolidated into a comprehensive META in the future (similar to the proposed Oracle META with 15+ issues and MSSQL META with 18+ issues).

PR #6069, opened 2024-05-07
feat: add dropUniqueIfExists

### LINKED ISSUE #2167: "drop constraint if exists" unavailable
Based on the documentation, there seems to be no way of including an "IF EXISTS" condition when dropping constraints such as foreign keys, so "ALTER TABLE 'x' DROP CONSTRAINT IF EXISTS 'x_y_foreign' ", for instance, can only be done using a raw string. Moreover, a "table.raw" method does not exist, so one has to make the entire ALTER TABLE statement raw. This can be a pretty awkward problem when running scripts that generate schemas. A convoluted workaround is available using whereExists, but it would be ideal if table.foreign etc had an argument to include the "IF EXISTS" condition.

louispotok:
closes #2167 

Could in principle do the same for `dropForeign` and `dropPrimary` but haven't done that yet.

DRI:@mercmobily


mercmobily:
I am onto it. Sorry for the long wait.

mercmobily:
I rebased it, I will wait to see what CI has to say and will get on to reviewing it

mercmobily:
It's still failing some tests. I am happy to deal with the test coverage and the linting, but some tests seem to be actually failing:

[Unit Tests / Unit Tests (16.x) (pull_request)](https://github.com/knex/knex/actions/runs/20154976279/job/57855501774?pr=6069)
Unit Tests / Unit Tests (16.x) (pull_request)Cancelled after 1m
[Unit Tests / Unit Tests (18.x) (pull_request)](https://github.com/knex/knex/actions/runs/20154976279/job/57855501783?pr=6069)
Unit Tests / Unit Tests (18.x) (pull_request)Cancelled after 1m
[Unit Tests / Unit Tests (20.x) (pull_request)](https://github.com/knex/knex/actions/runs/20154976279/job/57855501770?pr=6069)
Unit Tests / Unit Tests (20.x) (pull_request)Cancelled after 1m

Please have a look

@louispotok

louispotok:
@mercmobily Thanks - I think it should pass now.

mercmobily:
@louispotok I spent quite a lot of time on this, to make sure it actually closes the ticket; I also fixed the linting and test coverage problems, and a couple of bugs that were there. I am not done yet! I need to check everything over properly. I will let you know. I will also pass this on to another maintainer for review, since I did so much of it and am frankly losing subjectivity. Thank you.

louispotok:
@mercmobily Amazing, thanks for jumping in and helping out.

mercmobily:
> @mercmobily Amazing, thanks for jumping in and helping out.

@louispotok All good. I would be very grateful if you could jump on and actually check that things work in real code. I  know we have tests but...

mercmobily:
@kibertoad @myndzi This PR's scope grew a little.
The main unexpected change is that MariaDB DOES support `drop index if exists` and I _really_ didn't want Knex to lie and say that it wasn't supported. To do that, I had to add a check for MariaDB. This will also make these possible: 

* #5637 The reporter explicitly notes MariaDB ≥10.7 has a native UUID type and asks for a MariaDB-aware code path to use it. Our version/engine detection is the prerequisite to gate that behaviour without breaking MySQL.

* #5963 Sequences exist only on MariaDB ≥10.3. Emitting CREATE SEQUENCE/ NEXTVAL should be gated to MariaDB; detection is required to safely expose that API.

* #5979 The fix needs MariaDB-specific session flags (SET @@system_versioning_alter_history…) before ALTER TABLE. Detection would let us apply those guards only on MariaDB.

I thought about making a different PR for the detection, but then this feels like the right spot.

Since I have authored most of the code, I should't be the reviewer.  I will leave it open to others to pick up.

lib/dialects/mysql/index.js
```diff
@@ -191,6 +198,60 @@ class Client_MySQL extends Client {
...
+    this._versionCheckPromise = (async () => {
+      try {
+        const query = promisify(connection.query).bind(connection);
+        const rows = await query('select version() as version');
+        const rawVersion =
+          rows?.[0]?.version || rows?.[0]?.VERSION || rows?.[0]?.['VERSION()'];
+
+        if (!rawVersion) {
+          throw new Error('Unable to retrieve MySQL/MariaDB version');
+        }
```
myndzi, line 224: Failure to parse a version is our fault, not the user's; I think, in this case, it would be better to allow the version to be "unknown" and produce a warning instead.

Since this information is used to positively gate features, failing closed means that the library can still be used, except that the gated feature(s) will be unavailable.

If there's an oversight in our code, or if something changes on e.g. mysql/mariadb's end that makes the parsing no longer succeed, then the scope of the failure is reduced from "can't use knex at all" to "can't use a specific feature"

lib/dialects/mysql/index.js
```diff
@@ -191,6 +198,60 @@ class Client_MySQL extends Client {
       options: {},
     });
   }
+
+  async checkVersion(connection) {
+    if (this._versionCheckPromise) {
```
myndzi, line 206: Minor: since we're already mutating connection objects, it wouldn't really be any worse for us to store the version on the connection instead of the client.

We're not necessarily guaranteed that every connection ever returned by the pool is the same database, so doing it this way means that if, e.g., an operator fails over a database to a backup that's running an older version, knex may start doing the wrong thing by assuming a newer version than it is now connected to.

(minor because this appears to be how the other dialects do it, but I think it's still worth a mention)

lib/dialects/mysql/index.js
```diff
@@ -191,6 +198,60 @@ class Client_MySQL extends Client {
...
+      return this._versionCheckPromise;
+    }
+
+    this._versionCheckPromise = (async () => {
+      try {
+        const query = promisify(connection.query).bind(connection);
+        const rows = await query('select version() as version');
+        const rawVersion =
+          rows?.[0]?.version || rows?.[0]?.VERSION || rows?.[0]?.['VERSION()'];
```
myndzi: I'm a little confused by this syntax - do some versions of MySQL / MariaDB not return the column with the alias that was explicitly provided?

test/unit/schema-builder/cockroachdb.js
```diff
@@ -20,4 +20,56 @@ describe('CockroachDB SchemaBuilder', function () {
...
+  it('drop unique if exists, custom', function () {
+    tableSql = client
+      .schemaBuilder()
+      .table('users', function (table) {
+        table.dropUniqueIfExists(null, 'foo');
+      })
+      .toSQL();
+    equal(1, tableSql.length);
+    expect(tableSql[0].sql).to.equal(
+      'drop index if exists "users"@"foo" cascade '
```
myndzi: Same as ~~above~~ below (stray space?)

test/unit/schema-builder/cockroachdb.js
```diff
@@ -20,4 +20,56 @@ describe('CockroachDB SchemaBuilder', function () {
...
+  it('drop unique if exists', function () {
+    tableSql = client
+      .schemaBuilder()
+      .table('users', function (table) {
+        table.dropUniqueIfExists('foo');
+      })
+      .toSQL();
+    equal(1, tableSql.length);
+    expect(tableSql[0].sql).to.equal(
+      'drop index if exists "users"@"users_foo_unique" cascade '
```
myndzi: What's with the stray space at the end of this expectation? Is that an artifact of query construction or a typo? If it's a typo, why didn't the test fail?

myndzi (changes requested):
Looks pretty straightforward; left a couple questions but my main concern is the potential risk of a failed connection loop due to a version string. Reconnecting isn't likely to make the version parsing regex work any better the next time, so this change introduces a possible risk of "update knex -> now I can't connect to my database".

I'm about a 7/10 on that one since there's a clear path to graceful degredation instead of hard failure and this PR includes the only code that depends on it, but I recognize that the other dialect drivers that do version detection also have this behavior, so I can be talked down ;)

test/unit/schema-builder/cockroachdb.js
mercmobily: I traced that space back to the original submission, sorry I missed it -- I should have noticed.

test/unit/schema-builder/cockroachdb.js
mercmobily: I traced that space back to the original submission, sorry I missed it -- I should have noticed.

lib/dialects/mysql/index.js
mercmobily: No, this is on me. I didn't check Vs Code's autocomplete and I assumed that it was for good reason. I normally check those, but I had gotten tired at this point -- sorry.  I looked into it properly, and no, no good reason at all.  Deleted.

lib/dialects/mysql/index.js
mercmobily: I admit it, I am torn. So...

You are right - in theory. Yes, in theory the db could potentially change under the library's butt; however, I also think it would be *really* bad practice, to have a different (!) / lesser (?!?) version as an alternative. I've been around a while, and I've never -- ever -- seen that. If I *did* see it, I would probably question what the DBA was drinking (smoking?) when he set it up, and ask for some.

I don't know. I guess I am resisting this, since it would mean 1) Store the version in the connection, change the tests 2) Open an issue so that we fix other dialects 3) Understand that we are making one extra query per connection, every single connection, to ask something that really... really shouldn't change.

I might be missing something though.

lib/dialects/mysql/index.js
mercmobily: But it wouldn't be just "can't use the feature". It would actually and possibly open up features that shouldn't be opened up, and which will return errors. As dialects diverge, we might in time even end up sending the wrong SQL commands... I don't know about this one.  Do we want "Knex won't work, something really odd is happening and version strings no longer match" or "A probably ignored warning, and then you get weird behaviours and have no idea why"?

But, you feel strongly about this. So... I would much rather add a new PR which reads like "Uniform behaviour across dialects in case the version string can't be parsed", specifying what the defaults would be, rather than "fixing" it for a specific dialect now, leave the other one behaving differently...

test/unit/schema-builder/cockroachdb.js
myndzi: ... why didn't the test fail, if the space is incorrect? 🤔

Either we're not running the test, or the test is succeeding incorrectly

mercmobily: The query builder created the query with the space, hence the pass:

https://github.com/knex/knex/blob/b8ec46a539aa50885223491e37f0ede6fb71827c/lib/dialects/cockroachdb/crdb-tablecompiler.js#L33

lib/dialects/mysql/index.js
mercmobily: Test

lib/dialects/mysql/index.js
myndzi: (did you miss a commit / not push?)

lib/dialects/mysql/index.js
myndzi: I agree that I don't expect it to be likely, but that's what defines an edge case, y'know? "this one query is failing occasionally on some of our app instances and we can't figure out why - turns out it's failing on all instances of the app that were launched after we upgraded our database and were not redeployed after we rolled back in that incident to address the performance regression that took us over the version threshold".

That's the kind of thing that's unlikely to happen, but when it does, is _super_ painful to figure out.

Regardless, it's an edge case we already have; I'm like a 0/10 on fixing this now in this PR, but it might be worth noting for future improvements. I'm sold on the wiki page ....

lib/dialects/mysql/index.js
mercmobily: Yeah. In fact now I wonder which other PR will try and make this change, and fail for a conflict... UGH. Repushed.

lib/dialects/mysql/index.js
mercmobily: I am for adding a n issue and re-evaluate whether we fail on connection or not. In the same issue that will implement versioning for ALL DBs. Let me know if you agree and I will create the issue, close this comment.

lib/dialects/mysql/index.js
myndzi: Please bear with me, I'm going to try and organize my thoughts and reduce back-and-forth :) 

## Regex version parsing

Regex is so easy to mess up, and version strings can change their shape and content over time. You see distro-level builds adding their own strings and formatting onto things, backported security fixes, and other random garbage. They're usually well-structured, but if we _don't know_ the structure that they're expected to have, we're just guessing. I've dealt with production incidents that came down to "we deployed a new version of this thing, and now it produces data that no longer matches our expectation".

If we're just guessing, my opinion is that we should ensure that we don't cause a total outage if our guess is wrong. If we're not guessing, we need to support that by demonstrating (with documentation, source code, or empirical data) that we have good coverage. I'm not happy with "one test on one version string from one instance of maridb" for the empirical approach.

## Value vs risk

The value of this code, in this file, at this time, is that we're able to run one new query: "drop index if exists". The risk is a total outage. That seems pretty one-sided to me.

Example: User upgrades (or security-patches) database. Version string no longer matches regex. Application refuses to connect. User now has an incident to deal with, where their only actionable options are to roll back their database or monkey patch a library dependency.

## Consistency across dialects

I agree with you that we should treat things consistently, and would be bullish on a PR to do that!

The reason _this_ dialect in _this_ change is different is that it's being merged into a codebase with like 2 million downloads a week and it hasn't been out in the wild yet. So I'm erring on the side of caution and thinking through the lens of "migrating code safely in place".

## Unexpected behavior

I agree that, in general, "something unknown is wrong" can be interpreted to mean "we can't rely on anything and we should make the user look into it". The difference here is that we know what's wrong, and we know its impact. What's wrong is "we failed to parse the version string". We aslo know:

- The user has told us we're talking to mysql/mariadb
- The version data is used to _opt in_ to some behavior

There is no code that I know of that would "open up features that shouldn't be opened up", and I'd be hesitant to write any feature-detected code that does so. Acknowledging that I may be missing some real cases, I expect code that checks database versions to be of the form: `if ( i know that this works ) { do the thing }`. Which means that "we don't know the version" is equivalent to "we don't know for certain that it will work", and thus we "don't do the thing"

If you have something in mind here, let me know? I may well have missed something

myndzi: I should give you a clear actionable path forward here :sweat_smile: 

Here are some resolutions:
- A statement that the risk is acceptable (for any number of reasons, but if one of those reasons is "we'll fix it in another PR", a clear mechanism to hold us to that before publishing a release)
- Point out something I missed (such as a concrete use-case / reason that behaving cautiously here can cause a real problem)
- Or: we're agreed on adding a change to this PR

mercmobily: I tend to agree with you, but I do want to add my 2c.
I agree that checking if the version is obtainable, and failing if the version fetching fails, is very drastic.
However:

1) Who pushes onto production a version of their app after updating libraries (or even ONE library) or (worse) updating the *database server* without going through staging first? I mean, really, even a sole developer of a little app will at least run the app _once_ on their computer after a lib upgrade

OK, this is your example:

"Example: User upgrades (or security-patches) database. Version string no longer matches regex. Application refuses to connect. User now has an incident to deal with, where their only actionable options are to roll back their database or monkey patch a library dependency."

This means that the following needs to happen: 1) There is an urgent upgrade to the DB to be done 2) The upgrade is done with no testing. I mean, EVEN in a "security update, we are in a hurry" situation... If Knex refuses to connect, and provides a very clear message "Knex could connect, but failed to understand the database version. Expected: (regexp), received: [returned output].

What is worse, then: to deal with a clear error like that, or to try to figure out why we are pushing queries that no longer work in an obscure (but maybe critical) part of the application?

2) Assuming that there *are* issues with the versioning string, wouldn't a _total_ failure be actually favourable?

3) The main aim of changing the code, and query for a version in *each* connection, is in the fact that the DB server may have changed under the app's butt (see: the main server is down, and the replacement is a little different). How often is *that* gonna happen? And is this corner case (and calling it "corner" is doing it a favour...) worth making a change that will ask the versioning in *each* connection? OK I realise they are only a tiny query, but still, that's a tiny query for every knex app out there for every connection...

Now, am *I* missing something?

The only real problem I see, which you mention and makes a LOT of sense, is this:

"The reason this dialect in this change is different is that it's being merged into a codebase with like 2 million downloads a week and it hasn't been out in the wild yet."

So I propose this:

1) Keep the check at connection level. If the replacement DB is a different version with less functionality, sack the DBA
2) Have a `strictDbVersion` option that will cause the DB to crash if the version is not parseable. In that case, we need to take the most conservative route in terms of features. Issue a warning if the version is not parseable, or an error, depending on this flag
3) Add a PR that will add `strictDbVersion` to the other part of the app with version dependency

Yes?

lib/dialects/mysql/index.js
mercmobily: I just realised I am muddling things up. Moving things to the connection is linked to the "do we exit on versioning regex problems", but it solves a different issue: "DO we cover the case where a connection within the same app happens to be on a lesser/different DB version", in which case I say "No, I would opt to sack the DBA in that case"...

myndzi: Yeah, this thread is resolved. I just wanted your thoughts since we had the opportunity to do this one in a potentially better way.

lib/dialects/mysql/index.js
myndzi: Rather than reply to various points, I want to focus on what I think is the heart of our impedance mismatch:

> [...] try to figure out why we are pushing queries that no longer work in an obscure (but maybe critical) part of the application [...]

which motivates this:

> Have a strictDbVersion option that will cause the DB to crash if the version is not parseable [...]

You seem to believe that "I don't know the version" means that knex will spray weird queries everywhere, and I don't. Do you have a reason for that belief?

I called this out in my previous post, but it was amidst a bunch of other comments:

> I expect code that checks database versions to be of the form: `if ( i know that this works ) { do the thing }`

As best I can tell, either my expectation here is not true (e.g. there is code in knex that I haven't seen that performs queries that are _not supported_ by the database when some version check _does not pass_) -- or you disagree with the statement entirely (e.g. you believe that we _should_ be able to and/or want to be able to write code like this in the future)

Is it one of these? Or, why do you hold the belief that "not knowing the version" means we'd be "pushing queries that don't work"?

lib/dialects/mysql/index.js
mercmobily: Wait wait, if you think it's worthwhile doing, it probably is. At the very least, I will open an issue where we can discuss this properly and decide there -- agreed?

lib/dialects/mysql/index.js
mercmobily: In terms of "now", we have very little:

* Postgres: potentially the wrong DDL commands are issued with json/jsonb used/not used when they shouldn't/should be
* Oracle: DDLs might fail because of the 30 character limit

But what about the future? For Postgres we really should modernise how we make auto increment records. This is in existing issues. This would be behind a version check. I assume as we work on issues, we will  potentially gate more things behind version control. It's not a "now" problem but...

mercmobily: I don't feel strongly enough for my own argument and you have expressed in good detail that it's very unlikely that the "wrong sql" will be ever issued, especially if gating is done well.  So I will fix it. But the inconsistency with the other dialect really craps me, so I will make an issue about uniforming it/putting it behind an option

lib/dialects/mysql/index.js
myndzi: Sure. I just mean it's not particularly critical and certainly doesn't block this PR

mercmobily:
@kibertoad this is not finished yet: there are discussion on whether it's OK or not to make it fail if the version string cannot be parsed. The consensus is that it shouldn't fail since it might start failing for legit installations out there after release. I would like you to tell us what you think about this, and possibly review everything else about this PR.

lib/dialects/mysql/index.js
kibertoad: I think we need an escape hatch here, allowing user to explicitly opt-in into using certain features even if we fail to parse version or map version to what is supported wrongly. otherwise we may fail to support some specific flavours of same DB, which in theory do support things we want to use, but decided to use a different versioning convention.

and yes, we should log a warning but do not throw in case we fail to parse version, because again, maybe everything works just great, version number is an implementation detail. if something is actually broken, let it break explicitly.

I would even go as far as to suggest that if we fail to parse a version, by default assume that all the conditional features are supported - and let them break if they are not.

mercmobily: @kibertoad

The "allow everything" is a stance I didn't consider, but I like it. If parsing fails, assume super-latest version of the server. And yes to escape hatch.
So, a better approach would be:

* Offer the escape hatch which would effectively bypass any gating
* If version regexp fails, allow any gating.

The reason why I threw was because it would be consistent with what happens already in the codebase:

* Postgres throws if the version query doesn't work and effectively breaks if the string isn't parsed
* Oracle throws

Here is what I propose: 

* We add the escape hatch here as discussed
* If version is not parsed properly, we log a warning and we assume version = ∞
* I open another PR where I uniform behaviour of Postgres and Oracle. I could even add functions to the dialect's interface and get every dialect to implement them them, and make sure that *this* PR creates the foundation for this work

This way, EVERY dialect knows which version it is; if it doesn't, because the query fails or because the version parsing fails, then it assumes no gating.

Yes?

myndzi: I'm in favor of kibertoad's suggestion of just letting failing queries fail, unless at some point we identify a feature that will _behave wrongly_. I don't know of any cases like that with what the current code is doing though; they're all syntax things that will just produce a syntax error(?)

As for an escape hatch, that need is covered by the "explicitly specify a version" behavior. If the user specifies a version, then knex behaves as though it's talking to that database version. You might support something like `latest` as an explicit "just wide open do whatever you can". This would at least keep from tacking on "yet another option to the bag".

The simplest rendering of the ideas presented here is:

- Allow user to specify a database version. If they do so, behave in accordance with that version, and perform no introspection.
- Escape hatch: If they explicitly specify 'latest', behave as though the database is of the latest version we know/care about.
- If the user does not specify a database version, query the database and parse the version to figure it out
- If unparseable, behave as though the user had specified 'latest'

mercmobily:
@myndzi over to you. I think this is possibly good to go. Please note that I have created another issue, #6364, to make sure that behaviour and detection is actually standardised across dialect. I have already assigned it to me. I have a bunch of issues I will go through once I've reviewed the pending PRs. I want to get this one soon though, while the DB versioning stuff is fresh in my head.

mercmobily:
This is done, already approved by @kibertoad, and finally merged!

PR #6073, opened 2024-05-14
fix TypeError on close stream event

proydenko94:
Fix TypeError: Cannot read properties of undefined (reading '__knexUid'). this.onnection may not exist in some cases


nvedenyuk (approved):

mercmobily:
DUPE of #6197

PR #6088, opened 2024-06-21
refactor: remove resolve dependency

elliotttf:
This replaces the dependency on resolve-from with node's native `require.resolve` method.

If this and gulpjs/rechoir#47 are merged, a false positive related to [resolve's `monorepo-symlink-test`](https://github.com/browserify/resolve/issues/312) file will be resolved.


elliotttf:
Disregard, this PR isn't necessary (but neither is resolve-from if it's desirable to be removed)

PR #5741, opened 2023-11-17
Import knex as type in TS migration template

OliverWales:
To match `@typescript-eslint/consistent-type-imports` eslint rule.

PR #6094, opened 2024-07-02
Import knex as type in TS seed template

OliverWales:
Relates #5741

Saves fixing `@typescript-eslint/consistent-type-imports` eslint rule every time.


kibertoad (approved):

kibertoad:
thank you!

PR #6104, opened 2024-08-05
fix:drop only FKRefs which are related to changed column

SzymonLisowiec:
## Bug Description
When using `renameColumn()` in a migration for MySQL, Knex drops all foreign keys of the table, renames the specified column, and then re-adds all the foreign keys. The correct behavior should be for Knex to drop only the foreign keys referencing the column being renamed.

## Steps to Reproduce
1. Create the following tables:
   - `users` with columns `id`, `name`, and `email`
   - `user_books` with columns `id`, `userId`, and `name`. The `userId` column should have a foreign key referencing `users.id`.
2. Create a migration for `users` table that renames the `name` column to `nickname` using `table.renameColumn('name', 'nickname')`.

### DIFF
--- a/lib/dialects/mysql/schema/mysql-tablecompiler.js
+++ b/lib/dialects/mysql/schema/mysql-tablecompiler.js
@@ -83,7 +83,7 @@ class TableCompiler_MySQL extends TableCompiler {
       output(resp) {
         const column = resp[0];
         const runner = this;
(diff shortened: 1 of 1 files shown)


kibertoad (changes requested):
Thank you for your submission! Could you please add a test for it?

PR #6120, opened 2024-09-05
add MariaDB to README

robertsilen:
MariaDB is mentioned at https://knexjs.org/, so suggest to add MariaDB to github README too :) I asked in discussion https://github.com/knex/knex/issues/6119 about MariaDB support in general too.


robertsilen:
Any help in approving this?

Issue #6119, opened 2024-09-05
MariaDB support?


robertsilen:
Does knex support MariaDB?

tluyben:
It does, we are using it with mariadb; the only issues we found are on both mysql & mariadb; nothing mariadb specific yet.

mercmobily:
I was a little ahead of myself closing this. I think we will merge shortly, I will reopen otherwise.

PR #6127, opened 2024-09-19
add mention of to MariaDB README.md

robertsilen:
Add mention of MariaDB to README.md as per discussion https://github.com/knex/knex/issues/6119


mercmobily:
Dupe of #6119

PR #6130, opened 2024-09-22
fix: join clause literal value is treated as Double quotation mark in Postgres Database

YongSeok-Choi51:
# Related Issue

[Literal Value treated as Column name in Where Clause (Postgres Database)]
- https://github.com/knex/knex/issues/6124

## Issue Detail

```javscript
db // Knex instance

// Basic query that includes a join with literal join value
const qb = db('table_a').innerJoin('table_b', join => {
  join.onVal('table_b.id', '123')
})

// Executing select on this query results in proper formatting of literal value
// SQL: select "table_a"."id" from "table_a" inner join "table_b" on "table_b"."id" = '123' 👈👈 (This Allows in Postgres!)
console.log(qb.select('table_a.id').toQuery())

// Executing delete on this query results in literal value being formatted as if it were a column name
// SQL: delete from "table_a" using "table_b" where "table_b"."id" = "123"  👈👈 (This is Not Allowed in Postgres. 123 must be single quoted like '123')
console.log(qb.delete().toQuery())
```

## Description (Solution)

i fixed this issue by using `builder._queryContext` that cannot be shared to string format method (`wrapIdentifierImpl`)

### DIFF
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -63,8 +63,12 @@ class Client_PG extends Client {
     return require('pg');
   }
 
-  wrapIdentifierImpl(value) {
+  wrapIdentifierImpl(value, queryContext) {
     if (value === '*') return value;
+
+    const isLiteralValue = () => {
+      return queryContext && queryContext.pgValueFlag;
+    };
 
     let arrayAccessor = '';
     const arrayAccessorMatch = value.match(/(.*?)(\[[0-9]+\])/);
@@ -74,7 +78,9 @@ class Client_PG extends Client {
       arrayAccessor = arrayAccessorMatch[2];
     }
 
-    return `"${value.replace(/"/g, '""')}"${arrayAccessor}`;
+    return isLiteralValue()
+      ? `'${value.replace(/"/g, '""')}'${arrayAccessor}`
+      : `"${value.replace(/"/g, '""')}"${arrayAccessor}`;
   }
 
   _acquireOnlyConnection() {
--- a/lib/client.js
+++ b/lib/client.js
@@ -186,7 +186,7 @@ class Client extends EventEmitter {
     if (this.config.wrapIdentifier) {
       return this.config.wrapIdentifier(value, origImpl, queryContext);
     }
-    return origImpl(value);
+    return origImpl(value, queryContext);
   }
 
   wrapIdentifierImpl(value) {
--- a/lib/query/querycompiler.js
+++ b/lib/query/querycompiler.js
@@ -977,6 +977,12 @@ class QueryCompiler {
   // ------
 
   _valueClause(statement) {
+    // When DataBase is Postgres. it should be handled with single quotation.
+    if (this._getDBVendor() === 'pg') {
+      const exists = this.builder._queryContext || {};
+      this.builder.queryContext({ ...exists, pgValueFlag: true });
+    }
+
     return statement.asColumn
       ? wrap_(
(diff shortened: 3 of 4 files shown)


lib/dialects/postgres/index.js
```diff
@@ -74,7 +78,9 @@ class Client_PG extends Client {
       arrayAccessor = arrayAccessorMatch[2];
     }
 
-    return `"${value.replace(/"/g, '""')}"${arrayAccessor}`;
+    return isLiteralValue()
+      ? `'${value.replace(/"/g, '""')}'${arrayAccessor}`
```
YongSeok-Choi51, line 82: `wrapIdentifierImpl` method is implemented in each DBVender Class(ex. oracle, mysql, mssql) 

berows are other DB Class Implementations

[mssql - wrapIdentifierImpl](https://github.com/knex/knex/blob/176151d8048b2a7feeb89a3d649a5580786d4f4e/lib/dialects/mssql/index.js#L131) 

[mysql - wrapIdentifierImpl](https://github.com/knex/knex/blob/176151d8048b2a7feeb89a3d649a5580786d4f4e/lib/dialects/mysql/index.js#L59)

[sqlite3 - wrapIdentifierImpl](https://github.com/knex/knex/blob/176151d8048b2a7feeb89a3d649a5580786d4f4e/lib/dialects/sqlite3/index.js#L77)

mercmobily:
Thank you for taking the time to fix this bug and write a test! The fix does work and solves the immediate problem.

However, after reviewing the code more carefully, I think there's a cleaner approach that would be more maintainable and fix additional related bugs at the same time.

The current fix uses a flag in queryContext to change how `wrapIdentifier` behaves, but this feels like a workaround rather than addressing the root cause. The real issue is that the PostgreSQL DELETE compiler in `lib/dialects/postgres/query/pg-querycompiler.js:104-113` doesn't handle different join clause types properly.

Looking at how the standard join compiler works (in `lib/query/querycompiler.js:456`), it calls the appropriate handler method based on clause type:

```javascript
const val = this[clause.type](clause);
```

This means:
- `onBasic` clauses call `this.onBasic()` which wraps both sides as identifiers
- `onVal` clauses call `this.onVal()` which treats the value as a literal parameter
- `onIn` clauses call `this.onIn()` which handles arrays properly
- And so on...

The PostgreSQL DELETE code should follow the same pattern instead of forcing everything through `whereBasic()` with `asColumn: true`. This would fix not just `onVal`, but also `onIn`, `onBetween`, `onNull`, and other join clause types that are currently also broken in DELETE queries.

Would you be interested in updating your PR to use this approach instead? The change would be in `lib/dialects/postgres/query/pg-querycompiler.js`, replacing lines 104-113 with something like:

```javascript
const joinWheres = [];
for (const clause of join.clauses) {
  // Call the appropriate handler based on clause type, just like regular joins
  const clauseSql = this[clause.type](clause);
  if (clauseSql) {
    joinWheres.push(clauseSql);
  }
}
```

This would be a cleaner fix that follows the existing patterns in the codebase. What do you think?

YoungSeok-Choi:
@mercmobily Thanks for detail reviewing of this PR! 🙏🏻 i totally with you! i'll do more look the code base and write codes based on your direction.

YongSeok-Choi51:
> @mercmobily Thanks for detail reviewing of this PR! 🙏🏻 i totally with you! i'll do more look the code base and write codes based on your direction.

by the way, that is my second account :)

PR #6132, opened 2024-09-26
pg-native is now part of node-postgres for easier maintenance

melroyvandenberg:
See title.. 

And proof here: https://github.com/brianc/node-pg-native?tab=readme-ov-file#%EF%B8%8F-notice-%EF%B8%8F


mercmobily:
DUPE of #6121

PR #6136, opened 2024-10-04
fix: skip sqlite filename warning when connection is a function

innerdvations:
## Summary

When `connection` is a function, Knex stores it as a `connectionConfigProvider` and does not resolve settings synchronously in the constructor. The sqlite constructor was checking `config.connection.filename` directly, which is always `undefined` for a function — so users with a valid dynamic config got a false warning.

Guard the existing constructor check with `typeof config.connection !== 'function'`. Static misconfiguration (e.g. `connection: {}` without `filename`) still warns at construction.

## Test plan

- [x] `test/integration2/dialects/sqlite.spec.js` — `connection.filename warning` block for both `sqlite3` and `better-sqlite3`
- [x] Targeted mocha grep for the new tests — 6 passing

### DIFF
--- a/lib/dialects/better-sqlite3/index.js
+++ b/lib/dialects/better-sqlite3/index.js
@@ -9,8 +9,6 @@ class Client_BetterSQLite3 extends Client_SQLite3 {
 
   // Get a raw connection from the database, returning a promise with the connection object.
   async acquireRawConnection() {
-    this._warnIfMissingFilename();
-
     const options = this.connectionSettings.options || {};
 
     const db = new this.driver(this.connectionSettings.filename, {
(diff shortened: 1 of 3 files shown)


tgriesser:
I think the GPT generated code here is overkill, the ideal fix is just to just change:

```ts
if (config.connection && config.connection.filename === undefined) {
```
 
 to
 
```ts
if (config.connection && typeof config.connection !== 'function' && config.connection.filename === undefined) {
```

since what we're basically assuming is that if you've specified a function for connection, we don't want/need to warn.

innerdvations:
Ahh, that's what I had originally 😆 I figured if we defer we could still validate that it does eventually resolve to a filename. Either way works for me. Updated now to the simpler version.

PR #6138, opened 2024-10-09
Replace the use of knex with the knexInstance that is created using the config

ebenkok:
The sample code for the typescript implementation creates an instance of the knex using the config, but does not use this instance variable when querying the db. 

This results in a SQLITE_ERROR (using mssql)

This commit makes use of the instance and resolves the problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ const config: Knex.Config = {
 const knexInstance = knex(config);
 
 try {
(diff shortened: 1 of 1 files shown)


n0x001:
The fix is also relevant to the following PR: [https://github.com/knex/knex/pull/6191](https://github.com/knex/knex/pull/6191)

mercmobily:
DUPE of 6191

PR #6156, opened 2024-11-22
Add missing `peerDependencies` in `package.json`

stekycz:
The key [`peerDependencies`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependencies) should be used to define the version range for peer dependencies. It is missing but it seems to kind of work in the most common package managers even with [`peerDependenciesMeta`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependenciesmeta). However, `peerDependenciesMeta` are designed to define possible optionality of the dependency only (as of 2024-11-22 and according to my knowledge and understanding).

Missing the `peerDependencies` may possibly break usage of `knex` with build tools. We do face an issue using [Rush.js](https://rushjs.io), more specifically its [`rush deploy` command](https://rushjs.io/pages/commands/rush_deploy/). The effect of missing `peerDependencies` in the `package.json` of `knex` is the Rush.js omits the optional peer dependency from the deployable files even though the peer dependency is actually present.

You can see this as an issue in Rush.js but I think the `peerDependencies` should be defined at the level of `knex` anyway. I believe there is a specific versions range of each peer dependency the `knex` can work with rather than any version of(which is what I propose here as at least some and quick step forward).


Tharos:
Is there any update on resolving this issue? Due to the reasons mentioned by @stekycz, we're currently forced to patch our package.json as a workaround…

```
--- a/package.json
+++ b/package.json
@@ -76,7 +76,14 @@
     "tildify": "2.0.0"
   },
   "peerDependencies": {
-    "pg-query-stream": "^4.14.0"
+    "tedious": "*",
+    "mysql": "*",
+    "mysql2": "*",
+    "pg": "*",
+    "pg-native": "*",
+    "pg-query-stream": "^4.14.0",
+    "sqlite3": "*",
+    "better-sqlite3": "*"
   },
   "peerDependenciesMeta": {
     "tedious": {
```

PR #6159, opened 2024-12-04
fix: update types for migrator config

3ddyBoi:
## Why
In the docs and code it seems you can use a function to create a migration name. But the type does not reflect that.

See:
https://knexjs.org/guide/migrations.html#custom-migration-name
https://github.com/knex/knex/blob/176151d8048b2a7feeb89a3d649a5580786d4f4e/lib/migrations/migrate/MigrationGenerator.js#L42-L47

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -3149,6 +3149,7 @@ declare namespace Knex {
     loadExtensions?: readonly string[];
     migrationSource?: MigrationSource<unknown>;
     name?: string;
+    getNewMigrationName?: (name: string) => string;
   }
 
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -3149,6 +3149,7 @@ declare namespace Knex {
     loadExtensions?: readonly string[];
     migrationSource?: MigrationSource<unknown>;
     name?: string;
+    getNewMigrationName?: (name: string) => string;
```
kibertoad, line 3152: I'm not sure this is accurate. it's not a part of migrator config, to my knowledge

PR #6167, opened 2025-01-05
Update commander, drop Node 16 support

kibertoad:


myndzi:
@kibertoad what was the plan with this PR? I see that it is labeled a breaking change; should I hold onto it until the next... when? 

I have to figure that a straight dependency update shouldn't need a major semver bump, but the tests seem to be legitimately failing

PR #6169, opened 2025-01-14
Fix for foreign key's referencing table schemas.

tisdadd:
Thanks for such a great project!

I used patch-package with the following when I ran into an issue with foreign keys utilizing schemas as mentioned in the title, so the basic fix for those who use that in the meantime is below.

```diff
diff --git a/node_modules/knex/lib/schema/tablebuilder.js b/node_modules/knex/lib/schema/tablebuilder.js
index 99f849a..5d78396 100644
--- a/node_modules/knex/lib/schema/tablebuilder.js
+++ b/node_modules/knex/lib/schema/tablebuilder.js
@@ -109,8 +109,9 @@ class TableBuilder {
             },
           };
         }
-        foreignData.inTable = pieces[0];
-        foreignData.references = pieces[1];
+        
+        foreignData.references = pieces.pop();
+        foreignData.inTable = pieces.join('.');
         return returnObj;
       },
       withKeyName(keyName) {
```

After reading the contributing guidelines, I also created some migrations and tests for them based off of my main project where I was actually running through my own checklist of tests. As I only have set up for `mssql` so far, I have not tested against other drivers.

As for why I did `if(db === DRIVER_NAMES.MsSQL) {` instead of `if (isMsSql(knex)) {`, through some digging I found that will always return false at that level, since `knex` is undefined there.

During the digging, it looked like `test/util/db-helpers.js`  `function getDriverName(knex)` always had undefined for `knex` passed in, so I am not sure if the switch statements are working for other bits or not. If you would like me to dig into that, I can try to find time as well.

Anyway, thanks again and let me know if I should change anything.

### DIFF
--- a/lib/schema/tablebuilder.js
+++ b/lib/schema/tablebuilder.js
@@ -109,8 +109,8 @@ class TableBuilder {
             },
           };
         }
-        foreignData.inTable = pieces[0];
-        foreignData.references = pieces[1];
+        foreignData.references = pieces.pop();
+        foreignData.inTable = pieces.join('.');
         return returnObj;
       },
       withKeyName(keyName) {
--- a/test/integration2/migrate/migration-integration.spec.js
+++ b/test/integration2/migrate/migration-integration.spec.js
@@ -28,6 +28,7 @@ const { getAllDbs, getKnexForDb } = require('../util/knex-instance-provider');
 const {
   ensureTable,
 } = require('../../../lib/migrations/migrate/table-creator');
+const { DRIVER_NAMES } = require('../../../lib/constants');
 
 describe('Migrations', function () {
   getAllDbs().forEach((db) => {
@@ -80,6 +81,20 @@ describe('Migrations', function () {
               });
             });
         });
+
+        if(db === DRIVER_NAMES.MsSQL) {
+          it('should not fail creating a foreign key referencing a schema', () => {
+            return knex.migrate
+              .latest({
+                directory: 'test/integration2/migrate/test_foreign_key_with_schema',
+              })
+              .then(() => {
+                return knex.migrate.rollback({
+                  directory: 'test/integration2/migrate/test_foreign_key_with_schema',
+                });
+              });
+          });
+        }
 
         if (isPostgreSQL(knex)) {
           it('should not fail drop-and-recreate-column operation when using promise chain and schema', () => {
--- a/test/integration2/migrate/test_foreign_key_with_schema/20250113173201_schema_setup.js
+++ b/test/integration2/migrate/test_foreign_key_with_schema/20250113173201_schema_setup.js
@@ -1,0 +1,13 @@
+exports.up = function up(knex) {
+  if(knex.client.dialect !== 'mssql') {
+    throw new Error('mssql only supported for this test setup at the moment')
+  }
+  return knex.raw('CREATE SCHEMA testSchema')
+}
+
+exports.down = function down(knex) {
+  if(knex.client.dialect !== 'mssql') {
+    throw new Error('mssql only supported for this test setup at the moment')
+  }
+  return knex.raw('DROP SCHEMA testSchema');
+}
--- a/test/integration2/migrate/test_foreign_key_with_schema/20250113173202_foreign_keys_with_schema.js
+++ b/test/integration2/migrate/test_foreign_key_with_schema/20250113173202_foreign_keys_with_schema.js
@@ -1,0 +1,31 @@
+exports.up = function up(knex) {
+  return knex.schema
+    .withSchema('testSchema')
+    .createTable('LeftSide', (table) => {
+      table.increments('LeftSideId').primary();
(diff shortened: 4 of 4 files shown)


tisdadd:
Forgot to mention, that while searching for solutions to this I did find [PR 5651](https://github.com/knex/knex/pull/5651)as well, which handles something slightly different, and may not allow referencing another schema if needed. If both features are wanted, we may need to work through that. Microsoft has the WideWorldImporters example, which has several examples of Foreign Keys from the `[Application]` schema being utilized elsewhere, for instance `[Sales].[Customers].[AlternateContactPersonID]` references `[Application].[People].[PersonID]`, so the ability to specify what schema is referenced should generally be good, but we may consider needing to just check if there is a schema here, and if not and there is a schema from `withSchema` then utilize that one as per the other PR.

mercmobily:
@tisdadd Are you still around to help out with this?

mercmobily:
@tisdadd I think your PR is great! After reviewing both your solution and PR #5651, yours provides the better foundation by fixing the fundamental parsing issue.

Would you be willing to enhance your PR to also incorporate the automatic schema inheritance feature from PR #5651? Specifically:

1. After your parsing fix, add logic to check if no schema was explicitly specified
2. If not, and `withSchema()` context exists, automatically prepend it
3. This would look something like:

```javascript
// In tablebuilder.js, after your parsing logic
if (!foreignData.inTable.includes('.') && this._schemaName) {
  foreignData.inTable = `${this._schemaName}.${foreignData.inTable}`;
}
```

This combined approach would provide both:
- Your fix for explicit cross-schema references (`application.people`)
- Automatic schema inheritance when working within a single schema

Also, could you extend the tests to cover PostgreSQL alongside MSSQL?

If you're able to make these enhancements, we can close PR #5651 and move forward with your more comprehensive solution.

tisdadd:
Hello @mercmobily - I am still around and should be able to do so hopefully over the next few days. If things are moving ahead more, should I look over more local patches I have around Oracle Privileges and Schemas/Owners in migrations and make a separate PR for that as well? I would have already but I was not sure of the rhythm of things on this repo, and had not wanted to not add more in until then.

mercmobily:
Oh I am so glad you are still here... I was *really* hoping to have an active Oracle person around who could help! Things are moving quite a lot now. I am triaging every single open issue (yes, one by one). I should be finished in a couple of weeks. We will THEN move onto PRs -- luckily only post-2015 PRs are around, so it's all recent stuff.  I would absolutely be happy to have some, but please understand that I need to triage everything first, so that we KNOW what issue(s) each PR closes and we have a good idea of what is going on.

THANK YOU for being around!

tisdadd:
I am glad I am able to be here still. I have included the proposed, and looking at the Oracle tests can see what you mean. I did include in some code that is working in the project I have been working on but converted to the style here in order to also be able to run the tests for MySQL2 and OracleDB as well and the tests are not running as a high enough privileged user for one, and I realized that the Oracle would need the below added for the current version.

```sql
SELECT VALUE
FROM V$PARAMETER
WHERE NAME = 'common_user_prefix'
```
I started adding it and then saw some test errors with migrations being locked already are introduced, so reverted back to just what I run against 21.3 right now. 

If you would like me to dig in more or remove the code around that I can do so, but not sure how long digging in will take. As 18c is not supported now since June of 2021, I am not sure how far back this project should support.

Thank you for your hard work @mercmobily, and if you would like me to look at anything let me know. I am getting ready to release a product soon that will have this as a core part, and supports `mssql`, `mysql2`, `oracledb`, `postgres` and `better-sql3` drivers so I should be familiar enough to deal with some stuff around those as needed, as long as it is a reasonably modern version of the underlying database.

After this is closed, I will work on the PR for the other bits I have around Oracle.

mercmobily:
I am still in the triaging stage. Are you able to release your product and _then_ get this fixed? Or do you need to resolve this before release? I will prioritise things accordingly

tisdadd:
I am able to release and then get this fixed as needed. I already have patched in what I need with patch-package. I was just saying that I am familiar with several things and if it helps I should be able to do some more on this... right now I am doing the automatic deployment artifact tests for my stuff across the three major clouds, so I have time while infrastructure tears down and pulls up.

My product works great without requiring this is updated as I will not be supporting older versions, and I have a patch for any fixes I found that I need.

I just wanted to check if you wanted me to worry about the other things I put in for the tests for the future, and offer to help if you want me to touch anything else specific. Also, I was not sure if we want to run the other tests with higher privileges or not.

I am not waiting on anything from here.

mercmobily:
I would _absolutely_ want help with anything Oracle-related. I am considering driopping older versions of Oracle too; in general, I would love to have you around and even give us your 2c on anything Oracle would be amazing. Just give us time to finish the issue triaging first, so that we will actually _know_ what the actually outstanding issues are.

tisdadd:
Sounds excellent. :+1:

PR #6188, opened 2025-03-25
Types no longer allow knex to be called without tablename

jpike88:
`calling knex without a tableName is deprecated. Use knex.queryBuilder() instead.`

I've seen that error crop up in our code for many years. It's maybe time to knock it on the head.


myndzi (approved):
👍

PR #6191, opened 2025-03-29
Fix the TypeScript example code.

n0x001:
Fixes two errors:

1. use knexInstance for futher db manipuation
2. compilation error of usage of await outside of a module


myndzi (approved):
👍

Good change, the examples should work when copypasted!

Issue #4890, opened 2021-12-13
Task: All tests to date pass on Oracle 18c


code-ape:
## Overview

**What:** It was recently found with the addition of CI tests for Oracle 18c (#4888 and #4889) that not all tests are passing for it. This should be fixed for proper functionality on changes and releases.

**Why:** CI has not run for Oracle for over a year since Travis CI [shutdown their free OSS tier in November 2020](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing).

**How:** All broken tests are listed below in check list form to track their resolution.

## Tests to Fix

**Current status:**  304 passing, 61 pending, 34 failing. All known failing listed below with failure summary.

- Inserts / oracledb /
    - [ ] should handle multi inserts
        - Failure: `AssertionError: expected [ '2', '3' ] to deeply equal [ '1', '2' ]`
    - [ ] should take hashes passed into insert and keep them in the correct order
        - Failure: `AssertionError: expected [ '4', '5' ] to deeply equal [ '1', '2' ]`
    - [ ] should drop any where clause bindings
        - Failure: `AssertionError: expected [ '8' ] to deeply equal [ '1' ]`
    - [ ] should take an array of columns to return in oracle or postgres
        - Expected: `"test_table_two" ("account_id", "details", "status") values (?, ?, ?) returning "account_id","details" into ?,?' to equal 'insert into "test_table_two"`
        - Received: `"insert into "test_table_two" ("account_id", "details", "status")\n                 values (?, ?, ?) returning "account_id","details"\n                 into ?,?"`
    - [ ] should allow a * for returning in postgres and oracle
        - Failure: Array of element 1 different in object key `id`, expected `4` got `1`.
- Additional / oracledb /
    - [ ] other operations / should insert binary uuid and retreive it (typo in word "retreive")
        - Error: `create table "uuid_table" ("uuid_col_binary" binary(16)) - ORA-00907: missing right parenthesis`
- Joins / oracledb /
    - [ ] has a leftJoin method parameter to specify the join type
        - Failure: Numerous differences in objects returned. Specifically `id` and `details` fields.
    - [ ] supports join aliases with advanced joins
       - Failure: Different object returned.
- Selects / oracledb /
    - [ ] returns an array of a single column with `pluck`
        - Failure: `AssertionError: expected [ 7, 8, 9, 10, 11, 12 ] to deeply equal [ 1, 2, 3, 4, 5, 6 ]`
    - [ ] can pluck a qualified column name, #1619
[...]

PR #6200, opened 2025-04-29
Expose oracledb fetch array size

### LINKED ISSUE #4964: Support node-oracledb 'arraysize' for tuning multi-row fetch performance
# Environment

Knex version: master
Database + version: Oracle
OS: all

## Feature discussion / request

@atiertant the current recommended way to tune fetching of multiple rows is with [`fetchArraySize`](https://oracle.github.io/node-oracledb/doc/api.html#propdbfetcharraysize).  Knex can support this similar to how `prefetchRows` and `stmtCacheSize` are supported.  

Add 

```
if (client.connectionSettings.fetchArraySize) {
  client.driver.fetchArraySize = client.connectionSettings.fetchArraySize;
}
```

at around https://github.com/knex/knex/blob/master/lib/dialects/oracledb/index.js#L138

justin-cotarla:
Closes #4964

This PR exposes the oracledb `fetchArraySize` parameter used for tuning large requests. 

As far as I can tell, OracleDB tests are not in a good state (#4890); that being said this change does not affect any core logic other than exposing a driver-level parameter and making sure it is being leveraged when fetching data.


justin-cotarla:
@kibertoad mind taking a look?

Issue #6201, opened 2025-05-01
Add dialect support for Spanner Postgres


code-ape:
**Why:** GCP Spanner supports the Postgres dialect via their [PGAdapter](https://cloud.google.com/spanner/docs/pgadapter) proxy service. However, this supports a notably limited subset of Postgres functionality -- for example, the concept of schema's has been fully stripped as Spanner doesn't have schemas. This means things like migrations don't work for Spanner Postgres, because it uses `current_schema()` which causes an error.

**What:** There needs to be a way to express these limitations to Knex via configuration so that Spanner Postgres will work. This pattern will presumably be used in the future by other databases that expose limited Postgres interfaces, as this is a growing practice amongst database systems.

**Proposed Solution:** Add a new dialect `postgres-spanner` which is a wrapper around the standard `postgres` interface and overrides the few functions that needs to be slightly different.

**Next Steps:** Get feedback from maintainers (@kibertoad and @OlivierCavadenti). Discuss proposed solution versus alternative ideas. Open PR for this.

mercmobily:
Are you still happy to work on this?

PR #6206, opened 2025-05-19
Issue #6201: Add support for Spanner Postgres

code-ape:
## Related PRs and Issues

1. #6201

## Reason 

This PR addresses issue #6201 by adding a dedicated dialect for Google Cloud Spanner's flavor of "Postgres". This is because Spanner's Postgres implementation (via `pgadapter`) is a limited subset of the full Postgres dialect and thus minor changes are needed.

## Todo

- [ ] Code changes
- [ ] Add tests
- [ ] Update docs

## Changes

1. Added dialect `postgres-spanner` via `lib/constants.js`, `lib/dialects/index.js`, and `lib/dialects/postgres-spanner/`.
2. Added Docker Compose entries for the local Spanner emulator and PgAdapter which proxies the Postgres dialect requests, via `scripts/docker-compose.yml`.

## References
1. https://cloud.google.com/spanner/docs/postgresql-interface
2. https://cloud.google.com/spanner/docs/pgadapter

### DIFF
--- a/lib/dialects/index.ts
+++ b/lib/dialects/index.ts
@@ -11,6 +11,7 @@ const dbNameToDialectLoader: Record<string, () => any> = Object.freeze({
   oracledb: () => require('./oracledb'),
   pgnative: () => require('./pgnative'),
   postgres: () => require('./postgres'),
+  'postgres-spanner': () => require('./postgres-spanner'),
   redshift: () => require('./redshift'),
   sqlite3: () => require('./sqlite3'),
 });
--- a/scripts/docker-compose.yml
+++ b/scripts/docker-compose.yml
@@ -135,19 +135,18 @@ services:
       - 9010:9010
       - 9020:9020
 
-
   spanner-pgadapter:
     # Most recent container as of 2025-04-25
     image: gcr.io/cloud-spanner-pg-adapter/pgadapter:v0.46.2
     container_name: uis-test-spanner-pgadapter
     command:
-      - "-p test-project"
-      - "-i test-instance"
-      - "-r autoConfigEmulator=true"
-      - "-e spanner-emulator:9010"
-      - "-ddl=AutocommitExplicitTransaction"
-      - "-c \"\""
-      - "-x"
+      - '-p test-project'
+      - '-i test-instance'
+      - '-r autoConfigEmulator=true'
+      - '-e spanner-emulator:9010'
+      - '-ddl=AutocommitExplicitTransaction'
+      - '-c ""'
+      - '-x'
     # Mapping to port 25434 as the next available port following the incrementing of Postgres
     # services in this file
     ports:
--- a/scripts/docker-compose.yml
+++ b/scripts/docker-compose.yml
@@ -135,8 +135,7 @@ services:
       - 9010:9010
       - 9020:9020
 
-  # pgadapter exposes Spanner's PostgreSQL interface on the pg wire protocol.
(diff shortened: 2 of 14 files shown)


lib/dialects/index.ts
```diff
@@ -10,6 +10,7 @@ const dbNameToDialectLoader: Record<string, () => any> = Object.freeze({
   oracledb: () => require('./oracledb'),
   pgnative: () => require('./pgnative'),
   postgres: () => require('./postgres'),
+  'postgres-spanner': () => require('./postgres-spanner'),
```
kibertoad, line 14: can you also add it to documentation?

scripts/docker-compose.yml
```diff
@@ -84,6 +84,34 @@ services:
       - -c
       - 'until /usr/local/bin/psql postgres://testuser:knextest@postgres/knex_test -c "SELECT 1"; do sleep 5; done'
 
+  spanner-emulator:
```
kibertoad, line 131: can you configure integration tests to run on this?

(probably not all of them would be passing, and it's ok to skip the ones that are hard to fix immediately, but the more coverage we can get, the better)

OlivierCavadenti:
Picked this up and rebased on master. 

Fixed two things: the dialect mutated the base `pg` client prototype (broke all postgres), and `columnInfo` used `current_database()`/`current_schema()` which Spanner doesn't support.

The shared integration suite assumes schemas and migrations that Spanner lacks (~133 failures in the common setup), so I added a dedicated spec running on the Spanner emulator: CRUD, `where`, `columnInfo`, `hasTable`/`hasColumn`. Docs updated, CI green.

n1ghtmare:
Thank you @OlivierCavadenti for picking this up, much appreciated. 

One thing that this PR doesn't cover yet that we've had to patch downstream, is migration rollback ordering:

https://github.com/knex/knex/blob/8355ad758db9ebc00aacdb8ca6f6d56f28bb5f4a/lib/migrations/migrate/Migrator.js#L602-L617

`_getLastBatch()` orders the last batch by `id DESC`, which assumes the auto-increment `id` ascends with insertion order. That doesn't hold up on Spanner. The migrations table is created with `t.increments()`, which on the Postgres dialiect compiles to `SERIAL` - and per the spanner docs, [`SERIAL`/`AUTO_INCREMENT` are  DDL aliases for `IDENTITY` columns](https://cloud.google.com/spanner/docs/primary-key-default-value) backed by **bit-reversed sequences**, whose values are "evenly distributed across the key space so that they don't cause hotspots" rather than monotonic. This is by design - Spanner's [schema-design best practices](https://cloud.google.com/spanner/docs/schema-design) say *"Do not choose a column whose value monotonically increases or decreases as the first key part…"*

So on Spanner, `ORDER BY id DESC` does **not** return the last batch in reverse insertion order, and `migrate:rollback` can run the batch's `down()` steps out of order.

We patch this to order by `migration_time`. Suggested fix that's safe for **all** dialects (unless I'm missing something, sorry if that's the case):

  ```js
  .orderBy('migration_time', 'desc')
  .orderBy('id', 'desc');
  ```
  
(The `migration_time` tiebreaker matters because, I think, a fast batch can stamp several rows with the same `new Date();` the secondary `id` keeps ordering deterministic on dialects where it's monotonic, and is harmless on Spanner.)
  
  Since Spanner support is the goal of this PR, would it make sense to include this here?

PR #6244, opened 2025-08-20
docs: formulated behavior of knex.transaction more explicitly

0x0dd4b2:
Thank you for this great library! Working with the library I also use `knex.transaction` method. To understand the method's behavior, I read the respective part in knex's [guide](https://knexjs.org/guide/transactions.html).

After I understood how `knex.transaction` behaves, when given a handler, I felt that the guide's formulation could be improved a bit, hence this PR.

If there's anything else that needs to be done for this PR to be considered mergeable, let me know!


myndzi (approved):
👍

Issue #4939, opened 2022-01-17
API to abort a query/stream


ghost:
# Environment

Knex version: 0.21.19
Database + version: Postgres 10.17
OS: MacOS Monterey

# Feature discussion / request

1. I'm trying to handling an HTTP request that retrieves a lot of data with a stream from my database and returns the stream to the HTTP client. The response takes usually 7~10s with a payload of 5MB.
2. I've already tried by using `stream.end()` or `stream.destroy()` but it seems that the pool does not release the connection correctly and I stumble across two kind of errors:
   - `the pool is probably full` given that the connections are not correctly released
   - `write after stream end/destroyed` since any pending query whose stream has been ended tries to write to the stream after the pool gets free
3. I think something like `query.abort()` API should be fine

# What I've tried

Since there was no simple way to solve this issue I've tried to imitate the operations done by knex when a client is timed out (using `timeout()`), and I've use the tarn pool API to work around the problem:

```javascript
// === the handler of the request. EG: server.on('request')

          // in order to catch the pending acquiring from the pool
          // I had to wrap the pool.acquire function in order to decorate it
          // with my requestID
          const originalAcquire = knex.client.pool.acquire;
          knex.client.pool.acquire = function acquire() {
            const pendingAcquire = originalAcquire.apply(knex.client.pool);
            pendingAcquire.__requestId = requestId;
            return pendingAcquire;
          };

          // get the query stream
          const stream = query.stream();

          // save the stream just created in a map.
          this.activeQueries[requestId] = {
            stream
          };

          // when stream ends, remove it from the map
          stream.on('end', () => {
            delete this.activeQueries[requestId];
          });

          const that = this;
          // using tarn API wait for connection in order to get the resource of the pool and save it in the map
          const waitForConnection = function(_, resource) {
            that.activeQueries[requestId].resource = resource;
            knex.client.pool.removeListener('acquireSuccess', waitForConnection);
          };
          knex.client.pool.on('acquireSuccess', waitForConnection);

// === stop request handler
```

In the `activeQueries` map I have:
- `stream` the stream returned to the client
- `resource` the knex resource returned from the pool with the connection used

Here the code to stop the query:

```javascript
// === the handler of the request end. EG: request.on('end')

     // get the query data from the map
      if (!this.activeQueries[requestId]) {
        return;
      }

      const { stream, resource } = this.activeQueries[requestId];

      // search for a pending connection, yet to be acquired, with the same requestId
[...]

PR #6253, opened 2025-08-22
[GH COPILOT ASSISTED PR] Support Query Cancelation

### LINKED ISSUE #707: Support query cancellation

exoRift:
Closes #707

I wanted to test out GIthub's Copilot agent feature so I gave it the task of implementing something I and a lot of people have wanted in Knex which is query canceling.

- The code looks solid to me (maybe too many comments which can be trimmed down)
- I ran the tests and they passed (may change them to be `SELECT pg_sleep`s to actually ensure in-progress abortion)
- It looks like it properly uses existing well-defined cancelation infrastructure
- It adds documentation and types

The Agent PR entry can be viewed here: https://github.com/exoRift/knex/pull/1


mercmobily:
Oh wow... This is something that is really quite tricky. In the coming weeks (maybe couple of months), we intend to get into the swing of reviewing PRs very quickly and either merge them, or close them. But right now, we are sorting out the tickets, so this one is going to take a little longer because it is a very tricky one. I read all of the past issues and saw the relevant code, and it screams "edge cases!" everywhere. Sorry if we won't get to it as quickly as I would like.

mercmobily:
While you are around...

Thank you @exoRift for this  PR implementing query cancellation with AbortSignal! This is a long-awaited feature that many users have been requesting since 2015 (#707). The implementation looks good at a quick glance, and follows the exact pattern proposed in the original discussion.

I've reviewed the PR and noticed that it currently supports MySQL and PostgreSQL. Since we already have cancellation infrastructure for other drivers, it would be valuable to extend this PR to ensure full compatibility with all drivers that support query cancellation:

**Drivers that should be tested/verified:**
- YES PostgreSQL (`canCancelQuery: true` at lib/dialects/postgres/index.js:301)
- YES MySQL/MySQL2 (`canCancelQuery: true` at lib/dialects/mysql/index.js:203)
- YES PostgreSQL Native (`canCancelQuery: true` at lib/dialects/pgnative/index.js:9) - Should work but needs verification
- NO CockroachDB (inherits from PostgreSQL, has custom `cancelQuery()` at lib/dialects/cockroachdb/index.js:43) - Should work but needs verification

The tests in `test/integration2/query/misc/abort-on-signal.spec.js` correctly check for `knex.client.canCancelQuery` before running, which is good. However, we should ensure the feature works correctly with all four drivers that support cancellation.

**Suggested additions:**
1. Update the documentation to mention that PostgreSQL Native and CockroachDB are also supported (if testing confirms they work)
2. Consider adding a note about which drivers don't support cancellation (SQLite3, Better-SQLite3, MSSQL, Oracle, OracleDB, Redshift) to set clear expectations
3. Run the integration tests against PostgreSQL Native and CockroachDB to ensure compatibility

For MSSQL, whilst it doesn't currently have cancellation support in Knex, there's a working community implementation shown in #4939 that uses the tedious driver's request cancellation. This could be explored as a future enhancement, but shouldn't block this PR.

We need to decide what to do with the DB engine that do NOT support it. The consensus in the code is to throw an error when attempting to use cancellation features on drivers that don't support it, rather than silently failing.

Overall, this is a great addition to Knex that solves a critical need for many applications. With verification of the additional driver support, this PR would provide a comprehensive solution for query cancellation across all capable drivers.

exoRift:
What's the timeline on this PR being approved & merged?

(And the overall timeline on the next release, as well)

mercmobily:
We are going through ALL PRs in the coming weeks. We have slowed down in the review as we are getting everything ready for consistent releasing. It's all happening...

exoRift:
I noticed that releases have resumed on Knex. What's the timeline looking like for this feature?

exoRift:
Hi! Any updates? 😭

PR #6264, opened 2025-08-24
Fix migration CLI and cli tests

### LINKED ISSUE #6267: Migration CLI fails when there's nothing to do
https://github.com/knex/knex/pull/5541 introduced lifecycle hooks into the migrator, but avoids running them if there are no migrations to execute. It returns an empty array from `_runBatch` in this case, but if there is work to do it returns the result of `this._waterfallBatch`, which differs: `_waterfallBatch` returns `[0, []]` when it has nothing to do.

This causes a confusing error (`Uncaught TypeError: Cannot read properties of undefined (reading 'length')`) when a user runs migrations from the CLI but no migrations need to be applied (e.g. already at latest / earliest).

### LINKED ISSUE #6268: `npm run test:cli` does nothing and exits with a success code
https://github.com/knex/knex/pull/5138 bumped Jake, and as part of the migration moved to using `child_process` instead of `jake.createExec`. However, the events emitted by Jake and Node when a process exits are different, so the event handler intended to resolve the promise when the exec finishes never runs.

Since no handles hold the process open, it just exits with 0, doing nothing and performing no tests or assertions.

myndzi:
Fixes #6267 and also fixes #6268

Detailed descriptions are in the commit messages, but here's the high-level summary:

- Changes the jake task names to be more readable
  Before: `Starting 'reateamigrationfile'...`
  Now: `Starting 'create_a_migration_file'...`
- Fixes `migration-test-helper.js`'s `assertExec` and `assertExecError` functions to resolve their promises
- Fixes the `_runBatch` method of `Migrator` to return the same shape of data that it did before the migration hooks PR

Deferred:
- I don't really know how to add `npm test:cli` to CI, but I assume the repo owner has to set that up anyway. It'd be a good idea to have.

### DIFF
--- a/lib/migrations/migrate/Migrator.js
+++ b/lib/migrations/migrate/Migrator.js
@@ -378,7 +378,6 @@ class Migrator {
       );
 
       let batchNo = await this._latestBatchNumber(trx);
-      if (direction === 'up') batchNo++;
 
       // Run any hooks before/after this batch
       const beforeAll = this.config.beforeAll || (() => {});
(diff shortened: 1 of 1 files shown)


mercmobily:
Does this close an existing issue? Sorry I don't mean to be lazy, but it will be easier for you to answer :smile:

myndzi:
> Does this close an existing issue? Sorry I don't mean to be lazy, but it will be easier for you to answer 😄

No worries at all. I had _thought_ this was a prerequisite, since I was working on https://github.com/knex/knex/issues/6082 - and my expectation was "write a test along with the PR". Which of course led me to identifying what set of tests would be appropriate to add the new tests to. That's how I discovered that the tests didn't work :)

It doesn't close an existing thing, but it does cause (some of) the tests (which already exist) to be able to be run again.

mercmobily:
  The end result is that all tests now actually run -- and pass. Related PRs:  #6264 #6310 #6311 #6313 @kibertoad  please merge. @mindzi thank you....

PR #6265, opened 2025-08-24
Improve CLI error reporting in some edge cases

### LINKED ISSUE #6082: Error without actual reason in CLI due to TypeScript error
# Environment

Knex version: 3.1.0
Database + version: Postgres 16
OS: Debian 12

# Bug

I ran `knex seed:run` got error without actual reason, It should give me the detail of `Error`.

2. Error message

```tx
/home/chiro/Git/oracle/node_modules/colorette/index.cjs:70
          ("" + string).indexOf(close, at),
              ^
TypeError: Cannot convert object to primitive value
    at Object.red (/home/chiro/Git/oracle/node_modules/colorette/index.cjs:70:15)
    at exit (/home/chiro/Git/oracle/node_modules/knex/bin/utils/cli-config-utils.js:82:25)
error Command failed with exit code 1.
```

3. Suggest fix

I figured out that, `text` in `console.error(color.red(text));` is an TypeScript error but It isn't `instance of Error`.

```js
function exit(text) {
  if (text instanceof Error) {
    if (text.message) {
      console.error(color.red(text.message));
    }
    console.error(
      color.red(`${text.detail ? `${text.detail}\n` : ''}${text.stack}`)
    );
  } else {
    console.error(color.red(text));
  }
  process.exit(1);
}
```

So I prefer to change this line:

https://github.com/knex/knex/blob/176151d8048b2a7feeb89a3d649a5580786d4f4e/bin/utils/cli-config-utils.js#L82

To:

`console.error(color.red(typeof text === 'string' ? text : inspect(text))); `

myndzi:
Fixes #6082

Apparently CI isn't running the CLI tests, and also `npm run test:cli` doesn't run any of the mocha tests. I have [this related PR](https://github.com/knex/knex/pull/6264) to fix the Jakefile tests (the ones that were run when you use `npm run test:cli`), but the others (in `test/cli/*.spec.js`) don't appear to be run by any of the `npm run test` commands? Some of them are failing. I added new tests for this PR, which do pass, but I'm not sure if it's better to merge this and tackle the CI / CLI tests in a separate PR or what.

Detailed descriptions in the commit messages, here's the summary:

- Ensures that if a knexfile throws, the caught value is wrapped up in an Error, similar to what [other code does](https://github.com/knex/knex/blob/3.1.0/lib/migrations/seed/Seeder.js#L114-L127)
- Ensures that the [exit function in cli-config-utils.js](https://github.com/knex/knex/blob/3.1.0/bin/utils/cli-config-utils.js#L73-L85) can accept any kind of data and log something useful as a result

Here's an example:
```bash
$ node ./node_modules/.bin/knex seed:run --client=sqlite3
Failed to read config from knexfile
A non-Error value was thrown
    at default export (/home/myndzi/knex-user-test/knexfile.js)
Error: Failed to read config from knexfile
    at ensureError (/home/myndzi/knex/bin/cli.js:37:17)
    at openKnexfile (/home/myndzi/knex/bin/cli.js:66:11)
    at async initKnex (/home/myndzi/knex/bin/cli.js:85:7)
[Object: null prototype] { foo: 'bar' }
```

With this knexfile:
```js
module.exports = function () {
  throw Object.assign(Object.create(null), { foo: "bar" });
};
```

DRI:@myndzi

### DIFF
--- a/test/cli/cli-config-utils.spec.js
+++ b/test/cli/cli-config-utils.spec.js
@@ -5,7 +5,7 @@ const { execFile } = require('child_process');
 const { expect } = require('chai');
 
 /**
- * Execute a command, return a promise for its stdout/stderr
+ * Execute _justexit.js as a child process, return a promise for its stdout/stderr
  *
  * @param {...string} args
  * @returns {Promise<{error: import('child_process').ExecException|null, stdout: string, stderr: string}>}


myndzi:
TIL: ESM namespace objects have a null prototype. So that provides a likely explanation for the original bug - I was scratching my head over how people would be _accidentally_ encountering null-prototype objects. But if they return an import promise or namespace object, that'd do it.

mercmobily:
I am looking at this @myndzi  But [one line in the recently merged #6264](https://github.com/knex/knex/pull/6265/files#diff-8df847172b11012ffc21b5cd1bce51ce514cdee7227c3fa8203129400617e65cR29) conflicts with a [change in this one](https://github.com/knex/knex/pull/6265/files#diff-8df847172b11012ffc21b5cd1bce51ce514cdee7227c3fa8203129400617e65cR29). If you have a look, I will get onto this.

myndzi:
Ah, that might have leaked - I think I made the same change in the other PR for the same reason. I'll fix

bin/utils/cli-config-utils.js
```diff
@@ -78,8 +78,19 @@ function exit(text) {
     console.error(
       color.red(`${text.detail ? `${text.detail}\n` : ''}${text.stack}`)
     );
-  } else {
+    // give whatever info we can about a non-error value that was thrown
+    if (text.nonErrorValue) {
```
mercmobily: If nonErrorValue is falsy, we still won't enter here. 
Maybe this would be better:

```
if (Object.prototype.hasOwnProperty.call(text, 'nonErrorValue')) {
    console.error(text.nonErrorValue);
  }
```

test/cli/_justexit.js
```diff
@@ -0,0 +1,20 @@
+const { exit } = require('../../bin/utils/cli-config-utils');
+
+switch (process.argv[2]) {
+  case 'error':
+    exit(new Error('exit(Error)'));
+    break;
+  case 'string':
+    exit(new Error('exit(string)'));
```
mercmobily: This 'string'  branch currently calls exit(new Error('exit(string)')), so it routes through the Error-handling path again and never exercises the new typeof text ===  'string' branch. I wonder if we should switch this to exit('exit(string)') so we actually  cover the string-handling logic? If I'm missing an intentional reason to  keep it as an Error (very likely), happy to be corrected.

mercmobily:
I made a couple of comments in the code; please have a look.

Only one thing... With your recent PRs to re-enable the CLI tests, will these new specs be picked up in CI now, or do we still need to add anything here? Just want to be sure they run as part of the suite.

Other than that... WOW. thank you for chasing down the gnarly CLI edge cases and making the error reporting so much clearer. Wrapping non-Error throws, plumbing the details into exit, and adding focused CLI coverage gives users
  way better signals when their knexfile misbehaves. Also appreciate the breadth of fixtures you added to cover import/function/getter failures; that kind of thoroughness really pays off. This is the "generally improving the structure" you were talking about the other day!

mercmobily:
I just noticed the date... 25th of August...? Did I take THIS long? I would have been in the thick of dealing with the issue triaging back then... wow. I didn't think it was that long ago.

test/cli/_justexit.js
myndzi: This feels like an accidental commit, none of this makes sense

bin/utils/cli-config-utils.js
myndzi: Good callout on the truthy check, but I don't think we need HOP here since it's a known property. `!== undefined` should be enough (and more in line with the intent), and probably a prefix on the log message (I hate when something logs and I can't figure out where it's coming from, so some kind of context message would help here). That would make things a little bit more helpful in the event we capture a `throw 0` or something weird

myndzi:
> Only one thing... With your recent PRs to re-enable the CLI tests, will these new specs be picked up in CI now, or do we still need to add anything here? Just want to be sure they run as part of the suite.

I was hoping that I could get someone familiar with GH actions to figure that out. I don't think all tests are executed in CI, but some are intentionally not executed and others are accidentally not executed because the actions are fanned out to specific databases in some cases, so they don't go through the full "npm test" path. 

I can probably still at least review _which_ tests get run and check whether any are missing, even if I have no idea how to test and develop the actions themselves...

mercmobily:
> > Only one thing... With your recent PRs to re-enable the CLI tests, will these new specs be picked up in CI now, or do we still need to add anything here? Just want to be sure they run as part of the suite.
> 
> I was hoping that I could get someone familiar with GH actions to figure that out. I don't think all tests are executed in CI, but some are intentionally not executed and others are accidentally not executed because the actions are fanned out to specific databases in some cases, so they don't go through the full "npm test" path.
> 
> I can probably still at least review _which_ tests get run and check whether any are missing, even if I have no idea how to test and develop the actions themselves...

@kibertoad Are you familiar with this piece of the puzzle?

test/cli/_justexit.js
mercmobily: ...are you sure?

kibertoad:
@mercmobily I think I am, but need to find a slot to look into it

test/cli/_justexit.js
myndzi: It's been a while, I had to regenerate the context. I was confused by using "exit(Error)" as the string; I'll improve this to make it more understandable

test/cli/cli-config-utils.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    execFile(
+      'node',
+      [path.resolve(__dirname, '_justexit.js'), ...args],
+      (error, stdout, stderr) => {
+        resolve({ error, stdout, stderr });
+      },
+      { timeout: 1000, cwd: __dirname, env: {} }
+    );
+  });
+}
```
myndzi, line 24: cli-testlab doesn't allow for negative assertions on stderr output; but also, with the current state of JS and node, it's pretty ergonomic to just avoid the dependency entirely. Here, instead of deferring the assertions to a library that would have to be updated and published to support more things, we can just get the data directly and perform assertions using whatever assertion library _we_ are using

myndzi:
@mercmobily while addressing your comments I backed my way into a solution that just has a better separation of concerns entirely. I'm rather happier with this solution than the previous one.

mercmobily:
> @mercmobily while addressing your comments I backed my way into a solution that just has a better separation of concerns entirely. I'm rather happier with this solution than the previous one.

I am very happy I could (VERY HUMBLY) inspire. I will review, no worries.

mercmobily:
Well this was a rewrite. Wow... awesome. The code was not very much,  but I actually went on and compared the output of master and the PR. And... WOW. Yes very very very much needed. The extra class is awesome. Generally, amazing work.

mercmobily:
@kibertoad This is ready to merge,

kibertoad (approved):

PR #6273, opened 2025-08-26
Fix PostgreSQL numeric type preservation

### LINKED ISSUE #1001: PostgreSQL adds quotes around number literals.
Adding an issue here because I don't know what the correct behaviour is, but I noticed that PostgreSQL queries seem to add quotes around numbers.

```js
knex('table').where('id', 5).toString();
// "select * from "table" where "id" = 5"

pg('table').where('id', 5).toString();
// "select * from "table" where "id" = '5'"
```

Is this to handle a use case I'm not familiar with?

mercmobily:
This PR fixes the long-standing issue where PostgreSQL queries fail when numeric parameters are converted to strings by the underlying `pg` library. The solution adds automatic type casting for numeric values using PostgreSQL's native `::integer` and `::numeric` syntax.

## The Problem
The `pg` library converts all JavaScript numbers to strings before sending them to PostgreSQL. This causes failures in several scenarios:
- Functions with strict type requirements (e.g., `generate_series(1, 20, 1)`)
- Substring and other string functions requiring integer positions
- Foreign key constraints expecting numeric IDs
- Arithmetic operations in SQL

## The Solution
Override the `parameter` method in the PostgreSQL client to add explicit type casts (`::integer`, `::numeric`) for numeric values. This tells PostgreSQL to interpret the values as numbers, not strings.

## Why Not Just Modify `_valueClause`?

While modifying `_valueClause` in `pg-querycompiler.js` as suggested in [this comment](https://github.com/knex/knex/issues/1001#issuecomment-3223471896) would work for query builder operations, it has significant limitations:

### The `_valueClause` Approach Would Only Fix:
- ✅ WHERE clauses built with the query builder
- ✅ HAVING clauses built with the query builder

### But Would NOT Fix:
- ❌ **Raw queries** - The main complaint in the issue (`knex.raw('generate_series(?, ?, ?)', [1, 20, 1])`)
- ❌ **INSERT values** - Foreign key violations when inserting numeric IDs
- ❌ **UPDATE values** - Setting numeric columns
- ❌ **Function parameters** in raw queries (`substring`, `generate_series`, etc.)
- ❌ **Stored procedure calls** with numeric parameters
- ❌ **Complex raw SQL** with arithmetic operations

### Why the `parameter` Method is Superior:

1. **Universal Coverage**: The `parameter` method is called for ALL parameter bindings, whether from raw queries or the query builder
2. **Single Point of Control**: One method handles all numeric values consistently
3. **Cleaner Architecture**: Type casting logic belongs at the binding level, not scattered across different query builders
4. **Raw Query Support**: Most importantly, it fixes raw queries which are the primary use case in the issue

### Code Comparison:

**Limited `_valueClause` approach:**
```javascript
// Only affects WHERE/HAVING in query builder
_valueClause(statement) {
  if (!statement.asColumn && typeof statement.value === 'number') {
    // ... add casting
  }
}
```

**Comprehensive `parameter` approach (this PR):**
```javascript
// Affects ALL parameter bindings everywhere
parameter(value, builder, bindingsHolder) {
  if (this.useNumericCasting) {
    // ... add casting for all numeric parameters
  }
}
```

### New Files:
- `lib/dialects/postgres/parameter-casting.js` - Core casting logic
- `test/unit/dialects/postgres/parameter-casting.js` - Comprehensive test suite

### Modified Files:
- `lib/dialects/postgres/index.js` - Override `parameter` method to use casting

## Configuration

The feature is enabled by default but can be disabled for backwards compatibility:
```javascript
const knex = Knex({
  client: 'pg',
  useNumericCasting: false // Disable if needed
});
```

## Testing

All tests pass, including specific test cases for:
- `generate_series()` with numeric parameters
- `substring()` with integer position
- Foreign key inserts with numeric IDs
- Mixed type parameters
- Integer range detection (smallint vs integer vs bigint)
- Backwards compatibility mode

## Breaking Changes
None. The feature is enabled by default to fix the broken behaviour, but can be disabled via configuration if any edge cases arise.

## Fixes
Fixes #1001

### DIFF
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -15,6 +15,7 @@ const ViewBuilder = require('./schema/pg-viewbuilder');
 const SchemaCompiler = require('./schema/pg-compiler');
 const { makeEscape } = require('../../util/string');
 const { isString } = require('../../util/is');
+const { parameterWithCasting } = require('./parameter-casting');
 
 class Client_PG extends Client {
   constructor(config) {
@@ -26,7 +27,19 @@ class Client_PG extends Client {
     if (config.searchPath) {
       this.searchPath = config.searchPath;
     }
+    
+    // Control numeric type casting (default: true for fix #1001)
+    this.useNumericCasting = config.useNumericCasting !== false;
   }
+  
+  // Override parameter method to add type casting for numbers
+  parameter(value, builder, bindingsHolder) {
+    if (this.useNumericCasting) {
+      return parameterWithCasting(Client).call(this, value, builder, bindingsHolder);
+    }
+    return super.parameter(value, builder, bindingsHolder);
+  }
+  
   transaction() {
     return new Transaction(this, ...arguments);
   }
--- a/lib/dialects/postgres/parameter-casting.js
+++ b/lib/dialects/postgres/parameter-casting.js
@@ -1,0 +1,113 @@
+// PostgreSQL Parameter Casting Solution for Issue #1001
+// This module provides a simpler approach to preserve numeric types
+// by adding PostgreSQL type casts directly in the SQL
+
+// Override the parameter method to add casting for numbers
+function parameterWithCasting(Client) {
+  return function(value, builder, bindingsHolder) {
+    // First, let the parent handle special cases (functions, raw queries)
+    const baseResult = Client.prototype.parameter.call(
+      this,
+      value, 
+      builder, 
+      bindingsHolder
+    );
+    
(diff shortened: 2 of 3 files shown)


mercmobily:
It's all broken. I pushed this as I got fixated on it while triaging the issues, and I don't have time to fix it now. I will get to it after triaging.

mercmobily:
It's all broken. I pushed this as I got fixated on it while triaging the issues, and I don't have time to fix it now. I will get to it after triaging.

PR #6278, opened 2025-09-18
Add gin_trgm_ops to storageEngineIndexType

atlanteh:
postgres supports gin_trgm_ops, so we can add this to typing


atlanteh:
Nevermind this doesn't actually work

PR #6296, opened 2025-10-23
docs: replaces broken link in streams section

sunsheeppoplar:
Substack link doesn't work anymore. Not sure if this is a good enough swap in for it or if we'd rather leave it blank.

### DIFF
--- a/docs/src/guide/interfaces.md
+++ b/docs/src/guide/interfaces.md
@@ -107,7 +107,7 @@ knex
 
 ## Streams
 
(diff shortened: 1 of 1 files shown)


skimberk:
I also just started trying to fix that link but found your PR.

I was able to find the original markdown file from the `substack` repo: https://raw.githubusercontent.com/Raynos/stream-handbook/ad267fdf0895eb15d6136032032ec866ec82dbda/readme.markdown

Which seems to match the contents of the last saved repo on Internet Archive: https://web.archive.org/web/20220825174324/https://github.com/substack/stream-handbook

Anyway, it seems that link you added (https://github.com/JasonGhent/stream-handbook-epub) mostly matches the original (I ran a diff locally).

skimberk:
@rubengmurray I think this change is reasonable, it replaces a dead link (that is pretty visible from homepage -> streams link) with an alive link that is quite similar to the original.

rubengmurray:
I agree. Feels like an obvious roll-in to `master` on the next release as the new link also works for me.

rubengmurray (approved):
Looks good to me. New link works and provides good stream usage context.

rubengmurray:
@sunsheeppoplar you around to fix the conflict?

sunsheeppoplar:
> @sunsheeppoplar you around to fix the conflict?

Yes, I'm away from a computer at the moment but can resolve tonight

sunsheeppoplar:
@rubengmurray @skimberk fixed the conflict, thanks for the review.

PR #6309, opened 2025-11-11
fix: join clause literal value is treated as Double quotation mark in Postgres Database

YongSeok-Choi51:
# Related Issue

[Literal Value treated as Column name in Where Clause (Postgres Database)]
- https://github.com/knex/knex/issues/6124

[closed previous PR]
- https://github.com/knex/knex/pull/6130

## Issue Detail

```javscript
db // Knex instance

// Basic query that includes a join with literal join value
const qb = db('table_a').innerJoin('table_b', join => {
  join.onVal('table_b.id', '123')
})

// Executing select on this query results in proper formatting of literal value
// SQL: select "table_a"."id" from "table_a" inner join "table_b" on "table_b"."id" = '123' 👈👈 (This Allows in Postgres!)
console.log(qb.select('table_a.id').toQuery())

// Executing delete on this query results in literal value being formatted as if it were a column name
// SQL: delete from "table_a" using "table_b" where "table_b"."id" = "123"  👈👈 (This is Not Allowed in Postgres. 123 must be single quoted like '123')
console.log(qb.delete().toQuery())
```

## Description (Solution)

i solved this issue by fixing pg-querycompiler implementation

### DIFF
--- a/lib/dialects/postgres/query/pg-querycompiler.js
+++ b/lib/dialects/postgres/query/pg-querycompiler.js
@@ -103,19 +103,17 @@ class QueryCompiler_PG extends QueryCompiler {
 
         const joinWheres = [];
         for (const clause of join.clauses) {
-          joinWheres.push(
-            this.whereBasic({
-              column: clause.column,
-              operator: '=',
-              value: clause.value,
-              asColumn: true,
-            })
-          );
+          const clauseSql = this[clause.type](clause);
+          if (clauseSql) {
+            joinWheres.push(clauseSql);
+          }
         }
+
         if (joinWheres.length > 0) {
           wheres += (wheres ? ' and ' : 'where ') + joinWheres.join(' and ');
         }
       }
+
       if (tableJoins.length > 0) {
         using += (using ? ',' : 'using ') + tableJoins.join(',');
       }
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -66,10 +66,6 @@ class Client_PG extends Client {
   wrapIdentifierImpl(value, queryContext) {
     if (value === '*') return value;
 
-    const isLiteralValue = () => {
-      return queryContext && queryContext.pgValueFlag;
-    };
-
     let arrayAccessor = '';
     const arrayAccessorMatch = value.match(/(.*?)(\[[0-9]+\])/);
 
@@ -78,9 +74,7 @@ class Client_PG extends Client {
       arrayAccessor = arrayAccessorMatch[2];
     }
 
-    return isLiteralValue()
-      ? `'${value.replace(/"/g, '""')}'${arrayAccessor}`
-      : `"${value.replace(/"/g, '""')}"${arrayAccessor}`;
+    return `"${value.replace(/"/g, '""')}"${arrayAccessor}`;
   }
 
   _acquireOnlyConnection() {
--- a/lib/query/querycompiler.js
+++ b/lib/query/querycompiler.js
@@ -1020,14 +1020,7 @@ class QueryCompiler {
 
   // Where Clause
   // ------
-
   _valueClause(statement) {
-    // When DataBase is Postgres. it should be handled with single quotation.
-    if (this._getDBVendor() === 'pg') {
-      const exists = this.builder._queryContext || {};
-      this.builder.queryContext({ ...exists, pgValueFlag: true });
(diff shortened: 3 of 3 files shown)


lib/client.js
```diff
@@ -186,7 +186,7 @@ class Client extends EventEmitter {
     if (this.config.wrapIdentifier) {
       return this.config.wrapIdentifier(value, origImpl, queryContext);
     }
-    return origImpl(value);
+    return origImpl(value, queryContext);
```
YongSeok-Choi51, line 189: @mercmobily hi! 

i suggest the parameter `queryContext` pass to function `origImpl` because future usage (e.g. `wrapIdentifierImpl` or else)

please let me know what do you think 🙏

lib/dialects/postgres/query/pg-querycompiler.js
```diff
@@ -103,19 +103,17 @@ class QueryCompiler_PG extends QueryCompiler {
...
         for (const clause of join.clauses) {
-          joinWheres.push(
-            this.whereBasic({
-              column: clause.column,
-              operator: '=',
-              value: clause.value,
-              asColumn: true,
-            })
-          );
+          const clauseSql = this[clause.type](clause);
```
YongSeok-Choi51, line 106: @mercmobily  The Code that you suggested is totally works... But, frankly i don't know how it works. event that code pass the test code

i'll dig in the code base. 
please let me know if i need to do some fixing the code

lib/client.js
myndzi: `queryContext` is only [consumed by wrapIdentifier](https://knexjs.org/guide/#wrapidentifier), if the user has supplied such a function. So it doesn't need to be passed to the dialect clients

test/unit/query/builder.js
```diff
@@ -10863,6 +10863,25 @@ describe('QueryBuilder', () => {
...
+        })
+        .toQuery();
+
+      expect(query).to.equal(
+        `delete from "wp_posts" as "p" using "wp_postmeta" where "wp_postmeta"."id" = '123'`
+      );
+
+      expect(query).to.not.equal(
+        `delete from "wp_posts" as "p" using "wp_postmeta" where "wp_postmeta"."id" = "123"`
+      );
```
myndzi, line 10964:
```suggestion
```

This assertion is unnecessary :) It can never fail because the previous one must have succeeded to get here

lib/dialects/postgres/query/pg-querycompiler.js
myndzi: I don't blame you. The code is very abstract.

The QueryCompiler instances have methods for building fragments of the query; the grouped objects are descriptions of "which method to call and which parameters" that are created when the QueryBuilder methods are called.

So, something like this:
```js
knex.join('table', (join) => {
  join.onVal('table', 123);
});
```
creates an object like:
```js
{
  group: 'join',
  type: 'onVal',
  value: 123,
  ...etc
}
```
...and pushes it into an array of "stuff"

When compiling queries, this array is grouped by the `group` property, and farmed out to more methods. So `this.grouped.join` contains "all the objects which had a `group` property with the value `"join"`". Of these, _some_ are joins on literal values, some are joins on column references, and so on -- and which kind of join it is depends on the `type` property.

The code here, then, is saying "depending on which type of join this is, call the appropriate implementation to turn it into a string, manage the bindings, and so on". It implicitly depends on having a method that exists for every value that `type` could contain, and these individual methods are responsible for determining how to format and/or wrap the pieces of the join condition.

Since these things already exist and are used, it's reasonably safe here to use them again - and is certainly more correct.

myndzi (changes requested):
Hello there, and sorry for the delay. I see that you responded to a comment on a previous PR but I'm going through all the PR backlog in a somewhat arbitrary order (from previous triaging) and wasn't aware that you had picked this back up!

This looks good to go, but for cleanliness I'd prefer if you remove the unused queryContext parameter and the unnecessary assertion.

I'm waffling on whether we want to add more comprehensive tests for delete queries, but this definitely fixes a real problem -- and if any of the other join types are broken, they'd have been broken before this change too. So I'll just leave an issue to come back and do that later rather than bloat out this PR's scope.

mercmobily:
@YongSeok-Choi51 You still around? Please let us know if you are able to push this over the finishing line -- thank you!

PR #6310, opened 2025-11-14
Fix `test:cli` tests

myndzi:
- Run all CLI tests with `npm run test:cli`
- Fix tests that were failing locally due to ANSI color mismatch

This PR also refactors things a little to make targeting the CLI tests more straightforward:
- Extracts the mocha-based CLI tests to a separate file that can be readily run on its own
- Adds the mocha-based CLI tests to `npm run test:cli` (previously it was only running the jake tests; mocha-based CLI tests were included in the all-test-suite but not the `test:cli` command)

Note: the Jake tests don't actually do anything until https://github.com/knex/knex/pull/6264 is also merged.


mercmobily:

PR #6311, opened 2025-11-14
Ensure CRDB has the expected database before running integration tests

myndzi:
On a fresh checkout of Knex, using `npm db:start` and then `npm test` (or `npm test:db`, `npm test:cockroachdb`), all the CockroachDB integration tests fail due to a missing `test` database. This PR adds code to ensure this database exists and is in a consistent state when running the integration tests.

The code relies on the `integration2` helpers to actually get connected to the database and execute the necessary "create database" command, but is called by the mocha configs before actually executing the integration suites.

### DIFF
--- a/test/tape/index.js
+++ b/test/tape/index.js
@@ -15,20 +15,22 @@ require('./pool');
 require('./knex');
 require('./invalid-db-setup')(knexfile);
 
-Object.keys(knexfile).forEach(async function (key) {
-  const knex = makeKnex(knexfile[key]);
+(async () => {
+  for (const key of Object.keys(knexfile)) {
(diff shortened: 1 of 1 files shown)


myndzi:
I need a better solution than using describe/it in this hacky way - gonna revise this one. It causes failures when using `it.only`, for example

(It worked in this way to ensure ordering, but wasn't working as a before block)

myndzi:
This seems to work, and CI passed. Un-drafting

myndzi:
(this doesn't apply to the tape tests, refactoring to get those too)

mercmobily:

PR #6313, opened 2025-11-15
Fix various failing tests

myndzi:
This PR (along with https://github.com/knex/knex/pull/6264, https://github.com/knex/knex/pull/6310, and https://github.com/knex/knex/pull/6311) completes my goal of being able to run all tests locally to completion with a successful outcome.

Detailed descriptions are written in the commit messages for each file, but here's the high-level overview:

- Fixed a failing assertion against an Oracle error message that appears to have changed
- Dropped some additional testing tables before running the tests that depend on them not existing
- Fixed an orchestration logic error in the timeout tests against mysql
- Fixed various `tape` tests that were just broken
- Fixed the `tape` knex instantiation tests to account for `password` no longer being enumerable after  https://github.com/knex/knex/pull/5559

### Future work

I discovered various other problems, some of which may indicate actual bugs, while doing this. Here are some notes:

- I reduced the scope of my goal from "`npm test` always succeeds" to "`npm run db:stop && npm run db:start && npm run test` always succeeds" -- there are many tests that cannot succeed after they have already run, and at this moment I only wanted to have some way to ensure that all tests that should pass, do pass. I'm leaving it to a later effort to improve the developer experience to be less hasslesome.

- In various cases, there were failing tests that seem to _just not be run by CI_. In others, there are tests that pass in CI but fail locally (e.g. the mysql timeout tests). It's likely worth auditing the CI config to ensure that we're not missing swaths of code that should be tested but aren't failing CI -- and/or ensuring that the CI test environment and steps are reproducible with a single command locally.

- I noticed that in Oracle, sequences are being left behind after the test runs. By the code itself, it looks like `knex.schema.dropTable` _intends_ to remove these. This may be a bug in the code itself.
  - Newer versions of Oracle no longer require the trigger workaround that knex uses to implement `increments` columns so, depending on whatever the policy is for "how old of a database" we want to support, it might be preferable to just redo the `increments` implementation to not rely on trigger/sequence behavior at all

- Some tests are not self-contained; they will fail unless other tests have run prior to them. This is not ideal. There will be an issue to track revamping the tests systematically; this would be addressed there.

r? @mercmobily

### DIFF
--- a/test/tape/invalid-db-setup.js
+++ b/test/tape/invalid-db-setup.js
@@ -12,6 +12,7 @@ module.exports = (knexfile) => {
     // TODO: FIX ORACLE AND MSSQL TO WORK THE SAME WAY WITH OTHER DIALECTS IF POSSIBLE
     if (
       dialect !== 'sqlite3' &&
+      dialect !== 'better-sqlite3' &&
       dialect !== 'oracledb' &&
       dialect !== 'mssql' &&
       dialect !== 'cockroachdb'


mercmobily:

Issue #6074, opened 2024-05-15
limit() is not working with delete()


kilobyte2007:
# Environment

Knex version: 3.1.0
Database + version: MySQL 8.3.0
OS: Linux/MacOS

It seems like knex still doesn't support limit when using `delete`.

`.where(...).limit(100).delete()`

I see there's an older issue #96 about this which is closed as fixed but it still doesn't work.
One of the reasons stated in the original issue is the consistency across platforms. But from what I see in the docs - there's lots of things that are platform specific so I think it would make sense to implement this.

jokeyrhyme:
i just encountered this with PostgreSQL, where DELETE does not support LIMIT at all, but the `knex` query builder happily accepts `.delete(...).limit(...)` and simply removes the LIMIT from the query before execution

very surprising! :)

mercmobily:
I will get to this today or tomorrow

mercmobily:
As a favour, would you be able to do a search in the issue list (I know, it's a little big...) and see if there are duplicate issues saying the same thing?

ZeRego:
@mercmobily no issue with the keyword `limit` seems to match this scenario.

jokeyrhyme:
I have 2 ideas for how to improve this, but they are probably both breaking changes

- modify the TypeScript types so that it's a compiler error to try to use .limit() on a DELETE query builder (but this wouldn't be appropriate if other DBs do support this)

- throw an error if the DB is Postgres and there's an attempt to produce a `DELETE ... LIMIT ...` query

in my use case, I was attempting to delete a test batch of 8K rows, and knex silently discarded the LIMIT to produce a valid DELETE query, result in a query that deleted 45M rows instead (causing a production outage due to a combination of other queries running at the same time)

not blaming knex for the outage: it was my code and i should have understood and tested the output of the query builder and it's ramifications 

but it was indeed surprising :)

mercmobily:
> in my use case, I was attempting to delete a test batch of 8K rows, and knex silently discarded the LIMIT to produce a valid DELETE query, result in a query that deleted 45M rows instead (causing a production outage due to a combination of other queries running at the same time)

Oh my God... I can't even imagine.
In any case, I believe dropping the limit is potentially deadly (you can confirm that...) and it will need to be addressed; I am thinking of throwing, frankly.  It was my thought before, and it's even stronger now!

myndzi:
[...]

PR #6314, opened 2025-11-15
Prevent unexpected combinations of statements and clauses groups from executing

### LINKED ISSUE #4070: knex('table').having({ field: 'value' }).del() silently deletes ALL rows!
# Environment

Knex version: 0.21.5
Database + version: postgres 13.0
OS: alpine

# Bug

Explain what kind of behaviour you are getting and how you think it should do:

running
``` javascript
knex('table').having('some_relation_count', '>', 0).del()
```
silently removes "having" and results in
``` sql
delete from "table"
```
I executed a similar select query to check the effected rows and then executed the delete query.
fortunately I had backups!

I think knex could throw an error or let the database throw a syntax error by not removing the "having" part
Thank you for this great project

myndzi:
Fixes #4070, #6074

Clauses such as "HAVING" and "WHERE" have no effect on "DELETE" or "TRUNCATE" statements. Due to the architecture of knex, dialects override various base methods for building specific statements. The methods implementing these statements don't explicitly check for the presence of clauses that a user might expect to have an effect, but there isn't a good place to ensure that such queries fail systematically.

This commit introduces a "_preValidate" method on the root QueryCompiler class and a set of disallowed verb/clause combinations that will throw an error before doing any SQL-building work. This should make it consistent to prevent anything like `knex.having(...).del()` from unintentionally deleting all rows.

The current configuration disallows "HAVING" from "DELETE" statements, and both "WHERE" and "HAVING" from "TRUNCATE" statements.

Update:
Now also rejects "LIMIT" from "DELETE" and "TRUNCATE"

DRI:@myndzi

### DIFF
--- a/docs/src/guide/query-builder.md
+++ b/docs/src/guide/query-builder.md
@@ -479,6 +479,17 @@ knex.select('*').from('users').offset(10, { skipBinding: true }).toSQL().sql;
 
 ### limit
 
+**Important:** Knex may throw errors during SQL compilation when the query is unsound. This prevents unexpected data loss or unexpected behavior. "Limit" clauses may throw when:
+
+- A "limit" clause is present on a "delete" or "truncate" statement
+
+Examples of queries that would throw:
+
+```js
+knex('accounts').limit(10).del().toSQL();
+knex('logs').limit(10).truncate().toSQL();
+```
+
 **.limit(value, options={skipBinding: boolean})**
 
 Adds a limit clause to the query. An optional skipBinding parameter may be specified to avoid adding limit as a prepared value (some databases don't allow prepared values for limit).
@@ -1629,10 +1640,16 @@ declare module 'knex' {
 
 Several methods exist to assist in dynamic where clauses. In many places functions may be used in place of values, constructing subqueries. In most places existing knex queries may be used to compose sub-queries, etc. Take a look at a few of the examples for each method for instruction on use:
 
-**Important:** Supplying knex with an `undefined` value to any of the `where` functions will cause knex to throw an error during sql compilation. This is both for yours and our sake. Knex cannot know what to do with undefined values in a where clause, and generally it would be a programmatic error to supply one to begin with. The error will throw a message containing the type of query and the compiled query-string. Example:
+**Important:** Knex may throw errors during SQL compilation when the query is unsound. This prevents unexpected data loss or unexpected behavior. "Where" clauses may throw when:
+
+- Any "undefined" value is passed to a where clause
+- A "where" clause is present on a "truncate" statement
+
+Examples of queries that would throw:
 
 ```js
 knex('accounts').where('login', undefined).select().toSQL();
+knex('logs').where('server', 'dev').truncate().toSQL();
 ```
 
 ### where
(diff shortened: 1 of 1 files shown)


myndzi:
(This needs a re-run of the one CI task - it failed on a network error updating coveralls)

lib/query/querycompiler.js
```diff
@@ -1,6 +1,7 @@
 // Query Compiler
 // -------
 const helpers = require('../util/helpers');
+const HOP = require('../util/security').hasOwnProperty;
```
kibertoad: It would be preferable to avoid non-obvious abbreviations

lib/query/querycompiler.js
```diff
@@ -66,8 +77,44 @@ class QueryCompiler {
...
+    const verb = HOP.call(methodAliases, method)
+      ? methodAliases[method]
+      : method;
+    if (!HOP.call(invalidClauses, verb)) return;
+
+    // For certain verbs, certain clauses just don't exist / aren't supported. The list
+    // here is intentionally not complete; it's just checking the things that allow users
+    // to make dangerous errors.
+    const invalid = invalidClauses[verb];
+    for (let i = 0; i < invalid.length; i++) {
```
kibertoad, line 96: why not use for of?

lib/query/querycompiler.js
```diff
@@ -66,8 +77,44 @@ class QueryCompiler {
...
+      const clause = invalid[i];
+
+      const hasNonEmptyGrouped =
+        HOP.call(this.grouped, clause) && this.grouped[clause].length > 0;
+      const hasNonEmptySingle =
+        HOP.call(this.single, clause) && this.single[clause] != null;
+
+      if (hasNonEmptyGrouped || hasNonEmptySingle) {
+        throw new Error(
+          `Refused to compile query: \`${clause}\` has no effect on a \`${verb}\` statement`
```
kibertoad: Wording is a bit confusing, "Refused" can mean many things. I'd suggest making it more explicit by saying that query compilation was aborted because part of it would have had no effect

lib/util/security.js
```diff
@@ -21,6 +21,9 @@ function setHiddenProperty(target, source, propertyName = 'password') {
   });
 }
 
+const hasOwnProperty = Object.prototype.hasOwnProperty;
```
kibertoad: I'm not sure it's a good idea to do this reexport. the fact that it needs to be invoked via ".call" is not obvious.

helper function wpuld be clearer:

function hasOwn(obj, key) {
  return Object.prototype.hasOwnProperty.call(obj, key);
}

lib/query/querycompiler.js
myndzi: Combination of:
- presuming it's a hot path, so avoiding extra garbage and/or cpu overhead generated by the iterator protocol
- this loop pattern is generally so recognizable that I don't expect it to be a legibility disadvantage

lib/util/security.js
myndzi: I just find myself typing it so often that I wanted to make it more easy to access. This is fine too; the main thing is that I want to capture the value of `hasOwnProperty` at load time, which slightly reduces the risk that it gets monkey patched by something (or potentially exploited). I can do that and still export a function.

Once we go beyond node 16 as minimum version we can just use Object.hasOwn anyway

lib/query/querycompiler.js
myndzi: How about "Aborted query compilation: ${clause} has no effect on a ${verb} statement"?

I want to include both the thing they tried to add (e.g. `having`) and the statement they tried to add it to (e.g. `delete`) so that it's clear to the user what the problem is.  I would like to avoid general language like "part of it would have had no effect" (what part? what query?), and also make it clear that it was an intentional choice by the library (I "refused" to do what you asked, and here's why)

lib/query/querycompiler.js
myndzi: Let me know if you'd prefer I changed this

myndzi:
(rebased onto master)

PR #6315, opened 2025-11-16
Refactor SQLite transactions to allow setting the foreign_keys pragma for a transaction

### LINKED ISSUE #6213: Dropping a column in an Sqlite database via a migration triggers foreign key checks and causes data loss
# Bug

Running a migration that drops a column from a table triggers an "on delete cascade" foreign key in a different table, causing all rows to be dropped.

A migration like this
```js
exports.up = function (knex) {
    return knex.schema.alterTable('test_customers', function (table) {
        table.dropColumn('gender');
    });
};
```
is executed by creating a temp table with all columns from `test_customers` except `gender`, copying over data from `test_customers`, dropping `test_customers` and finally renaming the temp table to `test_customers`. Previously, foreign key checks are disabled. For some reason, the key is checked, and dropping the table causes a connected table to drop all rows.

### Environment
 
* Nodejs: 20.10.0
* Knex.js: 3.1.0
* better-sqlite3: 11.10.0
* OS: MacOS 15.5, 

### Test case 

https://github.com/nenad-mitic-bg/knex-test

myndzi:
The SQLite schema builder uses a pretty standard workaround to drop columns:
- Create a copy of the target table without the column in question
- Copy over the data
- Drop the old table and rename the new one

In order to prevent rows in other tables with ON CASCADE DELETE from being dropped when the old table gets dropped (sqlite executes an implicit DELETE on all rows when dropping a table), the DDL helper attempted to disable foreign key enforcement before performing this dance, and enable it after.

The DDL helper performs this work in a transaction, which is reasonable, except that "PRAGMA foreign_keys" has no effect in a transaction: https://sqlite.org/pragma.html#pragma_foreign_keys

In order to ensure that we can "have a transaction with foreign key enforcement disabled", I extended the Transaction_Sqlite class to accept a new config option, `enforceForeignCheck`. When set to a boolean, the transaction that the callback receives will have its `foreign_keys` pragma set to ON (for true) or OFF (for false). If not defined, no change is made.

In order to prevent future error, I also introduced a private "_strict()" method on the Client_SQLite3 class, which returns a version of the client which will refuse to create a transaction unless the `enforceForeignCheck` property is specified. This should guard against people making changes or adding new code but not knowing they need to use this mechanism. The `.ddl()` method always uses a "strict-mode" client, and the two uses of transactions in the SQLite3_DDL class were updated to supply the required option.

I had to refactor the base Transaction class a little so that the Transaction_Sqlite class could override and wrap the business logic with setup and cleanup code.

All changed code has 100% coverage.

Fixes #6213

Updated:
I added a new option to `config.transaction` on a migration file: `false_if_sqlite`. This allows users to disable transactions on a per-migration basis when the driver being used is `sqlite3` or `better-sqlite3`. Added the new option to the documentation and added tests to support it.

---

### Note for users

When in strict mode, Transaction_Sqlite will also throw an error if it attempts to create a nested transaction _and also_ needs to change the foreign_keys pragma. This may cause knex to start failing for some people who have been using it in a way that might have caused them data loss.

Users who start hitting this error are advised to update their migration to disable migration-runner transactions and instead use an explicit transaction like this:
```js
exports.up = function (knex) {
  return knex.transaction(function (trx) {
    return knex.schema.alterTable('users', function (table) {
      table.string('website');
    });
  });
};

exports.down = function (knex) {
  return knex.transaction(
    function (trx) {
      return knex.schema.alterTable('users', function (table) {
        table.dropColumn('website');
      });
    },
    { enforceForeignCheck: false }
  );
};
exports.config = { transaction: 'false_if_sqlite' };
```
DRI:@myndzi

### DIFF
--- a/lib/migrations/migrate/Migrator.js
+++ b/lib/migrations/migrate/Migrator.js
@@ -13,7 +13,7 @@ const { getSchemaBuilder } = require('./table-creator');
 const migrationListResolver = require('./migration-list-resolver');
 const MigrationGenerator = require('./MigrationGenerator');
 const { getMergedConfig } = require('./migrator-configuration-merger');
-const { isBoolean, isString, isFunction } = require('../../util/is');
+const { isBoolean, isFunction } = require('../../util/is');
 
 class LockError extends Error {
   constructor(msg) {
--- a/test/unit/migrations/migrate/migrator-use-transaction.js
+++ b/test/unit/migrations/migrate/migrator-use-transaction.js
@@ -1,8 +1,6 @@
 const { expect } = require('chai');
-const { DRIVER_NAMES } = require('../../../util/constants');
 const { isBoolean } = require('../../../../lib/util/is');
 const { Migrator } = require('../../../../lib/migrations/migrate/Migrator');
-const { getDriverName } = require('../../../util/db-helpers');
 
 describe('Migrator', () => {
   describe('_useTransaction', () => {
@@ -11,8 +9,7 @@ describe('Migrator', () => {
     });
 
     // eslint-disable-next-line no-undef
-    const knexSqlite = mockKnex(DRIVER_NAMES.SQLite);
-    const knexNotSqlite = mockKnex('no such thing');
+    const fakeKnex = mockKnex('no such thing');
 
     const mockMigrationContent = (transaction) => {
       const content = {
@@ -50,7 +47,7 @@ describe('Migrator', () => {
       it(`allTransactionsDisabled=${allTransactionsDisabled} config.transactions=${local} _useTransaction() = ${expected}`, () => {
         const migrationContent = mockMigrationContent(local);
         const mocked = {
-          knex: knexNotSqlite,
+          knex: fakeKnex,
         };
         const result = Migrator.prototype._useTransaction.call(
           mocked,


myndzi:
(some of the existing tests are failing, but I'm done for the day. I'll come back and run those down too later)

myndzi:
Existing tests all pass now, but I (or someone) needs to review whether it matters that the sqlite migration integration tests now run with `disableTransactions: true`. The issue that this PR is intended to fix suggests that some of the previous tests might have been unsound. It seems unlikely, but running the tests outside of a migration might also alter their semantics.

There might also need to be some new CLI tests to verify that `--disable-transactions` does indeed have the expected effect, since it's new.

test/integration2/util/knex-instance-provider.js
```diff
@@ -30,6 +33,36 @@ function getAllDbs() {
...
+ * should be executed.
+ * @param {(knex: Knex, name: DriverName) => void} cb
+ * @example
+ * withDbs([Db.SQLite], (knex) => {
+ *   describe('SQLite-specific tests', () => {
+ *     // ...
+ *   })
+ * })
+ */
+function withDbs(dbNames, cb) {
```
myndzi, line 51: This little helper is intended to replace the `getAllDbs().forEach(db => getKnexForDb(db))` pattern that the integration2 tests are using. It gives a `knex` instance directly to the callback and supports filtering, so that tests are only defined for database drivers that they are useful for, instead of conditionally skipping them.

test/integration2/util/knex-instance-provider.js
```diff
@@ -208,8 +249,44 @@ function getKnexForDb(db, configOverrides = {}) {
...
+  config.pool = {
+    ...config.pool,
+    afterCreate: function (connection, callback) {
+      connection.prepare(sql).run();
+      callback(null, connection);
+    },
+  };
+  return knex(config);
+}
+
```
myndzi, line 280: These functions parameterize the sqlite knex instance so that connections are initialized with foreign key checking either on or off. This is in support of the new tests that are added; this could be integrated into the other bits, but I wanted to keep this PR as self-contained as possible and not also involve refactoring everything.

test/jake/jakelib/migrate-test.js
```diff
@@ -270,6 +270,7 @@ test('migrate:rollback --all rolls back all completed migrations', (temp) => {
         `node ${KNEX} migrate:rollback --all \
       --client=sqlite3 \
       --connection=${temp}/db \
+      --disable-transactions \
```
myndzi, line 273: I didn't comprehensively add `--disable-transactions` to every instance of `knex migrate` in these tests, only the ones that needed it.

test/integration2/util/knex-instance-provider.js
```diff
@@ -153,16 +186,24 @@ const testConfigs = {
     client: 'sqlite3',
     connection: testConfig.sqlite3 || ':memory:',
     pool: poolSqlite,
-    migrations,
+    migrations: {
+      ...migrations,
+      disableTransactions: true,
```
myndzi: For the integration tests, I added `disableTransactions` here, which ensures that the tests that ran afoul of the new soundness check run successfully again, but there is not a simple way to target only the ones that need it, so all of the integration tests exercising sqlite migrations are now running with transactions disabled

mercmobily:
This is ready to be reviewed/merged now, right?

mercmobily:
This is NOT my review, but I just noticed that there is some light pestering from GitHub:

image

myndzi:
> This is NOT my review, but I just noticed that there is some light pestering from GitHub:

This is harmless; it's a stale comment from when this branch was stacked on top of https://github.com/knex/knex/pull/6264.

Also, that PR was just replacing an exec with an exec; so it didn't alter the exposure, just got rid of a dependency.

mercmobily:
@kibertoad This is officially mine for review, but I am falling behind and was wondering if you had time to give us our 2c about it.

docs/src/guide/migrations.md
```diff
@@ -306,7 +306,52 @@ exports.down = function (knex) {
...
 ```
 
-The same config property can be used for enabling transaction per-migration in case the common configuration has `disableTransactions: true`.
+The same config property can be used for enabling transaction per-migration in case the common configuration has `disableTransactions: false`.
+
+#### Sqlite3-specific concerns
+
+In some cases, such as writing migrations that use `.dropColumn()`, migration-level transactions can result in data loss (see [this PR](https://github.com/knex/knex/pull/6315) for details). Knex now throws an error in these cases.
+
+Rather than disable transactions for all migrations, or disable them for a single migration file on all drivers, you may specify the string `"false_if_sqlite"` as the value of `config.transaction`. Transactions will be disabled if executing via the `sqlite3` or `better-sqlite3` drivers, and use the default global behavior on other drivers:
```
kibertoad: @mercmobily personally I'm not a fan of such deeply opinionated, internally conditional flag. if we want conditionality, isn't it more easily expressed in JS terms?

```ts
{ transaction: !(knex.driver.dialect === 'sqlite') // (pseudocode, but you get the idea)
```

myndzi: It is certainly more _flexible_ to express this as a JS expression.

However, knowing _whether_ and _how_ to write that expression is the tricky part here. Why should the user have to know or care about the existence of `knex.driver.dialect`? Or the difference between `sqlite3` and `better-sqlite3`? What happens if they wrote their migration files when they were using `sqlite3`, everything worked, and then they change over to `better-sqlite3`?

My goal here was to make the configuration value stand out as unusual (it's neither `true` nor `false`), do exactly and only the correct thing, and clearly state what the expectation is ("If I'm using sqlite, I don't expect these migrations to run in a transaction"). They state something declaratively, and it's the library's responsibility to do the right thing.

So, I understand what you're describing but I don't prefer it as a solution to the specific problem I'm trying to address here.

kibertoad: is there a way to avoid having this a parameter altogether, and automagically determine we need to skip the transaction?

myndzi: It's been a while since I wrote this, but I believe the default is to execute in a transaction.

By the time any code that would care whether we're in a transaction or not runs, it's too late - thus the need to explicitly configure it on the exported property. It'd be a rather larger refactor to allow the code that needs to "not have a transaction" to be able to control what it receives.

Prior to this, I had a version of this PR where the advice I gave was to set `transaction: false` to disable transactions for migrations; the `false_if_sqlite` came out of feeling like that was too broad a brush and wanting a way to be more precise.

That larger refactor is likely something I am interested in doing later anyway; this whole thing is kind of a half measure to try and find the sweet spot of "touching as little as possible" while getting the job done safely and correctly. If this PR is blocked on performing that refactor, it's probably not going to merge for a while.

Since this is a fix to a data loss problem though, I prioritized finding the best compromise that I could.

kibertoad: I don't like the overly specific coupling to sqlite. if the idea is to have opinionated decisioning on knex level, maybe we can e. g. call it "smart" or "auto", leaving us a door to add extra edge case handlings there?

myndzi: If it's just the word `sqlite` that is bothering you, I'm cool with changing the naming, but I'll lay out one more motivating factor for you:

The specific and exact problem here is that `PRAGMA foreign_keys` in sqlite3 has no effect when it is run within a transaction. As a result, the only way for us to correctly run a migration that needs to alter the foreign_keys pragma (which happens implicitly via the sqlite client's implementation of things like dropColumn) is to ensure that it is altered outside of a transaction, before the transaction is opened. The problem that this setting _exists to solve_ in an sqlite problem, so I don't agree that abstracting away the clarity is gaining us anything.

By naming it this way, a person looking at the file knows everything they need to know to reason about it directly. They know what they _expect to happen_: "this code will not be run in a transaction if executed against an sqlite database". We can't convey that semantic intent with "auto" or "smart".

I personally value "not being surprised", especially when it comes to things like database migrations, which is why I feel reasonably strongly here. However, "being surprised" by losing data is certainly much worse than "being surprised" by my migration failing with an error because I didn't know to set a magic property.

If this is not convincing to you, just tell me what you want me to do to pass review and I'll implement whatever you want :)

kibertoad: which use-case are we solving for by not just setting value to "false"? users who run same code on multiple databases?

are there no other databases that would benefit from disabling transactions in this context? is there a way to frame this condition by scenario/db behaviour and not by db name?

myndzi: You've got it. I was worried that "false" was too coarse a setting - in general, I expect that we _want_ transactions where they can be used. So, by setting it to false, we don't get them in scenarios where they would be beneficial (any database but sqlite)

I am unable to think of any other SQL databases for which this exact combination of problems exists (support for DDL transactions, limited support for alter table requiring a table-copy-over workaround, inability to enable/disable foreign key enforcement inside of the transaction), but I can research it and follow up.

It should be answerable from knex's code; I can just look into whether any other clients are implementing that copyover behavior and dig deeper.

I was trying to think of a behavior but I'm not sure how to describe it. "if_supports_changing_foreign_key_enforcement_in_transaction" would be accurate, I guess?

myndzi: - CockroachDB and PostgreSQL support transactions on DDL statements, and don't support disabling of the constraints. The way to make "dirty" migration changes in postgres is to use something like deferred constraint enforcement to validate at commit time
- Redshift does not enforce foreign keys at all
- Sqlite3 either enforces or doesn't enforce foreign keys by default (depending on if you use `sqlite3` or `better-sqlite3`), but it can be enabled / disabled with pragma. In a transaction, the pragma is silently ignored (rather than causing an error), which is the source of the problem this PR attempts to solve
- Microsoft's docs are impossible, but the best I can get in general is that altering check constraints is risky and not transaction-scoped in general, but the actual enabling/disabling of the constraint does work inside a transaction
- MySQL, MariaDB, and Oracle auto-commit transactions on a DDL statement
- MySQL and MariaDB have the ability to enable/disable foreign key checks
- Oracle and MS SQL have the ability to enable/disable specific constraints rather than "foreign key checking behavior in general"

So I guess the value of wrapping _knex_ migrations in a transaction is to support knex client code's ability to "more safely" simulate unsupported operations. I was somewhat surprised to learn that postgres seems to be the only database that fully supports rolling back DDL operations and thus performing transactions over the DDL operations themselves.

It's pretty hard for me to identify "the set of knex methods that simulate unsupported DDL operations", but I think it's fairly straightforward to answer the other question:

sqlite does indeed seem to be the only database that has this specific problem.

While researching, I recognized that whether or not "changing foreign key enforcement inside a transaction" is supported, that's not really the problem here. The problem here is that sqlite _accepts a command_ that says "enable/disable foreign keys" and _doesn't do it_. Silently ignoring the command we sent without raising an error and continuing the transaction is what causes the problem.

Bringing things back to the topic under discussion, I think I can more clearly describe the impact of transaction on migrations:

`transaction: false` means that a migration run on most clients will have potentially undersirable outcomes that would otherwise have been presented in only very specific circumstances:
- the database is not postgres (or sqlite)
- the knex client's implementation of a DDL operation relies on multiple DML statements, with no DDL statements inbetween, that benefit from atomicity

As a result, I'm much less concerned about "transaction: false" being the solution to sqlite migrations and am changing my opinion to "let's just get rid of this entirely". I had falsely presumed that since transaction: true is _the default_, it was more useful than it is. In fact, transaction: false probably more closely represents what users _should_ be expecting of migrations on anything except postgres(!)

What do you think?

kibertoad: FWIW, there are more issues caused by using transactions for DDL operations, see this for an example: https://github.com/drizzle-team/drizzle-orm/issues/860

But that said, I can see how users might find transactional migrations on postgres beneficial, so I would still prefer to keep that as an option, but I don't mind it being opt-in.

What would be your recommendation after all the research you've done?

myndzi: I have maybe a warm take that I left in the wiki as a note for future work, which I think is that we should not automatically wrap transactions, or at most only allow auto-wrapping where it's known to be safe.

My reasoning is twofold:

1. The option gives the user a false sense of security: I know that I, using knex, in the past, thought it was a good and safe thing, but the actual behavior of any migrations I wrote wouldn't have been what I expected in the case of a failure. But it's not hard for the user to create their own transaction if that's what they actually want, and it puts the responsibility in _their_ hands, which means they're thinking about what the transaction means with respect to the queries they are executing.
2. In cases where we emulate DDL functionality, it makes sense to create transactions at the level of the code implementing the emulation where it matters. For example, an emulated statement that performs multiple DML operations should be atomic where possible, and the code we write should be specific about which parts are transactional. That forces us to write code that models what the database actually does, which means _we_ don't get surprised by the behavior of the database since we only use transactions where we know them to work, and have to be considerate of the order of our statements to maximize that safety

As for this PR, I'll go ahead with removing the `false_if_sqlite` flag and change the advice to suggest using `transaction: false` for sqlite. I think it would be a surprising change to alter the default, and any changes like described above probably belong to a future major version of the migration runner

kibertoad (approved):

PR #6316, opened 2025-11-18
[preview] POC demonstrating a refactoring of integration tests

myndzi:
## The problem

The existing tests use a pattern like this:

```js
describe('Inserts', function () {
  getAllDbs().forEach((db) => {
    describe(db, () => {
      let knex;

      before(async () => {
        knex = logger(getKnexForDb(db));
        // ...
      });

      after(async () => {
        // ..
        return knex.destroy();
      });
      
      // correct, if verbose usage
      it('this test runs', () => {
        if (isCertainDatabase(knex)) {
          this.skip();
          return;
        }

        // do things with `knex`
      });

      // and sometimes
      if (isCertainDatabase(knex)) {
        it('this test will never run!', () => {
          // do things with `knex`
        });
      }
    });
  });
});
```

There are a few problems here:
- The most important one is that since the test files are huge, it's easy to forget or miss the fact that `knex` is only defined _inside_ of `it` blocks. Inside of `describe` blocks, it will be undefined, and the helper functions don't do anything. There are tests in master right now that never run because of this.
- There's also a boatload of manual "after" hooks to destroy knex instances and a bunch of tests that are defined but do nothing (because they always get skipped), making instantiation and cleanup of knex instances error-prone
- By crossing the scope boundary like this, it's easy to have really confusing and odd errors; there are so many nested blocks and test scopes that it's really hard to reason about the order and availability of things

The intent here appears to be to solve the problem of "some tests only make sense with some database clients / backends" in conjunction with "I want to be able to run integration tests only for the declared set of clients" (e.g. with the DB environment variable).

I set about finding and fixing cases of tests that should have been running but weren't, and found myself building up a new helper suite for accomplishing these goals but in a more straightforward and useful way.

### The POC

The general idea is to provide a convenient way to declaratively specify which database drivers to exercise, while allowing for narrowing / nested blocks but supporting development by ensuring nothing gets dropped on the floor. Separate from the declaration, the declarations get filtered at test-time by the runtime environment's own selections, which may or may not overlap with the test declarations.

We want to "declare tests only for clients where they work" and also "run tests only for clients that should be tested in this run". 

Each nested produces only a subset of its parents, and crashes if it's never going to test anything. So for example:

```js
dbs.each(['sqlite3', 'pg'], () => {
  describe('title', () => {
    dbs.each('pg', (knex) => {
      // this test is defined _only_ for `pg`
    });
  });
  dbs.each('mysql', () => {
    // throws an error: 'mysql' is not present in the parent selection (sqlite3, pg)
  });
});
```

In addition, there are selectors for drivers with multiple clients:

```js
dbs.each('sqlite~', ...);
```

And inversion:

```js
dbs.except('sqlite~', ...);
```

The outermost block generates a new `describe` block in mocha with a title like `Driver: <drivername>` for each selected driver, and creates an `after` block to clean up the knex instances that it creates. So:

```js
dbs.each(`mysql~`, () => {
  it('succeeds', () => { ... });
});
```

Produces a report similar to:

```
Driver: mysql
  - succeeds
Driver: mysql2
  - succeeds
```

While:

```js
describe('describe title', () => {
  dbs.each('*', () => {
    it('succeeds', () => { ... })
  });
});
```

Produces a report similar to:
```
describe title
  - Driver: mysql
    - succeeds
  - Driver: mysql2
    - succeeds
  - ....
```

The names of the selectors were chosen for useful intellisense integration, so if you start typing `mysql` you will see `mysql`, `mysql2`, and `mysql~` as autocomplete options, to help clearly distinguish the difference between specific clients and client groups.

All the client references are rigorously verified; the tests will fail if there's a typo.

As an initial trial, I migrated the file `test/integration2/migrate/migration-integration.spec.js`, which is one of the ones that was incorrectly skipping some tests. That's what is demonstrated in this PR. The diff is only legible if you [view it with "hide whitespace" enabled](https://github.com/knex/knex/pull/6316/files?diff=unified&w=1), but you can see that it wouldn't be too hard to move the other tests over.

In general, there are two goals here:
1. Make it impossible to make an error
2. Make it more convenient to define, execute, and inspect test results

Let me know your thoughts...

### DIFF
--- a/test/integration2/migrate/migration-integration.spec.js
+++ b/test/integration2/migrate/migration-integration.spec.js
@@ -7,6 +7,7 @@ const { FileTestHelper } = require('cli-testlab');
 
 const equal = require('assert').equal;
 const fs = require('fs');
+const fs$ = require('fs/promises');
 const path = require('path');
 const rimraf = require('rimraf');
 const knexLib = require('../../../knex');
@@ -15,28 +16,20 @@ const config = require('../../knexfile');
 const delay = require('../../../lib/execution/internal/delay');
 const _ = require('lodash');
 const testMemoryMigrations = require('./memory-migrations');
-const {
-  isPostgreSQL,
-  isOracle,
-  isMssql,
-  isMysql,
-  isSQLite,
-  isRedshift,
-} = require('../../util/db-helpers');
+
 const { assertNumber } = require('../../util/assertHelper');
-const { getAllDbs, getKnexForDb } = require('../util/knex-instance-provider');
+
+const { dbs } = require('../util/test-harness/instance-provider');
+const { isMysql, isRedshift, isOracle } = require('../../util/db-helpers');
+
 const {
   ensureTable,
 } = require('../../../lib/migrations/migrate/table-creator');
-
-describe('Migrations', function () {
-  getAllDbs().forEach((db) => {
-    describe(db, () => {
-      let knex;
-
+{
+  dbs.each('*', (knex) => {
+    describe('Migrations', () => {
       before(async () => {
         rimraf.sync(path.join(__dirname, './migration'));
-        knex = logger(getKnexForDb(db));
         // make sure lock was not left from previous failed test run
         await knex.schema.dropTableIfExists('knex_migrations');
         await knex.schema.dropTableIfExists('migration_test_1');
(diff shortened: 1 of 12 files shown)


myndzi:
(CI is not expected to succeed at this stage; I modified `getDriverName` to throw an error when it receives an `undefined` knex instance, so tests that weren't being defined before are now definitely failing)

myndzi:
I rewrote the whole thing to be much more well-factored and added tests for 100% coverage of the new test harness.

I stacked this PR on top of https://github.com/knex/knex/pull/6318, which enables syntax features we're allowed to use but were failing lint, such as the nullish-coalescing operator (`??`). If y'all are uncertain about 6318, I can certainly leave that out and rewrite this to use older syntax.

PR #6323, opened 2025-12-09
Fix where in query with raw column

myndzi:
Rebased #5691 and fixed the tests

Original OP:
> Fixes https://github.com/knex/knex/issues/5690
> 
> In [b97d49b](https://github.com/knex/knex/commit/b97d49b1847791b3e1c437710af514d9fbc545fd) the order in which the columns and values where build changed.
> But the order is important if the raw columns statement contained value bindings.
> Because the value bindings from column statements are now added after the values.
> 
> This PR fixes this by building the columns before the values again.

The problem was essentially that `this._columnClause()` and `this.client.values()` implicitly push onto the bindings array, so the order of operations was backwards. The small change to `querycompiler.js` reorders them so that they get pushed in the same order as their placement in the constructed string.

### DIFF
--- a/test/integration2/query/select/where.spec.js
+++ b/test/integration2/query/select/where.spec.js
@@ -313,7 +313,9 @@ describe('Where', function () {
 
       it('handles "where in" raw cases', async () => {
         await knex('accounts')
-          .whereIn(knex.raw('?', [9]), [1, 2, 3])
+          // COALESCE() is used here simply to generate a valid left-hand-side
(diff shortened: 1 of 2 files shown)


mercmobily:
I spent ages on this, double-triple checking. @kibertoad this is an important bug fix, please merge.

mercmobily:
@kibertoad Ping in case you missed it

kibertoad (approved):

PR #6324, opened 2025-12-10
Update to non-screamy versions of dependencies that have npm audit failures

myndzi:
All the audit warnings were on dev dependencies, so I updated most of them to latest where applicable. `tap-spec` doesn't have a non-vulnerable version, so I switched to `faucet`, which doesn't look the same but still enumerates the names of each test for CI.

`node-toxiproxy-client` is only used by the stress testing infrastructure, and that code is broken anyway. I am working on an update/replacement so that they at least work, else we should ditch the dependency entirely.

I removed the `coveralls` dependency entirely, it does not appear to be used. Instead, the github actions use a coveralls github container to publish coverage stats.

r? @mercmobily

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "source-map-support": "^0.5.21",
     "sqlite3": "^5.0.11",
     "tape": "^5.6.0",
-    "tedious": "^19.1.3",
+    "tedious": "^18.2.1",
     "toxiproxy-node-client": "^4.0.0",
     "ts-node": "^10.9.1",
     "tsd": "^0.28.1",


mercmobily:
Some tests are failing but I wonder if it was a transient problem?

myndzi:
Nope, I'll have to look into the tedious thing later. It's probably an old node version thing.

package.json
```diff
@@ -107,23 +107,23 @@
     "@tsconfig/recommended": "^1.0.1",
     "@types/minimatch": "^5.1.2",
     "@types/node": "^20.4.0",
-    "better-sqlite3": "^9.1.1",
+    "better-sqlite3": "^10.1.0",
     "chai": "^4.3.6",
     "chai-as-promised": "^7.1.1",
     "chai-subset-in-order": "^3.1.0",
-    "cli-testlab": "^2.2.0",
-    "coveralls": "^3.1.1",
```
kibertoad, line 115: they are used here: https://github.com/knex/knex/blob/master/.github/workflows/coverage.yml

package.json
```diff
@@ -138,10 +138,9 @@
     "sinon-chai": "^3.7.0",
     "source-map-support": "^0.5.21",
     "sqlite3": "^5.0.11",
-    "tap-spec": "^5.0.0",
     "tape": "^5.6.0",
-    "tedious": "^16.6.1",
-    "toxiproxy-node-client": "^2.0.6",
+    "tedious": "^19.1.3",
```
kibertoad: is there a compromise version that still works on Node 16? we can go to Node 18 in the next semver major

package.json
myndzi: `npm run coveralls` doesn't actually run the `coveralls` binary, it runs the `coveralls` script:
https://github.com/knex/knex/blob/master/package.json#L19

which actually just runs `nyc` to produce lcov output consumed (presumably) by the next action:
https://github.com/knex/knex/blob/master/.github/workflows/coverage.yml#L59-L62

package.json
myndzi: I was able to find a compromise version for better-sqlite3, but had to set this down for the day. I'll see if I can find one for tedious later when I work to ensure CI passes fully.

PR #6326, opened 2025-12-13
DRI assignment logic in workflow

mercmobily:
@kibertoad This is the workflow script that will automatically bounce back a PR to the DRI once somebody commits something. I *think* this is what we want, although we might have to refine this later.

DRI:@mercmobily

### DIFF
--- a/.github/workflows/assign-dri.yml
+++ b/.github/workflows/assign-dri.yml
@@ -35,9 +35,18 @@ jobs:
               .replace(/^@+/, '');
             core.info(`Found DRI label: ${labelMatch} -> ${driUser}`);
 
+            const driUserLower = driUser.toLowerCase();
+            const assignees = (pr.assignees || [])
+              .map((assignee) => assignee.login?.toLowerCase())
+              .filter(Boolean);
+            if (assignees.includes(driUserLower)) {
+              core.info('DRI already assigned; skipping assignment.');
+              return;
+            }
+
             if (context.payload.action === 'synchronize') {
               const sender = context.payload.sender?.login || '';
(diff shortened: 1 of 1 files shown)


mercmobily:
I will need to change this, since I added the ability to set the DRI (and the "DRI is coder" flag) via labels. I am assigning this back to myself till I have actually completed it.

mercmobily:
I want to test it more on my own repo before asking for merging.

mercmobily:
@kibertoad I haven't tested this locally. Since it's only a workflow and I will likely have to play with it quite a lot, I was wondering if you were OK with me merging this (and probably merging again 3 or 4 times, till it works well)

mercmobily:
I am working on this. I should be able to have it ready for review today.

mercmobily:
@kibertoad @myndzi I tested this on my own repo. It was a little tricky because I had to make sure that I replicated the whole workflow over on my own fork, and had to make sure I tested the edge cases. 

As I used it, I realised that this is actually a good idea. Non-collaborators will make a commit after discussion, and the assignee will be once again the DRI, which is the person who is meant to check it. It handles cases where the DRI is also the author.

I mean, we will need to see how this works in real life. It might be great, or we might hate it. Let's see how it goes.

myndzi:
I suppose there's also the possibility that this will trigger when an out-of-sync PR is updated via the UI (like with a rebase)

mercmobily:
I agree that there are possible unwanted effects, but we will only really know once we start using it. We could make it depend on a comment with specific words, but then we would need external users to know how to, and a missed handover would probably be worse than an unexpected one... maybe. Maybe not.

Let's see!

PR #6327, opened 2025-12-14
Centralize query-specific response handling into Runner.query()

myndzi:
Part 1 of (2? 3?) of adding casting support to query columns in service of `bigint` support (and other datatypes). This is pre-work to centralize the query-response handling logic, in order to enable consistent treatment of column value casting.

Draft notes:

I need to understand clearly what `.upsert()` is expected to return from the MySQL client. Some tests may have been succeeding by accident since `processResponse` defaults to returning the result of `_query`, which used to be a tuple of `[rows, fields]` and now is just `rows` directly. The internal implementation details were probably leaking and this may become a breaking external api surface change as a consequence.

I _think_ `upsert()` should be like `insert()` where it returns a single-item array, since MySQL doesn't return a list of affected rows, just a summary object. We can keep it from becoming a breaking change by hacking the return value of upsert to return the same thing it would have returned before, but should definitely clean that up at next opportunity for breaking changes.

Also: how do we verify Redshift? (and any other clients that aren't run by CI?)

--

Before this change, every `Client.processRequest()` call independently called something like:

```js
if (obj.output) return obj.output.call(runner, response);
```

This was the only reason to pass `runner` into `processResponse`, but due to differing implementations of `Client._query` and `Client.processResponse`, lifting this logic out of the client required some changes to the client implementations. Most meaningfully, MySQL was returning a tuple of [rows, fields] as its response data, rather than using query.context like e.g. sqlite3.

After this change, calling `response.output` is the responsibility of the runner, and happens before calling `processResponse` on the clients.

Most clients don't account for `columnInfo` calls in their processResponse implementation, but some now have to.

I also added some JSDoc types to help sort out and understand the data flow here.

DRI:@myndzi

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -167,6 +167,15 @@ class Client extends EventEmitter {
     return sql;
   }
 
+  /**
+   * @abstract
+   * @param {import('../../query/querycompiler').ProcessResponseQueryObject} query
+   * @returns {any}
+   */
+  processResponse(query) {
+    throw new Error('not implemented');
+  }
+
   postProcessResponse(resp, queryContext) {
     if (this.config.postProcessResponse) {
       return this.config.postProcessResponse(resp, queryContext);
--- a/lib/dialects/mssql/index.js
+++ b/lib/dialects/mssql/index.js
@@ -19,6 +19,48 @@ const debug = require('debug')('knex:mssql');
 const SQL_INT4 = { MIN: -2147483648, MAX: 2147483647 };
 const SQL_BIGINT_SAFE = { MIN: -9007199254740991, MAX: 9007199254740991 };
 
+/**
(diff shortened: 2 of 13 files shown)


mercmobily:
@myndzi  You are both DRI and the assignee. I assume you're not ready to have it reviewed?

myndzi:
Correct, I'm hanging onto this until I've finished the other one I'm working on in case I learn new things; however, I made it a separate PR because we might still want to accept it even if we decide against the other

myndzi:
(and also I need to dig into this https://github.com/knex/knex/pull/6327/changes#diff-c46d41926fdd59ada0e2bb2f34472b1295493d6dad75e336385ef4d22c28c32cR196-R208)

Issue #6319, opened 2025-11-27
better-sqlite3: defaultSafeIntegers option is ignored


dong-jun-shin:
# Environment

Knex version: Latest
Database + version: better-sqlite3
OS: All

# Feature discussion / request

1. Explain what kind of behaviour you are getting and how you think it should do

When using the `better-sqlite3` dialect, the `defaultSafeIntegers` option passed in the connection configuration is ignored. This prevents users from enabling BigInt support for safe integer handling in `better-sqlite3`.

The expected behavior is that if `defaultSafeIntegers: true` is passed in the connection options, Knex should configure the `better-sqlite3` database instance to use safe integers (returning BigInts for integers).

2. Error message

No error message is thrown, but data precision loss occurs for large integers because they are returned as Javascript Numbers instead of BigInts.

Example:
- Input (BigInt): `9223372036854776603`
- Output (Number): `9223372036854776600` (Precision lost)

3. Reduced test code

```js
const Knex = require('knex');
const Database = require('better-sqlite3');

const knex = Knex({
  client: 'better-sqlite3',
  connection: {
    filename: ':memory:',
    defaultSafeIntegers: true, // This option is currently ignored
  },
});

async function main() {
  await knex.schema.createTable('test', (t) => {
    t.bigInteger('id');
  });

  const bigIntId = BigInt('9223372036854776603'); // Max BigInt
  console.log('Input:', bigIntId); // 9223372036854776603n

  await knex('test').insert({ id: bigIntId });

  const result = await knex('test').first();
  
  console.log('Type of id:', typeof result.id);
  console.log('Value of id:', result.id);
  // Actual Output: 9223372036854776600 (Number)
  // Expected Output: 9223372036854776603n (BigInt)

  if (typeof result.id === 'bigint') {
    console.log('PASS: Returned as BigInt');
  } else {
    console.log('FAIL: Returned as ' + typeof result.id);
  }

  await knex.destroy();
}

main();
```

# Relation Issues
- https://github.com/knex/knex/issues/5050

mercmobily:
Thank you for the PR. This effectively makes #5050 obsolete. I will give priority to this one since it came with a pull request. We have to merge several pull requests to fix the CI tests. I was meant to do it last week, and then work pounced on me. I will give this PR priority once we merge the CI PRs. Thank you!!!

Issue #6032, opened 2024-03-12
Stream postProcessResponse error is not catchable with .on('error')


Cerins:
# Environment

Knex version: 3.1.0
Database + version: Microsoft SQL Server 2019 + tedious 18.1.0
OS: Linux Mint 21

[@smorey2](https://github.com/smorey2)

# Bug

## Actual behaviour and the expected one

When using `.stream()` query and postProcessResponse, that throws an error, then it is impossible to catch the error with `.on('error'`. The expected behaviour is that `.on('error'` should also catch an error that happened in `postProcessResponse`. 

## Minimal code to cause the issue. 

Specific data is not needed. 

```javascript

const DATABASE = ''; // Name of the database
const USERNAME = ''; // Username
const PASSWORD = ''; // Password
const HOST = ''; // Hostname

function main() {
    const knex = require('knex')({
        client: 'mssql',
        connection: {
            database: DATABASE,
            user: USERNAME,
            password: PASSWORD,
            server: HOST,
        },
        // A showcase what happens if postProcessResponse is a sync function which throws
        postProcessResponse: (result, queryContext) => {
            throw new Error('This is a test error');
        }
    });
    // Query does not matter since the issue is with postProcessResponse
    const stream = knex.raw('SELECT 1 as result').stream();
    stream.on('data', (row) => {
        console.log('row', row);
    });
    // Should handle the error thrown in postProcessResponse
    // But it does not
    stream.on('error', (err) => {
        console.error('error', err);
    });
    stream.on('close', () => {
        console.log('closed');
        knex.destroy();
    });
}
process.on('uncaughtException', (reason) => {
    // This should not fire since the error is caught in the stream
    // But it does
    console.error('uncaughtException', reason);
    process.exit(1);
});
main();
```

If  `process.on('uncaughtException'` is removed from the minimal code example, then the node process crashes with output containing
`throw er; // Unhandled 'error' event`

mercmobily:
Hello! Thanks very much for reporting this issue with clear reproduction code, and even more thanks for submitting PR #6033 with a fix!

This is indeed a legitimate bug. When using `.stream()` with a `postProcessResponse` hook that throws an error, the error is not properly caught by the `.on('error')` handler on the stream, which causes an uncaught exception that crashes the Node process.

The problem is in lib/execution/runner.js:65-70. The Transform stream's `transform` function calls `postProcessResponse` but doesn't wrap it in a try-catch block:

```javascript
const stream = new Transform({
[...]

PR #6328, opened 2025-12-14
Support bigint mode when using `better-sqlite3`

myndzi:
Replaces https://github.com/knex/knex/pull/5482, fixes https://github.com/knex/knex/issues/6319

This change adds support for selecting how to treat integers (number or bigint) when using `better-sqlite3`. Adds code, documentation, and tests for using the knex factory's client options and query builder `.options` method to enable or disable numbers-as-bigints.

Further work incoming, but this is useful out of the box and isolated; some of the other work involves trickier refactoring.

Open question: should we use the property `defaultSafeIntegers` in the factory options? This is not a pass-through option to the driver no matter what it is named, so here I've kept it with the name `safeIntegers` in both places it's used.

Related: https://github.com/knex/knex/pull/6327

DRI:@myndzi

### DIFF
--- a/docs/src/guide/index.md
+++ b/docs/src/guide/index.md
@@ -125,9 +125,14 @@ const knex = require('knex')({
 :::
 
 ::: info Better-SQLite3
-With the Better-SQLite3 adapter, you can use `options.nativeBinding` to specify the location of the adapter's compiled C++ addon. This can be useful when your build system does a lot of transformation/relocation of files.
 
-Example use:
+For more information on connection options, see the [Better-SQLite3 documentation](https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#new-databasepath-options) on database connection options.
+
+The `better-sqlite3` adapter supports the following options:
+
+`options.nativeBinding`: specify the location of the adapter's compiled C++ addon. This can be useful when your build system does a lot of transformation/relocation of files.
+
+Example:
 
 ```js
 const knex = require('knex')({
@@ -141,7 +146,9 @@ const knex = require('knex')({
 });
 ```
 
-Additionally, you can open the database in read-only mode using `options.readonly`:
+`options.readonly`: open the database in read-only mode.
+
+Example:
 
 ```js
 const knex = require('knex')({
@@ -155,7 +162,23 @@ const knex = require('knex')({
 });
 ```
 
-For more information, see the [Better-SQLite3 documentation](https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#new-databasepath-options) on database connection options.
+`options.safeIntegers`: Configure the default type to use for integer columns. See [better-sqlite3's documentation](https://github.com/WiseLibs/better-sqlite3/blob/master/docs/integer.md#getting-bigints-from-the-database) for more information about this behavior.
+
+Example:
+
+```js
+const knex = require('knex')({
+  client: 'better-sqlite3',
+  connection: {
+    filename: '/path/to/db.sqlite3',
+    options: {
+      safeIntegers: true, // return integers as bigint
+    },
+  },
+});
+```
+
+Use [the `.options()` method](/guide/query-builder.html#options) on a query builder to override the default for a specific query.
 
 :::
 
--- a/docs/src/guide/query-builder.md
+++ b/docs/src/guide/query-builder.md
@@ -1561,6 +1561,26 @@ knex('accounts as a1')
   });
 ```
 
+::: info Better-SQLite3
+`better-sqlite3` can return integers as either plain JS `number` (the default) or as `bigint`. You can configure this per-query with `.options({safeIntegers: boolean})`.
+
+Example:
+
+```js
+// {balance: bigint}[]
+const rows = await knex('accounts')
+  .select('balance')
+  .options({ safeIntegers: true });
+
+// {balance: number}[]
+const rows = await knex('accounts')
+  .select('balance')
+  .options({ safeIntegers: false });
+```
+
+The default for all queries can be set via [global configuration options](/guide/#configuration-options)
+:::
+
 ### queryContext
 
 **.queryContext(context)**


dong-jun-shin:
@myndzi
Hi, I’ve taken a look at the work.
I had previously worked on #6320 as a PR for #6319, and it seems there may be some overlap in scope. I was wondering why a new PR was created that includes this work again.

myndzi:
> @myndzi Hi, I’ve taken a look at the work. I had previously worked on #6320 as a PR for #6319, and it seems there may be some overlap in scope. I was wondering why a new PR was created that includes this work again.

Hi there. In brief, I was working on a larger set of changes that would address this need systematically (with respect to the types returned by queries), and add safety to better-sqlite3 specifically by rejecting queries that return numbers that are not safe (in other words: changing the default to always use safeIntegers, casting the values back to number, and throwing an error if it would be unsafe to do so).

I prepared individual pieces that could be reviewed and/or merged separately, but I was not sure whether the work in this PR would interact with the other PRs I was preparing, so I had it staged separately until everything was working together. (This PR is not ready for review, it was just holding one of those pieces)

I'm leaning towards coming back to that deeper work later and focusing on getting PRs merged so that I don't wind up generating a bunch of merge conflicts, which means that I can likely close this in favor of your PR (it wouldn't wind up changing to integrate with the safety change).

If you'd like, I'm happy to work with you to get your original PR merged instead of this one?

The main requests I'd have are to support per-query configuration of this option and to utilize the existing integration testing framework instead of making ad-hoc knex instances in the tests. You can see what I mean in the code here.

dong-jun-shin:
@myndzi 
Thank you for the explanation. I’ve reviewed your comment and the PR code, and I’d be interested to hear your thoughts on the points and questions below.

1. Setting safeIntegers as the default and casting values back to number while intentionally blocking cases that would cause precision loss seems like a good approach. In this case, exposing an option to control whether bigint values are returned (e.g. options.safeIntegers: boolean) also sounds reasonable. Returning bigint when set to true, and returning number with precision checks when set to false, seems like it could address both cases where precision loss wasn’t previously an issue and cases where bigint support is needed.

2. Applying safeIntegers at the query level also makes sense. If point (1) is adopted, providing the bigint return option at the query options level would allow for more flexible usage.

3. Using the existing integration test framework, as shown in your PR, also seems beneficial. Centralized instance management and testing against the actual DB driver look like clear advantages. Since I have limited experience designing test code and would like to better understand Knex’s testing structure, I was wondering why integration tests were preferred over unit tests in this case.

If you’re able to help get this merged as PR #6032, I’d be happy to try incorporating the feedback you shared and iterating on the implementation.

myndzi:
@dong-jun-shin 

>   Setting safeIntegers as the default and casting values back to number while intentionally blocking cases that would cause precision loss seems like a good approach. In this case, exposing an option to control whether bigint values are returned (e.g. options.safeIntegers: boolean) also sounds reasonable. Returning bigint when set to true, and returning number with precision checks when set to false, seems like it could address both cases where precision loss wasn’t previously an issue and cases where bigint support is needed.

For now, I'm shelving the full "casting behavior" thing. It's mergeable with a few more tests, but I'm not happy with the compromises I had to make; there is a clear point to refactor that would clean up those compromises, but it would touch more code than I want to change until we've knocked down the PR backlog.

I definitely agree with you that the extra safety is an improvement, but I want to be clear that it's not expected to be solved by this PR.

> If you’re able to help get this merged as PR #6032, I’d be happy to try incorporating the feedback you shared and iterating on the implementation.

Let me come back to point 3 in a moment.

First, I should apologize for not having caught your PR in the first place. I'm still kinda in a mode where I expect all our PRs to be old and mostly abandoned. So I wound up creating a PR that does _exactly what I would want from yours_ already :sweat_smile:. This leaves me in an awkward position, because I'd rather merge yours and give you the contribution credit.

I'm very happy to spend some time helping you understand the decisions I made (which are by no means the final word or anything!) if you're interested. Especially if you're planning on contributing further ([we're seeking all the help we can get!](https://github.com/knex/knex/issues/6322)), there's a lot of value in getting folks up to speed on the codebase and how things work / what patterns are used and why.

With that said,

> Using the existing integration test framework, as shown in your PR, also seems beneficial. Centralized instance management and testing against the actual DB driver look like clear advantages. Since I have limited experience designing test code and would like to better understand Knex’s testing structure, I was wondering why integration tests were preferred over unit tests in this case.

When I skimmed your PR, I didn't actually notice that the tests you added were in the `unit` directory. This is because what you've created _are_ integration tests:

```js
     knexInstance = knex({
        client: 'better-sqlite3',
        useNullAsDefault: true,
        connection: {
          filename: ':memory:',
          options: {
            defaultSafeIntegers: true,
          },
        },
      });
```

This is actively "connecting" to a real sqlite3 instance using the `better-sqlite3` library and running queries through it.

Since you chose to do it this way, I believe you already understand why an integration test is correct here: the goal is to ensure that the code that was written actually has the desired effect, and the only way to do that is to actually exercise the interface between knex and the `better-sqlite3` library.

While the way that integration tests are run (and databases are selected) could use some improvement, it will be easiest to implement that improvement later, after the tests in all the pending PRs have been merged. Keeping them "working the same way as the current tests work" makes that job easier in the future.

### Next steps

- Let's get a PR merged as a stepwise improvement, simply allowing users to interact with the `safeIntegers` setting in `better-sqlite3`. The "safe bigint casting" behavior I described is out of scope for this improvement.
- It should support both a factory-level option and a query-level option
- If the user does not specify either option, it should not call `safeIntegers` at all. That is, the default behavior should be "the same behavior as before the PR", _not_ "default to safeIntegers = false" _or_ "default to safeIntegers = true" (so no `!!flag`)
- The tests, which are integration tests, should be in the `integration2` directory (by my best estimation, this is the most-current testing skeleton for integration tests)

If you update your PR to satisfy these things, it'll be good to go. Feel free to ask me questions here (or join the discord) if you need something, I'm happy to help get it there.

dong-jun-shin:
@myndzi
Thank you very much for the detailed explanation of the background and your perspective. 🙂
I’m also glad to hear that you prefer contributing to the repo through my PR.

> For now, I'm shelving the full "casting behavior" thing. It's mergeable with a few more tests, but I'm not happy with the compromises I had to make; there is a clear point to refactor that would clean up those compromises, but it would touch more code than I want to change until we've knocked down the PR backlog.

I frequently use Knex in the Node.js ecosystem, and if there are areas that need improvement, I’m more than willing to contribute as long as time allows.

> When I skimmed your PR, I didn't actually notice that the tests you added were in the `unit` directory. This is because what you've created _are_ integration tests
This is actively "connecting" to a real sqlite3 instance using the `better-sqlite3` library and running queries through it.
Since you chose to do it this way, I believe you already understand why an integration test is correct here: the goal is to ensure that the code that was written actually has the desired effect, and the only way to do that is to actually exercise the interface between knex and the `better-sqlite3` library.
While the way that integration tests are run (and databases are selected) could use some improvement, it will be easiest to implement that improvement later, after the tests in all the pending PRs have been merged. Keeping them "working the same way as the current tests work" makes that job easier in the future.

As you mentioned, the tests in my PR needed to verify interactions with the actual driver, so they should have been written as integration tests, and writing them as unit tests was not appropriate. Although it’s a bit embarrassing, I realize that my understanding of the tests I wrote was insufficient, and thanks to your detailed explanation, I now understand what I missed. Thank you. 🙏

> ### Next steps
> * Let's get a PR merged as a stepwise improvement, simply allowing users to interact with the `safeIntegers` setting in `better-sqlite3`. The "safe bigint casting" behavior I described is out of scope for this improvement.
> * It should support both a factory-level option and a query-level option
> * If the user does not specify either option, it should not call `safeIntegers` at all. That is, the default behavior should be "the same behavior as before the PR", _not_ "default to safeIntegers = false" _or_ "default to safeIntegers = true" (so no `!!flag`)
> * The tests, which are integration tests, should be in the `integration2` directory (by my best estimation, this is the most-current testing skeleton for integration tests)
> 
> If you update your PR to satisfy these things, it'll be good to go. Feel free to ask me questions here (or join the discord) if you need something, I'm happy to help get it there.

I strongly agree with the gradual improvement approach you suggested. I was planning to work on my PR (#6320) by incorporating the work from the current PR (#6328) and reflecting the next steps feedback, but it looks like it has already been merged… 😅

Should I proceed by working on this as a separate PR?

kibertoad:
@dong-jun-shin yup, iterating in a separate PR would be great!

PR #6332, opened 2025-12-20
New, vastly improved web site

mercmobily:
* The dialect selector at the top is now very important. For each query in the "QueryBuilder" and "Raw", it will now show the generated query for that dialect. It will also tell readers the errors thrown in case the query breaks, AND it will give a red mark on the menu on the left

* The red mark on the left is invaluable to know, at a glance, what a database server doesn't supports.

* Better navigation. When scrolling through the guide, a more modern indicator will tell you what you are looking at.

* Updated to the latest Vitepress, which gives up the bonus of nicer CSS and a more modern look. The homepage is slightly changed.

### DIFF
--- a/docs/.vitepress/theme/Layout.vue
+++ b/docs/.vitepress/theme/Layout.vue
@@ -60,6 +60,28 @@ const setSidebarActive = (headingId) => {
   }
 };
 
+const updateUrlHash = (headingId) => {
+  if (typeof window === 'undefined') {
+    return;
+  }
+
+  const pathname = window.location.pathname;
+  const currentHash = window.location.hash.replace(/^#/, '');
+  if (!headingId) {
+    if (!window.location.hash) {
+      return;
+    }
+    window.history.replaceState(null, '', pathname);
+    return;
+  }
+
+  if (currentHash === headingId) {
+    return;
+  }
+
+  window.history.replaceState(null, '', `${pathname}#${headingId}`);
+};
+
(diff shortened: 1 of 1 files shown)


mercmobily:
Don't review it just yet... I still need to go through it a little. But... I wanted to share it. It's awesome!

mercmobily:
@kibertoad Alright, this is a super massive improvement on the site. I spent hours and hours and hours on this. I did it because I felt I *really* needed a feel of what knex was capable of, and since knex is many dialects, I needed to know what each dialect was capable of.

So... the documentation is now updated to the latest VitePress. It looks basically identical: the only differences are in the homepage (which is closely related to the new default of VitePress) and some of the basic styling (it "looks" modern). More importantly, the dialect selector, which was basically unused in the original site, has now a HUGE role.:

* If you go to the QueryBuilder section, you will see, for EVERY example shown, what SQL gets generated depending on the dialect selected. I can't express how much I learned from this, and how much I love it. This, along with the client-side https://mobily-enterprises.github.io/knex-query-studio , just make me really *get it*. If a query fails for that dialect, the sidebar will display an exclamation mark

* MORE IMPORTANTLY, for every feature, it will show whether it's only for one/two dialects, or (if it's for 3 ore more) which dialects it's MISSING from. The output of this is awesommeeeeee.

Here is a taste of it:

image

I would love to see this reviewed/merged soon, since it improves the site massively.

PR #6339, opened 2025-12-23
  Fix VitePress build in CI (WebCrypto)

mercmobily:
Local npm run build succeeds, but CI fails with crypto.getRandomValues is not  a function during VitePress build. The CI Node environment doesn’t expose globalThis.crypto, so Vite/VitePress fails while resolving config.

I used a newer version of node as recommended by @myndzi 

This is isolated to the docs config and doesn’t touch the Knex runtime or main   build.


mercmobily:
@kibertoad @myndzi Someone please mark it as reviewed, I really want to see the new site online :D

kibertoad (approved):

PR #6344, opened 2025-12-23
Enable OIDC-based version publishing workflow

kibertoad:
This will automatically publish a new version using semver labels for deciding the version bump

ToDo - adjust PR template to include label adding step

### DIFF
--- a/lib/dialects/mssql/index.js
+++ b/lib/dialects/mssql/index.js
@@ -295,14 +295,11 @@ class Client_MSSQL extends Client {
 
       request.on('row', (row) => {
         stream.write(
-          row.reduce(
-            (prev, curr) => {
-              prev[curr.metadata.colName] = curr.value;
+          row.reduce((prev, curr) => {
+            prev[curr.metadata.colName] = curr.value;
 
-              return prev;
-            },
-            {}
-          )
+            return prev;
+          }, {})
         );
       });
       request.on('error', (err) => {


mercmobily:
I haven't had time to look at it. But... I would prefer that only PRs with a specific label actually turn into a release/NPM package.  This will give us more control.

mercmobily:
@kibertoad Sorry I pressed the button to rebase, please don't forget to pull before any changes. I will get to this as soon as I finish the site

myndzi:
I forgot to mention another mitigation step: using version tags on actions can also be compromised; we can pin the actions we run to a SHA instead, but if we don't, [things like this can happen](https://github.com/advisories/ghsa-mrrh-fwg8-r2c3)

kibertoad:
@myndzi I'd actually suggest disabling any pre/postinstall hooks. likely the only one we care about is better-sqlite native compilation, but hopefully it is possible to trigger it explicitly.
what is the benefit of lockfile, if we do not ship transitive dependencies to our end users?

I don't mind pinning github actions to a specific SHA.

OIDC has the benefit of only generating npm token at the moment of publishing, so technically there is nothing to steal. The key feasible attack vector is bypassing PR review flow and triggering publish flow on unsanctioned code, but I'm not sure how realistic that danger is.

Key benefit of this approach is how low-friction and traceable it is. You know which PR triggered release. There is no step at which you can substitute repo code with something tampered with on your own machine.

myndzi:
> Key benefit of this approach is how low-friction and traceable it is. You know which PR triggered release. There is no step at which you can substitute repo code with something tampered with on your own machine.

I very agree with this. Not that I've ever "accidentally published random files hanging out in my project directory to npm before" or anything like that 😅

I can talk details to some extent, but I'd like to refocus on the broader concern rather than digging in point-by-point. My comment (not my review) can be summed up as:

1. I _know_ that there are things I don't know (here be dragons)
2. I know also that setting this up must be done carefully, and that it must be _maintained_ carefully and aggressively over time by keeping up with security news (perhaps with support from operational monitoring and security scanning stuff)
3. While I listed _some_ of the things I _do_ know about, my concern is that I don't have enough knowledge to guide us to accomplishing (2) with confidence
4. The impact of an error here is large

With that in mind, do you believe the convenience is worth the risk? Do you have any specific things in mind in terms of how to ensure we are and _remain_ safe over time?

As an example of an alternative approach we could explore, we could automate _production of an artifact_ (so that we get the benefits of a from-scratch clean build that doesn't require credentials), and separately have _publishing_ that artifact be a manual step (perhaps with some automated cautionary steps, like validating npm audit and/or running the package through virustotal, etc)

kibertoad:
@myndzi I'm not sure I trust myself more than I trust combined security teams of GitHub and npm, who are the ones responsible for maintaining OIDC support from both sides. There is not much happening in our GitHub Action, as you can see, all the magic is happening elsewhere.

kibertoad:
That said, I don't mind introducing extra pre-publishing steps, e. g. for dependency security checks.

myndzi:
> @myndzi I'm not sure I trust myself more than I trust combined security teams of GitHub and npm, who are the ones responsible for maintaining OIDC support from both sides.

Ah, I think we're talking about slightly different things.  "OIDC support" is only part of the whole. The question I'm asking is "how do we, the people responsible for maitaining the git and npm projects, ensure that we're not exposing ourselves to a situation where malicious code can wind up being run on users' machines?". That is rather broader in scope, and includes a need to fully understand the "magic".

>  There is not much happening in our GitHub Action, as you can see, all the magic is happening elsewhere.

[This citation is a good example of what happens when you assume the magic is safe](https://yossarian.net/til/post/actions-checkout-can-leak-github-credentials/). Quoting from that post:

> GitHub knows about this and considers it intentional behavior

This is an unfortunate choice, but it's not unreasonable. Github has provided a tool that operates in a certain way; "checkout v4" is working as intended. But, for many users, it was not working as _expected_. That difference between Github's intention and the users' expectation is something that can be taken advantage of.

---

Browser session cookies are a good analogy here. As a mechanism, they're fine; it's unlikely that someone will "hack" me by brute forcing my session cookie. But if I let someone use my computer, they can copy the cookie out of devtools and have whatever privileges are granted by my session to some website. 

The same is true here, of Github actions. The OIDC token itself is fine, as is the authentication process to npm in order to publish a package. But if we let some code into the container that has the credentials, they can copy the credentials and have whatever privileges are granted by those credentials (in this case, both committing to this repository and publishing to the knex npm package).

Session cookies have nuance to using them safely (secure / httponly attributes, entropy minimums, same site, etc). The same is true here; there is nuance to "how we store and retrieve credentials, what we run in the context where the credentials are available, which actions we consume and how we configure them".

Many robust sites, including Github, will accept your session cookie to identify and authorize you for low-risk tasks (such as posting comments on an issue), but will force you to reauthorize when you attempt to perform high-risk tasks (such as adding an SSH key to your account). We have no equivalent here: a github action that automatically publishes with no user input is equivalent to shutting off 2fa on your account and eliminating the "sudo" reauthorization.

Sorry for the novel; I feel like it is pretty important to clearly explain what I'm talking about so that we're on the same page and can make some decisions.

Since I don't have a full and complete understanding of the landscape, my preference is, in order:

- Just don't have privileged secrets in github actions at all. Nobody can exploit them if they aren't there.
  - This means that we would have to continue to make merges and publishes manually, by hand
  - However, we can certainly make the _artifacts that we publish_ automated and built cleanly from scratch; in other words, we can still get the benefit you stated even if we don't automatically publish
- Consult an expert for advice, ideally someone with specific domain experience around github actions and supply chain attacks
- If none of the above are acceptable I can certainly get more knowledgeable than I am with some research, and may be able to suggest specific choices to minimize risk

myndzi:
OK, I did a bit of reading to get a better understanding of the landscape; here's what I learned about. Not all of this is fully verified, this is just my best understanding of the situation.

### Authentication and authorization for publishing to NPM with trusted publishing
- `npm publish` is the actual step that requests an OIDC token from Github. It chooses to do this via some sort of environmental detection that recognizes that it's running on Github Actions.
- The token has an expiration of O(minutes) and cannot be refreshed
- NPM revokes the token after use
- The token's identity claims that Github provides include the org/repository/workflow, and these claims are used to gate "which npm package can be published" according to configuration set up on the NPM side of things

In sum: the scope of risk for trusted publishing is:
- The attacker can only publish with an exfiltrated token if they use it before before the Github action does
- They can only publish to the NPM packages that the workflow was authorized for (irrelevant to us)
- Tokens can only be acquired within a job that has `permissions: {id-token: write}`

Caveats:
- It'd be theoretically possible for malicious code to request a token that is not the same one that `npm publish` requests and uses, so it'd at least be valid for around 5 minutes

Mitigations:
- Run `npm publish` in a separate job, reducing the scope of "things that run" that even have the capability to generate OIDC tokens

### Provenance
- `npm publish --provenance` adds information about the build contents and origin. OIDC says "who made the build", and provenance says "how the build was made" / what it produced
- Provenance does not provide for "ensuring that the contents of the artifact published to npm are what we expect them to be". The answer [to this question on a presentation about provenance](https://youtu.be/w1hgRapM-Q8?si=VOGlCu9hBRBqrjHp&t=1644) was pretty telling: "how would this have prevented some of the attacks that you demonstrated on the earlier slides?" -> "well... [it wouldn't have, that's not the problem we are solving]". Because npm is not a "curated" registry where the expectation is that the published artifacts are vetted by humans.
- The _identity_ claims of the OIDC token are cryptographically bound to the provenance attestation certificate(?)
- It's on _consumers_ of the published artifact to validate the provenance attestation (e.g. whether the hash of the package matches the attestation, whether they want to use it given the identity information, etc.). Tooling is built in for this, e.g. `npm audit signatures`, but it does not seem to happen automatically at the moment(?)
- A compromised OIDC token can be used to produce forged provenance attestation; provenance essentially validates that "the thing you downloaded was produced by this OIDC identity" and no more/less

### Workflows
While we wouldn't expect a maintainer to trigger a publish action from a third party pull request, it's not impossible to make a mistake in order-of-operations, or in the definition / configuration of the workflow itself. It'd be better to avoid building from the "pull request" trigger entirely, but it also makes our life easier _operationally_.

The example on [NPM's docs for trusted publishers](https://docs.npmjs.com/trusted-publishers#github-actions-configuration) uses `on: push: tags`, which makes a lot of sense. We handle PRs, and when we want to do a release, we just push a tag. CI builds the repo at that tag and publishes.

This eliminates the need for a complex conditional as in: https://github.com/knex/knex/pull/6344/changes#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R12

The SHA-pinning thing I mentioned is more useful for third-party workflow actions (which may be less trustworthy than Github's own actions), but is not really concerning for github-owned actions (`actions/etc`). It's likely better to use the version syntax instead so that we automatically receive any changes published by github to these actions.

### Provenance part 2
`npm publish --provenance` publishes provenance to _npm_, but we can also [create provenance attestations on _Github_](https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations). This dovetails nicely with "separation of build and publish", since it requires that we _have_ an artifact to attest.

### package.json and npm lifecycle scripts, dependencies
- We should be checking in `package-lock.json`
- We should be using `npm ci` to install dependencies from the `package-lock.json`, which ensures that we don't install newer versions of packages that have changed since we last had confidence that our dependencies were safe
- We should be using `--ignore-scripts` with `npm ci` to avoid running unexpected scripts, or switch to something like `pnpm` with an explicit allow-list of build scripts that can be run
  - This can help mitigate "npm worms" such as Shai Hulud v2
  - In general, we should _know_ and _deliberately decide to run_ all code that runs during the build step
- We should consider using `npm audit` to abort builds that contain known vulnerabilities before running the build step
  - There is a constant churn of vulnerabilities in the "tooling / building" libraries that are out there on NPM, many of which are low-risk, and this may generate some toil. There might be a case for allowing certain "vulnerable" packages to exist in cases where we don't care, such as a "regular expression denial of service" in a test framework.

### Monitoring
Dependabot can be configured to notify us about _actions_ as well as npm dependencies

### Putting it all together

The implicit assumptions here are:
- When we push a tag to be published, the codebase at that version is passing the full test suite. We could build that into the workflow as a requirement, but it would add a lot and the value is questionable
- We trust all code _executed by_ `npm build`, including dependencies we've taken on, as of the specific version we've checked into the `package-lock.json` file
  - When we update dependencies and check in a new version to `package-lock.json`, we are implicitly signing off on the trustworthiness of the new dependency version
- We do not trust any lifecycle scripts of any dependencies; if it turns out that we _must_ run some of them, we should opt in deliberately. Examples that might crop up include `esbuild` and `better-sqlite3`
- We trust anyone who can push a tag to the repo to publish an NPM package

The benefits are:
- Isolated privileges
  - Potentially-risky code runs with no access to any credentials, no matter how restricted the risk is
  - Job environments that perform privileged actions don't run any potentially-risky code
- Separation of concerns
  - The operational task of managing the codebase (reviewing, approving, merging code) can't produce published artifacts
  - The operational task of producing a published artifact is a task on its own that can't be conflated with PR management

Other notes:
- "fetch-depth: 0" actually [fetches everything](https://github.com/knex/knex/pull/6344/changes#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R23). The default is ["fetch-depth: 1"](https://github.com/actions/checkout#:~:text=%23%20Number%20of%20commits%20to%20fetch%2E%200%20indicates%20all%20history%20for%20all%20branches%20and%20tags%2E%20%23%20Default%3A%201%20fetch%2Ddepth%3A%20%27%27)
- This workflow doesn't automatically bump the version, so it doesn't include / rely on GITHUB_TOKEN. We could add another isolated job to e.g. set the `package.json` version based on the tag that triggered the push, and commit the version before running the build and publish jobs. That would also simplify and decouple the decision about "which version to publish" and separate it from the task of accepting and merging PRs. Alternately, we can just make a release PR with the release-related changes (version bump, changelog updates, etc) -- which can also be supported by tooling.

(untested, consider this pseudocode. based on [the version from npm's documentation](https://docs.npmjs.com/trusted-publishers#github-actions-configuration))

```yaml
name: Publish release 

on:
  push:
    tags:
      - "v*"

permissions:
  contents: read

jobs:
  build:
    runs-on: ubuntu-latest
    permissions:
      contents: read
    outputs:
      tarball: ${{ steps.pack.outputs.tarball }}
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-node@v6
        with:
          node-version: "24"

      - run: npm ci --ignore-scripts
      # - run: npm audit check of some kind? before "build"
      - run: npm run build

      - id: pack
        shell: bash
        run: |
          TARBALL="$(npm pack --silent)"
          echo "tarball=$TARBALL" >> "$GITHUB_OUTPUT"
          ls -la "$TARBALL"

      - uses: actions/upload-artifact@v6
        with:
          name: npm-tarball
          path: ${{ steps.pack.outputs.tarball }}
          if-no-files-found: error

  publish:
    runs-on: ubuntu-latest
    needs: build
    permissions:
      contents: read
      id-token: write        # Allow job to obtain OIDC tokens
      attestations: write    # Allow job to write Github artifact attestations
    steps:
      - uses: actions/download-artifact@v7
        with:
          name: npm-tarball
          path: .

      - uses: actions/attest-build-provenance@v3
        with:
          subject-path: ${{ needs.build.outputs.tarball }}

      - uses: actions/setup-node@v6
        with:
          node-version: "24"

      - run: npm publish --ignore-scripts --provenance "${{ needs.build.outputs.tarball }}"
```

myndzi:
Speaking of lockfiles, while I was rereading this and looking into a couple other things, I came across https://www.npmjs.com/package/lockfile-lint - which is probably worth including too. This is an attack vector that hadn't been in my mind before now :sweat_smile:

kibertoad:
@myndzi My question would be how far do we want to go. There was this meme about a person investing in super reliable top-notch encryption solution, for it only to be bypassed by a malicious actor punching him really hard so that he unlocked everything himself.

I don't see much value in solutions that are aimed at increasing protection against people with admin or write permissions. If you can get one malicious actor on the repo, you can get two, and at that point they can bypass anything and everything you could throw at them pipeline-wise. I'd say that pareto applies hard to security solutions, and you start getting diminishing returns and unjustified complexity overhead very quckly. My general recommendation would be to follow the best practices in Node.js space, find a major, reputable project that did a reasonable level of precautions, and use them as a reference. We do not need to be the most secure package in the space with breach being mathematically impossible.

If we can, I would avoid the complexity of introducing pnpm, and instead just umbrella-disable the execution of postinstall hooks (via `--ignore-scripts`), and only trigger the ones that we need manually. I also prefer not using lockfile, so that we are always running our CI on the latest in-range dependencies. What risks are you seeing there? Let us assume that `prettier` becomes compromised, and we run infected version in our CI for a couple of hours or days. What is the worst case scenario of damage that might cause? We are not bundling it with our software.

I am against aborting builds when encountering audit errors, there is A LOT of irrelevant noise in that space.

> Run npm publish in a separate job, reducing the scope of "things that run" that even have the capability to generate OIDC tokens

What kind of an access would attacker need to exploit our pipeline otherwise?

myndzi:
Rebased to pick up https://github.com/knex/knex/pull/6355 (build script no longer modifies .gitignore)

I created a bash script rather than piece together a bunch of separate Github actions and run commands, to tell a clear and coherent story about the build process. Left lots of comments.

I chose to use `npm install` instead of `npx` as we discussed because it's unclear if `npx` accepts the `--skip-scripts` argument or will always behave that way; however, `npm install`'s current behavior precludes targeted dependency installation so this required a bit of a workaround to generate a package.json that won't install "everything" when installing the build dependencies.

There's an extra step to validate that the pinned dependency versions specified by the build script (`build-for-release.sh`) are valid against the root `package.json`

@kibertoad how do you feel about this as a whole? I believe it satisfies the outcome of our discussion, though it's a bit more involved than I would like. I tried to leave clear comments explaining the choices all along the way and set things up so that it's difficult to do the wrong thing + clear what needs to happen if the workflow fails.

Edit: I also verified (with act) that the following situations skip the build and publish jobs:
- Closed PR without merge
- Merged PR without label
- Merged PR with invalid label

I verified (as far as act can take me) that the following situations succeed:
- Merged PR with correct label

I checked in the act command line and mock events used to perform these tests into the `./scripts/act-testing` directory

As a side note, we currently include `scripts/` in the NPM artifact, but shouldn't. It was originally added to deal with a `postinstall` script that no longer exists in package.json:

PR: https://github.com/knex/knex/pull/1737 to fix issue: https://github.com/knex/knex/issues/1736

That postinstall script was removed here with no apparent PR or explanation:
https://github.com/calvinmetcalf/knex/commit/a16639b156b316921696b6b811b9ebabba209dce

I've followed up with [a PR to clean that up](https://github.com/knex/knex/pull/6356).

myndzi:
Next testing steps would be to enable the commit-back behavior (but not npm publish) and be sure that it succeeds at committing the bump and formatted files to the repo + artifact attestation works.

Final step is to enable npm publish and verify that also succeeds.

mercmobily:
I only ducked out for a moment, and you two... :D
This is awesome, and we should get really ready to release. If you guys are happy with it, we should merge and probably give it a spin.

mercmobily:
Assigning to @kibertoad since I believe he's the current person with the ball in their court

package.json
```diff
@@ -115,6 +115,7 @@
     "*.js": "eslint --cache --fix"
   },
   "devDependencies": {
+    "@tsconfig/node12": "^1.0.11",
```
kibertoad, line 118: why are we using such an ancient config?

myndzi: This is just me encoding an implicit dependency explicitly. It's used here:

https://github.com/knex/knex/blob/master/tsconfig.json#L2

I wasn't inclined to change or migrate this, but I did record it as what's currently being depended on

PR #6346, opened 2025-12-24
Move __knexUid and __knexTxId away from connection instances

myndzi:
In general, mutating "someone else's" data is undesirable.

These properties appear to be used for log correlation and no actual functionality, so this change should be pretty safe.

Instead of mutating connection instances with properties, associate them indirectly via a weakmap and expose a singleton to allow retrieval. Motivated by #5288, this also fixes the "__knexTxId persists after the transaction is complete" behavior.

Open question: knex allows users to provide a connection directly, via `knex.connection(...)`. In this instance, we don't create a __knexUid for the passed-in connection; if it happens to be a connection that was acquired through knex's pool, it will have an associated id. If it does not, it won't. Is this behavior that we want?

We could assign a uid for consistency's sake if one isn't present, but I favor "the user can expect that what they get is what they give us"

cc @kibertoad - any concerns about the approach here?

### DIFF
--- a/lib/util/connection-tracker.js
+++ b/lib/util/connection-tracker.js
@@ -55,13 +55,16 @@ class ConnectionTracker {
   }
 
   withIds(resc, obj) {
-    const __knexUid = this.conns.get(resc) ?? 'unknown';
-    const res = { __knexUid };
+    const uid = this.uid(resc);
+    const trxid = this.trxid(resc);
 
-    if (this.trxs.has(resc)) {
-      res.__knexTxId = this.trxid(resc);
+    if (uid === undefined && trxid === undefined) {
+      return obj;
     }
 
+    const res = {};
+    if (uid !== undefined) res.__knexUid = uid;
+    if (trxid !== undefined) res.__knexTxId = trxid;
     return Object.assign(res, obj);
   }
 


lib/dialects/oracledb/transaction.js
```diff
@@ -71,28 +71,12 @@ module.exports = class Oracle_Transaction extends Transaction {
...
       } finally {
-        if (!configConnection) {
-          await this.client.releaseConnection(connection);
-        } else {
-          debugTx('%s: not releasing external connection', this.txid);
-        }
+        await connection.commitAsync().catch((err) => void this._rejecter(err));
       }
-    }
+    });
```
myndzi, line 80: This code was very confusing to follow and mostly duplicated the behavior in the parent class. This simplification appears to pass the tests that are relevant, but the integration tests are, in general, broken on Oracle. 

I expect to fix those tests up in a separate effort, and certainly don't mind blocking this PR on that work being done.

lib/execution/runner.js
```diff
@@ -127,9 +128,10 @@ class Runner {
   // to run in sequence, and on the same connection, especially helpful when schema building
   // and dealing with foreign key constraints, etc.
   async query(obj) {
-    const { __knexUid, __knexTxId } = this.connection;
+    const connection = this.connection;
```
myndzi, line 131: Here, `this.connection` was used in a few places, but with asynchronous "gaps" between usages. It's technically possible that its value could have changed between accesses, so I capture it once and reuse it since it makes sense to expect that we're referring to the same "connection" object throughout this method. The tests don't suggest it's breaking anything.

lib/execution/runner.js
```diff
@@ -179,13 +182,13 @@ class Runner {
...
         if (obj.cancelOnTimeout) {
-          cancelQuery = this.client.cancelQuery(this.connection);
+          cancelQuery = this.client.cancelQuery(connection);
         } else {
           // If we don't cancel the query, we need to mark the connection as disposed so that
           // it gets destroyed by the pool and is never used again. If we don't do this and
           // return the connection to the pool, it will be useless until the current operation
           // that timed out, finally finishes.
-          this.connection.__knex__disposed = error;
+          connection.__knex__disposed = error;
```
myndzi, line 191: I'd like to do a similar change for `.__knex_disposed`, but that one actually affects behavior, so I left it for a different PR at another time

lib/util/connection-tracker.js
```diff
@@ -0,0 +1,86 @@
...
+  trxid(resc) {
+    return this.trxs.get(resc)?.[0] ?? 'unknown';
+  }
+
+  isTransaction(resc) {
+    return this.trxs.has(resc);
+  }
+}
+
+const tracker = new ConnectionTracker();
```
myndzi, line 84: Initially, I had made this a subclass of Tarn's Pool and threaded it through the code. However, the existence of `knex.connection()` means we can't rely on `everywhere we have a connection, we have a pool`. I loosened things up a bit, which means we can't cleanly "remove all entries when a pool gets replaced or destroyed", but it's a weakmap so it will not retain references to connections that otherwise get GC'd and should lose any entries naturally in these cases.

lib/client.js
```diff
@@ -298,7 +297,9 @@ class Client extends EventEmitter {
       delete tarnPoolConfig.afterCreate;
     }
 
-    this.pool = new Pool(tarnPoolConfig);
+    const pool = new Pool(tarnPoolConfig);
+    tracker.addPool(pool);
```
myndzi, line 301: This essentially just binds some event handlers that add/remove entries when Tarn reports that connections are acquired or released. Tarn's docs specify that these event handlers are synchronous, and I verified experimentally that the event handlers run before e.g. `pool.acquire()` returns -- so we can be assured that the connection returned from Tarn has an id assigned to it by the time we first have a reference to it

lib/dialects/oracledb/transaction.js
```diff
@@ -71,28 +71,12 @@ module.exports = class Oracle_Transaction extends Transaction {
...
-
-    const connection =
-      configConnection || (await this.client.acquireConnection());
-    try {
-      connection.__knexTxId = this.txid;
-      connection.isTransaction = true;
-      return await cb(connection);
-    } finally {
-      debugTx('%s: releasing connection', this.txid);
-      connection.isTransaction = false;
```
myndzi, line 84: Since `.isTransaction = true` is only being set in the same place we assigned `.__knexTxId`, and we can now rely on the transaction id being correctly cleaned up, this becomes a query to the connection tracker about _whether_ a transaction id exists instead of an explicit bookkeeping task

tgriesser:
Great change. Fixing #5288 was one of the first things I was going to try and tackle as I ease back into things here - I'd also done the `delete connection.__knexTxId` in my own `patch-package` version of `knex` I've used elsewhere, but having a correct idea of whether the request is in a transaction can be really helpful from an observability standpoint. 

Completely agree we should not be mutating objects we don't own, but alas 2013 js was a different beast and there was no `WeakMap` back then and we were all still figuring this stuff out

> Open question: knex allows users to provide a connection directly, via knex.connection(...). In this instance, we don't create a __knexUid for the passed-in connection; if it happens to be a connection that was acquired through knex's pool, it will have an associated id. If it does not, it won't. Is this behavior that we want?

I'd kinda lean toward always having one for consistency, but that's just my perspective

lib/util/connection-tracker.js
```diff
@@ -0,0 +1,86 @@
...
+    pool.on('release', (resc) => {
+      if (this.trxs.has(resc)) {
+        debug(
+          'unassigning trxid=%s from uid=%s',
+          this.trxid(resc),
+          this.uid(resc)
+        );
+        this.trxs.delete(resc);
+      }
+    });
```
tgriesser, line 27: I was wondering for this whole `addPool` section if it was overkill or if we were just duplicating the job of the `WeakMap`, but I saw your other note about the guarantees of having the `id`

Is there a benefit to having this block managed by the pool - are there any cases you can think of where this wouldn't be handled by `untransact` via the transaction? Or if we're returning a connection to the pool for a connection that should still be associated with a transaction, would that be considered a bug?

myndzi: A lot of my PRs were aiming for "least code-change", and also I don't have a strong view of the complexities of the library as a whole throughout all the layers. I don't quite recall anymore, but it's likely that this is a case of "I know it needs to _happen_ here, and I can't easily reason clearly enough about everything to know that it's definitely correct to do it another way"

Issue #3636, opened 2020-01-23
Connection terminated unexpectedly (Postgresql / AWS Lambda)


jamesdixon:
# Environment

Knex version: 0.20.8
Database + version: Postgres 9.5.14
OS: AWS Lambda - Node 12.x

Recently, I've been seeing upticks in a `connection terminated unexpectedly` error. I can't find a discernible pattern at this point but unlike #3523, it doesn't seem to occur after long periods of activity. I see instances that are minutes apart in some cases. 

I added some logging to the Knex `afterCreate` method and am seeing two errors:

1. The `connection terminated unexpectedly` error
2. And `Error [ERR_STREAM_DESTROYED]: Cannot call write after a stream was destroyed` - this seems to happen on the 1st or second invocation after seeing the `connection terminated unexpectedly` error.

I've tried a number of configurations with Knex pooling:

1. `{ pool: { min: 1, max: 1 } }`
2. `{ pool: { min: 0  }`
3. `{ pool: { min: 1 } }`

I've also tried setting `{ pool: { idleTimeoutMillis: 300000 } }` as per #3523.

Additional information:
* I'm defining my Knex instance outside of my AWS handlers, so the pool should be reused per container.
* I'm using the following for my `afterCreate` function -- I'm not sure if there's a better way to get at the underlying error?

```js
function afterCreate(connection, done) {
  connection.on('error', (err) => {
    console.log('POSTGRES CONNECTION ERROR');
    console.error(err)
  });
  done()
}
```
Appreciate any help! This is driving me nuts :)

**UPDATE:** the only change that I've made is migrating the lambda function to Node 12 from Node 8 as AWS is removing support for 8. Unfortunately, it doesn't look like I can go back. My other functions in production are on Node 8 and not seeing any issues 🤔🤔🤔

**ISSUE FOUND:** for whatever reason, reverting back to Node 8.x fixes the issue. Still not certain on the underlying cause.

elhigu:
have you tried hooking some logging to:

knex.on('query', ...) , knex.on('query-success', ...), knex.on('query-error', ...), knex.on('error', ...)

Also you could wire tarn.js hooks to output what pool is doing when errors happen. Tarn instance is found from `knex.client.pool`

jamesdixon:
@elhigu I have not, but I'll give it a shot. Odd that this seems to only occur with Node 10/12. Thanks for the tips!

FreddyJD:
I'm having the same issue with MySQL. Same behavior

[...]

PR #6350, opened 2025-12-27
More complete queries for sqlite in documentation

mercmobily:


docs/scripts/format-sql-bindings.mjs
```diff
@@ -0,0 +1,108 @@
...
+function hasQuestionBindings(sql) {
+  const regex = /\\?\?\??/g;
+  let match = null;
+  while ((match = regex.exec(sql)) !== null) {
+    if (match[0] !== '\\?') {
+      return true;
+    }
+  }
+  return false;
+}
```
myndzi, line 31: This doesn't account for "the character before the ? is, itself, escaped"

If I'm reading this right, the goal is to locate instances of `?` that are not escaped?

A way to go about that is to seek an _even_ number of escape characters followed by the target character, e.g.:

```js
function hasQuestionBindings(sql) {
  const regex = /(?<!\\)(?:(\\\\)*)\?/;
  return regex.test(sql);
}
```

docs/scripts/resolve-schema-snippets.mjs
```diff
@@ -232,6 +235,31 @@ function writeOutput(filePath, data) {
 
 function createKnex(dialect, allowDb) {
   if (dialect.sqlite) {
+    // Docs builds run against the published Knex package (not repo source).
```
myndzi, line 238: I think you mentioned that generating the schema snippets is manual / checked-in, yeah? We should probably ensure that we have installed the latest knex in whatever npm script that does that; it'll be pretty easy to forget to `npm install knex@latest` and accidentally generate old docs

docs/scripts/format-sql-bindings.mjs
mercmobily: This was intentionally left simple. The idea is that this is NOT something that needs to work for knex as a whole, but just for the queries. We are already using raw() for something that was NOT intended to be used for...  If you feel strongly about it, I will make it more robust. But as I said, this needs to work for specific targeted queries, and if we start chasing formal correctness for this code we will enter a veeeeerrryyyyy ddeeeeppp rabbit hole :D

myndzi: I'll just say that the queries it needs to work against will change over time as the docs change. So, I do find correctness valuable because I don't want future-me to be writing an example of how to escape placeholders in queries and spend half an hour debugging why the docs are wrong, assuming I or someone notices -- since the content that winds up on the site isn't "seen" in the code review of the PR that creates them.

If you want the code to be simple and understandable, a straight up loop is both simpler and clearer here, while also being correct:

```js
function hasQuestionBindings(sql) {
  let escaped = false;
  for (const chr of sql) {
    if (escaped) {
      escaped = false;
    } else if (chr === '\\') {
      escaped = true;
    } else if (chr === '?') {
      return true;
    }
  }
  return false;
}
```

(note: the default iterator of string produces code points, so this is surrogate-pair safe even :) )

mercmobily: Point taken -- you are totally right.

docs/scripts/resolve-schema-snippets.mjs
mercmobily: Good point. I added a preresolve-schema-snippets step that runs npm install knex@latest before the resolver, and the resolver now logs the exact knex  version it’s using. That makes it hard to accidentally generate docs with an outdated package. If someone runs the script directly, the version is still logged so it’s visible.
(Is this the solution you were thinking about?)

mercmobily:
@myndzi  I think this is good to go.  However, I ended up in a much deeper and nastier rabbit hole... #6363 This took longer than fixing the stuff here... Along with Coverall insisting that I lost 0.0004% of coverage...

test/unit/util/format-sql-bindings.js
```diff
@@ -0,0 +1,108 @@
...
+    const { knex, raw } = makeKnexWithRawReturn(`RAW:${sql}`);
+    knex.client = {
+      _escapeBinding: sinon
+        .stub()
+        .throws(new Error('unexpected escape binding')),
+    };
+    const result = formatSqlWithBindings(sql, [1], 'sqlite3', knex);
+
+    expect(raw).to.have.been.calledOnceWithExactly(sql, [1]);
+    expect(result).to.equal(`RAW:${sql}`);
```
myndzi, line 30: I'm not sure this test is exercising the behavior it intends to?

`makeKnexWithRawReturn` mocks `knex.raw`, so `result` is just `whatever we said above`, and asserting on it is not going through any knex code. I was expecting the `?` in the sql statement to be filled with the `1` from the bindings, but we're neither doing that nor testing for it(?)

myndzi (commented):
I need to come at this fresh to understand what's going on with all the mocks and what code is actually being exercised here. Can't do it tonight though, just leaving this note

test/unit/util/format-sql-bindings.js
mercmobily: True. I've kept this unit test but clarified its intent: it's only asserting the control flow that `formatSqlWithBindings` delegates to `knex.raw(sql, bindings)` when `?` placeholders are present (and doesn't try to format them itself). That's the behaviour the docs generator depends on.

  To cover the “real formatting” concern, I added a separate real‑Knex test that  uses sqlite3 and asserts the actual `\\?` vs `?` replacement. It's currently `it.skip` with an Istanbul ignore and a link to #6363 (as you requested), so once the bug is fixed we can unskip and get the end‑to‑end check. (I am expecting surprises, but I am not gonna get into the rabbit hole of fixing #3636 to see if this one actually works, which is how I tend to function...)

myndzi: Gotcha, that makes sense then. I just didn't have enough brain left today to follow things around :sweat_smile:

PR #6352, opened 2025-12-27
Feat: improve safeIntegers support for better-sqlite3

dong-jun-shin:
## Description

This PR refines the `safeIntegers` support for the `better-sqlite3` dialect, building upon the initial work in #6320. It ensures robust handling of large integers (BigInt) and adds flexibility with query-level configuration.

## Changes

- **Improved Factory Configuration:** Refined how the global `safeIntegers` option is handled in connection settings to ensure consistent behavior.
- **Added Query-level Option:** Users can now specify `safeIntegers` per query using `.options({ safeIntegers: true })`.
- **Test Cleanup:** Removed incorrect unit tests (`test/unit/dialects/better-sqlite3.js`) and replaced them with comprehensive integration tests (`test/integration2/dialects/better-sqlite3.spec.js`) that verify both factory and query-level behaviors.

## Related Issue

- Related to #6319
- Addressing feedback and TODOs discussed in [this comment](https://github.com/knex/knex/pull/6328#issuecomment-3684590640).

## Checklist

- [x] Added integration tests
- [x] Updated documentation

### DIFF
--- a/test/integration2/dialects/better-sqlite3.spec.js
+++ b/test/integration2/dialects/better-sqlite3.spec.js
@@ -84,8 +84,7 @@ describe('better-sqlite3 safeIntegers', () => {
           table.bigInteger('value');
         });
 
-        const largeValue = 9007199254740992n; // Number.MAX_SAFE_INTEGER + 1
-        await knex('test_table').insert({ id: 1n, value: largeValue });
+        await knex('test_table').insert({ id: 1n, value: BIGINT });
 
         let queryBuilder = knex('test_table').select('*');
 


dong-jun-shin:
@kibertoad (cc. @myndzi)
Based on your explanation and my understanding, I went ahead and worked on this. I’d appreciate it if you could take a look. 🙏

myndzi (changes requested):
Thanks for following up on this! Please add one more test for factory-level true and query-level false; this logical case isn't covered....

---

Some explanatory notes:

### Why test all the cases?

My testing philosophy is to write tests that describe the behavior that _should be_. So, while we've technically covered the case that "one of the override cases works", and can assume, by the _current_ implementation, that "false also works" -- future work may change the implementation. Having thorough tests in place that assert the desired outcome in all cases supports maintenance over time and future development.

### Test ergonomics

This isn't a request for change, but you can decide if this technique is useful and adopt it if you like:

For combinatorial stuff like this, table testing is really helpful. There are four possible states for each value: unset, true, false, and set-to-undefined. Given the implementation of `_optSafeIntegers`, unset and undefined can be treated the same. So, there are 9 cases to be exercised, which can be easily expressed as data:

```js
const tests = [
  { name: 'name for this test', factory: undefined, query: undefined, expected: 'number' },
  { name: 'another name', factory: undefined, query: true, expected: 'bigint'},
  ...
];

for (const test of tests) {
  it(test.name, async () => {
    // setup table, perform query
    expect(typeof result).to.be(test.expected);
  });
 
}
```

Writing the tests in this way makes it easy to understand what cases are covered and what the expectations are (you can read it all in one place), and ensures that the code that executes the test is consistent / doesn't run afoul of copy-paste errors. It's supported natively in newer test frameworks by things like `it.each` in jest and vitest, but here since we're still on mocha, we have to kinda do it manually.

myndzi:
Ah, the documentation should also describe how to use the query-level option

dong-jun-shin:
@myndzi

> Please add one more test for factory-level true and query-level false; this logical case isn't covered....

Thank you for the kind and detailed explanation. It helped me understand things much better. The proposed improvements look good, and I was incorporate them into the next commit.

> Ah, the documentation should also describe how to use the query-level option

While I was preparing the documentation, I noticed that the query-level option is already described at L.1729 in [this commit](https://github.com/knex/knex/blob/a88c4207da7661bc31d9e6cb580734687fdff28e/docs/src/guide/query-builder.md).

It felt a bit odd that the documentation existed without a corresponding implementation. Based on the commit context, it make me wonder whether the implementation was intended to be merged later as a separate PR. (cc. @mercmobily)

mercmobily:
I have absolutely no idea how that bit of text leaked. This was a huge PR and I kept on going through all of the changes. But I really can't explain how that leaked there, honestly.  Apologies for the mistake.

myndzi:
Oh, that explains it. I was just looking at the diff, I didn't actually review the PR where that text made it in :sweat_smile:

test/integration2/dialects/better-sqlite3.spec.js
```diff
@@ -0,0 +1,105 @@
...
+      const knex = getKnexForBetterSqlite(false, knexConfig);
+
+      try {
+        await knex.schema.createTable('test_table', (table) => {
+          table.bigInteger('id').primary();
+          table.bigInteger('value');
+        });
+
+        const largeValue = 9007199254740992n; // Number.MAX_SAFE_INTEGER + 1
+        await knex('test_table').insert({ id: 1n, value: largeValue });
```
myndzi:
```suggestion
        await knex('test_table').insert({ id: 1n, value: BIGINT });
```

The point of defining the constant above was to reuse it here :)

myndzi: (to clarify, the assertions we're running are expecting the value we receive from the database to be "BIGINT". So, the value we insert into the database should also be "BIGINT". The reason I defined BIGINT and then made NUMBER from it, is because its value is intentionally overflowing. So, when we receive an overflowed value as a number from better-sqlite3, it will be equal to `Number(BIGINT)`)

dong-jun-shin:
> I have absolutely no idea how that bit of text leaked. This was a huge PR and I kept on going through all of the changes. But I really can't explain how that leaked there, honestly. Apologies for the mistake.

Thanks for the explanation. That clears up my confusion.

> Oh, that explains it. I was just looking at the diff, I didn't actually review the PR where that text made it in 😅

I was confused as well, since it wasn’t part of any PR I had reviewed or worked on.
Given that the documented behavior seems to align well with the implementation, I think it should be fine to keep and use it as is.

test/integration2/dialects/better-sqlite3.spec.js
dong-jun-shin: I’ve updated it to reuse the constant that was defined earlier. 🙂

I initially thought it would be better to redefine the expected value to keep the test more independent, but that was a mistake.

After reading your review again, I agree that since this is an integration test rather than a unit test, and the purpose is to verify precision loss caused by the DB driver, it makes more sense to reuse the constant as you suggested.

myndzi (approved):
LGTM 🎉 

Thanks!

test/integration2/dialects/better-sqlite3.spec.js
myndzi: For what it's worth, `Number.MAX_SAFE_INTEGER+1` is "safe-ish", so it's not actually losing precision:

> \> BigInt(Number.MAX_SAFE_INTEGER+1) === BigInt(Number.MAX_SAFE_INTEGER) + 1n
> true

However, it returns false from `Number.isSafeInteger()`. I believe it's a rounding artifact (the underlying data is definitely unsafe, but it just happens to output the expected value).

This test would work equally well with a value of 0, but I originally chose MAX_SAFE_INTEGER+1 as a way to make the test fail if something changes in the future such that we "conditially get a number or a bigint depending on the value". In retrospect I should have done +10 or something, but I saved that change for the casting PR I was working on, which would introduce its own tests anyway and maybe update the tests from this change.

mercmobily: "a way to make the test fail if something changes in the future such that we "conditially get a number or a bigint depending on the value".  I didn't expect any different from you :D

dong-jun-shin: Thanks for the clear explanation. To better understand it, I spent some time digging into how JavaScript Number values and IEEE-754 doubles behave.

I now see the intent behind using a boundary value to guard against future policy-level changes.

If you do open the casting-related PR you mentioned, I’d be interested in taking a look at how that idea is applied.

myndzi: @dong-jun-shin this is the backburnered PR, which you can follow if you like:
https://github.com/knex/knex/pull/6331

It probably won't see movement for a while, but that'd be the easiest way to follow along when I pick it back up :) 

@mercmobily I'm choosing a constant anyway, so it's no extra work to choose one that does "more" for me ;)

kibertoad (approved):

PR #6353, opened 2025-12-28
Clean up order by / group by

myndzi:
This PR is stacked on top of https://github.com/knex/knex/pull/5239 and can be merged instead of, or after, that PR to include the fixes to "nulls" ordering.

Here's the diff between this branch and the base PR:
https://github.com/OlivierCavadenti/knex/compare/fix/5135-order-nulls...myndzi:knex:myndzi/cleanup-orderby-groupby

---

Unwinds a lot of spaghetti and redundancy. Both "group by" and
"order by" were using the same code, though the "nulls first/last"
behavior only mattered to order by, which littered a lot of the
implementation details with conditional checks on group.type
and/or item.nulls.

Some of the behavior was defined in the base QueryCompiler class,
overridden in dialect clients, and then passed back to the base class.

Clients were using the "format this item for inclusion in a group
by / order by element" to "fake" adding an additional column to
"order by" by returning something _other_ than what they were
expected to, and then the base method was checking if it
thought they did that and re-adding the thing that the subclass
was supposed to have formatted.

Now, order by and group by each have their own logic. Classes
declaratively choose their "order by nulls" behavior from a few
prebuilt functions, with the option to implement their own
behavior as an expression (which oracle uses, via IIF)

### DIFF
--- a/test/unit/query/builder.js
+++ b/test/unit/query/builder.js
@@ -6302,7 +6302,7 @@ describe('QueryBuilder', () => {
   it('order by, null first', () => {
     testsql(qb().from('users').orderBy('foo', 'desc', 'first'), {
       mysql: {
-        sql: 'select * from `users` order by (`foo` is not null), `foo` desc',
+        sql: 'select * from `users` order by `foo` is not null, `foo` desc',
       },
       mssql: {
         sql: 'select * from [users] order by IIF([foo] is null,1,0) desc, [foo] desc',
@@ -6317,10 +6317,10 @@ describe('QueryBuilder', () => {
         sql: 'select * from "users" order by "foo" desc nulls first',
       },
       sqlite3: {
-        sql: 'select * from `users` order by (`foo` is not null), `foo` desc',
+        sql: 'select * from `users` order by `foo` is not null, `foo` desc',
       },
       cockroachdb: {
-        sql: 'select * from "users" order by ("foo" is not null), "foo" desc',
+        sql: 'select * from "users" order by "foo" is not null, "foo" desc',
       },
     });
   });
@@ -6332,7 +6332,7 @@ describe('QueryBuilder', () => {
         .orderBy([{ column: 'foo', order: 'desc', nulls: 'first' }]),
       {
         mysql: {
-          sql: 'select * from `users` order by (`foo` is not null), `foo` desc',
+          sql: 'select * from `users` order by `foo` is not null, `foo` desc',
         },
         mssql: {
           sql: 'select * from [users] order by IIF([foo] is null,1,0) desc, [foo] desc',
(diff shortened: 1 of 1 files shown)


test/integration2/query/select/selects.spec.js
```diff
@@ -597,137 +598,116 @@ describe('Selects', function () {
...
+        ];
+
+        // test mixed strings and nulls, mixed ints and nulls
+        for (const type of ['string', 'int']) {
+          // test all combinations of null-order and value-order
+          const column = `nullable_${type}`;
+
+          for (const { nulls, order, expected } of table) {
+            it(`nulls ${nulls}, order ${order}, values ${type}, array-of-objects syntax`, async () => {
+              await assertRows(
```
myndzi, line 689: We already have unit tests on the SQL construction, so using `.testSql` was redundant here. It also passed when a specific dialect was not included in the `tester(...)` calls even if the query would fail on that dialect. The new integration tests simply run on whatever databases we're running the tests against, and expect all dialects to be able to satisfy the query

test/integration2/query/select/selects.spec.js
```diff
@@ -597,137 +598,116 @@ describe('Selects', function () {
...
-            tester(
-              'mssql',
-              'select [id] from [OrderByNullTest] order by IIF([null_col] is null,1,0) asc, [string_col] asc',
-              [],
-              [1, 3, 2, 4]
-            );
-          });
-        await knex.schema.dropTable('OrderByNullTest');
+            it(`nulls ${nulls}, order ${order}, values ${type}, convenience syntax`, async () => {
+              await assertRows(
```
myndzi, line 701: I don't know why we had separate tests for mixed-null-and-integer vs mixed-null-and-string, so I just kept them

test/unit/query/formatter.js
```diff
@@ -1,31 +1,311 @@
...
+  describe('wrap', () => {
+    it('wraps an identifier as a string', () => {
+      for (const isParameter of [undefined, true, false]) {
+        const value = formatter().wrap('foo', isParameter);
+        expect(value).to.equal('"foo"');
+      }
+    });
+
+    for (const { driverName, formatter, ...rest } of forAllClients()) {
+      it(`escapes identifier values (driver=${driverName})`, () => {
```
myndzi, line 88: I couldn't find any direct tests on this behavior, which is a bit scary to have un-covered. This exercises the top-level "entry point" into "turn some stuff into a string we put in the raw SQL".

myndzi:
While reviewing the diff, I was nervous about some of the query construction so I dug into it in more detail. Added tests to cover some uncovered cases (specifically with parameter binding ordering in order by), and cleaned up the order by handling even more to ensure we do this correctly.

I also lurked in a bunch of tests around `Formatter`'s behavior since that was under-covered, and I now understand it better. I added a method to serve group by / order by in formatting their sql directly, instead of indirectly via `columnize`, which wasn't being used in the way it was intended.

I don't know what the CodeQL complaint about the regex thing is. It is a legitimate alert, but does not appear related to anything I changed. It's probably just scanning the "files I touched" and coming back with a newly-noticed thing that has been present for a long time.

I also have no idea what coveralls is complaining about because I can't even load the contents of the page it links to over multiple days of trying. It's probably some branch coverage on some of the defensive code patterns like `sql || ''` when sql is just never able to be falsy.

PR #6362, opened 2026-01-18
Postgres: clearer error when pg-query-stream is missing

### LINKED ISSUE #6358: Cannot find module 'pg-query-stream' when using query streams in Knex (PostgreSQL)
We’re using a streamed query in this handler:

```js
async function (request, reply) {
  const queryStream = await tasksRepository.createStream()
  const csvTransform = stringify({ header: true, columns: undefined })
  const gzipStream = createGzip()

  reply
    .header('Content-Type', 'application/gzip')
    .header(
      'Content-Disposition',
      `attachment; filename="${encodeURIComponent('tasks.csv.gz')}"`
    )

  return queryStream.pipe(csvTransform).pipe(gzipStream)
}
```

The repository implementation is:

```js
createStream () {
  return knex.select('*').from('tasks').stream()
}
```

All tests pass locally, but in CI the tests fail with:

```
Cannot find module 'pg-query-stream'
```

Run: [https://github.com/fastify/demo/actions/runs/20621476554/job/59224136187](https://github.com/fastify/demo/actions/runs/20621476554/job/59224136187)

Adding `pg-query-stream` as an explicit dependency fixes the CI failure.

It’s unclear why this works locally without the dependency but fails in CI. Is this relying on a transitive dependency that isn’t guaranteed to be present? If so, should `pg-query-stream` be declared explicitly when using `.stream()` with PostgreSQL?

Full code here:
[https://github.com/lraveri/demo/blob/feat/issue-74/src/plugins/app/tasks/tasks-repository.ts](https://github.com/lraveri/demo/blob/feat/issue-74/src/plugins/app/tasks/tasks-repository.ts)

[https://github.com/lraveri/demo/blob/feat/issue-74/src/routes/api/tasks/index.ts](https://github.com/lraveri/demo/blob/feat/issue-74/src/routes/api/tasks/index.ts)

lraveri:
Fixes #6358.

When using PostgreSQL query streaming without `pg-query-stream` installed, Knex now throws a clearer error explaining which package to install.

This keeps the dependency optional, but makes the failure mode easier to understand

### DIFF
--- a/test/unit/dialects/postgres.js
+++ b/test/unit/dialects/postgres.js
@@ -187,4 +187,107 @@ describe('Postgres Unit Tests', function () {
       await knexInstance.destroy();
     }
   });
+
+  it('does not require pg-query-stream in browser mode', async () => {
+    const knexInstance = knex({
+      client: 'postgresql',
+      version: '10.5',
+      connection: {
+        pool: {},
+      },
+    });
+
+    const Module = require('module');
+    const originalLoad = Module._load;
+    const loadStub = sinon
+      .stub(Module, '_load')
+      .callsFake((request, parent, isMain) => {
+        if (request === 'pg-query-stream') {
+          throw new Error('pg-query-stream should not be required in browser');
+        }
+        return originalLoad(request, parent, isMain);
+      });
+
+    const originalBrowser = process.browser;
+    process.browser = true;
+
+    try {
+      const connection = {
+        query() {
+          throw new Error('connection.query should not be called');
+        },
+      };
+
+      await knexInstance.client
+        ._stream(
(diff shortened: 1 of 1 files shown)


myndzi (approved):
LGTM

mercmobily:
I took the liberty to add a couple of tests since not all paths were covered. It now actually passes. @myndzi  @lraveri  Please check that you agree with them, then please just merge this so that we can close the issue and the PR :D

mercmobily:
@kibertoad Please merge this if you are OK with it all. I had assumed you had already given it your blessing, but I just realised that you didn't, so won't merge till you say yay.

PR #6365, opened 2026-01-22
Fixed error in documentation due to wrong conflict resolution

mercmobily:
I made the mistake of merging the docs via the UI; I was even impressed by how easy it was... and of course I didn't realise that I was actually merging it wrong. Apologies. This fixes it.


mercmobily:
@myndzi The error in the CI is due to some transient Windows problem.

PR #6368, opened 2026-01-23
Update docs deploy workflow (actions v4 + npm cache)

mercmobily:
This PR modernizes the docs deployment workflow: it bumps the GitHub Actions to v4 and switches caching to setup-node’s npm cache.

We drop the node_modules cache and the conditional install so npm ci always runs, using docs/package-lock.json as the cache key.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -7,6 +7,12 @@ on:
     paths:
       - docs/**
       - .github/workflows/**
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


mercmobily:
I tested this by running the action manually:

```
gh workflow run "Knex / Documentation deployment" --ref node-20-for-docs -f deploy=false
```

And it worked. I *think* it's good to go.

mercmobily:
Obsoleted by the not-yet-merged #6372

PR #6371, opened 2026-01-24
fix: handle escaped placeholders using even/odd backslash logic

### LINKED ISSUE #6363: Fix escaped placeholder parsing in raw SQL bindings across dialects (\\?, \\:name, etc)
While working on #6350 I entered another rabbit hole: escape characters.
Here is a hopefully comprehensive list of problems:

**All dialects (raw)**

This is the logic for array bindings:

```js
// lib/formatter/rawFormatter.js
const sql = raw.sql.replace(/\\?\?\??/g, function (match) {
  if (match === '\\?') return match;
  // ...consume binding for ? or ??
});
```

This only treats `\?` as escaped. It does not count backslashes. So `\\?` (escaped backslash + placeholder) is wrongly treated as escaped, and the
binding is not consumed. The same issue applies to any *even* number of backslashes (e.g. `\\\\?`).

This affects all dialects, but it also affects named bindings (object bindings), because `replaceKeyBindings` uses a similar “single backslash only” escape rule:

```js
// lib/formatter/rawFormatter.js
const regex = /\\?(:(\w+):(?=::)|:(\w+):(?!:)|:(\w+))/g;
```

So `\\:name` is affected too, as well as the identifier form `:name:` and cast form `:name::type` when escaped with an even number of backslashes.

**Other dialects**

As we know, for compiled queries Knex calls `client.positionBindings()` to convert `?` into the various `$1`, `@p0`, `:1`, etc. This is dialect-specific, and this is where "right vs wrong by dialect" applies.

**Some dialects are right...**

Postgres family — uses even/odd backslash counting:

```js
// lib/dialects/postgres/index.js
  if (escapes.length % 2) return '?';
  questionCount++;
  return `$${questionCount}`;
});
```

This correctly treats `\?` as literal `?`, and `\\?` as a real placeholder.
This correct logic is inherited by: postgres, pgnative, cockroachdb, redshift

**...some dialects appear to be wrong...**

MSSQL has the same problem as the raw formatter:

```js
// lib/dialects/mssql/index.js
return sql.replace(/\\?\?/g, (match) => {
  if (match === '\\?') return '?';
  return `@p${++questionCount}`;
});
```

This misinterprets `\\?` as escaped and skips the binding.

**...some dialects don't do it at all...**

Oracle / Oracledb seems to replace every `?` with `:n`, no escape logic:

```js
// lib/dialects/oracle/index.js
return sql.replace(/\?/g, function () {
  questionCount++;
  return `:${questionCount}`;
});
```

**...some dialects don't use positionBindings at all (they share Knex's "?" syntax).**

For mysql, mysql2, sqlite3, better-sqlite3, they keep `?` placeholders as-is (driver accepts them), so there’s no dialect-level conversion to judge. But they still use the shared raw formatter, so the raw `\\?` bug still applies.

Shivang731:
Fixes #6363

### DIFF
--- a/.github/workflows/assign-dri.yml
+++ b/.github/workflows/assign-dri.yml
@@ -1,0 +1,61 @@
+name: Assign PR to DRI
+
+on:
+  pull_request_target:
+    types: [opened, labeled, synchronize]
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  assign:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Assign based on DRI label
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const pr = context.payload.pull_request;
+            if (!pr) {
+              core.info('No pull request payload; exiting.');
+              return;
+            }
+
+            const labels = (pr.labels || []).map((label) => label.name);
+            const labelMatch = labels.find((name) => /^DRI:@/i.test(name));
+            if (!labelMatch) {
+              core.info('No DRI label found; skipping assignment.');
+              return;
+            }
+
+            const driUser = labelMatch
+              .replace(/^DRI:@/i, '')
+              .trim()
+              .replace(/^@+/, '');
+            core.info(`Found DRI label: ${labelMatch} -> ${driUser}`);
+
+            const driUserLower = driUser.toLowerCase();
+            const assignees = (pr.assignees || [])
+              .map((assignee) => assignee.login?.toLowerCase())
+              .filter(Boolean);
+            if (assignees.includes(driUserLower)) {
+              core.info('DRI already assigned; skipping assignment.');
+              return;
+            }
+
+            if (context.payload.action === 'synchronize') {
+              const sender = context.payload.sender?.login || '';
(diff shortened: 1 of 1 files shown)


mercmobily:
Made myself the DRI, since it's still fresh in my head. Thank you @Shivang731 , will review soon

mercmobily:
@Shivang731 Did you mean to close this PR?

Shivang731:
@mercmobily
Sorry about the confusion — I closed the PR intentionally after reading your review more carefully.

I realized the issue is broader than my initial understanding (especially around Oracle/Oracledb parity, skipped tests, and the regex concerns), and I don’t want to iterate noisily or leave half-baked fixes in review while I’m still getting fully comfortable with this part of the codebase.

I’m going to take some time to study the formatter and dialect implementations properly and will come back with a more complete, well-tested approach if I do reopen or submit a follow-up PR.

Thanks a lot for the detailed feedback and for taking the time to review — much appreciated.

mercmobily:
All good, no worries at all. I mean if you wanted a softer landing, having somebody review existing PRs will help us a lot!

PR #6372, opened 2026-01-25
Documentation Workflow: Release‑Only Publish and CI   Validation

mercmobily:
It's crucial that we have a clear workflow for documentation. This change makes sure the website is published ONLY when a new release is actually created. If a change adds or updates schema snippets, developers should regenerate `docs/generated/schema-snippets.json` and commit it as part of their PR, and then build the docs locally using the current working tree to verify the output. When they push a change, CI builds the docs and fails if a snippet is missing. This CI check does not regenerate snippets and does not publish the site.

In short, this PR does:

- Make docs builds reflect released packages.
- Make local docs/snippet validation use the current working tree.
- Add a docs build check on every push/PR; fail if "Snippet not available" is
  present in the built output, without deploying anything.
- Avoid SQLite snippet failures from a fake transaction stub.
- Improve `docs/README.md`, offering guidance on how to test documentation (plus
  one minor cosmetic change).
- Only inject the dummy `trx` stub when needed: the snippet uses
  `.transacting(` or references `trx` without declaring it.

### DIFF
--- a/lib/formatter/wrappingFormatter.js
+++ b/lib/formatter/wrappingFormatter.js
@@ -29,6 +29,8 @@ const operators = transform(
     'regexp',
     'not regexp',
     'match',
+    'similar to',
+    'not similar to',
     '&',
     '|',
     '^',
--- a/scripts/docker-compose.yml
+++ b/scripts/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3'
-
 services:
   mssql:
     image: mcr.microsoft.com/mssql/server:2019-latest
--- a/scripts/stress-test/docker-compose.yml
+++ b/scripts/stress-test/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3'
-
 services:
   toxiproxy:
     image: shopify/toxiproxy


mercmobily:
Assigning it to myself, I will try this on the forked repo to test it as much as possible

mercmobily:
@kibertoad @myndzi thoughts? The sooner we merge the release system, the better. We have a few things up in the air in terms of releasing/documentation/etc.

mercmobily:
I tested this on my own fork, and it worked 100%. I think we should merge this.
This means that the web site will only be published at each release, which makes total sense (we will add features, document them, and then the changes and new features will appear on the site).

Unassigning this from myself, and asking for review.

myndzi:
A few high-level thoughts:

- The code is triggered by a github release coming into existence, but that's not the same thing as "an NPM package" existing. The release can exist without the NPM package, causing a failure. We could install knex from the actual release file instead of NPM, but there's a better answer noted below

- The code is using the snippets file from the checkout, which is master, and not the release version; these may not have the same contents. Ideally we'd _generate_ the snippets from the version of knex installed as a dependency, but I understand that this is trickier. Alternate solutions include:
  - Checking out the tag that is created on release (probably least-bad, as long as we enforce a guarantee that the implementation cannot change when a tag is updated)
  - Creating a release artifact for the _snippets file_ (complicated)
  - Publishing the snippets with the knex package itself (seems bad - this file has no purpose there)

- The workflow is triggered by creating a Github release. The [other PR](https://github.com/knex/knex/pull/6344) doesn't do that, but could be made to. However, we still need a way to deploy documentation-site-only changes without bumping and releasing a new library version

Without being able to generate the docs via CI, this is unfortunately always going to be error-prone - requiring alignment between a checked-in artifact and a separate CI process that depends on it. Solving for "the ability to build the snippets in CI" would be the most fruitful thing - then we could definitively generate all necessary data from a concretely released version of knex. In lieu of that, one workaround I can think of that may be "least bad" is:

- Trigger a deployment / release of the documentation not on the creation of a Github release, but instead on a push to a tag. This would work with the pending release workflow as written
- Check out the tag, not master, when deploying documentation - this gives us a point-in-time snippets file that reflects the code we will be using to generate the documentation
- Assert that the knex version in the checked-out tag is the same as the version in master, so that we never deploy old documentation
- Git tags are _not_ immutable, but NPM artifacts are. The Git tag should contain the same library code that was released to NPM, so we need to ensure that pushes to tags are rejected unless the only changes are to files within the `docs` directory. We could automate this assertion with a workflow check; not sure if we can do it with protection rules

This would allow us to release fixes to the docs to a released version by updating an existing tag to include documentation changes that should be applied to that version.

myndzi:
cc @mercmobily it's not an actual review since there are some unanswered questions that I think we should sort out

myndzi:
(on further reflection, this problem mostly goes away / doesn't even exist once we merge the OIDC PR, because every code release goes to NPM anyway. That's honestly the best answer here. Once that is how we operate, then master is "always published to NPM" and thus it's not a problem for the documentation to reflect master)

PR #6378, opened 2026-02-11
feat: add UPDATE with JOIN support for PostgreSQL and SQLite

### LINKED ISSUE #2796: Update operations do not support joins

wingding12:
## Summary

Fixes #2796 — UPDATE operations now properly support JOIN clauses on PostgreSQL and SQLite, where they were previously silently ignored.

### Changes

- **Base query compiler** (`lib/query/querycompiler.js`): Includes `this.join()` in the `update()` method, placing joins between the table name and SET clause (MySQL-style syntax as default)
- **PostgreSQL** (`lib/dialects/postgres/query/pg-querycompiler.js`): Converts JOIN clauses into `FROM` + WHERE conditions, mirroring the existing `DELETE...USING` pattern. Handles multiple joins and works alongside the existing `updateFrom()` API.
- **SQLite3** (`lib/dialects/sqlite3/query/sqlite-querycompiler.js`): Converts JOIN clauses into `FROM` + WHERE conditions (SQLite 3.33.0+ supports `UPDATE...FROM` syntax)
- **Tests** (`test/unit/query/builder.js`): Adds 3 new test cases and updates 2 existing test expectations

### SQL Generation Examples

```javascript
knex('users')
  .join('photos', 'photos.id', 'users.id')
  .where({ 'users.active': true })
  .update({ email: 'foo' })
```

| Dialect | Generated SQL |
|---------|--------------|
| MySQL | `UPDATE users INNER JOIN photos ON photos.id = users.id SET email = ? WHERE users.active = ?` |
| PostgreSQL | `UPDATE users SET email = ? FROM photos WHERE users.active = ? AND photos.id = users.id` |
| MSSQL | `UPDATE users SET email = ? FROM users INNER JOIN photos ON photos.id = users.id WHERE users.active = ?` |
| SQLite | `UPDATE users SET email = ? FROM photos WHERE users.active = ? AND photos.id = users.id` |

### Test plan

- [x] All 459 existing unit tests pass
- [x] 3 new tests added: single join, join without WHERE, multiple joins
- [x] 2 existing test expectations updated for PG to reflect correct (non-silent-ignore) behavior

### DIFF
--- a/docs/src/guide/migrations.md
+++ b/docs/src/guide/migrations.md
@@ -56,7 +56,7 @@ $ knex migrate:make --stub
 ```
 
 - if a stub path is provided, it must be relative to the knexfile.\[js, ts, etc\] location
-- if a is used, the stub is selected by its file name. The CLI will look for this file in the config.migrations.directory folder. If the config.migrations.directory is not defined, this operation will fail
+- if a file name is used, the stub is selected by its file name. The CLI will look for this file in the config.migrations.directory folder. If the config.migrations.directory is not defined, this operation will fail
 
 Once you have finished writing the migrations, you can update the database matching your `NODE_ENV` by running:
 
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -187,9 +187,25 @@ class Client_PG extends Client {
   _stream(connection, obj, stream, options) {
     if (!obj.sql) throw new Error('The query is empty');
 
-    const PGQueryStream = process.browser
-      ? undefined
-      : require('pg-query-stream');
+    let PGQueryStream;
+    if (process.browser) {
+      PGQueryStream = undefined;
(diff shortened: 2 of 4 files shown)


kibertoad:
@myndzi could you please review?

PR #6379, opened 2026-02-11
fix(mysql): preserve expression defaults when renaming columns

### LINKED ISSUE #1888: Renaming columns with a default value of CURRENT_TIMESTAMP generates invalid SQL
Knex version: 0.12.6
MySql version: 5.7.13

Attempting to rename the column will generate SQL with single quotes around `'CURRENT_TIMESTAMP'`, which causes it to be a static string and therefore an invalid default value for the `DATETIME`/`TIMESTAMP` column:

```
{ [Error: alter table `Example` change `ts` `timestamp` timestamp NOT NULL DEFAULT 'CURRENT_TIMESTAMP' - ER_INVALID_DEFAULT: Invalid default value for 'timestamp']
  code: 'ER_INVALID_DEFAULT',
  errno: 1067,
  sqlState: '42000',
  index: 0 }
```

To reproduce:

 ```
var knex = require('knex');

var config = {
  client: 'mysql',
  connection: {
    host: '127.0.0.1',
    user: 'root',
    password: '',
    database: 'knex_test',
    charset: 'utf8mb4',
    timezone: '0000',
  },
  pool: {
    min: 2,
    max: 10,
  },
};

var connection = knex(config);

connection.schema
  .createTableIfNotExists('Example', (table) => {
    table.timestamp('ts')
      .defaultTo(connection.fn.now());
  })
  .table('Example', (table) => {
    table.renameColumn('ts', 'timestamp');
  })
  .then(() => {
    connection.destroy();
  })
  .catch((err) => {
    console.error(err);
    connection.destroy();
  });
```

This does not appear to be a problem with the `connection.fn.now()` function, as creating the column with:
```
table.timestamp('ts')
  .defaultTo(connection.raw('CURRENT_TIMESTAMP'));
```
will still cause the error to occur when renaming the column.

This error occurs for `DATETIME` and `TIMESTAMP` column types of all precision, that is, changing the column type to `DATETIME(3)` and the default value to `CURRENT_TIMESTAMP(3)` will cause the same error to occur.

### PR DESCRIPTION
**Pull request by wingding12**
## Summary

Fixes #1888 -- `renameColumn()` on MySQL now correctly handles expression-based default values like `CURRENT_TIMESTAMP`, `NOW()`, `UUID()`, and MySQL 8.0+ parenthesized expression defaults.

### Problem

When renaming a column that has an expression default (e.g. `DEFAULT CURRENT_TIMESTAMP`), the MySQL dialect's `renameColumn` method wraps **all** default values in single quotes:

```sql
-- Generated (broken):
ALTER TABLE example CHANGE ts timestamp TIMESTAMP NOT NULL DEFAULT 'CURRENT_TIMESTAMP'
-- ER_INVALID_DEFAULT: Invalid default value for 'timestamp'

-- Expected (correct):
ALTER TABLE example CHANGE ts timestamp TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
```

Additionally, the `ON UPDATE CURRENT_TIMESTAMP` clause was silently lost during rename operations.

### Solution

1. **`isExpressionDefault(defaultValue, extra)`** -- Detects expression defaults using:
   - The `EXTRA` field's `DEFAULT_GENERATED` flag (MySQL 8.0.13+)
   - Regex pattern matching for known MySQL expressions (`CURRENT_TIMESTAMP`, `NOW()`, `UUID()`, `UTC_TIMESTAMP`, parenthesized expressions, etc.)

2. **`parseExtra(extra)`** -- Extracts structured attributes from the `SHOW FULL FIELDS` Extra column:
   - `isAutoIncrement` -- replaces the previous loose equality check
   - `onUpdateValue` -- preserves the `ON UPDATE` clause
   - `isDefaultGenerated` -- MySQL 8.0.13+ expression default marker

3. **Updated `renameColumn()`** -- Uses these utilities to emit expression defaults unquoted and preserve the `ON UPDATE` clause.

### Files Changed

| File | Change |
|------|--------|
| `lib/dialects/mysql/schema/internal/isExpressionDefault.js` | New utility |
| `lib/dialects/mysql/schema/internal/parseExtra.js` | New utility |
| `lib/dialects/mysql/schema/mysql-tablecompiler.js` | Updated renameColumn |
| `test/unit/dialects/mysql/isExpressionDefault.js` | 27 test cases |
| `test/unit/dialects/mysql/parseExtra.js` | 13 test cases |

### Test plan

- [x] 40 new unit tests pass (expression detection + Extra parsing)
- [x] All 459 existing query builder tests pass
- [x] Covers: temporal functions, UUID functions, MySQL 8.0+ parenthesized expressions, DEFAULT_GENERATED detection, literal value preservation, edge cases (null, empty, numeric, JSON defaults)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,23 +4,23 @@
 
 ### Bug fixes
 
-- andWhereNotJsonObject calling wrong function (#5683)
-- PostgreSQL: fix error when setting query_timeout (#5673)
-- MySQL: Missing comments on delete, update and insert (#5738)
-- MySQL: Fixed issue with bigincrements not working with composite primary key - #5341 (#5343)
+- andWhereNotJsonObject calling wrong function ([#5683](https://github.com/knex/knex/issues/5683))
+- PostgreSQL: fix error when setting query_timeout ([#5673](https://github.com/knex/knex/issues/5673))
+- MySQL: Missing comments on delete, update and insert ([#5738](https://github.com/knex/knex/issues/5738))
+- MySQL: Fixed issue with bigincrements not working with composite primary key - [#5341](https://github.com/knex/knex/issues/5341) ([#5343](https://github.com/knex/knex/issues/5343))
 
 ### Types
 
-- Add type definitions for orHavingNull and orHavingNotNull (#5669)
-- Import knex as type in TS migration template (#5741)
-- Fix conditional constraint error (#5747)
-- PostgreSQL: Fix typing to reflect pg typing change (#5647)
+- Add type definitions for orHavingNull and orHavingNotNull ([#5669](https://github.com/knex/knex/issues/5669))
+- Import knex as type in TS migration template ([#5741](https://github.com/knex/knex/issues/5741))
+- Fix conditional constraint error ([#5747](https://github.com/knex/knex/issues/5747))
+- PostgreSQL: Fix typing to reflect pg typing change ([#5647](https://github.com/knex/knex/issues/5647))
 
 ### New features
 
-- Add transactor.parentTransaction (#5567)
-- MySQL: Added implementation for upsert (#5743)
-- Oracle: Support Object Names Greater than 30 Characters for Oracle DB Versions 12.2 and Greater (#5197)
+- Add transactor.parentTransaction ([#5567](https://github.com/knex/knex/issues/5567))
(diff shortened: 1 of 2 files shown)


kibertoad:
thanks! please add an integration test

wingding12:
> thanks! please add an integration test

added!

PR #6380, opened 2026-02-11
fix: prevent pre-existing WHERE clauses from leaking into upsert

### LINKED ISSUE #5257: Ambiguous column error with onConflict merge
# Environment

Knex version: 2.1.0
Database + version: Postgres 14
OS: MacOS

# Bug

When querying a table like so:

```js
 knex('option')
      .insert({ key, value: JSON.stringify(value) })
      .onConflict("key")
      .merge(["value"])
```

I get this error:

> error: insert into "option" ("key", "value") values ($1, $2) on conflict ("key") do update set "value" = excluded."value" where "key" = $3 - column reference "key" is ambiguous

I've tried running that query directly in postgres and it's the last `where "key" = $3` that it doesn't like. It should be `where "option"."key" = $3` to distinguish it from `excluded.key`. But I don't think the `where` clause is required at all since the `on conflict update` is scoped to the current row anyway.

When I build the same query on an unconnected Knex instance, it does not add the bogus `where` clause, as demonstrated here: https://runkit.com/tamlyn/knex-onconflict-merge

Also, interestingly, if this is the first query I run after starting the app, it works. But if I query the table first then subsequently run the query above, it fails. It's like something is introspecting the database and then adding this incorrect `where` clause.

For completeness, here's the table schema:

```sql
CREATE TABLE option (
    key text PRIMARY KEY,
    value jsonb NOT NULL 
  )
```

wingding12:
## Summary

Fixes #5257 — When a query builder instance is reused, WHERE clauses added before `.insert()` no longer leak into the `onConflict().merge()` upsert statement.

### Problem

```javascript
const builder = knex('option').where('key', 'test');
builder.insert({ key: 'new', value: 'bar' }).onConflict('key').merge(['value']);

// Before fix — leaked WHERE produces ambiguous column error:
// insert into "option" (...) values (...)
//   on conflict ("key") do update set "value" = excluded."value"
//   where "key" = ?  ← leaked!

// After fix — clean upsert:
// insert into "option" (...) values (...)
//   on conflict ("key") do update set "value" = excluded."value"
```

The intentional `.merge().where()` conditional upsert feature (PostgreSQL) still works:

```javascript
knex('users').insert({...}).onConflict('email').merge().where('users.active', true)
// → ... on conflict ("email") do update set ... where "users"."active" = ?
```

### Approach

When `merge()` is called on the `OnConflictBuilder`, it snapshots the current WHERE clause count as `mergeWhereOffset`. The new `_mergeWheres()` method on the base query compiler then only includes WHERE clauses added **after** that offset, filtering out any that existed before the upsert chain began.

### Files changed (6)

| File | Change |
|------|--------|
| `lib/query/querybuilder.js` | Record `mergeWhereOffset` in `OnConflictBuilder.merge()` |
| `lib/query/querycompiler.js` | Add `_mergeWheres()` method to base compiler |
| `lib/dialects/postgres/query/pg-querycompiler.js` | Use `_mergeWheres()` in insert |
| `lib/dialects/sqlite3/query/sqlite-querycompiler.js` | Use `_mergeWheres()` in insert |
| `lib/dialects/mysql/query/mysql-querycompiler.js` | Use `_mergeWheres()` in insert |
| `test/unit/query/builder.js` | 4 new test cases |

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -131,8 +131,8 @@ jobs:
           git add .
           git commit -m "chore: release v$(node -p "require('./package.json').version")"
           git tag "v$(node -p "require('./package.json').version")"
-          echo :::: git push
-          echo :::: git push --tags
+          git push
+          git push --tags
 
       - uses: actions/attest-build-provenance@v3
         with:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,23 +4,23 @@
 
 ### Bug fixes
 
-- andWhereNotJsonObject calling wrong function (#5683)
-- PostgreSQL: fix error when setting query_timeout (#5673)
-- MySQL: Missing comments on delete, update and insert (#5738)
-- MySQL: Fixed issue with bigincrements not working with composite primary key - #5341 (#5343)
+- andWhereNotJsonObject calling wrong function ([#5683](https://github.com/knex/knex/issues/5683))
+- PostgreSQL: fix error when setting query_timeout ([#5673](https://github.com/knex/knex/issues/5673))
+- MySQL: Missing comments on delete, update and insert ([#5738](https://github.com/knex/knex/issues/5738))
+- MySQL: Fixed issue with bigincrements not working with composite primary key - [#5341](https://github.com/knex/knex/issues/5341) ([#5343](https://github.com/knex/knex/issues/5343))
 
 ### Types
 
-- Add type definitions for orHavingNull and orHavingNotNull (#5669)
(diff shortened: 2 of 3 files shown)


kibertoad:
thank you! could you please add an integration test?

wingding12:
> thank you! could you please add an integration test?

added!

tgriesser:
@wingding12 can you give a more concrete situation where this was an issue? Just want to understand why you'd want to reuse the builder without the wheres, feels like it's maybe special case/going against the patterns everywhere else where you'd otherwise just `.clone().clearWhere()`

PR #6381, opened 2026-02-12
fix: handle escaped placeholder parsing across all dialects (#6363)

### LINKED ISSUE #6363: Fix escaped placeholder parsing in raw SQL bindings across dialects (\\?, \\:name, etc)

wingding12:
## Summary

Fixes #6363.

The raw formatter and several dialect-specific `positionBindings()` implementations only handled a single backslash before `?` placeholders using `/\\?\?/g`. This meant `\\?` (escaped backslash + real placeholder) was incorrectly treated as an escaped question mark, leaving the binding unconsumed.

This PR applies the even/odd backslash counting approach (already used by PostgreSQL's `positionBindings`) consistently across all affected layers:

### Changes

**`lib/formatter/rawFormatter.js`** — Both array and named binding replacement regexes now use `/(\\*)(...)/g` to capture arbitrarily many leading backslashes and check parity:
- **Array bindings**: `/(\\*)(\?\??)/g` replaces the old `/\\?\?\??/g`
- **Named bindings**: `/(\\*)(:(\w+):(?=::)|:(\w+):(?!:)|:(\w+))/g` replaces `/\\?(:(\w+):...)/g`

**`lib/dialects/mssql/index.js`** — `positionBindings()` regex updated from `/\\?\?/g` to `/(\\*)(\?)/g` with modulo check, matching the PostgreSQL pattern.

**`lib/dialects/oracle/index.js`** — `positionBindings()` had **no escape handling at all** (`/\?/g`). Now uses `/(\\*)(\?)/g` with modulo check.

**`lib/client.js`** — Base `positionBindings()` (inherited by MySQL, SQLite, and other dialects that use `?` natively) now strips escape backslashes so `\?` in raw SQL becomes a literal `?` rather than `\?` being sent to the driver.

### Behavior

| Input (in raw SQL string) | Backslash count | Parity | Meaning |
|---|---|---|---|
| `?` | 0 | even | Real placeholder |
| `\?` | 1 | odd | Escaped — literal `?` |
| `\\?` | 2 | even | Real placeholder |
| `\\\?` | 3 | odd | Escaped — literal `?` |

### Tests

- Added `positionBindings` tests for Oracle, PostgreSQL, and MSSQL with even/odd backslash scenarios
- Added builder-level `testquery` test for the end-to-end `\\?` case

### DIFF
--- a/docs/src/guide/migrations.md
+++ b/docs/src/guide/migrations.md
@@ -56,7 +56,7 @@ $ knex migrate:make --stub
 ```
 
 - if a stub path is provided, it must be relative to the knexfile.\[js, ts, etc\] location
-- if a is used, the stub is selected by its file name. The CLI will look for this file in the config.migrations.directory folder. If the config.migrations.directory is not defined, this operation will fail
+- if a file name is used, the stub is selected by its file name. The CLI will look for this file in the config.migrations.directory folder. If the config.migrations.directory is not defined, this operation will fail
 
 Once you have finished writing the migrations, you can update the database matching your `NODE_ENV` by running:
 
--- a/lib/dialects/postgres/index.js
+++ b/lib/dialects/postgres/index.js
@@ -187,9 +187,25 @@ class Client_PG extends Client {
   _stream(connection, obj, stream, options) {
     if (!obj.sql) throw new Error('The query is empty');
 
-    const PGQueryStream = process.browser
-      ? undefined
-      : require('pg-query-stream');
+    let PGQueryStream;
+    if (process.browser) {
+      PGQueryStream = undefined;
+    } else {
+      try {
+        PGQueryStream = require('pg-query-stream');
+      } catch (e) {
+        if (
+          e instanceof Error &&
+          e.code === 'MODULE_NOT_FOUND' &&
+          e.message.includes('pg-query-stream')
+        ) {
+          throw new Error(
+            "knex PostgreSQL query streaming requires the 'pg-query-stream' package. Please install it (e.g. `npm i pg-query-stream`)."
+          );
+        }
+        throw e;
+      }
+    }
     const sql = obj.sql;
 
     return new Promise(function (resolver, rejecter) {
--- a/package.json
+++ b/package.json
@@ -87,6 +87,9 @@
     "tarn": "^3.0.2",
     "tildify": "2.0.0"
   },
+  "peerDependencies": {
+    "pg-query-stream": "^4.2.4"
+  },
   "peerDependenciesMeta": {
     "tedious": {
       "optional": true
@@ -101,6 +104,9 @@
       "optional": true
     },
     "pg-native": {
+      "optional": true
+    },
+    "pg-query-stream": {
       "optional": true
     },
     "sqlite3": {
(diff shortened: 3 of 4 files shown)


OlivierCavadenti:
hi, thanks !
One thing I noticed: in positionBindings (MSSQL, Oracle, base client.js), even backslashes like \\? get the binding substituted but the \\ themselves are never stripped, they just stay in the output. The rawFormatter handles this correctly by only returning the replacement value, but the positionBindings paths return the full match including the backslashes. Worth checking if that's intentional.

Also, the "raw named bindings with escaped backslash before placeholder" builder test doesn't actually have any backslashes in the query string ('col = :val'), so it's not really testing the fix.

PR #6384, opened 2026-02-18
Add lateral join methods (joinLateral, leftJoinLateral, crossJoinLateral)

### LINKED ISSUE #3732: Add lateral join methods
I've been monkey-patching knex to include some lateral join methods but it'd be nice if the library actually supported them. Adding some kind of extra parameter to the existing join methods would be a potentially breaking change, not to mention the existing method signatures aren't all compatible (lateral joins have to use a subquery). So I propose adding 8 additional methods:

* innerJoinLateral
* leftJoinLateral
* leftOuterJoinLateral
* rightJoinLateral
* rightOuterJoinLateral
* outerJoinLateral
* fullOuterJoinLateral
* crossJoinLateral

Signatures would look something like:

```
(raw: Raw): QueryBuilder;
(subquery: QueryCallback, on: JoinCallback): QueryBuilder;
(subquery: QueryCallback, on: Raw): QueryBuilder;
(subquery: QueryCallback, columns: { [key: string]: string | number | boolean | Raw }): QueryBuilder;
(subquery: QueryCallback, column1: string, column2: string): QueryBuilder;
(subquery: QueryCallback, column1: string, column2: Raw): QueryBuilder;
(subquery: QueryCallback, column1: string, operator: string, column2: string): QueryBuilder;
```

I can try to put together a PR, but I wanted to confirm this sort of change would be acceptable first.

wingding12:
## Summary

Adds support for LATERAL joins across PostgreSQL, MySQL, and MSSQL dialects, resolving a long-standing feature request.

### New methods

- `joinLateral(table, onCallback)` / `innerJoinLateral(table, onCallback)`
- `leftJoinLateral(table, onCallback)` / `leftOuterJoinLateral(table, onCallback)`
- `crossJoinLateral(table, onCallback)`

### Cross-dialect compilation

- **PostgreSQL / MySQL**: Standard SQL `LATERAL` keyword is appended after the join type (e.g. `inner join lateral (...)`, `left join lateral (...)`, `cross join lateral (...)`).
- **MSSQL**: Translated to the equivalent `APPLY` syntax. Inner/cross lateral becomes `CROSS APPLY`, left/left outer lateral becomes `OUTER APPLY`. The MSSQL query compiler overrides `join()` to handle this, skipping `ON` clauses for APPLY since APPLY does not use them.

### Changes

| File | Change |
|------|--------|
| `lib/query/joinclause.js` | Added `_lateral` property to JoinClause |
| `lib/query/querybuilder.js` | Added 5 lateral join methods + `_lateral()` flag helper |
| `lib/query/querycompiler.js` | Emit `lateral` keyword when `join._lateral` is true |
| `lib/dialects/mssql/query/mssql-querycompiler.js` | Override `join()` to emit `CROSS APPLY` / `OUTER APPLY` for lateral joins |
| `types/index.d.ts` | Added TypeScript signatures for all lateral join methods |
| `test/unit/query/builder.js` | 5 new unit tests covering all lateral join variants and dialect outputs |

### Example usage

```js
// PostgreSQL / MySQL
knex('users')
  .select('users.id', 'top_orders.total')
  .joinLateral(
    knex.select('total').from('orders')
      .whereRaw('"orders"."user_id" = "users"."id"')
      .orderBy('total', 'desc')
      .limit(3)
      .as('top_orders'),
    function() { this.on(knex.raw('1 = 1')); }
  );
// => select ... from "users" inner join lateral (...) as "top_orders" on 1 = 1

// MSSQL
// => select ... from [users] cross apply (...) as [top_orders]
```

Closes #3732

### DIFF
--- a/lib/dialects/mssql/query/mssql-querycompiler.js
+++ b/lib/dialects/mssql/query/mssql-querycompiler.js
@@ -8,6 +8,8 @@ const isEmpty = require('lodash/isEmpty');
 const Raw = require('../../../raw.js');
 const {
   columnize: columnize_,
+  wrap: wrap_,
+  unwrapRaw: unwrapRaw_,
 } = require('../../../formatter/wrappingFormatter');
 
 const components = [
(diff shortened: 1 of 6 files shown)


lib/dialects/mssql/query/mssql-querycompiler.js
```diff
@@ -60,6 +62,66 @@ class QueryCompiler_MSSQL extends QueryCompiler {
     return result;
   }
 
+  join() {
```
OlivierCavadenti, line 65: it's a full copy-paste of the base compiler's join() (60 lines). If the base method ever gets a fix or a new clause type, MSSQL will silently diverge. Worth looking into calling super.join() for the non-lateral path

test/unit/query/builder.js
```diff
@@ -4585,6 +4585,144 @@ describe('QueryBuilder', () => {
...
+        .from('users')
+        .crossJoinLateral(
+          qb().select('id').from('orders').as('o'),
+          function () {
+            this.on(raw('1 = 1'));
+          }
+        ),
+      {
+        pg: {
+          sql: 'select * from "users" cross join lateral (select "id" from "orders") as "o" on 1 = 1',
```
OlivierCavadenti, line 4711: The on '1 = 1) here is a workaround, not a real test, CROSS JOIN LATERAL shouldn't need an ON clause. The base compiler always emits ON for non-raw joins, so is forced to pass this dummy callback. This should be fixed in the compiler so that crossJoinLateral works without a callback, same as crossJoin already does in some dialects.

PR #6387, opened 2026-02-28
Add missing semver package

joshkel:
Needed by [test/cli/esm-interop.spec.js](https://github.com/knex/knex/blob/c18fb1ba2dc3001ee0fb2a79c126a32e6cd831a5/test/cli/esm-interop.spec.js#L7)


kibertoad (approved):

kibertoad:
thank you!

PR #6388, opened 2026-02-28
Fix pg-query-stream 4.13.0 incompatibility

joshkel:
pg-query-stream 4.13.0 updates its query stream handling to invoke its callback under more circumstances. This causes awaiting on a Knex Postgres stream to return early.

See https://github.com/brianc/node-postgres/pull/2810

This is reproducible with the current Knex test suite, if run against pg-query-stream 4.13.0 instead of an older version.


kibertoad:
@joshkel Thank you for the PR! this is fixed in a slightly different way in the latest knex release

PR #6386, opened 2026-02-26
fix(postgres): postgres updateFrom value binding order

Duod-Notyap:
## Summary
Fixes #6376 

### Issue
`QueryCompiler.where()` was being called prior to `QueryCompiler_PG._updateFrom()`, subsequently confusing parameter binding order when using a `UPDATE ... FROM VALUES` statement in postgres.

### Resolution
Moved order of function calls slightly to avoid problem.

PR #6390, opened 2026-03-06
Fix update from raw

dertieran:
The `updateFrom` query was compiled after the `where` clause this caused the order of the bindings to be wrong.


dertieran:
Just saw that this is a duplicate of #6386 
MB for not checking first, closing this PR.

PR #6392, opened 2026-03-11
Fix: Properly Escape Aliases in Analytic Functions

dertieran:
**Problem**

The **analytic functions** (`row_number`, `rank`, etc.) do not correctly escape aliases when generating SQL.

For example:

```js
knex('users').select('*').rowNumber('rowNumber', 'email', 'firstName');
```

produces:

```sql
select *, row_number() over (partition by "firstName" order by "email") as rowNumber from "users"
```

Unquoted aliases can cause errors in SQL dialects that require escaping for camelCase or special-character names.

**Solution**

This PR updates the query builder to use the `columnize` method when generating aliases for analytic functions.

After this change, the generated SQL is properly escaped:

```sql
select *, row_number() over (partition by "firstName" order by "email") as "rowNumber" from "users"
```

All existing unit tests for analytic functions were updated to reflect this change.
If needed, I can add a dedicated test to explicitly verify that aliases are escaped correctly.


OlivierCavadenti (commented):
Thanks for the fix! Small note: the rest of querycompiler.js uses `wrap` for aliases (see `_aggregate` line 390), not `columnize`. Both work for a single string but `wrap` would be more consistent:

```js
sql += ' as ' + self.formatter.wrap(stmt.alias);
```

OlivierCavadenti (approved):

Issue #6403, opened 2026-03-23
Bug: ESM types not resolved — missing `types/index.d.mts` (knex 3.2.3)


thetutlage:
### Description

`package.json` exports declare a `.d.mts` types entry for the ESM import path, but the file does not exist:

```jsonc
// package.json exports
".": {
  "import": {
    "types": "./types/index.d.mts",  // ← does not exist
    "default": "./knex.mjs"
  },
  "require": {
    "types": "./types/index.d.ts",
    "default": "./knex.js"
  }
}
```

Shipped files in `types/`:

```
types/
├── index.d.ts
├── result.d.ts
└── tables.d.ts
```

No `.d.mts` files are published.

### Impact

Any ESM TypeScript project using `import { Knex } from 'knex'` gets no type resolution — everything from `knex` becomes `any`:

```
error TS7016: Could not find a declaration file for module 'knex'.
  There are types at 'node_modules/knex/types/index.d.ts', but this
  result could not be resolved when respecting package.json "exports".
```

### Fix

Either:

1. Ship `types/index.d.mts` (can be a re-export of `index.d.ts`), or
2. Point the ESM types entry to the existing `.d.ts` file:

```diff
 "import": {
-  "types": "./types/index.d.mts",
+  "types": "./types/index.d.ts",
   "default": "./knex.mjs"
 }
```

kibertoad:
@thetutlage can you please try 3.2.4?

B4nan:
I had the same error in CI, with 3.2.4 it fails too, but differently:

```
Run yarn build
Error: src/raw.ts(2,15): error TS2724: '"knex"' has no exported member named 'Knex'. Did you mean 'knex'?
The command failed in workspace @mikro-orm/knex-compat@workspace:packages/knex-compat with exit code 2
The command failed for workspaces that are depended upon by other workspaces; can't satisfy the dependency graph
Failed with errors in 2m 43s
Error: Process completed with exit code 1.
```

https://github.com/mikro-orm/mikro-orm/actions/runs/23438075267/job/68181213099?pr=7380

---

Clearly, this should have been a major bump with some RC phase 🙃

kibertoad:
@B4nan tbf, I'm not sure people actively use rc and errors would actually get reported

kibertoad:
will fix when I reach the hotel

B4nan:
[...]

PR #6406, opened 2026-03-24
fix: add type exports

tgriesser:
Restores `types/*.d.ts` & `types/*` imports as mentioned by @colatkinson in #6403

Additionally, adds `bin/*` just incase anyone was relying on the code in there as a public interface for whatever reason. If we want to make further changes to lock down the public API we can do that in a breaking change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,19 +4,19 @@
 
 ### Bug fixes
 
-- Fix module exports (#6405)
+- Fix module exports [#6406](https://github.com/knex/knex/issues/6406)
 
 # 3.2.5 - 23 March, 2026
 
 ### Bug fixes
 
-- Fix ESM exports (#6405)
+- Fix ESM exports [#6405](https://github.com/knex/knex/issues/6405)
 
 # 3.2.4 - 23 March, 2026
 
 ### Bug fixes
 
-- Fix ESM type exports
+- Fix ESM type exports [#6404](https://github.com/knex/knex/issues/6404)
 
 # 3.2.1 - 22 March, 2026
 


package.json
```diff
@@ -36,7 +40,8 @@
     "coveralls": "nyc report --reporter=lcov",
     "lint": "eslint --cache .",
     "lint:fix": "eslint --cache --fix .",
-    "lint:types": "tsd && tstyche --target 5.4 test-tstyche/esm-types.tst.ts",
+    "prelint:types": "cd test-tstyche && npm i",
```
kibertoad, line 43: why have npm install as a part of a specific script?

tgriesser: Because it’s npm installing to resolve the `file:..` in the `test-tstyche` dir to properly test the types. Unless I was missing something, the paths in the tsconfig that was previously in there wasn’t suitable for being able to test the export path behavior correctly.

tgriesser: @kibertoad I could put it all in `lint:types` I suppose but then it’s 2 cd commands. The `pre*` hooks run automatically before the commands they’re named for. Either way is fine with me 🤷‍♂️

kibertoad: what I mean is we don't do it for any other commands. why not just call it explicitly from workflows where necessary?

tgriesser: > what I mean is we don't do it for any other commands

Because no other commands need to do an `npm install` before they run

If you pull down the repo and `npm install` and then run `npm run lint:types` it will fail without it which feels incorrect.

Looking at the scripts in there, it'd probably be nicer DX wise to move the `db:start:$DB` and `db:stop:$DB` to `pretest:$DB` and `posttest:$DB` so you can just pull down and run `npm run test:mssql` and everything just works ™️ 

image

Just an idea. Sorry, not trying to change too many things at once while jumping back in here, just went with what I would typically do to get a fix out quicker, happy to change

PR #6408, opened 2026-03-25
fix: sqlite DDL operations failing inside transactions #6402

### LINKED ISSUE #6402: Bug: SQLite DDL operations fail inside transactions (knex 3.2.3)
This behavior was introduced in this PR https://github.com/knex/knex/pull/6315

### Reproduction

```js
  client: 'sqlite3',
  connection: { filename: ':memory:' },
  useNullAsDefault: true,
});

await knex.schema.createTable('users', (t) => {
  t.increments('id');
});
await knex.schema.createTable('accounts', (t) => {
  t.increments('id');
  t.integer('user_id').unsigned().references('users.id');
});

await trx.schema.table('accounts', (t) => {
  t.dropForeign(['user_id']); // throws
});
```

```
Error: Refusing to create transaction: unable to change `foreign_keys` pragma inside a nested transaction
```

Same error for `dropColumn`, `renameColumn`, `setNullable`, `dropNullable` — anything that goes through `SQLite3_DDL`.

### Cause

`ddl()` in `lib/dialects/sqlite3/index.js` always uses `this._strict()`, so `strictForeignKeyPragma = true`.

When DDL reads table metadata via `getTableSql()`, it creates a nested transaction with `{ enforceForeignCheck: null }` (meaning "don't touch the pragma").

`_setForeignCheck(conn, null)` correctly returns `null` — "no change made." But the guard in `_evaluateContainer` checks:

```js
restoreForeignCheck !== undefined  // null !== undefined → true
```

So it throws even though no pragma change was needed.

### Fix

```diff
-       restoreForeignCheck !== undefined
+       restoreForeignCheck != null
```

This is consistent with `_restoreForeignCheck` which already treats `null` as "nothing to restore" via `typeof enable !== 'boolean'`.

tgriesser:
Fixes #6402

@thetutlage I added your repro as the test case to make sure it was fixed, it seemed we also needed to also change the behavior in `alter` to fully fix it.

### DIFF
--- a/lib/dialects/sqlite3/execution/sqlite-transaction.js
+++ b/lib/dialects/sqlite3/execution/sqlite-transaction.js
@@ -88,7 +88,7 @@ class Transaction_Sqlite extends Transaction {
       if (
         strictForeignKeyPragma &&
         hasOuterTransaction &&
-        restoreForeignCheck !== undefined
+        restoreForeignCheck != null
       ) {
         throw new Error(
           `Refusing to create transaction: unable to change \`foreign_keys\` pragma inside a nested transaction`
--- a/lib/dialects/sqlite3/schema/ddl.js
+++ b/lib/dialects/sqlite3/schema/ddl.js
@@ -334,6 +334,10 @@ class SQLite3_DDL {
   }
 
   async alter(newSql, createIndices, columns) {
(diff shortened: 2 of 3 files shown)


thetutlage:
Fixed. Thanks 🙏

PR #6413, opened 2026-03-26
feat: consolidate all testing under vitest (very wip)

tgriesser:
With the help of Claude for a lot of the grunt work, took a pass at converting our disparate test suites (`mocha`, `tape`, `jake`) to as `vitest`

Did my best to keep the files in place, to ensure we didn't change much about the existing testing patterns. This way it's less of a pain as we merge other PRs down, there should be fewer conflicts. Once the dust has settled a bit we can start to standardize directory naming and move things around.

Figure eventually we can start to use fancier features like snapshot testing, running more tests in parallel, etc.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,7 +1,13 @@
 ---
 name: Coverage
 
-on: ['push', 'pull_request']
+on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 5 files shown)


tgriesser:
Closing for now - we'll get there eventually but too ambitious while we're still getting through the issues

Issue #6292, opened 2025-10-18
Proposal: Experimental Pooling Integration Mode (selectable, backwards-compatible)


mercmobily:
Context
- Related: #6241 (META: Connection and pooling issues), #6286 (META: External Connection Pooling and Custom Pool Integration).
- Idea: introduce a "new shiny" pooling subsystem as an opt-in, marked experimental, so we can iterate aggressively without impacting existing users. The default remains the current tarn-based integration.

Goals
- Keep tarn as the underlying pool implementation.
- Provide a selectable integration mode that rewrites Knex’s pooling glue to address reliability issues tracked in #6241.
- Default behavior unchanged; users can opt in to the experimental integration.

Configuration (proposal)
- New option: `pool.integration: 'legacy' | 'experimental'` (default: `legacy`).
- Alternative naming bikeshed: `pool.mode`, `pool.impl`, `pool.engine` (but intent is integration, not replacing tarn).

Architecture Sketch
- Introduce a thin adapter around tarn that centralizes acquire/release/destroy/validate and exposes small hooks (`createRawConnection`, `destroyRawConnection`, `validateConnection`, `afterCreate`).
- Experimental integration keeps tarn but tightens the Knex-side lifecycle and error handling.

Key Improvements in the Experimental Integration
- Safe release guard: `releaseConnection` becomes a no-op when given `null`/`undefined` (prevents `__knexUid` crashes; see #6052).
- Destroy fast‑fail: track a destroyed flag and throw a clear "pool destroyed" error on acquire, instead of timing out (see #5628).
- Generic disposal on connection‑level errors: when queries fail with connection errors, mark `connection.__knex__disposed = err` so tarn prunes the resource (helps #5186, #3523, #5192).
- Stronger dialect validation/listeners:
  - MySQL: add `'end'` listener to mark disposed; guard `removeAllListeners()` calls; broaden validate beyond just `connection.state` (helps #3447, #5595).
  - MySQL2: ensure `'error'`/`'end'` listeners mark disposed; validate against `_fatalError`/`_protocolError`/`_closing`/`stream.destroyed`.
  - Postgres: rely on existing `'error'`/`'end'` listeners and add disposal-on-error for query failures.
- Optional (off by default): light pre-acquire liveness heuristics via tarn `validate` (no extra query).
- Better error messages and optional pool stats hooks.

Rollout Plan
- Phase 1 (safe defaults):
[...]

PR #6414, opened 2026-03-27
feat: add connectionPool option for bringing an external pool

### LINKED ISSUE #2665: Use driver's pool instead of knex one
# Environment

Knex version: 0.14.2
Database + version: Oracle 11gR2
OS: linux

@atiertant 

knex recycle connection in a pool so to get efficient resource usage. So far so good. Now with dialect driver's getting rich in functonnality, they offer not only pool management but other feature (proxy authentication) which mix well with it (see https://github.com/oracle/node-oracledb/blob/master/doc/api.md#pool-proxy-authentication). 

See also #2660 for the feature request about proxy authentication.

So, I would propose to delegate pool management to the driver's dialect like so (pool config following):
pool: {
  delegateToDriversDialect: true,
  min: 1
  max: 10
  options: {
   ...
  }
}

In this example, the dialect would translate poolMin/poolMax to the correct driver's property (for oracle poolMin and poolMax). options would be for dialect-specific options, or driver-specific options. 

Would you accept a pr in that way ?

Thanks

tgriesser:
Closes #2665, #6292

Allows bringing a native connection pool from PG / MySQL / Oracle, or providing your own via a new `knex.KnexPool`

This would allow for better interop with other libraries like Kysely since you can just share the same pool instance between both libraries: https://kysely.dev/docs/getting-started#instantiation

Example with Native Pool:

```ts
const { Pool } = require('pg');

const pool = new Pool({
  host: '127.0.0.1',
  user: 'your_database_user',
  password: 'your_database_password',
  database: 'myapp_test',
  max: 20,
});

  client: 'pg',
  connectionPool: pool,
});
```

Example with `knex.KnexPool` / Tarn Compatible Pool:

```ts
const knex = require('knex');
const { Client } = require('pg');

const sharedPool = new knex.KnexPool({
  create: () => {
    const client = new Client({ host: '127.0.0.1', database: 'myapp' });
    return client.connect().then(() => client);
  },
  destroy: (connection) => connection.end(),
  validate: (connection) => !connection._ending,
  min: 0,
  max: 10,
});

// Both instances share the same underlying pool
const db1 = knex({ client: 'pg', connectionPool: sharedPool });
const db2 = knex({ client: 'pg', connectionPool: sharedPool });
```


Nebu1eto:
I've experienced multiple bugs when using Knex and PostgreSQL together due to Knex's internal connection resource management. It's so fantastic, and I hope this PR get merged.

lib/dialects/mysql/index.js
```diff
@@ -109,6 +110,41 @@ class Client_MySQL extends Client {
...
+          if (typeof connection.destroy === 'function') {
+            connection.destroy();
+          }
+          return true;
+        }
+        if (typeof connection.release === 'function') {
+          connection.release();
+          return true;
+        }
+        return false;
```
OlivierCavadenti, line 139: release return here a boolean, but oracledb a promise:  return Promise.resolve(false), the code don't await release, but can be a problem later.

OlivierCavadenti (commented):
add oracledb integration test. 
also 14 lines no covered

PR #6415, opened 2026-03-27
feat: mariadb driver support

erulabs:
Adds support for the mariadb mysql driver: https://github.com/mariadb-corporation/mariadb-connector-nodejs

Addresses https://github.com/knex/knex/issues/5331 and is a prerequisite for https://github.com/knex/knex/issues/5963

I haven't used this yet in production, but all tests pass. I'll take a shot at trialing this with real traffic in the next few days.


tgriesser:
@erulabs let us know when you get a chance to test this out. Integration tests pass with good coverage, I'm good merging it as soon as you think it's ready

erulabs:
@tgriesser We're now running this branch in production with the `mariadb` driver! So far so good!

OlivierCavadenti:
Hi !

so in CI now, 3 mariadb jobs fail (9 integration tests). 
The mariadb dialect extends mysql and isMysql() includes mariadb, so the suite treats MariaDB exactly like MySQL. 

Failures:
knex.fn.uuid()
=> FunctionHelper.uuid() has no mariadb case and throws. MariaDB has UUID(); just add it next to mysql/mysql2.

.using() join
=> the mariadb driver throws on duplicate result columns (checkDuplicate); mysql2 doesn't. Pass checkDuplicate: false in the dialect's connection options.

noWait() & select from subquery
=> MySQL-only error checks. MariaDB returns Lock wait timeout exceeded / errno

with update query, drop checks, json path join
=> genuine MariaDB SQL differences (no CTE on UPDATE, inline check-constraint syntax, JSON compare). Need dialect-specific handling or skip.

OlivierCavadenti (approved):
LGTM !

Issue #6277, opened 2025-09-16
Bindings are not in the correct order when joining a subquery in a delete operation


glejune:
When joining a subquery in a delete operation, the bindings for the "where" come before the bindings for the "join", causing the resulting query not to be what is expected. This could be dangerous because it could cause the wrong data to be deleted.

Knex version: `3.1.0`

Here is the sample code

```
import Knex from "knex"

const knex = Knex({ client: 'mysql' })

async function main() {
    const subQuery = knex.from('tableA').where('fieldA', 'valueForFieldA').as('subQuery')
    const knexQuery = knex.delete()
        .from(knex.raw('tableB'))
        .innerJoin(subQuery, "subQuery.id", "=", "tableB.relatedId")
        .where('tableA.fieldB', 'valueForFieldB')
    console.log(knexQuery.toSQL())

}

main()
    .then(() => knex.destroy())
    .catch(err => {
        console.error(err);
        return knex.destroy();
    });
```

And the resulting query generated (we would expect `valueForFieldA` to come first in the bindings):

```
{
  method: 'del',
  options: {},
  timeout: false,
  cancelOnTimeout: false,
  bindings: [ 'valueForFieldB', 'valueForFieldA' ],
  __knexQueryUid: 'IwfHyZEac_9o2mnqZFVJQ',
  sql: 'delete tableB from tableB inner join (select * from `tableA` where `fieldA` = ?) as `subQuery` on `subQuery`.`id` = `tableB`.`relatedId` where `tableA`.`fieldB` = ?'
}
```

We can see [in the code](https://github.com/knex/knex/blob/01b177c485d696f1b72858dee728ba143c4fad76/lib/query/querycompiler.js#L819) that the `this.where()` call is before the `this.join()` call, this explains why the bindings for the "where" come first in the resulting query.

I think this is related to #873

mercmobily:
Thank you for this excellent bug report! I looked into it and can confirm it's a bug that's been present since DELETE with JOIN was implemented in PR #4568 (July 2021).

The bug is in `lib/query/querycompiler.js` at lines 819-835 in the `del()` method:

```javascript
del() {
  // Make sure tableName is processed by the formatter first.
  const { tableName } = this;
  const withSQL = this.with();
  const wheres = this.where();  // Line 823: Called FIRST
  const joins = this.join();     // Line 824: Called SECOND
  // When using joins, delete the "from" table values as a default
  const deleteSelector = joins ? tableName + ' ' : '';
  return (
[...]

PR #6438, opened 2026-04-28
fix(pg,mssql): preserve binding order in delete and update queries

AbdelrahmanHafez:
re #6277, #6412

Some delete and update compilers built bindings in a different order than the generated SQL placeholders, so values could be matched with the wrong `?` placeholders.

This fixes MSSQL delete binding order in both the standard and trigger-modification paths when joins contribute bindings, MSSQL update binding order when a CTE appears before `update top (?)`, and PostgreSQL delete binding order for joins that are rewritten into `USING` plus `WHERE`, including raw join predicates with their own bindings.

The generated SQL text is unchanged for these cases, this only changes when the side-effectful compiler helpers append bindings.


tgriesser (approved):

tgriesser:
Thanks!

PR #6444, opened 2026-05-01
use ecr for mysql/pg images to reduce docker rate limiting

tgriesser:
Seeing if we can avoid some of the docker rate limiting on our integration tests


tgriesser:
Closing in favor of #6445

PR #6450, opened 2026-05-06
Update homepage urls

giannisp:
I have just noticed that a couple of links point to the old documentation urls, so minor fix here!


tgriesser (approved):

tgriesser:
Thanks!

PR #6456, opened 2026-05-27
micro-optimization in wrappingFormatter

erulabs:
Start of an evidence-based performance optimization campaign. First up,`wrapString`.

- Avoid allocating a one-element array in `columnize()` for the common single-column path
- Replace `value.toLowerCase().indexOf(' as ')` with a no-allocation case-insensitive alias scan
- Avoid `split('.')` array allocation in `wrapString()` and wraps dotted identifiers by scanning in place


erulabs:
Happy to provide my benchmarking setup but here are some results:

| metric | original | new | delta |
| -- | -- | -- | -- |
| latency (median) | ~314 ns | ~185 ns | −41% |
| throughput | 3.19 M ops/s | 5.42 M ops/s | +70% |
| alloc bytes/op | 807.9 B | 363.8 B | −55% |

tgriesser:
@erulabs would definitely be interested to see the benchmarking setup if you had a simple repo or gist. Seems like a reasonable change though

erulabs:
@tgriesser Here ya go: https://gist.github.com/erulabs/f14d34470f148c6e0ed394190f4f5029 👯

tgriesser (approved):

PR #6457, opened 2026-05-29
replace nanoid with crypto.randomUUID

erulabs:
- crypto.randomUUID() was added in Node.js v14.17.0 (now that our minimum nodejs supported version is 16, this is safe to use)
- `nanoid()` is ~2.8% of toSQL CPU
- Perf: ~8x faster (2.7M → 22.4M ops/s on node 22)
- No regression of [#4089](https://github.com/knex/knex/issues/4089): that was a Date-based UUID; randomUUID() is CSPRNG-based, so fake-timer tests are unaffected.
- `nanoid()` is 21 chars, crypto.randomUUID is 36 chars, but I dont see why this would be an issue
- get to delete more code than we add! 💃


lib/raw.js
```diff
@@ -112,7 +112,7 @@ class Raw extends EventEmitter {
       );
     }
 
-    obj.__knexQueryUid = nanoid();
+    obj.__knexQueryUid = randomUUID();
```
rubengmurray, line 115: Looks like a good change to me

lib/util/randomString.js
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Generate a numeric-only random string. This uses the non-secure,
+ * predictable random generator and is intended for low-collision-sensitivity
+ * uses such as generating temporary identifiers.
+ *
+ * @param size Length of the string. The default size is 21.
+ * @returns A random numeric string.
+ */
+function randomNumericString(size = 21) {
```
rubengmurray, line 11: Do we need to replace nanoid here?

erulabs: This is the same function as [used to live in nanoid.js](https://github.com/knex/knex/pull/6457/changes#diff-684f9de15c937d080f52e9be8de15edd0f5bdf6cbeb73e763b90ce71e6ddd2e9L31), just renamed since that file doesnt use nanoid anymore.

rubengmurray (approved):
Looks good to me.

lib/util/randomString.js
rubengmurray: Got you. Misread from me. That's what you get for reviewing on a phone!

PR #6463, opened 2026-06-09
fix(types): #6452 - .where type regression for invalid types

### LINKED ISSUE #6452: Regression in 3.2.0 no longer picking up invalid parameters in `.where({ ... })`
TypeScript version: 6.0.3

With 3.1.0 knex was reasonably safe from bad usage of unknown columns when using `.where({ ... })`

Image

With 3.2.0 and forward this is no longer caught

Image

Minimal-ish repro:

```ts
import { knex } from 'knex';
declare module 'knex/types/tables' {
  export interface Tables {
    accounts: { id: string };
  }
}

const isTest = (await knex({ client: 'pg' })('accounts')
                       // 
        .where({ account_id: id })
        .first());
```

The same applies when using the generic variant eg. `knex<MyInterface>` as well.

tgriesser:
Fixes #6452

### DIFF
--- a/test-tsd/tsd-tests3.test-d.ts
+++ b/test-tsd/tsd-tests3.test-d.ts
@@ -1,5 +1,5 @@
 import { knex, Knex } from '../types';
-import { expectType } from 'tsd';
+import { expectError, expectType } from 'tsd';
 
 const clientConfig = {
   client: 'sqlite3',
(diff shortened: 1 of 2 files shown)


SimonSchick:
I think the same issue also affects other operators such as `groupBy`, `insert` etc. I will create another tracking issue but you may want to investigate.

PR #6471, opened 2026-06-13
Update docs: timeout section

### LINKED ISSUE #6006: Confusion around purpose `timeout(1000, { cancel: false })` in documentation
In the documentation, the behaviour of `cancel: false` is undocumented.

Whilst logically it is obvious - the timeout wouldn't apply? - I don't understand what `false` would achieve over just removing the timeout method completely?

https://knexjs.org/guide/query-builder.html#timeout

Perhaps it would be better to give an example of why `cancel: false` would be useful, rather than giving an example of `cancel: true`, which is just a verbose version of `timeout()`?

Excerpt:

```markdown
.timeout(ms, options={cancel: boolean})

Sets a timeout for the query and will throw a TimeoutError if the timeout is exceeded. The error contains information about the query, bindings, and the timeout that was set. Useful for complex queries that you want to make sure are not taking too long to execute. Optional second argument for passing options:* cancel: if true, cancel query if timeout is reached.

WARNING

Only supported in MySQL and PostgreSQL for now.

knex.select()
  .from('books')
  .timeout(1000)

knex.select()
  .from('books')
  .timeout(1000, { 
    cancel: true // MySQL and PostgreSQL only
  }) 
```

rubengmurray:
- Make the default | `{ cancel: false }` case clearer.
- Attempts to resolve #6006 which I raised a while ago


docs/src/guide/query-builder.md
```diff
@@ -118,31 +121,23 @@ queryBuilder.select('name').then((users) => {
...
 
-::: warning
-Only supported in PostgreSQL for now.
-:::
-
-```js
-knex('users', { only: true }).select('*');
-knex.select('*').from('users', { only: true });
-```
+**.timeout(ms, options={cancel: boolean})**
```
rubengmurray, line 136: Separate non-blocking comment for this PR:

- What governs the `PG+MY only` badge in the `#### timeout` header? I can't see it in the markdown.

Image

That seems incorrect according to the test cases:

https://github.com/rubengmurray/knex/blob/e714ae0df910b0250bdec9f2e2ccb117c7101699/test/integration2/query/misc/additional.spec.js#L769

`TimeoutError` there works for: `Cockroach DB`, `Postgres`, `MySQL`, `MSSQL` & `Oracle`.

docs/src/guide/query-builder.md
```diff
@@ -118,31 +121,23 @@ queryBuilder.select('name').then((users) => {
...
 
-### only (knex/from option) [-MY -SQ -MS -OR -CR -RS]
-
-The `only` option is a second argument you can pass to `knex(tableName, options)` or `.from(tableName, options)`. When `only: true` is set, Knex prefixes the table name with the `ONLY` keyword, which tells PostgreSQL to read rows from the named table only, excluding rows inherited from child tables. This is useful when table inheritance is in use and you want to avoid pulling in data from descendants.
+#### timeout
 
-::: warning
-Only supported in PostgreSQL for now.
-:::
-
```
rubengmurray, line 128: I think the updated text makes it clearer what the default behaviour is. For me this is unexpected. I would expect the timeout to stop query execution to protect the database so I think this is worth being explicit about.

docs/src/guide/query-builder.md
myndzi: I don't know where the code is, but i think `-OR` for example means "except oracle". @mercmobily would know

myndzi (approved):

docs/src/guide/query-builder.md
rubengmurray: Ah I see there are `-*` flags next to the headers

tgriesser: https://github.com/rubengmurray/knex/blob/e714ae0df910b0250bdec9f2e2ccb117c7101699/test/integration2/query/misc/additional.spec.js#L861-L867

Looks like it's only MySQL & PG, the others throw

PR #6497, opened 2026-07-02
Guidance for MySQL related changes

rubengmurray:
The general documentation states that `migrations` run in `transactions`. 

Whilst this is true it may lead the user to believe that everything that runs inside that migration will be rolled back in full if it fails.

In reality, this is dependent on the underlying database engine.

I have experienced this first-hand with MySQL and thought at least the documentation could do with an update.

Others have also mentioned:

https://github.com/knex/knex/issues/4678
https://github.com/knex/knex/issues/5818
https://github.com/knex/knex/issues/1596


docs/src/guide/migrations.md
```diff
@@ -336,6 +336,36 @@ exports.down = function (knex) {
...
+  throw new Error('Some bad code');
+
+  await knex.schema.createTable('my_table_2', (table) => {
+    table.bigInteger('id').unsigned().nullable().alter();
+  });
+};
+```
+
+In this case, the migration will error and won't be classed as complete, but you will be left with `my_table` in your database.
+
```
rubengmurray, line 362: We could advise here:

`For this reason, as best practice, every DDL change should be isolated to it's own file for clean "complete status" and rollback strategy.`

docs/src/guide/migrations.md
```diff
@@ -336,6 +336,36 @@ exports.down = function (knex) {
 exports.config = { transaction: false };
 ```
 
+#### MySQL-specific concerns
+
+If you are making DDL / schema changes to MySQL tables you should be aware that a sudden failure does not rollback any successful schema changes made in the same migration file.
+
+This is because MySQL uses an `Implicit Commit` for many schema related commands changes.
```
rubengmurray:
```suggestion
This is because MySQL uses an `Implicit Commit` for many schema related commands. 
```

rubengmurray:
A step towards helping with these:

https://github.com/knex/knex/issues/4678
https://github.com/knex/knex/issues/5818
https://github.com/knex/knex/issues/1596

PR #6501, opened 2026-07-16
fix(sqlite3): generateDdlCommands returns incomplete plan for drop/add constraint ops

### LINKED ISSUE #6291: `toSQL()` method returns incomplete SQL for SQLite3 DDL operations
# Bug Report: `toSQL()` method returns incomplete SQL for SQLite3 DDL operations

## Bug Description

The `toSQL()` method for SQLite3 schema alterations (column alterations, dropping columns, foreign key operations, etc.) only returns the first SQL statement instead of all the required statements needed to perform the operation. This is problematic because SQLite3 doesn't support direct column alterations and requires a multi-step process involving creating temporary tables, copying data, dropping original tables, and renaming.

## Environment

- **Knex version**: [any version with SQLite3 support]
- **Database + version**: SQLite3
- **OS**: Any

### Example 1: Column Alteration

```javascript
  client: 'sqlite3',
  connection: { filename: ':memory:' }
});

// Create a table
await knex.schema.createTable('users', table => {
  table.increments('id');
  table.string('name', 50);
  table.string('email');
});

// Try to alter a column and get SQL
const alterQuery = knex.schema.alterTable('users', table => {
  table.string('name', 100).alter();
});

const sql = alterQuery.toSQL();
console.log(sql);
// Expected: Multiple SQL statements for the complete SQLite3 DDL process
// Actual: Only returns "PRAGMA table_info(`users`)"
```

### Example 2: Drop Column

```javascript
const dropQuery = knex.schema.alterTable('users', table => {
  table.dropColumn('email');
});

const sql = dropQuery.toSQL();
console.log(sql);
// Expected: All statements needed to drop column in SQLite3
// Actual: Only returns "PRAGMA table_info(`users`)"
```

## Expected Behavior

`toSQL()` should return an array containing all SQL statements that would be executed during the actual operation, including:

1. Pragma statements to check foreign keys
2. Disable foreign key checks (if needed)
3. Create temporary table with new schema
4. Copy data from original to temporary table
5. Drop original table
6. Rename temporary table to original name
7. Recreate indexes
8. Re-enable foreign key checks (if needed)
9. Foreign key constraint validation (if needed)

## Actual Behavior

`toSQL()` only returns the initial PRAGMA statement used to gather table information, missing all the subsequent DDL operations that would actually be executed.

## Root Cause Analysis

The issue lies in the SQLite3 table compiler and DDL implementation:

1. **SQLite3 Table Compiler** (sqlite-tablecompiler.js): Uses `statementsProducer` and `output` functions to generate complex DDL operations, but these are only executed during actual query execution, not during `toSQL()` compilation.

2. **SQLite3 DDL Class** (`ddl.js`): The `generateAlterCommands` method returns a complex object with `pre`, `sql`, `post`, and `check` arrays, but this structure isn't compatible with the standard `toSQL()` format.

3. **Base Table Compiler**: Doesn't have logic to handle SQLite3's special multi-statement DDL requirements when generating SQL for `toSQL()`.

### Solution 1: Modify DDL Class to Support toSQL()

Update the `generateAlterCommands` method in `ddl.js` to include a `toSQL()` method:

```javascript
async generateAlterCommands(newSql, createIndices, columns) {
  // ... existing code ...
  
  const allStatements = [...pre, ...sql, ...(check ? [check] : []), ...post];
  
  return { 
    pre, 
    sql, 
    check, 
    post,
    // Add method for toSQL() compatibility
    toSQL: () => allStatements.map(stmt => ({ sql: stmt, bindings: [] })),
    statements: allStatements
  };
}
```

### Solution 2: Enhance Table Compiler

Add `resolveStatements` methods to the SQLite3 table compiler to properly handle multi-statement operations:

```javascript
this.pushQuery({
  sql: `PRAGMA table_info(${this.tableName()})`,
  statementsProducer(pragma, connection) {
    return compiler.client
      .ddl(compiler, pragma, connection)
      .alterColumn(columnsInfo);
  },
  // Add this method
  async resolveStatements(connection) {
    const pragma = await compiler.client.raw(`PRAGMA table_info(${compiler.tableName()})`).connection(connection);
    const result = await compiler.client
      .ddl(compiler, pragma, connection)
      .alterColumn(columnsInfo);
    
    if (result && typeof result.toSQL === 'function') {
      return result.toSQL();
    } else if (result && result.statements) {
      return result.statements.map(stmt => ({ sql: stmt, bindings: [] }));
    }
    return [];
  }
});
```

### Solution 3: Base TableCompiler Enhancement

Modify the base `TableCompiler` class to check for and use `resolveStatements` when generating SQL for `toSQL()` calls.

## Impact

This bug affects:
- Developers using `toSQL()` for debugging SQLite3 schema operations
- Migration tools that need to preview SQL before execution
- Code generation tools that rely on `toSQL()` output
- Testing frameworks that validate generated SQL

## Workaround

Currently, there's no clean workaround. Developers must execute the actual schema alteration to see what SQL would be generated, which is not ideal for debugging or preview purposes.

nabsei:
Fixes #6291.

`dropColumn`, `dropForeign`, `dropPrimary`, and `primary` (adding a PK) all built their multi-statement rebuild plan via `alter()`, which executes a real transaction immediately, exposed to the table compiler through an `output` hook. Only `statementsProducer` is recognized by `generateDdlCommands()` (and by the runner's generic pre/sql/check/post execution path used for real execution too), so these four operations only ever returned the initial `PRAGMA table_info(...)` when previewed via `generateDdlCommands()`, instead of the full rebuild plan — exactly the bug reported (and, per the maintainer's own comment, suspected to affect "maybe more" than the three named in the title; `primary()` had the identical issue).

**Fix:** switch these four to build the plan via `generateAlterCommands()` — the same preview-only method `alterColumn`/`setNullable`/`foreign` already use — and expose it through `statementsProducer` instead of `output`. The runner already knows how to execute a `statementsProducer`-tagged query generically (run `pre`, wrap `sql` in a transaction, run `check`, run `post`), so real execution keeps working exactly as before, just through the same shared path the other operations already use. The old `alter()` method (which executed a transaction directly) is no longer called by anything and has no other callers, so I left it in place rather than removing it — happy to delete it if you'd prefer.

**Testing:**
- Manually reproduced the bug and confirmed the fix with a script exercising all four operations through both `generateDdlCommands()` (preview) and real execution — preview now returns the full `{pre, sql, check, post}` plan, and real execution still applies correctly (verified column/constraint changes actually take effect).
- Ran the existing sqlite3 schema-builder unit tests (163 passing) and the relevant integration suites (`alter.spec.js`, `misc.spec.js`, `index.spec.js`, `set-nullable.spec.js` — 233 passing total, 0 regressions). Two pre-existing failures in `alter.spec.js` (`Invalid Chai property: rejectedWith`) are unrelated to this change — confirmed identical on a clean `master` checkout (missing `chai-as-promised` registration in this environment).
- Added 4 new tests in `alter.spec.js` asserting `generateDdlCommands()` returns the full multi-statement plan for `dropColumn`, `dropForeign`, `dropPrimary`, and adding a primary key.
- `eslint` clean on all changed files.

Parts of this fix (root-cause tracing through the table compiler/DDL/runner code) were produced with the help of Claude Code — credited via the `Co-authored-by` trailer on the commit. I reviewed and understood the change before opening this PR.


nabsei:
Just checking in on this one — happy to make any changes if something's needed, or let me know if it's not a good fit. No rush either way.

PR #6503, opened 2026-07-19
fix(mysql): #3012 - handle case-insensitive text column values

### LINKED ISSUE #3012: Setting text column type in MySQL is case sensitive
# Environment

Knex version: 0.16.3
Database + version: MySQL 5.7.21
OS: macOS 10.13.6

# Bug

Setting the type of a MySQL TEXT column only works if the type is lower case. MySQL is case insensitive for this kind of thing, so I'd expect knex to do the same.

```javascript
const knex = require('knex')({ client: 'mysql' });

const statement1 = knex.schema.createTable('foo1', (table) => {
  table.text('bar', 'MEDIUMTEXT');
});
console.log(statement1.toSQL());
// [ { sql: 'create table `foo1` (`bar` text)', bindings: [] } ]

const statement2 = knex.schema.createTable('foo2', (table) => {
  table.text('bar', 'mediumtext');
});
console.log(statement2.toSQL());
// [ { sql: 'create table `foo2` (`bar` mediumtext)', bindings: [] } ]
```

Severian0:
Fixes #3012.

MySQL is case-insensitive for SQL keywords and data types, but Knex only recognised lowercase values correctly in the MySQL column compiler. This means that function calls like 

```js
table.text('foo','MEDIUMTEXT');
```

would incorrectly fall back to `text` instead of `mediumtext`.

*Fix*: Added a ternary operator to normalise the `column` input to lowercase:
```js
const normalisedText = column ? column.toLowerCase() : column;
```
and then run the switch statement with `normalizedText` instead of `column`.

*Testing*: Added new regression test in `test\unit\schema-builder\mysql.js` to 'test adding case-insensitive text types'.

### DIFF
--- a/lib/dialects/mysql/schema/mysql-columncompiler.js
+++ b/lib/dialects/mysql/schema/mysql-columncompiler.js
@@ -40,7 +40,8 @@ class ColumnCompiler_MySQL extends ColumnCompiler {
   }
 
   text(column) {
-    switch (column) {
+    const normalizedText = column ? column.toLowerCase() : column;
+    switch (normalizedText) {
       case 'medium':
       case 'mediumtext':
         return 'mediumtext';
--- a/test/unit/schema-builder/mysql.js
+++ b/test/unit/schema-builder/mysql.js
@@ -904,6 +904,20 @@ module.exports = function (dialect) {
       expect(tableSql[0].sql).to.equal('alter table `users` add `foo` text');
     });
 
+    it('test adding case-insensitive text types', function () {
+      tableSql = client
+        .schemaBuilder()
+        .table('users', function () {
+          this.text('foo', 'MEDIUMTEXT');
+        })
+        .toSQL();
+
+      equal(1, tableSql.length);
+      expect(tableSql[0].sql).to.equal(
+        'alter table `users` add `foo` mediumtext'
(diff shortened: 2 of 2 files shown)


marceli1404 (commented):
The fix is on the right track — `getColumnType()` forwards the raw user argument to `text()` via `tail(this.args)`, and MySQL treats these type names case-insensitively, so normalizing before the switch matches DB semantics. `mysql2` inherits this column compiler (`Client_MySQL.columnCompiler()`), so it's covered too.

A few suggestions:

1. **Type-safety**: `column ? column.toLowerCase() : column` throws `TypeError: column.toLowerCase is not a function` if a truthy non-string is passed (e.g. `table.text('foo', 123)`), whereas the old `switch` would fall through to the `default` branch and return `'text'`. Consider `typeof column === 'string' ? column.toLowerCase() : column`.

2. **Test coverage**: the added test only exercises `MEDIUMTEXT`. Worth also adding the no-arg case (`table.text('foo')` → `'text'`) to cover the falsy guard, and a `LONGTEXT` case to cover the second branch of the switch.

3. Minor: `tinytext` still falls through to `'text'` — pre-existing and arguably out of scope for this issue, but worth noting.

Issue #1966, opened 2017-03-13
Empty error


RWOverdijk:
```js
var knex = require('knex')({
  client: 'sqlite3',
  useNullAsDefault: true,
  connection: { filename: "./mydb.sqlite" }
});

knex.transaction(transaction => {
  transaction.rollback().catch(() => {
    console.log('Not called.');
  });
}).catch(() => console.log('This is where it all goes to.. Poop.'));
```

Remove the last catch, and you'll find yourself an unhandled rejection undefined. It took me a long time to track down.

Why does the rollback not fail? What's the purpose of .transaction() being a promise that can reject here?

**Update:** I haven't tracked down the place that actually rejects.

wubzz:
@RWOverdijk The 'catch' handlers argument is `undefined` because `.rollback` was not given an argument such as an Error object it would normally get. Basically issuing a rollback without a given "error" reason. Of course this is ok and works, but the end result is that the error handler receives `undefined`, which is expected behavior at that point.

The rejection happens here: https://github.com/tgriesser/knex/blob/master/src/transaction.js#L125

And the rejection always happens on the "main" promise generated from `knex.transaction((tr) => {})` which is why the outer catch is called, so calling `.catch` on `.rollback` has no effect, . Also, errors are already caught in the query runner: https://github.com/tgriesser/knex/blob/master/src/transaction.js#L117

I'm pretty sure @tgriesser had thoughts on addressing some of these limitations in 1.0, unless I'm mistaken?

RWOverdijk:
@wubzz Indeed, I found that earlier today. It's weird to me, though. I'd expect it to use whatever is returned from within the callback itself. Not actual rollbacks. I've made my peace with the implementation and worked around it.

elhigu:
Mikael Lepistö @elhigu 10:08
@RWOverdijk afaik transaction is started at the same moment when knex.transaction() is called... what exception are you getting when running rollback() I think it should never fail, just silently ignore for example if connection was broken before running it... what do you think that would be the correct functionality here?

Wesley Overdijk @RWOverdijk 10:20
@elhigu I expect no exception at all.
But if I do get an exception, I expect it to not be undefined.

Mikael Lepistö @elhigu 10:21
hmm.. doesn't rollback causes transaction to reject? so exception is undefined because you didn't pass argument to rollback?

Wesley Overdijk @RWOverdijk 10:22
@elhigu No. If you actually do something the rollback doesn't cause an exception.

[...]

PR #6507, opened 2026-07-29
fix(transaction): reject explicit commit() when the COMMIT statement fails (#6141)

### LINKED ISSUE #6141: transaction commit does not throw errors when constraints are deferred
# Environment

Knex version: `knex@^3.1.0`
Database + version: `PostgreSQL 16.2 (Debian 16.2-1.pgdg120+2) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit`
OS: postgres: `16.2-bookworm (docker image)`

# Bug

### Details

We have sort of foreign key deadlock where `user` need to have an `organization_id` and `organization` needs to have `created_by` id. So I have made those columns `DEFERRABLE INITIALLY DEFERRED` to make constraints checks at the end of the transaction.

```sql
ALTER TABLE "users"
ALTER CONSTRAINT users_organizations_id_foreign
  DEFERRABLE INITIALLY DEFERRED;


ALTER TABLE "organizations"
ADD CONSTRAINT "organizations_created_by_foreign"
  FOREIGN KEY ("created_by")
  REFERENCES users(id)
  DEFERRABLE INITIALLY DEFERRED;
```

### Issue

We have a seed  file to add master organization which is a dummy organization to satisfy the master `user`'s `organizations_id` and the master users. Following is the seed file.

```typescript
import { Knex } from 'knex'

export async function seed(knex: Knex): Promise<void> {

  try {
    const [{ id: orgId }] = await trx('organizations')
      .insert([
        {
          name: 'name',
          email: 'CONTACT@EMAIL.COM',
          created_by: -1,
        },
      ])
      .returning('id')

    const [{ id: userId }] = await trx('users')
      .insert([
        {
          email: 'MASTER@EMAIL.COM',
          organizations_id: orgId,
        },
      ])
      .onConflict('email')
      .merge()
      .returning('id')

    // await trx('organizations')
    //   .update({
    //     created_by: userId,
    //   })
    //   .where('id', orliTechOrgId)

    console.log('>>>>>>>>>>>>> commit')
    await trx.commit()
  } catch (err) {
    console.log('>>>>>>>>>>>>> rollback', err)
    await trx.rollback()
  }
}
```

In above, I have commented out the section where `userId` added to the `organization.created_by`. So transaction should fail when committing. However it does not fail but in the db, **organization nor the user was created**. 

Following is the console output:

```shell
❯ NODE_ENV=development npx knex seed:run
Requiring external module ts-node/register
>>>>>>>>>>>>> commit
Ran 13 seed files
```

If I uncomment the lines above, I get exactly the same output but in the DB, a new organization is created. **However,** any other foreign key constraint (for all the other foreign keys I have not set any `deferrable` upon table creation so I assume they are by default `DEFERRABLE INITIALLY IMMEDIATE`) are throwing errors for instance when I pass `-1` to `users.countries_id` which is another foreign key in `users` table throws an error and goes to the catch block.

### Following however, works!

I passed the callback function to `transaction` and this way, it throws the error correctly

```typescript
import { Knex } from 'knex'

export async function seed(knex: Knex): Promise<void> {
  await knex.transaction(async (trx) => {
    const [{ id: orgId }] = await trx('organizations')
      .insert([
        {
          name: 'name',
          email: 'CONTACT@EMAIL.COM',
          created_by: -1,
        },
      ])
      .returning('id')

    const [{ id: mikesId }] = await trx('users')
      .insert([
        {
          email: 'MASTER@EMAIL.COM',
          organizations_id: orgId,
        },
      ])
      .onConflict('email')
      .merge()
      .returning('id')

    // await trx('organizations')
    //   .update({
    //     created_by: mikesId,
    //   })
    //   .where('id', orliTechOrgId)
  })
}
```

```shell
❯ NODE_ENV=development npx knex seed:run
Requiring external module ts-node/register
Error while executing "/home/s1n7ax/Workspace/orlitech-be/db/seeds/05_seed_add_master_users_and_org.ts" seed: COMMIT; - insert or update on table "organizations" violates foreign key constraint "organizations_created_by_foreign"
Error: Error while executing "/home/s1n7ax/Workspace/orlitech-be/db/seeds/05_seed_add_master_users_and_org.ts" seed: COMMIT; - insert or update on table "organizations" violates foreign key constraint "organizations_created_by_foreign"
    at Seeder._waterfallBatch (/home/s1n7ax/Workspace/orlitech-be/node_modules/knex/lib/migrations/seed/Seeder.js:118:23)
error: COMMIT; - insert or update on table "organizations" violates foreign key constraint "organizations_created_by_foreign"
    at Parser.parseErrorMessage (/home/s1n7ax/Workspace/orlitech-be/node_modules/pg-protocol/src/parser.ts:369:69)
    at Parser.handlePacket (/home/s1n7ax/Workspace/orlitech-be/node_modules/pg-protocol/src/parser.ts:188:21)
    at Parser.parse (/home/s1n7ax/Workspace/orlitech-be/node_modules/pg-protocol/src/parser.ts:103:30)
    at Socket.<anonymous> (/home/s1n7ax/Workspace/orlitech-be/node_modules/pg-protocol/src/index.ts:7:48)
    at Socket.emit (node:events:520:28)
    at Socket.emit (node:domain:488:12)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Socket.Readable.push (node:internal/streams/readable:390:5)
    at TCP.onStreamRead (node:internal/stream_base_commons:191:23)
```

- I believe `trx.commit()` should throw errors when transaction commit fails

Develop-KIM:
Fixes #6141.

### Problem

With the manual transaction pattern, a failing `COMMIT` is silently swallowed — the returned promise resolves as if it had succeeded:

```js
await trx('child').insert({ id: 1, parent_id: 999 }); // deferred FK, not yet checked
await trx.commit(); // resolves, even though the COMMIT is rejected by the DB
```

This is most visible with PostgreSQL `DEFERRABLE INITIALLY DEFERRED` constraints, which are only validated at commit time. The callback pattern (`knex.transaction(async trx => { ... })`) already rejects correctly for the same input; only the explicit-commit path is affected.

### Cause

`transactor.commit()` returned the promise from the low-level `query()`. For the commit/rollback queries that promise always *resolves*, and the real outcome is reported only through the transaction's internal execution promise. The callback pattern awaits that internal promise, so it sees the error; the manual caller awaits `commit()`'s return value, so it doesn't.

### Fix

- The internal auto-commit/rollback chain now goes through `_commit` / `_rollback`, preserving its exact current behaviour (so a failed commit in the callback pattern does **not** trigger a spurious rollback / unhandled rejection).
- The public `commit()` still runs the commit and additionally returns the execution promise, so an explicit `await trx.commit()` rejects when the COMMIT fails.

`rollback()` and the callback pattern are unchanged.

### Behaviour change

`commit()` used to resolve with whatever the driver returned for the `COMMIT` statement; it now resolves with `value`, the same thing the transaction promise resolves with. That return value was undocumented and dialect-specific, but callers inspecting it will see different data.

Related: an async container that both `await trx.commit()`s and returns a value now rejects loudly on the second commit instead of swallowing it. That is the intended consequence of `commit()` reporting the real outcome, but it is a behaviour change worth a changelog note.

### Tests

Added an integration test in `set-deferrable-constraint.spec.js` (PostgreSQL/Oracle) that commits a transaction violating a deferred FK and asserts `trx.commit()` rejects. It fails on `master` (`expected promise to be rejected but it was fulfilled with undefined`) and passes with this change.

`DEFERRABLE` only exists on those two dialects, so the explicit-commit happy path is covered separately in `test/integration/execution/transaction.js`, which runs on every dialect: it asserts `await trx.commit()` resolves and the work is persisted. I also ran the unit suite and the full sqlite3 integration suite locally — green, including the existing manual/callback/rollback transaction tests.


marceli1404 (commented):
Nice catch — the fix is correct: `_commit`/`_rollback` are the raw low-level query paths, and making the public `commit()` return `transactor.executionPromise` means `await trx.commit()` will now reject when the `COMMIT` fails. The auto-commit path in `_onAcquire` is correctly switched to `_commit`, keeping the old chained semantics (avoiding a rollback-after-failed-COMMIT), and `executionPromise` is assigned (transaction.js:225) before the container runs (line 232). The connection release is also correct — `_evaluateContainer`'s `finally` awaits `executionPromise`.

A few points worth addressing before this lands:

1. **API change in resolution value**: `commit()` previously resolved with the raw `COMMIT` query result; now it resolves with just the `value` argument. Callers inspecting the return value (rare, but possible) will see different data. Worth documenting in the PR description and considering whether to keep the raw result when the commit succeeds.

2. **Asymmetry with `rollback()`**: `rollback()` still returns the raw query promise rather than `executionPromise`, so `await trx.rollback()` won't reject on a failed `ROLLBACK` the way `commit()` now does. If the intent is symmetric behavior, `rollback()` should be updated too.

3. **Test coverage is dialect-gated**: the new test only runs under `isPostgreSQL || isOracle`. The explicit-commit path exists on sqlite/mysql/mssql too (sqlite-transaction only overrides `_onAcquire`, not commit) — could the test be extended to cover at least one non-gated dialect, or is there a reason the assertion only holds on PG/Oracle?

4. **PG aborted-transaction state**: after a failed `COMMIT`, a PostgreSQL session is left in an aborted state until `ROLLBACK`. The test exercises this implicitly via re-acquire — worth verifying the connection is usable afterward and not leaking a poisoned session from the pool.

5. **Pre-existing double-commit footgun**: with this change, an async container that both `await trx.commit()`s and returns a value will now reject loudly instead of silently swallowing the second commit. That's arguably better, but it's a behavior change that callers may rely on — worth a note in the changelog.

Develop-KIM:
Thanks for reading it that closely. Going through the points:

**1 & 5 — resolution value / double-commit.** Both real, and I've added them to the PR description. `commit()` used to resolve with whatever the driver returned for the `COMMIT` statement, which was never documented and differs per dialect; it now resolves with `value`, the same thing the transaction promise resolves with. I'd rather keep it that way than special-case the success path — the point of the change is that `trx.commit()` and the transaction promise settle identically. Happy to be overruled on that.

**2 — asymmetry with `rollback()`.** This one is deliberate. `executionPromise` *rejects* on rollback (that's how knex signals an aborted transaction), so pointing `rollback()` at it would make `await trx.rollback()` throw the rollback reason — and the usual `try { … } catch { await trx.rollback() }` shape would start blowing up in the catch block. There's an existing test pinning that (`does not reject promise when rolling back a transaction`, and `Rollback without an error should not reject with undefined #1966`). So a failing `ROLLBACK` staying quiet is a separate problem that needs its own signal, not the same promise.

**3 — dialect gating.** `DEFERRABLE` only exists on PG and Oracle, so the *failure* can't be reproduced elsewhere — but the explicit-commit path itself is dialect-independent, and you're right that it wasn't covered. I've added a test in `test/integration/execution/transaction.js`, which runs on every dialect, asserting `await trx.commit()` resolves and the work is persisted. It's a guard on the happy path rather than a repro; the full sqlite3 integration suite is green with it (462 passing).

**4 — PG aborted session.** The test already exercises that: after the rejected commit it runs `knex.schema.table(...).dropForeign(...)`, which goes back through the pool. If the connection came back poisoned that teardown would fail, so a leaked aborted session wouldn't go unnoticed.

Develop-KIM:
I'm winding down my open PRs, so I'm closing this rather than leaving it to go stale. Thanks for taking a look at it.

PR #6508, opened 2026-07-30
fix(postgres,sqlite3): guard columns===true in _merge to emit valid SQL (#6504)

### LINKED ISSUE #6504: onConflict().merge() without column args emits invalid SQL — columns===true guard missing in _merge()
## Bug

Calling `.onConflict().merge()` (no column argument) generates invalid SQL in the PostgreSQL and SQLite3 dialects because `_merge()` lacks the `columns === true` guard that `_ignore()` already has.

## Root cause

`onConflict()` called with no arguments sets the internal `columns` to `true` (the sentinel for "no target"). Both `_ignore()` and `_merge()` receive this value, but only `_ignore()` guards for it:

**`lib/dialects/postgres/query/pg-querycompiler.js` line 154–157**
```js
_ignore(columns) {
  if (columns === true) {          // ← handles the no-arg case
    return ' on conflict do nothing';
  }
  ...
}

_merge(updates, columns, insert) {  // ← no guard
  let sql = ` on conflict ${this._onConflictClause(columns)} do update set `;
  ...
}
```

`_onConflictClause(true)` → `columnize(true)` → `wrap(true)` → `"true"` (a quoted identifier).

## Reproduction

```js
import knex from 'knex';
const pg = knex({ client: 'pg' });

pg('users').insert({ email: 'foo@example.com' }).onConflict().merge().toSQL().sql
// → insert into "users" ("email") values (?) on conflict ("true") do update set "email" = excluded."email"
//                                                       ^^^^^^ should be an error, not a quoted identifier
```

Same issue in `lib/dialects/sqlite3/query/sqlite-querycompiler.js` line ~181.

## Suggested fix

Mirror the guard from `_ignore()` in `_merge()`:

```js
_merge(updates, columns, insert) {
  if (columns === true) {
    throw new Error(
      '.onConflict().merge() requires a conflict target — pass column name(s) to .onConflict(columns)'
    );
  }
  let sql = ` on conflict ${this._onConflictClause(columns)} do update set `;
  ...
}
```

PostgreSQL does not allow `DO UPDATE SET` without a conflict target, so throwing is correct. The SQLite3 dialect can apply the same fix.

marceli1404:
## Description

When onConflict() is called without arguments, _columns is set to 	rue (auto-detect). The _merge() method passed this directly to _onConflictClause(), producing invalid SQL like ON CONFLICT ("true") DO UPDATE SET ....

The _ignore() method already had a columns === true guard, but _merge() was missing it.

## Fix

Added a guard at the start of _merge() in both the **postgres** and **sqlite3** query compilers that resolves auto-detected conflict columns from the insert data when columns === true.

## Tests

- Added unit test for single-row insert without explicit onConflict columns
- Added unit test for multi-row insert without explicit onConflict columns
- All 477 existing unit tests pass

Fixes #6504


marceli1404:
@elhigu @kibertoad — friendly ping on this one. `.onConflict().merge()` without column args previously emitted invalid SQL (`ON CONFLICT ("true") DO UPDATE SET `). The fix resolves conflict columns from the insert data in both the postgres and sqlite3 dialects, and includes 2 regression tests. All 477 unit tests pass. Would appreciate a review!

PR #6514, opened 2026-08-11
fix(types): #6513 - infer type of aliased raw in keyed select

### LINKED ISSUE #6513: QOL Feat: Infer types from typed raw in keyed select
## Problem:

In the following query:

```ts
const result = await knex('table')
  .select({
    alias: db.raw<string>('foo::text')
  })
  .finally()
```

`result[number].alias` is of type any.

## Solution:

Infer the `string` type from the raw in deferred key selection

lazerg:
`select({ alias: knex.raw<string>('foo::text') })` typed the alias as `any`, because `MappedAliasType` only resolved alias values that are column names of the base record. It now falls back to the raw's own result type, so a typed raw keeps that type through the alias. `knex.ref()` in the same position still resolves to `any`, as before.

Fixes #6513

### DIFF
--- a/test-tsd/select.test-d.ts
+++ b/test-tsd/select.test-d.ts
@@ -46,6 +46,12 @@ const main = async () => {
     await knex('users_composite').select([knex.ref('name'), knex.ref('id')])
   );
 
+  expectType<{ identifier: string }[]>(
+    await knex<User>('users').select({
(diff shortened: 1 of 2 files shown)


lazerg:
Hey @tgriesser, could you review this type fix for aliased raw values in a keyed select? It's been sitting for a month.

PR #6530, opened 2026-09-05
Separate class for tidb

finwo:
While working on getting directus running on tidb, ran into some issues like tidb not supporting dropping and adding a primary key in a single query.

That pattern is not coming directly from knex, but since tidb does have quirks not present in mysql/mariadb (like no generally available foreign keys <8.5.0), it's worth starting to track it separately.

If this is not the place, feel free to close this PR


finwo:
Closing, just read https://github.com/knex/knex/blob/master/CONTRIBUTING.md#i-would-like-to-add-support-for-new-dialect-to-knex-is-it-possible

PR #6538, opened 2026-09-20
fix(sqlite): do not split quoted table names containing dots (#2762)

### LINKED ISSUE #2762: Knex does not play nice with dot-notation for table names (sqlite)
# Environment

Knex version: `v0.15.2`
Database + version: `sqlite3`
OS: MacOS and Linux

# Bug

  Knex splits table name variables on period characters, interpretting them as dot-notated identifiers (despite being used explicitly as a `tableName` variable). I've attempted various ways of escaping the entire table name, but it seems that those efforts are disregarded by a basic `split` on `/\./g`. Even when using a custom `wrapIdentifier` the dot-delimited words get split first. I couldn't find any documented way to keep the table name as a literal string. Doing this in sqlite directly is simple:

```sql
CREATE TABLE "my.table.name" (
  id int primary key,
  data text
);
```

Doing this the knex way fails:
```js
knex.schema.createTable(`"my.table.name"`, table => {
  table.integer('id')
  table.text('data')
  // etc...
})
```

I've attempted to pass in `tableName` wrapped in double-quotes, but that does nothing.

2. Error message

```
'create table `"my`.`table`.`name"` (`id` int primary key, `data` text) - SQLITE_ERROR: near ".": syntax error'
```

3. Reduced test code:
https://runkit.com/embed/emvu0aryey1m

dyk1454683243-sudo:
Fixes #2762.

SQLite allows `.` in a table name when the identifier is quoted. Knex always split identifiers on `.` in `wrapString`, so `"my.table.name"` was wrapped as three segments (`"my`, `table`, `name"`) and produced invalid SQL.

This change treats a fully quoted identifier as a single name in the SQLite dialect (`sqlite3` / `better-sqlite3`). Unquoted dotted names such as `schema.table` are unchanged.

## Reproduction

```js
knex.schema.createTable('"my.table.name"', (table) => {
  table.integer('id');
  table.text('data');
});
```

Compiled SQL after this fix:

```
create table `my.table.name` (`id` integer, `data` text)
```

## Tests

- Unit: formatter wrapping and SQLite schema builder SQL
- Integration: create / insert / select / hasTable against sqlite3 and better-sqlite3


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

